Issue #5, opened 2021-04-16
Enhance response time by re-using VMs between queries


hoh:
The response time of code running in VMs would be largely decreased if the same VM was reused between calls.

Achieving this could be done by:
- Keeping a VM for a certain time after the last response (timeout)
- Forwarding successive requests to the same VM
- Turning off the VM after a specific timeout (fixed or dynamic)

Host performance might be enhanced further with using the [snapshot feature of Firecracker](https://github.com/firecracker-microvm/firecracker/blob/main/docs/snapshotting/versioning.md)

PR #15, opened 2021-05-17
Feature: Keep a VM ready to accept another call

hoh:
After every execution, a VM is kept running for in case the same function gets called again.
The VM is stopped if no call is made during a configurable duration.


hoh:
Fixes #5

PR #74, opened 2021-07-25
[doc] fix caddy configuration port

aliel:


hoh:
Aleph VM uses port 4020 as the main endpoint. Port 4021 is used for the internal component `vm_connector`, and should not be exposed publicly.

PR #121, opened 2022-01-29
Update README.md

egk10:
Make a exclusive guide to install and run a CNR


hoh:
The README has been refactored.

PR #138, opened 2022-02-10
Fix: forget_vm called twice when VM init times out

hoh:
Solution: Allow execution to be already missing.

See Sentry error https://sentry.io/organizations/alephim/issues/3001246903/?project=6193637&query=is%3Aunresolved

### DIFF
--- a/vm_supervisor/pool.py
+++ b/vm_supervisor/pool.py
@@ -47,7 +47,10 @@ class VmPool:
             return None
 
     def forget_vm(self, vm_hash: VmHash) -> None:
-        self.executions.pop(vm_hash)
+        try:
+            del self.executions[vm_hash]
+        except KeyError:
+            pass
 
     async def stop(self):
         """Stop all VMs in the pool."""


vm_supervisor/pool.py
```diff
@@ -47,7 +47,7 @@ async def get_running_vm(self, vm_hash: VmHash) -> Optional[VmExecution]:
             return None
 
     def forget_vm(self, vm_hash: VmHash) -> None:
-        self.executions.pop(vm_hash)
+        self.executions.pop(vm_hash, None)
```
odesenfans, line 50:
```
try:
    del executions[vm_hash]
except KeyError:
    pass
```

would convey a better meaning IMO since you don't use the returned value, but your solution works as well.

odesenfans (commented):
Nitpick.

vm_supervisor/pool.py
hoh: More obvious, sounds good.

PR #143, opened 2022-02-14
Fix: VM resource seconds field was ignored

### LINKED ISSUE #139: Use VM timeout as the lowest between settings and VM duration

hoh:
Solution: Use the minimum between the VM resource 'seconds' and the executor's timeout as the reuse timeout of the VM.

Fixes #139


hoh:
@moshemalawach would this impact some existing VMs ?

moshemalawach:
I am not sure it would impact VMs, depends on the default value used so far I guess

hoh:
I was mistaken in the behaviour, the two fields mean something different:
- program.resources.seconds = timeout per request
- settings.REUSE_TIMEOUT = timeout without request

The change will should no affect currently running VMs that I know of, but I will send a warning in the dev channel anyway.

PR #144, opened 2022-02-14
Fix: VM resource seconds field was ignored

### LINKED ISSUE #139: Use VM timeout as the lowest between settings and VM duration

hoh:
Solution: Timeout the request after the number of seconds has passed.

Fixes #139


odesenfans (approved):
Looks alright :+1:

PR #150, opened 2022-03-22
Feature: Record VM execution times and log details on files.

hoh:
The resources used by a VM during its execution are recorded in an SQLite database.

The detailed information about VM executions can optionally be stored in JSON files on disk.

### DIFF
--- a/packaging/aleph-vm/DEBIAN/control
+++ b/packaging/aleph-vm/DEBIAN/control
@@ -3,6 +3,6 @@ Version: 0.1.8
 Architecture: all
 Maintainer: Aleph.im
 Description: Aleph.im VM execution engine
-Depends: python3,python3-pip,python3-aiohttp,python3-msgpack,python3-aiodns,python3-setproctitle,redis,python3-aioredis,python3-psutil,sudo,acl,curl,systemd-container,squashfs-tools,debootstrap
+Depends: python3,python3-pip,python3-aiohttp,python3-msgpack,python3-aiodns,python3-sqlalchemy,python3-setproctitle,redis,python3-aioredis,python3-psutil,sudo,acl,curl,systemd-container,squashfs-tools,debootstrap
 Section: aleph-im
 Priority: Extra
--- a/vm_supervisor/conf.py
+++ b/vm_supervisor/conf.py
@@ -95,6 +95,11 @@ class Settings(BaseSettings):
     RUNTIME_CACHE = FilePath(join(CACHE_ROOT, "runtime"))
     DATA_CACHE = FilePath(join(CACHE_ROOT, "data"))
 
+    EXECUTION_ROOT: FilePath = FilePath("/var/lib/aleph/vm")
+    EXECUTION_DATABASE: FilePath = FilePath(join(EXECUTION_ROOT, "executions.sqlite3"))
+    EXECUTION_LOG_ENABLED: bool = False
+    EXECUTION_LOG_DIRECTORY: FilePath = FilePath(join(EXECUTION_ROOT, "executions"))
+
     PERSISTENT_VOLUMES_DIR = FilePath(
         join("/var/lib/aleph/vm/volumes", "volumes", "persistent")
     )
--- a/vm_supervisor/conf.py
+++ b/vm_supervisor/conf.py
@@ -95,10 +95,10 @@ class Settings(BaseSettings):
     RUNTIME_CACHE = FilePath(join(CACHE_ROOT, "runtime"))
     DATA_CACHE = FilePath(join(CACHE_ROOT, "data"))
 
-    EXECUTION_ROOT: FilePath = FilePath("/var/lib/aleph/vm")
-    EXECUTION_DATABASE: FilePath = FilePath(join(EXECUTION_ROOT, "executions.sqlite3"))
-    EXECUTION_LOG_ENABLED: bool = False
-    EXECUTION_LOG_DIRECTORY: FilePath = FilePath(join(EXECUTION_ROOT, "executions"))
+    EXECUTION_ROOT = FilePath("/var/lib/aleph/vm")
+    EXECUTION_DATABASE = FilePath(join(EXECUTION_ROOT, "executions.sqlite3"))
+    EXECUTION_LOG_ENABLED = False
+    EXECUTION_LOG_DIRECTORY = FilePath(join(EXECUTION_ROOT, "executions"))
 
     PERSISTENT_VOLUMES_DIR = FilePath(
         join("/var/lib/aleph/vm/volumes", "volumes", "persistent")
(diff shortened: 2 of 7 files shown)


vm_supervisor/conf.py
```diff
@@ -95,6 +95,11 @@ class Settings(BaseSettings):
     RUNTIME_CACHE: FilePath = FilePath(join(CACHE_ROOT, "runtime"))
     DATA_CACHE: FilePath = FilePath(join(CACHE_ROOT, "data"))
 
+    EXECUTION_ROOT: FilePath = FilePath("/var/lib/aleph/vm")
```
odesenfans, line 98: You can remove the type annotations, they're obvious in this case.

vm_supervisor/metrics.py
```diff
@@ -0,0 +1,99 @@
...
+    engine = create_engine(f"sqlite:///{settings.EXECUTION_DATABASE}", echo=True)
+    Base.metadata.create_all(engine)
+    session_maker = sessionmaker(bind=engine)
+    return session_maker
+
+
+class ExecutionRecord(Base):
+    __tablename__ = "records"
+
+    uuid = Column(String, primary_key=True)
```
odesenfans, line 36: Add `nullable=False` to all columns that cannot be null. The default is `True` in SqlAlchemy iirc.

vm_supervisor/metrics.py
```diff
@@ -0,0 +1,99 @@
...
+    os.makedirs(settings.EXECUTION_LOG_DIRECTORY, exist_ok=True)
+    filepath = join(settings.EXECUTION_LOG_DIRECTORY, f"{execution_uuid}.json")
+    with open(filepath, "w") as fd:
+        fd.write(execution_data)
+
+
+async def save_record(record: ExecutionRecord):
+    """Record the resource usage in database"""
+    sessionmaker = get_database_sessionmaker()
+    session = sessionmaker()
```
odesenfans, line 90: It's better to use a context manager:

```
with sessionmaker() as session:
   ...
```

That way you don't need to call `session.close()`, which you forgot. If I remember correctly, not closing sessions ends up drying up the connection pool of SQLAlchemy, not that it should matter with SQLite.

vm_supervisor/metrics.py
```diff
@@ -0,0 +1,99 @@
...
+from sqlalchemy.ext.declarative import declarative_base
+from sqlalchemy.orm import sessionmaker
+
+from .conf import settings
+
+logger = logging.getLogger(__name__)
+
+
+Base = declarative_base()
+session_maker: Optional[sessionmaker] = None
```
odesenfans, line 18: The idiomatic name for this is `Session` in the docs. You can also create it directly from this level, I don't see the benefit of `get_database_sessionmaker`.

odesenfans: More info: https://docs.sqlalchemy.org/en/13/orm/session_basics.html#what-does-the-session-do

packaging/aleph-vm/DEBIAN/control
```diff
@@ -3,6 +3,6 @@ Version: 0.1.8
 Architecture: all
 Maintainer: Aleph.im
 Description: Aleph.im VM execution engine
-Depends: python3,python3-pip,python3-aiohttp,python3-msgpack,python3-aiodns,python3-setproctitle,redis,python3-aioredis,python3-psutil,sudo,acl,curl,systemd-container,squashfs-tools,debootstrap
+Depends: python3,python3-pip,python3-aiohttp,python3-msgpack,python3-aiodns,python3-sqlalchemy,python3-setproctitle,redis,python3-aioredis,python3-psutil,sudo,acl,curl,systemd-container,squashfs-tools,debootstrap
```
odesenfans, line 6: It would be better if you could fix the version of packages in a requirements.txt file or something. SQLAlchemy is moving towards v2.0 at the moment, there are already massive differences between 1.3 and 1.4 (the current version) so I'm pretty confident that we will have compatibility issues at some point.

vm_supervisor/metrics.py
```diff
@@ -0,0 +1,99 @@
...
+
+
+def get_database_sessionmaker() -> sessionmaker:
+    global session_maker
+    if session_maker:
+        return session_maker
+
+    # engine = create_engine('sqlite:///:memory:', echo=True)
```
odesenfans, line 26: Remove commented code.

vm_supervisor/metrics.py
```diff
@@ -0,0 +1,99 @@
...
+    engine = create_engine(f"sqlite:///{settings.EXECUTION_DATABASE}", echo=True)
+    Base.metadata.create_all(engine)
```
odesenfans, line 28: While this works, I'd create the tables from main instead for consistency and repeatability.

vm_supervisor/metrics.py
```diff
@@ -0,0 +1,99 @@
...
+            "cpu_time_user": self.cpu_time_user,
+            "cpu_time_system": self.cpu_time_system,
+            "io_read_count": self.io_read_count,
+            "io_write_count": self.io_write_count,
+            "io_read_bytes": self.io_read_bytes,
+            "io_write_bytes": self.io_write_bytes,
+            "vcpus": self.vcpus,
+            "memory": self.memory,
+            "network_tap": self.network_tap,
+        }
```
odesenfans, line 76: For reference, you could implement a generic `to_dict` method like this:

```
return {c.name: getattr(self, c.name) for c in self.__table__.c}
```

vm_supervisor/metrics.py
odesenfans: Also, note that the typical pattern is to create the session at the highest possible level and pass it as argument to functions so that you can use the same session in multiple functions. I don't know if that's strictly necessary here though.

vm_supervisor/supervisor.py
```diff
@@ -18,7 +18,7 @@
     about_login,
     about_executions,
     about_config,
-    status_check_fastapi,
+    status_check_fastapi, about_execution_records,
```
odesenfans, line 21: Run black.

odesenfans (changes requested):
A few things to improve in the usage of SQLAlchemy, looks good and clean otherwise.

vm_supervisor/metrics.py
hoh: Is https://github.com/aleph-im/aleph-vm/pull/150/commits/6ab03aae7849eb1f252edec1221b4c096c1c63ba solving this ?

hoh (commented):
Please see if all comments are addressed in https://github.com/aleph-im/aleph-vm/pull/150/commits/6ab03aae7849eb1f252edec1221b4c096c1c63ba so I can squash it.

vm_supervisor/metrics.py
odesenfans: Yes, with a few notes regarding the instantiation of the engine and Session objects.

PR #153, opened 2022-03-24
Problem: Developers had no easy way to know which Python dependencies to use

hoh:
Solution: Extract Python dependencies from Debian and Ubuntu container images.


odesenfans (approved):
So this just freezes the Python requirements to give additional information to the dev? Seems OK to me, too bad that we cannot control versions properly using this install method.

PR #168, opened 2022-04-26
Fix: Users could not tell if redacted fields were empty

hoh:


vm_supervisor/conf.py
```diff
@@ -192,7 +192,7 @@ def display(self) -> str:
         annotations = self.__annotations__.copy()
 
         for attr in annotations.keys():
-            if attr in self.SENSITIVE_FIELDS:
+            if getattr(self, attr) and attr in self.SENSITIVE_FIELDS:
```
odesenfans, line 195: I'm not sure I get the point here. Isn't the point of redacting the attributes that users should have no information whatsoever on the value, even if it is empty?

odesenfans (commented):
The code is okay, I just don't understand why this is better than the previous implementation.

PR #169, opened 2022-04-26
Fix: Process metrics crashed if process was already stopped

hoh:


odesenfans (commented):
Small comment on the log.

odesenfans (approved):

PR #173, opened 2022-04-26
Fix: Database access during benchmarks errored

hoh:
Solution: Initiate the database in the benchmark procedure.

An alternative would be to make the database optional and ignore it in benchmarks.


odesenfans (approved):
Seems alright to me. I think it's better to run the benchmarks with the DB so that we stay close to the behavior in production.

PR #186, opened 2022-05-11
Allow to run CI in a Droplet for nested virtualization

hoh:
The first step to run end-to-end integration tests of the project on a platform that supports nested virtualization.

Instead of attempting to run the self-hosted GitHub Actions runner on a machine in a stateful manner, this creates a new system and uses SSH to configure it.

Choices to discuss in the future:
- Create on demand and destroy or reset and keep running for demo ?
- Configure DNS to have a domain for HTTPS end-user testing ?


.github/workflows/test-on-droplet.yml
```diff
@@ -0,0 +1,74 @@
...
+          --ssh-keys 18:09:36:58:79:44:bb:84:45:c8:6f:9a:f6:b8:0a:c5
+
+      - name: Build Debian Package
+        run: |
+          cd packaging && make all-podman-debian-11 && cd ..
+          ls packaging/target
+
+#      - name: Wait for the system to setup and boot
+#        run: sleep 60
+# Disabled since the build takes approximately as much time
```
odesenfans, line 47: Anyway, it would be better to have a script that checks for connectivity on the VM, for example by verifying the VM is reachable every X seconds. Out of context for this PR, just saying.

odesenfans (approved):
Minor comments, looks good otherwise. Cool CI/CD feature!

.github/workflows/test-on-droplet.yml
hoh: Thanks for the remark, updated.

PR #198, opened 2022-06-22
Fix: Recording VM usage crashed if process had stopped beforehand

hoh:
Not the cleanest code due to duplication, but the most simple I could think of.

### DIFF
--- a/vm_supervisor/models.py
+++ b/vm_supervisor/models.py
@@ -191,25 +191,49 @@ class VmExecution:
                 execution_uuid=self.uuid, execution_data=self.to_json()
             )
         pid_info = self.vm.to_dict()
-        await save_record(
-            ExecutionRecord(
-                uuid=str(self.uuid),
-                vm_hash=self.vm_hash,
-                time_defined=self.times.defined_at,
-                time_prepared=self.times.prepared_at,
-                time_started=self.times.started_at,
-                time_stopping=self.times.stopping_at,
-                cpu_time_user=pid_info["process"]["cpu_times"].user,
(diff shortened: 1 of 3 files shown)


vm_supervisor/models.py
```diff
@@ -191,25 +191,55 @@ async def record_usage(self):
...
-                memory=self.vm.hardware_resources.memory,
-                network_tap=self.vm.fvm.network_tap,
+        # Traceback (most recent call last):
+        #   File "/root/aleph-vm/vm_supervisor/models.py", line 141, in expire
+        #     await self.stop()
+        #   File "/root/aleph-vm/vm_supervisor/models.py", line 156, in stop
+        #     await self.record_usage()
+        #   File "/root/aleph-vm/vm_supervisor/models.py", line 202, in record_usage
+        #     cpu_time_user=pid_info["process"]["cpu_times"].user,
+        # TypeError: 'NoneType' object is not subscriptable
```
odesenfans, line 201: Do you want to keep the traceback in the code?

vm_supervisor/vm/firecracker_microvm.py
```diff
@@ -243,7 +243,7 @@ def to_dict(self):
                     "num_ctx_switches": p.num_ctx_switches(),
                 }
             except psutil.NoSuchProcess:
-                logger.warning("Cannot read process metrics (process not found)", exc_info=True)
+                logger.warning("Cannot read process metrics (process not found)", exc_info=False)
```
odesenfans, line 246: That's the default value, right? Why keep it?

PR #200, opened 2022-06-22
Fix: Settings with implicit annotations were not displayed

hoh:
`self.__annotations__` did not include attributes with implicit type annotation (the type of the default value), and those were therefore not displayed when using `--print-settings`


odesenfans (approved):
Seems alright.

PR #202, opened 2022-06-23
Chore: Update kernel to 5.10

hoh:
Firecracker supports Linux kernels 4.20 and 5.10.

This updates the Linux kernel used to 5.10.

The configuration is based on the Firecracker config [1]
on which `make menuconfig` is run to disable the following
fields: CONFIG_INPUT_KEYBOARD CONFIG_INPUT_MISC CONFIG_INPUT_FF_MEMLESS
and CONFIG_SERIO.

[1] https://github.com/firecracker-microvm/firecracker/blob/0fa080b137fd29e5bcd95073473b0a57c3868d86/resources/guest_configs/microvm-kernel-x86_64-5.10.config


hoh:
Builds are expected to fail since the new kernel can only be updated on a release. 

It has been tested as passing when copying the binary in the commit: https://github.com/aleph-im/aleph-vm/actions/runs/2548185491

kernels/build-kernel.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+set -euf -o pipefail
+
+curl -OL "https://cdn.kernel.org/pub/linux/kernel/v5.x/linux-5.10.124.tar.xz"
+curl -OL "https://cdn.kernel.org/pub/linux/kernel/v5.x/linux-5.10.124.tar.sign"
+unxz linux-5.10.124.tar.xz
+
+#gpg --verify linux-5.10.124.tar.sign linux-5.10.124.tar
```
odesenfans, line 8: Why is this line commented? We always want to check the signature IMO.

odesenfans (commented):
Why is the gpg signature check commented?

Okay for the rest, I didn't check the kernel config.

PR #204, opened 2022-07-04
Fix: Pubsub queues were not discarded after use

hoh:


vm_supervisor/pubsub.py
```diff
@@ -25,10 +25,25 @@ async def msubscribe(self, *keys):
...
+
+        # Cleanup: remove the queue from the subscribers
+        for key in keys:
+            for subscriber in self.subscribers.values():
+                subscriber.discard(queue)
+                # Remove keys with no remaining queue
+                if not self.subscribers.get(key):
+                    self.subscribers.pop(key)
+
+        return
```
odesenfans, line 46: Don't you need to implement the same logic in `subscribe`? You add an entry to `subscribers` there as well.

odesenfans (approved):

PR #209, opened 2022-07-04
Chore: Cleanup internal logging

hoh:


vm_supervisor/tasks.py
```diff
@@ -96,7 +95,6 @@ async def watch_for_messages(dispatcher: PubSub, reactor: Reactor):
     )
 
     async for message in retry_generator(subscribe_via_ws(url)):
-        logger.info(f"Websocket received message: {message.item_hash}")
```
odesenfans, line 99: Not that I'm against it, but are you sure we're not gonna miss these logs some day?

PR #212, opened 2022-07-14
Fix: Stability issues when checking crash handling in monitored endpoint

hoh:
## Problem:
The endpoint '/status/check/fastapi' is monitored by multiple operators
and therefore called very frequently. The crash handling check results
in a new virtual machine being started at every call, and memory leaks
(likely pipes accumulating) can cause the supervisor to crash after
some time with 'Too many open files'.

## Solution:
In the short term, avoid calling the crash check in the monitored endpoint.
In the longer term, investigate and fix the accumulating file descriptors.


odesenfans (approved):
PR too big, please try to split it. JK, approved!

PR #215, opened 2022-07-29
Fix: Watching for updates could be started multiple times

hoh:
This would have been the source of many duplicate asyncio tasks.


odesenfans (approved):
Looks alright.

PR #219, opened 2022-08-06
Update README.md

mrhouzlane:
option for macOs users

### DIFF
--- a/tutorials/README.md
+++ b/tutorials/README.md
@@ -36,12 +36,34 @@ Some cryptographic functionalities of Aleph use curve secp256k1 and require inst
 Archiving programs and volumes requires
 [Squashfs user space tools](https://github.com/plougher/squashfs-tools).
 
-    sudo apt-get install -y python3-pip libsecp256k1-dev squashfs-tools
+- Linux : 
+
+``` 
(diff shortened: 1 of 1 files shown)


tutorials/README.md
```diff
@@ -36,12 +36,34 @@ Some cryptographic functionalities of Aleph use curve secp256k1 and require inst
...
 and the [Python Aleph client](https://github.com/aleph-im/aleph-client) for it's command-line tools:
 
-    pip3 install uvicorn[standard] aleph-client fastapi eth_account
+- Linux : 
+ 
+``` 
+pip3 install uvicorn[standard] aleph-client fastapi eth_account
+``` 
+
+- macOs :
```
hoh, line 62: Can you explain why different instructions for Linux and macOS when using pip3 ?

MHHukiewitz: It works with quotes for Linux as well

hoh: Then I do not understand why having different sections for Linux and a macOS, when they could be identical.

mrhouzlane: should be one section but with quotes , that's the update

hoh: These quotes also work on Linux. Can you update the document to use them for both OS ?

PR #227, opened 2022-09-13
Fix: Minor inconsistencies and improve logging

hoh:


vm_supervisor/vm/firecracker_microvm.py
```diff
@@ -148,7 +148,7 @@ async def download_code(self):
             self.code_path = await get_code_path(code_ref)
         except ClientResponseError as error:
             raise ResourceDownloadError(error)
-        assert isfile(self.code_path)
+        assert isfile(self.code_path), f"Code not found on '{self.code_path}'"
```
odesenfans, line 151: Could be changed with pathlib, but ok.

odesenfans (approved):

vm_supervisor/vm/firecracker_microvm.py
hoh: Yes, removing the `FilePath` type everywhere should be the topic of another branch.

PR #229, opened 2022-09-13
Fix: Exceptions in coroutines were not logged

hoh:
The traceback for these exceptions did not appear anywhere.

This adds helpers to run coroutines in tasks with exception catching and logging and helps with spotting errors in coroutines.

It also registers an exception handler on the event loop, and a CLI option to enable the debug mode of Asyncio.


hoh:
I removed the ineffective exception handler and created a separate issue for that topic.

odesenfans (approved):

PR #230, opened 2022-09-13
Fix: Internal PubSub `msubscribe` failed

hoh:
Mutating the object being iterated caused inconsistency when iterating over subscribers.

### DIFF
--- a/vm_supervisor/pubsub.py
+++ b/vm_supervisor/pubsub.py
@@ -11,7 +11,7 @@ logger = logging.getLogger(__name__)
 
 
 class PubSub:
-    subscribers: Dict[Hashable, Set[asyncio.Queue]]
+    subscribers: Dict[Hashable, Set[asyncio.Queue[set]]]
 
     def __init__(self):
         self.subscribers = {}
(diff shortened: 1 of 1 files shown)


vm_supervisor/pubsub.py
```diff
@@ -45,12 +45,11 @@ async def msubscribe(self, *keys):
 
         # Cleanup: remove the queue from the subscribers
         for key in keys:
-            for subscriber in self.subscribers.values():
+            for subscriber in list(self.subscribers.values()):
                 subscriber.discard(queue)
                 # Remove keys with no remaining queue
-                if not self.subscribers.get(key):
+                if self.subscribers.get(key) == set():
```
odesenfans, line 51: Mmh this could be dangerous if we ever use a different type than set. Maybe just check `if not self.subscribers:` ?

odesenfans (commented):
One minor comment, seems good otherwise.

vm_supervisor/pubsub.py
hoh: As discussed, I did this instead (and an additional comment):

image

odesenfans (approved):

PR #235, opened 2022-09-28
Internal: DB migration system

odesenfans:
Problem: some updates require to modify the DB models.
Solution: add a migration system based on alembic (the SQLAlchemy migration system).

Developers can now create migrations by running
`alembic -c vm_supervisor/alembic.ini revision -m "<description>" --autogenerate`.

The migrations are automatically run at startup time.


hoh:
The dependencies in requirements.txt files are frozen from official Debian / Ubuntu packages and should match the versions available in the distributions:

https://packages.debian.org/bullseye/alembic
https://packages.debian.org/bullseye/py-cpuinfo

If these versions cause an issue (missing important features), we can investigate their dependency trees to see if we should ship that package inside our .deb packages.

odesenfans:
Ok I'll have a look after Chicago, unless you need the feature asap.

MHHukiewitz:
Requires conflict resolution. Has this any priority currently? Has someone experienced troubles recently because of this problem?

vm_supervisor/migrations/env.py
```diff
@@ -0,0 +1,70 @@
...
+# ... etc.
+
+
+def run_migrations_offline() -> None:
+    """Run migrations in 'offline' mode.
+
+    This configures the context with just a URL
+    and not an Engine, though an Engine is acceptable
+    here as well.  By skipping the Engine creation
+    we don't even need a DBAPI to be available.
```
hoh, line 32: This is very low level, how can a person new to the project understand the difference between offline and online ?

vm_supervisor/migrations/env.py
```diff
@@ -0,0 +1,70 @@
...
+
+    Calls to context.execute() here emit the given string to the
+    script output.
+
+    """
+    url = make_db_url()
+    context.configure(
```
hoh, line 39: New to Alembic. Is this call purely stateful ?

vm_supervisor/migrations/script.py.mako
```diff
@@ -0,0 +1,24 @@
+"""${message}
```
hoh, line 1: The `mako` template engine should appear in the frozen `packaging/requirements*` files.

vm_supervisor/alembic.ini
```diff
@@ -0,0 +1,102 @@
...
+
+[handler_console]
+class = StreamHandler
+args = (sys.stderr,)
+level = NOTSET
+formatter = generic
+
+[formatter_generic]
+format = %(levelname)-5.5s [%(name)s] %(message)s
+datefmt = %H:%M:%S
```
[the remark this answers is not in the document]
odesenfans: This file is automatically generated, these are the defaults.

vm_supervisor/migrations/env.py
odesenfans: RTFM :smile: This file is auto-generated by alembic. Basically online mode runs your Python migration scripts directly, whereas in offline mode you use alembic to generate SQL scripts and apply that however you like to your DB. We're only ever gonna use `online` mode in our case.

vm_supervisor/__main__.py
```diff
@@ -203,6 +204,12 @@ async def fake_read() -> bytes:
     print("Event result", result)
 
 
+def run_db_migrations():
+    from vm_supervisor.conf import make_db_url
```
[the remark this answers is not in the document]
odesenfans: There were some import issues when I tried to configured this, I don't remember exactly which ones. Let me check.

vm_supervisor/migrations/env.py
odesenfans: Yes I think so. Don't fret about the details too much here, most of this file is automatically generated.

vm_supervisor/migrations/script.py.mako
odesenfans: I'll update the requirements.

vm_supervisor/__main__.py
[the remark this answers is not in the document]
hoh: Any update on this ?

vm_supervisor/metrics.py
```diff
@@ -10,7 +10,7 @@
 from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy.orm import sessionmaker
 
-from .conf import settings
+from vm_supervisor.conf import make_db_url, settings
```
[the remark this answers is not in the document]
odesenfans: I don't know why I thought that was necessary back then, reverted.

packaging/aleph-vm/DEBIAN/control
```diff
@@ -1,5 +1,5 @@
 Package: aleph-vm
-Version: 0.1.8
+Version: 0.2.5-26-g358c9ef
```
hoh, line 2: Why commit this ?

odesenfans: My bad, I'll revert it. It's a bit annoying that it can even be committed at all.

vm_supervisor/conf.py
```diff
@@ -221,5 +221,9 @@ class Config:
         env_file = ".env"
 
 
+def make_db_url():
+    return f"sqlite:///{settings.EXECUTION_DATABASE}"
```
hoh, line 225: Would it make more sense to put this as a `@property` of the settings ?

```python
class Settings(BaseSettings):
    @property
    def execution_database_url(self):
        eturn f"sqlite:///{self.EXECUTION_DATABASE}"

odesenfans: Doesn't make a lot of difference, maybe the advantage is that the caller does not have to get the singleton himself.

hoh:
Replaced with https://github.com/aleph-im/aleph-vm/pull/304 since this Pull Requests uses a fork.

PR #240, opened 2022-10-05
Cleanup: Code cleanup

hoh:
Cleanup multiple warnings without affecting the behaviour.

### DIFF
--- a/guest_api/__main__.py
+++ b/guest_api/__main__.py
@@ -71,6 +71,7 @@ async def repost(request: web.Request):
 
 async def properties(request: web.Request):
     logger.debug("Forwarding signing properties")
+    _ = request
 
     url = f"{ALEPH_VM_CONNECTOR}/properties"
     async with aiohttp.ClientSession() as session:
@@ -150,7 +151,7 @@ async def list_keys_from_cache(request: web.Request):
 
     redis: aioredis.Redis = await get_redis()
     result = await redis.keys(f"{prefix}:{pattern}")
-    keys = [key.decode()[len(prefix) + 1 :] for key in result]
+    keys = [key.decode()[len(prefix) + 1:] for key in result]
     return web.json_response(keys)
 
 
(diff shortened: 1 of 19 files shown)


guest_api/__main__.py
```diff
@@ -71,6 +71,7 @@ async def repost(request: web.Request):
 
 async def properties(request: web.Request):
     logger.debug("Forwarding signing properties")
+    _ = request
```
odesenfans, line 74: Some tool is complaining about unused variables?

vm_supervisor/__init__.py
```diff
@@ -2,6 +2,10 @@
 from subprocess import check_output, CalledProcessError
 from typing import Optional
 
+from . import supervisor
+
+assert supervisor
```
odesenfans, line 7: Why?

odesenfans (approved):
Two minor comments, looks good otherwise. I'm surprised the missing typing imports were not caught sooner.

vm_supervisor/__init__.py
hoh: To avoid unused variable warnings from flake8

guest_api/__main__.py
hoh: Exactly

vm_supervisor/__init__.py
odesenfans: I guess this is related to the global app creation. Couldn't you use a create_app() function instead and call that from main()?

odesenfans: If that's not feasible, add a comment.

hoh: It is not required, it just makes it easier to import it from the package (`from vm_supervisor import supervisor, ...).

We should probably add it for most of the package, like [this FastAPI](https://github.com/tiangolo/fastapi/blob/12132276672c51412f3fc38bd5e228cf4a87ecfb/fastapi/__init__.py) `__init__.py` does.

odesenfans: Oh yeah no issues but then we need to deactivate flake8 checks for imports like these. Fine for now, just add a comment though so that we don't forget why the assert is here.

odesenfans (approved):

PR #241, opened 2022-10-05
Feature: VMs could not be persisted

hoh:
The supervisor only supported on request execution of virtual machines.

This add support for an external scheduler to send a list of VMs that should be started and persisted.

A temporary simple HTTP authentication method is used to authenticate the scheduler, this is planned to move to P2P messages from Core Channel Nodes.

### DIFF
--- a/vm_supervisor/conf.py
+++ b/vm_supervisor/conf.py
@@ -106,6 +106,11 @@ class Settings(BaseSettings):
     MAX_PROGRAM_ARCHIVE_SIZE = 10_000_000  # 10 MB
     MAX_DATA_ARCHIVE_SIZE = 10_000_000  # 10 MB
 
+    # hashlib.sha256(b"secret-token").hexdigest()
+    ALLOCATION_TOKEN_HASH = (
+        "151ba92f2eb90bce67e912af2f7a5c17d8654b3d29895b042107ea312a7eebda"
+    )
+
     FAKE_DATA_PROGRAM: Optional[Path] = None
     BENCHMARK_FAKE_DATA_PROGRAM = Path(
         abspath(join(__file__, "../../examples/example_fastapi"))
--- a/vm_supervisor/resources.py
+++ b/vm_supervisor/resources.py
@@ -1,5 +1,6 @@
 from datetime import datetime, timezone
 from functools import lru_cache
+from typing import Set, Optional
 from typing import Tuple
 
 import cpuinfo
@@ -117,3 +118,9 @@ async def about_system_usage(request: web.Request):
     return web.json_response(
         text=usage.json(exclude_none=True),
     )
+
+
+class Allocation(BaseModel):
+    persistent_vms: Set[str]
+    on_demand_vms: Optional[Set[str]] = None
+    jobs: Optional[Set] = None
--- a/vm_supervisor/run.py
+++ b/vm_supervisor/run.py
@@ -214,3 +214,30 @@ async def run_code_on_event(vm_hash: VmHash, event, pubsub: PubSub):
             execution.stop_after_timeout(timeout=settings.REUSE_TIMEOUT)
         else:
             await execution.stop()
+
+
+async def start_persistent_vm(vm_hash: VmHash, pubsub: PubSub) -> VmExecution:
+    execution: Optional[VmExecution] = await pool.get_running_vm(vm_hash=vm_hash)
+
+    if not execution:
+        logger.info(f"Starting persistent VM {vm_hash}")
+        execution = await create_vm_execution(vm_hash=vm_hash)
+    # If the VM was already running in lambda mode, it should not expire
+    # as long as it is also scheduled as long-running
+    execution.marked_as_persistent = True
+    execution.cancel_expiration()
+
+    await execution.becomes_ready()
+
+    if settings.WATCH_FOR_UPDATES:
+        execution.start_watching_for_updates(pubsub=pubsub)
+
+    return execution
+
+
+async def stop_persistent_vm(vm_hash: VmHash) -> Optional[VmExecution]:
+    logger.info(f"Stopping persistent VM {vm_hash}")
+    execution = await pool.get_running_vm(vm_hash)
+    if execution:
+        await execution.stop()
+    return execution
--- a/vm_supervisor/run.py
+++ b/vm_supervisor/run.py
@@ -224,7 +224,7 @@ async def start_persistent_vm(vm_hash: VmHash, pubsub: PubSub) -> VmExecution:
         execution = await create_vm_execution(vm_hash=vm_hash)
     # If the VM was already running in lambda mode, it should not expire
     # as long as it is also scheduled as long-running
-    execution.marked_as_persistent = True
+    execution.persistent = True
     execution.cancel_expiration()
 
     await execution.becomes_ready()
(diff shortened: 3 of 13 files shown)


hoh:
Depends on https://github.com/aleph-im/aleph-vm/pull/240

vm_supervisor/resources.py
```diff
@@ -117,3 +118,9 @@ async def about_system_usage(request: web.Request):
     return web.json_response(
         text=usage.json(exclude_none=True),
     )
+
+
+class Allocation(BaseModel):
+    persistent_vms: Set[str]
+    on_demand_vms: Optional[Set[str]] = None
```
odesenfans, line 125: "on-demand" vs "ephemeral"?

vm_supervisor/run.py
```diff
@@ -216,3 +214,30 @@ async def run_code_on_event(vm_hash: VmHash, event, pubsub: PubSub):
             execution.stop_after_timeout(timeout=settings.REUSE_TIMEOUT)
         else:
             await execution.stop()
+
```
odesenfans, line 217: Unrelated to the PR but I opened the project in PyCharm and I had a nice warning about execution.stop_after_timeout not being awaited, so that's probably a bug you want to take care of.

vm_supervisor/run.py
```diff
@@ -216,3 +214,30 @@ async def run_code_on_event(vm_hash: VmHash, event, pubsub: PubSub):
...
+async def start_persistent_vm(vm_hash: VmHash, pubsub: PubSub) -> VmExecution:
+    execution: Optional[VmExecution] = await pool.get_running_vm(vm_hash=vm_hash)
+
+    if not execution:
+        logger.info(f"Starting persistent VM {vm_hash}")
+        execution = await create_vm_execution(vm_hash=vm_hash)
+    # If the VM was already running in lambda mode, it should not expire
+    # as long as it is also scheduled as long-running
+    execution.marked_as_persistent = True
```
odesenfans, line 227: Maybe a function could be clearer, like execution.mark_as_persistent()?

vm_supervisor/views.py
```diff
@@ -166,3 +169,53 @@ async def status_check_version(request: web.Request):
         )
     else:
         return web.HTTPForbidden(text=f"Outdated: version {current} < {reference}")
+
+
+def authenticate_api_request(request: web.Request) -> bool:
+    """Authenticate an API request to update the VM allocations."""
+    signature: bytes = request.headers.get("X-Auth-Signature").encode()
+    # body: bytes = await request.read()
```
odesenfans, line 177: Remove commented code.

vm_supervisor/conf.py
```diff
@@ -105,6 +105,11 @@ class Settings(BaseSettings):
     MAX_PROGRAM_ARCHIVE_SIZE = 10_000_000  # 10 MB
     MAX_DATA_ARCHIVE_SIZE = 10_000_000  # 10 MB
 
+    # hashlib.sha256(b"secret-token").hexdigest()
+    ALLOCATION_TOKEN_HASH = (
+        "151ba92f2eb90bce67e912af2f7a5c17d8654b3d29895b042107ea312a7eebda"
```
odesenfans, line 111: Same question you had for the RabbitMQ username/password on CCNs: how de we guarantee that node operators do not blindly use the default value?

vm_supervisor/views.py
```diff
@@ -166,3 +169,53 @@ async def status_check_version(request: web.Request):
...
+    if not signature:
+        raise web.HTTPUnauthorized(text="Authentication token is missing")
+
+    # Use a simple authentication method: the hash of the signature should match the value in the settings
+    if sha256(signature).hexdigest() != settings.ALLOCATION_TOKEN_HASH:
+        raise web.HTTPUnauthorized(text="Authentication token received is invalid")
+
+    return True
```
odesenfans, line 185: Seems like this function will either return True or raise, and usages expect it to return True/False. Either change it to return None and raise, or catch exceptions and return False.

vm_supervisor/views.py
```diff
@@ -166,3 +169,53 @@ async def status_check_version(request: web.Request):
...
+
+
+async def update_allocations(request: web.Request):
+    if not authenticate_api_request(request):
+        return web.HTTPUnauthorized(text="Invalid authentication")
+
+    try:
+        data = await request.json()
+        allocation = Allocation(**data)
```
odesenfans, line 194: Use Allocation.parse_obj(data)?

vm_supervisor/views.py
```diff
@@ -166,3 +169,53 @@ async def status_check_version(request: web.Request):
...
+    except ValidationError as error:
+        return web.json_response(
+            data=error.json(), status=web.HTTPBadRequest.status_code
+        )
+
+    pubsub = request.app["pubsub"]
```
odesenfans, line 200: Add a type hint for clarity.

vm_supervisor/views.py
```diff
@@ -166,3 +169,53 @@ async def status_check_version(request: web.Request):
...
+        vm_hash = VmHash(vm_hash)
+        logger.info(f"Starting long running VM {vm_hash}")
+        await start_persistent_vm(vm_hash, pubsub)
+
+    # Stop VMs
+    for execution in pool.get_persistent_executions():
+        if execution.vm_hash not in allocation.persistent_vms:
+            logger.info(f"Stopping long running VM {execution.vm_hash}")
+            await execution.stop()
+            execution.marked_as_persistent = False
```
odesenfans, line 213: What's the point of doing this? Does this not create a possibility of the app being launched as an ephemeral VM if the user hits this CRN with a request next?

vm_supervisor/views.py
```diff
@@ -166,3 +169,53 @@ async def status_check_version(request: web.Request):
...
+
+    # Log unsupported features
+    if allocation.on_demand_vms:
+        logger.info("Not supported yet: 'allocation.on_demand_vms'")
```
odesenfans, line 217: warning?

odesenfans (changes requested):

vm_supervisor/run.py
hoh: I had the same, it is a PyCharm error, probably due to the return value of the following definition:
```python
def stop_after_timeout(self, timeout: float = 5.0) -> Optional[Task]:
```

vm_supervisor/conf.py
hoh: This is the hash of a secret token stored on the scheduler, not a password.

hoh: It is also temporary, and will be deprecated when switching to consensus scheduling with P2P.

vm_supervisor/views.py
hoh: Persistent VMs may also be launched as ephemeral VMs if configured this way, that should not be an issue.

vm_supervisor/run.py
hoh: An extra function just to set an attribute with a similar name ? That looks like overhead-Java to me.

odesenfans: Yeah I don't know, having a passive name like "marked as ..." begs for a function to me, style-wise. Otherwise I'd just name the attribute "persistent" or "is_persistent". But that's complete nitpicking, you can ignore this comment if you like it that way.

vm_supervisor/conf.py
odesenfans: Ah yeah my bad, a bit tired. Ok for me.

vm_supervisor/run.py
hoh: Renamed

odesenfans (approved):

PR #250, opened 2022-10-18
Refactor: Move from iptables to nft as iptables is deprecated

tomribbens:
Since iptables is deprecated, a move to the replacement nftables was implemented in this change. This is in line with various network improvements that are ongoing as discussed in Issue #237. The use of nftables will facilitate some future enhancements, especially related to IPv6 support.

At the same time, this commit refactors all the networking and firewall code into its own package "network", as multiple packages were depending on making network changes. The new package is now an interface for all networking related tasks.


firecracker/microvm.py
```diff
@@ -11,6 +11,7 @@
 from tempfile import NamedTemporaryFile
 from typing import Optional, Tuple, List
 
+from network.network import Network
```
hoh, line 14: Use a relative import when importing from the local project, else it could collide or be confused with a third party library.
```suggestion
from ..network.network import Network
```

network/firewall.py
```diff
@@ -0,0 +1,396 @@
+import logging
+import json
+from typing import List, Dict
+
+import network.network as nw
+from nftables import Nftables
+
+logger = logging.getLogger(__name__)
+
+nft = Nftables()
```
hoh, line 10: Can we avoid this global variable ?

network/firewall.py
```diff
@@ -0,0 +1,396 @@
...
+
+
+class Firewall:
+    firewall_has_been_setup = False
+    postrouting_info: Dict[str, str] = {}
+    forwarding_info: Dict[str, str] = {}
+    vm_info: Dict[int, Dict] = {}
+
+    @classmethod
+    def execute_json_nft_commands(cls, commands: List[Dict]) -> int:
```
hoh, line 25: Why create a class method that does not use anything from the class instead of a standalone function ?

network/firewall.py
```diff
@@ -0,0 +1,396 @@
...
+        """Executes a list of nftables commands, and returns the exit status"""
+        commands_dict = {"nftables": commands}
+        try:
+            logger.debug("Validating nftables rules")
+            nft.json_validate(commands_dict)
+        except Exception as e:
+            logger.error(f"Failed to verify nftables rules: {e}")
+
+        logger.debug("Inserting nftables rules")
+        rc, output, error = nft.json_cmd(commands_dict)
```
hoh, line 35: What does `rc` stand for ? Should the answert to that question be in the code for future readings ?

network/firewall.py
```diff
@@ -0,0 +1,396 @@
...
+        return nft_ruleset
+
+    @classmethod
+    def get_base_chains_for_hook(cls, hook: str, family: str = "ip") -> List:
+        """Looks through the nftables ruleset and creates a list of
+        all chains that are base chains for the specified hook"""
+        nft_ruleset = cls.get_existing_nftables_ruleset()
+        chains = []
+
+        for entry in nft_ruleset["nftables"]:
```
hoh, line 59: What type are these entries ? This could help in the return type `List[EntryType]`.

network/firewall.py
```diff
@@ -0,0 +1,396 @@
...
+                            },
+                            {"accept": None},
+                        ],
+                    }
+                }
+            }
+        )
+
+        cls.execute_json_nft_commands(commands)
+        cls.firewall_has_been_setup = True
```
hoh, line 177: Do I understand that there is a class `Firewall` that only has class methods and whose state is modified ? That class uses a global `nft` variable.
Why not use normal methods and mutate an instance instead ? And create `self.nft` in the `__init__` of the class ?

hoh (commented):
The overall code looks good, however there are some some structural changes to make:

- This is not Java, it is okay to have standalone functions in a module instead of a class namespace. See the [Zen of Python](https://en.wikipedia.org/wiki/Zen_of_Python) for the philosophical approach. Using static and class methods here does not make much sense.

- The name of the `network` package is too generic and can be confused with system libraries or one of the following https://pypi.org/search/?q=network

firecracker/microvm.py
tomribbens: PyCharm automatically renamed it to that when I moved the file. And as I saw it being used like this in firecracker_microvm.py for other stuff, I left it as is. But I'll change it.

network/firewall.py
tomribbens: I'm assuming you're talking about the nft one, not the logger. Yeah that one should probably just move into the execute_json_nft_commands(). Before I had a single function to take care of executing the commands, having the configuration of on top made more sense.

If you're talking about the logger, it's because it's done like that in all files.

network/firewall.py
tomribbens: rc = return code. For me that was very obvious when I saw that exact code in example code on how to use the nftables lib: https://github.com/aborrero/python-nftables-tutorial/blob/main/nft-read-json.py

Together with the very next line where it tests for non-zeroness and if not outputs the error, I don't believe changing it would make it really clearer. naming it return_code or similar only would make it more verbose in my opinion, not clearer.

network/firewall.py
tomribbens: They're dicts, with potentially other dicts in them, but potentially sometimes it contains lists. And depending on the dict, it might contain another layer of dicts or lists. Furtest I've seen it nesting is about 5 levels deep.

network/firewall.py
tomribbens: Yes. As noted above, nft should just move into the single function even. And the one other function that still uses it (get_existing_nftables_ruleset() ) should be rewritten to use the other function as well.

tomribbens:
I don't agree that classmethods don't make sense. There's state information that needs to be kept, while these methods are needed in very different parts of the codebase, so passing around that state information would litter the code too much. This way makes that much easier, and much clearer in the code in my opinion. I don't really understand why you linked to the Zen of Python for this, as I don't see any principle really against what I wrote, but multiple in favor of what I wrote. As indicated, yes, there are some things that might be improved upon, but you've not convinced me why it would need to be restructured.

Some more functions maybe should be @staticmethod instead of @classmethod, but moving them out of the class would make it more confusing to read later on.

True, the name network is generic. But it does cover what it does. I'm not opposed to changing it, but I'm not sure to what. Do you have suggestions?

firecracker/microvm.py
```diff
@@ -434,22 +407,7 @@ async def teardown(self):
...
-            system(
-                f"iptables -t nat -D POSTROUTING -o {self.network_interface} -j MASQUERADE"
-            )
-            system(
-                "iptables -D FORWARD -m conntrack --ctstate RELATED,ESTABLISHED -j ACCEPT"
-            )
-            system(
-                f"iptables -D FORWARD -i {self.network_tap} -o {self.network_interface} -j ACCEPT"
-            )
+        await Network.remove_tap_interface(self.vm_id)
```
hoh, line 410: We may want to store a reference to this function instead of importing it, removing circular imports between packages.

vm_supervisor/ip.py
```diff
@@ -0,0 +1,530 @@
...
+        if vm_id not in self.vm_info:
+            logger.error(f"No firewall configuration found to teardown for vm {vm_id}")
+            return
+
+        self.remove_chain(self.vm_info[vm_id]["nat_chain"])
+        self.remove_chain(self.vm_info[vm_id]["filter_chain"])
+
+
+class Network:
+    firewall: Firewall = Firewall()
```
hoh, line 475: Initializing a variable in the class definition stores the instance on the class and not the instance.
This is usually not what we are looking for, especially when running tests.

```suggestion
    firewall: Firewall
    def __init__(self):
        self.firewall = Firewall()
```

hoh (commented):
As a general rule, I try to limit to 30 lines per function and 300 lines per file. Splitting the `ip.py` file in multiple modules within the same directory/package would be clearer.

hoh:
The `AlephFirecrackerVM` class is responsible of managing VM related resources that are specific to Aleph. I feel like it would make sense to move/call much of the firewall management code from there. This removes some indirections as discussed.

https://github.com/aleph-im/aleph-vm/commit/170e1a420514c9815728b516a5089ef2123150c8

vm_supervisor/ip.py
```diff
@@ -0,0 +1,530 @@
...
+    ipv4_forward_state_before_setup = None
+    address_pool: IPv4NetworkWithInterfaces = IPv4NetworkWithInterfaces("172.16.0.0/12")
+    network_size = 24
+    network_initialized = False
+    external_interface = "eth0"
+    vm_info: Dict = {}
```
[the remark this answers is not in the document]
tomribbens: The dict is already unused, it's a remaining artifact. I'm deleting it.

vm_supervisor/ip.py
```diff
@@ -0,0 +1,530 @@
...
+                f.write("1")
+
+    def reset_ipv4_forwarding_state(self) -> None:
+        """Returns the hosts IPv4 forwarding state how it was before we enabled it"""
+        logger.debug("Resetting IPv4 forwarding state to state before we enabled it")
+        if self.ipv4_forward_state_before_setup != get_ipv4_forwarding_state():
+            with open("/proc/sys/net/ipv4/ip_forward", "w") as f:
+                f.write(str(self.ipv4_forward_state_before_setup))
+
+    def initialize(
```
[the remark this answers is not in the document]
tomribbens: Yep, doing now.

tomribbens:
> As a general rule, I try to limit to 30 lines per function and 300 lines per file. Splitting the `ip.py` file in multiple modules within the same directory/package would be clearer.

After you split them out, I had some issues with them being too tightly coupled and having circular dependencies, so I joined them together again to resolve that. However, those dependencies seem to no longer be a problem, so I split them out again.

tomribbens:
> The `AlephFirecrackerVM` class is responsible of managing VM related resources that are specific to Aleph. I feel like it would make sense to move/call much of the firewall management code from there. This removes some indirections as discussed.
> 
> [170e1a4](https://github.com/aleph-im/aleph-vm/commit/170e1a420514c9815728b516a5089ef2123150c8)

So, the main reason it is the way it currently is, is because the code is not assuming nftables is set up in the default way, but actually will inspect the existing ruleset to determine some settings. This information is then stored in the instance of the Firewall class.  Currently there is some other information stored in there as well, but I've been thinking about it, and I think that can be removed as those things are simply based of the vm_id, and so can be recalculated.

We can re-inspect the existing ruleset every time we do a change to the firewall rules, and if we do, the whole Firewall class may be removed, and make all its methods just normal functions. I'm a bit torn on it, as I don't like the idea of recalculating this info every time since it comes from an external system, but on the other hand, it would eliminate passing the reference down as the code is currently doing.

tomribbens:
I removed the Firewall class. I did it in a seperate branch as I wanted to keep it seperate until certain. See the diff here: https://github.com/tomribbens/aleph-vm/compare/tomribbens-migrate-to-nftables...tomribbens:aleph-vm:tomribbens-migrate-to-nftables-no-firewall-object

This doesn't put anything in AlephFirecrackerVM though. Not sure if that's actually something that's actually needed. This seems fine as is to me.

hoh:
Can you confirm that the object 'Nftables' returned by `get_customized_nftables()` and can be instantiated multiple times without any issue ? If that is the case, removing the class indeed makes the code cleaner. 
Caching that function by decorating it with `lru_cache()`[https://docs.python.org/3/library/functools.html#functools.lru_cache] could remove the performance issue when calling the same function multiple times.

Regarding the `# TODO: gracefully exit`, the common approach in Python is to raise an exception and catch it where it should be handled. Can you explain in which situations this could happen ?

tomribbens:
> Can you confirm that the object 'Nftables' returned by `get_customized_nftables()` and can be instantiated multiple times without any issue ? If that is the case, removing the class indeed makes the code cleaner. 

I don't see why not. It's just loading and unloading the interface lib file each time.

> Caching that function by decorating it with `lru_cache()`[https://docs.python.org/3/library/functools.html#functools.lru_cache] could remove the performance issue when calling the same function multiple times.

That's a possibility. get_customized_nftables() takes about 1ms to execute according to my tests. In the objectless way, it gets run basically every time a firewall rule is added or removed, which would be at startup and teardown of a vm.  From testing, I get that it is about 9KiB of data in memory. Both are small numbers, so I'm not sure what side of the tradeoff is best.

 
> Regarding the `# TODO: gracefully exit`, the common approach in Python is to raise an exception and catch it where it should be handled. Can you explain in which situations this could happen ?

2 different chains could register with the kernel in the same hook. If they do, both chains would be called. But if they do, and don't agree on what to do with a certain packet, weird stuff would happen. From what I understand, it's a situation that isn't really that useful, just technically possible in the way the netfilter architecture is setup. In practice, I don't think we would encounter the situation ever.

hoh (approved):

PR #252, opened 2022-11-22
Add an example for Node.js

### LINKED ISSUE #255: The Typescript SDK is not compatible with the runtime Nodejs 12
**Describe the bug**

The Typescript SDK is incompatible with the current runtime since only Nodejs 12 is provided - and the SDK requires a more recent version.

We should install a recent version of Nodejs when creating the runtime rootfs instead of using the version provided by Debian 11 Bullseye.

BjrInt:
This example is based off the FastAPI example with some differences:

- The `state` example is using an ephemeral keypair (used to send message) instead of a counter
- The "post a message on Aleph" and "retrieve http post data" are merged into a single example
- There is no "cache"


hoh:
The goal of the `state` example is to mute data in a file on disk inside a mutable volume. Why change its logic ?

hoh:
How can we build and package this application ?

moshemalawach:
Not reviewed at all, this was a spam comment by a spam account @NostalgicGareth. He goes and spam a lot of projects like this.

oksome (approved):

oksome (approved):

hoh (commented):
Forgot to submit this review 🫣

MHHukiewitz:
Can we get confirmation that this still works as intended?

We talked about how we should support more languages/frameworks in the VMs and also to improve examples/docs for deploying these. Merging this PR should help with that.

odesenfans:
Closing as part of a cleanup of pre-#1000 PRs. This branch predates the 2026-02-18 force-push of main, so it can no longer be rebased (GitHub shows the whole history as the diff). Nothing in the branch can be rebased onto the current tree, and the open question from review (how CI builds and tests the example) was never settled. A Node.js example would be a fresh PR against the current examples/ layout.

PR #256, opened 2022-12-28
Feature: version endpoint

odesenfans:
Problem: retrieving the version of the VM supervisor directly can be useful for monitoring and debugging.

Solution: add an endpoint that returns the software version. The response has the same format as the /version endpoint of Core Channel Nodes.


hoh:
The version of the VM supervisor is already available in the HTTP headers of every HTTP response:
```
server: aleph-vm/0.2.5-5-g15e0b60
```
Is this not sufficient ?

MHHukiewitz:
I assume this can be closed?

PR #260, opened 2023-01-31
Fix: Network route should not contain IP mask

hoh:
This was breaking the `ip route add` command in the initialisation of VMs.


nesitor (approved):
:rocket:

AmozPay (approved):
Looks good to me

PR #261, opened 2023-01-31
Add requirements to support Docker in runtime

hoh:


runtimes/aleph-alpine-3.13-python/init0.sh
```diff
@@ -29,6 +29,12 @@ mount -t tmpfs run /run -o mode=0755,nosuid,nodev
 mount -t devpts devpts /dev/pts -o mode=0620,gid=5,nosuid,noexec
 mount -t tmpfs shm /dev/shm -omode=1777,nosuid,nodev
 
+# Required by Docker
+cgroupfs-mount
+update-alternatives --set iptables /usr/sbin/iptables-legacy
+update-alternatives --set ip6tables /usr/sbin/ip6tables-legacy
+echo '{\n"storage-driver": "overlay2"\n}\n' > /etc/docker/daemon.json
```
nesitor, line 36: For our actual indexers, we need to have it in vfs I think.

AmozPay: Same thought, I believe forcing use of overlay might lead to unpredictable behavior as not all systems are compatible. Overlay2 is the de facto storage driver so it would not be needed.

Here is the storage driver preference order detection:
<html><body>


Linux distribution | Recommended storage drivers | Alternative drivers
-- | -- | --
Ubuntu | overlay2 | overlay¹, devicemapper², aufs³, zfs, vfs
Debian | overlay2 | overlay¹, devicemapper², aufs³, vfs
CentOS | overlay2 | overlay¹, devicemapper², zfs, vfs
Fedora | overlay2 | overlay¹, devicemapper², zfs, vfs
SLES 15 | overlay2 | overlay¹, devicemapper², vfs
RHEL | overlay2 | overlay¹, devicemapper², vfs


</body>
</html>

> The vfs storage driver is usually not the best choice, and primarily intended for debugging purposes in situations where no other storage-driver is supported.

Although not the most performant, vfs is the most compatible one.

https://docs.docker.com/storage/storagedriver/select-storage-driver/

hoh: VFS works everywhere but it copies everything when a user creates a container from an image, duplicating the entire storage.

Forcing the storage driver makes Docker fail instead of switching to VFS without notice.

The only issue with overlay2 inside VMs is that it cannot run on top of another overlay filesystem, which is the case for `/` in the runtime.

I will remove that line from this so we can discuss it further in another issue / topic.

PR #266, opened 2023-02-09
Start design to support containers in VMs.

hoh:


nesitor:
This PR is pretty old and we found another solution.

PR #267, opened 2023-02-13
Fix: Command `systemd-resolve` does not exist in Ubuntu 22.04

hoh:
THe command `resolvectl` however appears to be present in both Ubuntu 20.04 and Ubuntu 22.04.


hoh:
Replaced within #268

PR #270, opened 2023-02-13
Fix: Supervisor started without /dev/kvm

### LINKED ISSUE #269: Running without KVM does not fail early

hoh:
Problem: The supervisor could start while the virtualization device /dev/kvm was absent from the system.

Solution: Add a check in the initialization of the configuration for /dev/kvm to be present.

Fixes #269


nesitor (approved):
:rocket:

PR #283, opened 2023-03-15
Fixed the virtual machine hostname that causes the virtual machine to…

aliel:
@Todo after merge:
-  Create a new runtime


hoh:
We should not crop the vm item_hash. If the source of this bug cannot be resolved, we should then use another representation of the item_hash such as it's base32 encoding.

aliel:
Tested with this code =>

```
#!/bin/bash

python3.9 -m http.server 8080
```

PR #292, opened 2023-04-04
Added modifications to enable instance messages

nesitor:
PR based on the latest release.
Related with these PRs:
- https://github.com/aleph-im/aleph-message/pull/47
- https://github.com/aleph-im/aleph-message/pull/48


firecracker/microvm.py
```diff
@@ -278,6 +284,34 @@ def enable_rootfs(self, path_on_host: str) -> str:
         else:
             return path_on_host
 
+    def mount_rootfs(self, path_on_host: str) -> str:
```
hoh, line 287: Using `pathlib.Path` for paths makes them more convenient to manipulate.

It also includes handy tools to concatenate paths such as `Path('/dir') / "filename")` and `Path('hello').symlink_to('world').

firecracker/microvm.py
```diff
@@ -278,6 +284,34 @@ def enable_rootfs(self, path_on_host: str) -> str:
...
+            rootfs_filename = Path(path_on_host).name
+            jailer_path_on_host = f"{DEVICE_BASE_DIRECTORY}/{rootfs_filename}"
+            if not check_device_exists(Path(f"{self.jailer_path}/{jailer_path_on_host}")):
+                device_vm_path = f"{DEVICE_BASE_DIRECTORY}/{self.vm_id}"
+                subprocess.run(["mkdir", "-p", device_vm_path],
+                               check=True,
+                               capture_output=True)
+                path_to_mount = f"{device_vm_path}/{rootfs_filename}"
+                os.symlink(path_on_host, path_to_mount)
+                os.system(f"cp -vap /dev/dm* {self.jailer_path}/dev/")
```
hoh, line 301: Why mix `os.system` with `subprocess.run` ?

vm_supervisor/storage.py
```diff
@@ -155,28 +161,98 @@ def create_ext4(path: Path, size_mib: int) -> bool:
...
 
 
+def check_device_exists(path: Path) -> bool:
+    try:
+        return stat.S_ISBLK(os.stat(path).st_mode)
+    except Exception:
+        return False
+
+
+async def create_devmapper(volume: PersistentVolume, namespace: str) -> Path:
```
hoh, line 171: This is a very long function. Should it be split in multiple, easier to understand functions ?

firecracker/microvm.py
```diff
@@ -278,6 +284,34 @@ def enable_rootfs(self, path_on_host: str) -> str:
...
             return path_on_host
 
+    def mount_rootfs(self, path_on_host: str) -> str:
+        """Mount a rootfs to the VM.
+        """
+        self.mounted_fs = path_on_host
+        if self.use_jailer:
+            rootfs_filename = Path(path_on_host).name
+            jailer_path_on_host = f"{DEVICE_BASE_DIRECTORY}/{rootfs_filename}"
+            if not check_device_exists(Path(f"{self.jailer_path}/{jailer_path_on_host}")):
```
hoh, line 294:
```suggestion
            if not (self.jailer_path / jailer_path_on_host).is_block_device():
```

vm_supervisor/storage.py
```diff
@@ -155,28 +161,98 @@ def create_ext4(path: Path, size_mib: int) -> bool:
     return True
 
 
+def check_device_exists(path: Path) -> bool:
```
hoh, line 164: Can this function be replaced with `pathlib.Path(...).is_block_device()` ?

vm_supervisor/INSTANCES.md
```diff
@@ -0,0 +1,44 @@
+
+# Instance Messages
+
+I have added support to Instance message in the aleph-message repository, in this PR:
```
hoh, line 4: Please use the passive form in documentation. "Support of Instance message in the aleph-message repository was added in this PR: ..." instead of "I have..."

hoh (commented):
I had a first look around and posted some comments.

Why is there a new runtime named `aleph-docker-rootfs` ?

odesenfans:
@nesitor Could you rebase this PR?

firecracker/microvm.py
nesitor: Because the `os.system` doesn't return any output, but `subprocess.run` does, and sometimes we don't want to check the system output.

vm_supervisor/storage.py
nesitor: I can split, but it doesn't make a lot of sense to do it in my opinion, because the command execution needs to be done in order to work.

firecracker/microvm.py
```diff
@@ -132,6 +135,8 @@ def prepare_jailer(self):
         system(f"chown jailman:jailman {self.jailer_path}/tmp/")
         #
         system(f"mkdir -p {self.jailer_path}/opt")
+        system(f"mkdir -p {self.jailer_path}/dev/mapper")
+        system(f"mkdir -p {self.jailer_path}/dev/aleph-mapper")
```
odesenfans, line 139: Why do we prepare the dev mapper in `prepare_jailer()`? Shouldn't this be its own step?

firecracker/microvm.py
```diff
@@ -262,6 +267,34 @@ def enable_rootfs(self, path_on_host: str) -> str:
...
+        else:
+            return path_on_host
+
+    def mount_rootfs(self, path_on_host: Path) -> Path:
+        """Mount a rootfs to the VM.
+        """
+        self.mounted_fs = path_on_host
+        if self.use_jailer:
+            rootfs_filename = Path(path_on_host).name
+            jailer_path_on_host = Path(f"{DEVICE_BASE_DIRECTORY}/{rootfs_filename}")
```
odesenfans, line 280: Nitpick: instead of using an f-string, you can do the following:


```suggestion
            jailer_path_on_host = Path(DEVICE_BASE_DIRECTORY) / rootfs_filename
```

firecracker/microvm.py
```diff
@@ -262,6 +267,34 @@ def enable_rootfs(self, path_on_host: str) -> str:
...
+            if not (self.jailer_path / jailer_path_on_host).is_block_device():
+                device_vm_path = f"{DEVICE_BASE_DIRECTORY}/{self.vm_id}"
+                subprocess.run(["mkdir", "-p", device_vm_path],
+                               check=True,
+                               capture_output=True)
```
odesenfans, line 285: Use pathlib to create the folder directly: https://docs.python.org/3/library/pathlib.html#pathlib.Path.mkdir

firecracker/microvm.py
```diff
@@ -262,6 +267,34 @@ def enable_rootfs(self, path_on_host: str) -> str:
...
+                path_to_mount = f"{device_vm_path}/{rootfs_filename}"
+                os.symlink(path_on_host, path_to_mount)
```
odesenfans, line 287: https://docs.python.org/3/library/pathlib.html#pathlib.Path.symlink_to

firecracker/microvm.py
```diff
@@ -262,6 +267,34 @@ def enable_rootfs(self, path_on_host: str) -> str:
...
+                os.system(f"cp -vap /dev/dm* {self.jailer_path}/dev/")
```
odesenfans, line 288: Use `shutil.copy` or any other standard Python function. Calling bash processes is not needed here.

firecracker/microvm.py
```diff
@@ -400,6 +433,17 @@ async def teardown(self):
         if self.stderr_task:
             self.stderr_task.cancel()
 
+        if self.mounted_fs:
+            logger.debug("Waiting for one second for the VM to shutdown")
```
odesenfans, line 437: Is there no signal you could use from Firecracker, like the PID of the VM or something like that? It would be better to wait until the VM is down and not an arbitrary delay.

firecracker/microvm.py
```diff
@@ -400,6 +433,17 @@ async def teardown(self):
...
+            await asyncio.sleep(1)
+            root_fs = self.mounted_fs.name
+            os.system(f"dmsetup remove {root_fs} {root_fs}_base")
+            if self.use_jailer:
+                os.system(f"umount {self.jailer_path}/dev/mapper")
+                os.system(f"umount {self.jailer_path}/dev/aleph-mapper")
+                os.system(f"rm -rf {self.jailer_path}/dev")
```
odesenfans, line 444: Use pathlib whenever possible, using `os.system` with string interpolation should be avoided if possible.

runtimes/aleph-docker-rootfs/create-debian-disk.sh
```diff
@@ -0,0 +1,105 @@
...
+echo ttyS0 > /etc/securetty
+EOT
+
+echo "PermitRootLogin yes" >> ./rootfs/etc/ssh/sshd_config
+
+# Generate SSH host keys
+#systemd-nspawn -D ./rootfs/ ssh-keygen -q -N "" -t dsa -f /etc/ssh/ssh_host_dsa_key
+#systemd-nspawn -D ./rootfs/ ssh-keygen -q -N "" -t rsa -b 4096 -f /etc/ssh/ssh_host_rsa_key
+#systemd-nspawn -D ./rootfs/ ssh-keygen -q -N "" -t ecdsa -f /etc/ssh/ssh_host_ecdsa_key
+#systemd-nspawn -D ./rootfs/ ssh-keygen -q -N "" -t ed25519 -f /etc/ssh/ssh_host_ed25519_key
```
odesenfans, line 63: Delete commented code. Or at least add a comment to explain why it is commented.

vm_supervisor/INSTANCES.md
```diff
@@ -0,0 +1,44 @@
...
+Also, I added support to run a writable root filesystem in Firecracker. I have bypassed all the parts that we pass and
+use the **_"code"_** properties, like the encoding or the entrypoint.
+
+A new instance message example has been added in **_examples/instance_message_from_aleph.json_**.
+
+### Current status
+
+Now the Dev-mapper device works well, Firecracker loads it in write state, but we need to fix 2 things:
+- Route the requests from the CRN to the Firecracker VM on any port, not only using the 8080.
+~~- Use the entire hard disk inside VM, because now only detects the size of the rootfs.~~(Fixed)
```
odesenfans, line 44: What do you mean by "Fixed"?

vm_supervisor/models.py
```diff
@@ -187,16 +190,27 @@ def start_watching_for_updates(self, pubsub: PubSub):
...
-            *(
-                volume.ref
-                for volume in (self.original.volumes or [])
-                if hasattr(volume, "ref")
-            ),
-        )
+        if self.is_instance:
+            await pubsub.msubscribe(
+                self.original.rootfs.parent,
+                self.original.data.ref if hasattr(self.original, "data") and self.original.data else None,
```
odesenfans, line 196: Instances do not have a data volume.

vm_supervisor/models.py
```diff
@@ -187,16 +190,27 @@ def start_watching_for_updates(self, pubsub: PubSub):
...
-            self.original.data.ref if self.original.data else None,
-            *(
-                volume.ref
-                for volume in (self.original.volumes or [])
-                if hasattr(volume, "ref")
-            ),
-        )
+        if self.is_instance:
+            await pubsub.msubscribe(
+                self.original.rootfs.parent,
```
odesenfans, line 195: I don't think we care about updates to the root file system for instances. We create the instance as a copy of the rootfs at that point in time and then let it live it's life. The user may have written data to their root partition that we should not overwrite.

vm_supervisor/models.py
```diff
@@ -187,16 +190,27 @@ def start_watching_for_updates(self, pubsub: PubSub):
...
+                self.original.runtime.ref,
+                self.original.data.ref if self.original.data else None,
+                *(
+                    volume.ref
+                    for volume in (self.original.volumes or [])
+                    if hasattr(volume, "ref")
+                ),
+            )
         logger.debug("Update received, stopping VM...")
         await self.stop()
```
odesenfans, line 215: Do we really want to stop instances on volume updates? I'm not sure the whole concept of updating the volumes of an instance makes sense. To be discussed with @hoh.

vm_supervisor/storage.py
```diff
@@ -155,28 +161,93 @@ def create_ext4(path: Path, size_mib: int) -> bool:
...
+async def create_devmapper(volume: PersistentVolume | RootfsVolume, namespace: str) -> Path:
+    volume_name = volume.name if isinstance(volume, PersistentVolume) else "rootfs"
+    volume_dev_name = f"{namespace}_{volume_name}"
+    path_volume_dev_name = Path(f"/dev/mapper/{volume_dev_name}")
+    if path_volume_dev_name.is_block_device():
+        return path_volume_dev_name
+    path = Path(
+        join(settings.PERSISTENT_VOLUMES_DIR, namespace, f"{volume_name}.ext4")
+    )
+    if not os.path.isfile(path):
```
odesenfans, line 173: `path.is_file()` works.

odesenfans (changes requested):
Okay, so what's the best way to test this?
I still have to review the device mapper code, I'll read a bit on it and take a second look.

vm_supervisor/models.py
hoh: Interesting question. 
- Immutable volumes: could be updated in place, ignored or updated after a reboot.
- Persistent volumes: it could be handy allow resizing (size increase).

If an instance uses a volume from a third party, that third party would be able to initiate the reboot of the instances, which may be problematic.

When users use `use_latest = True` on volumes, they expect updates to the volumes to be applied to running functions. We could keep the same behavior for instances but put that field by default to `False`, hence mitigating the security vulnerability.

vm_supervisor/storage.py
```diff
@@ -155,28 +161,93 @@ def create_ext4(path: Path, size_mib: int) -> bool:
     return True
 
 
+async def create_devmapper(volume: PersistentVolume | RootfsVolume, namespace: str) -> Path:
```
odesenfans, line 164: This function is quite complex, some comments and documentation would be more than welcome.

vm_supervisor/storage.py
```diff
@@ -155,28 +161,93 @@ def create_ext4(path: Path, size_mib: int) -> bool:
...
+    volume_name = volume.name if isinstance(volume, PersistentVolume) else "rootfs"
+    volume_dev_name = f"{namespace}_{volume_name}"
+    path_volume_dev_name = Path(f"/dev/mapper/{volume_dev_name}")
```
odesenfans, line 167: Weird name. Maybe `mapped_volume_path` would be clearer?

vm_supervisor/storage.py
```diff
@@ -155,28 +161,93 @@ def create_ext4(path: Path, size_mib: int) -> bool:
     return True
 
 
+async def create_devmapper(volume: PersistentVolume | RootfsVolume, namespace: str) -> Path:
+    volume_name = volume.name if isinstance(volume, PersistentVolume) else "rootfs"
+    volume_dev_name = f"{namespace}_{volume_name}"
```
odesenfans, line 166: Naming: `mapped_volume_name`?

vm_supervisor/storage.py
```diff
@@ -155,28 +161,93 @@ def create_ext4(path: Path, size_mib: int) -> bool:
...
+    path_volume_dev_name = Path(f"/dev/mapper/{volume_dev_name}")
+    if path_volume_dev_name.is_block_device():
+        return path_volume_dev_name
+    path = Path(
+        join(settings.PERSISTENT_VOLUMES_DIR, namespace, f"{volume_name}.ext4")
+    )
```
odesenfans, line 172: I'm not 100% sure but isn't this incorrect when jailer is used?

odesenfans, line 172: Avoid mixing `os.path` and `pathlib`.

```suggestion
    path = Path(settings.PERSISTENT_VOLUMES_DIR) / namespace / f"{volume_name}.ext4"
```

vm_supervisor/storage.py
```diff
@@ -155,28 +161,93 @@ def create_ext4(path: Path, size_mib: int) -> bool:
...
+        loop_user_data = subprocess.run(
+            ["losetup", "--find", "--show", path],
+            check=True,
+            capture_output=True,
+            encoding="UTF-8").stdout.strip()
+        table_command = f"0 {root_size} linear {loop_base} 0\n{root_size} {volume_data_size} zero"
+        base_dev_name = f"{namespace}_{volume_name}_base"
+        os.system(f" printf \"{table_command}\" | dmsetup create {base_dev_name}")
+        path_base_dev_name = f"/dev/mapper/{base_dev_name}"
+        table_command = f"0 {volume_data_size} snapshot {path_base_dev_name} {loop_user_data} P 8"
```
odesenfans, line 204: How did you determine the chunk size (8)?

vm_supervisor/storage.py
```diff
@@ -155,28 +161,93 @@ def create_ext4(path: Path, size_mib: int) -> bool:
...
+        os.system(f" printf \"{table_command}\" | dmsetup create {volume_dev_name}")
+        os.system(f"e2fsck -fy {path_volume_dev_name.__str__()}")
+        os.system(f"resize2fs {path_volume_dev_name.__str__()}")
```
odesenfans, line 207: Do you need to check the file system when creating it? Maybe this is something that we should only do when the root filesystem already exists?

odesenfans (changes requested):
Reviewed the device mapping part.

odesenfans:
What happens if the parent rootfs has multiple partitions?

vm_supervisor/storage.py
nesitor: We need to check it always, because if not the `resize2fs` command doesn't work.

vm_supervisor/storage.py
nesitor: I don't know exactly, I set it using [this](https://community.aleph.im/t/deploying-mutable-vm-instances-on-aleph/56/2) example command, we can investigate how we can ensure that is the correct chunk size.

vm_supervisor/storage.py
nesitor: No, because this method is to set up the volume on the host. After, in the `mount_rootfs` ([here](https://github.com/aleph-im/aleph-vm/blob/andres_enable-instance-message/firecracker/microvm.py#L274)) we set the proper URL for the mapped device if we use Jailer.

vm_supervisor/models.py
nesitor: I only left the subscription for volumes, to cover the use case that @hoh comments.

vm_supervisor/storage.py
nesitor: I just added some comments and link to documentation.

vm_supervisor/run.py
```diff
@@ -145,7 +145,7 @@ async def run_code_on_request(
         headers.update(
             {
                 "Aleph-Program-ItemHash": execution.vm_hash,
-                "Aleph-Program-Code-Ref": execution.program.code.ref,
+                "Aleph-Program-Code-Ref": execution.message.code.ref if not execution.is_instance else None,
```
hoh, line 148: Why this change ?

vm_supervisor/storage.py
```diff
@@ -155,28 +162,100 @@ def create_ext4(path: Path, size_mib: int) -> bool:
...
+        encoding="UTF-8").stdout.strip()
+    extended_block_size = subprocess.run(
+        ["blockdev", "--getsz", path],
+        check=True,
+        capture_output=True,
+        encoding="UTF-8").stdout.strip()
+    # Creates the first mapped device using the base disk image
+    table_command = f"0 {base_block_size} linear {base_loop_device} 0\n{base_block_size} {extended_block_size} zero"
+    base_volume_name = f"{namespace}_{volume_name}_base"
+    os.system(f"printf \"{table_command}\" | dmsetup create {base_volume_name}")
```
hoh, line 209: Can we avoid using `os.system` ?

```suggestion
    subprocess.run(cmd=f"dmsetup create {base_volume_name}", input=table_command, text=True, shell=True, check=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE
)
```

vm_supervisor/storage.py
```diff
@@ -155,28 +162,100 @@ def create_ext4(path: Path, size_mib: int) -> bool:
...
+    # the resize command doesn't run.
+    os.system(f"e2fsck -fy {path_mapped_volume_name}")
+    os.system(f"resize2fs {path_mapped_volume_name}")
+    if settings.USE_JAILER:
+        os.system(f"chown jailman:jailman {path_base_device_name}")
+        os.system(f"chown jailman:jailman {path_mapped_volume_name}")
+    return path_mapped_volume_name
+
+
+async def get_existing_file(ref: str) -> Path:
```
hoh, line 225: What is the purpose of this function compared to `download_file` ?

vm_supervisor/storage.py
```diff
@@ -155,28 +162,100 @@ def create_ext4(path: Path, size_mib: int) -> bool:
...
-        if settings.FAKE_DATA_PROGRAM and settings.FAKE_DATA_VOLUME:
-            return Path(settings.FAKE_DATA_VOLUME)
-
-        cache_path = Path(join(settings.DATA_CACHE, ref))
-        url = f"{settings.CONNECTOR_URL}/download/data/{ref}"
-        await download_file(url, cache_path)
-        return cache_path
-    elif isinstance(volume, PersistentVolume):
+        return await get_existing_file(ref)
+    elif isinstance(volume, PersistentVolume) | isinstance(volume, RootfsVolume):
```
hoh, line 241: Why `|` and not `or` ?

vm_supervisor/run.py
nesitor: Because an instance don't have a code reference, because don't have any code.

vm_supervisor/storage.py
nesitor: This function already exist before, and is to check if this reference file exists, also to return the fake data program and to set the correct permissions if we use Jailer. Also the download_file method is only to download the files only.

.github/workflows/test-on-droplet-debian-11.yml
```diff
@@ -73,6 +74,14 @@ jobs:
...
           curl --retry 5 --max-time 10 --fail "http://${DROPLET_IPV4}:4020/status/check/fastapi"
 
+      - name: Schedule an instance on the Droplet by faking a call from the scheduler
+        run: |
+          export DROPLET_IPV4="$(doctl compute droplet get aleph-vm-ci-debian-11 --output json | ./.github/scripts/extract_droplet_ipv4.py)"
+          curl --retry 5 --max-time 10 --fail -X POST -H "Content-Type: application/json" \
+            -H "X-Auth-Signature: test" \
+            -d '{"persistent_vms": [], "instances": ["INSTANCE-HASH-TODO-FIXME"]}' \
+            "http://${DROPLET_IPV4}:4020/control/allocations"
+
```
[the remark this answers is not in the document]
nesitor: We need to have a message already published from an instance to be able to test it.

runtimes/aleph-alpine-3.13-python/init1.py
```diff
@@ -57,12 +57,12 @@ class Volume:
...
 class ConfigurationPayload:
-    code: bytes
-    encoding: Encoding
-    entrypoint: str
     input_data: bytes
     interface: Interface
     vm_hash: str
+    code: Optional[bytes] = None
+    encoding: Optional[Encoding] = None
+    entrypoint: Optional[str] = None
```
hoh, line 65: Can we group these fields together in order to make it explicit that they are used for running programs/functions ?

firecracker/microvm.py
```diff
@@ -262,30 +267,68 @@ def enable_rootfs(self, path_on_host: str) -> str:
...
+        if not self.use_jailer:
+            return path_on_host
+
+        rootfs_filename = Path(path_on_host).name
+        jailer_path_on_host = Path(DEVICE_BASE_DIRECTORY) / rootfs_filename
+        if not (self.jailer_path / jailer_path_on_host).is_block_device():
+            device_vm_path = Path(DEVICE_BASE_DIRECTORY) / str(self.vm_id)
+            device_vm_path.mkdir(exist_ok=True, parents=True)
+            path_to_mount = device_vm_path / rootfs_filename
+            if not path_to_mount.is_symlink():
```
hoh, line 287: We discussed using `pathlib.Path(...).resolve()` to use the target of symlinks directly and refactor this secion.

firecracker/microvm.py
```diff
@@ -262,30 +267,68 @@ def enable_rootfs(self, path_on_host: str) -> str:
...
+                # Mount and rbind (recursively bind) /dev/aleph-mapper folder to the Jailer path, but only for the
+                # target VM devices
+                subprocess.run(
+                    ["mount", "--rbind", device_vm_path, jailer_path_to_mount],
+                    check=True,
+                    capture_output=True)
+                os.system(f"chown -Rh jailman:jailman {self.jailer_path}/dev")
+
+        return jailer_path_on_host
+
```
hoh, line 310: We must ensure that only the device mappers relevant to a VM are available inside its jailer.

firecracker/microvm.py
```diff
@@ -400,6 +443,17 @@ async def teardown(self):
...
 
+        if self.mounted_fs:
+            logger.debug("Waiting for one second for the VM to shutdown")
+            await asyncio.sleep(1)
+            root_fs = self.mounted_fs.name
+            os.system(f"dmsetup remove {root_fs} {root_fs}_base")
+            if self.use_jailer:
+                os.system(f"umount {self.jailer_path}/dev/mapper")
+                os.system(f"umount {self.jailer_path}/dev/aleph-mapper")
+                os.system(f"rm -rf {self.jailer_path}/dev")
```
hoh, line 454: What would be left in this directory and prevent using rmdir ?

hoh, line 454: Use `shutil.rmtree` or similar instead of `os.system("rm -rf ...")`.

hoh (commented):
We must ensure that only the device mappers relevant to a VM are available inside its jailer.

firecracker/microvm.py
hoh: Moved to a new issue so we can merge this.

hoh (approved):

hoh (approved):

Issue #293, opened 2023-04-20
ASGI Lifespan Protocol needs to be implemented


MHHukiewitz:
**Is your feature request related to a problem? Please describe.**
When using lifespan events for executing code on startup or shutdown of a VM, like described in the [FastAPI Docs](https://fastapi.tiangolo.com/advanced/events/), then this will work on apps being run with uvicorn, but not when deployed on Aleph.

**Describe the solution you'd like**
- On-demand, as well as persistent VMs, should have their ASGI lifespan events invoked.
- Ideally, we can autodetect lifespan events like in [Uvicorn](https://github.com/encode/uvicorn/blob/2a94a96e43a3830aa2d0fe90f3e21ac6d06bd788/uvicorn/lifespan/on.py)
- Lifespan events should belong to the code subject to timeout requirements and may be interrupted if not respecting those

**Describe alternatives you've considered**
- Optionally add a "lifespan" config for VMs to explicitly enforce that these events are being sent to the VM or not
- Stop VM execution if `lifespan.startup` failed and "lifespan" config is set to `true`

**Additional context**
Currently, timeouts are realized at the [`firecracker_microvm.py`](https://github.com/aleph-im/aleph-vm/blob/main/vm_supervisor/vm/firecracker_microvm.py#L478) level:
```python
return await asyncio.wait_for(
    communicate(reader, writer, scope),
    timeout=self.hardware_resources.seconds,
)
```
This should be rather on the [`VmExecution`](https://github.com/aleph-im/aleph-vm/blob/main/vm_supervisor/models.py#L39) level, given that we might also use other VM jailers. If for any reason startup and shutdown procedures are taking longer than expected, then this should not go "unbilled".

hoh:
This was merged.

PR #294, opened 2023-04-20
Implement ASGI Lifespan protocol

MHHukiewitz:
Should solve #293 in part. Goal of this PR is to have the most basic implementation of the ASGI Lifespan protocol possible without interfering with existing VMs that do not support the lifespan call.

### DIFF
--- a/vm_supervisor/models.py
+++ b/vm_supervisor/models.py
@@ -7,6 +7,7 @@ from dataclasses import dataclass
 from datetime import datetime
 from typing import NewType, Optional, Dict
 
+import msgpack
 from aleph_message.models import ProgramContent
 
 from .conf import settings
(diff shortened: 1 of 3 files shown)


vm_supervisor/models.py
```diff
@@ -120,6 +122,7 @@ async def create(
             await vm.start()
             await vm.configure()
             await vm.start_guest_api()
+            await self.run_code({"type": "lifespan.startup"})
```
hoh, line 125: Why not launch this event from the start of the VM init1.py ?

MHHukiewitz: You mean from here? https://github.com/aleph-im/aleph-vm/blob/main/vm_supervisor/vm/firecracker_microvm.py#L133

MHHukiewitz:
Superseded by #295

PR #295, opened 2023-04-20
Fix: ASGI processes did not support lifetime events

### LINKED ISSUE #293: ASGI Lifespan Protocol needs to be implemented

hoh:
ASGI specification https://asgi.readthedocs.io/en/latest/specs/lifespan.html

Fixes #293
Replaces #294


vm_supervisor/__main__.py
```diff
@@ -166,6 +166,7 @@ async def fake_read() -> bytes:
     settings.REUSE_TIMEOUT = 0.1
     for path in (
         "/",
+        "/lifetime",
```
MHHukiewitz, line 169: Would rename all occurrences of "lifetime" to "lifespan"

MHHukiewitz (commented):
LGTM, but would like to test it out locally tomorrow

hoh:
Needs a fix in aleph-client to work.

hoh:
Test by accessing `https://aleph-vm-lab-2.aleph.cloud/vm/test/lifespan`.

Commits squashed.

hoh:
Tests fail due to the issue in aleph-client.

MHHukiewitz:
@hoh Is there a reason why you did not implement the same for the Debian runtime?

hoh:
> @hoh Is there a reason why you did not implement the same for the Debian runtime?

I did. The init1.py of the Debian runtime is a symbolic link to the file I modified.

> It also fails in the GH action (but I cannot really make out why that is, I'm still struggling with making the VM supervisor run on my machine due to Conda, duh)

The reason why the tests fail still need investigation.

hoh:
This needs to be rebased due to recent changes.

MHHukiewitz:
Any idea why the tests might be failing?

runtimes/aleph-debian-11-python/init1.py
```diff
@@ -62,9 +73,9 @@ class ConfigurationPayload:
     input_data: bytes
     interface: Interface
     vm_hash: str
-    code: bytes = None
-    encoding: Encoding = None
-    entrypoint: str = None
+    code: bytes
+    encoding: Encoding
+    entrypoint: str
```
odesenfans, line 78: If these are always specified, okay. I don't see what these lines have to do with this PR though.

runtimes/aleph-debian-11-python/init1.py
```diff
@@ -171,7 +182,9 @@ def setup_input_data(input_data: bytes):
             os.system("unzip -q /opt/input.zip -d /data")
 
 
-def setup_authorized_keys(authorized_keys: List[str]) -> None:
+def setup_authorized_keys(authorized_keys: Optional[List[str]]) -> None:
+    if authorized_keys is None:
+        return
```
odesenfans, line 187: Same, seems unrelated to the PR. + I don't understand what's gained from passing an optional and having to check for None.

odesenfans (changes requested):
Lots of unrelated code, and the demo does not work.

runtimes/aleph-debian-11-python/init1.py
hoh: Moved to https://github.com/aleph-im/aleph-vm/pull/408

runtimes/aleph-debian-11-python/init1.py
hoh: Moved to https://github.com/aleph-im/aleph-vm/pull/408
This branch will need a rebase before continuing.

Issue #313, opened 2023-05-26
/usr/bin/ip not found on supervisord


aliel:
**Describe the bug**
after upgrading to version 0.2.6 

**To Reproduce**
Upgrade aleph-vm to 0.2.6 on debian 11.6

**Screenshots**

**Server (please complete the following information):**
 - OS:Debian 11.6
 - Version 11.6

PR #314, opened 2023-05-26
Fix ip command path #313

aliel:


vm_supervisor/network/interfaces.py
```diff
@@ -27,18 +28,19 @@ def host_ip(self) -> IPv4Interface:
     async def create(self):
         logger.debug("Create network interface")
 
-        run(["/usr/bin/ip", "tuntap", "add", self.device_name, "mode", "tap"])
+        ip_command = whereis_binary("ip")
```
hoh, line 31: Why not use the builtin `shutil.which` ? https://docs.python.org/3.9/library/shutil.html#shutil.which

aliel: done

hoh (approved):

PR #315, opened 2023-05-26
Fix: CI did not test on Ubuntu 20.04

hoh:


hoh:
We decided to drop support for Ubuntu 20.04 in #331

PR #326, opened 2023-06-13
Fix warnings from black, isort and mypy

hoh:
Depends on https://github.com/aleph-im/aleph-vm/pull/325

This fixes all basic warnings reported by `black` `isort` and `mypy` and uses `pathlib.Path` more generally.

### DIFF
--- a/vm_supervisor/storage.py
+++ b/vm_supervisor/storage.py
@@ -4,15 +4,11 @@ This module is in charge of providing the source code corresponding to a 'code i
 In this prototype, it returns a hardcoded example.
 In the future, it should connect to an Aleph node and retrieve the code from there.
 """
-import asyncio
 import hashlib
 import json
 import logging
-import os
 import re
-import subprocess
 import sys
-from os.path import isfile, join
 from pathlib import Path
 from shutil import make_archive
 from typing import Union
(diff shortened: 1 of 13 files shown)


vm_supervisor/utils.py
```diff
@@ -52,3 +53,23 @@ async def run_and_log_exception(coro: Coroutine):
 def create_task_log_exceptions(coro: Coroutine, *, name=None):
     """Ensure that exceptions running in coroutines are logged."""
     return asyncio.create_task(run_and_log_exception(coro), name=name)
+
+
+async def run_in_subprocess(
+    command: List[str], check: bool = True, stdin_input: Optional[bytes] = None
```
nesitor, line 59: The encoding parameter is needed for some commands related with devmapper

vm_supervisor/vm/firecracker/program.py
```diff
@@ -297,7 +299,11 @@ async def configure(self) -> None:
             resources=self.resources, drives=self.fvm.drives
         )
         interface: Interface = Interface.from_entrypoint(self.resources.code_entrypoint)
-        input_data: Optional[bytes] = read_input_data(self.resources.data_path)
+        # input_data: Optional[bytes] = read_input_data(self.resources.data_path)
```
nesitor, line 302: This comment should be removed?

vm_supervisor/storage.py
```diff
@@ -188,44 +195,34 @@ async def create_loopback_device(path: Path, read_only: bool = False) -> str:
     if read_only:
         command_args.append("--read-only")
     command_args.append(str(path))
-    loop_device = subprocess.run(
-        command_args, check=True, capture_output=True, encoding="UTF-8"
-    ).stdout.strip()
+    stdout = await run_in_subprocess(command_args)
+    loop_device = stdout.strip().decode()
```
nesitor, line 199: We need to decode to utf-8 here also

vm_supervisor/utils.py
hoh: Stdin and stdout are always expected to be bytes. Passing an extra parameter to ask to `encode()` / `decode()` would not make the code clearer or easier to manipulate.

PR #331, opened 2023-06-19
Fix: Drop support of Ubuntu 20.04 due to NFTables issues

hoh:
Incompatibilities in the version of NFTables prevent the current codebase from working correctly on that system.

Since the newer LTS, Ubuntu 22.04, shows no issue and many node operators upgraded to that system without any problem, we decided to drop support for Ubuntu 20.04.


doc/INSTALL-Ubuntu-22.04.md
```diff
@@ -0,0 +1,170 @@
+# Installing Aleph-VM on a server / Ubuntu 22.04 Jammy Jellyfish
+
+## 0. Introduction
+
+For production using official Debian packages.
```
odesenfans, line 5: A bit weird to mention "Debian" packages when installing on Ubuntu, don't you think? I know that's the name, but maybe "apt" packages could lead to less confusion?

doc/INSTALL-Ubuntu-22.04.md
```diff
@@ -0,0 +1,170 @@
...
+
+## 1. Requirements
+
+- A [supported Linux server](../vm_supervisor/README.md#1-supported-platforms)
+- A public domain name from a registrar and top level domain you trust. 
+
+In order to run an official Aleph.im Compute Resource Node (CRN), you will also need the following resources:
```
odesenfans, line 12: Nitpick: use the correct capitalization for aleph.im.

doc/INSTALL-Ubuntu-22.04.md
```diff
@@ -0,0 +1,170 @@
...
+## 2. Installation
+
+Run the following commands:
+
+First install the [VM-Connector](../vm_connector/README.md) using Docker:
+```shell
+sudo apt update
+sudo apt upgrade
+sudo apt install -y docker.io
+docker run -d -p 127.0.0.1:4021:4021/tcp --restart=always --name vm-connector alephim/vm-connector:alpha
```
odesenfans, line 34: Unrelated, but we should really version the connector image.

odesenfans (approved):

doc/INSTALL-Ubuntu-22.04.md
hoh: https://en.wikipedia.org/wiki/Deb_(file_format)
> Debian packages are standard [Unix](https://en.wikipedia.org/wiki/Unix) [ar archives](https://en.wikipedia.org/wiki/Ar_(Unix)) that (...)

doc/INSTALL-Ubuntu-22.04.md
hoh: And move it into a VM to remove the dependency on Docker

doc/INSTALL-Ubuntu-22.04.md
hoh: To fix in another nitpick branch.

PR #334, opened 2023-06-20
Fork `get_runtime_path`

hoh:
- Fix: When the users runs a test instance, it cannot use fake runtime Solution: Update general settings adding FAKE_DATA_INSTANCE variable to use the fake rootfs or the real one.
- Fix: Conditions differ for fake path to runtime or rootfs


nesitor:
This PR seems to be dependent on this one: https://github.com/aleph-im/aleph-vm/pull/336

PR #336, opened 2023-06-20
Fix: CLI launch of instances was limited

hoh:
Solution:

Use 2 commands: `--run-test-instance` to start an instance from the network and `--run-fake-instance` to start an instance from local files.

Add a setting with `--fake-instance-base` to specify the base of the local fake instance root filesystem.


vm_supervisor/__main__.py
```diff
@@ -270,7 +296,10 @@ def main():
         ALLOW_VM_NETWORKING=args.allow_vm_networking,
         FAKE_DATA_PROGRAM=args.fake_data_program,
         DEBUG_ASYNCIO=args.debug_asyncio,
+        FAKE_INSTANCE_BASE=args.fake_instance_base,
     )
+    if args.run_fake_instance:
+        settings.USE_FAKE_INSTANCE_BASE = True
```
nesitor, line 302:
```suggestion
        settings.FAKE_INSTANCE_BASE = True
```
You need to use the same setting name here: https://github.com/aleph-im/aleph-vm/blob/hoh-extend-instance-run/vm_supervisor/storage.py#L157

nesitor:
Also, I don't understand the purpose of `--fake-instance-base` argument, because it's a boolean and should be set to `True` if we use the `--run-fake-instance` argument.

vm_supervisor/__main__.py
hoh: I think that it would be better to have two settings:
`USE_FAKE_INSTANCE_BASE: bool` is a boolean that defines if the instance base from the message should be used or if a local file (rootfs.ext4) should be used as base.
`FAKE_INSTANCE_BASE` is the path to that rootfs base file, and has a sensible default for development and test purposes.

hoh:
`--fake-instance-base` is a string (path), not a boolean.

nesitor (approved):

PR #337, opened 2023-06-20
Feature: assign public IPv6 ranges to VMs

odesenfans:
Problem: we want each VM to be reachable on the internet, with no restriction in terms of ports, protocols, etc.

Solution: slice up the IPv6 range attributed to each CRN and allocate a subnet to each VM.

By default, a /124 IPv6 range is assigned to each VM. The IPv6 range is based on the item hash of the VM. For a VM with item hash `cafecafecafecafe...` on a CRN with 1111:2222:3333:4444::/64, the VM will be reachable at 1111:2222:3333:4444:1:cafe:cafe:caf1.

We use NDP proxying by default, as some cloud operators do not route the full IPv6 range to the server.

This PR does not manage the problem of isolating the IPv6 address of the host. It is advised to isolate the host range in its own /80 (ex: `1111:2222:3333:4444:0::/80`). This can be achieved by either running the following script in a tmux/screen session:

```shell
ip addr del 1111:2222:3333:4444::/64 dev eth0
ip addr add 1111:2222:3333:4444:0::/80 dev eth0
```

or by modifying the configuration in /etc/networking or /etc/netplan.

### DIFF
--- a/runtimes/aleph-debian-11-python/init1.py
+++ b/runtimes/aleph-debian-11-python/init1.py
@@ -31,6 +31,7 @@ import msgpack
 
 logger.debug("Imports finished")
 
+__version__ = "2.0.0"
 ASGIApplication = NewType("ASGIApplication", Any)
 
 
@@ -86,6 +87,7 @@ s.listen()
 # Send the host that we are ready
 s0 = socket.socket(socket.AF_VSOCK, socket.SOCK_STREAM)
 s0.connect((2, 52))
+s0.sendall(msgpack.dumps({"version": __version__}))
 s0.close()
 
 # Configure aleph-client to use the guest API
--- a/runtimes/aleph-debian-11-python/init1.py
+++ b/runtimes/aleph-debian-11-python/init1.py
@@ -65,7 +65,9 @@ class ConfigurationPayload:
     encoding: Encoding = None
     entrypoint: str = None
     ip: Optional[str] = None
+    ipv6: Optional[str] = None
     route: Optional[str] = None
+    ipv6_gateway: Optional[str] = None
     dns_servers: List[str] = field(default_factory=list)
     volumes: List[Volume] = field(default_factory=list)
     variables: Optional[Dict[str, str]] = None
@@ -108,7 +110,11 @@ def setup_variables(variables: Optional[Dict[str, str]]):
 
 
 def setup_network(
-    ip: Optional[str], route: Optional[str], dns_servers: Optional[List[str]] = None
+    ip: Optional[str],
+    ipv6: Optional[str],
+    route: Optional[str],
+    ipv6_gateway: Optional[str],
+    dns_servers: Optional[List[str]] = None,
 ):
     """Setup the system with info from the host."""
     dns_servers = dns_servers or []
@@ -116,29 +122,41 @@ def setup_network(
         logger.info("No network interface eth0")
         return
 
-    if not ip:
+    if not (ip or ipv6):
         logger.info("No network IP")
         return
 
-    logger.debug("Setting up networking")
+    # Configure loopback networking
     system("ip addr add 127.0.0.1/8 dev lo brd + scope host")
     system("ip addr add ::1/128 dev lo")
     system("ip link set lo up")
-    if "/" in ip:
-        # Forward compatibility with future supervisors that pass the mask with the IP.
-        system(f"ip addr add {ip} dev eth0")
-    else:
-        logger.warning(
-            "Not passing the mask with the IP is deprecated and will be unsupported"
-        )
-        system(f"ip addr add {ip}/24 dev eth0")
-    system("ip link set eth0 up")
 
-    if route:
-        system(f"ip route add default via {route} dev eth0")
-        logger.debug(f"IP and route set: {ip} via {route}")
-    else:
-        logger.warning("IP set with no network route")
+    if ip:
+        logger.debug("Setting up IPv4")
+
+        # Forward compatibility with future supervisors that pass the mask with the IP.
+        if "/" not in ip:
+            logger.warning(
+                "Not passing the mask with the IP is deprecated and will be unsupported"
+            )
+            ip = f"{ip}/24"
+
+        system(f"ip addr add {ip} dev eth0")
+
+        if route:
+            system(f"ip route add default via {route} dev eth0")
+            logger.debug(f"IP and route set: {ip} via {route}")
+        else:
+            logger.warning("IPv4 set with no network route")
+
+    if ipv6:
+        logger.debug("Setting up IPv6")
+        system(f"ip addr add {ipv6} dev eth0")
+        system(f"ip -6 route add default via {ipv6_gateway} dev eth0")
+        logger.debug(f"IPv6 setup to address {ipv6}")
+
+    if ip or ipv6:
+        system("ip link set eth0 up")
 
     with open("/etc/resolv.conf", "wb") as resolvconf_fd:
         for server in dns_servers:
@@ -478,7 +496,13 @@ def setup_system(config: ConfigurationPayload):
 
     setup_variables(config.variables)
     setup_volumes(config.volumes)
-    setup_network(config.ip, config.route, config.dns_servers)
+    setup_network(
+        ip=config.ip,
+        ipv6=config.ipv6,
+        route=config.route,
+        ipv6_gateway=config.ipv6_gateway,
+        dns_servers=config.dns_servers,
+    )
     setup_input_data(config.input_data)
     logger.debug("Setup finished")
 
--- a/vm_supervisor/network/hostnetwork.py
+++ b/vm_supervisor/network/hostnetwork.py
@@ -1,26 +1,160 @@
 import logging
+from ipaddress import IPv6Network
 from pathlib import Path
+from typing import Optional, Protocol
 
+from aleph_message.models import ItemHash
+
+from vm_supervisor.conf import IPv6AllocationPolicy
+
+from ..vm.vm_type import VmType
 from .firewall import initialize_nftables, setup_nftables_for_vm, teardown_nftables
 from .interfaces import TapInterface
 from .ipaddresses import IPv4NetworkWithInterfaces
+from .ndp_proxy import NdpProxy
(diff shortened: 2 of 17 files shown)


vm_supervisor/network/hostnetwork.py
```diff
@@ -1,26 +1,144 @@
...
 
 from .firewall import initialize_nftables, setup_nftables_for_vm, teardown_nftables
 from .interfaces import TapInterface
 from .ipaddresses import IPv4NetworkWithInterfaces
+from .ndp_proxy import NdpProxy
 
 logger = logging.getLogger(__name__)
 
 
+def _read_proc_file_value(config_file: Path) -> int:
```
hoh, line 18: This does not seem to be specific to proc files
```suggestion
def _read_file_as_int(config_file: Path) -> int:
```

vm_supervisor/network/hostnetwork.py
```diff
@@ -1,26 +1,144 @@
...
+    allocation_policy: IPv6AllocationPolicy, address_pool: str, subnet_prefix: int
+) -> IPv6Allocator:
+    allocator_cls = (
+        StaticIPv6Allocator
+        if allocation_policy == IPv6AllocationPolicy.static
+        else DynamicIPv6Allocator
+    )
+    return allocator_cls(
+        ipv6_range=IPv6Network(address_pool), subnet_prefix=subnet_prefix
+    )
```
hoh, line 98: This pattern of a `maker` function does not seem very pythonic... 🤔

Static methods come to my mind first.
```python
class IPv6Allocator:

    @staticmethod
    def from_policy(allocation_policy: IPv6AllocationPolicy, address_pool: str, subnet_prefix: int
) -> IPv6Allocator:
        ...
```

vm_supervisor/network/hostnetwork.py
```diff
@@ -1,26 +1,144 @@
...
+
+
 def get_ipv4_forwarding_state() -> int:
-    """Reads the current ipv4 forwarding setting from the hosts, converts it to int and returns it"""
-    return int(Path("/proc/sys/net/ipv4/ip_forward").read_text())
+    """Reads the current IPv4 forwarding setting from the host, converts it to int and returns it"""
+    return _read_proc_file_value(Path("/proc/sys/net/ipv4/ip_forward"))
+
+
+def get_ipv6_forwarding_state() -> int:
```
hoh, line 27: Since that file can only contain 0 or 1, should we use booleans instead ?

vm_supervisor/network/hostnetwork.py
```diff
@@ -1,26 +1,144 @@
...
+            )
+        if subnet_prefix < 124:
+            raise ValueError("The IPv6 subnet prefix cannot be larger than /124.")
+
+        self.ipv6_range = ipv6_range
+        self.subnet_prefix = subnet_prefix
+
+    def allocate_vm_ipv6_subnet(self, vm_id: int, vm_hash: ItemHash) -> IPv6Network:
+        ipv6_elems = self.ipv6_range.exploded.split(":")[:4]
+        ipv6_elems += ["1"]  # Magic number for microVMs
```
hoh, line 61: Magic number ?

vm_supervisor/network/hostnetwork.py
```diff
@@ -1,26 +1,144 @@
...
+        ipv6_elems += [vm_hash[0:4], vm_hash[4:8], vm_hash[8:11] + "0"]
```
hoh, line 62: Can you document what this does in a comment ?

vm_supervisor/network/hostnetwork.py
hoh, line 98: Storing the class into a variable here does not seem to make the code simpler to read...

```suggestion
    if allocation_policy == IPv6AllocationPolicy.static:
        return StaticIPv6Allocator(ipv6_range=IPv6Network(address_pool), subnet_prefix=subnet_prefix)
    else:
        return DynamicIPv6Allocator(ipv6_range=IPv6Network(address_pool), subnet_prefix=subnet_prefix)
```

vm_supervisor/network/interfaces.py
```diff
@@ -40,12 +58,25 @@ async def create(self):
...
         run([ip_command, "link", "set", self.device_name, "up"])
+        self.ndp_proxy.add_range(self.device_name, ipv6_gateway.network)
         logger.debug(f"Network interface created: {self.device_name}")
 
     async def delete(self) -> None:
         """Asks the firewall to teardown any rules for the VM with id provided.
         Then removes the interface from the host."""
         logger.debug(f"Removing interface {self.device_name}")
         await asyncio.sleep(0.1)  # Avoids Device/Resource busy bug
+        self.ndp_proxy.delete_range(self.device_name)
```
hoh, line 81: Should this go before the `asyncio.sleep` ?

vm_supervisor/network/ndp_proxy.py
```diff
@@ -0,0 +1,56 @@
...
+
+@dataclass
+class NdpRule:
+    address_range: IPv6Network
+
+
+class NdpProxy:
+    def __init__(self, host_network_interface: str):
+        self.host_network_interface = host_network_interface
+        self.interface_address_range_mapping = {}
```
hoh, line 29: What is the goal of this (untyped) dict ?

runtimes/aleph-debian-11-python/init1.py
```diff
@@ -107,35 +109,52 @@ def setup_variables(variables: Optional[Dict[str, str]]):
 
 
 def setup_network(
-    ip: Optional[str], route: Optional[str], dns_servers: Optional[List[str]] = None
+    ip: Optional[str],
+    ipv6: Optional[str],
+    route: Optional[str],
+    ipv6_gateway: Optional[str],
```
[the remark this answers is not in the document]
odesenfans: Because that's the correct name IMO. I didn't change the name of route for backwards compatibility, but gateway is much clearer I think.

runtimes/aleph-debian-11-python/init1.py
```diff
@@ -107,35 +109,52 @@ def setup_variables(variables: Optional[Dict[str, str]]):
...
+            ip = f"{ip}/24"
+
         system(f"ip addr add {ip} dev eth0")
-    else:
-        logger.warning("Not passing the mask with the IP is deprecated and will be unsupported")
-        system(f"ip addr add {ip}/24 dev eth0")
-    system("ip link set eth0 up")
+        system("ip link set eth0 up")
 
     if route:
```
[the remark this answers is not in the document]
odesenfans: Because I wasn't sure it was the right place. I wanted to enable the interface only after all addresses were set up, so `if ip` is not necessarily the best place. The only "problem" here is that the interface will be up even if the VM does not support networking, but that seems like a minor issue.

vm_supervisor/network/hostnetwork.py
odesenfans: We might, but then we have to convert between booleans and integers. I kept the previous behaviour, feel free to change it in another PR.

vm_supervisor/network/hostnetwork.py
odesenfans: As described [here](https://community.aleph.im/t/vm-ipv6-routing/63/3) I used the xxxx:xxxx:xxxx:xxxx:1::/80 subnet for microVMs. We should add a parameter to the function to support the different VM types.

vm_supervisor/network/ndp_proxy.py
odesenfans: This is a cache to avoid having to read the `/etc/ndppd.conf` file. With this, when updating the config, we can just overwrite the file entirely. And when removing a mapping, we can just pop the corresponding entry from this dictionary and then regenerate the file.

Easier (and probably faster) than parsing the file.

runtimes/aleph-debian-11-python/init1.py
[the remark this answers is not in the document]
hoh: Got it, thanks.

vm_supervisor/network/interfaces.py
odesenfans: I was wondering what this sleep fixes? At first glance it just seems like this occurs because we don't down the device before deleting it.

runtimes/aleph-debian-11-python/init1.py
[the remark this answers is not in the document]
odesenfans: Solved in your PR, marking this as resolved.

vm_supervisor/network/hostnetwork.py
odesenfans: We now use the VM type to determine this number, the discussion is moot.

vm_supervisor/network/hostnetwork.py
odesenfans: As mentioned on Telegram: I prefer this pattern for factories that generate different children classes based on the input and/or factories that take multiple arguments. We're definitely in both cases here so I don't see a clear point to change, unless you want to enforce a specific style. I do think this pattern is as pythonic as factory methods.

vm_supervisor/vm/firecracker/instance.py
```diff
@@ -170,15 +170,18 @@ def _create_network_file(self) -> bytes:
...
+        ipv6_gateway = self.get_vm_ipv6_gateway()
 
         network = {
             "network": {
                 "ethernets": {
                     "eth0": {
                         "dhcp4": False,
                         "dhcp6": False,
-                        "addresses": [str(ip)],
+                        "addresses": [str(ip), ipv6],
```
nesitor, line 182:
```suggestion
                        "addresses": [ip, ipv6],
```
Remove the unneeded `str` method, but we need to ensure that any IP is received

nesitor (approved):
:clap: :rocket:

hoh:
I cannot access the index page anymore over IPv6 when using this branch.

`curl -4 -i "https://aleph-vm-lab-2.aleph.cloud/` works

`curl -6 -i "https://aleph-vm-lab-2.aleph.cloud/` hangs

hoh (approved):

PR #271, opened 2023-02-15
Added intermediate loading page if the VM is not ready yet

nesitor:

PR #338, opened 2023-06-21
Add VM loading page

nesitor:
Problem: When the VM takes a bit long to load, the final user don't see anything, only the loading page.

Solution: Create a loading intermediate page that will be refreshing until the VM is up.

Note: This PR is a replacement of the old PR https://github.com/aleph-im/aleph-vm/pull/271, updated with last main branch changes and with Aleph style design (https://www.figma.com/file/qNnjyRzhXcmj6nwbL9XP5K/ALEPH-Cloud-Website?type=design&node-id=1761%3A25399&t=pgoxZgyAMG1lbwNn-1)


runtimes/aleph-debian-11-python/init1.py
```diff
@@ -325,6 +325,23 @@ async def make_request(session, scope):
     return headers, body
 
 
+def show_loading():
+    with open("/root/loading.html", "r") as f:
```
hoh, line 329: Why not read this as bytes ?

runtimes/aleph-debian-11-python/init1.py
```diff
@@ -325,6 +325,23 @@ async def make_request(session, scope):
...
+        body = {
+            "body": f.read()
```
hoh, line 331: Since the entire body is loaded in memory, would it not be easier to skip the file descriptor ?

```suggestion
            "body": Path(""/root/loading.html").read_bytes()
```

runtimes/aleph-debian-11-python/loading.html
```diff
@@ -0,0 +1,346 @@
+<html lang="en">
```
hoh, line 1: This is a pretty large file, would it make sense to load the CSS and SVG from a CDN ? (ex: IPFS)

runtimes/aleph-debian-11-python/init1.py
```diff
@@ -340,7 +357,7 @@ async def run_executable_http(scope: dict) -> Tuple[Dict, Dict, str, Optional[by
                 headers, body = await make_request(session, scope)
             except aiohttp.ClientConnectorError:
                 if tries > 20:
-                    raise
```
hoh, line 343: Should the error still appear in the logs ?

runtimes/aleph-debian-11-python/init1.py
nesitor: As I commented in the previous comment, I think that is better to use string format instead bytes, but yes, if you want I can skip the file descriptor by `Path(""/root/loading.html").read_text()`, but I don't see the advantage.

runtimes/aleph-debian-11-python/loading.html
nesitor: Yes, we can move it to an IPFS CDN, but for simplicity for the users, I think that is better to maintain all the imports inside the file, and allow our users to know better how it works and how to customize.

runtimes/aleph-debian-11-python/init1.py
nesitor: I don't think so, because cannot be an error, can be that the VM needs some time to be ready. When we implement the VM logs, we will see there if the VM is loading well or not.

runtimes/aleph-debian-11-python/init1.py
hoh: Simpler code is better. The overhead created by using a manager would be avoided.

runtimes/aleph-debian-11-python/init1.py
hoh: OK

runtimes/aleph-debian-11-python/init1.py
nesitor: Okay, I just changed it

hoh (approved):

PR #344, opened 2023-06-27
Fix: make runtime forward-compatible with new parameters

odesenfans:
Problem: the existing runtime is not compatible with the addition of new parameters in `ConfigurationPayload`. Adding a parameter will result in an invalid argument error with older runtimes.

Solution: add a factory method that takes a dictionary and filters out useless fields.


hoh:
We prefer using another approach with more validation.

PR #350, opened 2023-06-29
Fix: Root login with shared password is insecure

hoh:
Problem: With the incoming support of IPv6, allowing login into virtual machines using SSH with a password becomes unacceptable.

Solution: Remove SSH password login from runtimes, as well as the root default password.


hoh:
image

PR #352, opened 2023-07-02
Add default NODE_PATH env

MHHukiewitz:
Include `/opt/node_modules` in the module resolution for node.js in order to allow the [Dependency Volume Builder](https://github.com/aleph-im/aleph-cloud-tools/tree/main/dependency_builder) to target this directory when building a dependency volume.

### DIFF
--- a/vm_supervisor/README.md
+++ b/vm_supervisor/README.md
@@ -118,6 +118,10 @@ or in debug mode:
 ```shell
 python3 -m vm_supervisor -vv --system-logs
 ```
+or in test mode to run a single function:
+```shell
+python3 -m vm_supervisor -vv --system-logs -f examples/example_fastapi
+```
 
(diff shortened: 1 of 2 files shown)


vm_supervisor/README.md
```diff
@@ -118,6 +118,10 @@ or in debug mode:
 ```shell
 python3 -m vm_supervisor -vv --system-logs
 ```
+or in test mode to run a single function:
+```shell
+python3 -m vm_supervisor -vv --system-logs -f examples/example_fastapi
+```
```
hoh, line 124: This should be the topic of another pull request. It would not work as is as using the "fake program" mode also requires building a runtime locally as well as an example squashfs volume. That procedure is documented in `/tutorials/TESTING.md`

hoh (approved):

PR #363, opened 2023-07-04
Fix: Interface must be up before route creation

hoh:


runtimes/aleph-debian-11-python/init1.py
```diff
@@ -154,13 +157,14 @@ def setup_network(
 
     if ipv6:
         logger.debug("Setting up IPv6")
-        system(f"ip addr add {ipv6} dev eth0")
+        system(f"ip -6 addr add {ipv6} dev eth0")
+
+        # Interface must be up before a route can use it
+        system("ip -6 link set eth0 up")
```
odesenfans, line 163: TBD if setting the interface up before assigning addresses could do harm in any way whatsoever. It would be cleaner code-wise to set the interface up after setting up the loopback address, whether IPv4 or IPv6 are enabled.

hoh: I refactored the approach to setup the IP addresses in order to remove duplication. This will also help in the future when adding additional links between VMs.

odesenfans (approved):
Okay, looks cleaner!

PR #364, opened 2023-07-05
Fix: VM does not boot if no SSH key is specified

odesenfans:
Problem: The authorized_keys file could not be generated if the `config.authorized_keys` settings was None because it was used in a list comprehension.

Solution: check for None early.


hoh:
Should we make the authorized keys a `List` with default `[]` instead of allowing `None` ?

hoh (approved):

PR #368, opened 2023-07-06
Fix: diagnostic checks using aleph SDK were broken

odesenfans:
Problem: the Python SDK does not check environment variables when configuring the API client and VM cache objects like the previous implementation did.

Solution: pass the correct configuration explicitly when constructing these objects in the FastAPI example.

### DIFF
--- a/examples/example_fastapi/main.py
+++ b/examples/example_fastapi/main.py
@@ -30,7 +30,14 @@ logger.debug("imports done")
 
 http_app = FastAPI()
 app = AlephApp(http_app=http_app)
-cache = VmCache()
+cache = VmCache(
+    session=aiohttp.ClientSession(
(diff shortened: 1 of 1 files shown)


examples/example_fastapi/main.py
```diff
@@ -30,7 +30,14 @@
 
 http_app = FastAPI()
 app = AlephApp(http_app=http_app)
-cache = VmCache()
+cache = VmCache(
+    session=aiohttp.ClientSession(
+        connector=aiohttp.UnixConnector(path=os.environ["ALEPH_API_UNIX_SOCKET"])
```
hoh, line 35: The user of the SDK should not have to worry about that, the same code should work without a change on the developer's system and inside a microvm.

hoh:
This has been solved in the SDK, and a new version of the SDK should be used.

PR #370, opened 2023-07-11
Create Ubuntu rootfs

nesitor:
Updated creation script to create a BTRFS image of Ubuntu filesystem.


runtimes/instance-rootfs/create-debian-12-disk.sh
```diff
@@ -0,0 +1,51 @@
...
+mkdir -p "$MOUNT_ORIGIN_DIR"
+mkdir -p "$MOUNT_DIR"
+
+# Download Debian image
+echo "Downloading Debian 12 image"
+curl -L "$IMAGE_URL" -o "$IMAGE_NAME"
+
+# Create rootfs.btrfs file
+echo "Creating rootfs.btrfs file"
+dd if=/dev/zero of="$ROOTFS_FILE" bs=1G count=2
```
hoh, line 28: We should avoid using `dd` in our scripts when possible: https://eklitzke.org/the-cult-of-dd

```suggestion
# Allocate a 1GB file
fallocate -l 1G "$ROOTFS_FILE"
```

hoh: Why 1GB ? Is there a way to shrink it to the minimum required ?

runtimes/instance-rootfs/create-ubuntu-22-04-disk.sh
```diff
@@ -0,0 +1,35 @@
...
+mkdir -p "$MOUNT_DIR"
+mkdir -p "$ROOTFS_DIR"
+
+# Download Ubuntu image
+echo "Downloading Ubuntu 22.04 image"
+curl -L --remote-name "$IMAGE_URL"
+
+# Create rootfs.btrfs file
+echo "Creating rootfs.btrfs file"
+dd if=/dev/zero of="$ROOTFS_FILE" bs=1M count=1400
```
hoh, line 26: Same comment about `dd`

runtimes/instance-rootfs/create-debian-12-disk.sh
```diff
@@ -0,0 +1,51 @@
...
+mount "$ROOTFS_FILE" "$MOUNT_DIR"
+
+# Extract Debian image
+echo "Extracting Debian 12 image"
+tar xvf "$IMAGE_NAME"
+
+# Mount first partition of Debian Image
+LOOPDISK=$(losetup --find --show $IMAGE_RAW_NAME)
+partx -u $LOOPDISK
+mount "$LOOPDISK"p1 "$MOUNT_ORIGIN_DIR"
```
hoh, line 39: Why not using `btrfs-convert $LOOPDISK"p1` instead of extracting the files and putting them in a new filesystem ?

nesitor:
Removed use of `dd` and replaced by `fallocate` tool

runtimes/instance-rootfs/create-debian-11-disk.sh
```diff
@@ -0,0 +1,51 @@
+#!/bin/bash
+
+set -euf
```
odesenfans, line 3: Add `-o pipefail` ?

runtimes/instance-rootfs/create-debian-11-disk.sh
```diff
@@ -0,0 +1,51 @@
...
+mkdir -p "$MOUNT_ORIGIN_DIR"
+mkdir -p "$MOUNT_DIR"
+
+# Download Debian image
+echo "Downloading Debian 11 image"
+curl -L "$IMAGE_URL" -o "$IMAGE_NAME"
+
+# Allocate 2GB rootfs.btrfs file
+echo "Allocate 2GB rootfs.btrfs file"
+fallocate -l 1G "$ROOTFS_FILE"
```
odesenfans, line 28: Seems like you allocate 1GB, despite what you announce above.

runtimes/instance-rootfs/create-ubuntu-22-04-disk.sh
```diff
@@ -0,0 +1,38 @@
...
+mount "$ROOTFS_FILE" "$MOUNT_DIR"
+
+# Extract Ubuntu image to rootfs
+echo "Extracting Ubuntu 22.04 image"
+tar xvf "$IMAGE_NAME" -C "$MOUNT_DIR"
+
+# Cleanup and unmount
+umount "$MOUNT_DIR"
+rm -rf "$ROOTFS_DIR"
+rm "$IMAGE_NAME"
```
odesenfans, line 38: Are there many differences between the 3 scripts. It seems to me that they are extremely similar. Maybe you could factorize the flesh of the files as a function and import it/use it in these scripts?

Like

```
# create-rootfs.sh
function create_rootfs()
{
  image_url="$1"
  size="$2"
  ... # Download etc
}

# create-ubuntu-22-04-disk.sh
source ./create-rootfs.sh

create_rootfs("https://cloud-images.ubuntu.com/jammy/current/jammy-server-cloudimg-amd64-root.tar.xz", 1400M)
```

It's ok if we don't do it now.

odesenfans (changes requested):
Nothing major.

runtimes/instance-rootfs/create-ubuntu-22-04-disk.sh
nesitor: Yes, will be good to optimize it, but I think that we can do it later.

runtimes/instance-rootfs/create-debian-11-disk.sh
odesenfans: It makes sure that any failure in a piped operation (ex: `echo "hello" | cat`) causes the script to stop.

runtimes/instance-rootfs/create-debian-11-disk.sh
```diff
@@ -0,0 +1,51 @@
+#!/bin/bash
+
+set -euf
+
+# Variables
+ROOTFS_FILE="./rootfs.btrfs"
```
odesenfans, line 6: Could you use a different name for each distribution?

runtimes/instance-rootfs/create-debian-11-disk.sh
nesitor: Okay, but here don't make sense IMO, because we don't use any pipe command.

odesenfans (approved):

PR #378, opened 2023-07-13
Feature: allow IPv6 DNS

1yam:


vm_supervisor/vm/firecracker/program.py
```diff
@@ -388,9 +390,13 @@ async def _setup_configuration(
         ipv6 = self.get_vm_ipv6()
         ipv6_gateway = self.get_vm_ipv6_gateway()
 
-        if not settings.DNS_NAMESERVERS:
+        dns_servers = settings.DNS_NAMESERVERS
```
odesenfans, line 393: I think it would be great to have a `dns_servers()` method in `Settings` that uses `ipaddress` to perform the conversion to `IPv4/IPv6 address objects only once.

odesenfans (changes requested):

vm_supervisor/conf.py
```diff
@@ -285,6 +285,19 @@ def display(self) -> str:
             f"{attribute:<27} = {value}" for attribute, value in attributes.items()
         )
 
+    def dns_servers(self, ipv4_only: bool) -> Optional[List[str]]:
```
hoh, line 288: Why is there no explaination on what this function does as docstring ?

vm_supervisor/conf.py
```diff
@@ -285,6 +285,19 @@ def display(self) -> str:
...
+        dns_servers = self.DNS_NAMESERVERS
+        if not dns_servers:
+            raise ValueError("Invalid configuration: DNS nameservers missing")
+
+        # Apply DNS IPv6 filtering here if needed
```
hoh, line 293: What does "DNS IPv6 filtering" mean and why do it ?

vm_supervisor/vm/firecracker/program.py
```diff
@@ -387,9 +388,7 @@ async def _setup_configuration(
         route = self.get_vm_route()
         ipv6 = self.get_vm_ipv6()
         ipv6_gateway = self.get_vm_ipv6_gateway()
-
-        if not settings.DNS_NAMESERVERS:
-            raise ValueError("Invalid configuration: DNS nameservers missing")
+        dns_servers = settings.dns_servers(False)
```
hoh, line 391: How can a node operator disable DNS servers for tests ?

olethanh (changes requested):
Sorry I'm new here so I'm missing a lot of context, could you please explain to me what problem this is resolving and how the new feature is used?

odesenfans:
@olethanh This re-enables IPv6 DNS inside VMs. For a while, VMs only supported IPv4 networking and could therefore not reach IPv6 DNS. As we copied the list of DNS from the host, it could happen that the host used an IPv6 DNS, meaning that the VM would not be able to resolve domain names. Long story short, we simply filtered out IPv6 DNS. This PR removes this workaround.

olethanh:
@odesenfans Thanks for the answer but I still don't understand. I cannot find in the code in the main branch where the filtering was done before and where it is removed now?
Also shouldn't the filtering be done on the basis of whether ipv6 is enabled for the VM then?

vm_supervisor/conf.py
```diff
@@ -308,9 +308,6 @@ def setup(self):
         os.makedirs(self.EXECUTION_LOG_DIRECTORY, exist_ok=True)
         os.makedirs(self.PERSISTENT_VOLUMES_DIR, exist_ok=True)
 
-        if not self.NETWORK_INTERFACE:
-            self.NETWORK_INTERFACE = get_default_interface()
-
```
olethanh, line 313: Is that intender or was this removed by error ?

odesenfans:
@olethanh The filtering is performed [here](https://github.com/aleph-im/aleph-vm/blob/a3d5b52addd93747e0886b13f9ba3d8d9a794722/src/aleph/vm/conf.py#L67). Seems like a lot of code has moved since this PR was opened so a rebase is much needed.

Regarding your other comment, you are right that this should be done on a per-VM config, checking first if the supervisor configuration allows for it and then whether the VM runtime supports IPv6.

MHHukiewitz:
Superseded by #455

PR #381, opened 2023-07-14
Feature: Implement Instance snapshots

nesitor:
Feature: Upload VM snapshots using Aleph STORE message.
Feature: Restore existing snapshots as STORE Aleph message.

### DIFF
--- a/vm_supervisor/ipfs.py
+++ b/vm_supervisor/ipfs.py
@@ -1,0 +1,66 @@
+import logging
+from pathlib import Path
+from typing import Optional
+
+import aioipfs
+from aleph.sdk.chains.common import get_fallback_private_key
+from aleph.sdk.chains.ethereum import ETHAccount
+from aleph.sdk.client import AuthenticatedAlephClient
+from aleph.sdk.types import StorageEnum
+from aleph_message.models import ItemHash
+from aleph_message.status import MessageStatus
+
+logger = logging.getLogger(__name__)
+
+
+async def ipfs_upload_file(path: Path) -> str:
+    """Push a file to the IPFS service."""
+    # Use local ipfs node by default. TODO: Add a general setting to allow to customize it
+    client = aioipfs.AsyncIPFS(host="localhost", port=5001)
+    ipfs_hash = None
+    async for added_file in client.add(str(path), recursive=True):
+        ipfs_hash = added_file["Hash"]
+        logger.debug(f"Pushed {path} file to IPFS with hash {ipfs_hash}")
+
+    await client.close()
+    assert ipfs_hash, "File not pinned on IPFS"
+
+    return ipfs_hash
+
+
+async def ipfs_remove_file(name: str, ipfs_hash: str) -> None:
+    """Remove a file to the IPFS service."""
+    # Use local ipfs node by default. TODO: Add a general setting to allow to customize it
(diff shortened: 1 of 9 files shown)


vm_supervisor/snapshots.py
```diff
@@ -29,9 +34,13 @@ def __init__(self, path: Path, algorithm: SnapshotCompressionAlgorithm):
...
-    async def upload(self) -> ItemHash:
-        # TODO: Upload snapshots to Aleph Network
-        pass
+    async def upload(self, vm_hash: ItemHash) -> ItemHash:
+        account = get_fallback_account()
+        async with AuthenticatedAlephClient(account=account, api_server="https://official.aleph.cloud") as client:
+            message, status = await client.create_store(
+                file_path=self.path, storage_engine=StorageEnum.ipfs, sync=True, ref=f"snapshot_{vm_hash}"
+            )
+            assert status == MessageStatus.PROCESSED
```
odesenfans, line 43: I think it's important here to delete old snapshots from the network as well. A call to `client.forget()`would make a lot of sense when we delete the previous snapshot. We'll implement some depth to this (keep the last N snapshots) later.

nesitor: Good point, thanks, I just Implemented it.

vm_supervisor/ipfs.py
```diff
@@ -0,0 +1,66 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+async def ipfs_upload_file(path: Path) -> str:
+    """Push a file to the IPFS service."""
+    # Use local ipfs node by default. TODO: Add a general setting to allow to customize it
+    client = aioipfs.AsyncIPFS(host="localhost", port=5001)
+    ipfs_hash = None
+    async for added_file in client.add(str(path), recursive=True):
```
hoh, line 21: Why is this recursive if the goal is to upload a single file ?

vm_supervisor/ipfs.py
```diff
@@ -0,0 +1,66 @@
...
+            ref=ref,
+        )
+        assert status != MessageStatus.REJECTED
+        return message.item_hash
+
+
+async def send_forget_ipfs_message(
+    item_hash: ItemHash, reason: Optional[str] = ""
+) -> None:
+    pkey = get_fallback_private_key()
```
hoh, line 59: This function should never be used in production: the creation of the key should be explicit and during installation. A key that is missing should cause an error.

vm_supervisor/ipfs.py
nesitor: Because it accepts different files, and it return different results.

vm_supervisor/ipfs.py
nesitor: How I can get it?

vm_supervisor/ipfs.py
hoh: In practice I only see it used in one place, `snapshot_hash = await ipfs_upload_file(self.path)`, with a single file.

nesitor: Yes, I mean that the `client.add` method call, it accepts different files, and return different results.

hoh:
I am closing this PR since we are facing performance issues in this implementation and want to evaluate the use of Qcow2 as a replacement.

PR #385, opened 2023-07-15
Feature: install Kubo on CRNs

odesenfans:


nesitor (approved):
LGTM and also tested that works perfect.

PR #397, opened 2023-07-27
Solved some VM errors

nesitor:
Solved shutdown errors that prevent the VM to try again to be scheduled on the CRN.
Errors:
- Property `_unix_socket` is not used on instances yet, so initialize to `None` by default.
- If the VM don't respond to a ping, should forget the VM to be able to try to power up later.
- Improvement: Use `fallocate` instead `dd` command to create empty files.
- Fixed cloud-init issue that only executes the configuration on the first boot. Now, if the VM ID has changed, the cloud-init configuration re-runs again.
- Ensure that the allocation process don't stop, either, on a VM failure.

### DIFF
--- a/vm_supervisor/run.py
+++ b/vm_supervisor/run.py
@@ -49,7 +49,6 @@ async def build_event_scope(event) -> Dict[str, Any]:
 
 
 async def create_vm_execution(vm_hash: ItemHash) -> VmExecution:
-    execution: Optional[VmExecution] = None
     message, original_message = await load_updated_message(vm_hash)
     pool.message_cache[vm_hash] = message
 
@@ -81,7 +80,7 @@ async def create_vm_execution(vm_hash: ItemHash) -> VmExecution:
         logger.exception(error)
         pool.forget_vm(vm_hash=vm_hash)
 
-    if not execution or execution.vm:
+    if not execution.vm:
         raise ValueError("The VM has not been created")
 
     return execution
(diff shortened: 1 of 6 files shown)


odesenfans:
Would be nice if you could list the problems you solved in this PR in the description. I see that you solved quite a few of them and if would be nice to track them.

nesitor:
> Would be nice if you could list the problems you solved in this PR in the description. I see that you solved quite a few of them and if would be nice to track them.

I just updated the description with the error list.

vm_supervisor/run.py
```diff
@@ -75,6 +76,9 @@ async def create_vm_execution(vm_hash: ItemHash) -> VmExecution:
         logger.exception(error)
         pool.forget_vm(vm_hash=vm_hash)
         raise HTTPInternalServerError(reason="Error during runtime initialisation")
+    except HostNotFoundError as error:
+        logger.exception(error)
+        pool.forget_vm(vm_hash=vm_hash)
```
hoh, line 81: Why don't you raise an exception here ?

nesitor: Because if you raise the exception, the supervisor crashes and don't forget the VM, but needs to forget it first.

hoh: And raising the exception after launching `forget_vm` does not work, as for the other exceptions that are handled ?

nesitor: If we raise an exception after, the rest of the instances will not start it's deploying, and there are sometimes that the instance takes a bit more time to start.

hoh: The fact that the exception blocks other instances from starting is a good point, but it should not be handled here since any other exception will also block that.

PR #408, opened 2023-09-20
Cleanup: Fix Mypy errors in init1.py

hoh:
The only Mypy error left is related to the use of the type `ASGIApplication`, defined as:

```python
ASGIApplication = NewType("ASGIApplication", Any)
```

```
init1.py:46: error: Argument 2 to NewType(...) must be subclassable (got "Any")  [valid-newtype]
```

I did not find any easy solution for that. It does not seem blocking.

### DIFF
--- a/runtimes/aleph-debian-11-python/init1.py
+++ b/runtimes/aleph-debian-11-python/init1.py
@@ -24,7 +24,7 @@ from enum import Enum
 from io import StringIO
 from os import system
 from shutil import make_archive
-from typing import Any, AsyncIterable, Dict, List, NewType, Optional, Tuple, Union
+from typing import Any, AsyncIterable, Dict, List, NewType, Optional, Tuple, Union, cast
 
 import aiohttp
 import msgpack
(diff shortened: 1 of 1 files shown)


runtimes/aleph-debian-11-python/init1.py
MHHukiewitz, line 1: Can't you just replace [Line 35](https://github.com/aleph-im/aleph-vm/pull/408/files#diff-9a70c1e5d16c2b448e42ea1eeaa4235a1f8272a8e1f1a5970c28faec9e6a1d51L35) with
```py
ASGIApplication = Any
```
in this case? Reading up on [NewType](https://docs.python.org/3/library/typing.html#newtype), you are not loosing anything important anyways in terms of type safetly, when subclassing `Any`.

MHHukiewitz:
And also, can you add the debian runtime to the paths to be checked by mypy and others in the GH action?

hoh:
> And also, can you add the debian runtime to the paths to be checked by mypy and others in the GH action?

Done in https://github.com/aleph-im/aleph-vm/pull/411

PR #420, opened 2023-09-28
Fix: install FastAPI from PyPI in Debian 12 runtime

odesenfans:
Problem: Debian 12 runtime has FastAPI 0.92.0 (installed from apt) while Debian 11 runtime has FastAPI 0.95.1. Some VMs previously running on the Debian 11 runtime do not run on the Debian 12 runtime because of this issue.

Solution: install the latest version of FastAPI from PyPI.

### DIFF
--- a/runtimes/aleph-debian-12-python/create_disk_image.sh
+++ b/runtimes/aleph-debian-12-python/create_disk_image.sh
@@ -33,7 +33,7 @@ pip3 install --break-system-packages 'fastapi~=0.103.1'
 
 echo "Pip installing aleph-sdk-python"
 mkdir -p /opt/aleph/libs
-pip3 install --target /opt/aleph/libs 'aleph-sdk-python==0.7.0'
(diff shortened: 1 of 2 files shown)


runtimes/aleph-debian-12-python/create_disk_image.sh
```diff
@@ -17,21 +17,20 @@ apt-get install -y --no-install-recommends --no-install-suggests \
...
   cgroupfs-mount \
   nftables \
-  \
   iputils-ping curl
 
+pip3 install --break-system-packages 'fastapi~=0.103.1'
+
 echo "Pip installing aleph-sdk-python"
 mkdir -p /opt/aleph/libs
 pip3 install --target /opt/aleph/libs 'aleph-sdk-python==0.7.0'
```
hoh, line 36: Why not just this ?
```suggestion
pip3 install --target /opt/aleph/libs 'aleph-sdk-python==0.7.0' 'fastapi~=0.103.1'
```

hoh (approved):

PR #427, opened 2023-10-05
Fix: Set hostname encoding as string bytes

nesitor:


vm_supervisor/vm/firecracker/instance.py
```diff
@@ -237,7 +237,9 @@ def _create_network_file(self) -> bytes:
     def _create_metadata_file(self) -> bytes:
         """Creates metadata configuration file for cloud-init tool"""
 
-        hostname = base64.b32encode(self.vm_hash).decode().strip("=").lower()
+        hostname = (
+            base64.b32encode(str.encode(self.vm_hash)).decode().strip("=").lower()
```
hoh, line 241: Why ?

nesitor: Because `self.vm_hash` is an `ItemHash` type, so it fails. Only getting `str()` isn't enough because the `b32encode()` method needs a byte string to work.

hoh: Understand. But are you decoding the `vm_hash` correctly here ? In `init1.py`, the logic is:

```python
   540    item_hash_binary: bytes = base64.b16decode(config.vm_hash.encode().upper())                                                                                                                                               
   541    hostname = base64.b32encode(item_hash_binary).decode().strip("=").lower()                                                                                                                                                 
```

It looks like your code is not decoding the hexadecimal first.

nesitor: I think that casting the `vm_hash` as string bytes, it's not needed to convert it first to a hexadecimal format, because if you do `str(vm_hash)`, you get well the hexadecimal format directly.

hoh: Your code is not decoding the hexadecimal and taking the bytes directly from the string. 
See this example:

```python
>>> from aleph_message.models import ItemHash
>>> import base64
>>> vm_hash = ItemHash("3fc0aa9569da840c43e7bd2033c3c580abb46b007527d6d20f2d4e98e867f7af")
>>> base64.b32encode(str.encode(vm_hash)).decode().strip("=").lower()
'gntggmdbme4tknrzmrqtqnbqmm2dgzjxmjsdembtgnrtgyzvhaygcytcgq3gembqg42ten3egzsdemdggjsdizjzhbstqnrxmy3wczq'
>>> item_hash_binary: bytes = base64.b16decode(vm_hash.encode().upper())
>>> base64.b32encode(item_hash_binary).decode().strip("=").lower()                                                                                                                                                 
'h7akvflj3kcayq7hxuqdhq6fqcv3i2yaout5nuqpfvhjr2dh66xq'
```

PR #428, opened 2023-10-05
Fix: Give in total 60 second to the VM to respond the ping

nesitor:
Problem: Sometimes initiate a VM take a bit more time that used to, but finally starts.

Solution: Just to avoid HostNotFound errors, increase the ping attempts and the timeout to 60 seconds in total.

### DIFF
--- a/vm_supervisor/vm/firecracker/instance.py
+++ b/vm_supervisor/vm/firecracker/instance.py
@@ -142,8 +142,8 @@ class AlephFirecrackerInstance(AlephFirecrackerExecutable):
 
         ip = ip.split("/", 1)[0]
 
-        attempts = 10
-        timeout_seconds = 1.0
+        attempts = 30
+        timeout_seconds = 2.0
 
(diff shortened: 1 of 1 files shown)


vm_supervisor/vm/firecracker/instance.py
```diff
@@ -142,8 +142,8 @@ async def wait_for_init(self) -> None:
 
         ip = ip.split("/", 1)[0]
 
-        attempts = 10
-        timeout_seconds = 1.0
+        attempts = 30
+        timeout_seconds = 2.0
```
hoh, line 146:
```suggestion
        timeout_seconds = 0.5
        timeout_total = 60
        attempts = timeout_total // timeout_seconds
        
```

nesitor: This will increase the ping error logs, andI think that 2 seconds is a good timeout for ping on instances.

PR #440, opened 2023-10-19
Feature: Create CLI command to run permanent VMs

nesitor:
Problem: We need to run an instance independently of the supervisor process.

Solution: Create an Instance CLI command tool to manage it.


hoh:
The concept of the VM pool is to have a global object that keeps track of the VMs that are running. Using that object in the wrapper that launches a single VM seems weird. Would it not make more sense to directly launch the VM without the pool object ?

PR #445, opened 2023-10-23
Refactor VMType path

nesitor:
Problem: Class VMType are coupled to orchestration path.

Solution: Decouple VMType class to be outside orchestrator path.


hoh:
Would it not make sense to move this into aleph-message instead ?

nesitor:
> Would it not make sense to move this into aleph-message instead ?

Can be, but for the moment we need to merge this refactor just to advance on next steps. Later we can move it easily to aleph-message repository.

MHHukiewitz (approved):
LGTM

MHHukiewitz:
Merging for now, we will do bigger refactorings after we fixed it all. No sense in refactoring buggy code.

PR #449, opened 2023-10-24
Fix: Dates did not use UTC

hoh:
Problem: The timezone of the system was often used for dates. This can cause inconsistencies and should be avoided in favour of UTC.


olethanh (approved):
LGTM

PR #450, opened 2023-10-24
Cleanup code for standards

hoh:
This branch is based on the feedback of `ruff`, a code linter.

The commits below fix about half of the issues reported by `ruff`.

- Fix: Pool singleton caused issues in tests
- Fix: Don't create a dummy object when could be Optional
- Fix: Error remapping did not extend the original error
- Fix: Function arguments not to standard
- Fix: Dates did not use UTC
- Fix: Path.link_to is deprecated
- Fix: Hardcoded HTTP error codes were used
- Fix: Minor code cleanups
- Fix: subprocess.run() check argument was implicit
- Fix: Fixes for Python < 3.8 are now useless
- Fix: Useless print()
- Fix: Relative imports of parent not recommended
- Fix: Exception classes must end with `Error`


hoh:
I moved the problematic change to a separate PR: https://github.com/aleph-im/aleph-vm/pull/456

PR #452, opened 2023-10-25
Problem: CACHE_ROOT & EXECUTION_ROOT not respected

olethanh:
The value set by the user for CACHE_ROOT and EXECUTION_ROOT were not respected when calculating the default value of MESSAGE_CACHE, CODE_CACHE, EXECUTION_DATABASE, etc...

Solution:
This was caused by the default value being defined at class definition, this patch move the defaults computation into Settings.check() so we have the user passed settings.

Reproduce the problem:
1. Without setting MESSAGE_CACHE
2. set CACHE_ROOT (via environment variable ALEPH_VM_CACHE_ROOT)
3. Print the settings  e.g ALEPH_VM_CACHE_ROOT=/tmp/ aleph-vm --do-not-run --print-settings

What was happening:
MESSAGE_CACHE = /var/cache/aleph/vm/message

What is expected and is happening now:
MESSAGE_CACHE = /tmp/message

### DIFF
--- a/src/aleph/vm/conf.py
+++ b/src/aleph/vm/conf.py
@@ -173,17 +173,26 @@ class Settings(BaseSettings):
     CONNECTOR_URL = Url("http://localhost:4021")
 
     CACHE_ROOT = Path("/var/cache/aleph/vm")
-    MESSAGE_CACHE = CACHE_ROOT / "message"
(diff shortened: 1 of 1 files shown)


hoh (approved):

hoh:
Can you fix the `mypy` error ?

PR #455, opened 2023-11-07
Feature: allow IPv6 DNS

olethanh:
Problem IPv6 DNS were automatically filtered when detected from resolvectl

Solution: Nameservers are now split into ipv4 and ipv6 and can be passed to the VM accordingly At the moment we pass them if the ipv6 parameter is present on the tap interface but we need a more robust detection method.

This is an alternative implementation of #378.
As @odesenfans said in the other thread ideally , we should have the supervisor and runtime communicate to us if they support ipv6 and then only pass the ipv6 nameservers not sure if someone know a way to do that or if it's a new spec we should come up with? Also don't know if it's a blocker

### DIFF
--- a/tests/supervisor/test_resolvectl_dns_servers.py
+++ b/tests/supervisor/test_resolvectl_dns_servers.py
@@ -2,7 +2,7 @@
 import os
 from unittest import mock
 
-from aleph.vm.conf import resolvectl_dns_servers, resolvectl_dns_servers_ipv4
+from aleph.vm.conf import resolvectl_dns_servers
 
 os.environ["ALEPH_VM_ALLOW_VM_NETWORKING"] = "False"
 


MHHukiewitz (approved):

hoh:
This needs to be rebased on top of the current main branch.

src/aleph/vm/conf.py
```diff
@@ -94,7 +83,7 @@ def obtain_dns_ips(dns_resolver: DnsResolver, network_interface: str) -> list[st
         # Use a try-except approach since resolvectl can be present but disabled and raise the following
         # "Failed to get global data: Unit dbus-org.freedesktop.resolve1.service not found."
         try:
-            return list(resolvectl_dns_servers_ipv4(interface=network_interface))
+            return list(resolvectl_dns_servers(interface=network_interface))
```
hoh, line 86: Should we only change this behavior if IPv6 is available in the VMs ?

Referring to the docstring "VMs only support IPv4 networking for now, we must exclude IPv6 DNS from their config."

olethanh: not sure I understand the comment, we filter that in the  instance.py (see below) not at the obtain_dns_ip step

PR #456, opened 2023-11-08
Fix: subprocess.run() check argument was implicit

hoh:
Problem: Not passing the `check` argument to `subprocess.run` implicitly does not check for errors.

Solution: Specify when the command should succeed or not.


nesitor (approved):

hoh:
Replaced by https://github.com/aleph-im/aleph-vm/pull/566/files

PR #457, opened 2023-11-08
Fix: Users could fill all RAM using queues

hoh:
Users could open an unlimited number of queues containing an unlimited number of messages.


olethanh (approved):
tested, good for me

PR #471, opened 2023-11-16
Fix: Instances and instance rootfs were stored indifferently

hoh:
But the two are used very differently.


hoh:
This would cause backward compatibility issues with existing nodes.

PR #479, opened 2023-11-21
Added Systemd launcher for instances

nesitor:
Problem: The instance VMs should be launched independently of the orchestrator.

Solution: Launch VMs using systemd manager.


src/aleph/vm/models.py
```diff
@@ -174,10 +191,17 @@ async def create(self, vm_id: int, tap_interface: Optional[TapInterface] = None)
...
+                # for persistent and instances we will use SystemD manager
+                await vm.configure()
+            except Exception:
+                await vm.teardown()
+                raise
+
         try:
-            await vm.setup()
-            await vm.start()
-            await vm.configure()
```
olethanh, line 180: maybe we could keep this and just check with 
 if not self.persistent: like we do above?

nesitor: For Instances don't make sense, because always are persistent.

olethanh: yeah I mean just as a way to only have one code path but that's just cosmetic

olethanh:
I haven tested but is the record_usage stuff working? Since it's need the pid of the process in to_dict.

src/aleph/vm/controllers/firecracker/executable.py
```diff
@@ -251,8 +260,37 @@ async def wait_for_init(self) -> None:
...
+        """Save VM configuration to be used by the controller service"""
+        with open(f"{settings.EXECUTION_ROOT}/{self.vm_hash}-controller.json", "wb") as controller_config_file:
+            controller_config_file.write(
+                self.controller_configuration.json(by_alias=True, exclude_none=True, indent=4).encode()
+            )
+            controller_config_file.flush()
+            config_file_path = Path(controller_config_file.name)
+            config_file_path.chmod(0o644)
+            return config_file_path
+
```
hoh, line 273: This can be simplified

```suggestion
    def save_controller_configuration(self) -> Path:
        """Save VM configuration to be used by the controller service"""
        config_file_path = Path(f"{settings.EXECUTION_ROOT}/{self.vm_hash}-controller.json")
        with config_file_path.open("w") as controller_config_file:
            controller_config_file.write(
                self.controller_configuration.json(by_alias=True, exclude_none=True, indent=4)
            )
        config_file_path.chmod(0o644)
        return config_file_path
```

Changes made:

1.  Simplified File Handling: The Path object is created at the beginning of the function, and file opening is done with the open method of the Path object. This makes the handling of the file path and file object more straightforward.

2.    Removed Unnecessary Encoding and Flushing: Since the file is opened in text mode ("w"), there's no need to encode the JSON string manually. Also, the flush method is not explicitly required here, as closing the file (which is automatically done by the with block) ensures that the data is written to the file.

3.    Streamlined Permission Setting: The file permissions are set after the file is closed. This separation of concerns makes the code easier to understand.

hoh: Also, why is this a method of the class and not an independent function ?

src/aleph/vm/controllers/firecracker/executable.py
```diff
@@ -251,8 +260,37 @@ async def wait_for_init(self) -> None:
...
+            )
+
+            configuration = Configuration(
+                vm_id=self.vm_id,
+                settings=settings,
+                vm_configuration=vm_configuration,
+            )
+
+            self.controller_configuration = configuration
+            self.save_controller_configuration()
```
hoh, line 293: This pattern of saving the configuration in an attribute and then calling a function that uses that value just after seems weird, especially since that is the only place where that value is read.

Why not make it a separate function and call this ?

```suggestion
            save_controller_configuration(self.vm_hash, configuration)
```

src/aleph/vm/controllers/firecracker/instance.py
```diff
@@ -79,6 +79,7 @@ def __init__(
             enable_console,
             hardware_resources or MachineResources(),
             tap_interface,
+            True,
```
hoh, line 82: The last argument is not trivial.

```suggestion
            persistent=True,
```

src/aleph/vm/pool.py
```diff
@@ -152,11 +159,21 @@ async def stop_vm(self, vm_hash: ItemHash) -> Optional[VmExecution]:
...
             return execution
         else:
             return None
 
+    async def stop_persistent_execution(self, execution):
+        """Stop persistent VMs in the pool."""
+        assert execution.persistent, "Execution isn't persistent"
+        self.systemd_manager.stop_and_disable(execution.controller_service)
+        await execution.stop()
+        execution.persistent = False
```
hoh, line 175: What would happen if one of the two lines above would crash ?

src/aleph/vm/pool.py
```diff
@@ -170,10 +187,14 @@ def forget_vm(self, vm_hash: ItemHash) -> None:
             pass
 
     async def stop(self):
-        """Stop all VMs in the pool."""
-
+        """Stop ephemeral VMs in the pool."""
         # Stop executions in parallel:
-        await asyncio.gather(*(execution.stop() for vm_hash, execution in self.executions.items()))
+        await asyncio.gather(*(execution.stop() for vm_hash, execution in self.get_ephemeral_executions()))
```
hoh, line 192: What would be the procedure to stop all running VMs, including persistent VMs ?

src/aleph/vm/pool.py
```diff
@@ -170,10 +187,14 @@ def forget_vm(self, vm_hash: ItemHash) -> None:
...
+
+    def get_ephemeral_executions(self) -> Iterable[VmExecution]:
+        for _vm_hash, execution in self.executions.items():
+            if not execution.persistent and execution.is_running:
+                yield execution
```
hoh, line 197: Would it make sense to use a generator comprehension here ?

```suggestion
def get_ephemeral_executions(self) -> Iterable[VmExecution]:
    return (execution for _vm_hash, execution in self.executions.items()
            if execution.is_running and not execution.persistent)
```

src/aleph/vm/systemd.py
```diff
@@ -0,0 +1,73 @@
...
+import sys
+
+import dbus
+from dbus import SystemBus
+from dbus.proxies import Interface
+
+logger = logging.getLogger(__name__)
+
+
+class SystemDManager:
```
hoh, line 15: Can you add a docstring explaining the goal of this class ?

src/aleph/vm/systemd.py
```diff
@@ -0,0 +1,73 @@
...
+    bus: SystemBus
+    manager: Interface
```
hoh, line 17: Why is the instance of `Interface` called `manager` and not `interface` ?

src/aleph/vm/systemd.py
```diff
@@ -0,0 +1,73 @@
...
+        self.manager = dbus.Interface(systemd, "org.freedesktop.systemd1.Manager")
+
+    def stop_and_disable(self, service: str) -> None:
+        try:
+            if self.is_service_active(service):
+                self.stop(service)
+            if self.is_service_enabled(service):
+                self.disable(service)
+        except Exception:
+            raise
```
hoh, line 31: Why this error catching code ? It does not do anything :thinking:

src/aleph/vm/systemd.py
```diff
@@ -0,0 +1,73 @@
...
+        logger.debug(f"Restarted {service} service")
+
+    def disable(self, service: str) -> None:
+        self.manager.DisableUnitFiles([service], False)
+        logger.debug(f"Disabled {service} service")
+
+    def is_service_enabled(self, service: str) -> bool:
+        try:
+            return self.manager.GetUnitFileState(service) == "enabled"
+        except:
```
hoh, line 56: What error is this expected to handle ? Using global `except` with no precise error catching is a bad practice.

src/aleph/vm/systemd.py
```diff
@@ -0,0 +1,73 @@
...
+            return False
+
+    def is_service_active(self, service: str) -> bool:
+        try:
+            self.manager.GetUnit(service)
+            return True
+        except:
```
hoh, line 63: Same commend as above

src/aleph/vm/systemd.py
```diff
@@ -0,0 +1,73 @@
...
+            return False
+
+    def enable_and_start(self, service: str) -> None:
+        try:
+            if not self.is_service_enabled(service):
+                self.enable(service)
+            if not self.is_service_active(service):
+                self.start(service)
+        except Exception:
```
hoh, line 72: Same comment about the useless error catching.

src/aleph/vm/controllers/firecracker/executable.py
nesitor: Because it depends on the Instance/Program controller configuration and the `vm_hash`. We can create an independent method, but anyway we will need to pass the desired config. Also, it is only used internally to save the configuration, so it's better to have it inside the class.

olethanh: in general if you want to just dump the content into a file I would say you don't even need to bother using a With and everything you can just do

```
        path = Path(f"{settings.EXECUTION_ROOT}/{self.vm_hash}-controller.json")
        path.open("w").write(self.controller_configuration.json(by_alias=True, exclude_none=True, indent=4))
        path.chmod(0o644)
        return path
```

olethanh:
I also  wonder if it would not have more send to split the save configuration in to a serialization that is inside the controler and the saving to a file  inside the dbus manager.

olethanh:
I'm in the middle of testing it, via my usual lab install (which don't use the .deb), I had to copy the .service file to `/etc/systemd/system/aleph-vm-controller@.service`

Using 
`sudo cp packaging/aleph-vm/etc/systemd/system/aleph-vm-controller@.service /etc/systemd/system/aleph-vm-controller@.service `

It would be nice to have more documentation  on the feature and how to test it like I did in QEMU.md


When I try to launch via the allocation API I get:
```
2023-12-07 14:03:10,325 | ERROR | org.freedesktop.systemd1.NoSuchUnit: Unit aleph-vm-controller@decadecadecadecadecadecadecadecadecadecadecadecadecadecadecadeca.service not loaded.
```

I'm getting the following error in my journald:
```
Dec 07 14:06:13 alephtest systemd[1]: Started Aleph VM decadecadecadecadecadecadecadecadecadecadecadecadecadecadecadeca Controller.
Dec 07 14:06:13 alephtest systemd[320492]: aleph-vm-controller@decadecadecadecadecadecadecadecadecadecadecadecadecadecadecadeca.service: Changing to the requested working directory failed: No such file or directory
Dec 07 14:06:13 alephtest systemd[320492]: aleph-vm-controller@decadecadecadecadecadecadecadecadecadecadecadecadecadecadecadeca.service: Failed at step CHDIR spawning /usr/bin/python3: No such file or directory
Dec 07 14:06:13 alephtest systemd[1]: aleph-vm-controller@decadecadecadecadecadecadecadecadecadecadecadecadecadecadecadeca.service: Main process exited, code=exited, status=200/CHDIR
Dec 07 14:06:13 alephtest systemd[1]: aleph-vm-controller@decadecadecadecadecadecadecadecadecadecadecadecadecadecadecadeca.service: Failed with result 'exit-code'.
```

It seems to retry to launch it in a loop and spam the journal log, even when aleph-vm is stopped. I'm not sure if that's a good idea.


I also get the following warning:
`Dec 07 14:06:14 alephtest systemd[1]: /etc/systemd/system/aleph-vm-controller@.service:9: Standard output type syslog is obsolete, automatically updating to journal. Please update your unit file, and consider removing the setting altogether`
I'm guessing this is caused by the following line
```
StandardOutput=syslog
StandardError=syslog
```

olethanh:
ok I figured the issue, in the `.service` file the path to /opt/aleph-vm , the python binary and the controller file in /var/lib/ are hardcoded, which means it doesn't respect the alpeh settings for directories nor does it works with virtualenvs  or custom python install. We need to find a way around this. For now I have just hardcoded my own paths in the .service file.

nesitor:
> ok I figured the issue, in the `.service` file the path to /opt/aleph-vm , the python binary and the controller file in /var/lib/ are hardcoded, which means it doesn't respect the alpeh settings for directories nor does it works with virtualenvs or custom python install. We need to find a way around this. For now I have just hardcoded my own paths in the .service file.

Yes, the paths are hard-coded because it always installs from the package. Also, if you install the service manually, you need to refresh the services using that command: `sudo systemctl daemon-reload`

olethanh:
> Yes, the paths are hard-coded because it always installs from the package. 
Perhaps we should just remove theses settings then if using them is going to break things.

> Also, if you install the service manually, you need to refresh the services using that command: `sudo systemctl daemon-reload`
Thanks

hoh:
Can you squash / rework the commits ?

pyproject.toml
```diff
@@ -43,7 +43,8 @@ dependencies = [
   "packaging~=23.2",
   "jsonschema==4.19.1",
   "qmp==0.0.1",
-  "systemd-python"
+  "dbus-python==1.3.2",
+  "systemd-python",
```
hoh, line 47: Can we fix the version of `systemd-python` ?

src/aleph/vm/controllers/firecracker/instance.py
```diff
@@ -168,7 +164,11 @@ def _get_hostname(self) -> str:
     def _encode_user_data(self) -> bytes:
         """Creates user data configuration file for cloud-init tool"""
 
-        ssh_authorized_keys = self.resources.message_content.authorized_keys or []
+        ssh_authorized_keys: list[str] | None
+        if settings.USE_DEVELOPER_SSH_KEYS:
```
hoh, line 168: We should ensure that we only allow developer SSH keys if they come from the CLI argument, not from the environment variables.

```suggestion
        if settings.USE_DEVELOPER_SSH_KEYS == ALLOW_DEVELOPER_SSH_KEYS:
```

nesitor: We are already doing it here:
https://github.com/aleph-im/aleph-vm/blob/main/src/aleph/vm/orchestrator/cli.py#L308

src/aleph/vm/pool.py
nesitor: The same that happened before, I just added a method to use systemd instead of the old stop() method.
For the stop method on systemd, it is already controlled to not raise exceptions.

src/aleph/vm/pool.py
nesitor: This method is just called to stop the running VMs when the orchestrator goes down, and this should not stop the persistent VMs.
We don't have a method to stop all VMs, and also I think that we don't need it.

hoh: Should we at least add a CLI or script that stops persistent VMs in case we need it ?

src/aleph/vm/controllers/firecracker/instance.py
nesitor: Just for that, there we set the value, and only here we use it, so if at the moment to set it we do the job, here we don't need to check it again, because if not the `USE_DEVELOPER_SSH_KEYS` variable can be removed and only use `ALLOW_DEVELOPER_SSH_KEYS` variable to make the checks.

hoh: Move to https://github.com/aleph-im/aleph-vm/issues/506

src/aleph/vm/pool.py
hoh: Move to https://github.com/aleph-im/aleph-vm/issues/505

hoh:
I rebased this on `main`. 

I have the following error when deployed on Debian 12:
```
Traceback (most recent call last):
  File "<frozen runpy>", line 189, in _run_module_as_main
  File "<frozen runpy>", line 148, in _get_module_details
  File "<frozen runpy>", line 112, in _get_module_details
  File "/root/aleph-vm-remote/src/aleph/vm/orchestrator/__init__.py", line 3, in <module>
    from . import (
  File "/root/aleph-vm-remote/src/aleph/vm/orchestrator/reactor.py", line 7, in <module>
    from aleph.vm.pool import VmPool
  File "/root/aleph-vm-remote/src/aleph/vm/pool.py", line 12, in <module>
    from aleph.vm.systemd import SystemDManager
  File "/root/aleph-vm-remote/src/aleph/vm/systemd.py", line 8, in <module>
    import dbus
ModuleNotFoundError: No module named 'dbus'
```

src/aleph/vm/controllers/configuration.py
```diff
@@ -0,0 +1,47 @@
...
+    settings: Settings
+    vm_configuration: Union[QemuVMConfiguration, VMConfiguration]
+    hypervisor: HypervisorType = HypervisorType.firecracker
+
+
+def save_controller_configuration(vm_hash: str, configuration: Configuration) -> Path:
+    """Save VM configuration to be used by the controller service"""
+    config_file_path = Path(f"{settings.EXECUTION_ROOT}/{vm_hash}-controller.json")
+    with config_file_path.open("w") as controller_config_file:
+        controller_config_file.write(configuration.json(by_alias=True, exclude_none=True, indent=4))
```
hoh, line 45: When attempting to schedule a program instead of an instance from the scheduler API, I get an error 500.
While this is not expected to be done, it should not crash with an error 500.

```python
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/aiohttp/web_protocol.py", line 433, in _handle_request
    resp = await request_handler(request)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/aiohttp/web_app.py", line 504, in _handle
    resp = await handler(request)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/aiohttp/web_middlewares.py", line 117, in impl
    return await handler(request)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/root/aleph-vm-remote/src/aleph/vm/orchestrator/supervisor.py", line 52, in server_version_middleware
    resp: web.StreamResponse = await handler(request)
                               ^^^^^^^^^^^^^^^^^^^^^^
  File "/root/aleph-vm-remote/src/aleph/vm/orchestrator/views/__init__.py", line 290, in update_allocations
    await start_persistent_vm(instance_item_hash, pubsub, pool)
  File "/root/aleph-vm-remote/src/aleph/vm/orchestrator/run.py", line 267, in start_persistent_vm
    execution = await create_vm_execution(vm_hash=vm_hash, pool=pool, persistent=True)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/root/aleph-vm-remote/src/aleph/vm/orchestrator/run.py", line 60, in create_vm_execution
    execution = await pool.create_a_vm(
                ^^^^^^^^^^^^^^^^^^^^^^^
  File "/root/aleph-vm-remote/src/aleph/vm/pool.py", line 100, in create_a_vm
    await execution.create(vm_id=vm_id, tap_interface=tap_interface)
  File "/root/aleph-vm-remote/src/aleph/vm/models.py", line 218, in create
    await vm.configure()
  File "/root/aleph-vm-remote/src/aleph/vm/controllers/firecracker/program.py", line 327, in configure
    await super().configure()
  File "/root/aleph-vm-remote/src/aleph/vm/controllers/firecracker/executable.py", line 285, in configure
    save_controller_configuration(self.vm_hash, configuration)
  File "/root/aleph-vm-remote/src/aleph/vm/controllers/configuration.py", line 45, in save_controller_configuration
    controller_config_file.write(configuration.json(by_alias=True, exclude_none=True, indent=4))
                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "pydantic/main.py", line 504, in pydantic.main.BaseModel.json
  File "/usr/lib/python3.11/json/__init__.py", line 238, in dumps
    **kw).encode(obj)
          ^^^^^^^^^^^
  File "/usr/lib/python3.11/json/encoder.py", line 202, in encode
    chunks = list(chunks)
             ^^^^^^^^^^^^
  File "/usr/lib/python3.11/json/encoder.py", line 432, in _iterencode
    yield from _iterencode_dict(o, _current_indent_level)
  File "/usr/lib/python3.11/json/encoder.py", line 406, in _iterencode_dict
    yield from chunks
  File "/usr/lib/python3.11/json/encoder.py", line 406, in _iterencode_dict
    yield from chunks
  File "/usr/lib/python3.11/json/encoder.py", line 439, in _iterencode
    o = _default(o)
        ^^^^^^^^^^^
  File "pydantic/json.py", line 90, in pydantic.json.pydantic_encoder
TypeError: Object of type 'object' is not JSON serializable
```

hoh:
The command `btrfstune` is apparently not installed by the Debian package.

```python
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/aiohttp/web_protocol.py", line 433, in _handle_request
    resp = await request_handler(request)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/aiohttp/web_app.py", line 504, in _handle
    resp = await handler(request)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/aiohttp/web_middlewares.py", line 117, in impl
    return await handler(request)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/root/aleph-vm-remote/src/aleph/vm/orchestrator/supervisor.py", line 52, in server_version_middleware
    resp: web.StreamResponse = await handler(request)
                               ^^^^^^^^^^^^^^^^^^^^^^
  File "/root/aleph-vm-remote/src/aleph/vm/orchestrator/views/__init__.py", line 290, in update_allocations
    await start_persistent_vm(instance_item_hash, pubsub, pool)
  File "/root/aleph-vm-remote/src/aleph/vm/orchestrator/run.py", line 267, in start_persistent_vm
    execution = await create_vm_execution(vm_hash=vm_hash, pool=pool, persistent=True)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/root/aleph-vm-remote/src/aleph/vm/orchestrator/run.py", line 60, in create_vm_execution
    execution = await pool.create_a_vm(
                ^^^^^^^^^^^^^^^^^^^^^^^
  File "/root/aleph-vm-remote/src/aleph/vm/pool.py", line 91, in create_a_vm
    await execution.prepare()
  File "/root/aleph-vm-remote/src/aleph/vm/models.py", line 164, in prepare
    await resources.download_all()
  File "/root/aleph-vm-remote/src/aleph/vm/controllers/firecracker/instance.py", line 49, in download_all
    await asyncio.gather(
  File "/root/aleph-vm-remote/src/aleph/vm/controllers/firecracker/instance.py", line 45, in download_runtime
    self.rootfs_path = await create_devmapper(self.message_content.rootfs, self.namespace)
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/root/aleph-vm-remote/src/aleph/vm/storage.py", line 340, in create_devmapper
    await resize_and_tune_file_system(path_mapped_volume_name, mount_path)
  File "/root/aleph-vm-remote/src/aleph/vm/storage.py", line 286, in resize_and_tune_file_system
    await run_in_subprocess(["btrfstune", "-m", str(device_path)])
  File "/root/aleph-vm-remote/src/aleph/vm/utils.py", line 82, in run_in_subprocess
    process = await asyncio.create_subprocess_exec(
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/asyncio/subprocess.py", line 218, in create_subprocess_exec
    transport, protocol = await loop.subprocess_exec(
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/asyncio/base_events.py", line 1694, in subprocess_exec
    transport = await self._make_subprocess_transport(
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/asyncio/unix_events.py", line 207, in _make_subprocess_transport
    transp = _UnixSubprocessTransport(self, protocol, args, shell,
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/asyncio/base_subprocess.py", line 36, in __init__
    self._start(args=args, shell=shell, stdin=stdin, stdout=stdout,
  File "/usr/lib/python3.11/asyncio/unix_events.py", line 810, in _start
    self._proc = subprocess.Popen(
                 ^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/subprocess.py", line 1024, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/usr/lib/python3.11/subprocess.py", line 1901, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: 'btrfstune'
```

Fixed in https://github.com/aleph-im/aleph-vm/pull/508

src/aleph/vm/controllers/configuration.py
nesitor: I already tested it on Debian 11 and 12, but I cannot reproduce that issue.

hoh: OK, I will reset my test setup and check again.

src/aleph/vm/controllers/qemu/instance.py
```diff
@@ -366,26 +325,11 @@ async def print_logs():
...
-            if not resp == {}:
-                logger.warning("unexpected answer from VM", resp)
-            client.close()
-        self.qmp_socket_path = None
-
     def get_log_queue(self) -> asyncio.Queue:
         queue, canceller = make_logs_queue(self._journal_stdout_name, self._journal_stderr_name)
         self._queue_cancellers[queue] = canceller
         # Limit the number of queues per VM
+        # TODO : fix
```
hoh, line 332: I do not understand this comment

hoh (approved):

hoh (approved):

PR #481, opened 2023-11-22
Make JAILER_BASE_DIR dependend on execution root

olethanh:
it was hardcoded before on /var/lib/aleph so it wasn't respecting the proper user settings and was failling if the user didn't have the correct permission for this folder and was producing warning

### DIFF
--- a/src/aleph/vm/hypervisors/firecracker/microvm.py
+++ b/src/aleph/vm/hypervisors/firecracker/microvm.py
@@ -238,7 +238,7 @@ class MicroVM:
             "--gid",
             gid,
             "--chroot-base-dir",
-            self.jailer_base_directory,
+            str(self.jailer_base_directory),
             "--",
             "--config-file",
             "/tmp/" + str(self.config_file_path.name),
--- a/src/aleph/vm/hypervisors/firecracker/microvm.py
+++ b/src/aleph/vm/hypervisors/firecracker/microvm.py
@@ -94,10 +94,6 @@ class MicroVM:
         return f"vm-{self.vm_id}"
 
     @property
-    def jailer_base_directory(self) -> Path:
-        return self.execution_root_folder / "jailer"
-
-    @property
     def namespace_path(self) -> str:
         firecracker_bin_name = os.path.basename(self.firecracker_bin_path)
         return str(self.jailer_base_directory / firecracker_bin_name / str(self.vm_id))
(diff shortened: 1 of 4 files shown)


src/aleph/vm/hypervisors/firecracker/microvm.py
```diff
@@ -95,9 +94,13 @@ def __str__(self):
         return f"vm-{self.vm_id}"
 
     @property
-    def namespace_path(self):
+    def jailer_base_directory(self) -> Path:
+        return self.execution_root_folder / "jailer"
```
hoh, line 98: Why use this approach instead of building this path in the settings ?

olethanh:
I had the impression this path variable wasn't seems to be used
consistantly, I.e. we using depending wether we use jailer or not. So doing
it this way I feel this is more flexible for latter refactor.



Le mer. 22 nov. 2023, 11:04, Hugo Herter ***@***.***> a
écrit :

olethanh:
I'm pondering if we should just make a directory per VM and just pass it to
the MicroVm under the principle that that class won't make any
modifications on disk outside of it.  What do you think?

Le mer. 22 nov. 2023, 17:57, Olivier Le Thanh Duong ***@***.***> a
écrit :

olethanh:
After discussion with Andres did the modification with the settings approach

nesitor (approved):

src/aleph/vm/hypervisors/firecracker/microvm.py
```diff
@@ -95,9 +94,9 @@ def __str__(self):
         return f"vm-{self.vm_id}"
 
     @property
-    def namespace_path(self):
+    def namespace_path(self) -> str:
         firecracker_bin_name = os.path.basename(self.firecracker_bin_path)
-        return f"{JAILER_BASE_DIRECTORY}/{firecracker_bin_name}/{self.vm_id}"
+        return str(self.jailer_base_directory / firecracker_bin_name / str(self.vm_id))
```
hoh, line 99: Why `str()` instead of `.name` ?

```suggestion
        return (self.jailer_base_directory / firecracker_bin_name / str(self.vm_id)).name
```

src/aleph/vm/hypervisors/firecracker/microvm.py
```diff
@@ -237,7 +238,7 @@ async def start_jailed_firecracker(self, config_path: Path) -> asyncio.subproces
             "--gid",
             gid,
             "--chroot-base-dir",
-            JAILER_BASE_DIRECTORY,
+            str(self.jailer_base_directory),
```
hoh, line 241: Why str() instead of .name ?

olethanh: It's using str() for `firecracker_bin_path`  and `self.jailer_bin_path` just above so it's consistent.

And frankly, I always used str() for Path because and I didn't really know `.name` existed before.
If you prefer to standardize one one or the other it's okay for me I will adapt.
But with str() I know it just work to convert a path to as str, for `.name` I couldn't tell without checking if it's just going to return the filename of the leaf or the whole path, a string or something else.

PR #458, opened 2023-11-08
Improve VM control by owner

hoh:
- WIP: Fix control enpoints
- WIP: Use pydantic classes in control endpoints
- WIP: Use pydantic classes in control endpoints

PR #485, opened 2023-11-22
Fix minor issues in #458

MHHukiewitz:
See #458
- Return HTTP 403 instead of 401 where appropriate
- Fix typos, clarify some variable names


BjrInt (approved):
Thanks for looking after our typos :sweat_smile: 

LGTM

PR #489, opened 2023-11-27
Allow accessing then logs endpoint using the allocation key auth method

olethanh:
Not sure this is something we want in general and if it fit our architecture and security design, but I found this modification useful when doing development, it allow accessing the `/control/machines/X/logs` endpoint by passing the same  `X-Auth-Signature` http header that we use for `/control/allocation/` endpoint.

What do you think?


MHHukiewitz (approved):
LGTM :+1:

hoh:
Good idea. Would it be worth refactoring the code to make it more readable ? (many levels of nesting)

PR #490, opened 2023-11-29
QEmu support continuation

olethanh:
continuation of #487 

Allow launching VM Instance via QEMU instead of firecracker.

This works by adding a new Controller for Qemu alongside  AlephFirecrackerProgram and AlephFirecrackerInstance and launch it if the message == Instance + hypervisor == qemu.

I'm opening this so we can get the review and discussion started but from discussion I understand it won't be for the next release since we are focusing on bugfixes for now. I can clean it up the git history afterward if needed. Please play with it plenty to find any problem I might have missed.

There is a corresponding PR in aleph-message: https://github.com/aleph-im/aleph-message/pull/78

## how to test
See this pretty complete readme on how to test it https://github.com/aleph-im/aleph-vm/blob/qemu_support/src/aleph/vm/controllers/qemu/QEMU.md

Necessary change in aleph-message were released in 0.4.1

## Modification to the code

I had to make a few change outside the Qemu controller itself to provide compatibility between all controller:

- New abstract class: AlephControllerInterface which define the shared interface between Firecracker and Qemu controllers for sharing and typing.
- Add field `support_snapshot` on controller so the controler can declare support to the SnapShotManager without the different guessing from the method we had till now.
- a Mixin to manage the cloud init config, I intended to have it used between all the controllers that need it  but at the moment I had to tweak the cloud init configuration so it's not done yet
- `get_log_queue` and `unregister_queue` so the operator can register to the Log queues without knowing the internal logic of the VM (which is different since Qemu don't use MicroVM

Refer to QEMU.md for a list of supported feature at the moment.
IMHO the main thing missing is automated testing.
moment I had to tweak the cloud init configuration so it's not done yet
    get_log_queue and unregister_queue so the operator can register to the Log queues without knowing the internal logic of the VM (which is different since Qemu don't use MicroVM

Refer to QEMU.md for a list of supported feature at the moment.
IMHO the main thing missing is automated testing.


src/aleph/vm/controllers/interface.py
```diff
@@ -0,0 +1,91 @@
...
+    "local resource for the machine"
+    enable_console: bool
+    enable_networking: bool
+    "enable networking for this VM"
+    hardware_resources: MachineResources
+    support_snapshot: bool
+    "Does this controller support snapshotting"
+    guest_api_process: Optional[Process] = None
+    tap_interface: Optional[TapInterface] = None
+    "Network interface used for this VM"
```
MHHukiewitz, line 31: Shouldn't docstrings be triple quoted?

src/aleph/vm/controllers/interface.py
```diff
@@ -0,0 +1,91 @@
...
+from aleph_message.models import ItemHash
+from aleph_message.models.execution.environment import MachineResources
+
+from aleph.vm.controllers.firecracker.snapshots import CompressedDiskVolumeSnapshot
+from aleph.vm.network.interfaces import TapInterface
+
+logger = logging.getLogger(__name__)
+
+
+class AlephControllerInterface(ABC):
```
MHHukiewitz, line 16: Maybe rather `AlephVmControllerInterface`?

src/aleph/vm/controllers/interface.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def get_vm_ipv6(self) -> Optional[str]:
+        if self.tap_interface:
+            return self.tap_interface.guest_ipv6.with_prefixlen
+        return None
+
+    def get_vm_ipv6_gateway(self) -> Optional[str]:
+        if self.tap_interface:
+            return str(self.tap_interface.host_ipv6.ip)
+        return None
```
MHHukiewitz, line 51: Remove the `_vm` from the function name?

src/aleph/vm/controllers/interface.py
```diff
@@ -0,0 +1,91 @@
...
+        """Configuration done before the VM process is started"""
+        raise NotImplementedError()
+
+    async def start(self):
+        """Start the VM process"""
+        raise NotImplementedError()
+
+    async def wait_for_init(self) -> None:
+        """Wait for the init process of the virtual machine to be ready.
+        May be empty."""
```
MHHukiewitz, line 67: > May be empty.

What may be empty?

src/aleph/vm/controllers/interface.py
```diff
@@ -0,0 +1,91 @@
...
+        raise NotImplementedError()
+
+    async def stop_guest_api(self):
+        raise NotImplementedError()
+
+    async def teardown(self) -> Coroutine:
+        raise NotImplementedError()
+
+    async def create_snapshot(self) -> CompressedDiskVolumeSnapshot:
+        "Must be implement if self.support_snapshot is True"
```
MHHukiewitz, line 84: Triple quotes

src/aleph/vm/controllers/qemu/cloudinit.py
```diff
@@ -0,0 +1,138 @@
...
+
+import yaml
+from aleph_message.models import ItemHash
+
+from aleph.vm.conf import settings
+from aleph.vm.controllers.interface import AlephControllerInterface
+from aleph.vm.hypervisors.firecracker.config import Drive
+from aleph.vm.utils import is_command_available, run_in_subprocess
+
+# https://cloudinit.readthedocs.io/en/latest/reference/datasources/nocloud.html
```
MHHukiewitz, line 15: How is this related to the code? Maybe you can add a bit of explanation (for n00bs like me)

src/aleph/vm/controllers/qemu/cloudinit.py
```diff
@@ -0,0 +1,138 @@
...
+                user_data_config_file.name,
+                metadata_config_file.name,
+            ]
+        )
+
+
+class CloudInitMixin(AlephControllerInterface):
+    async def _create_cloud_init_drive(self) -> Drive:
+        """Creates the cloud-init volume to configure and setup the VM"""
+        # assert self.enable_networking and self.tap_interface, f"Network not enabled for VM {self.vm_id}"
```
MHHukiewitz, line 106: Why is this commented out? If not necessary, you can remove it

src/aleph/vm/controllers/qemu/instance.py
```diff
@@ -0,0 +1,348 @@
...
+
+    async def make_writable_volume(self, parent_image_path, volume: PersistentVolume | RootfsVolume):
+        "Create a new qcow2 image file based on the passed one, that we give to the VM to write onto"
+        qemu_img_path = shutil.which("qemu-img")
+        volume_name = volume.name if isinstance(volume, PersistentVolume) else "rootfs"
+
+        # detect the image format
+        out_json = await run_in_subprocess([qemu_img_path, "info", str(parent_image_path), "--output=json"])
+        out = json.loads(out_json)
+        parent_format = out.get("format", "")
```
MHHukiewitz, line 44: What happens below if `format` is empty?

src/aleph/vm/controllers/qemu/instance.py
```diff
@@ -0,0 +1,348 @@
...
+                "create",
+                "-f",  # Format
+                "qcow2",
+                "-F",
+                parent_format,
+                "-b",
+                str(parent_image_path),
+                str(dest_path),
+            ]
+        )
```
MHHukiewitz, line 65: It seems we do not take the `PersistentVolume.size_mib` into account.

src/aleph/vm/controllers/qemu/instance.py
```diff
@@ -0,0 +1,348 @@
...
+                }
+            except psutil.NoSuchProcess:
+                logger.warning("Cannot read process metrics (process %s not found)", self.qemu_process)
+                pid_info = None
+        else:
+            pid_info = None
+
+        return {
+            "process": pid_info,
+            **self.__dict__,
```
MHHukiewitz, line 138: Having all the `__dict__` values unpack into this can be dangerous. Defining `process` anywhere as a field can completely override your collected info. Maybe consider not unpacking it and instead make another field `properties` or similar

src/aleph/vm/controllers/qemu/instance.py
```diff
@@ -0,0 +1,348 @@
...
+
+        for attempt in range(attempts):
+            try:
+                await ping(ip, packets=1, timeout=timeout_seconds)
+                return
+            except HostNotFoundError:
+                if attempt < (attempts - 1):
+                    continue
+                else:
+                    raise
```
MHHukiewitz, line 255: Raise what?

src/aleph/vm/hypervisors/firecracker/microvm.py
```diff
@@ -50,6 +50,10 @@ def system(command):
 
 
 async def setfacl():
+    """Give current user permission to access /dev/kvm via acl"""
+    if os.access("/dev/kvm", os.R_OK | os.W_OK):
+        return
+
```
MHHukiewitz, line 56: Why? What does this in regards to QEMU?

src/aleph/vm/models.py
```diff
@@ -182,7 +210,8 @@ async def create(self, vm_id: int, tap_interface: Optional[TapInterface] = None)
             self.times.started_at = datetime.now(tz=timezone.utc)
             self.ready_event.set()
             return vm
-        except Exception:
+        except Exception as exception:
+            print(exception)
             await vm.teardown()
             raise
```
MHHukiewitz, line 216:
```suggestion
            raise exception
```

src/aleph/vm/orchestrator/cli.py
```diff
@@ -253,6 +253,7 @@ async def run_instances(instances: list[ItemHash]) -> None:
 
     await asyncio.gather(*[start_instance(item_hash=instance_id) for instance_id in instances])
     await asyncio.Event().wait()  # wait forever
+    # TODO : should we really wait forever?
```
MHHukiewitz, line 256: poetic.

MHHukiewitz (commented):
LGTM so far, left a few comments but nothing critical

src/aleph/vm/controllers/interface.py
olethanh: I thought it was triple quote only for functions? Black didn't really complain but I can change if if we want?

src/aleph/vm/controllers/interface.py
olethanh: The function code block I think. This is just one of the method I transferred "as is" from the existing code in FirecrackerExecutable

src/aleph/vm/controllers/qemu/cloudinit.py
olethanh: Keep the link there for easy reference, but yeah we should have a docstring for the whole module, will add it

src/aleph/vm/controllers/qemu/instance.py
olethanh: A bare "raise" in an "Except:" block will re-raise the currently handled exception.

src/aleph/vm/controllers/qemu/instance.py
olethanh: The next qemu-img command would fail. I will make a bit more safeguard in the python, it will provide more helpful debug information.

src/aleph/vm/controllers/qemu/instance.py
olethanh: One thing that bother me is that we endup in the `/about/executions` endpoint` with stuff like in the
```tap_interface	"<aleph.vm.network.interfaces.TapInterface object at 0x7fb95d0cffa0>"``` 

But it's totally outside of the scope of this pr

Comment (MHHukiewitz) [reply src/aleph/vm/controllers/qemu/instance.py:144]: https://github.com/aleph-im/aleph-vm/issues/494

File: src/aleph/vm/models.py [hunk 15]
```diff
@@ -64,7 +67,7 @@ class VmExecution:
     original: ExecutableContent
     message: ExecutableContent
     resources: Optional[AlephFirecrackerResources] = None
-    vm: Optional[AlephFirecrackerExecutable] = None
+    vm: Optional[AlephFirecrackerExecutable | AlephQemuInstance] = None
```
MHHukiewitz, line 70: The `|` syntax is not supported for Python versions < 3.10, but we still support 3.9: https://github.com/aleph-im/aleph-vm/blob/main/pyproject.toml#L96

MHHukiewitz (changes requested):
LGTM, just make sure CI/CD passes and we keep Python 3.9 compatibility

src/aleph/vm/models.py
olethanh: good catch, fixed it

nesitor (approved):
LGTM, maybe at the merge moment with the Systemd feature, we will need to make slight changes, but for now it is so good.

MHHukiewitz (approved):
LGTM

PR #491, opened 2023-11-29
Fix: Concurrent creation resulted in no `execution.vm`

hoh:
This caused an error while the processes calling this do check for the execution to become ready.


MHHukiewitz (approved):
LGTM

PR #498, opened 2023-12-08
Fix: Diagnostic could be improved

hoh:
Solution:
- Add checks on the host connectivity, to help with nodes without IPv6 connectivity
- Add links to troubleshooting docs
- Move IPv6 Egress in details section, it was confusing to have it on the main page.
image_2023-12-08_11-27-37
image_2023-12-08_11-27-37-2
image_2023-12-08_11-27-37-3


hoh:
Latest update: 
image

BjrInt (commented):
My feeling with these changes is that we're progressing in the right direction for CRN operators, offering more ways to visualize potential issues.

However the separation between presentation and logic in this specific view is becoming blurrier. We should make sure this does not worsen with new features.

nesitor:
> My feeling with these changes is that we're progressing in the right direction for CRN operators, offering more ways to visualize potential issues.
> 
> However the separation between presentation and logic in this specific view is becoming blurrier. We should make sure this does not worsen with new features.

I completely agree with @BjrInt, maybe for the next changes on these page, we should improve that view improving and refactoring the code to avoid feature issues.

src/aleph/vm/orchestrator/views/__init__.py
```diff
@@ -169,6 +177,25 @@ async def status_check_fastapi(request: web.Request):
...
+            "egress": await check_host_egress_ipv4(),
+            "dns": await check_dns_ipv4(),
+            "domain": await check_domain_resolution_ipv4(),
+        },
+        "ipv6": {
+            "egress": await check_host_egress_ipv6(),
+            "dns": await check_dns_ipv6(),
+            "domain": await check_domain_resolution_ipv6(),
+        },
+    }
```
nesitor, line 194: Maybe we need to improve the performance of that method just calling the method `resolve_dns` just twice instead of calling it for every field.

src/aleph/vm/orchestrator/views/host_status.py
```diff
@@ -0,0 +1,91 @@
...
+        "ipv4": ipv4,
+        "ipv6": ipv6,
+    }
+
+
+async def check_dns_ipv4() -> bool:
+    """Check if DNS resolution is working via IPv4."""
+    resolution = await resolve_dns(settings.CONNECTIVITY_DNS_HOSTNAME)
+    ipv4 = resolution["ipv4"]
+    return bool(ipv4)
```
nesitor, line 52: Why return just a `bool` value instead to return the value itself and control it on the upper method? Maybe we can re-use that code to show some node data returned by these methods.

src/aleph/vm/orchestrator/views/__init__.py
hoh: It is only called 2 times. It is indeed not ideal in terms of performance, but I didn't find a way to make the code simpler without this duplication. Any idea ?

nesitor: It can be to call the method `resolve_dns` on `status_check_host` and deal with the result on the other methods.

MHHukiewitz (approved):
LGTM

src/aleph/vm/conf.py
```diff
@@ -127,6 +127,11 @@ class Settings(BaseSettings):
     WATCH_FOR_UPDATES = True
 
     API_SERVER = "https://official.aleph.cloud"
+    # Connect to the Quad9 VPN provider using their IPv4 and IPv6 addresses.
+    CONNECTIVITY_IPV4_URL = "https://9.9.9.9/"
+    CONNECTIVITY_IPV6_URL = "https://[2620:fe::fe]/"
+    CONNECTIVITY_DNS_HOSTNAME = "example.org"
```
MHHukiewitz, line 133: https://www.iana.org/help/example-domains
Maybe adding fallback addresses to test would be useful?
> We provide a web service on the example domain hosts to provide basic information on the purpose of the domain. These web services are provided as best effort, but are not designed to support production applications. While incidental traffic for incorrectly configured applications is expected, please do not design applications that require the example domains to have operating HTTP service.

MHHukiewitz (approved):
LGTM

PR #500, opened 2023-12-11
Fix: Kubo IPFS used default config

hoh:
This disables features that are not expected on
aleph.im Compute Resource Nodes and limits the number of CIDs published on the P2P network.


MHHukiewitz:
Something seems to be wrong with the Digital Ocean checks, no matter how much one retries... Something about the service file and the VM requiring restart?

hoh:
> Something seems to be wrong with the Digital Ocean checks, no matter how much one retries... Something about the service file and the VM requiring restart?

That is an issue with Kubo 0.24.0 trying to read some `allow_list` in a directory that is not allowed to be accessed on Debian 11, linked to the security hardening of the service file.

See https://github.com/aleph-im/aleph-vm/blob/main/packaging/aleph-vm/etc/systemd/system/ipfs.service#L28-L52

hoh:
I moved the upgrade of Kubo to https://github.com/aleph-im/aleph-vm/pull/504

PR #501, opened 2023-12-11
Fix: CI frequently failed due to new versions

hoh:
Solution: Fix the version of all dependencies used in the project.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,7 +7,7 @@ name = "aleph-vm"
 dynamic = ["version"]
 description = "Aleph.im VM execution engine"
 readme = "README.md"
-requires-python = ">=3.9"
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -7,7 +7,7 @@ name = "aleph-vm"
 dynamic = ["version"]
 description = "Aleph.im VM execution engine"
 readme = "README.md"
-requires-python = ">=3.9"
+requires-python = "==3.9"
```
nesitor, line 10: I think that this cannot be added, because new Debian versions comes with Python 3.11 by default.

hoh: Indeed, I went a bit too hasty.

nesitor (approved):

PR #511, opened 2023-12-14
Load already running persistent VMs

nesitor:
Problem: When the node operator restarts the orchestrator with already running persistent VMs, at the start, it doesn't get already running persistent VMs.

Solution: Use the SQLite database to save all the executions and re-create it from the start.

### DIFF
--- a/src/aleph/vm/models.py
+++ b/src/aleph/vm/models.py
@@ -148,7 +148,7 @@ class VmExecution:
     def to_json(self, indent: Optional[int] = None) -> str:
         return dumps_for_json(self.to_dict(), indent=indent)
 
-    async def prepare(self, download: bool = True):
+    async def prepare(self):
         """Download VM required files"""
         async with self.preparation_pending_lock:
             if self.resources:
@@ -168,8 +168,7 @@ class VmExecution:
             if not resources:
                 msg = "Unknown executable message type"
                 raise ValueError(msg, repr(self.message))
-            if download:
-                await resources.download_all()
+            await resources.download_all()
             self.times.prepared_at = datetime.now(tz=timezone.utc)
             self.resources = resources
 
--- a/src/aleph/vm/models.py
+++ b/src/aleph/vm/models.py
@@ -348,8 +348,8 @@ class VmExecution:
                     vcpus=self.vm.hardware_resources.vcpus,
                     memory=self.vm.hardware_resources.memory,
                     network_tap=self.vm.tap_interface.device_name if self.vm.tap_interface else "",
-                    message=self.message,
-                    original_message=self.original,
+                    message=self.message.json(),
+                    original_message=self.original.json(),
                     persistent=self.persistent,
                 )
             )
--- a/src/aleph/vm/models.py
+++ b/src/aleph/vm/models.py
@@ -10,7 +10,6 @@ from aleph_message.models import (
     ExecutableContent,
     InstanceContent,
     ItemHash,
-    MessageType,
     ProgramContent,
 )
 from aleph_message.models.execution.environment import HypervisorType
@@ -28,6 +27,7 @@ from aleph.vm.controllers.qemu.instance import AlephQemuInstance, AlephQemuResou
 from aleph.vm.network.interfaces import TapInterface
 from aleph.vm.orchestrator.metrics import (
     ExecutionRecord,
+    delete_record,
     save_execution_data,
     save_record,
 )
@@ -37,6 +37,7 @@ from aleph.vm.utils import create_task_log_exceptions, dumps_for_json
 
 if TYPE_CHECKING:
     from aleph.vm.controllers.firecracker.snapshot_manager import SnapshotManager
+    from aleph.vm.systemd import SystemDManager
 
 logger = logging.getLogger(__name__)
 
@@ -83,7 +84,11 @@ class VmExecution:
 
     @property
     def is_running(self):
-        return self.times.starting_at and not self.times.stopping_at
+        return (
+            self.times.starting_at and not self.times.stopping_at
+            if not self.persistent
+            else self.systemd_manager.is_service_active(self.controller_service)
+        )
 
     @property
     def is_program(self):
@@ -116,6 +121,7 @@ class VmExecution:
         message: ExecutableContent,
         original: ExecutableContent,
         snapshot_manager: "SnapshotManager",
+        systemd_manager: "SystemDManager",
         persistent: bool,
     ):
         self.uuid = uuid.uuid1()  # uuid1() includes the hardware address and timestamp
@@ -130,6 +136,7 @@ class VmExecution:
         self.preparation_pending_lock = asyncio.Lock()
         self.stop_pending_lock = asyncio.Lock()
         self.snapshot_manager = snapshot_manager
+        self.systemd_manager = systemd_manager
         self.persistent = persistent
 
     def to_dict(self) -> dict:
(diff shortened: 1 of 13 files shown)


src/aleph/vm/models.py
```diff
@@ -161,15 +168,17 @@ async def prepare(self):
             if not resources:
                 msg = "Unknown executable message type"
                 raise ValueError(msg, repr(self.message))
-            await resources.download_all()
+            if download:
```
hoh, line 171: Why add this `download` flag ? Is `download_all()` not checking for existing resources ? What would happen if a node operator clears `/var/lib/cache` before restarting the orchestrator ?

src/aleph/vm/orchestrator/supervisor.py
```diff
@@ -107,6 +107,10 @@ async def stop_all_vms(app: web.Application):
 def run():
     """Run the VM Supervisor."""
     settings.check()
+
+    engine = setup_engine()
+    create_tables(engine)
```
hoh, line 112: Are there migrations to run ?

src/aleph/vm/orchestrator/metrics.py
```diff
@@ -35,10 +43,11 @@ def create_tables(engine: Engine):
 
 
 class ExecutionRecord(Base):
-    __tablename__ = "records"
+    __tablename__ = "executions"
```
hoh, line 46: Why rename the table ? What would the old table become after the upgrade ?
Do we need both a table for the history of executions and for the currently running VMs ?

src/aleph/vm/pool.py
```diff
@@ -58,6 +62,10 @@ def __init__(self):
         logger.debug("Initializing SnapshotManager ...")
         self.snapshot_manager.run_snapshots()
 
+        logger.debug("Loading existing executions ...")
+        loop = asyncio.get_event_loop()
+        loop.run_until_complete(self._load_persistent_executions())
```
hoh, line 67: Why not use the simpler `asyncio.run()` ?

```suggestion
        asyncio.run(self._load_persistent_executions())
```

src/aleph/vm/orchestrator/metrics.py
```diff
@@ -81,6 +94,16 @@ async def save_record(record: ExecutionRecord):
         session.close()
 
 
+async def delete_record(execution_uuid: str):
+    """Delete the resource usage in database"""
+    session = Session()  # undefined name 'Session'
```
hoh, line 99: Why not use an [AsyncSession](https://docs.sqlalchemy.org/en/20/orm/extensions/asyncio.html#sqlalchemy.ext.asyncio.AsyncSession) ?

src/aleph/vm/pool.py
```diff
@@ -188,6 +199,46 @@ def forget_vm(self, vm_hash: ItemHash) -> None:
         except KeyError:
             pass
 
+    async def _load_persistent_executions(self):
+        """Load persistent executions from the database."""
+        saved_executions = await get_execution_records()
+        for saved_execution in saved_executions:
```
hoh, line 205: In this use case, would it make sense to use an [AsyncSession](https://docs.sqlalchemy.org/en/20/orm/extensions/asyncio.html#sqlalchemy.ext.asyncio.AsyncSession) instead of a synchronous one ?

The original implementation of `get_execution_records()` was synchronous because all data was always returned at once to the HTTP client, but that is not the case anymore with this.

src/aleph/vm/models.py
nesitor: Yes, it checks that already exists. Was the runtime that I was worried about, but it also checks if already exists. I will remove that argument.

src/aleph/vm/orchestrator/supervisor.py
nesitor: I don't think so, also the `EXECUTIONS_DATABASE` parameter was not well set, and I already fixed it to use the correct path, so it will use a new dabase.

src/aleph/vm/orchestrator/metrics.py
nesitor: I have changed de table name because it was only used to know the current executions. Wasn't used as a history either before, because was only used on VM stopping. Anyways, the VM record data was stored on files on VM stop also, and we don't expose either that information.
Now that database is just the table to use to store the current running VMs.

src/aleph/vm/orchestrator/metrics.py
nesitor: I have followed the same pattern than the rest of the methods. Maybe can you add me here an example of how to use it?

src/aleph/vm/pool.py
nesitor: Why to complicate it? Which are the advantages of doing it with AsyncSession?

src/aleph/vm/orchestrator/metrics.py
MHHukiewitz: Wouldn't this require an alembic migration then? Or is this table just too ephemeral for us to care?

src/aleph/vm/orchestrator/metrics.py
MHHukiewitz: I think it is like a regular Session but with `await` everywhere?

src/aleph/vm/orchestrator/supervisor.py
MHHukiewitz: But the renamed database above might be a problem

src/aleph/vm/pool.py
MHHukiewitz: > "When using the AsyncSession object from SQLAlchemy’s asyncio extension, this object is only a thin proxy on top of a Session, and the same rules apply; it is an unsynchronized, mutable, stateful object, so it is not safe to use a single instance of AsyncSession in multiple asyncio tasks at once."

One `Session` per thread, one `AsyncSession` per task is what is intended. An AsyncSession in an `async` function [is exactly what we want](https://docs.sqlalchemy.org/en/20/orm/session_basics.html#is-the-session-thread-safe-is-asyncsession-safe-to-share-in-concurrent-tasks).

MHHukiewitz: Also, consider a Session to be a transaction for the DB. If you have mutliple Session objects across multiple async tasks, this will lead to unnecessary blocking running python code, while the database is executing a transaction. Using AsyncSession, this will be handled gracefully.

MHHukiewitz (commented):
Agreeing with Hugo on most parts, otherwise LGTM

src/aleph/vm/orchestrator/metrics.py
nesitor: I didn't see where we use the old execution data, and previously, we were saving it just when the VM stops, so now, after this PR, we maintain the current execution's registry fine. Also, I can maintain the old database to see the already stopped executions, but I don't see the advantage to do it.

src/aleph/vm/orchestrator/supervisor.py
nesitor: I don't think so, because we aren't using it now.

src/aleph/vm/pool.py
hoh: > Why to complicate it? Which are the advantages of doing it with AsyncSession?

Using the Session object makes database calls blocking on the main thread of the application. This means that all tasks running on the orchestrator (proxying HTTP requests, starting new VMs, ...) will be waiting for that database call to return.  Using an AsyncSession would make this wait in the background and allow the main event loop to handle other events in the meantime.

nesitor: Okay, this is solved in https://github.com/aleph-im/aleph-vm/pull/512 PR

src/aleph/vm/orchestrator/metrics.py
MHHukiewitz: I guess we reverted to `Session` then?

MHHukiewitz (approved):
LGTM

PR #512, opened 2023-12-15
Work in progress: Monitor payments via balances and streams

hoh:
This is the skeleton for monitoring payments via balances and via token streams.

It will stop VMs until the balance of the user is sufficient. The order is still not set (should be the last resources created stopped first).

The implementation of the payment specific functions is still missing, see `payment.py`.


src/aleph/vm/pool.py
```diff
@@ -213,3 +213,17 @@ def get_instance_executions(self) -> Iterable[VmExecution]:
             for _vm_hash, execution in self.executions.items()
             if execution.is_running and execution.is_instance
         )
+
+    def get_executions_by_sender(self, payment_type: PaymentType) -> Iterable[Tuple[str, list[VmExecution]]]:
+        """Return all executions of the given type, grouped by sender and by chain."""
+        executions_by_sender: dict[str, list[VmExecution]] = {}
+        for vm_hash, execution in self.executions.items():
```
nesitor, line 220: I think that this should only get the persistent VMs for PAYG.

src/aleph/vm/orchestrator/tasks.py
```diff
@@ -127,3 +134,46 @@ async def stop_watch_for_messages_task(app: web.Application):
...
+    pool: VmPool = app["vm_pool"]
+    while True:
+        await asyncio.sleep(settings.PAYMENT_MONITOR_INTERVAL)
+
+        # Check if the balance held in the wallet is sufficient holder tier resources
+        for sender, executions in pool.get_executions_by_sender(payment=PaymentType.hold):
+            balance = await get_balance(sender)
+
+            # Stop executions until the required balance is reached
+            required_balance = await get_required_balance(executions)
```
nesitor, line 150: This mechanism is not valid for hold tiers, because it only takes into account the executions on that CRN, but means that some user with some ALEPH tokens can allocate different VMs on the different CRNs with the less amount of tokens.

src/aleph/vm/pool.py
MHHukiewitz: Why? We can just calculate 1/10th of the price for on-demand VMs with PAYG, at least for now.

Later on we really need more accurate measurements and stream stop/start when using on-demand, or some different kind of payment method alltogether, if it doesn't prove feasible to rapidly start and stop streams.

src/aleph/vm/orchestrator/tasks.py
```diff
@@ -127,3 +134,46 @@ async def stop_watch_for_messages_task(app: web.Application):
...
+
+
+async def monitor_payments(app: web.Application):
+    logger.debug("Monitoring balances")
+    pool: VmPool = app["vm_pool"]
+    while True:
+        await asyncio.sleep(settings.PAYMENT_MONITOR_INTERVAL)
+
+        # Check if the balance held in the wallet is sufficient holder tier resources
+        for sender, executions in pool.get_executions_by_sender(payment=PaymentType.hold):
```
MHHukiewitz, line 146: Should we explicitly sort them chronologically before checking them? What should be the default behavior when looking for the first VMs to shut down? Latest first?

src/aleph/vm/orchestrator/tasks.py
MHHukiewitz: How? This does mean that either the scheduler or CCN is currently setting the used ALEPH balance for each user?

MHHukiewitz: Even if so, this check here is still a good sanity check, just-in-case.

MHHukiewitz (commented):
A few questions regarding the balance checks

MHHukiewitz:
Should `main` be indeed the target branch? @hoh @nesitor

src/aleph/vm/orchestrator/payment.py
```diff
@@ -0,0 +1,160 @@
...
+    return size
+
+
+def get_required_flow(executions: Iterable[VmExecution]) -> Decimal:
+    """Compute the flow required for the resources of the user from the messages and the pricing aggregate"""
+    flow = Decimal(0)
+    for execution in executions:
+        flow += compute_execution_flow_cost(execution)
+
+    return Decimal(flow)
```
MHHukiewitz, line 132: Looks like a duplicate to `get_required_balance()`. Maybe fuse them into one function?

src/aleph/vm/orchestrator/payment.py
```diff
@@ -0,0 +1,160 @@
...
+    compute_units_required = _get_nb_compute_units(execution)
+    compute_unit_multiplier = _get_compute_unit_multiplier(execution)
+
+    compute_unit_price = (
+        Decimal(compute_units_required) * Decimal(compute_unit_multiplier) * Decimal(compute_unit_cost_second)
+    )
+
+    price = compute_unit_price + _get_additional_storage_flow_price(execution)
+
+    return Decimal(price)
```
MHHukiewitz, line 148: Looks also like some code duplication compared to `compute_execution_hold_cost()`.

src/aleph/vm/orchestrator/payment.py
```diff
@@ -0,0 +1,160 @@
...
+def _get_additional_storage_flow_price(execution: VmExecution) -> Decimal:
+    additional_storage_hour_price = 0.000000977  # TODO: Get from PAYG aggregate
+    additional_storage_second_price = additional_storage_hour_price / Hour
+    nb_compute_units = execution.vm.hardware_resources.vcpus
+    free_storage_per_compute_unit = 2 * GiB if not execution.persistent else 20 * GiB
+
+    total_volume_size = _get_execution_storage_size(execution)
+    additional_storage = max(total_volume_size - (free_storage_per_compute_unit * nb_compute_units), 0)
+    price = additional_storage / additional_storage_second_price / MiB
+    return Decimal(price)
```
MHHukiewitz, line 160: Shouldn't all price calculations, in every function, be done with Decimal? I see potential for inaccurate prices here.

src/aleph/vm/orchestrator/tasks.py
```diff
@@ -127,3 +134,51 @@ async def stop_watch_for_messages_task(app: web.Application):
...
+                # Stop executions until the required balance is reached
+                required_balance = get_required_balance(executions)
+                while balance < required_balance:
+                    last_execution = executions.pop(-1)
+                    logger.debug(f"Stopping {last_execution} due to insufficient stream")
+                    await last_execution.stop()
+                    required_balance = get_required_balance(executions)
+
+        # Check if the balance held in the wallet is sufficient stream tier resources
+        for sender, chains in pool.get_executions_by_sender(payment_type=PaymentType.superfluid).items():
```
MHHukiewitz, line 159: why not use `.is_stream()` instead in `.get_executions_by_sender()`?

src/aleph/vm/orchestrator/views/__init__.py
```diff
@@ -253,7 +253,13 @@ async def update_allocations(request: web.Request):
     allocations = allocation.persistent_vms | allocation.instances
     # Make a copy since the pool is modified
     for execution in list(pool.executions.values()):
-        if execution.vm_hash not in allocations and execution.is_running:
+        if (
+            execution.vm_hash not in allocations
+            and execution.is_running
+            and (
+                not execution.message.payment or (execution.message.payment and not execution.message.payment.is_stream)
```
MHHukiewitz, line 260: Hard to read, can you make a little helper function for that plz?

src/aleph/vm/orchestrator/views/__init__.py
```diff
@@ -318,3 +324,61 @@ async def update_allocations(request: web.Request):
...
+        data = await request.json()
+        vm_notification = VMNotification.parse_obj(data)
+    except ValidationError as error:
+        return web.json_response(data=error.json(), status=web.HTTPBadRequest.status_code)
+
+    pubsub: PubSub = request.app["pubsub"]
+    pool: VmPool = request.app["vm_pool"]
+
+    # First free resources from persistent programs and instances that are not scheduled anymore.
+    instance = vm_notification.instance
```
MHHukiewitz, line 343: I don't see where you are freeing resources here

MHHukiewitz (commented):
Some comments, but overall LGTM

src/aleph/vm/orchestrator/payment.py
```diff
@@ -0,0 +1,160 @@
...
+    # TODO: Convert chain str to ID
+    chain_id = 43113
+    superfluid_instance = CFA_V1(settings.PAYMENT_RPC_SERVER, chain_id)
+
+    super_token: HexAddress = to_normalized_address(settings.PAYMENT_SUPER_TOKEN)
+    sender_address: HexAddress = to_normalized_address(sender)
+    receiver_address: HexAddress = to_normalized_address(receiver)
+
+    flow_data: Web3FlowInfo = superfluid_instance.get_flow(super_token, sender_address, receiver_address)
+    stream = from_wei(flow_data["flowRate"], "ether")
```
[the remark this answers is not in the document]
nesitor: Is flow per seconds, but the value is given using the Wei format, it has 18 decimals and this is because I convert it to `ether`.

src/aleph/vm/orchestrator/payment.py
```diff
@@ -0,0 +1,160 @@
...
+        raise ValueError("Unknown format {}, attempted to normalize to {}".format(value, hex_address))
+
+
+def get_required_balance(executions: Iterable[VmExecution]) -> Decimal:
+    """Get the balance required for the resources of the user from the messages and the pricing aggregate."""
+    balance = Decimal(0)
+    for execution in executions:
+        balance += compute_execution_hold_cost(execution)
+
+    return Decimal(balance)
```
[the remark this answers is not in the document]
nesitor: This is done on the previous step, so here it's not necessary IMO

src/aleph/vm/orchestrator/payment.py
```diff
@@ -0,0 +1,160 @@
...
+    nb_compute_units = execution.vm.hardware_resources.vcpus
+    free_storage_per_compute_unit = 2 * GiB if not execution.persistent else 20 * GiB
+
+    total_volume_size = _get_execution_storage_size(execution)
+    additional_storage = max(total_volume_size - (free_storage_per_compute_unit * nb_compute_units), 0)
+    price = Decimal(additional_storage) / 20 / MiB
+    return price
+
+
+def _get_nb_compute_units(execution: VmExecution) -> int:
```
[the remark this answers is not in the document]
nesitor: I think that means **number**, but I just used the same names from the CCN repo.

src/aleph/vm/orchestrator/payment.py
nesitor: Yes, but anyway, all these parts should be replaced soon by a centric pricing API instead implement it in every service.

src/aleph/vm/orchestrator/tasks.py
nesitor: This is the loop to get the executions by sender and by chain, and the `is_stream` is a property of the payment field on the message, so it's not related.

src/aleph/vm/orchestrator/views/__init__.py
nesitor: Here I don't free any resources, this endpoint is only to notify the CRN to create the instance. Anyway, we need to decide first if we will use that endpoint or will use the Scheduler approach.

src/aleph/vm/orchestrator/payment.py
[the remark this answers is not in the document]
MHHukiewitz: Would be good to document this, it's not really obvious from the function or variable name

src/aleph/vm/orchestrator/tasks.py
MHHukiewitz: Yes, but my point was to use `is_stream` inside of `get_executuons_by_sender`, instead of passing the `PaymentType` as a parameter. Something like:

```py
get_executions_by_sender(is_stream=True)
```

MHHukiewitz (changes requested):
Still has open issues

src/aleph/vm/orchestrator/payment.py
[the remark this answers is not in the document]
nesitor: For now, I've added a comment to document it in the docs later.

src/aleph/vm/orchestrator/payment.py
```diff
@@ -0,0 +1,160 @@
...
+    memory = math.ceil(execution.vm.hardware_resources.memory / 2048)
+    nb_compute_units = cpu if cpu >= memory else memory
+    return nb_compute_units
+
+
+def _get_compute_unit_multiplier(execution: VmExecution) -> int:
+    compute_unit_multiplier = 1
+    if not execution.persistent and execution.message.environment.internet:
+        compute_unit_multiplier += 1
+    return compute_unit_multiplier
```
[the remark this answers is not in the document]
nesitor: Yes, this was totally confirmed by @moshemalawach

src/aleph/vm/orchestrator/payment.py
nesitor: It's a different method, for a different purpose, and I don't think that merge will be good for code understanding.

src/aleph/vm/orchestrator/payment.py
```diff
@@ -0,0 +1,153 @@
...
+    price = compute_unit_price + _get_additional_storage_flow_price(execution)
+
+    return Decimal(price)
+
+
+def _get_additional_storage_flow_price(execution: VmExecution) -> Decimal:
+    # TODO: Use PAYMENT_PRICING_AGGREGATE when possible
+    additional_storage_hour_price = 0.000000977
+    additional_storage_second_price = Decimal(additional_storage_hour_price) / Decimal(Hour)
+    nb_compute_units = execution.vm.hardware_resources.vcpus
```
MHHukiewitz, line 145:
```suggestion
    nb_compute_units = _get_nb_compute_units(execution)
```

src/aleph/vm/utils.py
```diff
@@ -136,3 +139,38 @@ def check_disk_space(bytes_to_use: int) -> bool:
 
 class NotEnoughDiskSpaceError(OSError):
     pass
+
+
+def get_path_size(path: Path) -> int:
```
hoh, line 144: Add a docstring

src/aleph/vm/utils.py
```diff
@@ -136,3 +139,38 @@ def check_disk_space(bytes_to_use: int) -> bool:
...
+    if path.is_dir():
+        return sum([f.stat().st_size for f in path.glob("**/*")])
+    elif path.is_block_device():
+        return get_block_device_size(str(path))
+    elif path.is_file():
+        return path.stat().st_size
+    else:
+        raise NotImplementedError
```
hoh, line 152: What scenario could lead to this ?

src/aleph/vm/utils.py
```diff
@@ -136,3 +139,38 @@ def check_disk_space(bytes_to_use: int) -> bool:
...
+
+
+def get_block_device_size(device: str) -> int:
+    output = subprocess.run(
```
hoh, line 156: This must be run asynchronously

src/aleph/vm/conf.py
```diff
@@ -201,6 +203,25 @@ class Settings(BaseSettings):
...
+    )
+    PAYMENT_RECEIVER_ADDRESS: str = Field(
+        default="",
+        description="Address of the account receiving payments",
+    )
+    PAYMENT_SUPER_TOKEN: str = Field(
+        default="0x1290248e01ed2f9f863a9752a8aad396ef3a1b00",
+        description="Address of the ALEPH SuperToken on SuperFluid",
+    )
+    PAYMENT_PRICING_AGGREGATE: str = ""  # TODO: Missing
```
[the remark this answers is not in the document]
nesitor: Using an ItemHas I fail if it's not set. So, We can change it to ItemHash when we implement the aggregate. I will update the comment.

src/aleph/vm/conf.py
```diff
@@ -201,6 +203,25 @@ class Settings(BaseSettings):
...
+    PAYMENT_MONITOR_INTERVAL: float = Field(
+        default=60.0,
+        description="Interval in seconds between payment checks",
+    )
+    PAYMENT_RECEIVER_ADDRESS: str = Field(
+        default="",
+        description="Address of the account receiving payments",
+    )
+    PAYMENT_SUPER_TOKEN: str = Field(
+        default="0x1290248e01ed2f9f863a9752a8aad396ef3a1b00",
```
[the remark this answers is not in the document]
nesitor: For now the Token is only added on the Testnet, so when we put into production, we can put the good one.

src/aleph/vm/orchestrator/views/__init__.py
```diff
@@ -318,3 +318,60 @@ async def update_allocations(request: web.Request):
...
+    instance_item_hash = ItemHash(instance)
+    try:
+        await start_persistent_vm(instance_item_hash, pubsub, pool)
+        successful = True
+    except vm_creation_exceptions as error:
+        logger.exception(error)
+        scheduling_errors[instance_item_hash] = error
+        successful = False
+
+    failing = set(scheduling_errors.keys())
```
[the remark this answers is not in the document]
nesitor: It uses the same structure as the `update_allocation` method.

src/aleph/vm/orchestrator/views/__init__.py
```diff
@@ -318,3 +318,60 @@ async def update_allocations(request: web.Request):
...
+        ResourceDownloadError,
+        FileTooLargeError,
+        VmSetupError,
+        MicroVMFailedInitError,
+        HostNotFoundError,
+    )
+
+    scheduling_errors: dict[ItemHash, Exception] = {}
+
+    instance_item_hash = ItemHash(instance)
```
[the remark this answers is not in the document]
nesitor: No, because the `VMNotification.instance` field is a `string`, so we need to convert to an `ItemHash`, exactly that it makes on the `update_allocation` method.

packaging/Makefile
```diff
@@ -15,7 +15,7 @@ debian-package-code:
 	cp ../examples/instance_message_from_aleph.json ./aleph-vm/opt/aleph-vm/examples/instance_message_from_aleph.json
 	cp -r ../examples/data ./aleph-vm/opt/aleph-vm/examples/data
 	mkdir -p ./aleph-vm/opt/aleph-vm/examples/volumes
-	pip3 install --target ./aleph-vm/opt/aleph-vm/ 'aleph-message==0.4.1' 'jwskate==0.8.0' 'eth-account==0.9.0' 'sentry-sdk==1.31.0' 'qmp==1.1.0'
+	pip3 install --target ./aleph-vm/opt/aleph-vm/ 'aleph-message@git+https://github.com/aleph-im/aleph-message.git@mhh-pay-as-you-go-v1#egg=aleph-message' 'jwskate==0.8.0' 'eth-account==0.9.0' 'sentry-sdk==1.31.0' 'qmp==1.1.0' 'superfluid@git+https://github.com/Godspower-Eze/superfluid.py#egg=superfluid'
```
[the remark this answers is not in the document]
nesitor: It isn't, so I had to add as a Git repository directly

packaging/aleph-vm/etc/aleph-vm/supervisor.env
```diff
@@ -1,2 +1,3 @@
 ALEPH_VM_PRINT_SYSTEM_LOGS=True
 ALEPH_VM_DOMAIN_NAME=vm.example.org
+ALEPH_VM_PAYMENT_RECEIVER_ADDRESS=
```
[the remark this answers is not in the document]
nesitor: We agreed to have it twice for double checks.

src/aleph/vm/orchestrator/tasks.py
```diff
@@ -127,3 +134,51 @@ async def stop_watch_for_messages_task(app: web.Application):
         await app["messages_listener"]
     except asyncio.CancelledError:
         logger.debug("Task messages_listener is cancelled now")
+
+
+async def monitor_payments(app: web.Application):
+    logger.debug("Monitoring balances")
+    pool: VmPool = app["vm_pool"]
+    while True:
+        await asyncio.sleep(settings.PAYMENT_MONITOR_INTERVAL)
```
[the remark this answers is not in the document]
nesitor: We can do it after the POC, for the final release.

src/aleph/vm/models.py
```diff
@@ -110,6 +110,10 @@ def vm_id(self) -> Optional[int]:
     def controller_service(self) -> str:
         return f"aleph-vm-controller@{self.vm_hash}.service"
 
+    @property
+    def is_payment_stream(self) -> bool:
+        return self.message.payment.is_stream if self.message.payment else False
```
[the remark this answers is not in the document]
nesitor: No in this case, because the payment field is not set always.

src/aleph/vm/orchestrator/payment.py
```diff
@@ -0,0 +1,153 @@
...
+async def get_balance(address: str) -> Decimal:
+    """Get the balance of the user from the PyAleph."""
+    # See https://github.com/aleph-im/pyaleph/blob/master/src/aleph/web/controllers/routes.py#L62
+    # "/api/v0/addresses/{address}/balance"
+    async with aiohttp.ClientSession() as session:
+        url = f"{settings.API_SERVER}/api/v0/{address}/balance"
+        resp = await session.get(url)
+
+        if not resp.ok:
+            return Decimal(0)
```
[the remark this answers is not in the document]
nesitor: IMO I think that is better to return 0, but yes, for the release we can implement it.

src/aleph/vm/orchestrator/payment.py
```diff
@@ -0,0 +1,153 @@
...
+
+        resp_data = await resp.json()
+        return resp_data["balance"] if resp_data["balance"] else 0
+
+
+def get_stream(sender: str, receiver: str, chain) -> Decimal:
+    # See https://community.aleph.im/t/pay-as-you-go-using-superfluid/98/11
+
+    # TODO: Convert chain str to ID
```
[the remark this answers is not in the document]
nesitor: We will add the implementation on the SDK in theory

src/aleph/vm/orchestrator/payment.py
```diff
@@ -0,0 +1,153 @@
...
+    # "/api/v0/addresses/{address}/balance"
+    async with aiohttp.ClientSession() as session:
+        url = f"{settings.API_SERVER}/api/v0/{address}/balance"
+        resp = await session.get(url)
+
+        if not resp.ok:
+            return Decimal(0)
+
+        resp_data = await resp.json()
+        return resp_data["balance"] if resp_data["balance"] else 0
```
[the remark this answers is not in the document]
nesitor: Yes, but is pending to add the SDK as a dependency.

src/aleph/vm/orchestrator/payment.py
```diff
@@ -0,0 +1,153 @@
...
+    total_volume_size = _get_execution_storage_size(execution)
+    additional_storage = max(total_volume_size - (free_storage_per_compute_unit * nb_compute_units), 0)
+    price = Decimal(additional_storage) / 20 / MiB
+    return price
+
+
+def _get_nb_compute_units(execution: VmExecution) -> int:
+    cpu = execution.vm.hardware_resources.vcpus
+    memory = math.ceil(execution.vm.hardware_resources.memory / 2048)
+    nb_compute_units = cpu if cpu >= memory else memory
```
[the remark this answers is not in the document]
nesitor: It's exactly the same implementation that PyAleph. The compute units only depends on the CPU and RAM, we include the storage, but it changes depending on the type of VM.

src/aleph/vm/orchestrator/payment.py
```diff
@@ -0,0 +1,153 @@
...
+    compute_unit_multiplier = 1
+    if not execution.persistent and execution.message.environment.internet:
+        compute_unit_multiplier += 1
+    return compute_unit_multiplier
+
+
+def _get_execution_storage_size(execution: VmExecution) -> int:
+    size = 0
+
+    if execution.is_instance:
```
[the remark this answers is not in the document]
nesitor: All this will be removed when we implement the pricing library, that for sure we need to implement it soon.

src/aleph/vm/orchestrator/payment.py
```diff
@@ -0,0 +1,153 @@
...
+        size += execution.message.rootfs.size_mib * MiB
+    elif execution.is_program:
+        if isinstance(execution.resources, AlephProgramResources):
+            size += get_path_size(execution.resources.code_path)
+            if execution.resources.data_path:
+                size += get_path_size(execution.resources.data_path)
+
+    for volume in execution.resources.volumes:
+        size += get_path_size(volume.path_on_host)
```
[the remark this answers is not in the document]
nesitor: Now the `get_path_size` uses an `await`, so we cannot put it in one line.

src/aleph/vm/orchestrator/supervisor.py
```diff
@@ -127,7 +134,9 @@ def run():
     try:
         if settings.WATCH_FOR_MESSAGES:
             app.on_startup.append(start_watch_for_messages_task)
+            app.on_startup.append(start_payment_monitoring_task)
             app.on_cleanup.append(stop_watch_for_messages_task)
+            app.on_cleanup.append(stop_balances_monitoring_task)
```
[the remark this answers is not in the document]
nesitor: You put that name :sweat_smile:

src/aleph/vm/orchestrator/payment.py
```diff
@@ -0,0 +1,153 @@
+import logging
```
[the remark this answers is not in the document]
nesitor: I had to prepare it in days for the POC, we can add it for the release. Anyway, as I put in other comments, all the pricing methods will be moved to a single library that will be imported here.

packaging/Makefile
```diff
@@ -15,7 +15,7 @@ debian-package-code:
 	cp ../examples/instance_message_from_aleph.json ./aleph-vm/opt/aleph-vm/examples/instance_message_from_aleph.json
 	cp -r ../examples/data ./aleph-vm/opt/aleph-vm/examples/data
 	mkdir -p ./aleph-vm/opt/aleph-vm/examples/volumes
-	pip3 install --target ./aleph-vm/opt/aleph-vm/ 'aleph-message==0.4.1' 'jwskate==0.8.0' 'eth-account==0.9.0' 'sentry-sdk==1.31.0' 'qmp==1.1.0'
+	pip3 install --target ./aleph-vm/opt/aleph-vm/ 'aleph-message@git+https://github.com/aleph-im/aleph-message.git@mhh-pay-as-you-go-v1#egg=aleph-message' 'jwskate==0.8.0' 'eth-account==0.9.0' 'sentry-sdk==1.31.0' 'qmp==1.1.0' 'superfluid@git+https://github.com/Godspower-Eze/superfluid.py#egg=superfluid' 'sqlalchemy[asyncio]' 'aiosqlite==0.19.0' 'alembic==1.13.1'
```
hoh, line 18: We already install `alembic` using the Debian packages, see `packaging/aleph-vm/DEBIAN/control`. This would create a conflict of versions. [Debian 10 alembic package](https://packages.debian.org/bullseye/alembic). I am open to only use the vendored version if it brings enough value compared to the one from the distribution.

packaging/Makefile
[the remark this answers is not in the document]
hoh: That is annoying. We must at least pin the exact git commit we use in such cases, preferably also fork the repository and install from our fork or publish it ourserlves on PyPI. Motivation: We want to ensure it stays available and auditable in case of supply chain attack.

src/aleph/vm/conf.py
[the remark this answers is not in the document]
hoh: What about `Optional[ItemHash]` then ?

src/aleph/vm/orchestrator/payment.py
```diff
@@ -0,0 +1,153 @@
...
+    nb_compute_units = _get_nb_compute_units(execution)
+    free_storage_per_compute_unit = 2 * GiB if not execution.persistent else 20 * GiB
+
+    total_volume_size = _get_execution_storage_size(execution)
+    additional_storage = max(total_volume_size - (free_storage_per_compute_unit * nb_compute_units), 0)
+    price = Decimal(additional_storage) / 20 / MiB
+    return price
+
+
+def _get_nb_compute_units(execution: VmExecution) -> int:
```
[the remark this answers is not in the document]
hoh: What about the storage part of the compute unit ?

src/aleph/vm/models.py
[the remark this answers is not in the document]
hoh: The two codes are identical in their result. I think it's safe to commit my simplification and take responsability for it.

src/aleph/vm/utils.py
hoh: Then we should use an `assert False`. This gives the impression that we're going to implement something more later.

packaging/Makefile
hoh: Moved to a separate issue.

hoh, line 18: Should 'aleph-message==0.4.2rc1' not work ?

```suggestion
	pip3 install --target ./aleph-vm/opt/aleph-vm/ 'aleph-message==0.4.2rc1' 'jwskate==0.8.0' 'eth-account==0.9.0' 'sentry-sdk==1.31.0' 'qmp==1.1.0' 'superfluid@git+https://github.com/Godspower-Eze/superfluid.py#egg=superfluid' 'sqlalchemy[asyncio]' 'aiosqlite==0.19.0' 'alembic==1.13.1'
```

src/aleph/vm/orchestrator/tasks.py
MHHukiewitz: Sorry, but I have to insist on that one. If we bring Solana support in the next months, we will have to do this anyways, as we won't use "superfluid" there, but some other protocol, which will then also return `True` on `execution.is_stream()`.

MHHukiewitz (approved):
Nothing too major now. I will approve it but not merge it yet, as there is one issue left that I kind of care about.

MHHukiewitz (approved):

MHHukiewitz:
Need to rebase/merge :sweat_smile:

nesitor:
Yes, it's needed because you merged the other PR into main, and this should
be merged before. Anyway I will update the branch and solve the conflicts...

hoh:
I rebased this branch on `main` and will force push. 

The old branch is available here: https://github.com/aleph-im/aleph-vm/pull/new/hoh-monitor-payments-before-rebase

src/aleph/vm/orchestrator/views/__init__.py
```diff
@@ -319,3 +319,61 @@ async def update_allocations(request: web.Request):
         },
         status=status_code,
     )
+
+
+async def notify_allocation(request: web.Request):
+    """Notify instance allocation, only used for Pay as you Go feature"""
+    if not authenticate_api_request(request):
+        return web.HTTPUnauthorized(text="Authentication token received is invalid")
```
hoh, line 327: Why do we ask for scheduler authentication here ?

hoh: This function should however check the balance of the flow.

src/aleph/vm/orchestrator/payment.py
```diff
@@ -0,0 +1,156 @@
...
+
+    # TODO: Convert chain str to ID
+    chain_id = 43113
+    superfluid_instance = CFA_V1(settings.PAYMENT_RPC_API, chain_id)
+
+    super_token: HexAddress = to_normalized_address(settings.PAYMENT_SUPER_TOKEN)
+    sender_address: HexAddress = to_normalized_address(sender)
+    receiver_address: HexAddress = to_normalized_address(receiver)
+
+    flow_data: Web3FlowInfo = superfluid_instance.get_flow(super_token, sender_address, receiver_address)
```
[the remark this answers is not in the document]
hoh: Fixed in my improvements branch with commit https://github.com/aleph-im/aleph-vm/pull/517/commits/7eb8311d10f96d3ab3cc8f3ee6bd7ca53e787199

src/aleph/vm/orchestrator/views/__init__.py
hoh: https://github.com/aleph-im/aleph-vm/pull/517/commits/56f09ec95a9a9bc2c10842ae77628c619cc20ab2

src/aleph/vm/orchestrator/payment.py
```diff
@@ -0,0 +1,156 @@
...
+async def get_required_flow(executions: Iterable[VmExecution]) -> Decimal:
+    """Compute the flow required for the resources of the user from the messages and the pricing aggregate"""
+    flow = Decimal(0)
+    for execution in executions:
+        flow += await compute_execution_flow_cost(execution)
+
+    return Decimal(flow)
+
+
+async def compute_execution_flow_cost(execution: VmExecution) -> Decimal:
```
hoh, line 126: Requiring a VM execution instead of the message makes these functions difficult to test and to reuse.

hoh:
**TODO**: 
Call the future API of pyaleph that computes the pricing of an instance based on the item_hash.
@nesitor will prepare the API for this purpose.

src/aleph/vm/orchestrator/tasks.py
hoh: Let's refactor that later then ;-)

PR #530, opened 2024-02-02
Add diagnostic for legacy VM

hoh:
- Fix: Status check using FastAPI used hardcoded VM ID
- Fix: Compatibility with the legacy VM could not be checked


MHHukiewitz (approved):
LGTM

PR #534, opened 2024-02-08
Solve issues on Operator API

nesitor:
Problem: Now the operator API just allow to reboot well ephemeral VMs, not the persistent ones.

Solution: Use the VM Pool to control reboot, stop and erase operations.


MHHukiewitz (approved):
LGTM

PR #535, opened 2024-02-13
Fix: TypeError: catching classes that do not inherit from BaseException is not allowed

hoh:
This code used the wrong class for the timeout exception.


MHHukiewitz (approved):
LGTM

PR #536, opened 2024-02-13
Fix: An IndexError was raised if no execution was left for this sender.

hoh:
Solution: Catch and handle the IndexError.


MHHukiewitz (approved):
LGTM

PR #537, opened 2024-02-13
Fix: Status check fastapi view could crash due to ServerDisconnectedError.

hoh:
Solution: Catch the ServerDisconnectedError and return an appropriate error message instead.


MHHukiewitz (approved):
LGTM

PR #538, opened 2024-02-13
Fix: Restarting the NDP Proxy sometimes crashed

hoh:
This is likely due because it was restarted too quickly.

Solution: Wrap the error in a custom class and retry after 5 seconds.

This is still a work in progress as a better approach should limit the frequency of restarts instead, for example using an event or queue.


odesenfans:
Closing as part of a cleanup of pre-#1000 PRs. This branch predates the 2026-02-18 force-push of main, so it can no longer be rebased (GitHub shows the whole history as the diff). Superseded: the NDP proxy on main (src/aleph/vm/network/ndp_proxy.py) already handles the subprocess error and debounces restarts, which is the better approach this PR's description called for.

PR #540, opened 2024-02-14
Fix: `socket.getaddrinfo` does not always return 2 values

hoh:
A user reported that this function crashed on his host due to `ValueError: too many values to unpack (expected 2)`.

Solution: Process each returned tuple instead of assuming that two of them are always returned.

This fixes the issue both in the orchestrator and in the diagnostic VM.


MHHukiewitz (commented):
LGTM, don't be too hard on yourself, it is well readable

MHHukiewitz (approved):

PR #541, opened 2024-02-14
Detect already running Persistent VMs

nesitor:
Problem: Persistent VMs running were not detected after the orchestrator reboot.

Solution: Don't delete the entire table on the start process. Also added small fixes about guest_api and solved code quality issue.

### DIFF
--- a/src/aleph/vm/controllers/firecracker/executable.py
+++ b/src/aleph/vm/controllers/firecracker/executable.py
@@ -307,7 +307,10 @@ class AlephFirecrackerExecutable(Generic[ConfigurationType], AlephVmControllerIn
 
     async def stop_guest_api(self):
         if self.guest_api_process and self.guest_api_process.is_alive():
-            self.guest_api_process.join(timeout=10)
+            self.guest_api_process.terminate()
+            await asyncio.sleep(5)
+            if self.guest_api_process.is_alive():
+                self.guest_api_process.kill()
 
     async def teardown(self):
         if self.fvm:
--- a/src/aleph/vm/controllers/firecracker/executable.py
+++ b/src/aleph/vm/controllers/firecracker/executable.py
@@ -306,8 +306,8 @@ class AlephFirecrackerExecutable(Generic[ConfigurationType], AlephVmControllerIn
         logger.debug(f"started guest API for {self.vm_id}")
 
     async def stop_guest_api(self):
-        if self.guest_api_process and self.guest_api_process._popen:
-            self.guest_api_process.terminate()
+        if self.guest_api_process and self.guest_api_process.is_alive():
+            self.guest_api_process.join(timeout=10)
 
     async def teardown(self):
         if self.fvm:
--- a/src/aleph/vm/orchestrator/views/operator.py
+++ b/src/aleph/vm/orchestrator/views/operator.py
@@ -170,7 +170,7 @@ async def operate_reboot(request: web.Request, authenticated_sender: str) -> web
     if execution.is_running:
         logger.info(f"Rebooting {execution.vm_hash}")
         if execution.persistent:
-            await pool.systemd_manager.restart(execution.controller_service)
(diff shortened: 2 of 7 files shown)


src/aleph/vm/orchestrator/views/operator.py
```diff
@@ -197,7 +197,7 @@ async def operate_erase(request: web.Request, authenticated_sender: str) -> web.
 
     # Stop the VM
     await pool.stop_vm(execution.vm_hash)
-    await pool.forget_vm(execution.vm_hash)
+    pool.forget_vm(execution.vm_hash)
```
hoh, line 200: I am wondering whether this makes sense here... :thinking: 

We use an event, `VmExecution.stop_event`, to automatically forget a VM once it has stopped.
See https://github.com/aleph-im/aleph-vm/blob/0.4.0a3/src/aleph/vm/pool.py#L131-L135

The line above, `await pool.stop_vm(execution.vm_hash)`, should await `execution.stop()`, which ends with `self.stop_event.set()`, triggering the forget.

Is there an issue in that logic ?

src/aleph/vm/controllers/firecracker/executable.py
```diff
@@ -306,8 +306,8 @@ async def start_guest_api(self):
         logger.debug(f"started guest API for {self.vm_id}")
 
     async def stop_guest_api(self):
-        if self.guest_api_process and self.guest_api_process._popen:
-            self.guest_api_process.terminate()
+        if self.guest_api_process and self.guest_api_process.is_alive():
+            self.guest_api_process.join(timeout=10)
```
hoh, line 310: Why remove `.terminate()` ?

According to [the Python documentation](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.pool.Pool.join):

> `join()`: Wait for the worker processes to exit. One must call [close()](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.pool.Pool.close) or [terminate()](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.pool.Pool.terminate) before using [join()](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.pool.Pool.join).

src/aleph/vm/orchestrator/views/operator.py
```diff
@@ -170,7 +170,7 @@ async def operate_reboot(request: web.Request, authenticated_sender: str) -> web
     if execution.is_running:
         logger.info(f"Rebooting {execution.vm_hash}")
         if execution.persistent:
-            await pool.systemd_manager.restart(execution.controller_service)
+            pool.systemd_manager.restart(execution.controller_service)
```
hoh, line 173: I see that the underlying function calls `dbus.Interface(...).RestartUnit(service, "replace")`.

Is that call IO limited by communicating via DBus in a synchronous manner ? In that case, we should open another issue to make the methods of `SystemDManager` asynchronous and prevent them from blocking the entire orchestrator.

src/aleph/vm/orchestrator/views/operator.py
nesitor: This event is only used for ephemeral VMs, because for permanent we don't use anymore if we restart the orchestrator, because is called on `create_a_vm` that implements a task to call `forget_on_stop` when it's stopped.
To improve it, we can implement the `forget_on_stop` method to the entire pool receiving an extra `vm_hash` argument, and create the task on the `VMExecution` instance creation.
Anyway, the `forget_vm` method has a try catch if the execution doesn't exist, so will never fail.

src/aleph/vm/orchestrator/views/operator.py
nesitor: Yes, we can open issues and improve it later to avoid block the entire orchestrator loop.

hoh: https://app.clickup.com/t/86bxjfcp2

hoh (approved):

PR #542, opened 2024-02-14
Added CORS support on supervisor endpoints

nesitor:
Fix: Implemented CORS support on supervisor endpoints.


hoh:
Thanks !

Can we use the version of aiohttp-cors shipped in Debian ?
https://packages.debian.org/bullseye/python3-aiohttp-cors

Should we remove the duplication with a custom decorator that wraps `custom_cors(...)` ?

hoh (approved):

PR #543, opened 2024-02-14
Fix: Linux kernel from package could not be on a different device

hoh:
If the Linux kernel provided is was another device than the execution root, creating a hardlink within the jailer directory would fail.

Solution: Copy the kernel to the execution root during setup.

### DIFF
--- a/src/aleph/vm/conf.py
+++ b/src/aleph/vm/conf.py
@@ -16,7 +16,7 @@ from pydantic import BaseSettings, Field, HttpUrl
 from pydantic.env_settings import DotenvType, env_file_sentinel
 from pydantic.typing import StrPath
 
-from aleph.vm.utils import is_command_available
+from aleph.vm.utils import file_hashes_differ, is_command_available
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


src/aleph/vm/conf.py
```diff
@@ -356,6 +357,14 @@ def setup(self):
 
         os.makedirs(self.EXECUTION_ROOT, exist_ok=True)
 
+        # If the Linux kernel provided is on another device than the execution root,
+        # copy it to the execution root to allow hardlink creation within jailer directories.
+        if os.stat(self.LINUX_PATH).st_dev != os.stat(self.EXECUTION_ROOT).st_dev:
+            logger.info("The Linux kernel is on another device than the execution root. Creating a copy.")
+            linux_path_on_device = self.EXECUTION_ROOT / "vmlinux.bin"
+            shutil.copy(self.LINUX_PATH, linux_path_on_device)
```
nesitor, line 365: Here also adding a `self.LINUX_PATH.isfile()` check, should be good to avoid not existing errors on copy

hoh: `shutil.copy()` should override the existing file. We could optimize the process to only override it if the two files differ :thinking:

nesitor (approved):

PR #544, opened 2024-02-16
Add CPU information (model, vendor, frequency) and memory details (clock, size, type) to API

aliel:
Update CPU and memory details by switching to lshw method instead of cpuinfo
Add CPU information (model, vendor, frequency) and memory details (clock, size, type) to API


src/aleph/vm/orchestrator/machine.py
```diff
@@ -0,0 +1,77 @@
...
+def get_cpu_info():
+    hw = get_hardware_info()
+
+    cpu_info = hw["cpu"]
+    architecture = cpu_info["width"]
+
+    if "x86_64" in cpu_info["capabilities"] or "x86-64" in cpu_info["capabilities"]:
+        architecture = "x86_64"
+    elif "arm64" in cpu_info["capabilities"] or "arm-64" in cpu_info["capabilities"]:
+        architecture = "arm64"
```
hoh, line 36: Is this really the best approach to get the architecture ? Seems weird to look for it in the CPU capabilities.
What if a 32 bit system is running on a x86-64 capable system ?

hoh (commented):
Can we have automated tests that ensure that this is working fine ?

Maybe in the `.github/workflows/test-on-droplets-matrix.yml` file ?

hoh:
Can you also provide an example of the JSON format returned by the API ?

src/aleph/vm/orchestrator/machine.py
```diff
@@ -0,0 +1,77 @@
...
+import re
+import subprocess
+from functools import lru_cache
+
+import psutil
+
+
+@lru_cache
+def get_hardware_info():
+    lshw = subprocess.Popen(["lshw", "-sanitize", "-json"], stdout=subprocess.PIPE, shell=False)
```
hoh, line 11: Using `subprocess.Popen` would block the entire orchestrator since `.communicate()` is a synchronous blocking call. This should use https://docs.python.org/3/library/asyncio-subprocess.html instead.

hoh, line 11: A check in the settings that verifies that `lshw` is available on the machine would be a good addition.

src/aleph/vm/orchestrator/machine.py
```diff
@@ -0,0 +1,77 @@
...
+    output, _ = lshw.communicate()
+    data = json.loads(output)
+
+    hw_info = {}
```
hoh, line 15: Can we add type hints on this dictionary ?

src/aleph/vm/orchestrator/machine.py
```diff
@@ -0,0 +1,77 @@
...
+        except:
+            pass
+
+    return {
+        "size": mem_info["size"],
+        "units": mem_info["units"],
+        "type": memory_type,
+        "clock": memory_clock,
+        "clock_units": "Hz",
+    }
```
hoh, line 77: Why not return an instance of `MemoryProperties` ?

src/aleph/vm/orchestrator/machine.py
```diff
@@ -0,0 +1,77 @@
...
+    elif "Advanced Micro Devices [AMD]" in vendor:
+        vendor = "AuthenticAMD"
+
+    return {
+        "architecture": architecture,
+        "vendor": vendor,
+        "model": cpu_info["product"],
+        "frequency": cpu_info["capacity"],
+        "count": psutil.cpu_count(),
+    }
```
hoh, line 52: Why return a dict and not instances of the Properties / Capabilities classes defined blow in the diff ?

src/aleph/vm/orchestrator/machine.py
aliel: here we want to know the architecture supported by the processor, not by the system
Obviously, 64-bit processors support 32-bit systems, but not vice versa.

src/aleph/vm/orchestrator/machine.py
```diff
@@ -0,0 +1,82 @@
...
+@lru_cache
+async def get_memory_info():
+    hw = await get_hardware_info()
+    mem_info = hw["memory"]
+
+    memory_type = ""
+    memory_clock = ""
+
+    for bank in mem_info["children"]:
+        memory_clock = bank["clock"]
```
olethanh, line 67: On  CRN-TEST-2204 I have this in mem_info. No clock key. Should I remove the information conditionally?
```
 {'capabilities': {'ecc': 'Multi-bit error-correcting code (ECC)'},
            'children': [{'claimed': True,
                          'class': 'memory',
                          'description': 'DIMM RAM',
                          'handle': 'DMI:1100',
                          'id': 'bank',
                          'physid': '0',
                          'size': 17179869184,
                          'slot': 'DIMM 0',
                          'units': 'bytes',
                          'vendor': 'QEMU'}],
            'claimed': True,
            'class': 'memory',
            'configuration': {'errordetection': 'multi-bit-ecc'},
            'description': 'System Memory',
            'handle': 'DMI:1000',
            'id': 'memory',
            'physid': '1000',
            'size': 17179869184,
            'units': 'bytes'}}
```

odesenfans:
Closing as part of a cleanup of pre-#1000 PRs. This branch predates the 2026-02-18 force-push of main, so it can no longer be rebased (GitHub shows the whole history as the diff). Superseded: src/aleph/vm/agent/machine.py on main queries lshw and exposes CPU vendor, model and frequency plus memory size, clock and type through the API.

PR #545, opened 2024-02-20
Fix: CI did not run `pytest`

hoh:
We have integration tests, and some unit tests using pytest but the CI does not run them.

Our CI should run `hatch run testing:test`  and ensure that it succeeds.

This requires fixing bugs in the unit tests, the pipeline itself.


MHHukiewitz (approved):
LGTM

PR #549, opened 2024-02-27
Solve random 500 errors on ephemeral VMs

nesitor:
Problem: Sometimes, when a VM expires, the execution is not removed.

Solution: Check if the namespace stills there and if not just remove the execution.

### DIFF
--- a/src/aleph/vm/orchestrator/run.py
+++ b/src/aleph/vm/orchestrator/run.py
@@ -119,6 +119,12 @@ async def run_code_on_request(vm_hash: ItemHash, path: str, pool: VmPool, reques
 
     execution: Optional[VmExecution] = await pool.get_running_vm(vm_hash=vm_hash)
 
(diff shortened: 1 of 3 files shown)


src/aleph/vm/orchestrator/run.py
```diff
@@ -119,6 +119,12 @@ async def run_code_on_request(vm_hash: ItemHash, path: str, pool: VmPool, reques
 
     execution: Optional[VmExecution] = await pool.get_running_vm(vm_hash=vm_hash)
 
+    # Prevent execution issues if the execution resources are empty
+    # TODO: Improve expiration process to avoid that kind of issues.
+    if execution and not execution.has_resources:
```
hoh, line 124: I am afraid that this could create bugs when the VM is called in other tasks and would like to review this logic more extensively.

hoh (approved):

PR #553, opened 2024-03-05
Update the diagnostic VM

hoh:
We published multiple changes to the diagnostic VM recently but none of these was released.

This provides a new diagnostic VM, based on a new runtime [1], with fixes:

- Reading messages with the newer SDK
- Better handling of IPv6 detection errors
- Two different tests for signing messages (local and remote)
- aleph-message version was not specified
- fetching a single message was not tested

### DIFF
--- a/examples/example_fastapi/main.py
+++ b/examples/example_fastapi/main.py
@@ -5,12 +5,19 @@ import os
 import socket
 import subprocess
 import sys
-from datetime import datetime
+from datetime import datetime, timezone
 from os import listdir
 from pathlib import Path
-from typing import List, Optional
+from typing import Any, Optional
 
 import aiohttp
+from aleph_message.models import (
+    MessagesResponse,
+    PostMessage,
+    ProgramMessage,
+    StoreMessage,
+)
+from aleph_message.status import MessageStatus
 from fastapi import FastAPI
 from fastapi.middleware.cors import CORSMiddleware
 from fastapi.responses import PlainTextResponse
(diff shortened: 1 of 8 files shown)


runtimes/aleph-debian-12-python/create_disk_image.sh
```diff
@@ -31,7 +31,7 @@ apt-get install -y --no-install-recommends --no-install-suggests \
 
 echo "Pip installing aleph-sdk-python"
 mkdir -p /opt/aleph/libs
-pip3 install --target /opt/aleph/libs 'aleph-sdk-python==0.7.0' 'fastapi~=0.103.1'
+pip3 install --target /opt/aleph/libs 'aleph-sdk-python==0.9.0' 'aleph-message==0.4.4' 'fastapi~=0.109.2'
```
nesitor, line 34: Better to not be so strict with version, use `~=` will be better than `==`

hoh: In the case of the runtimes, we want to know and be able to communicate the precise version of every file we ship.

examples/example_fastapi/main.py
```diff
@@ -159,9 +163,21 @@ async def read_internet():
             return {"result": resp.status, "headers": resp.headers}
 
 
-@app.get("/post_a_message")
-async def post_a_message():
-    """Post a message on the Aleph network"""
+@app.get("/get_a_message")
```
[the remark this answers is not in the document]
hoh: I don't understand this comment, can you reformulate ?

src/aleph/vm/orchestrator/views/__init__.py
```diff
@@ -199,6 +199,9 @@ async def status_check_fastapi(request: web.Request, vm_id: Optional[ItemHash] =
                 "index": await status.check_index(session, fastapi_vm_id),
                 "environ": await status.check_environ(session, fastapi_vm_id),
                 "messages": await status.check_messages(session, fastapi_vm_id),
+                # Using the remote account currently causes issues
+                # "post_a_message": await status.check_post_a_message(session, fastapi_vm_id),
+                # "sign_a_message": await status.check_sign_a_message(session, fastapi_vm_id),
```
nesitor, line 204: Why create that methods (`check_post_a_message` and `check_sign_a_message`) that aren't used finally on the fastapi endpoint?

hoh: Because I am afraid we still have the issue with the `RemoteAccount` not working as we had when running with pytest. I don't want to make that a blocking issue.

nesitor (approved):

PR #554, opened 2024-03-05
Fix issues with stopping VMs

hoh:
- Fix: A VM was only marked as stopping after all runs completed
- Fix: Stopping executions could be returned


hoh:
Only CodeCov tests are failing

nesitor (approved):

PR #558, opened 2024-03-06
Fix: Execution creation was not tested

hoh:
Solution: Add a test that creates a new VM execution and checks that it starts properly.


tests/supervisor/test_execution.py
```diff
@@ -0,0 +1,41 @@
...
+        snapshot_manager=None,
+        systemd_manager=None,
+        persistent=False,
+    )
+
+    # Downloading the resources required may take some time, limit it to 10 seconds
+    await asyncio.wait_for(execution.prepare(), timeout=30)
+
+    execution.create(vm_id=3, tap_interface=None)
+    # await asyncio.wait_for(execution.start(), timeout=30)
```
nesitor, line 41: I think that test the start method will be needed here also to ensure that covers the main use case.

hoh:
I managed to get the VM working, only one blocking issue left:

```python
# FIXME: This import fails to work in a VM when using pytest
from aleph.sdk.chains.remote import RemoteAccount
```

Antonyjin:
Is there a way to test this locally? I'm trying with `hatch run testing:test` but it doesn't work.

hoh:
I added doc to run the tests in `TESTING.md` https://github.com/aleph-im/aleph-vm/blob/hoh-test-create-vm/TESTING.md

TESTING.md
```diff
@@ -0,0 +1,63 @@
+# Testing aleph-vm
+
+This proedure describes how to run tests on a local system.
```
olethanh, line 3: procedure
```suggestion
This procedure describes how to run tests on a local system.
```

.github/workflows/test-on-droplets-matrix.yml
```diff
@@ -37,6 +37,32 @@ jobs:
         run: |
           sudo apt-get install libsystemd-dev cmake libdbus-1-dev libglib2.0-dev
 
+      - name: Download and build required files for running tests
+        run: |
```
olethanh, line 41: Lot of commands and magic  version numbers, there is a risk that we don't remember why this is needed or to  properly keep  this up to date, we should probably add more comments or link in the doc where these instructions come from.

tests/supervisor/test_execution.py
```diff
@@ -0,0 +1,57 @@
...
+
+from aleph.vm.conf import settings
+from aleph.vm.controllers.firecracker import AlephFirecrackerProgram
+from aleph.vm.models import VmExecution
+from aleph.vm.orchestrator import metrics
+from aleph.vm.storage import get_message
+
+
+@pytest.mark.asyncio
+async def test_create_execution():
```
olethanh, line 15: Add a warning that this will make change to the local system (like in TESTING.md)
Maybe also add a mark to no make execution optional since it need to be run as root?

tests/supervisor/test_execution.py
```diff
@@ -0,0 +1,57 @@
...
+
+    # Ensure that the settings are correct and required files present.
+    settings.setup()
+    settings.check()
+
+    # The database is required for the metrics and is currently not optional.
+    engine = metrics.setup_engine()
+    await metrics.create_tables(engine)
+
+    vm_hash = ItemHash("cafecafecafecafecafecafecafecafecafecafecafecafecafecafecafecafe")
```
olethanh, line 35: should take this value from the settings maybe?

tests/supervisor/test_execution.py
```diff
@@ -0,0 +1,57 @@
...
+    )
+
+    # Downloading the resources required may take some time, limit it to 10 seconds
+    await asyncio.wait_for(execution.prepare(), timeout=30)
+
+    vm = execution.create(vm_id=3, tap_interface=None)
+
+    # Test that the VM is created correctly. It is not started yet.
+    assert isinstance(vm, AlephFirecrackerProgram)
+    assert vm.vm_id == 3
```
olethanh, line 54: why 3?

olethanh (changes requested):
Add a few place were more documentation is needed. Otherwise look good to me, a part from the problem in the example that still fail

tests/supervisor/test_execution.py
hoh: Should the tests use temporary directories for the storage paths and set them in the settings as to not make change to the local system ?

Regarding tests running as root, while this test may not need it, I am afraid that many tests will require it for network configuration.

tests/supervisor/test_execution.py
hoh: This is a dummy hash that can be anything, I thought it would be quite explicit.

I would not put test ids in the global settings. What would you suggest ?

tests/supervisor/test_execution.py
hoh: To ensure that is is not the default from the settings. This should indeed be documented if considered as relevant.

TESTING.md
```diff
@@ -0,0 +1,63 @@
...
+
+```shell
+python3 -m venv /opt/venv
+source /opt/venv/bin/activate
+
+# Inside the venv
+pip install hatch
+```
+
+## 3. Initialize hatch for running the tests
```
olethanh, line 30: is there actually much value to use hatch? Especially if we have to patch it up like this.

tests/supervisor/test_execution.py
olethanh: My bad I had assumed it had to match the FAKE_INSTANCE_ID in the configuration

TESTING.md
hoh: Hatch is still quite handy as it ensures that the required dependencies are present.

I am open to suggestions for improvements.

.github/workflows/test-on-droplets-matrix.yml
hoh: These are copied from `packaging/Makefike`

olethanh (approved):
LGTM now

PR #568, opened 2024-03-06
Fix: Tests on Droplets ran even when tests failed

hoh:
Solution: Only run integration tests on Droplets if code quality and pytest succeed.


nesitor (approved):
Maybe the workflow is going a bit big, and we need to start thinking about how to improve it.

hoh:
> Maybe the workflow is going a bit big, and we need to start thinking about how to improve it.

Agreed, a refactoring would be useful.

PR #569, opened 2024-03-11
Fix: Depencency aleph-message was outdated

hoh:


nesitor (approved):
Maybe we should also publish the new release to the Github repo? 
Because the last one that appears is the 0.4.2.
https://github.com/aleph-im/aleph-message

hoh:
> Maybe we should also publish the new release to the Github repo? Because the last one that appears is the 0.4.2. https://github.com/aleph-im/aleph-message

Done

PR #571, opened 2024-03-12
Fix: Network errors were hard to inspect

hoh:
Context: We recently added the argument `check=True` to `subprocess.run(command)` to ensure we do not ignore errors in the setup or teardown of network interfaces.

Problem: Analyzing the errors that may happen during the setup or teardown of network interfaces is difficult due to the little amount of information provided by the `ip` command. It is therefore difficult to react accordingly.

Solution: Switch to use `pyroute2`, a pure Python netlink library that provides the required functions with more fined grained error reports.

The library `Pyroute2` is available in Debian 12 [1], however that version (0.7.2-2) is not documented anymore. It is absent from Debian 11. We therefore use the latest stable version from PyPI.

[1] https://packages.debian.org/bookworm/python3-pyroute2

### DIFF
--- a/.github/workflows/test-on-droplets-matrix.yml
+++ b/.github/workflows/test-on-droplets-matrix.yml
@@ -37,10 +37,12 @@ jobs:
         run: |
           sudo apt-get install libsystemd-dev cmake libdbus-1-dev libglib2.0-dev
 
+        # Unit tests create and delete network interfaces, and therefore require to run as root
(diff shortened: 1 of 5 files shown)


hoh:
Looks like `pytest` fails due to `'Operation not permitted'`. Looking into it ;-)

nesitor (approved):
LGTM

PR #572, opened 2024-03-13
Fix `FAKE_DATA_PROGRAM` issue on instances.

nesitor:
Fix: If the `FAKE_DATA_PROGRAM` variable have a default value, when we start the controllers, it takes that default value instead to avoid it, and it fails on the checks.

Solution: Put `FAKE_DATA_PROGRAM` variable to None by default.

### DIFF
--- a/src/aleph/vm/conf.py
+++ b/src/aleph/vm/conf.py
@@ -258,7 +258,7 @@ class Settings(BaseSettings):
 
     # Tests on programs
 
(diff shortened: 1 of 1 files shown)


nesitor:
> What about adding a boolean `USE_FAKE_DATA` instead ? Would that not make more sense ?

This will add another env variable to know, and we have lots.
I think that provide the `None` option by default makes more sense because we already specify that option as an argument when we want to use a fake program.

PR #573, opened 2024-03-13
Implement default hypervisor options on settings

nesitor:
Problem: The unique way to set the default hypervisor to use is hard-coded on the `models.py` class, so we can't select another one by default.

Solution: Implement `INSTANCE_DEFAULT_HYPERVISOR` field on the orchestrator configuration variables.

### DIFF
--- a/src/aleph/vm/conf.py
+++ b/src/aleph/vm/conf.py
@@ -361,9 +361,6 @@ class Settings(BaseSettings):
             assert is_command_available(
                 "qemu-system-x86_64"
             ), "Command `qemu-system-x86_64` not found, run `apt install qemu-system-x86`"
-        else:
-            # If QEmu is not supported, ignore the setting and use Firecracker by default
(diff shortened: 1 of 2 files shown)


src/aleph/vm/conf.py
```diff
@@ -356,6 +361,9 @@ def check(self):
             assert is_command_available(
                 "qemu-system-x86_64"
             ), "Command `qemu-system-x86_64` not found, run `apt install qemu-system-x86`"
+        else:
+            # If QEmu is not supported, ignore the setting and use Firecracker by default
+            settings.INSTANCE_DEFAULT_HYPERVISOR = HypervisorType.firecracker
```
hoh, line 366: Since this changes the settings, it should go in `setup()` instead of `check()`

hoh (commented):
One comment, the rest looks fine.

hoh (approved):

PR #576, opened 2024-03-14
Fix: Deleting a missing interface crashed

hoh:
Calling `delete_tap_interface` on an interface that is not present raised an error. This logs a debug message instead.

When the interface is not found for adding an IP address or setting a link up, a more explicit error is raised.

### DIFF
--- a/src/aleph/vm/network/interfaces.py
+++ b/src/aleph/vm/network/interfaces.py
@@ -64,7 +64,7 @@ def set_link_up(ipr: IPRoute, device_name: str):
     interface_index: list[int] = ipr.link_lookup(ifname=device_name)
     if not interface_index:
         raise MissingInterfaceError(f"Interface {device_name} does not exist, can't set it up.")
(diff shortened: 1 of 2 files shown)


src/aleph/vm/network/interfaces.py
```diff
@@ -52,11 +61,18 @@ def add_ip_address(ipr: IPRoute, device_name: str, ip: Union[IPv4Interface, IPv6
 
 def set_link_up(ipr: IPRoute, device_name: str):
     """Set the given interface up."""
+    interface_index: list[int] = ipr.link_lookup(ifname=device_name)
+    if not interface_index:
+        raise MissingInterfaceError(f"Interface {device_name} does not exist, can't set it up.")
     ipr.link("set", index=ipr.link_lookup(ifname=device_name)[0], state="up")
```
nesitor, line 67:
```suggestion
    ipr.link("set", index=interface_index[0], state="up")
```

hoh: Thanks !

nesitor (approved):

PR #577, opened 2024-03-18
Fix: Tests were run twice

hoh:
The command `hatch run testing:cov` includes running `hatch run testing:test-cov`.

This removes the duplication


Antonyjin (approved):
LGTM

PR #585, opened 2024-03-21
Fix: Unknown errors in interface configuration crashed the VM setup

hoh:
Solution: Log an error instead of raising an exception.

We'll need to investigate these errors further since this is not the expected behaviour.


nesitor:
This changes are already applied on this PR https://github.com/aleph-im/aleph-vm/pull/586

PR #586, opened 2024-03-21
Solve network issues on Ubuntu hosts

nesitor:
Problem: After the change of the new interface library to handle network interfaces, when you call 2 VMs at the same time, it assigns to you the same VM_ID and the network setup fails.

Solution: Implement semaphores on the execution creation to avoid creating VMs at the same time.
* Avoid raising exceptions on network issues and just log it as error.
* Added missing await async on VM request handling

### DIFF
--- a/src/aleph/vm/orchestrator/views/__init__.py
+++ b/src/aleph/vm/orchestrator/views/__init__.py
@@ -57,7 +57,7 @@ from packaging.version import InvalidVersion, Version
 logger = logging.getLogger(__name__)
 
 
-def run_code_from_path(request: web.Request) -> Awaitable[web.Response]:
+async def run_code_from_path(request: web.Request) -> web.Response:
     """Allow running an Aleph VM function from a URL path
 
     The path is expected to follow the scheme defined in `app.add_routes` below,
@@ -68,7 +68,7 @@ def run_code_from_path(request: web.Request) -> Awaitable[web.Response]:
 
     message_ref = ItemHash(request.match_info["ref"])
     pool: VmPool = request.app["vm_pool"]
-    return run_code_on_request(message_ref, path, pool, request)
+    return await run_code_on_request(message_ref, path, pool, request)
 
 
 async def run_code_from_hostname(request: web.Request) -> web.Response:
--- a/src/aleph/vm/pool.py
+++ b/src/aleph/vm/pool.py
@@ -3,7 +3,6 @@ from __future__ import annotations
 import asyncio
 import json
 import logging
-import threading
 from collections.abc import Iterable
 from datetime import datetime, timezone
 from typing import Optional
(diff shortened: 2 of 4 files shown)


src/aleph/vm/orchestrator/views/__init__.py
```diff
@@ -57,7 +57,7 @@
 logger = logging.getLogger(__name__)
 
 
-def run_code_from_path(request: web.Request) -> Awaitable[web.Response]:
+async def run_code_from_path(request: web.Request) -> web.Response:
```
hoh, line 60: What is the issue with this ?

src/aleph/vm/pool.py
```diff
@@ -87,6 +89,8 @@ async def create_a_vm(
     ) -> VmExecution:
         """Create a new Aleph Firecracker VM from an Aleph function message."""
 
+        await self.creation_semaphore.acquire()
```
hoh, line 92: Why use a semaphore and not a lock ?

src/aleph/vm/pool.py
```diff
@@ -130,9 +134,13 @@ async def create_a_vm(
             # ensure the VM is removed from the pool on creation error
             self.forget_vm(vm_hash)
             raise
+        finally:
+            self.creation_semaphore.release()
```
hoh, line 138: Why not use a context manager (`async with ...`) instead of a try... finally ?

src/aleph/vm/orchestrator/views/__init__.py
nesitor: It doesn't await for the returning response, so needs to await it

src/aleph/vm/pool.py
```diff
@@ -43,10 +44,12 @@ class VmPool:
     network: Optional[Network]
     snapshot_manager: Optional[SnapshotManager] = None
     systemd_manager: SystemDManager
+    creation_lock: threading.Lock
```
hoh, line 47:
```suggestion
    creation_lock: asyncio.Lock
```

src/aleph/vm/pool.py
```diff
@@ -87,53 +90,54 @@ async def create_a_vm(
...
-            execution = VmExecution(
-                vm_hash=vm_hash,
-                message=message,
-                original=original,
-                snapshot_manager=self.snapshot_manager,
-                systemd_manager=self.systemd_manager,
-                persistent=persistent,
-            )
-            self.executions[vm_hash] = execution
+        with self.creation_lock:
```
hoh, line 93:
```suggestion
        async with self.creation_lock:
```

hoh (changes requested):
Looks almost good, the only issue is that you used `threading.Lock` instead of `asyncio.Lock` while this is an asyncio application.

hoh (approved):

PR #588, opened 2024-04-04
Fix: Test launch a VM without building it locally

Antonyjin:
Deleting the line with mksquashfs which was building the runtimes locally.

### DIFF
--- a/.github/workflows/test-build-examples.yml
+++ b/.github/workflows/test-build-examples.yml
@@ -34,7 +34,6 @@ jobs:
           ls
           pwd
           pip3 install -t /opt/packages -r ./examples/example_pip/requirements.txt
-          mksquashfs /opt/packages packages.squashfs
 
 #      - run: |
 #          ipfs add packages.squashfs


tests/supervisor/test_execution_online.py
```diff
@@ -0,0 +1,48 @@
...
+from aleph.vm.models import VmExecution
+from aleph.vm.orchestrator import metrics
+from aleph.vm.storage import get_message
+
+
+@pytest.mark.asyncio
+async def test_create_execution():
+    """
+    Create a new VM execution and check that it starts properly.
+    """
```
hoh, line 17:
```suggestion
async def test_create_execution_from_network():
    """
    Create a new VM execution from the diagnostic VM and check that it starts properly.
    This will download the resources from the aleph.im network if they are not present locally.
    """
```

hoh (commented):
Looks great. Apart from my suggestion, what do you think about moving the test function in the same file `test_execution.py`, as the other test ?

tests/supervisor/test_execution_online.py
Antonyjin: Yes I like this suggestion, i'll make the changes!

PR #594, opened 2024-04-10
Fix: vm-connector used obsolete software

hoh:
The vm-connector service used to run with obsolete versions of the aleph-sdk and Python.

Solution: Use the latest the version of the SDK and pin the version of dependencies.


olethanh (approved):

olethanh:
we need test for this.

foxpatch-aleph (approved):
The PR successfully modernizes the vm-connector service by updating Python from 3.9 to 3.13, switching from the deprecated `aleph-client` to `aleph-sdk-python`, pinning all dependency versions, and adding new endpoints (`/compute/latest_amend`, `/publish/data`, `/properties`, `/sign`). The code is generally clean and the refactoring consolidates duplicate download logic. Minor issues include the use of `--reload` in the production Dockerfile, bare `assert` statements for validation in `compute_latest_amend`, and the SDK version being pinned to 1.4.0 rather than the latest available (2.3.4) despite the PR description promising the latest version.

**`docker/vm_connector.dockerfile`** (line 15): The `--reload` flag enables auto-reload on code changes which is intended for development only. Consider removing it from the production Docker image.

**`docker/vm_connector.dockerfile`** (line 8): Pinned `aleph-sdk-python==1.4.0` but the latest available version is 2.3.4. If 1.4.0 is intentionally chosen (e.g., API compatibility), the PR description should clarify this instead of saying "Use the latest the version of the SDK".

**`vm_connector/main.py`** (line 142): Using bare `assert` for message validation in `compute_latest_amend`. Asserts are stripped when Python runs with `-O`, potentially bypassing validation. Use proper conditional checks with exceptions instead.

**`vm_connector/main.py`** (line 29): `get_latest_message_amend` creates a new `aiohttp.ClientSession()` on every call, similar to other functions in this file. Consider using a shared session for better performance under load.

**`vm_connector/conf.py`** (line 12): Default API server changed to `https://official.aleph.cloud` without explanation. The previous default was `https://api2.aleph.im` which at least pointed to a well-known domain. This change may silently break deployments relying on the old default.

PR #595, opened 2024-04-10
Problem: dbus call were not async

olethanh:
dbus call to systemd were not called asyncronously and thus were hanging the whole thing

Modify all call to asynchronous, this required switching from
python-dbus to dbus-fast which offer an asyncio backend.

Another adventage of dbus-fast is that is in pure python so won't require libdbus or compilation (also it is packaged in ubuntu & debian )

### DIFF
--- a/packaging/Makefile
+++ b/packaging/Makefile
@@ -15,7 +15,7 @@ debian-package-code:
 	cp ../examples/instance_message_from_aleph.json ./aleph-vm/opt/aleph-vm/examples/instance_message_from_aleph.json
 	cp -r ../examples/data ./aleph-vm/opt/aleph-vm/examples/data
 	mkdir -p ./aleph-vm/opt/aleph-vm/examples/volumes
-	pip3 install --target ./aleph-vm/opt/aleph-vm/ 'aleph-message==0.4.4' 'jwskate==0.8.0' 'eth-account==0.9.0' 'sentry-sdk==1.31.0' 'qmp==1.1.0' 'superfluid==0.2.1' 'sqlalchemy[asyncio]' 'aiosqlite==0.19.0' 'alembic==1.13.1' 'aiohttp_cors==0.7.0' 'pyroute2==0.7.12'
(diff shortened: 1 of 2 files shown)


hoh (commented):
This seem to bring nice properties in addition to fixing the core issues.

Looks like you wrote a lot of Systemd specific boilerplate :thinking: 
Is there an easy way to push that upstream ?

packaging/Makefile
```diff
@@ -15,7 +15,7 @@ debian-package-code:
 	cp ../examples/instance_message_from_aleph.json ./aleph-vm/opt/aleph-vm/examples/instance_message_from_aleph.json
 	cp -r ../examples/data ./aleph-vm/opt/aleph-vm/examples/data
 	mkdir -p ./aleph-vm/opt/aleph-vm/examples/volumes
-	pip3 install --target ./aleph-vm/opt/aleph-vm/ 'aleph-message==0.4.4' 'jwskate==0.8.0' 'eth-account==0.9.0' 'sentry-sdk==1.31.0' 'qmp==1.1.0' 'superfluid==0.2.1' 'sqlalchemy[asyncio]' 'aiosqlite==0.19.0' 'alembic==1.13.1' 'aiohttp_cors==0.7.0' 'pyroute2==0.7.12'
+	pip3 install --target ./aleph-vm/opt/aleph-vm/ 'aleph-message==0.4.4' 'jwskate==0.8.0' 'eth-account==0.9.0' 'sentry-sdk==1.31.0' 'qmp==1.1.0' 'superfluid==0.2.1' 'sqlalchemy[asyncio]' 'aiosqlite==0.19.0' 'alembic==1.13.1' 'aiohttp_cors==0.7.0' 'pyroute2==0.7.12' 'dbus-fast==1.90.1'
```
hoh, line 18: Should we drop support for Debian 11 ?

olethanh:
> This seem to bring nice properties in addition to fixing the core issues.
> 
> Looks like you wrote a lot of Systemd specific boilerplate 🤔 Is there an easy way to push that upstream ?

Upstream is a generic lib for dbus and not fixated on any particular interface. I could move this in another python file if you prefer

packaging/Makefile
olethanh: probably but I wouldn't block the PR on it

hoh (approved):

PR #596, opened 2024-04-15
Problem: Vm execution failed due to network interface

olethanh:
occasionaly vm creation vm because the assigned tap network inteface was already existing. Probably not properly teared down from a previous execution or from a concurrency issue

Displayed Error was :
 OSError: [Errno 16] Device or resource busy

it was then looped as a retry and blocked the whole thing

Solution:
When assigning a vm_id, check that the network interface for that vm doesn't already exists This act as a double check for a variety of issues

### DIFF
--- a/src/aleph/vm/pool.py
+++ b/src/aleph/vm/pool.py
@@ -161,6 +161,12 @@ class VmPool:
             # anymore.
             currently_used_vm_ids = {execution.vm_id for execution in self.executions.values() if execution.is_running}
             for i in range(settings.START_ID_INDEX, 255**2):
+
+                if self.network:
+                    # Check the network interface don't already exists, otherwise it will cause a crash
+                    if self.network.interface_exists(i):
+                        continue
+
                 if i not in currently_used_vm_ids:
                     return i
             else:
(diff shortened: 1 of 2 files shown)


hoh:
Can we easily get this code tested ?

olethanh:
Not really automatically as it is at low level but if you want to test it manually you can create a network interface, prior to launching using

```
sudo ip tuntap add mode tap vmtap4
```

src/aleph/vm/pool.py
```diff
@@ -161,6 +161,12 @@ def get_unique_vm_id(self) -> int:
             # anymore.
             currently_used_vm_ids = {execution.vm_id for execution in self.executions.values() if execution.is_running}
             for i in range(settings.START_ID_INDEX, 255**2):
+
+                if self.network:
+                    # Check the network interface don't already exists, otherwise it will cause a crash
+                    if self.network.interface_exists(i):
+                        continue
```
hoh, line 168: Would it make more sense to delete the interface and then recreating it with the proper configuration ?

hoh (approved):

PR #598, opened 2024-04-15
Solve CORS issue after latest update

nesitor:
Fix: Solved CORS issues on PAYG creation.

### DIFF
--- a/src/aleph/vm/orchestrator/supervisor.py
+++ b/src/aleph/vm/orchestrator/supervisor.py
@@ -13,8 +13,8 @@ from pathlib import Path
 from secrets import token_urlsafe
 from typing import Callable
 
-import aiohttp_cors
 from aiohttp import web
+from aiohttp_cors import ResourceOptions, setup
 
 from aleph.vm.conf import settings
 from aleph.vm.pool import VmPool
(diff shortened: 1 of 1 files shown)


nesitor:
> Thanks for fixing this.
> 
> Can you explain what the changes from the "latest update" consist in and how they broke the CORS ? Did this happen when updating pinned versions of dependencies or after updating them ? Can we use automated tests to ensure that this does not happen anymore ?

Initially my feeling was that a CORS lib update (https://github.com/aio-libs/aiohttp-cors) caused that issue, because in a previous PR that I made it was working well. But after checking the library seems that any recent update was applied, so it means that other broke the CORS on the supervisor API, but I don't know exactly who or what.
The method to check that this don't happen again is to add a check to get the request headers and ensure that we always return `Access-Control-Allow-Origin` header on that API endpoints.

olethanh:
@nesitor Hello. Is there any documentation on how to test the PayG stuff for testing?
Which are then endpoints called by the PAYG stuff? Who is calling them? The ccn?

nesitor:
> @nesitor Hello. Is there any documentation on how to test the PayG stuff for testing? Which are then endpoints called by the PAYG stuff? Who is calling them? The ccn?

Hey, there aren't yet, but I think that these questions are out of the scope of that PR.

hoh (approved):

PR #600, opened 2024-04-16
Problem: Makefile for publishing example were not working

olethanh:
'aleph program' now need an 'update' argument.
Solution: Update makefile and documentation

### DIFF
--- a/examples/example_http_js/Makefile
+++ b/examples/example_http_js/Makefile
@@ -16,4 +16,4 @@ docker-publish:
 
 publish:
 	chmod +x ./src/run.sh
-	aleph program ./src "run.sh"
+	aleph program upload ./src "run.sh"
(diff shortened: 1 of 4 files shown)


examples/program_message_from_aleph.json
```diff
@@ -14,7 +14,7 @@
         "allow_amend": false,
         "code": {
             "encoding": "squashfs",
-            "entrypoint": "main:app",
+            "entrypoint": "run.sh",
```
nesitor, line 17: Modify this endpoint on that file is generic for all the examples. So maybe instead modify it, should be better to create a new message example file with needed endpoint.

olethanh: sorry it was added by error, will remove it

hoh (approved):

PR #606, opened 2024-04-26
Fix: Unkwnown hashes raised exception

hoh:
Problem: Many crawlers called URLs that do not exist on CRNs.

The current implementation raises an error when the hash of the VM cannot be found, which fills the logs on Sentry.

Solution: Return an HTTP Not Found status instead.

### DIFF
--- a/src/aleph/vm/orchestrator/views/__init__.py
+++ b/src/aleph/vm/orchestrator/views/__init__.py
@@ -11,7 +11,7 @@ from typing import Optional
 import aiodns
 import aiohttp
 from aiohttp import web
-from aiohttp.web_exceptions import HTTPNotFound
+from aiohttp.web_exceptions import HTTPBadRequest, HTTPNotFound
 from aleph_message.exceptions import UnknownHashError
 from aleph_message.models import ItemHash, MessageType
 from pydantic import ValidationError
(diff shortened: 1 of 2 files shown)


nesitor (approved):

src/aleph/vm/orchestrator/views/__init__.py
```diff
@@ -98,8 +101,10 @@ async def run_code_from_hostname(request: web.Request) -> web.Response:
             try:
                 message_ref = ItemHash(await get_ref_from_dns(domain=f"_aleph-id.{request.host}"))
                 logger.debug(f"Using DNS TXT record to obtain '{message_ref}'")
-            except aiodns.error.DNSError as error:
-                raise HTTPNotFound(reason="Invalid message reference") from error
+            except aiodns.error.DNSError:
+                return HTTPNotFound(reason="Invalid message reference")
+            except UnknownHashError:
+                return HTTPNotFound(reason="Invalid message reference")
```
nesitor, line 107: Maybe can we add a test to check that Not Found status?

hoh: Absolutely :+1:

hoh:
Hi @nesitor , I added tests and requested a new review ;-)

tests/supervisor/views/test_run_code.py
```diff
@@ -0,0 +1,46 @@
...
+        "/vm/" + item_hash,
+        match_info={
+            "ref": item_hash,
+            "suffix": "/some/suffix",
+        },
+        headers={"Host": settings.DOMAIN_NAME},
+        app=app,
+    )
+    with pytest.raises(HTTPBadRequest):
+        await run_code_from_path(invalid_hash_request)
```
nesitor, line 38: I just checked the test, and it isn't checking the condition itself to raise the `HTTPBadRequest`, because you are mocking the `aiohttp` request to always raise the error. Instead, I think that will be better to just put an invalid `ItemHash` to ensure the try/catch block receiving the request.

hoh: I do not understand your comment. Can you provide an example ?

nesitor: In my comment I already tell to you an example about what will be better to do. 

> Instead, I think that will be better to just put an invalid ItemHash to ensure the try/catch block receiving the request.

PR #608, opened 2024-04-26
Make DBus calls Async (2nd attempt)

hoh:
Follow up on https://github.com/aleph-im/aleph-vm/pull/595/files

- Problem: dbus call were not async
- use mypy procotol method for typing
- fix CI, specify bus_type
-  fix init in async, isort
- mypy
- dbus fast is not in debian 11


hoh:
@olethanh should we drop support for Python 3.9 / Debian 11 for this PR ?

olethanh:
> @olethanh should we drop support for Python 3.9 / Debian 11 for this PR ?

Didn't we already have this discussion ? I'm not against it but I have no idea of what the processus or the consequence for this is :D

Anyway I have fixed the problem

olethanh:
Pr is ready for review. Updated with main

Note that there is a chance of behavior: Now in is_running for persistent vms we only check the stop time like we do for other VMs, to stay consistent. The behaviour is load_persistent_executions has been properly adapted

src/aleph/vm/pool.py
```diff
@@ -239,8 +240,10 @@ async def load_persistent_executions(self):
                 persistent=saved_execution.persistent,
             )
 
-            if execution.is_running:
-                # TODO: Improve the way that we re-create running execution
+            if await self.systemd_manager.is_service_active(
+                execution.controller_service
+            ):  # TODO: Improve the way that we re-create running execution
```
nesitor, line 245: IMO having different ways to check if an instance is running here or at the method used on the `VMExecution` class, can provide some issues in the future, because we don't follow always the same pattern to check it.

src/aleph/vm/systemd.py
```diff
@@ -2,75 +2,231 @@
...
+
+
+@runtime_checkable
+class UnitProxy(Protocol):
+    """for typing.
+
+    for description of methods see
+    https://www.freedesktop.org/software/systemd/man/latest/org.freedesktop.systemd1.html#Service%20Unit%20Objects"""
+
+    async def get_active_state(self) -> ActiveStateLiteral: ...
```
nesitor, line 137: I think that all these classes should be on another file, not here, because are part of the way of using the `dbus_fast` library.

nesitor:
Also, will be good to have some tests about these refactor to prevent implementation issues.

olethanh:
> Also, will be good to have some tests about these refactor to prevent implementation issues.

on which part?

src/aleph/vm/pool.py
nesitor: You are correct, but IMO having the same way to check if a VM is running independently of the method used (systemd or direct) it's better that rehydrate fields and trust on that rehydration.

nesitor:
> > Also, will be good to have some tests about these refactor to prevent implementation issues.
> 
> on which part?

On the whole `Systemd` class and on the main methods of the application that calls the systemd start, enable, stop and disable methods.

PR #609, opened 2024-04-29
Problem /about/usage/system was not tested

olethanh:
Solution: Add a test in pytest and a  test in droplet

I had to refactor the setup of the web.app test since pytest-aiohttp use a new async loop between each tests.
Also fixed an issue in `get_machine_properties` where the returned format is not the same between all the machines. note that this code/dep  is going to be dropped in #544 so that's not a blocking issue

### DIFF
--- a/.github/workflows/test-on-droplets-matrix.yml
+++ b/.github/workflows/test-on-droplets-matrix.yml
@@ -238,6 +238,14 @@ jobs:
             -d '{"persistent_vms": [], "instances": ["${{ matrix.check_vm.item_hash }}"]}' \
             "http://${DROPLET_IPV4}:4020/control/allocations"
 
+      - name: Get system usage
+        run: |
+          export DROPLET_IPV4="$(doctl compute droplet get aleph-vm-ci-${{ matrix.os_config.alias }}-${{ matrix.check_vm.alias }} --output json | ./.github/scripts/extract_droplet_ipv4.py)"
+          curl -X GET -H "Content-Type: application/json" \
+            -H "X-Auth-Signature: test" \
+            "http://${DROPLET_IPV4}:4020/about/usage/system"
+
+
       - name: Export aleph logs
         if: always()
         run: |
--- a/.github/workflows/test-on-droplets-matrix.yml
+++ b/.github/workflows/test-on-droplets-matrix.yml
@@ -238,11 +238,10 @@ jobs:
             -d '{"persistent_vms": [], "instances": ["${{ matrix.check_vm.item_hash }}"]}' \
             "http://${DROPLET_IPV4}:4020/control/allocations"
 
-      - name: Get system usage
+      - name: Fetch system usage endpoint
         run: |
           export DROPLET_IPV4="$(doctl compute droplet get aleph-vm-ci-${{ matrix.os_config.alias }}-${{ matrix.check_vm.alias }} --output json | ./.github/scripts/extract_droplet_ipv4.py)"
           curl -X GET -H "Content-Type: application/json" \
-            -H "X-Auth-Signature: test" \
             "http://${DROPLET_IPV4}:4020/about/usage/system"
 
 
--- a/src/aleph/vm/orchestrator/resources.py
+++ b/src/aleph/vm/orchestrator/resources.py
@@ -87,8 +87,8 @@ def get_machine_properties() -> MachineProperties:
     cpu_info = cpuinfo.get_cpu_info()  # Slow
     return MachineProperties(
         cpu=CpuProperties(
-            architecture=cpu_info["raw_arch_string"],
-            vendor=cpu_info["vendor_id"],
+            architecture=cpu_info.get("raw_arch_string", cpu_info.get("arch_string_raw")),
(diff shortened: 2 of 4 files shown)


.github/workflows/test-on-droplets-matrix.yml
```diff
@@ -238,6 +238,13 @@ jobs:
             -d '{"persistent_vms": [], "instances": ["${{ matrix.check_vm.item_hash }}"]}' \
             "http://${DROPLET_IPV4}:4020/control/allocations"
 
+      - name: Fetch system usage endpoint
+        run: |
+          export DROPLET_IPV4="$(doctl compute droplet get aleph-vm-ci-${{ matrix.os_config.alias }}-${{ matrix.check_vm.alias }} --output json | ./.github/scripts/extract_droplet_ipv4.py)"
+          curl -X GET -H "Content-Type: application/json" \
+            "http://${DROPLET_IPV4}:4020/about/usage/system"
+
+
```
nesitor, line 247: I think that adding that check on the CI without check the results, doesn't provide so much inside about if it's working well or not. I think that having a unit test for it with some results covered is enough to check if it's working well.

tests/supervisor/test_views.py
```diff
@@ -24,3 +25,16 @@ async def test_allocation_fails_on_invalid_item_hash(aiohttp_client):
...
+async def test_system_usage(aiohttp_client):
+    """Test that the usage system endpoints responds. No auth needed"""
+    app = setup_webapp()
+    client = await aiohttp_client(app)
+    response: web.Response = await client.get("/about/usage/system")
+    assert response.status == 200
+    # check if it is valid json
+    resp = await response.json()
+    assert "cpu" in resp
+    assert resp["cpu"]["count"] > 0
```
nesitor, line 40: To make the test reliable on the future, maybe we can mock the system functions that returns the CPU specs and test here the options to have a match of the same results.

nesitor:
I think that on Alie's PR (https://github.com/aleph-im/aleph-vm/pull/544) he had installed also the needed packages needed on the installation step.
 I don't know if that PR merges also his changes, but if it does, please, can you also add it to this PR? And if not, can you link as dependent this PR to the old one to merge both at the same time?

olethanh:
> I think that on Alie's PR (#544) he had installed also the needed packages needed on the installation step. I don't know if that PR merges also his changes, but if it does, please, can you also add it to this PR? And if not, can you link as dependent this PR to the old one to merge both at the same time?

Alie's  PR is not included into this one, in that other PR the deps have changed but here they stay the same so normally the installation step shouldn't change, unless I have missed something?

.github/workflows/test-on-droplets-matrix.yml
olethanh: It check  that the endpoint doesn't crash in "real" conditions, which for me is important for the integration part of test suite and is complementary with the unit test, especially if we are going to mock the lib in the unit test.
I agree ideally  we should check the content here also but at the moment we can already check visually when there is a problem the whole droplet things need more automation

src/aleph/vm/orchestrator/resources.py
```diff
@@ -87,8 +87,8 @@ def get_machine_properties() -> MachineProperties:
     cpu_info = cpuinfo.get_cpu_info()  # Slow
     return MachineProperties(
         cpu=CpuProperties(
-            architecture=cpu_info["raw_arch_string"],
-            vendor=cpu_info["vendor_id"],
+            architecture=cpu_info.get("raw_arch_string", cpu_info.get("arch_string_raw")),
+            vendor=cpu_info.get("vendor_id", cpu_info.get("vendor_id_raw")),
```
nesitor, line 91: Sorry, I got confused by that part, but seems that was just a typo.

```suggestion
            architecture=cpu_info.get("arch_string_raw"),
            vendor=cpu_info.get("vendor_id_raw"),
```

.github/workflows/test-on-droplets-matrix.yml
nesitor: This endpoint seems to not be so relevant, so just check that returns "something" and don't crash isn't so useful from my point of view, and also it adds more usage time on the Digital Ocean droplets. I think that we need to agree with the best way to execute the CI tests and with kind of tests.

src/aleph/vm/orchestrator/resources.py
olethanh: not a typo it's a fallback

.github/workflows/test-on-droplets-matrix.yml
olethanh: I had send a reply by e-mail but it seems it didn't pass. 

> I think that we need to agree with the best way to execute the CI tests and with kind of tests.

I'm guessing you meant 'which' and not 'with' ? For me unit tests and integration tests are complementary.
I understand we want to reign in on the DO cost but we are going to do more comprehensive testing on them in the future so we should look at other way  to reduce the cost. But if you feel that strongly about it I can remove it from this PR

src/aleph/vm/orchestrator/resources.py
nesitor: Okay

nesitor (approved):

PR #611, opened 2024-05-02
Fix: Backquote in shell script executed command

hoh:
When executed using `bash`, the `create_disk_image` was interrupted by a Python REPL due to the `python -OO` command being surrounded by backquotes.


olethanh (approved):

olethanh:
@hoh is it like a fish only error? Because I don't have the issue with Debian's bash

nesitor (approved):

PR #612, opened 2024-05-02
Fix: Branch `main` could not be tested easily

hoh:
This deploys the main branch automatically on the staging servers for system testing.

### DIFF
--- a/.github/workflows/deploy-main-on-staging.yml
+++ b/.github/workflows/deploy-main-on-staging.yml
@@ -1,0 +1,62 @@
+# This workflow automatically deploys main on staging
+name: "Deploy `main` automatically on staging"
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  deploy_staging_servers:
+    name: "Deploying on ${{ matrix.staging_servers.hostname }}"
+    runs-on: ubuntu-latest
+    strategy:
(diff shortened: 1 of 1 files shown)


olethanh:
both server are on debian 12? Wouldn't it make sense to have one on ubuntu?

olethanh (approved):
LGTM.
I trust you have tested it somehow

.github/workflows/deploy-main-on-staging.yml
```diff
@@ -0,0 +1,62 @@
+# This workflow automatically deploys main on staging
+name: "Deploy `main` automatically on staging"
+
+on:
+  push:
+    branches:
+      - main
```
nesitor, line 7: Maybe we can do it using a tag instead automatically? Just to have better control on it.

hoh:
> both server are on debian 12? Wouldn't it make sense to have one on ubuntu?

Yes, that would make sense. However, would it then be difficult to know if a bug comes from the OS or from the hosting provider :thinking: ?

.github/workflows/deploy-main-on-staging.yml
hoh: I think automatically would make things easier: we'll always know that the latest version runs there.

PR #614, opened 2024-05-03
Fix: System testing on DO took too many resources

hoh:
Problem: Running system testing on DigitalOcean for every push consumed a lot of resources and failed frequently.

We now start to have integration testing using `pytest`, which provides a better confidence that things actually work.

Solution: Only test on DO open Pull Requests and not every push. In the future, consider only running when merged on `main`.


MHHukiewitz (approved):
LGTM, simple restructuring

PR #615, opened 2024-05-03
Added Qemu automatic tests

nesitor:
Feature: Added automatic tests to check if Qemu runs, also with network interface.
The tests bypass systemd manager to reduce complexity.
In later tests we will test the entire flow also using systemd.

### DIFF
--- a/src/aleph/vm/conf.py
+++ b/src/aleph/vm/conf.py
@@ -281,13 +281,15 @@ class Settings(BaseSettings):
     )
 
     USE_FAKE_INSTANCE_BASE = False
-    FAKE_INSTANCE_BASE = Path(abspath(join(__file__, "../../runtimes/instance-debian-rootfs/rootfs.ext4")))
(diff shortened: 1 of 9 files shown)


src/aleph/vm/conf.py
```diff
@@ -281,13 +281,15 @@ class Settings(BaseSettings):
     )
 
     USE_FAKE_INSTANCE_BASE = False
-    FAKE_INSTANCE_BASE = Path(abspath(join(__file__, "../../runtimes/instance-debian-rootfs/rootfs.ext4")))
+    FAKE_INSTANCE_BASE = Path(abspath(join(__file__, "../../runtimes/instance-rootfs/rootfs.ext4")))
```
Antonyjin, line 284: In my case, I'm using FAKE_INSTANCE_BASE
and I think the path is wrong, so I had to change it to : 
`FAKE_INSTANCE_BASE = Path(abspath(join(__file__, "../../../../runtimes/instance-rootfs/debian-12.btrfs")))`

I don't know if it's relevant to change it in this PR because in your test you override the path. I'll certainly do it in a future PR from a branch that will take the changes from this PR.
I just wanted to note it now in case you want to change it immediately.

PR #616, opened 2024-05-07
601 creating instance tests

Antonyjin:
Feature: Added tests to check if Firecracker instance runs with network interface.
The tests bypass systemd manager

### DIFF
--- a/.github/workflows/test-using-pytest.yml
+++ b/.github/workflows/test-using-pytest.yml
@@ -41,7 +41,7 @@ jobs:
         run: |
           sudo mkdir --parents /opt/firecracker/
           sudo curl -fsSL -o "/opt/firecracker/vmlinux.bin" "https://ipfs.aleph.cloud/ipfs/bafybeiaj2lf6g573jiulzacvkyw4zzav7dwbo5qbeiohoduopwxs2c6vvy"
-          
+
           rm -fr /tmp/firecracker-release
           mkdir --parents /tmp/firecracker-release /opt/firecracker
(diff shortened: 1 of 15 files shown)


.github/workflows/test-using-pytest.yml
```diff
@@ -50,20 +50,26 @@ jobs:
...
+          cd runtimes/instance-rootfs && sudo ./create-debian-12-disk.sh && cd ../..
 
       - name: "Build example volume"
         run: |
           cd examples/volumes && bash build_squashfs.sh
 
-        # Unit tests create and delete network interfaces, and therefore require to run as root
+      - name: Create user jailman
+        run: |
+          sudo useradd jailman
```
nesitor, line 70: I think that this line can be added at final in a previous step, like in the `Download and build required files for running tests. Copied from packaging/Makefile.` that install jailer.

src/aleph/vm/models.py
```diff
@@ -81,14 +79,17 @@ class VmExecution:
...
 
     @property
     def is_running(self) -> bool:
         return (
-            bool(self.times.starting_at and not self.times.stopping_at)
-            if not self.persistent
-            else self.systemd_manager.is_service_active(self.controller_service)
+            self.systemd_manager.is_service_active(self.controller_service)
+            if self.persistent and self.systemd_manager
+            else bool(self.times.starting_at and not self.times.stopping_at)
```
nesitor, line 92: This has been changed on main, so shouldn't be here. Be careful the next time with the merges.

nesitor (approved):

PR #617, opened 2024-05-16
Added `USE_CONFIDENTIAL_COMPUTING` check

nesitor:
Problem: A node operator cannot check or add support for confidential computing.

Solution: Implemented a setting to allow node operator to enable confidential computing. A check ensure that the system is well configured, and it shows that configuration on the /public/config endpoint.

Implementations:
- Added `USE_CONFIDENTIAL_COMPUTING` setting with a `False` default value. If the user enables it, it checks the module availability on the system.
- Added a new computing section on `/public/config` endpoint with relevant information about computing.
- Added `PAYMENT_MONITOR_INTERVAL` field on payment section on `/public/config` endpoint to show the payment interval check.

### DIFF
--- a/src/aleph/vm/conf.py
+++ b/src/aleph/vm/conf.py
@@ -16,7 +16,7 @@ from pydantic import BaseSettings, Field, HttpUrl
 from pydantic.env_settings import DotenvType, env_file_sentinel
 from pydantic.typing import StrPath
 
-from aleph.vm.utils import file_hashes_differ, is_command_available
+from aleph.vm.utils import check_system_module, file_hashes_differ, is_command_available
 
 logger = logging.getLogger(__name__)
 
@@ -261,6 +261,12 @@ class Settings(BaseSettings):
         description="Default hypervisor to use on running instances, can be Firecracker or QEmu",
     )
 
+    USE_CONFIDENTIAL_COMPUTING: bool = Field(
+        default=False,
+        description="Enable Confidential Computing using AMD-SEV. It will test if the host is compatible "
+        "with SEV and SEV-ES",
+    )
+
     # Tests on programs
 
     FAKE_DATA_PROGRAM: Optional[Path] = None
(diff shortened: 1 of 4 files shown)


olethanh:
I suggest we change the name to `ENABLE_CONFIDENTIAL_COMPUTING` or `ENABLE_CONFIDENTIAL_SUPPORT`, because "use" for me imply that it will be used by default, while  "enable" tell to make it available.

For example `USE_JAILER` : mean jailer is used for all program, the user launching a program cannot choose when it's used or not.
But `ENABLE_QEMU_SUPPORT` make the backend available for users

src/aleph/vm/conf.py
```diff
@@ -336,6 +342,12 @@ def check(self):
                 int(ipv4_pool_length) <= settings.IPV4_NETWORK_PREFIX_LENGTH
             ), "The IPv4 address pool prefix must be shorter than an individual VM network prefix"
 
+        if self.USE_CONFIDENTIAL_COMPUTING:
```
olethanh, line 345: confidential require qemu so we should check that the qemu support is also properly enabled

src/aleph/vm/utils.py
```diff
@@ -130,6 +130,14 @@ def is_command_available(command):
         return False
 
 
+def check_system_module(module_path) -> str:
+    try:
+        output = subprocess.check_output(["cat", "/sys/module", module_path], stderr=subprocess.STDOUT)
+        return str(output)
+    except subprocess.CalledProcessError:
+        return ""
```
olethanh, line 138: You don't need to use cat you can just open() and then read() it in python like a regular file

```suggestion
from pathlib import Path

def check_system_module(module_path) -> bool:
    p = Path('/sys/module/') / module_path
    if not p.exists():
        return False
    content = p.open().read().strip()
    return content == 'Y'
```

Note this return directly a bool and not a str anymore

olethanh (changes requested):

olethanh (approved):
sorry I thought I had approved with my last comment

PR #618, opened 2024-05-28
Add platform confidential directory on Settings

nesitor:
Problem: The server doesn't have a directory to save the platform certificates generated by `sevctl`.

Solution: Set that directory field on settings class and ensure to create the folder on initialization step.

### DIFF
--- a/src/aleph/vm/conf.py
+++ b/src/aleph/vm/conf.py
@@ -267,6 +267,11 @@ class Settings(BaseSettings):
         "with SEV and SEV-ES",
     )
 
+    CONFIDENTIAL_DIRECTORY: Path = Field(
+        None,
+        description="Confidential Computing default directory. Default to EXECUTION_ROOT/confidential",
+    )
+
     # Tests on programs
 
     FAKE_DATA_PROGRAM: Optional[Path] = None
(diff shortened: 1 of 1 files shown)


olethanh:
I'm not actually sure we would need a decitated dir for the certificates, Platform certificate we can just generate in a temporary file or in the main aleph dir (there is only one file now) and certificate we get for the VM should go in the VM directory.

nesitor:
> I'm not actually sure we would need a decitated dir for the certificates, Platform certificate we can just generate in a temporary file or in the main aleph dir (there is only one file now) and certificate we get for the VM should go in the VM directory.

I think that is better to have it cached, to avoid request it all the time

olethanh (approved):

hoh:
> I think that is better to have it cached, to avoid request it all the time

If this data is cached, should it be in `CACHE_ROOT` instead ?

```python
        if not self.CONFIDENTIAL_DIRECTORY:
            self.CONFIDENTIAL_DIRECTORY = self.CACHE_ROOT / "confidential"
```

hoh:
`/var/cache` contains data that may be deleted (ex: can be downloaded again or recreated) while `/var/lib` contains data that may not be lost.

PR #619, opened 2024-05-28
Implement get platform certificates endpoint

nesitor:
Problem: There isn't an endpoint to be able to get the confidential platform certificates to start the VM key exchange.

Solution: Create that endpoint and return the platform certificates generated by the `sevctl` command.

### DIFF
--- a/src/aleph/vm/conf.py
+++ b/src/aleph/vm/conf.py
@@ -267,6 +267,11 @@ class Settings(BaseSettings):
         "with SEV and SEV-ES",
     )
 
+    CONFIDENTIAL_DIRECTORY: Path = Field(
+        None,
+        description="Confidential Computing default directory. Default to EXECUTION_ROOT/confidential",
+    )
+
     # Tests on programs
 
     FAKE_DATA_PROGRAM: Optional[Path] = None
@@ -409,6 +414,7 @@ class Settings(BaseSettings):
 
         os.makedirs(self.EXECUTION_LOG_DIRECTORY, exist_ok=True)
         os.makedirs(self.PERSISTENT_VOLUMES_DIR, exist_ok=True)
+        os.makedirs(self.CONFIDENTIAL_DIRECTORY, exist_ok=True)
 
         self.API_SERVER = self.API_SERVER.rstrip("/")
 
@@ -467,6 +473,8 @@ class Settings(BaseSettings):
             self.EXECUTION_LOG_DIRECTORY = self.EXECUTION_ROOT / "executions"
         if not self.JAILER_BASE_DIR:
             self.JAILER_BASE_DIR = self.EXECUTION_ROOT / "jailer"
+        if not self.CONFIDENTIAL_DIRECTORY:
+            self.CONFIDENTIAL_DIRECTORY = self.EXECUTION_ROOT / "confidential"
 
     class Config:
         env_prefix = "ALEPH_VM_"
--- a/src/aleph/vm/conf.py
+++ b/src/aleph/vm/conf.py
@@ -387,6 +387,7 @@ class Settings(BaseSettings):
             assert (
                 check_system_module("kvm_amd/parameters/sev_es") == "Y"
             ), "SEV-ES feature isn't enabled, enable it in BIOS"
+            assert is_command_available("sevctl"), "Command `sevctl` not found, run `cargo install sevctl`"
 
             assert self.ENABLE_QEMU_SUPPORT, "Qemu Support is needed for confidential computing and it's disabled, "
             "enable it setting the env variable `ENABLE_QEMU_SUPPORT=True` in configuration"
--- a/src/aleph/vm/orchestrator/resources.py
+++ b/src/aleph/vm/orchestrator/resources.py
@@ -11,6 +11,7 @@ from aleph_message.models.execution.environment import CpuProperties
 from pydantic import BaseModel, Field
 
 from aleph.vm.conf import settings
+from aleph.vm.sevclient import SevClient
 from aleph.vm.utils import cors_allow_all
 
 
@@ -129,10 +130,10 @@ async def about_certificates(request: web.Request):
     if not settings.ENABLE_CONFIDENTIAL_COMPUTING:
         return web.HTTPBadRequest(reason="Confidential computing setting not enabled on that server")
 
-    sev_client = request.app["sev_client"]
+    sev_client: SevClient = request.app["sev_client"]
 
     if not sev_client.certificates_archive.is_file():
-        sev_client.export_certificates()
+        await sev_client.export_certificates()
 
     return web.FileResponse(sev_client.certificates_archive)
 
--- a/src/aleph/vm/orchestrator/resources.py
+++ b/src/aleph/vm/orchestrator/resources.py
@@ -122,6 +122,21 @@ async def about_system_usage(_: web.Request):
     return web.json_response(text=usage.json(exclude_none=True))
 
 
+@cors_allow_all
+async def about_certificates(request: web.Request):
+    """Public endpoint to expose platform certificates for confidential computing."""
+
+    if not settings.ENABLE_CONFIDENTIAL_COMPUTING:
+        return web.HTTPBadRequest(reason="Confidential computing setting not enabled on that server")
+
+    sev_client = request.app["sev_client"]
+
+    if not sev_client.certificates_archive.is_file():
+        sev_client.export_certificates()
+
+    return web.FileResponse(sev_client.certificates_archive)
+
+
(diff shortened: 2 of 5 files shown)


src/aleph/vm/conf.py
```diff
@@ -387,6 +387,7 @@ def check(self):
             assert (
                 check_system_module("kvm_amd/parameters/sev_es") == "Y"
             ), "SEV-ES feature isn't enabled, enable it in BIOS"
+            assert is_command_available("sevctl"), "Command `sevctl` not found, run `cargo install sevctl`"
```
hoh, line 390: Should this command be installed globally on the system or is it a standalone binary whose path should be in the settings ?

src/aleph/vm/orchestrator/resources.py
```diff
@@ -122,6 +122,21 @@ async def about_system_usage(_: web.Request):
...
 
 
+@cors_allow_all
+async def about_certificates(request: web.Request):
+    """Public endpoint to expose platform certificates for confidential computing."""
+
+    if not settings.ENABLE_CONFIDENTIAL_COMPUTING:
+        return web.HTTPBadRequest(reason="Confidential computing setting not enabled on that server")
+
+    sev_client = request.app["sev_client"]
```
hoh, line 132:
```suggestion
    sev_client: SevClient = request.app["sev_client"]
```

src/aleph/vm/sevclient.py
```diff
@@ -0,0 +1,22 @@
...
+
+class SevClient:
+    def __init__(self, sev_dir: Path):
+        self.sev_dir = sev_dir
+        self.certificates_dir = sev_dir / "platform"
+        self.certificates_dir.mkdir(exist_ok=True, parents=True)
+        self.certificates_archive = self.certificates_dir / "certs_export.cert"
+
+    def sevctl_cmd(self, *args) -> subprocess.CompletedProcess:
+        result = subprocess.run(
```
hoh, line 13: Using the `subprocess` module is synchronous and would block the entire event loop.

[asyncio.subprocess](https://docs.python.org/3/library/asyncio-subprocess.html#asyncio-subprocess) should be used instead ;-)

src/aleph/vm/conf.py
nesitor: It is supposed to be already on the system, as we commented on the sync meeting, we will create a binary to be downloaded and installed directly on the system, and also on the client side.

hoh: In this case, we should add a setting `SEVCTL_PATH` similar to `FIRECRACKER_PATH` for binaries that are provided by our package.

hoh: I opened a Pull Request to build and bundle the `sevctl` program as part of the Debian packages.
See https://github.com/aleph-im/aleph-vm/pull/629

This is not blocking and can be marked as resolved.

src/aleph/vm/sevclient.py
```diff
@@ -0,0 +1,22 @@
...
+        self.certificates_dir.mkdir(exist_ok=True, parents=True)
+        self.certificates_archive = self.certificates_dir / "certs_export.cert"
+
+    async def sevctl_cmd(self, *args) -> bytes:
+        result = await run_in_subprocess(
+            ["sevctl", *args],
+            check=True,
+        )
+
+        return result
```
hoh, line 19: Method 'sevctl_cmd' may be 'static' . It would be better to make this method a separate function since it does not depend on the class.

hoh, line 19: This code can be simplified.

```suggestion
    async def sevctl_cmd(self, *args) -> bytes:
        return await run_in_subprocess(
            ["sevctl", *args],
            check=True,
        )
```

src/aleph/vm/sevclient.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+class SevClient:
+    def __init__(self, sev_dir: Path):
+        self.sev_dir = sev_dir
+        self.certificates_dir = sev_dir / "platform"
+        self.certificates_dir.mkdir(exist_ok=True, parents=True)
+        self.certificates_archive = self.certificates_dir / "certs_export.cert"
+
+    async def sevctl_cmd(self, *args) -> bytes:
```
hoh, line 13: This method is missing a docstring to explain what it does and why.

tests/supervisor/test_views.py
```diff
@@ -121,3 +126,50 @@ def get_persistent_executions(self):
...
+    with mock.patch(
+        "pathlib.Path.is_file",
+        return_value=False,
+    ) as is_file_mock:
+        with mock.patch(
+            "aleph.vm.sevclient.run_in_subprocess",
+            return_value=True,
+        ) as export_mock:
+            app = setup_webapp()
+            sev_client = SevClient(settings.CONFIDENTIAL_DIRECTORY)
```
hoh, line 162: Why pass a setting and not directly a temporary directory ?

tests/supervisor/test_views.py
```diff
@@ -121,3 +126,50 @@ def get_persistent_executions(self):
...
+    assert await response.text() == "400: Confidential computing setting not enabled on that server"
+
+
+@pytest.mark.asyncio
+async def test_about_certificates(aiohttp_client):
+    """Test that the certificates system endpoint responds. No auth needed"""
+
+    settings.ENABLE_QEMU_SUPPORT = True
+    settings.ENABLE_CONFIDENTIAL_COMPUTING = True
+    settings.CONFIDENTIAL_DIRECTORY = Path().resolve()
```
hoh, line 150: This path is used to create side effects on the disc. Why not use a [temporary directory](https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile) ?

src/aleph/vm/sevclient.py
```diff
@@ -0,0 +1,22 @@
...
+        result = await run_in_subprocess(
+            ["sevctl", *args],
+            check=True,
+        )
+
+        return result
+
+    async def export_certificates(self):
```
hoh, line 21: Use "-> None" if function does not return a value

```suggestion
    async def export_certificates(self) -> None:
```

src/aleph/vm/orchestrator/resources.py
```diff
@@ -122,6 +123,21 @@ async def about_system_usage(_: web.Request):
...
+
+    if not settings.ENABLE_CONFIDENTIAL_COMPUTING:
+        return web.HTTPBadRequest(reason="Confidential computing setting not enabled on that server")
+
+    sev_client: SevClient = request.app["sev_client"]
+
+    if not sev_client.certificates_archive.is_file():
+        await sev_client.export_certificates()
+
+    return web.FileResponse(sev_client.certificates_archive)
```
hoh, line 138: Would it not be preferable to move the caching logic inside the `SevClient` class ?

```suggestion
    return web.FileResponse(await sev_client.get_certificates())
```

hoh (commented):
Base looks good,

src/aleph/vm/sevclient.py
nesitor: Yes, some methods can be static, but I prefer to have it in the same class, specially if we will add the `SEV_CTL_PATH` in the future.

src/aleph/vm/sevclient.py
nesitor: The proper name of the method explains perfectly what it does, so I think that add a docstring explanation is totally useless here.

src/aleph/vm/sevclient.py
hoh: My IDE will keep complaining but I mark this as resolved ;-)

src/aleph/vm/orchestrator/resources.py
```diff
@@ -122,6 +123,18 @@ async def about_system_usage(_: web.Request):
...
+@cors_allow_all
+async def about_certificates(request: web.Request):
+    """Public endpoint to expose platform certificates for confidential computing."""
+
+    if not settings.ENABLE_CONFIDENTIAL_COMPUTING:
+        return web.HTTPBadRequest(reason="Confidential computing setting not enabled on that server")
+
+    sev_client: SevClient = request.app["sev_client"]
+
+    return web.FileResponse(await sev_client.get_certificates())
```
hoh, line 135: Is any error expected to be potentially raised by  `await sev_client.get_certificates()` ? That would result in an error 500 (and a log in Sentry).

nesitor: Shouldn't be, because it only executes `sevctl` command, that is already checked on the supervisor initialization, and also the folder is already created at the start.

hoh (approved):

PR #620, opened 2024-05-29
Ol qemu confidential controller

olethanh:
Add the Qemu confidential controler, and endpoints to retrieve the measurements.

### DIFF
--- a/src/aleph/vm/controllers/configuration.py
+++ b/src/aleph/vm/controllers/configuration.py
@@ -29,6 +29,20 @@ class QemuVMConfiguration(BaseModel):
     interface_name: Optional[str]
 
 
+class QemuConfidentialVMConfiguration(BaseModel):
+    qemu_bin_path: str
+    cloud_init_drive_path: Optional[str]
+    image_path: str
+    monitor_socket_path: Path
+    qmp_socket_path: Path
+    vcpu_count: int
+    mem_size_mb: int
+    interface_name: Optional[str]
+    ovmf_path: Path
+    sev_session_file: Path
+    sev_dh_cert_file: Path
+
+
 class HypervisorType(str, Enum):
     qemu = "qemu"
     firecracker = "firecracker"
@@ -37,7 +51,7 @@ class HypervisorType(str, Enum):
 class Configuration(BaseModel):
     vm_id: int
     settings: Settings
-    vm_configuration: Union[QemuVMConfiguration, VMConfiguration]
+    vm_configuration: Union[QemuConfidentialVMConfiguration, QemuVMConfiguration, VMConfiguration]
     hypervisor: HypervisorType = HypervisorType.firecracker
 
 
--- a/src/aleph/vm/controllers/qemu/instance.py
+++ b/src/aleph/vm/controllers/qemu/instance.py
@@ -149,7 +149,6 @@ class AlephQemuInstance(Generic[ConfigurationType], CloudInitMixin, AlephVmContr
     is_instance: bool
     qemu_process: Optional[Process]
     support_snapshot = False
-    qmp_socket_path = None
     persistent = True
     _queue_cancellers: dict[asyncio.Queue, Callable] = {}
     controller_configuration: Configuration
@@ -219,7 +218,7 @@ class AlephQemuInstance(Generic[ConfigurationType], CloudInitMixin, AlephVmContr
 
         logger.debug(f"Making  Qemu configuration: {self} ")
         monitor_socket_path = settings.EXECUTION_ROOT / (str(self.vm_id) + "-monitor.socket")
-        self.qmp_socket_path = qmp_socket_path = settings.EXECUTION_ROOT / (str(self.vm_id) + "-qmp.socket")
+
         cloud_init_drive = await self._create_cloud_init_drive()
 
         image_path = str(self.resources.rootfs_path)
@@ -237,7 +236,7 @@ class AlephQemuInstance(Generic[ConfigurationType], CloudInitMixin, AlephVmContr
             cloud_init_drive_path=cloud_init_drive_path,
             image_path=image_path,
             monitor_socket_path=monitor_socket_path,
-            qmp_socket_path=qmp_socket_path,
+            qmp_socket_path=self.qmp_socket_path,
             vcpu_count=vcpu_count,
             mem_size_mb=mem_size_mb,
             interface_name=interface_name,
@@ -246,7 +245,7 @@ class AlephQemuInstance(Generic[ConfigurationType], CloudInitMixin, AlephVmContr
         configuration = Configuration(
             vm_id=self.vm_id, settings=settings, vm_configuration=vm_configuration, hypervisor=HypervisorType.qemu
         )
-
+        logger.debug(configuration)
         save_controller_configuration(self.vm_hash, configuration)
 
     def save_controller_configuration(self):
@@ -259,6 +258,10 @@ class AlephQemuInstance(Generic[ConfigurationType], CloudInitMixin, AlephVmContr
     @property
     def _journal_stdout_name(self) -> str:
         return f"vm-{self.vm_hash}-stdout"
+
+    @property
+    def qmp_socket_path(self) -> Path:
+        return settings.EXECUTION_ROOT / f"{self.vm_id}-qmp.socket"
 
     @property
     def _journal_stderr_name(self) -> str:
@@ -276,7 +279,6 @@ class AlephQemuInstance(Generic[ConfigurationType], CloudInitMixin, AlephVmContr
         if not ip:
             msg = "Host IP not available"
             raise ValueError(msg)
-
         ip = ip.split("/", 1)[0]
 
         attempts = 30
--- a/src/aleph/vm/controllers/qemu_confidential/instance.py
+++ b/src/aleph/vm/controllers/qemu_confidential/instance.py
@@ -26,6 +26,7 @@ logger = logging.getLogger(__name__)
 
 
 class AlephQemuConfidentialResources(AlephQemuResources):
+    # TODO: Implement download of the custom OVMF bootloader to use if specified, if not only use the default one.
     pass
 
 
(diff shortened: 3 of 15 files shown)


src/aleph/vm/orchestrator/run.py
```diff
@@ -57,31 +57,12 @@ async def create_vm_execution(vm_hash: ItemHash, pool: VmPool, persistent: bool
...
-        pool.forget_vm(vm_hash=vm_hash)
-        raise HTTPInternalServerError(reason="Error during vm initialisation") from error
-    except MicroVMFailedInitError as error:
-        logger.exception(error)
-        pool.forget_vm(vm_hash=vm_hash)
-        raise HTTPInternalServerError(reason="Error during runtime initialisation") from error
-    except HostNotFoundError as error:
-        logger.exception(error)
-        pool.forget_vm(vm_hash=vm_hash)
-        raise HTTPInternalServerError(reason="Host did not respond to ping") from error
```
nesitor, line 84: This method is also used on the method `start_persistent_vm()` and there aren't any other try catch handler there, so these lines are needed I think.
Maybe we can leave it here and remove it in the other `create_vm_execution_or_raise_http_error()` method, that does exactly the same controls.

src/aleph/vm/controllers/configuration.py
```diff
@@ -29,6 +29,20 @@ class QemuVMConfiguration(BaseModel):
...
 
+class QemuConfidentialVMConfiguration(BaseModel):
+    qemu_bin_path: str
+    cloud_init_drive_path: Optional[str]
+    image_path: str
+    monitor_socket_path: Path
+    qmp_socket_path: Path
+    vcpu_count: int
+    mem_size_mb: int
+    interface_name: Optional[str]
```
hoh, line 40: Should this inherit/extend the class `QemuVMConfiguration` ?

src/aleph/vm/controllers/qemu/instance.py
```diff
@@ -260,6 +259,10 @@ def save_controller_configuration(self):
     def _journal_stdout_name(self) -> str:
         return f"vm-{self.vm_hash}-stdout"
 
+    @property
+    def qmp_socket_path(self) -> str:
```
hoh, line 263:
```suggestion
    def qmp_socket_path(self) -> Path:
```

src/aleph/vm/controllers/qemu_confidential/instance.py
```diff
@@ -0,0 +1,111 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class AlephQemuConfidentialResources(AlephQemuResources):
+    pass
+
+
+class AlephQemuConfidentialInstance(AlephQemuInstance):
+    vm_id: int
```
hoh, line 28: Lots of fields are redefined. Does inheritance make sense then or should we rather use a [Protocol](https://docs.python.org/3/library/typing.html#typing.Protocol) ?

src/aleph/vm/controllers/qemu_confidential/instance.py
```diff
@@ -0,0 +1,111 @@
...
+        )
+
+        configuration = Configuration(
+            vm_id=self.vm_id, settings=settings, vm_configuration=vm_configuration, hypervisor=HypervisorType.qemu
+        )
+
+        save_controller_configuration(self.vm_hash, configuration)
+
+    @property
+    def qmp_socket_path(self) -> str:
```
hoh, line 105:
```suggestion
    def qmp_socket_path(self) -> Path:
```

src/aleph/vm/orchestrator/views/operator.py
```diff
@@ -179,12 +185,59 @@ async def operate_reboot(request: web.Request, authenticated_sender: str) -> web
...
         return web.Response(status=200, body=f"Rebooted VM with ref {vm_hash}")
     else:
         return web.Response(status=200, body="Starting VM (was not running) with ref {vm_hash}")
 
 
+@cors_allow_all
+@require_jwk_authentication
+async def operate_sev_measurement(request: web.Request, authenticated_sender: str = None) -> web.Response:
+    """
+    Reboots the virtual machine, smoothly if possible.
```
hoh, line 198:
```suggestion
    Forgot to update the docstring :rofl: 
```

hoh (commented):
Nice start !

I commented on simple stuff not to forget since the PR is still in Draft.

src/aleph/vm/orchestrator/run.py
olethanh: There is no catch handler in `start_persistent_vm()` but they are on the function that call it. The only place I found that used `start_persistent_vm()` and didn't have a try_catch was in `operate_reboot` so I switched it to use `create_vm_execution_or_raise_http_error()` (there is another place but it is in the debug cli so it make no sense to raise an HttpError) . in the end it is the same thing but the other way around (which make more senses with the name)

olethanh:
@hoh Thanks ! but I just opened it so we could sync with @nesitor, I will refactor the whole thing and improve the style then so no need to waste time in  stylistic review for now

src/aleph/vm/orchestrator/run.py
olethanh: I know this isn't related to confidential but it was masking an error while I was debugging

src/aleph/vm/controllers/qemu_confidential/instance.py
olethanh: Doesn't make sense to use a protocol since we want to use the method but I will remove the duplicate field

olethanh: Since it can only be launched via systemd now it will probably make senses at some point to separate what is generating and serializing the config from what is used to launch the emulator. But I don't know enough yet about the firecracker controller   which can be lauched inside and outside systemd to make that split right now.

src/aleph/vm/controllers/qemu_confidential/instance.py
```diff
@@ -0,0 +1,117 @@
...
+    ConfigurationType,
+    logger,
+)
+from aleph.vm.network.interfaces import TapInterface
+
+logger = logging.getLogger(__name__)
+
+
+class AlephQemuConfidentialResources(AlephQemuResources):
+    pass
```
nesitor, line 29:
```suggestion
    # TODO: Implement download of the custom OVMF bootloader to use if specified, if not only use the default one.
    pass
```

src/aleph/vm/controllers/qemu_confidential/instance.py
```diff
@@ -0,0 +1,117 @@
...
+
+        logger.debug(f"Making  Qemu configuration: {self} ")
+        monitor_socket_path = settings.EXECUTION_ROOT / (str(self.vm_id) + "-monitor.socket")
+
+        cloud_init_drive = await self._create_cloud_init_drive()
+
+        image_path = str(self.resources.rootfs_path)
+        vcpu_count = self.hardware_resources.vcpus
+        mem_size_mib = self.hardware_resources.memory
+        mem_size_mb = str(int(mem_size_mib / 1024 / 1024 * 1000 * 1000))
```
nesitor, line 82: Maybe extract all these code to another method to avoid code duplication with `AlephQemuInstance `.

src/aleph/vm/controllers/qemu_confidential/instance.py
```diff
@@ -0,0 +1,117 @@
...
+            vm_id=self.vm_id, settings=settings, vm_configuration=vm_configuration, hypervisor=HypervisorType.qemu
+        )
+        logger.debug(configuration)
+
+        save_controller_configuration(self.vm_hash, configuration)
+
+    async def wait_for_init(self) -> None:
+        """Wait for the init process of the instance to be ready."""
+        # FIXME: Cannot ping since network is not set up yet.
+        return
```
nesitor, line 117: Now the network is working, so we can remove that section :-) .

src/aleph/vm/hypervisors/qemu_confidential/qemuvm.py
```diff
@@ -0,0 +1,122 @@
...
+            # Linux kernel 6.9 added a control on the RDRAND function to ensure that the random numbers generation
+            # works well, on Qemu emulation for confidential computing the CPU model us faked and this makes control
+            # raise an error and prevent boot. Passing the argument --cpu host instruct the VM to use the same CPU
+            # model than the host thus the VM's kernel knows which method is used to get random numbers (Intel and
+            # AMD have different methods) and properly boot.
+            "-cpu",
+            "host",
+            # Uncomment following for debug
+            # "-serial", "telnet:localhost:4321,server,nowait",
+            # "-snapshot",  # Do not save anything to disk
```
nesitor, line 105: Some of that commands are so similar with the standard `QemuVM` class, so maybe will be better to extract to a method and avoid code duplication.

src/aleph/vm/models.py
```diff
@@ -236,7 +239,7 @@ def create(
                 )
             elif self.hypervisor == HypervisorType.qemu:
                 assert isinstance(self.resources, AlephQemuResources)
-                self.vm = vm = AlephQemuInstance(
+                self.vm = vm = AlephQemuConfidentialInstance(
```
nesitor, line 242: These should be fixed to avoid using always a confidential class. You can use the property `is_confidential` to make the filter.

src/aleph/vm/controllers/qemu_confidential/instance.py
olethanh: I guess it depends on at what step we want to make this check. After unencryption ? The machine take a few minutes to boot up so we will probably timeout the http request beforehand.

src/aleph/vm/controllers/qemu_confidential/instance.py
olethanh: we can probably have the default one specified in the client, like we do for runtime, so we would always download one (or take it from the cache)

src/aleph/vm/controllers/qemu_confidential/instance.py
nesitor: In this case, we can increase the timeout, like to 2 minutes or something similar, but ensuring that the VM runs.

nesitor (approved):
We can leave the tests and the refactor for a later step and continue advancing on that feature.

PR #623, opened 2024-05-29
Update test and linting dependencies

hoh:
This updates test dependencies in pyproject.toml.

Support for Python 3.9 is removed in favour of Python 3.10 minimum.

Dependencies for running the software are untouched.


nesitor (approved):
I don't know if we have node operator that runs `aleph-vm` on Debian 3.9. To avoid issues, maybe will be better to announce the deprecation before merge that changes.

olethanh (approved):

PR #624, opened 2024-05-30
Fix: No .deb was built for Ubuntu 24.04

hoh:
This adds support in the packaging/Makefile and in the CI to build packages for Ubuntu 24.04


olethanh (approved):
lgtm.
Outside of the scope of this PR but it doesn't really make sense that the packet we are testing in the matrix droplet as the one we build here.

PR #627, opened 2024-06-04
Implement Start Confidential endpoint

nesitor:
Problem: Now isn't possible as a VM operator to get the client session certificates to initialize a confidential VM.

Solution: Create an operator start endpoint that receive the confidential session files and starts the QEMU VM to continue with the certificate exchange methods.

### DIFF
--- a/src/aleph/vm/orchestrator/views/operator.py
+++ b/src/aleph/vm/orchestrator/views/operator.py
@@ -8,7 +8,6 @@ from aiohttp.web_urldispatcher import UrlMappingMatchInfo
 from aleph_message.exceptions import UnknownHashError
 from aleph_message.models import ItemHash
 from aleph_message.models.execution import BaseExecutableContent
-from anyio import asyncPath
 
 from aleph.vm.conf import settings
 from aleph.vm.models import VmExecution
@@ -165,15 +164,17 @@ async def operate_start(request: web.Request, authenticated_sender: str) -> web.
     if session_file_content:
         return web.Response(status=403, body=f"Session file required for VM with ref {vm_hash}")
 
-    session_file_path = asyncPath(vm_session_path / "vm_session.b64")
-    await session_file_path.write_bytes(session_file_content.file.read())
+    session_file_path = vm_session_path / "vm_session.b64"
+    with open(session_file_path, "wb") as session_file:
+        session_file.write(session_file_content.file.read())
 
     godh_file_content = post.get("godh")
     if godh_file_content:
         return web.Response(status=403, body=f"GODH file required for VM with ref {vm_hash}")
 
-    godh_file_path = asyncPath(vm_session_path / "vm_godh.b64")
-    await godh_file_path.write_bytes(godh_file_content.file.read())
+    godh_file_path = vm_session_path / "vm_godh.b64"
+    with open(godh_file_path, "wb") as godh_file:
+        godh_file.write(godh_file_content.file.read())
 
     pool.systemd_manager.enable_and_start(execution.controller_service)
 
--- a/src/aleph/vm/orchestrator/views/operator.py
+++ b/src/aleph/vm/orchestrator/views/operator.py
@@ -17,7 +17,7 @@ from aleph.vm.orchestrator.views.authentication import (
     require_jwk_authentication,
 )
 from aleph.vm.pool import VmPool
-from aleph.vm.utils import cors_allow_all, write_bytes_file
+from aleph.vm.utils import cors_allow_all
 
 logger = logging.getLogger(__name__)
 
--- a/src/aleph/vm/orchestrator/views/operator.py
+++ b/src/aleph/vm/orchestrator/views/operator.py
@@ -1,4 +1,5 @@
 import logging
+import os
 from datetime import timedelta
 
 import aiohttp.web_exceptions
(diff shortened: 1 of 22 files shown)


src/aleph/vm/utils.py
```diff
@@ -218,3 +218,8 @@ def file_hashes_differ(source: Path, destination: Path, checksum: Callable[[Path
         return True
 
     return checksum(source) != checksum(destination)
+
+
+def write_bytes_file(file_path: Path, file_content: bytes):
+    """Save a file on disk in async way."""
+    file_path.write_bytes(file_content)
```
hoh, line 225: I do not understand this function 🤨
1. How is it better to use `write_bytes_file(path, bytes)` instead of `path.write_bytes(bytes)` ?
2. How is this asynchronous ? Should [aiofiles](https://pypi.org/project/aiofiles/) be used instead ?

src/aleph/vm/pool.py
```diff
@@ -123,7 +123,7 @@ async def create_a_vm(
                 await execution.start()
 
                 # Start VM and snapshots automatically
-                if execution.persistent:
+                if execution.persistent and not execution.is_confidential:
```
hoh, line 126: Can you add a comment to explain why confidential VMs are excluded ?

src/aleph/vm/orchestrator/views/operator.py
```diff
@@ -133,6 +135,52 @@ async def operate_expire(request: web.Request, authenticated_sender: str) -> web
...
+    if execution.is_running:
+        return web.Response(status=403, body=f"VM with ref {vm_hash} already running")
+
+    if not execution.is_confidential:
+        return web.Response(status=403, body=f"Operation not allowed for VM {vm_hash} because it isn't confidential")
+
+    post = await request.post()
+
+    vm_session_path = settings.CONFIDENTIAL_SESSION_DIRECTORY / vm_hash
+    os.makedirs(vm_session_path, exist_ok=True)
```
hoh, line 161:
```suggestion
    vm_session_path.mkdir(exist_ok=True)
```

src/aleph/vm/orchestrator/views/operator.py
```diff
@@ -133,6 +135,52 @@ async def operate_expire(request: web.Request, authenticated_sender: str) -> web
...
+
+    session_file_content = post.get("session")
+    if session_file_content:
```
hoh, line 164: Why is a bad status returned if there is a session file content ?

src/aleph/vm/orchestrator/views/operator.py
```diff
@@ -133,6 +135,52 @@ async def operate_expire(request: web.Request, authenticated_sender: str) -> web
...
+        return web.Response(status=403, body=f"Session file required for VM with ref {vm_hash}")
+
+    session_file_path = vm_session_path / "vm_session.b64"
+    with open(session_file_path, "wb") as session_file:
+        session_file.write(session_file_content.file.read())
```
hoh, line 169: If the entire content is read in memory, you might as well use pathlib.Path helpers:
```suggestion
    session_file_path.write_bytes(session_file_content.file.read())
```

src/aleph/vm/orchestrator/views/operator.py
```diff
@@ -133,6 +135,52 @@ async def operate_expire(request: web.Request, authenticated_sender: str) -> web
...
+
+    godh_file_content = post.get("godh")
+    if godh_file_content:
```
hoh, line 172: Same question: if there is a content, why is a 403 returned ?

src/aleph/vm/orchestrator/views/operator.py
```diff
@@ -133,6 +135,52 @@ async def operate_expire(request: web.Request, authenticated_sender: str) -> web
...
+        return web.Response(status=403, body=f"GODH file required for VM with ref {vm_hash}")
+
+    godh_file_path = vm_session_path / "vm_godh.b64"
+    with open(godh_file_path, "wb") as godh_file:
+        godh_file.write(godh_file_content.file.read())
```
hoh, line 177: If the entire content is read in memory, you might as well use pathlib.Path helpers:
```suggestion
    godh_file_path.write_bytes(godh_file_content.file.read())
```

hoh (commented):
I guess I cannot test this yet ?

src/aleph/vm/pool.py
nesitor: We need to wait for the user session files (session.b64 and godh.b64 files) before start the instance using systemd, this is why I'm creating the new endpoint, to receive that files and start the systemd service once we have it.

src/aleph/vm/utils.py
nesitor: Initially I tried to write files in async way, and I created that method for that thing. But after I get that the files received by the endpoint are in memory and are so small, so isn't needed to write the files in async way. So finally I made it sync and I forgot to remove that method. Now I have removed it.

src/aleph/vm/orchestrator/views/operator.py
nesitor: Things that happens to write the code faster hahaha, now it's fixed.

nesitor:
> I guess I cannot test this yet ?

Yes, you can test it manually, but I need also to add automatic tests on that PR as I commented on the chat.

olethanh:
I'm merging in my branch so I can test further but I have two remarks, not linked to code quality but with the architecure

1.   Not  too fond about naming this `/start`, it only works with Confidential and the VM isn't really started, only the qemu process is launched. So it will probably not do what the user expect. I think you had suggested `init_encrypted_sessions` or maybe `provide_session_certificates` ?
2. The number of directory in the Settings are starting to explode, so do really need this configurable by the user? Can't we just have a folder by VM and put everything in there ? Eventually in a sub directory.

olethanh (approved):

PR #629, opened 2024-06-05
Fix: Binary `sevctl` was absent from debian packages

hoh:
Solution: Build `sevctl` using an upstream version of Rust (the version in Debian is not supported), and bundle it in the Debian packages.

Add a setting in aleph-vm with the path of the bundled binary.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -14,3 +14,4 @@
 **/data.tgz
 /pydantic/
 **/target
+/packaging/sevctl/target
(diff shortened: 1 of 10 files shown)


nesitor (approved):

olethanh:
Test are failing
``` cd ./sevctl && cargo build --release --target x86_64-unknown-linux-gnu
error: could not find `Cargo.toml` in `/opt/packaging/sevctl` or any parent directory
make: *** [Makefile:51: target/sevctl] Error 101
make: *** [Makefile:93: all-podman-debian-12] Error 2
ls: cannot access 'packaging/target': No such file or directory
```

nesitor (approved):
If we enable git submodules on the repo, maybe we need to also add it on the docs.

PR #630, opened 2024-06-07
Problem: Crash when trying to auth via websocket

olethanh:
The auth function in websocket was crashing when the header  "X-Auth-Signature" wasn't passed, even  that authentification method wasn't used.

Make it an option in the method so we can still have error raised in normal header auth

### DIFF
--- a/src/aleph/vm/orchestrator/views/__init__.py
+++ b/src/aleph/vm/orchestrator/views/__init__.py
@@ -337,11 +337,11 @@ async def status_public_config(request: web.Request):
     )
 
 
-def authenticate_api_request(request: web.Request) -> bool:
+def authenticate_api_request(request: web.Request, raises_on_missing_header=True) -> bool:
     """Authenticate an API request to update the VM allocations."""
     signature: bytes = request.headers.get("X-Auth-Signature", "").encode()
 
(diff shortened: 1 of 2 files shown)


src/aleph/vm/orchestrator/views/__init__.py
```diff
@@ -337,11 +337,11 @@ async def status_public_config(request: web.Request):
     )
 
 
-def authenticate_api_request(request: web.Request) -> bool:
+def authenticate_api_request(request: web.Request, raises_on_missing_header=True) -> bool:
```
hoh, line 340: What about making this the default behavior and skipping the extra argument ?

olethanh: The update allocation endpoint rely on that behaviour

hoh:
Calling `authenticate_api_request` from a websocket endpoint does not make sense, and it seems more appropriate to now allow header authentication on that endpoint.

My proposal on this branch: https://github.com/aleph-im/aleph-vm/pull/632

hoh (changes requested):
My proposal on this branch: https://github.com/aleph-im/aleph-vm/pull/632

PR #631, opened 2024-06-12
Fix: Correct string formatting in VM startup response

1yam:
Problem:
{vm_hash} was not being replaced with the actual value.

image


olethanh (approved):

hoh:
Tests failed due to the DigitalOcean token not available in forks.

```
>>> doctl version v1.107.0 installed to /opt/hostedtoolcache/doctl/1.107.0/x64
Error: Input required and not supplied: token
```

PR #632, opened 2024-06-12
Fix: Endpoint authenticate_api_request is not adequate for Websocket connections

hoh:
…t connections

This caused issues when header "X-Auth-Signature" was not passed.


olethanh:
iirc I had added this  auth method because it was easier when doing debugging python script but you are right maybe it's not needed anymore since we will have client code for this now

PR #638, opened 2024-06-19
Provide confidential example image

olethanh:
I have moved the creation scripts from Olivier Desenfans , and improved them
They now  clean up ressources  the loop device and mount on crash.
Compared to the original method it is not necessary anymore to log into the original image to upgrade the kernel and install cryptsetup before launching the scripts.
The whole procedure is also now documented

I have also fixed a bug with the permission that broke `sudo` and other suid executable inside the VM.


nesitor (approved):
LGTM, but maybe we can move that scripts to the runtimes directory, because indeed it creates a runtime instead of an example program.

PR #640, opened 2024-06-19
Disable printing system logs on deb package as per default configuration recommendation

aliel:
https://github.com/aleph-im/aleph-vm/blob/main/src/aleph/vm/conf.py#L139


hoh:
As suggested in the chat, deleting this line would make more sense.

hoh (approved):

nesitor:
> These logs are indeed not so useful anymore for debugging, at least not enough to be enabled by default.

This is not true, we are using also this line for printing SystemD logs on instances. So maybe we need to add another argument to print that systemd logs.

hoh:
Hum... I should not have merged this so late in the evening, gonna review this more in details.

PR #644, opened 2024-06-21
Problem : Log was not working on system.  Unify logging method

olethanh:
Restore the logging method using systemd and use it on all execution types. This method use the systemd journal library to open a file descriptor, which we pass to the subprocess. This should provide a performance boost since there is no python handling of the process output. 
 

- Restored the method of logging in journald in Qemu
- Ported it to Firecracker hypervisor
- Now all Execution (program / instance   qemu / firecracker)  use the same unified logging method. 
- Streaming Logging endpoint now works properly and return all logs.
-  As per discussion: PRINT_SYSTEM_LOGS only affect Program execution, since we decided it made no sense or performance impact on Instances.
- When logs are disabled connect the Firecracker VM to  `/dev/null` which should further improve performances.


nesitor (approved):
🚀

PR #645, opened 2024-06-21
Problem: Websocket were required to fetch logs

olethanh:
Solution: Provide a view that works in get

not sure about the endpoint name.
It's just a proposal to see if it is interesting to have it Still have to fix  the style and doc but want to gather your feed back first

### DIFF
--- a/src/aleph/vm/orchestrator/supervisor.py
+++ b/src/aleph/vm/orchestrator/supervisor.py
@@ -49,6 +49,7 @@ from .views import (
 from .views.operator import (
     operate_erase,
     operate_expire,
+    operate_logs,
     operate_reboot,
     operate_stop,
     stream_logs,
--- a/src/aleph/vm/orchestrator/supervisor.py
+++ b/src/aleph/vm/orchestrator/supervisor.py
@@ -102,6 +102,7 @@ def setup_webapp():
         # /control APIs are used to control the VMs and access their logs
         web.post("/control/allocation/notify", notify_allocation),
         web.get("/control/machine/{ref}/logs", stream_logs),
+        web.get("/control/machine/{ref}/logs2", operate_logs),
         web.post("/control/machine/{ref}/expire", operate_expire),
         web.post("/control/machine/{ref}/stop", operate_stop),
         web.post("/control/machine/{ref}/erase", operate_erase),
--- a/src/aleph/vm/orchestrator/supervisor.py
+++ b/src/aleph/vm/orchestrator/supervisor.py
@@ -101,8 +101,8 @@ def setup_webapp():
         web.get("/about/config", about_config),
         # /control APIs are used to control the VMs and access their logs
         web.post("/control/allocation/notify", notify_allocation),
-        web.get("/control/machine/{ref}/logs", stream_logs),
-        web.get("/control/machine/{ref}/logs2", operate_logs),
+        web.get("/control/machine/{ref}/stream_logs", stream_logs),
+        web.get("/control/machine/{ref}/logs", operate_logs),
         web.post("/control/machine/{ref}/expire", operate_expire),
         web.post("/control/machine/{ref}/stop", operate_stop),
         web.post("/control/machine/{ref}/erase", operate_erase),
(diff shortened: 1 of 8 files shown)


src/aleph/vm/utils/logs.py
```diff
@@ -56,3 +58,28 @@ def do_cancel():
         r.close()
 
     return queue, do_cancel
+
+
+def get_past_vm_logs(stdout_identifier, stderr_identifier, skip_past=False) -> Generator[EntryDict, None, None]:
+    """Create a queue which streams the logs for the process.
```
hoh, line 64: Obsolete docsting ?

src/aleph/vm/orchestrator/views/operator.py
```diff
@@ -87,6 +88,27 @@ async def stream_logs(request: web.Request) -> web.StreamResponse:
...
+    execution = get_execution_or_404(vm_hash, pool=pool)
+    if not is_sender_authorized(authenticated_sender, execution.message):
+        return web.Response(status=403, body="Unauthorized sender")
+
+    response = web.StreamResponse()
+    response.headers["Content-Type"] = "text/plain"
+    await response.prepare(request)
+
+    for entry in execution.vm.past_logs():
+        msg = entry["__REALTIME_TIMESTAMP"].isoformat() + "> " + entry["MESSAGE"] + "\n"
```
hoh, line 106: The formatting can be discussed, but it looks acceptable to me.

nesitor (approved):

src/aleph/vm/orchestrator/supervisor.py
```diff
@@ -101,6 +102,7 @@ def setup_webapp():
         # /control APIs are used to control the VMs and access their logs
         web.post("/control/allocation/notify", notify_allocation),
         web.get("/control/machine/{ref}/logs", stream_logs),
+        web.get("/control/machine/{ref}/logs2", operate_logs),
```
nesitor, line 105: I think that will be better to maintain the same endpoint and detect if we are using Websockets or not.

hoh: I agree with this.

olethanh: I'm not too sure the behaviour are quite different, one is streaming and returning json, the other is only sending existing logs as text. We could merge both and have some kind of behemoth that can handle everything (text/json/streaming and not streaming) but that would multiply the complexity

olethanh: it's doable but it  is getting a bit out of scope for  a simple debug endpoint

nesitor: So maybe use a different endpoint path, like `steam_logs` for WebSockets endpoint and normal `logs` for REST endpoint.

olethanh:
I have renamed the endpoint as suggested and added tests for them.

src/aleph/vm/orchestrator/views/operator.py
```diff
@@ -87,6 +89,27 @@ async def stream_logs(request: web.Request) -> web.StreamResponse:
             execution.vm.unregister_queue(queue)
 
 
+@cors_allow_all
+@require_jwk_authentication
+async def operate_logs(request: web.Request, authenticated_sender: str) -> web.StreamResponse:
```
hoh, line 94: `operate_logs` seems a weird name :thinking: 

```suggestion
async def get_vm_logs(request: web.Request, authenticated_sender: str) -> web.StreamResponse:
```

src/aleph/vm/orchestrator/views/operator.py
```diff
@@ -87,6 +89,27 @@ async def stream_logs(request: web.Request) -> web.StreamResponse:
...
+    if not is_sender_authorized(authenticated_sender, execution.message):
+        return web.Response(status=403, body="Unauthorized sender")
+
+    response = web.StreamResponse()
+    response.headers["Content-Type"] = "text/plain"
+    await response.prepare(request)
+
+    for entry in execution.vm.past_logs():
+        msg = f'{entry["__REALTIME_TIMESTAMP"].isoformat()}> {entry["MESSAGE"]}'
+        await response.write(msg.encode())
```
hoh, line 108: I was wondering if this could cause performance issues because the generator is not async, but the presence of an `await` in the loop should at least give control back to the event loop frequently.

hoh (approved):
Neat code. I added two nitpicks, but good overall.

I see a test failing however.

src/aleph/vm/orchestrator/views/operator.py
olethanh: I don't like it much but it stay consistant with the other naming of the method

PR #646, opened 2024-06-21
Implement custom OVMF firmware

nesitor:
Problem: A user cannot specify which OVMF firmware want to use for the confidential instances.

Solution: Use new `aleph-message` version that includes that data schema and implement it on the qemu confidential resources.
It also implements the extra volume use for qemu instances.


olethanh (approved):

olethanh:
I made a few fix after testing, waiting to see if the CI pass now

olethanh:
Had to make some more modification to make the volumes working

PR #647, opened 2024-06-26
Move `domain` payload field to Operation token instead PubKey one

nesitor:
Problem: If a user wants to manage different operations for a different CRNs, they have to sign a new pubkey token for every CRN, and this is so bad for the user experience.

Solution: Move the `domain` field to the operation token payload instead the pubkey one, just to improve the user experience and maintain the security integrity.


hoh:
This proposal trades security versus convenience, we need to double check it:

What happens if the SignedPubKeyPayload is leaked ? With a domain builtin, only VMs on a specific CNR could be messed up with. There is no revocation mechanism with this approach to authentication, so the user would not be able to do much.

hoh (approved):

hoh:
@nesitor can you merge this when you think appropriate ?

Issue #652, opened 2024-07-04
AttributeError: 'MicroVM' object has no attribute 'send_shutdown_message'


hoh:
On staging-ovh, version `0.4.1-53-g4a9eabd`.

```
Jul 04 13:11:02 staging-ovh vm-3176162d9646d0987e656cf6237bff5d2a0612b42ee379b10cd798a57e9ed8b5-stderr[2525280]: thread 'main' panicked at 'Jailer error: Failed to create /dev/userfaultfd via mknod inside the jail: File exists (os error 17)', src/jailer/src/main.rs:377:27
Jul 04 13:11:02 staging-ovh vm-3176162d9646d0987e656cf6237bff5d2a0612b42ee379b10cd798a57e9ed8b5-stderr[2525280]: note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Jul 04 13:11:02 staging-ovh systemd[1]: aleph-vm-controller@3176162d9646d0987e656cf6237bff5d2a0612b42ee379b10cd798a57e9ed8b5.service: Main process exited, code=exited, status=1/FAILURE
Jul 04 13:11:02 staging-ovh systemd[1]: aleph-vm-controller@3176162d9646d0987e656cf6237bff5d2a0612b42ee379b10cd798a57e9ed8b5.service: Failed with result 'exit-code'.
Jul 04 13:11:05 staging-ovh systemd[1]: aleph-vm-controller@f6f5bea7e32a44521fe203ea04d47b9afe57cfe1af010e00919b2ba693aaa5e4.service: Scheduled restart job, restart counter is at 2299.
Jul 04 13:11:05 staging-ovh systemd[1]: Stopped aleph-vm-controller@f6f5bea7e32a44521fe203ea04d47b9afe57cfe1af010e00919b2ba693aaa5e4.service - Aleph VM f6f5bea7e32a44521fe203ea04d47b9afe57cfe1af010e00919b2ba693aaa5e4 Controller.
Jul 04 13:11:05 staging-ovh systemd[1]: Started aleph-vm-controller@f6f5bea7e32a44521fe203ea04d47b9afe57cfe1af010e00919b2ba693aaa5e4.service - Aleph VM f6f5bea7e32a44521fe203ea04d47b9afe57cfe1af010e00919b2ba693aaa5e4 Controller.
Jul 04 13:11:06 staging-ovh python3[2525284]: Traceback (most recent call last):
[...]

PR #653, opened 2024-07-04
Fix: AttributeError: 'MicroVM' object has no attribute 'send_shutdown_message'

hoh:
Error reported on https://github.com/aleph-im/aleph-vm/issues/652

### DIFF
--- a/src/aleph/vm/hypervisors/qemu/qemuvm.py
+++ b/src/aleph/vm/hypervisors/qemu/qemuvm.py
@@ -127,5 +127,5 @@ class QemuVM:
             client.close()
 
     async def teardown(self):
-        """Stop the VM, cleanup network interface and remove data directory."""
+        """Stop the VM."""
         self.send_shutdown_message()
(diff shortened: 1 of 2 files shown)


nesitor (approved):

src/aleph/vm/hypervisors/qemu/qemuvm.py
```diff
@@ -125,3 +125,7 @@ def send_shutdown_message(self):
                 logger.warning("unexpected answer from VM", resp)
             print("shutdown message sent")
             client.close()
+
+    async def teardown(self):
+        """Stop the VM, cleanup network interface and remove data directory."""
```
olethanh, line 130: The docstring is wrong
it's not actually doing any of that except stopping the vm.

Even on microvm it's not cleaning up the interface

olethanh (changes requested):

src/aleph/vm/hypervisors/qemu/qemuvm.py
olethanh, line 130:
```suggestion
        """Stop the VM."""
```

olethanh (approved):

PR #654, opened 2024-07-04
Improve instances code

nesitor:
Problem: If an instance is created and removed, one of the base mounted devices remains on the system. Also, a user cannot create a `Qemu` disk from Debian image.

Solution: Remove the useless device from the system and create the two needed script to create `Qemu` Debian runtimes.


olethanh (approved):
lgtm

PR #658, opened 2024-07-17
Solve Firecracker reboot issues

nesitor:
Problem: If the frontend or a user send the request to reboot a Firecracker instance, it stops but doesn't start again.

Solution: Change method to restart it just stopping the process, cleaning the firecracker run files and starting it again.


olethanh:
The lifecyle of the Execution in the different use case are not clear to me. Is the tear down ever called for Firecracker VM when it's running inside systemd?

nesitor:
> The lifecyle of the Execution in the different use case are not clear to me. Is the tear down ever called for Firecracker VM when it's running inside systemd?

Yes, previously to the PR, the Firecracker VM was removed completely on a service stop or restart. Now is only stopping the process and running it again. Only in the case that the instance is removed completely from the system should do the `teardown` process.

PR #659, opened 2024-07-17
Small fixes noticed on new installations.

nesitor:
Problem: If the server doesn't have access to Internet, it just crashes. Also, if an instance is removed, it fails removing the base block device.

Solution: Handle the crash and log as an error about the Internet connection. Check if the mounted block devices exist before removing it.

### DIFF
--- a/src/aleph/vm/hypervisors/firecracker/microvm.py
+++ b/src/aleph/vm/hypervisors/firecracker/microvm.py
@@ -472,12 +472,16 @@ class MicroVM:
         if self.stderr_task:
             self.stderr_task.cancel()
 
+        # Clean mounted block devices
         if self.mounted_rootfs:
             logger.debug("Waiting for one second for the VM to shutdown")
             await asyncio.sleep(1)
-            root_fs = self.mounted_rootfs.name
-            system(f"dmsetup remove {root_fs}")
-            system(f"dmsetup remove {root_fs}_base")
+            if self.mounted_rootfs.is_block_device():
+                root_fs = self.mounted_rootfs.name
+                system(f"dmsetup remove {root_fs}")
+            base_device = Path(self.mounted_rootfs.name.replace("_rootfs", "_base"))
+            if base_device.is_block_device():
+                system(f"dmsetup remove {base_device}")
             if self.use_jailer and Path(self.jailer_path).is_dir():
                 shutil.rmtree(self.jailer_path)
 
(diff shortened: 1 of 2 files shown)


src/aleph/vm/orchestrator/status.py
```diff
@@ -120,6 +120,10 @@ async def check_internet(session: ClientSession, vm_id: ItemHash) -> bool:
     try:
         response: dict = await get_json_from_vm(session, vm_id, "/internet")
 
+        if not hasattr(response, "headers"):
```
hoh, line 123: Is there no error message to check instead of looking for a missing key ?

src/aleph/vm/hypervisors/firecracker/microvm.py
```diff
@@ -472,12 +472,16 @@ async def teardown(self):
...
             self.stderr_task.cancel()
 
+        # Clean mounted block devices
         if self.mounted_rootfs:
             logger.debug("Waiting for one second for the VM to shutdown")
             await asyncio.sleep(1)
-            root_fs = self.mounted_rootfs.name
-            system(f"dmsetup remove {root_fs}")
-            system(f"dmsetup remove {root_fs}_base")
+            if self.mounted_rootfs.is_block_device():
```
hoh, line 479: Can you add a comment in the code explaining this ?

hoh (commented):
These are two completely unrelated fixes, can you at least make two distinct commits, preferably two PRs ?

src/aleph/vm/orchestrator/status.py
nesitor: If the request times out and don't return a valid response, the key `headers` is missing, so to prevent that I have put that condition to avoid having more issues.

src/aleph/vm/hypervisors/firecracker/microvm.py
nesitor: I have put the comment up: https://github.com/aleph-im/aleph-vm/pull/659/files/17d5ab5c511430f0467dceda1a9c12b82fddcd2c#diff-81d0c3cbfe4e4c759d3074c29ddc23188851005f7372b1c8ac83eb1d35f80857R475

nesitor:
> These are two completely unrelated fixes, can you at least make two distinct commits, preferably two PRs ?

This is not totally unrelated, are issues found installing a new server.
Why put it in different commits or different PRs? It's just 6 lines of code in total.

PR #662, opened 2024-07-29
Enable Qemu support by default

nesitor:
Problem: If a user wants to launch a Qemu instance inside a CRN, by default the support is disabled.

Solution: Enable the support by default and disable snapshots because aren't used.


philogicae (approved):
LGTM

PR #679, opened 2024-08-21
Check message status before checking the payment

nesitor:
Problem: If a user allocates a VM and later forgets the VM, the payment task fails because cannot get the price for that execution.

Solution: Check the message status before checking the price and remove the execution if it is forgotten or on a different status than `processed`.

### DIFF
--- a/src/aleph/vm/orchestrator/messages.py
+++ b/src/aleph/vm/orchestrator/messages.py
@@ -1,7 +1,7 @@
 import asyncio
 import copy
 
-from aiohttp import ClientSession, ClientConnectorError, ClientResponseError
+from aiohttp import ClientConnectorError, ClientResponseError, ClientSession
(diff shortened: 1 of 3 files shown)


src/aleph/vm/orchestrator/messages.py
```diff
@@ -69,3 +71,15 @@ async def load_updated_message(
     message = copy.deepcopy(original_message)
     await update_message(message)
     return message, original_message
+
+
+async def get_message_status(item_hash: ItemHash) -> MessageStatus:
+    """Fetch the status of an execution from the reference API server."""
```
olethanh, line 77: The rest of the code fetch messages from connector but here it is using the api server. I'm supposing there is a reason for that but it would be nice to have it explained and written.  Especially for maintainability

nesitor: Indeed, so I added the explanation on the method description :smiley:

olethanh (approved):

PR #682, opened 2024-08-23
Problem: FirecrackerVM not working if /var/lib and /var/cache on two …

olethanh:
…separate partion

The prepare step for jailer was failing because it couldn't create a link to rootfs file when the CACHE and EXECUTION were not on the same partition

This was due do trying to make a hardlink instead of as soft symlink (contrary to what the docstring indicated)

Solution: Make a symlink

### DIFF
--- a/src/aleph/vm/hypervisors/firecracker/microvm.py
+++ b/src/aleph/vm/hypervisors/firecracker/microvm.py
@@ -324,7 +324,7 @@ class MicroVM:
             rootfs_filename = Path(path_on_host).name
             jailer_path_on_host = f"/opt/{rootfs_filename}"
             try:
-                os.link(path_on_host, f"{self.jailer_path}/{jailer_path_on_host}")
+                os.symlink(path_on_host, f"{self.jailer_path}/{jailer_path_on_host}")
             except FileExistsError:
                 logger.debug(f"File {jailer_path_on_host} already exists")
             return Path(jailer_path_on_host)
--- a/src/aleph/vm/hypervisors/firecracker/microvm.py
+++ b/src/aleph/vm/hypervisors/firecracker/microvm.py
@@ -1,4 +1,5 @@
 import asyncio
+import errno
 import json
 import logging
 import os.path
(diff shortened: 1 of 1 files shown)


src/aleph/vm/hypervisors/firecracker/microvm.py
```diff
@@ -324,7 +324,7 @@ def enable_file_rootfs(self, path_on_host: Path) -> Path:
             rootfs_filename = Path(path_on_host).name
             jailer_path_on_host = f"/opt/{rootfs_filename}"
             try:
-                os.link(path_on_host, f"{self.jailer_path}/{jailer_path_on_host}")
+                os.symlink(path_on_host, f"{self.jailer_path}/{jailer_path_on_host}")
```
nesitor, line 327: I think that this needs to be a hard link, because for Firecracker instances, the rootfs file is a devmapper device, and using jailer the unique way to make it work is doing a hard link, not a symlink.

hoh:
What would happen if the cache is cleared while a VM is working and using a file stored there then ?
With hardlinks, this would not be an issue since the file would still exist.

olethanh:
> What would happen if the cache is cleared while a VM is working and using a file stored there then ? With hardlinks, this would not be an issue since the file would still exist.

While the VM is working the file is open, and as long as a program has a fd to a file Linux won't actually delete it from the disk so it isn't a problem. But anyway we don't use a symlink anymore

src/aleph/vm/hypervisors/firecracker/microvm.py
```diff
@@ -318,15 +319,23 @@ def enable_rootfs(self, path_on_host: Path) -> Path:
...
-        Creates a symlink to the rootfs file if jailer is in use.
+        If jailer is in use, try to create a hardlink
+        If it is not possible to create a link because the dir are in separate device made a copy.
         """
         if self.use_jailer:
             rootfs_filename = Path(path_on_host).name
             jailer_path_on_host = f"/opt/{rootfs_filename}"
             try:
-                os.link(path_on_host, f"{self.jailer_path}/{jailer_path_on_host}")
+                os.symlink(path_on_host, f"{self.jailer_path}/{jailer_path_on_host}")
```
nesitor, line 329:
```suggestion
                os.link(path_on_host, f"{self.jailer_path}/{jailer_path_on_host}")
```
So here we need to restore the hardlink also.

PR #686, opened 2024-08-28
Fix CI slowness and correct  execution tests

olethanh:
This PR solve a few issue with the tests:

- test_create_execution Was VERY VERY slow, more than 10 minutes to run. -> Now it run in less than 1 minutes
- test_create_execution_online  and test_create_execution_legacy were actually running the same test as test_create_execution. (because of a settings contamination) Which compelled the problem and made the whole test suite take 3/4h to run
- Additionally the code path of running a real program wasn't tested
- The Workflow name was not correct
- Execution test  of a program were failing on python 3.12

## Changes
see above

## How to test
Everything is in CI

### DIFF
--- a/.github/workflows/test-using-pytest.yml
+++ b/.github/workflows/test-using-pytest.yml
@@ -7,6 +7,12 @@ jobs:
   tests-python:
     name: "Test Python code"
     runs-on: ubuntu-22.04
+    services:
+      # Run vm connector for the execution tests
+      vm-connector:
+        image: alephim/vm-connector:alpha
+        ports:
+          - 4021:4021
 
     steps:
       - uses: actions/checkout@v4
--- a/.github/workflows/test-using-pytest.yml
+++ b/.github/workflows/test-using-pytest.yml
@@ -1,4 +1,4 @@
-name: "Test on DigitalOcean Droplets"
+name: "py.test and linting"
 
 on:
   push
--- a/runtimes/aleph-debian-11-python/init1.py
+++ b/runtimes/aleph-debian-11-python/init1.py
@@ -247,6 +247,7 @@ async def setup_code_asgi(code: bytes, encoding: Encoding, entrypoint: str) -> A
         module = __import__(module_name)
         for level in module_name.split(".")[1:]:
             module = getattr(module, level)
+        logger.debug("import done")
         app = getattr(module, app_name)
     elif encoding == Encoding.plain:
         # Execute the code and extract the entrypoint
--- a/src/aleph/vm/hypervisors/firecracker/microvm.py
+++ b/src/aleph/vm/hypervisors/firecracker/microvm.py
@@ -331,8 +331,8 @@ class MicroVM:
                 logger.debug(f"File {jailer_path_on_host} already exists")
             except OSError as err:
                 if err.errno == errno.EXDEV:
-                    #  Invalid cross-device link:
-                    # cannot make hard link between partition. Make a copy instead
+                    # Invalid cross-device link: cannot make hard link between partition.
+                    # In this case, copy the file instead:
                     shutil.copyfile(path_on_host, f"{self.jailer_path}/{jailer_path_on_host}")
                 else:
                     raise
--- a/src/aleph/vm/hypervisors/firecracker/microvm.py
+++ b/src/aleph/vm/hypervisors/firecracker/microvm.py
@@ -1,4 +1,5 @@
 import asyncio
+import errno
 import json
 import logging
 import os.path
@@ -318,7 +319,8 @@ class MicroVM:
     def enable_file_rootfs(self, path_on_host: Path) -> Path:
         """Make a rootfs available to the VM.
 
-        Creates a symlink to the rootfs file if jailer is in use.
+        If jailer is in use, try to create a hardlink
+        If it is not possible to create a link because the dir are in separate device made a copy.
         """
         if self.use_jailer:
             rootfs_filename = Path(path_on_host).name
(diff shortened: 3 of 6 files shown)


.github/workflows/test-using-pytest.yml
```diff
@@ -7,6 +7,13 @@ jobs:
   tests-python:
     name: "Test Python code"
     runs-on: ubuntu-22.04
+    services:
+      # Run vm connector for the execution tests
+      vm-connector:
+        image: alephim/vm-connector:alpha
+#        env:
```
hoh, line 14: Commented out ?

runtimes/aleph-debian-11-python/init1.py
```diff
@@ -247,6 +247,7 @@ async def setup_code_asgi(code: bytes, encoding: Encoding, entrypoint: str) -> A
         module = __import__(module_name)
         for level in module_name.split(".")[1:]:
             module = getattr(module, level)
+        logger.debug("import done")
```
hoh, line 250: Should this be kept ?

src/aleph/vm/hypervisors/firecracker/microvm.py
```diff
@@ -489,7 +498,12 @@ async def teardown(self):
         if self._unix_socket:
             logger.debug("Closing unix socket")
             self._unix_socket.close()
-            await self._unix_socket.wait_closed()
+            try:
+                await asyncio.wait_for(self._unix_socket.wait_closed(), 2)
+            except asyncio.TimeoutError:
+                # In  Python < 3.11 wait_closed() was broken and returned immediatly
+                # It is supposedly fixed in python  3.12.1 but hang indefinelty during the tests.
```
hoh, line 505:
```suggestion
                # It is supposedly fixed in Python 3.12.1, but it hangs indefinitely during tests.
```

tests/supervisor/test_execution.py
```diff
@@ -4,29 +4,35 @@
...
-async def test_create_execution():
+async def test_create_execution(mocker):
     """
     Create a new VM execution and check that it starts properly.
     """
+    mock_settings = Settings()
+    mocker.patch("aleph.vm.conf.settings", new=mock_settings)
+    mocker.patch("aleph.vm.storage.settings", new=mock_settings)
+    mocker.patch("aleph.vm.controllers.firecracker.executable.settings", new=mock_settings)
+    mocker.patch("aleph.vm.controllers.firecracker.program.settings", new=mock_settings)
```
hoh, line 24: Would a loop help here ?

```suggestion
    for path in [
        "aleph.vm.conf.settings",
        "aleph.vm.storage.settings",
        "aleph.vm.controllers.firecracker.executable.settings",
        "aleph.vm.controllers.firecracker.program.settings"
    ]:
        mocker.patch(path, new=Settings())
```

tests/supervisor/test_execution.py
```diff
@@ -73,29 +80,34 @@ async def test_create_execution_online(vm_hash: ItemHash = None):
...
     assert isinstance(vm, AlephFirecrackerProgram)
+    vm.enable_console = True
+    vm.fvm.enable_log = True
     assert vm.vm_id == 3
 
     await execution.start()
     await execution.stop()
 
 
+# This test depends on having a vm-connector running on port 4021
```
hoh, line 110: Should the test assert that the `vm-connector` can be reached ?

tests/supervisor/test_execution.py
```diff
@@ -73,29 +80,34 @@ async def test_create_execution_online(vm_hash: ItemHash = None):
...
+        original=original_message.content,
         snapshot_manager=None,
         systemd_manager=None,
         persistent=False,
     )
 
-    # Downloading the resources required may take some time, limit it to 10 seconds
-    await asyncio.wait_for(execution.prepare(), timeout=30)
+    # Downloading the resources required may take some time, limit it to 120 seconds
+    # it's a bit slower in github ci
```
hoh, line 95:
```suggestion
    # since it is a bit slow in GitHub Actions
```

src/aleph/vm/hypervisors/firecracker/microvm.py
```diff
@@ -327,6 +329,13 @@ def enable_file_rootfs(self, path_on_host: Path) -> Path:
                 os.link(path_on_host, f"{self.jailer_path}/{jailer_path_on_host}")
             except FileExistsError:
                 logger.debug(f"File {jailer_path_on_host} already exists")
+            except OSError as err:
+                if err.errno == errno.EXDEV:
+                    #  Invalid cross-device link:
+                    # cannot make hard link between partition. Make a copy instead
```
hoh, line 335:
```suggestion
                    # Invalid cross-device link: cannot make hard link between partition. 
                    # In this case, copy the file instead:
```

tests/supervisor/test_execution.py
olethanh: Won't really change anything

tests/supervisor/test_execution.py
olethanh: Frankly I find it less readable

.github/workflows/test-using-pytest.yml
olethanh: fixed

runtimes/aleph-debian-11-python/init1.py
olethanh: The underlying problem isn't really fixed so it will helps when the issue occur again.

nesitor (approved):

tests/supervisor/test_execution.py
nesitor: Yes, I see better to be more explicit here also.

PR #689, opened 2024-09-03
Fix 199 ruff errors

hoh:
Explain what problem this PR is resolving

The python linter [ruff](https://github.com/astral-sh/ruff) is not enable on this project because it raises too many errors to fix at once. At the time of writing, `ruff check src` raises 375 errors.

## Self proofreading checklist

Not relevant for these changes.

- [ ] The new code clear, easy to read and well commented.
- [ ] New code does not duplicate the functions of builtin or popular libraries.
- [ ] An LLM was used to review the new code and look for simplifications.
- [ ] New classes and functions contain docstrings explaining what they provide.
- [ ] All new code is covered by relevant tests.
- [ ] Documentation has been updated regarding these changes.

## Changes

The first commit contains the result of `ruff check src --fix`, which contains safe automatic fixes to the codebase.

I then ran `ruff check src --fix --unsafe-fixes` and added the following commits with the changes that I reviewed manually and deem ready to commit:
- Fix: Generator loop -> 'yield from'
- Fix: Type annotations could be improved
- Fix: Python < 3.10 is no supported anymore
- Fix: Ruff errored on exception raising

## How to test

Run `ruff check src` or `hatch run testing:ruff`.

`ruff check src` on the code found 176 errors after the changes.

PR #690, opened 2024-09-03
ci/fix(test-using-pytest): ensure hatch is always installed when needed

Psycojoker:
Sometime in test-using-pytest CI workflow the previous job failed, for example
[here](https://github.com/aleph-im/aleph-vm/actions/runs/10687238972/job/29624379941?pr=689),
and this line with `if: always()` won't be able to run because hatch is
installed in the previous step but this step might get skipped.

## Changes

Always install hatch in this `always` step of the job.

## How to test

Make a PR that will fail this job, for example by not respecting black or use
[act](https://github.com/nektos/act) to test that locally.

### DIFF
--- a/.github/workflows/test-using-pytest.yml
+++ b/.github/workflows/test-using-pytest.yml
@@ -72,16 +72,19 @@ jobs:
       - name: "Build example volume"
         run: |
           cd examples/volumes && bash build_squashfs.sh
-        
+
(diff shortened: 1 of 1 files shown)


Psycojoker:
@hoh this is related to your #689 PR

hoh (approved):

PR #700, opened 2024-09-18
Problem: Solana wallet  couln't be used to control the VM

olethanh:
Add Solana authentication method for the VM operator endpoint.

First version on how to implement this. It is the same method as with Ethereum using an ephemeral key, but that key is signed with Solana private key instead of Ethereum.

On the  SignedPubKeyPayload a  `chain` argument that can be either `ETH` or `SOL` so we know which chain signing method to validate against. It default to ETH as to remain backward compatible with current client implementations.

Also added a docstring on the `require_jwk_authentication` decorator. 

Related Jira tickets : ALEPH-50

## How to test

Not really much to test as the client part (Python/ web) is not developed yet. So you will have to rely on the tests.

## Print screen / video

N/A

## Notes

This add the new dependency Solathon to sign and validate the signature via a Solana wallet, it itself use the nacl.sign module (which is a binding on top of libsodium). An alternative would be to use the Solders lib which is running in rust. The API we use is quite small. So it shouldn't be much trouble to port from one to the other.

### DIFF
--- a/src/aleph/vm/orchestrator/views/authentication.py
+++ b/src/aleph/vm/orchestrator/views/authentication.py
@@ -1,6 +1,6 @@
 """Functions for authentications
 
-See /doc/operator_auth.md for the explaination of how the operator authentication works.
+See /doc/operator_auth.md for the explanation of how the operator authentication works.
 
 Can be enabled on an endpoint using the @require_jwk_authentication decorator
 """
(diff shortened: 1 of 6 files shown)


src/aleph/vm/orchestrator/views/authentication.py
```diff
@@ -55,6 +56,7 @@ class SignedPubKeyPayload(BaseModel):
     # alg: Literal["ECDSA"]
     address: str
     expires: str
+    chain: Literal["ETH"] | Literal["SOL"] = "ETH"
```
hoh, line 59: Should this use an Enum ? For example `aleph_message.models.Chain.ETH` and ``aleph_message.models.Chain.SOL` ?

src/aleph/vm/orchestrator/views/authentication.py
```diff
@@ -89,12 +91,23 @@ def check_expiry(cls, values) -> dict[str, bytes]:
...
         """Check that the signature is valid"""
-        signature: bytes = values["signature"]
+        signature: list = values["signature"]
         payload: bytes = values["payload"]
         content = SignedPubKeyPayload.parse_raw(payload)
-        if not verify_wallet_signature(signature, payload.hex(), content.address):
-            msg = "Invalid signature"
-            raise ValueError(msg)
+        if content.chain == "SOL":
+            from solathon.utils import verify_signature
```
hoh, line 98: Is it the best practice to import `solathon` here ? What about trying to import it in the imports section else display a warning ? If this is a required feature, should a normal import be used instead ?

hoh (commented):
Looks good overall !

Small comments regarding the use of Enums and best practices for optional imports.

The dependencies pulling `qrcode` and `pillow` seem a bit overkill but well :roll_eyes: ...

olethanh:
Thanks for the review I will make the correction.  

I can try porting using Solder if you want, should have less deps.

src/aleph/vm/orchestrator/views/authentication.py
olethanh: Correct, was a oversight

src/aleph/vm/orchestrator/views/authentication.py
```diff
@@ -89,12 +98,23 @@ def check_expiry(cls, values) -> dict[str, bytes]:
...
+
+        if content.chain == Chain.SOL:
+
+            try:
+                verify_signature(content.address, signature, payload.hex())
+            except BadSignatureError:
+                msg = "Invalid signature"
+                raise ValueError(msg)
+        elif content.chain == "ETH":
+            if not verify_wallet_signature(signature, payload.hex(), content.address):
```
nesitor, line 113: Maybe unify both methods for SOL and ETH into one unique method?
Also abstracting this common method and use one or another implementation depending on the chain will be a bit cleaner and easy to integrate with next chains.

olethanh: done

nesitor (approved):

PR #701, opened 2024-09-18
Fix mypy error with dataclasses

olethanh:
Had do ignore since mypy expect a DataClassIstance that don't exist

### DIFF
--- a/src/aleph/vm/utils/__init__.py
+++ b/src/aleph/vm/utils/__init__.py
@@ -75,7 +75,7 @@ def to_json(o: Any):
     elif hasattr(o, "dict"):  # Pydantic
         return o.dict()
     elif is_dataclass(o):
-        return dataclass_as_dict(o)
+        return dataclass_as_dict(o)  # type: ignore
     else:
         return str(o)
 


src/aleph/vm/utils/__init__.py
```diff
@@ -75,7 +75,7 @@ def to_json(o: Any):
     elif hasattr(o, "dict"):  # Pydantic
         return o.dict()
     elif is_dataclass(o):
-        return dataclass_as_dict(o)
+        return dataclass_as_dict(o)  # type: ignore
```
hoh, line 78: What is the error and isn't there a more elegant way to solve the issue ?
`# type: ignore` looks a bit like "force-ignore-this".

Psycojoker: And shouldn't the return type be defined for this function?

olethanh: @hoh That's explained in the PR description why I had to force ignore.

error was
```
src/aleph/vm/utils/__init__.py:78: error: No overload variant of "asdict" matches argument type "type[DataclassInstance]"  [call-overload]
src/aleph/vm/utils/__init__.py:78: note: Possible overload variants:
src/aleph/vm/utils/__init__.py:78: note:     def asdict(obj: DataclassInstance) -> dict[str, Any]
src/aleph/vm/utils/__init__.py:78: note:     def [_T] asdict(obj: DataclassInstance, *, dict_factory: Callable[[list[tuple[str, Any]]], _T]) -> _T
Found 1 error in 1 file (checked 71 source files)
```
DataclassInstance is not something that exist, so for me it looks like a mypy bug.  But if you have another fix I would be happy to accept it.

@Psycojoker which function?

Psycojoker: @olethanh github doesn't allow me to comment on any line but the exact function you are editing

image

olethanh: yeah it probably should but it's totally outside the scope of this PR

Psycojoker: Ended up doing it here https://github.com/aleph-im/aleph-vm/pull/704

Psycojoker:
I'm a bit confused, I can't reproduce it locally, is it using "hatch run lint:typing"? Or with a specific version of mypy or on another code base or PR?

src/aleph/vm/utils/__init__.py
hoh: I merged #704

olethanh:
It was one the solana branch, which added the dep : `solathon==1.0.2`

PR #702, opened 2024-09-19
Problem: error org.freedesktop.systemd1.NoSuchUnit: Unit aleph-vm-con…

olethanh:
Problem: error org.freedesktop.systemd1.NoSuchUnit: Unit aleph-vm-controller@...

Jira issue: ALEPH-113

This happened because we tried to check if the service was running before checking if it was enabled. It resulted in no proper error but was creating an error message in the logs

Solution: Check if service is enabled before checking if it is active

How to test:
Launch instances. Restart the aleph-vm process, stop them. That error message should not display

### DIFF
--- a/src/aleph/vm/systemd.py
+++ b/src/aleph/vm/systemd.py
@@ -60,7 +60,10 @@ class SystemDManager:
 
     def is_service_active(self, service: str) -> bool:
         try:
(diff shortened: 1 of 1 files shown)


hoh (approved):

hoh:
Would it make sense to invest time adding tests on this ?

olethanh:
For this bug in particular no, it was just a message with no effect visible for the user. For the whole systemd subsystem absolutely yes.

PR #713, opened 2024-10-28
Problem: IGNORE_TRACEBACK_FROM_DIAGNOSTICS broken

olethanh:
Symptom:
The CustomError from the diagnostics VM was printed even if if IGNORE_TRACEBACK_FROM_DIAGNOSTICS was set to True (the default)

Analysis:
This was caused by the refactoring of the fastapi_example/main.py file done in fe9235ac658915eea20d5371ae45cedabe1f7b17
Which changed the output used to detect the error to catch

Solution:
Fix detection string


Psycojoker (approved):
I'm confused about why the builds fails since this seems totally unrelated to this PR (rust error for some reason) but otherwise LGTM

PR #721, opened 2024-11-11
716 updating pydantic version

Antonyjin:
Pydantic need to be updated to v2

Jira tickets : ALEPH-150

## Changes

Upgrading to pydantic 2.
- Removing every deprecated functions
- Adapt the parsing with pydantic 2 which is more strict (model_dump / model_dump_json)
- Modify the usage of dotenv to adapt with pydantic 2


src/aleph/vm/storage.py
```diff
@@ -132,7 +132,7 @@ async def get_latest_amend(item_hash: str) -> str:
     if settings.FAKE_DATA_PROGRAM:
         return item_hash
     else:
-        url = f"{settings.CONNECTOR_URL}/compute/latest_amend/{item_hash}"
+        url = f"{settings.CONNECTOR_URL}compute/latest_amend/{item_hash}"
```
olethanh, line 135: would be nice to have something in the settings that add that / if it doesn't exist

olethanh (commented):
Did a quick read of the code and it look mostly good to me. But of course it will need quite a bit more real testing since it a big infra change

src/aleph/vm/storage.py
Antonyjin: Yeah, I could add a validator for this method with something like this

```py
@validator("CONNECTOR_URL", pre=True)
    def ensure_trailing_slash(cls, value: str) -> str:
        if not value.endswith("/"):
            return f"{value}/"
        return value
```

This would add automatically a slash if missing after calling `settings.CONNECTOR_URL`

Antonyjin: Yep can do a function to do that but I also checked the documentation of Pydantic 2, the HttpUrl ensures that the argument given is well formated.
That way it will automatically add an extra slash if needed.

So is this really necessary to create a function knowing that pydantic will add automatically a slash?

olethanh: Didn't know pydantic did it but if that's the case no it is not needed

PR #735, opened 2024-12-20
Problem Deleted Instance not stopping - monitor_payment

olethanh:
Make monitory_payment never stop

= Problem
Reported by user “Roby” on Telegram with his own CRN and VMs. PAYG Instance he had deleted from the web frontend never stopped running.

= Analysis
After some investigation and dicussion with the user, the probable cause is that the monitor payment task has crashed and stopped. Thus VM didn’t get stopped anymore this way. (when forgotten or failing payment)

Upon investigation there is no error handling over the whole task thus in case of uncatched exception the whole task stopped.

= Solution
Wrap the monitor_payment task around a try_catch so it never stop running


philogicae (approved):
LGTM

PR #736, opened 2025-01-08
Fix VM logs websocket endpoint getting stuck

olethanh:
Jira ticket: ALEPH-316

At some point the websocket endpoint stopped sending log entries to client

## Original report from Arnaud:
"aleph instance logs" output gets stuck
VM started and is accessible, but logs seem frozen when querying them via CLI.

## Analysis:
The log entries are in journald and retrievable but they don't get put at on the internal queue. My guess is that there is a lock bug between the journal polling and the queue max length, probably that if the queue is full and get emptied, the polling doesn't reset unless new message arrive in journald.

## Solution:
Rework the journald to Queue implementation.

## Note:
I have tried a lot of combinaison and variation of the logic and rewrote that code block multiple times, this version works as far as I can tell.

The interaction between asyncio select handling, journald fd reset and async task is complicated and confusing.

Other versions broke in some way or others, sometimes subtle. For examples:
- Past logs working but new log entries not getting consumed
- Queue (re)filling starting only when a new entrie was created
- infinite tasks always popping up
- multiple task consuming the same fd
- Polling never resetting and consumer being called in a loop

Some of theses issues get hidden by the TCP buffer on the websocket and only popup after some time or depending on the network or how the client interact.

So beware if you try to rewrite that logic.
Also DO NOT TRUST ChatGPT or Deepseek on this, they will produce nice looking code that do not works properly.

## To test:
Start an instance or a program on your dev CRN. Fetch the logs using `aleph instance logs` with the --domain option.

For testing you can insert additional log entries using the systemd-cat command
 ```bash
echo 1 $(date)|systemd-cat -t vm-63faf8b5db1cf8d965e6a464a0cb8062af8e7df131729e48738342d956f29ace-stdout
```
or

```bash
for ((i=1; i<=400; i++));
do
 echo $i
 echo $i |systemd-cat -t vm-63faf8b5db1cf8d965e6a464a0cb8062af8e7df131729e48738342d956f29ace-stdout
 sleep 1
done
```

Log server side can be  checked using
```
journalctl -t vm-63faf8b5db1cf8d965e6a464a0cb8062af8e7df131729e48738342d956f29ace-stderr -t vm-63faf8b5db1cf8d965e6a464a0cb8062af8e7df131729e48738342d956f29ace-stdout -f
```

For reproducing or debugging issues adding an asyncio.sleep(0.2) inside the websocket handling code
usually helps, most inner loop of `stream_logs` method inside src/aleph/vm/orchestrator/views/operator.py

### DIFF
--- a/src/aleph/vm/utils/logs.py
+++ b/src/aleph/vm/utils/logs.py
@@ -1,8 +1,8 @@
 import asyncio
 import logging
 from collections.abc import Callable, Generator
-from datetime import datetime
-from typing import TypedDict
+from datetime import datetime, timedelta
+from typing import List, TypedDict
 
 from systemd import journal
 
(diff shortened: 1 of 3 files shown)


src/aleph/vm/utils/logs.py
```diff
@@ -35,27 +35,38 @@ def make_logs_queue(stdout_identifier, stderr_identifier, skip_past=False) -> tu
...
-    queue: asyncio.Queue = asyncio.Queue(maxsize=1000)
+    queue: asyncio.Queue = asyncio.Queue(maxsize=5)
+    tasks: List[asyncio.Future] = []
 
-    def _ready_for_read() -> None:
-        change_type = r.process()  # reset fd status
-        if change_type != journal.APPEND:
-            return
+    async def process_messages() -> None:
+        loop.remove_reader(r.fileno())
```
Psycojoker, line 42: Can you maybe comment this function lines (and the next one) so it looks less like black magic? At least if you understand something about it and so we have some hope of maintaining it in the future v_v

Psycojoker, line 42: Can you maybe comment this function lines (and the next one) so it looks less like black magic? At least if you understand something about it and so we have some hope of maintaining it in the future v_v

hoh (approved):

Psycojoker (approved):

PR #738, opened 2025-01-14
Display proper error for message not found

olethanh:
or not being downloadable

Related ClickUp, GitHub or Jira tickets : ALEPH-353

## Problem:
When calling an program on a CRN, whose message the CRN couldn’t reach, it returned the error
```
500: Unhandled error during initialisation
```

instead of a proper explanation error

## Solution:
Catch that error. Catch also the other possible error when downloading

This also now works when failing to download a runtime or volume

## testing
Call your CRN via http with an invalid item hash

### DIFF
--- a/src/aleph/vm/orchestrator/run.py
+++ b/src/aleph/vm/orchestrator/run.py
@@ -3,7 +3,7 @@ import logging
 from typing import Any
 
 import msgpack
-from aiohttp import web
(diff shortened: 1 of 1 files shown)


Psycojoker (approved):

olethanh:
correction there was a typo this fix JIRA ALEPH-352 not ALEPH-353

PR #741, opened 2025-01-21
Fix: Runtime emitted wrong reboot signal

hoh:
The runtime init `init1.py` emitted the signal `0x4321fedc`
to the Linux kernel when shutting down.

The correct procedure [according to the Firecracker docs]
(https://github.com/firecracker-microvm/firecracker/blob/dfb45dc4213bcb1c9704435457e233d3a210dce2/docs/getting-started.md?plain=1#L298) is to issue a `reboot`.

The syscall number for a reboot on Linux is `0x1234567`.

Tested using Firecracker independently of `aleph-vm`.


philogicae (approved):
Works properly now

olethanh (approved):

nesitor (approved):

PR #746, opened 2025-02-03
Garbate collector: Free disk space from inactive VM

olethanh:
Add a script to manually  list and remove volume linked to inactive VM. It fetches data from the scheduler and pyaleph main's node as to fetch information on the status of the VM. Then display them to the user to determine if they can be removed safely.

JIRA ticket ALEPH-37

This solution is a stop gap to run manualy till we have  more confident in it and can automatically run it in the background

### DIFF
--- a/src/aleph/vm/garbage_collector.py
+++ b/src/aleph/vm/garbage_collector.py
@@ -19,6 +19,8 @@ TEST_HASHES = [
     "fake_vm_fake_vm_fake_vm_fake_vm_fake_vm_fake_vm_fake_vm_fake_vm_",
     "cafecafecafecafecafecafecafecafecafecafecafecafecafecafecafecafe",
     "decadecadecadecadecadecadecadecadecadecadecadecadecadecadecadeca",
+    "63faf8b5db1cf8d965e6a464a0cb8062af8e7df131729e48738342d956f29ace",
+    "67705389842a0a1b95eaa408b009741027964edc805997475e95c505d642edd8",
 ]
 
 api_server = [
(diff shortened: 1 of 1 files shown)


src/aleph/vm/garbage_collector.py
```diff
@@ -0,0 +1,145 @@
...
+        proc_status = "stopped"
+    else:
+        proc_status = "error"
+        print("Unknown process state", exit_code)
+    # to remove
+
+    if proc_status != "running":
+        # not running and forgotten
+
+        if message_status == "forgotten" or message_status == "rejected":
```
nesitor, line 114: In theory, a rejected message cannot be never scheduled on a CRN.

olethanh: Yes in theory but it is better to handle all cases.
Also we have seen strange thing with the network disconnects,  like one API  with the message process and the other rejected.

src/aleph/vm/garbage_collector.py
```diff
@@ -0,0 +1,145 @@
...
+import subprocess
+from pathlib import Path
+
+import requests
+
+# following hashes are used in tests or debug VM, we can ignore them.
+TEST_HASHES = [
+    "fake_vm_fake_vm_fake_vm_fake_vm_fake_vm_fake_vm_fake_vm_fake_vm_",
+    "cafecafecafecafecafecafecafecafecafecafecafecafecafecafecafecafe",
+    "decadecadecadecadecadecadecadecadecadecadecadecadecadecadecadeca",
```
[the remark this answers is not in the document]
olethanh: I will add it

PR #747, opened 2025-02-10
Implemented GPU compatibility system

nesitor:
Implemented a settings aggregate query to check the compatible GPUs that we support and also other general settings.

- [X] The new code clear, easy to read and well commented.
- [X] New code does not duplicate the functions of builtin or popular libraries.
- [X] An LLM was used to review the new code and look for simplifications.
- [X] New classes and functions contain docstrings explaining what they provide.

## Changes

Feature: Added options to get GPU compatibilities from a settings aggregate.

## How to test

Go to the `/about/usage/system` endpoint on a CRN with at least 1 GPU and check if the `compatible` field appears and also the `model` one.

### DIFF
--- a/src/aleph/vm/conf.py
+++ b/src/aleph/vm/conf.py
@@ -9,7 +9,7 @@ from enum import Enum
 from os.path import abspath, exists, isdir, isfile, join
 from pathlib import Path
 from subprocess import CalledProcessError, check_output
-from typing import Any, Literal, NewType, List
+from typing import Any, List, Literal, NewType
 
 from aleph_message.models import Chain
 from aleph_message.models.execution.environment import HypervisorType
--- a/src/aleph/vm/conf.py
+++ b/src/aleph/vm/conf.py
@@ -9,7 +9,7 @@ from enum import Enum
 from os.path import abspath, exists, isdir, isfile, join
 from pathlib import Path
 from subprocess import CalledProcessError, check_output
-from typing import Any, Literal, NewType
+from typing import Any, Literal, NewType, List
 
 from aleph_message.models import Chain
 from aleph_message.models.execution.environment import HypervisorType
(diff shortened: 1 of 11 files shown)


src/aleph/vm/conf.py
```diff
@@ -280,6 +280,11 @@ class Settings(BaseSettings):
         description="Enable GPU pass-through support to VMs, only allowed for QEmu hypervisor",
     )
 
+    # Settings to get from the network aggregates
+    SETTINGS_AGGREGATE_ADDRESS: str = "0xA07B1214bAe0D5ccAA25449C3149c0aC83658874"
+    COMMUNITY_WALLET_ADDRESS: str | None = None
+    COMPATIBLE_GPUS: List[dict[str, str]] = []
```
olethanh, line 286: The Settings object is not the correct place to save this, it cannot be set by the user and will create confusion
It should be placed in a separate singleton.

src/aleph/vm/orchestrator/resources.py
```diff
@@ -134,6 +135,8 @@ async def about_system_usage(request: web.Request):
     """Public endpoint to expose information about the system usage."""
     period_start = datetime.now(timezone.utc).replace(second=0, microsecond=0)
     machine_properties = get_machine_properties()
+    # Refresh and get latest settings aggregate
+    await update_aggregate_settings()
```
olethanh, line 139: instead of refreshing the aggregate every time this page , we should be a timeout period and only refresh if the data is older than that period.

olethanh:
The  More information is strangely aligned with the gpu list

image

src/aleph/vm/orchestrator/resources.py
nesitor: I moved the code to service initialization, so now only executes it at boot time.

nesitor:
> The More information is strangely aligned with the gpu list
> 
> image

I have already fixed the CSS for Chrome and Firefox and now it shows well.

gpu screenshot

PR #776, opened 2025-03-11
Gpu Reservation system

olethanh:
Implement a resource (GPU) reservation system. 
Please do not squash when merging.

This new feature prevent resources being taken between the moment the user initiate the payment flow and the moment the VM start.

Jira tickets : ALEPH-421

## Changes

- Add a new endpoint `/control/reserve_resource` which reserve the required resource to start the VM for the user. At the moment it only reserve GPU.
- Added `reservations` attributes on `VmPool` that handle the list of reservations.
- Check and free the reservation when staring a new VM. 
- When inspecting the earlier code, I noticed that if no GPU were available the  VM probably started without the GPU, now the VM doesn't start and raise an error
- pool.setup() is now async, to allow working in the unit tests.

## How to test

Need a CRN with GPU. I tested  principally via the unit tests.

## Notes

I added a mock_app_with_pool fixture that you might find useful for your future tests, it create a server and a VmPool  (with a GPU inside)

### DIFF
--- a/src/aleph/vm/pool.py
+++ b/src/aleph/vm/pool.py
@@ -73,7 +73,7 @@ class VmPool:
         if settings.SNAPSHOT_FREQUENCY > 0:
             self.snapshot_manager = SnapshotManager()
 
-    def setup(self) -> None:
+    async def setup(self) -> None:
         """Set up the VM pool and the network."""
         if self.network:
             self.network.setup()
@@ -81,10 +81,9 @@ class VmPool:
         if self.snapshot_manager:
             logger.debug("Initializing SnapshotManager ...")
             self.snapshot_manager.run_in_thread()
-
         if settings.ENABLE_GPU_SUPPORT:
             # Refresh and get latest settings aggregate
-            asyncio.run(update_aggregate_settings())
+            await update_aggregate_settings()
             logger.debug("Detecting GPU devices ...")
             self.gpus = get_gpu_devices()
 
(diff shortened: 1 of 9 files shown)


src/aleph/vm/resources.py
```diff
@@ -71,6 +72,8 @@ def get_gpu_model(device_id: str) -> bool | None:
 def is_gpu_compatible(device_id: str) -> bool:
     """Checks if a GPU is compatible based on vendor and model IDs."""
     compatible_gpu_set = {gpu["device_id"] for gpu in get_compatible_gpus()}
+    logger = logging.getLogger(__name__)
+    logger.info(get_compatible_gpus())
```
nesitor, line 76:
```suggestion
    logger = logging.getLogger(__name__)
    logger.debug(get_compatible_gpus())
```
I think this log output can be moved to debug instead info

src/aleph/vm/pool.py
```diff
@@ -340,3 +350,66 @@ def get_executions_by_sender(self, payment_type: PaymentType) -> dict[str, dict[
                 executions_by_sender.setdefault(sender, {})
                 executions_by_sender[sender].setdefault(chain, []).append(execution)
         return executions_by_sender
+
+    class Reservation:
```
nesitor, line 354: I think that we can move that class outside the `VMPool` class as it's more generic

src/aleph/vm/resources.py
olethanh: you are right, I had put that there because I had an issue with my debuger setup at the time. I remove the line completely

nesitor (approved):

PR #778, opened 2025-03-11
Hoh upgrade kubo

hoh:
This bumps their versions to
- Kubo 0.23.0 -> 0.33.2

The list of changes regarding Kubo too large to be
 mentioned here, but I mostly expect performance
 improvements as the main API has not changed much.


hoh:
Do not squash ;-)

olethanh (approved):

PR #793, opened 2025-04-10
Retry instances not running add new Execution list endpoint

olethanh:
Jira ticket: ALEPH-287

This PR improve how Instance are started, making it more reliable and properly cleaning when they fail to start. It also allow to start again the VM if it failed to start the first time.

Changes introduced by this PR:
1. Decouple instance wait_for_init from allocation. It now return directly after starting the VM controller and run wait_for_init in the background. wait_for_init has been renamed wait_for_boot in VM
2. Stop the controller when the Instance fail to start. Before if the instance failed to respond to ping the Instance was reported as not running and removed from the list but the Instance systemd controller was still. This tied up resource and prevented the VM from starting again. Also it caused issue with the network. 
3. Introduce a new endpoint `/v2/about/executions/list` which also include the starting, started, end times etc.. so we can inspect the VM state. and includ non running VM It will allow us in the future to improve the clients
4. Fix VM inconsistant state calculation between the boot and the end of boot.  
5. Properly clean up ressource if the VM crash and we tried to start it again. Earlier if it was not is_running we just started a new one regardless if it was shutting down or crashed and the resource were not cleaned.
6. Save debug token inside a protected file so we can still access it if the log rotated  
7. Improve the instances tests:
     - Reduce the problem of contamination of settings between tests
     - ensure we tests in a separat temp dir  we don't reuse the cache. Before that the tests missed issue where ressource could not be downloaded because of a coding error but the resource was already cached
     - Better output and debug info

## Changes
In addition to what was listed above, others refactor changes:
* in tests Setup webapp take the vm_pool argument
* vm_pool don't take the loop anymore (not needed since Python 3.10)
* Execution call .enable_and_start and .stop_and_disable, not VmPool
* .enable_and_start is now async
* Remove some unused code
* Better debugging output

## How to test

Allocate VM, try allocating VM that fail to start and reallocate again. Kill VM during boot

### DIFF
--- a/src/aleph/vm/controllers/firecracker/instance.py
+++ b/src/aleph/vm/controllers/firecracker/instance.py
@@ -21,13 +21,7 @@ from aleph.vm.hypervisors.firecracker.config import (
 from aleph.vm.hypervisors.firecracker.microvm import setfacl
 from aleph.vm.network.interfaces import TapInterface
 from aleph.vm.storage import create_devmapper, create_volume_file
-from aleph.vm.utils import (
-    HostNotFoundError,
-    NotEnoughDiskSpaceError,
-    check_disk_space,
-    ping,
-    run_in_subprocess,
-)
+from aleph.vm.utils import NotEnoughDiskSpaceError, check_disk_space, run_in_subprocess
 
 from .executable import (
     AlephFirecrackerExecutable,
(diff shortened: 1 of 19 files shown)


olethanh:
I will rework the commit cleanly and update the desc but for me this is ready for review

nesitor (approved):

PR #816, opened 2025-06-18
Ol aleph 565 ipv4 domain support

olethanh:

## Changes

## How to test

Explain how to test your PR.

## Print screen / video

## Notes

### DIFF
--- a/packaging/aleph-vm/etc/haproxy/https_domains.map
+++ b/packaging/aleph-vm/etc/haproxy/https_domains.map
@@ -1,0 +1,2 @@
+dualstack-to-instance.aleph.sh 172.16.6.2:443
+vm-lab.aleph.sh 127.0.0.1:4443
--- a/src/aleph/vm/haproxy.py
+++ b/src/aleph/vm/haproxy.py
@@ -305,16 +305,17 @@ async def fetch_list_() -> list[dict]:
             return []
         return instances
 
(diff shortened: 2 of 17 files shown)


src/aleph/vm/haproxy.py
```diff
@@ -0,0 +1,335 @@
...
+    async with aiohttp.ClientSession() as client:
+        resp = await client.get(url=str(settings.DOMAIN_SERVICE_URL))
+        resp.raise_for_status()
+        instances = await resp.json()
+        if len(instances) == 0:
+            return []
+        return instances
+
+
+async def fetch_list_and_update2(socket_path, local_vms: list[str], force_update):
```
nesitor, line 296: Why this method is called `fetch_list_and_update2`, I think should be called `fetch_list_and_update`

packaging/aleph-vm/etc/haproxy/https_domains.map
```diff
@@ -0,0 +1,2 @@
+dualstack-to-instance.aleph.sh 172.16.6.2:443
+vm-lab.aleph.sh 127.0.0.1:4443
```
nesitor, line 2: I think this line should be removed and should be the user domain name.

nesitor (commented):
Looks good overall except the 2 comments I added, but I think that we should include also tests for the new class added checking that creates the mapping files well.

nesitor (approved):

PR #819, opened 2025-07-11
fix GH-814 gradual increase in waiting time

aliel:
fix GH-814 gradual increase in waiting time


nesitor (approved):

nesitor (commented):
The CI tests are failing after that change, it's needed to fix them before merging it.

PR #905, opened 2026-03-13
WIP: cold migration

odesenfans:

## Changes

## How to test

Explain how to test your PR.

## Print screen / video

## Notes

### DIFF
--- a/src/aleph/vm/controllers/configuration.py
+++ b/src/aleph/vm/controllers/configuration.py
@@ -73,9 +73,30 @@ class Configuration(BaseModel):
     hypervisor: HypervisorType = HypervisorType.firecracker
 
 
+def get_controller_configuration_path(vm_hash: str) -> Path:
+    """Get the path to the controller configuration file for a VM."""
+    return Path(f"{settings.EXECUTION_ROOT}/{vm_hash}-controller.json")
+
+
+def load_controller_configuration(vm_hash: str) -> Configuration | None:
+    """Load VM configuration from the controller service configuration file.
+
+    :param vm_hash: The VM hash identifying the configuration file
+    :return: The Configuration object, or None if the file doesn't exist
+    """
+    config_file_path = get_controller_configuration_path(vm_hash)
+
+    if not config_file_path.exists():
+        logger.warning(f"Controller configuration file not found for {vm_hash}")
+        return None
+
+    with config_file_path.open("r") as f:
+        return Configuration.model_validate_json(f.read())
+
+
 def save_controller_configuration(vm_hash: str, configuration: Configuration) -> Path:
     """Save VM configuration to be used by the controller service"""
-    config_file_path = Path(f"{settings.EXECUTION_ROOT}/{vm_hash}-controller.json")
+    config_file_path = get_controller_configuration_path(vm_hash)
     with config_file_path.open("w") as controller_config_file:
         controller_config_file.write(
             configuration.model_dump_json(
--- a/src/aleph/vm/controllers/firecracker/executable.py
+++ b/src/aleph/vm/controllers/firecracker/executable.py
@@ -310,7 +310,7 @@ class AlephFirecrackerExecutable(Generic[ConfigurationType], AlephVmControllerIn
         return
 
     async def configure(self):
-        """Configure the VM by saving controller service configuration"""
+        """Configure the VM by saving controller service configuration."""
         if self.persistent:
             firecracker_config_path = await self.fvm.save_configuration_file(self._firecracker_config)
             vm_configuration = VMConfiguration(
--- a/src/aleph/vm/controllers/interface.py
+++ b/src/aleph/vm/controllers/interface.py
@@ -73,7 +73,7 @@ class AlephVmControllerInterface(ABC):
         pass
 
     async def configure(self) -> None:
-        """Configuration done after the VM process is started"""
+        """Configuration done after the VM process is started."""
         raise NotImplementedError()
 
     async def load_configuration(self) -> None:
--- a/src/aleph/vm/controllers/qemu/client.py
+++ b/src/aleph/vm/controllers/qemu/client.py
@@ -1,6 +1,7 @@
 import asyncio
 import json
 import logging
+from enum import Enum
 from pathlib import Path
 from typing import cast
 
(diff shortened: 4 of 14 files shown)


foxpatch-aleph (changes requested):
This PR adds cold migration functionality for QEMU VMs. The implementation is comprehensive with good test coverage, but has a security vulnerability (path traversal in download endpoint), missing migration state update on destination, and some code quality issues that should be addressed before merging.

**`src/aleph/vm/orchestrator/views/migration.py`** (line 419): **SECURITY ISSUE: Path Traversal Vulnerability**

The `filename` parameter comes directly from the URL path without validation. An attacker could use `../` sequences to access files outside the intended directory.

**Fix:** Add path validation:
```python
# Prevent path traversal
if ".." in filename or filename.startswith("/"):
    return web.HTTPBadRequest(text="Invalid filename")

# Ensure the resolved path is within volumes_dir
export_path = (volumes_dir / f"{filename}.export.qcow2").resolve()
if not str(export_path).startswith(str(volumes_dir.resolve())):
    return web.HTTPBadRequest(text="Invalid filename")
```

**`src/aleph/vm/orchestrator/views/migration.py`** (line 501): **Missing Migration State Update**

After successful import, the destination VM's `migration_state` is never set to `MigrationState.COMPLETED`. The state remains at `NONE` (default) or `FAILED` if there was a previous attempt. This makes it impossible to track migration history.

**Fix:** After `pool.create_a_vm()`, set:
```python
if execution:
    execution.migration_state = MigrationState.COMPLETED
    execution.export_token = None
```

**`src/aleph/vm/orchestrator/views/migration.py`** (line 267): **Potential Race Condition**

The QMP connection is closed immediately after sending `system_powerdown`, but the loop continues checking if the service is active. If the QMP socket is deleted during shutdown, subsequent operations might fail. Consider keeping the connection alive during the wait or handling connection errors gracefully.

**`src/aleph/vm/orchestrator/views/migration.py`** (line 238): **Global Mutable State**

`_export_state` and `_export_cleanup_tasks` are module-level globals. In production with multiple processes or workers, this could cause issues. Consider using a proper cache with TTL support or a database-backed state store.

**`src/aleph/vm/orchestrator/views/migration.py`** (line 189): **Missing Import State Tracking**

The `MigrationState.IMPORTING` state is defined but never used. The import endpoint should set `execution.migration_state = MigrationState.IMPORTING` at the start and handle state transitions on success/failure.

**`src/aleph/vm/hypervisors/qemu/qemuvm.py`** (line 150): **Breaking Change Warning**

Adding `-machine pc-i440fx-6.2` hardcoded to all non-GPU VMs could break existing VMs that were created with different machine types. Consider:
1. Only applying this to new VMs
2. Making it configurable
3. Adding migration compatibility checks before applying

**`src/aleph/vm/controllers/qemu/cloudinit.py`** (line 42): **Air-gapped Environment Concern**

Adding `package_update: True` to cloud-init config could cause issues in air-gapped environments where package updates aren't available. Consider making this configurable or conditional.

**`src/aleph/vm/pool.py`** (line 163): **Logic Simplification Needed**

The condition `if current_execution.is_running and not current_execution.is_stopping` should also check for migration state. A VM being exported should not be returned as 'running':
```python
if (current_execution.is_running and 
    not current_execution.is_stopping and
    current_execution.migration_state == MigrationState.NONE):
```

**`src/aleph/vm/orchestrator/views/migration.py`** (line 123): **Missing Type Hint**

The `VmStatus.is_running` property should have an explicit return type hint for consistency with other properties.

**`tests/supervisor/views/test_migration.py`** (line 1): **Missing Test Coverage**

Consider adding tests for:
1. Path traversal attack on download endpoint
2. Import failure after partial download (cleanup verification)
3. Concurrent migration attempts on same VM
4. TTL cleanup task execution
5. Migration state transitions (IMPORTING -> COMPLETED/FAILED)

foxpatch-aleph (commented):
The cold migration implementation is well-structured with good test coverage. The migration flow (export -> download -> import -> cleanup) is logical and the state machine is clean. However, there are several issues worth addressing: a missing `os` import that is unused, a potential race condition in the `migration_lock` lazy initialization, the `_export_state` global mutable dict lacks synchronization, and the cloud-init `install_guest_agent` logic has a bug where `packages` and `runcmd` overwrite each other on repeated calls. These are mostly code quality and edge-case issues rather than blockers.

**`src/aleph/vm/orchestrator/views/migration.py`** (line 13): Unused import: `os` is imported but never used in this file. Remove it.

**`src/aleph/vm/orchestrator/views/migration.py`** (line 41): Race condition: `migration_lock` is lazily initialized on line 429-431 inside the endpoint. Between the `None` check and the assignment, two concurrent requests could each create a different Lock instance. Initialize at module level: `migration_lock = asyncio.Lock()`.

**`src/aleph/vm/orchestrator/views/migration.py`** (line 47): The `_export_state` global dict is accessed from multiple async contexts (the download endpoint and the TTL cleanup task) without any synchronization. While CPython's GIL makes simple dict operations atomic in practice, this is fragile. Consider using a lock for mutations, or at least add a comment noting this assumption.

**`src/aleph/vm/orchestrator/views/migration.py`** (line 144): TTL cleanup task references `_export_cleanup_tasks` by name in its body, but if the module is reloaded or imported in a way that creates a new module object, this could break. This is a minor concern but worth noting.

**`src/aleph/vm/orchestrator/views/migration.py`** (line 310): The `namespace` variable on line 309 is assigned but the same value is accessed via `execution.vm_hash` throughout. Consider using `namespace` consistently or removing the redundant assignment.

**`src/aleph/vm/orchestrator/views/migration.py`** (line 314): If `volumes_dir` doesn't exist (no volumes attached), the code falls through to the 'No disk files found' error. This is technically correct but could be confusing — a VM with zero volumes would get a 500 error. Consider returning a 400 with a clearer message.

**`src/aleph/vm/orchestrator/views/migration.py`** (line 396): Path traversal risk: `filename` from `request.match_info` is used directly to construct `export_path` on line 408 without sanitization. An attacker could craft a URL like `/disk/../../../etc/passwd` to read arbitrary files. Add path validation: `if Path(filename).is_absolute() or '..' in filename: return HTTPBadRequest`.

**`src/aleph/vm/orchestrator/views/migration.py`** (line 487): The scheme selection logic (`https` for port 443, `http` otherwise) is fragile. A non-standard HTTPS port would fall back to insecure HTTP. Consider accepting a `source_protocol` field in the request, or at minimum validate that non-443 ports are expected to be HTTP.

**`src/aleph/vm/orchestrator/views/migration.py`** (line 512): After `pool.create_a_vm()` succeeds, the code doesn't set `execution.migration_state = MigrationState.COMPLETED`. The state remains at whatever the source host set it to. Consider updating the state on the destination host for observability.

**`src/aleph/vm/orchestrator/views/migration.py`** (line 536): Error cleanup only removes `downloaded_files` but not the `dest_dir` directory itself. If the import fails after creating the directory, stale empty directories accumulate under `PERSISTENT_VOLUMES_DIR`.

**`src/aleph/vm/hypervisors/qemu/qemuvm.py`** (line 95): The existing code sets `machine_type = 'q35' if self.gpus else 'pc'` at line 95. Then at line 180-181, for non-GPU VMs, it overrides with `-machine pc-i440fx-6.2`. This means the `-machine pc` from line 95-99 is immediately overwritten. The first `-machine` flag is what QEMU uses, so the later one takes precedence. This works but is confusing — consider removing the initial `-machine pc` for non-GPU VMs or restructure the logic.

**`src/aleph/vm/hypervisors/qemu/qemuvm.py`** (line 107): Added `file.locking=off` to the root disk drive. This is appropriate for migration (prevents lock conflicts when the same image exists on multiple hosts), but note that this disables all file-level locking. If VMs could ever run concurrently on the same host, this would risk corruption. Since cold migration is stop/start, this is safe.

**`src/aleph/vm/controllers/qemu/cloudinit.py`** (line 89): Bug: `config['packages']` and `config['runcmd']` are set unconditionally when `install_guest_agent=True`. If this function is called multiple times (e.g., for different VM types), the second call would overwrite the first. Since `install_guest_agent` defaults to `True`, all non-confidential QEMU VMs will get qemu-guest-agent installed. This is the intended behavior, but the conditional `if install_guest_agent` means confidential VMs (which pass `install_guest_agent=False`) won't get it — which is correct since confidential VMs already handle this differently.

**`src/aleph/vm/controllers/qemu/client.py`** (line 157): The `QemuVmClient` constructor calls `client.connect()` synchronously. If the connect call blocks or hangs, it will block the calling coroutine. Consider making this async or adding a timeout.

**`src/aleph/vm/pool.py`** (line 328): The `create_a_vm` change from `self.get_running_vm(vm_hash)` to `self.executions.get(vm_hash)` is a behavior change. If `get_running_vm` had additional logic (e.g., checking `is_running` flag), this could change behavior. Verify that `get_running_vm` was only doing a dict lookup.

**`tests/supervisor/views/test_migration.py`** (line 1): Missing test: no test for the `migration_export` error path that attempts to restart the VM after a failed export. Also missing: test for `test_export_no_volumes_dir` (when volumes_dir doesn't exist at all). Consider adding these for completeness.

**`tests/supervisor/test_qemu_client.py`** (line 1): Good test coverage for VmStatus and VmRunStatus. Consider adding a test for `VmStatus.is_error` with `VmRunStatus.PAUSED` to confirm it returns False (negative assertion).

odesenfans:
Superseded by #936.

PR #924, opened 2026-03-19
Fix cloud-init ISO not visible to guest on Q35 machine type

aliel:
On Q35, -cdrom creates an unattached IDE device that the guest
  cannot see, so cloud-init never runs and the VM gets no network.

  Use virtio drive for the cloud-init ISO and explicit virtio-net-pci
  device syntax, both compatible with i440FX and Q35.

### DIFF
--- a/src/aleph/vm/hypervisors/qemu/qemuvm.py
+++ b/src/aleph/vm/hypervisors/qemu/qemuvm.py
@@ -133,11 +133,24 @@ class QemuVM:
             # "-snapshot",  # Do not save anything to disk
         ]
         if self.interface_name:
-            # script=no, downscript=no tell qemu not to try to set up the network itself
-            args += ["-net", "nic,model=virtio", "-net", f"tap,ifname={self.interface_name},script=no,downscript=no"]
+            # Use explicit device + netdev syntax for Q35 compatibility.
+            # The legacy -net syntax causes NIC renaming (eth0 -> enp0s1)
+            # on Q35 which breaks cloud-init network config.
+            args += [
+                "-device",
+                "virtio-net-pci,netdev=net0",
+                "-netdev",
+                f"tap,id=net0,ifname={self.interface_name},script=no,downscript=no",
+            ]
 
         if self.cloud_init_drive_path:
-            args += ["-cdrom", f"{self.cloud_init_drive_path}"]
+            # Use explicit drive syntax instead of -cdrom for Q35
+            # compatibility. On Q35, -cdrom creates an unattached IDE
+            # device that the guest can't see.
+            args += [
+                "-drive",
+                f"file={self.cloud_init_drive_path},media=cdrom,readonly=on,if=virtio",
+            ]
 
         args += self._get_host_volumes_args()
         args += self._get_gpu_args()
--- a/src/aleph/vm/hypervisors/qemu_confidential/qemuvm.py
+++ b/src/aleph/vm/hypervisors/qemu_confidential/qemuvm.py
@@ -129,11 +129,16 @@ class QemuConfidentialVM(QemuVM):
             # "-snapshot",  # Do not save anything to disk
         ]
         if self.interface_name:
-            # script=no, downscript=no tell qemu not to try to set up the network itself
-            args += ["-net", "nic,model=virtio", "-net", f"tap,ifname={self.interface_name},script=no,downscript=no"]
+            args += [
+                "-device", "virtio-net-pci,netdev=net0",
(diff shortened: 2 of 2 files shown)


foxpatch-aleph (approved):
This PR correctly fixes the Q35 machine type compatibility issue for cloud-init ISO and network device configuration. The changes use the modern `-drive` syntax with `if=virtio` for CD-ROM devices and the explicit `-device`/`-netdev` syntax for network interfaces, both of which are compatible with both Q35 and i440FX (pc) machine types. The fix is minimal, well-commented, and addresses the root cause of the issue. No security vulnerabilities are introduced - the same string interpolation pattern was already used for `image_path` and other configuration values.

**`src/aleph/vm/hypervisors/qemu_confidential/qemuvm.py`** (line 131): The confidential VM implementation still uses the legacy `-net` syntax and `-cdrom` at lines 131-136. While this likely works (since it doesn't explicitly use Q35), consider applying the same fix here for consistency and to avoid potential future issues if confidential VMs ever use Q35 machine type.

foxpatch-aleph (approved):
This PR correctly fixes a real bug where cloud-init ISO was invisible to guests on Q35 machine type due to unattached IDE devices. The solution using `-drive ... if=virtio` is the proper approach and works on both i440FX and Q35. The network device change to explicit `-device`/`-netdev` syntax is also correct and modern. No security concerns. Existing integration tests cover the affected code paths.

**`src/aleph/vm/hypervisors/qemu_confidential/qemuvm.py`** (line 132): Minor style inconsistency: The formatting here puts multiple arguments on the same line, while the equivalent code in qemu/qemuvm.py (lines 139-143) uses one argument per line. Consider making these consistent for readability.

**`src/aleph/vm/hypervisors/qemu/qemuvm.py`** (line 136): Good explanation in the comment. Consider also noting that the `-device virtio-net-pci` syntax is the modern/recommended QEMU approach (deprecated `-net nic` syntax).

foxpatch-aleph (approved):
This PR correctly addresses a Q35 machine type compatibility issue where the legacy `-cdrom` flag creates unattached IDE devices that guests cannot see, and the legacy `-net` syntax causes NIC renaming issues. The fix uses modern explicit device/netdev syntax which is the recommended approach for Q35. The changes are minimal, well-commented, and consistent across both standard and confidential VM implementations. The additional `q35` specification in the confidential VM machine string is necessary and correct.

**`src/aleph/vm/hypervisors/qemu_confidential/qemuvm.py`** (line 133): Minor style nit: The formatting here differs from the standard QEMU implementation (lines 139-144 in qemuvm.py) which uses multi-line formatting for readability. Consider making them consistent for maintainability.

**`src/aleph/vm/hypervisors/qemu/qemuvm.py`** (line 147): Good fix. The `-cdrom` to `-drive if=virtio` change ensures the cloud-init ISO is properly attached on Q35. The `readonly=on` parameter is correctly preserved.

**`src/aleph/vm/hypervisors/qemu_confidential/qemuvm.py`** (line 118): The explicit `q35,confidential-guest-support=sev0` is correct. Note that Q35 is now required for confidential VMs anyway (due to SEV requirements), so this change aligns with that constraint.

foxpatch-aleph (approved):
This PR correctly fixes cloud-init ISO visibility and network device configuration for Q35 machine type (required for GPU passthrough and AMD SEV confidential VMs). The changes from legacy `-cdrom`/`-net` syntax to explicit `-drive`/`-device`+`-netdev` syntax are the proper QEMU approach for Q35 compatibility. The code is clean, well-commented, and the fixes are minimal and targeted.

**`src/aleph/vm/hypervisors/qemu/qemuvm.py`** (line 137): Minor comment improvement: The comment mentions NIC renaming (eth0 → enp0s1) breaks cloud-init, but the actual fix is primarily about using explicit device syntax for Q35 compatibility. The NIC naming is a side effect. Consider clarifying that the explicit syntax ensures consistent device attachment on Q35.

**`src/aleph/vm/hypervisors/qemu_confidential/qemuvm.py`** (line 118): Good catch on explicitly specifying `q35,` prefix. Confidential VMs (AMD SEV) require Q35 for proper CPU feature passthrough. This makes the machine type explicit and consistent with the regular QEMU VM code path when GPUs are present.

**`src/aleph/vm/hypervisors/qemu/qemuvm.py`** (line 152): The `-drive` syntax with `if=virtio` is correct for Q35. Note that this means the cloud-init ISO will appear as a virtio-blk device (typically `/dev/vdb`) rather than an IDE CD-ROM (`/dev/sr0`). This is actually preferable as virtio drivers are standard in modern Linux distributions.

PR #926, opened 2026-03-25
Fix WebSocket connection to Aleph API: heartbeat, reconnect, error handling

aliel:
The WebSocket listener could silently lose connection with no
reconnect, leaving the supervisor deaf to aggregate changes (port-forwarding, domains) and message updates.

  - Add 30s heartbeat ping to detect silent disconnects
  - Fix backoff: min() not max() so delay caps at 8s instead of jumping to it
  - Reset retry delay on successful message receipt
  - Catch all exceptions in retry loop to prevent listener death
  - Replace assertions with logging for unexpected message formats
  - Log reconnect attempts, connection events, and close reasons
  - Handle CLOSE/CLOSING/CLOSED WebSocket message types

### DIFF
--- a/src/aleph/vm/orchestrator/tasks.py
+++ b/src/aleph/vm/orchestrator/tasks.py
@@ -67,20 +67,25 @@ Value = TypeVar("Value")
 COMMUNITY_STREAM_RATIO = Decimal(0.2)
 
 
-async def retry_generator(generator: AsyncIterable[Value], max_seconds: int = 8) -> AsyncIterable[Value]:
+async def retry_generator(generator: AsyncIterable[Value], max_seconds: float = 8.0) -> AsyncIterable[Value]:
     retry_delay = 0.1
     while True:
         try:
             async for value in generator:
+                # Reset delay on successful message
+                retry_delay = 0.1
                 yield value
         except RuntimeError as e:
             if "Event loop is closed" in str(e):
                 logger.debug("retry_generator exiting: event loop closed")
                 return
             raise
+        except Exception:
+            logger.exception("WebSocket generator error, reconnecting")
 
+        logger.warning("WebSocket disconnected, reconnecting in %.1fs", retry_delay)
         await asyncio.sleep(retry_delay)
-        retry_delay = max(retry_delay * 2, max_seconds)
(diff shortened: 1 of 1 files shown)


foxpatch-aleph (approved):
This PR correctly fixes the silent WebSocket disconnection issue that could leave the supervisor unable to receive aggregate changes. The key fixes are: (1) adding 30s heartbeat to detect stale connections, (2) fixing the backoff bug where `max()` was incorrectly used instead of `min()`, causing delays to jump to 8s immediately rather than capping at 8s, (3) resetting retry delay on successful message receipt for better responsiveness, (4) catching all exceptions in the retry loop to prevent listener death, (5) replacing assertions with logging for unexpected message formats, and (6) proper handling of WebSocket CLOSE/CLOSING/CLOSED message types. The changes are well-tested in practice (the existing codebase has extensive usage of this code path) and the logging improvements provide good observability.

**`src/aleph/vm/orchestrator/tasks.py`** (line 70): Minor: The type hint `max_seconds: int` is slightly inconsistent since `retry_delay` is initialized as `0.1` (float). Consider changing the hint to `max_seconds: float = 8.0` for clarity, though this is not a functional issue.

**`src/aleph/vm/orchestrator/tasks.py`** (line 145): Nit: `ws.exception()` may return `None` if no exception was set, which would log "WebSocket error: None". Consider `logger.error("WebSocket error: %s", ws.exception() or "unknown")` for slightly cleaner logs.

foxpatch-aleph (approved):
This PR correctly fixes the WebSocket connection reliability issues. The critical backoff bug (max→min) is properly fixed, heartbeat detection is added to prevent silent disconnects, exception handling prevents listener death, and assertions are replaced with logging. The code is clean, well-structured, and follows existing patterns in the codebase.

**`src/aleph/vm/orchestrator/tasks.py`** (line 86): Minor: The warning log on line 86 fires for both exceptions and normal generator exits. Consider moving it inside the except block if you want it to only log on actual errors.

**`src/aleph/vm/orchestrator/tasks.py`** (line 147): Minor: You might want to include the close code/reason in the log message: `logger.warning("WebSocket closed by server: %s", msg)` for better debugging.

PR #927, opened 2026-04-08
qemu: enable virtio-balloon free-page-reporting on instances

aliel:


foxpatch-aleph (approved):
This is a clean, well-documented change that enables virtio-balloon free-page-reporting for QEMU VMs without GPU passthrough. The logic correctly skips ballooning when GPUs are present (since vfio-pci pins guest memory, conflicting with page reclaim). The inline comment clearly explains the what, how, and why. The code follows existing patterns in the file and uses safe hardcoded strings with no injection risks.

PR #928, opened 2026-04-08
pool: add capacity admission control for memory, vCPU and disk

aliel:
- Refuse to create a VM when accepting it would push the host above its memory, vCPU or disk capacity. Memory and vCPU allow overcommit via configurable factors (MEMORY_OVERCOMMIT_FACTOR=1.1, VCPU_OVERCOMMIT_FACTOR=4.0); disk is strict. All running executions count toward committed totals so instances and programs are treated uniformly.

  The check runs authoritatively inside create_a_vm under creation_lock, before execution.prepare() downloads anything, and advisorily from notify_allocation and reserve_resources so clients get a fast rejection before payment. Cold program starts now return 503 Service Unavailable on capacity failure instead of 400 Bad Request.

Replaces the previous check_sufficient_resources helper.

### DIFF
--- a/src/aleph/vm/pool.py
+++ b/src/aleph/vm/pool.py
@@ -185,6 +185,13 @@ class VmPool:
         if isinstance(message, InstanceContent) and message.rootfs:
             required_disk_mib += message.rootfs.size_mib
         if message.volumes:
+            # Immutable volumes reference a file on Aleph storage via
+            # ``ref`` and do not carry a ``size_mib`` field, so they are
+            # not counted here and the admission estimate is best-effort
+            # for messages that include them. The authoritative disk
+            # check happens later via ``calculate_available_disk`` /
+            # ``get_disk_usage_delta``, which measures the real on-disk
+            # size of each downloaded file.
             for volume in message.volumes:
                 required_disk_mib += getattr(volume, "size_mib", 0) or 0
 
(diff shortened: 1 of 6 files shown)


src/aleph/vm/pool.py
```diff
@@ -125,6 +127,142 @@ def teardown(self) -> None:
...
+            return
+
+        required_memory_mib = message.resources.memory
+        required_vcpus = message.resources.vcpus
+        required_disk_mib = 0
+        if isinstance(message, InstanceContent) and message.rootfs:
+            required_disk_mib += message.rootfs.size_mib
+        if message.volumes:
+            for volume in message.volumes:
+                required_disk_mib += getattr(volume, "size_mib", 0) or 0
```
odesenfans, line 196: FYI this is not accurate, some volumes are files on Aleph and do not have a size field. So you're gonna miss bytes here and there.

src/aleph/vm/pool.py
```diff
@@ -125,6 +127,142 @@ def teardown(self) -> None:
...
+                f"physical {physical_memory_mib} MiB, "
+                f"host_reserved {host_reserved_mib} MiB, "
+                f"program_reserved {program_reserved_mib} MiB)"
+            )
+        if required_vcpus > available_vcpus:
+            errors.append(
+                f"vCPUs: required {required_vcpus}, available {available_vcpus} "
+                f"(committed {committed_vcpus} of {vcpu_cap} cap, "
+                f"physical {physical_cores} x factor {settings.VCPU_OVERCOMMIT_FACTOR})"
+            )
```
odesenfans: I don't think this takes your overcommit factor into account?

src/aleph/vm/pool.py
aliel: immutable volumes are currently not counted because they don't carry size_mib
i will add a comment flagging the limitation

src/aleph/vm/pool.py
aliel: it is applied, vcpu_cap is computed as int(physical_cores * settings.VCPU_OVERCOMMIT_FACTOR) a few lines above
the code is not clear enough, i will do some adjustment for readability

odesenfans (approved):

PR #930, opened 2026-04-13
tasks: fix WebSocket reconnect using exhausted generator

aliel:
Pass a factory function instead so each retry creates a fresh subscribe_via_ws call with a new WebSocket connection.


foxpatch-aleph (approved):
This PR correctly fixes a bug where exhausted async generators couldn't be reused on retry. By passing a factory function instead of a pre-created generator, each retry creates a fresh WebSocket connection. The fix is minimal, well-documented, and correctly typed.

PR #935, opened 2026-04-28
tests: fix invalid ItemHash in operator test fixtures

aliel:


foxpatch-aleph (approved):
This PR correctly fixes invalid ItemHash values in operator test fixtures. The old value `fake_vm_fake_vm_fake_vm_fake_vm_fake_vm_fake_vm_fake_vm_fake_vm_` contains non-hex characters ('v', 'm', '_'), which would fail `ItemHash` validation in `get_itemhash_or_400` (src/aleph/vm/orchestrator/views/operator.py:241) since ItemHash expects a 64-char hex string representing a SHA256 digest. The new value `decadecadecadecadecadecadecadecadecadecadecadecadecadecadecadeca` is a valid 64-char hex string and matches `settings.FAKE_INSTANCE_ID`. All assertions using the hash are updated consistently.

PR #938, opened 2026-04-30
status: fix egress check and add hostname-based HTTP probes with fallback

aliel:


foxpatch-aleph (commented):
The PR is well-structured and correct overall. It replaces Quad9 with Cloudflare for raw-IP probes, adds hostname-based HTTP connectivity checks with sensible fallbacks, and properly marks the new HTTP checks as advisory so transient third-party failures don't cause false-positive offline alerts. The HTTP/1.1-only ALPN SSL context is a thoughtful fix. No bugs or security issues found. A few minor suggestions for robustness and naming.

**`src/aleph/vm/orchestrator/views/host_status.py`** (line 33): The function name `check_ip_connectivity` is now misleading since it's used for both raw-IP probes (e.g. 1.1.1.1) and hostname-based URLs (e.g. detectportal.firefox.com). Consider renaming to something like `check_http_endpoint` or `check_connectivity` to better reflect that it handles both raw IPs and hostnames. This is a minor nit.

**`src/aleph/vm/orchestrator/views/host_status.py`** (line 58): `check_http_connectivity_with_fallback` creates a new `ClientSession` + `TCPConnector` for each URL attempt. For the 3 fallback URLs this is fine, but if efficiency matters you could reuse a single session/connector across all attempts. Not a bug — just a note.

**`src/aleph/vm/conf.py`** (line 144): The `CONNECTIVITY_HTTP_IPV4_URLS` and `CONNECTIVITY_HTTP_IPV6_URLS` lists are identical. If they're always the same, consider a single `CONNECTIVITY_HTTP_URLS` setting with both IPv4 and IPv6 checks using the same list. The per-family override is useful in theory (e.g. IPv6-only endpoints), but adds config surface for no current benefit. Minor suggestion.

**`tests/supervisor/test_status.py`** (line 1): The new `check_host_http_ipv4`, `check_host_http_ipv6`, and `check_http_connectivity_with_fallback` functions have no test coverage. Consider adding unit tests that mock `aiohttp` to verify: (1) the fallback mechanism stops at the first success, (2) all URLs are tried on failure, (3) the `@return_false_on_timeout` decorator correctly returns False on timeout.

foxpatch-aleph (approved):
This PR is well-structured and addresses two goals: (1) switching the raw-IP egress probes from Quad9 to Cloudflare DNS resolvers that have valid TLS certs for the raw IP, and (2) adding hostname-based HTTP connectivity checks with fallback URLs. The new `check_http_endpoint` function cleanly replaces the old `check_ip_connectivity`, the SSL context with ALPN set to http/1.1 is a good defensive measure to avoid h2 negotiation issues, and the `@return_false_on_timeout` decorator on the new functions ensures timeouts return False instead of raising. The status endpoint correctly marks the new HTTP checks as advisory (not affecting the 200/503 status). The test suite covers the fallback iteration logic and timeout behavior well. The code is clean, well-documented, and the changes are focused.

src/aleph/vm/orchestrator/views/host_status.py
```diff
@@ -9,6 +10,14 @@
...
 logger = logging.getLogger(__name__)
 
+# SSL context that negotiates HTTP/1.1 only via ALPN.
+# Without this, newer Python ssl advertises h2 as the preferred protocol.
+# Some connectivity probe endpoints accept h2 in the TLS handshake but only
+# speak HTTP/1.1, causing them to return 505 when they receive HTTP/1.1
+# frames over a connection negotiated as h2. Plain HTTP URLs are unaffected.
+_HTTP11_SSL_CONTEXT = ssl.create_default_context()
+_HTTP11_SSL_CONTEXT.set_alpn_protocols(["http/1.1"])
+
```
odesenfans: Did you encounter a situation where this was necessary? Seems unlikely.

src/aleph/vm/conf.py
```diff
@@ -141,10 +141,18 @@ class Settings(BaseSettings):
 
     API_SERVER: str = "https://official.aleph.cloud"
     IPFS_GATEWAY: str = "https://ipfs.aleph.im/ipfs"
-    # Connect to the Quad9 VPN provider using their IPv4 and IPv6 addresses.
-    CONNECTIVITY_IPV4_URL: str = "https://9.9.9.9/"
-    CONNECTIVITY_IPV6_URL: str = "https://[2620:fe::fe]/"
+    # Cloudflare DNS resolvers — have valid TLS certs for the raw IP and a
+    # real HTTP/1.1 web server, making them reliable egress connectivity probes.
+    CONNECTIVITY_IPV4_URL: str = "https://1.1.1.1/"
+    CONNECTIVITY_IPV6_URL: str = "https://[2606:4700:4700::1111]/"
```
odesenfans, line 147: I still don't understand why we do HTTP checks on DNS servers.

aliel: The idea is to test HTTPS egress without relying on DNS resolution, we need a raw IP with a valid TLS certificate.
DNS servers are the only widely available servers that have one, because they expose DNS-over-HTTPS (DoH) on their IP

src/aleph/vm/orchestrator/views/host_status.py
aliel: it was a defensive measure. removed

foxpatch-aleph (commented):
The review is complete. Here's my assessment:

**Verdict: COMMENT** — The code is correct and well-structured with good test coverage. No bugs or security issues require changes. A few observational comments:

**Summary:** The PR correctly fixes egress connectivity checks by switching from Quad9 to Cloudflare DNS resolvers (which have valid TLS certs for raw IPs), adds hostname-based HTTP probes with fallback URLs to verify DNS+egress end-to-end, and makes the new HTTP checks advisory so transient third-party failures don't cause false-positive offline alerts.

**Comments:**

1. **`host_status.py:37`** — `resp.status < 600` is intentionally loose to accept any HTTP response as proof of connectivity. The comment explains this well. No issue.

2. **`host_status.py:32`** — The `except aiohttp.ClientConnectorError` only catches connection-level failures. Other `aiohttp` exceptions (e.g., TLS errors, `InvalidUrl`) would propagate and escape the `@return_false_on_timeout` decorator (which only catches `TimeoutError`). Consider catching `aiohttp.ClientError` for broader resilience. In practice this is low-risk since the URLs are hardcoded, but it's worth hardening.

3. **`__init__.py:376`** — The `required_keys` set approach is clean. Set iteration order doesn't matter since all keys must be `True`.

4. **`conf.py:151`** — The mix of HTTP and HTTPS in `CONNECTIVITY_HTTP_URLS` is intentional (captive portal detection uses HTTP). Good design.

src/aleph/vm/orchestrator/views/host_status.py
```diff
@@ -21,29 +21,74 @@ async def wrapper(*args: Any, **kwargs: Any) -> bool:
...
         try:
-            async with session.get(url) as resp:
-                # We expect the Quad9 endpoints to return a 404 or 400 error, but other endpoints may return a 200
-                if resp.status not in (200, 400, 404):
-                    resp.raise_for_status()
-                return True
+            async with s.get(url) as resp:
+                # Any HTTP response confirms connectivity — the status code is
+                # irrelevant since probe endpoints may return 200, 204, 400, or 404.
+                return resp.status < 600
```
odesenfans: Mmh can't you just `return True` ? There's no HTTP status code above 600, you're just checking that you got an HTTP response.

odesenfans:
```suggestion
                return True
```

PR #939, opened 2026-05-01
fix host egress check

aliel:


foxpatch-aleph (approved):
This PR improves the host egress connectivity check by switching from Quad9 to Cloudflare DNS resolvers (which have valid TLS certs for raw IP), broadening error handling from ClientConnectorError to the base ClientError to catch all aiohttp client errors (including SSL, timeout, and response errors), and adding hostname-based HTTP connectivity checks with fallback URLs. The new HTTP checks are correctly marked as advisory so transient third-party endpoint failures don't make the CRN appear offline. The refactoring to reuse a single session across fallback URLs is clean and efficient. Tests are well-structured.

**`src/aleph/vm/orchestrator/views/host_status.py`** (line 29): The docstring says "any valid HTTP status code" but the function accepts all non-raising responses, including 5xx. This is intentional (any response proves connectivity), but the docstring could be slightly more precise, e.g. "Return True if the server responds with any HTTP status (even 4xx/5xx — a response proves egress works)."

**`tests/supervisor/test_status.py`** (line 42): The tests mock both `check_http_endpoint` and `aiohttp.ClientSession`. The session mocks aren't strictly needed since `check_http_endpoint` is patched — but they don't hurt. Just noting this is a minor redundancy in the test setup.

foxpatch-aleph (approved):
Solid improvement to the connectivity probe system. The PR replaces Quad9 with Cloudflare DNS resolvers (both have valid TLS certs for raw IP access), adds hostname-based HTTP probes with fallback URLs to verify DNS+egress end-to-end, and correctly marks the new 'http' checks as advisory so transient third-party failures don't falsely mark a CRN as offline. The refactor from check_ip_connectivity to the more flexible check_http_endpoint (with session reuse) is clean, and the broader aiohttp.ClientError catch is more defensive than the previous ClientConnectorError-only. Tests cover the key fallback and timeout scenarios well.

**`src/aleph/vm/orchestrator/views/host_status.py`** (line 34): Minor: the response body is never read/drained before the context manager exits. For HTTP/2 or keep-alive connections, this could leave data in the socket buffer. Not a functional issue for a connectivity probe, but worth noting. If this becomes a problem in practice, adding `await resp.release()` or `await resp.read()` would fix it.

**`tests/supervisor/test_status.py`** (line 47): Nit: the mock sets `__aexit__` to return `False`. In real Python semantics, returning `False` from `__aexit__` means exceptions propagate. The tests work because `check_http_endpoint` is also patched, so the real `_do_get` is never called, but it's slightly misleading. Not a bug.

foxpatch-aleph (approved):
Clean, minimal fix with two improvements: (1) removing the unused `resp` variable from the context manager, and (2) broadening exception handling from `ClientConnectorError` to `ClientError` (the base class). This correctly catches additional failure modes like SSL errors (`ClientConnectorSSLError`), response errors (`ClientResponseError`), and other client-side exceptions that weren't previously handled. The comment already documents that any HTTP response confirms connectivity, so catching all `ClientError` subclasses is the right semantics. The existing tests mock at a higher level and aren't affected by the exception type change. No issues found.

foxpatch-aleph (approved):
PR #939 makes two focused, correct improvements. (1) The status check in `views/__init__.py` now evaluates only a defined set of required health-check keys instead of `all(result.values())`, so non-essential checks like `cache`, `persistent_storage`, and `messages` no longer can force a 503. The `if k in result` guard correctly handles `lifespan` being absent in retro-compat mode. (2) The exception catch in `check_http_endpoint` is broadened from `ClientConnectorError` to `ClientError`, and an unused `resp` variable is removed — this correctly catches SSL errors, redirects, and other HTTP-level failures in addition to connection failures.

**`src/aleph/vm/orchestrator/views/__init__.py`** (line 343): Minor note: `lifespan` is in `required_keys` but is only added to `result` when `retro_compatibility` is `False`. The `if k in result` guard handles this correctly — it simply won't check `lifespan` on older runtimes. This is likely intentional but could use a comment for clarity.

**`src/aleph/vm/orchestrator/views/__init__.py`** (line 343): The `messages` and `get_a_message` keys are not in `required_keys`. If message connectivity should block the status, they should be added. If not, this is fine as-is.

**`src/aleph/vm/orchestrator/views/host_status.py`** (line 38): Good change — catching the broader `ClientError` base class also handles SSL errors, redirect-too-many, and other aiohttp client-side errors that `ClientConnectorError` alone would miss.

PR #942, opened 2026-05-05
refactor: extract allocation auth helpers to dedicated module

odesenfans:
## Summary

- Move `authenticate_api_request` and `requires_allocation_auth` out of `views/__init__.py` into a new `views/allocation_auth.py` module.
- Pure move, no behaviour change. Prep work for an upcoming signature-based auth scheme so that the cryptographic change can land as a focused diff.

## Details

- **New file:** `src/aleph/vm/orchestrator/views/allocation_auth.py` hosts both helpers verbatim.
- **`views/__init__.py`:** removes the two definitions; re-imports `authenticate_api_request` for its three direct call sites (`update_allocations`, `recreate_network`, `regenerate_proxy`); drops the `sha256` import that became unused.
- **`views/migration.py`:** imports `requires_allocation_auth` directly from the new module instead of re-routing through `views/__init__.py`.
- **`tests/supervisor/views/test_migration.py`:** the 4 `mocker.patch` targets and the explanatory docstring move from `aleph.vm.orchestrator.views.authenticate_api_request` to `aleph.vm.orchestrator.views.allocation_auth.authenticate_api_request`. The decorator now lives in `allocation_auth`, so the name lookup resolves there — `mock.patch` must target the lookup site.

## Test plan

- [x] `pytest tests/supervisor/views/test_migration.py` — 44 passed
- [x] `pytest tests/supervisor/test_views.py::test_allocation_*` and `test_regenerate_proxy_*_token` — 5 passed (covers the `update_allocations` and `regenerate_proxy` direct call paths)
- [x] `ruff check` and `ruff format --check` clean on the changed files

## Why this PR exists

The current shared-secret auth (one bearer token whose SHA-256 hash is configured per CRN) is being replaced with per-signer EIP-191 signatures. Splitting the move from the crypto change keeps both PRs easy to review.

### DIFF
--- a/src/aleph/vm/orchestrator/views/__init__.py
+++ b/src/aleph/vm/orchestrator/views/__init__.py
@@ -3,7 +3,6 @@ import binascii
 import http
 import logging
 from decimal import Decimal
-from hashlib import sha256
 from json import JSONDecodeError
 from packaging.version import InvalidVersion, Version
 from pathlib import Path
@@ -55,6 +54,7 @@ from aleph.vm.orchestrator.utils import (
     is_after_community_wallet_start,
     update_aggregate_settings,
 )
+from aleph.vm.orchestrator.views.allocation_auth import authenticate_api_request
 from aleph.vm.orchestrator.views.authentication import require_jwk_authentication
 from aleph.vm.orchestrator.views.host_status import (
     check_dns_ipv4,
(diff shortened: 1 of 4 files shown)


foxpatch-aleph (approved):
Clean, surgical refactoring that moves two auth helpers out of views/__init__.py into a dedicated module. All import sites are correctly updated (views/__init__.py, views/migration.py), the sha256 import was properly removed from __init__.py, and the test mock patch targets were correctly updated to point to the new module location. The patch target fix is particularly important — since requires_allocation_auth now lives in allocation_auth.py, mock.patch must target that module's namespace to intercept the lookup. No behavior change, no missing imports, no stray references.

PR #943, opened 2026-05-06
feat: add vm_type to /about/executions/list responses

odesenfans:
## Summary
- Add a `vm_type` field to each entry of `/about/executions/list` (v1) and `/v2/about/executions/list` (v2)
- Values come from the existing `VmType` enum: `microvm`, `persistent_program`, `instance`
- Removes the need to guess from other fields when consuming these endpoints

## Test plan
- [ ] `pytest tests/supervisor/test_views.py -k executions_list` (existing v2 list tests updated to assert the new key)
- [ ] Hit `/about/executions/list` and `/v2/about/executions/list` on a node running a mix of executions and confirm `vm_type` is present per entry

### DIFF
--- a/src/aleph/vm/orchestrator/views/__init__.py
+++ b/src/aleph/vm/orchestrator/views/__init__.py
@@ -77,6 +77,7 @@ from aleph.vm.utils import (
     get_ref_from_dns,
 )
 from aleph.vm.version import __version__
+from aleph.vm.vm_type import VmType
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


foxpatch-aleph (approved):
Clean, minimal change that adds a `vm_type` field (string from the existing `VmType` enum) to both v1 and v2 execution list endpoints. The `VmType.from_message_content()` method already exists and handles the three expected types (`InstanceContent`, `ProgramContent` with/without persistent). The v2 tests are correctly updated to assert the new field. No bugs or security concerns.

**`tests/supervisor/test_views.py`** (line 368): The v2 endpoint tests (`test_v2_executions_list_one_vm`, `test_v2_executions_list_vm_network`) are updated with the new `vm_type` assertion. There is no dedicated test for the v1 `/about/executions/list` endpoint — consider adding one for completeness, though this is not a blocker.

PR #944, opened 2026-05-06
packaging: harden supervisor shutdown to prevent dpkg upgrade failures

aliel:


foxpatch-aleph (changes requested):
The PR correctly addresses the dpkg upgrade failure issue through three complementary changes. The prerm script hardening and pool.stop() return_exceptions are solid. However, the systemd RestartPreventExitStatus uses invalid syntax — systemd expects numeric exit codes (143, 142) or special keywords ('success', 'core'), not signal names like 'SIGTERM'. This means the directive is silently ignored, and if the process is killed by SIGTERM outside of graceful shutdown, it would still restart. Fix to use numeric exit codes. The NDP proxy debounce is a nice bonus improvement with correct implementation.

**`packaging/aleph-vm/etc/systemd/system/aleph-vm-supervisor.service`** (line 16): **Bug: RestartPreventExitStatus does not accept signal names.**

systemd expects numeric exit status codes or the special keywords `success` and `core`. The signal names `SIGTERM` and `SIGINT` are silently ignored, making this directive a no-op.

When a process is terminated by SIGTERM, systemd sees exit code 143 (128 + 15). For SIGINT it's 142.

Fix:
```
RestartPreventExitStatus=0 143 142
```

Note: `Restart=on-failure` already prevents restart on normal exit (code 0), so the `0` here covers the case where the directive actually takes effect. Include it for clarity. See: https://www.freedesktop.org/software/systemd/man/systemd.service.html#RestartPreventExitStatus=

**`packaging/aleph-vm/DEBIAN/prerm`** (line 5): Minor: the `|| true` on `systemctl stop` means dpkg won't fail even if stop fails, but with `set -e` the `systemctl disable` on line 4 could still fail (though unlikely). The `reset-failed` on line 6 is a nice touch to clean up stale failed state.

**`src/aleph/vm/network/ndp_proxy.py`** (line 60): Nice debounce implementation. One concern: `_schedule_restart` is called from within `_update_ndppd_conf` which runs under `self._lock`. The asyncio task created here runs independently — if the event loop is shutting down (e.g. during `pool.stop()`), a task created here could raise `RuntimeError: Event loop is closed`. Consider adding a try/except around `asyncio.create_task` or cancelling the pending task in a teardown method.

**`src/aleph/vm/pool.py`** (line 769): Good use of `return_exceptions=True`. The `isinstance(result, BaseException)` check is the correct pattern. Minor: consider using `logger.warning` with `exc_info=True` if `result` is an Exception with a traceback, instead of just logging the string representation.

foxpatch-aleph (commented):
**Verdict: APPROVE**

This PR correctly addresses the dpkg upgrade failure caused by the systemd restart loop. The four changes work together well:

1. **prerm** — `|| true` on `disable`/`stop` tolerates systemctl failures; `reset-failed` clears any lingering failed state.
2. **service file** — `Restart=on-failure` + `RestartPreventExitStatus=0 130 143` prevents restart on clean exit, SIGINT, or SIGTERM. `TimeoutStopSec=30s` gives a reasonable shutdown window.
3. **ndp_proxy** — Debounces ndppd restarts across bursts of VM add/delete, and guards against closed event loops during shutdown.
4. **pool.stop() / supervisor.stop_all_vms()** — `return_exceptions=True` ensures one failing VM cleanup doesn't abort the rest; the supervisor catches any remaining exceptions to avoid a non-zero exit.

**Minor comments (not blocking):**

- **supervisor.py:263** — `logger.exception("... %s", error)` is slightly redundant since `logger.exception` already includes the traceback. Consider `logger.exception("Error stopping VMs during shutdown")` without the `error` arg. Purely cosmetic.
- **ndp_proxy.py:30** — 0.5s debounce is fine for typical workloads. If many VMs stop sequentially beyond that window, multiple restarts can still fire — but each picks up the latest config, so this is acceptable.
- **service file:19** — Exit code `0` in `RestartPreventExitStatus` is redundant with `Restart=on-failure` (which only restarts on non-zero). Keeping it for documentation clarity is reasonable.

foxpatch-aleph (approved):
This PR hardens the supervisor shutdown path to prevent systemd restart loops during dpkg package upgrades. The changes are well-motivated and correctly implemented. The prerm script now tolerates missing services, the systemd unit uses `Restart=on-failure` with explicit exit codes to prevent restarts on intentional shutdowns, `VmPool.stop()` catches individual VM cleanup errors without aborting, the supervisor's `stop_all_vms` catches all exceptions, and the ndppd restart is debounced to coalesce bursts. There are no correctness bugs or security issues.

**`packaging/aleph-vm/etc/systemd/system/aleph-vm-supervisor.service`** (line 19): `RestartPreventExitStatus=0` is redundant with `Restart=on-failure` — systemd only restarts on non-zero exits with this policy. Consider adding a comment explaining this is a safety net against someone later switching to `Restart=always`, or simply remove it to avoid confusion.

odesenfans (approved):

PR #945, opened 2026-05-08
auth: Aleph-EIP191-V1 signature path for scheduler endpoints

odesenfans:
## Summary

Adds a per-signer EIP-191 signature path (`Authorization: Aleph-EIP191-V1 sig=...,payload=...`) to the 8 scheduler-only endpoints, alongside the legacy `X-Auth-Signature` shared token. Both paths run during rollout; the legacy path logs a deprecation warning so operators can spot un-migrated scheduler clients.

Stacks on PR #942 (allocation auth helpers extraction, already merged).

## Why

Today every CRN holds the same shared token. A single compromised CRN can spoof the scheduler against every other CRN. The new scheme has CRNs hold only public values (an explicit list of authorized signer addresses), and binds each request via signature recovery + monotonic-iat replay protection.

## What's in this PR (supervisor side)

- New settings: `AUTHORIZED_ALLOCATION_SIGNERS`, `ALLOCATION_SIGNATURE_MAX_AGE_SECONDS`. Address checksum validation at startup.
- `_verify_aleph_signature` in `allocation_auth.py`, with method/path/body binding, ±5min absolute time window, and a per-signer `_last_accepted_iat` floor.
- `authenticate_api_request` rewritten as an async dispatcher; presence of the new header disables the legacy fallback (no spoofing the migration).
- 30+ unit tests + 3 integration tests on `update_allocations`.

## What's NOT in this PR

- Scheduler-side signing client. Tracked in a separate change in the scheduler repo.
- Removal of the legacy path. Phase 3, separate change.

## Test plan

- [x] Unit tests for the verifier (parser, time window, method/path, signer authorization, body hash, replay, malformed headers, tampered payload).
- [x] Dispatcher tests (route to signature path, no fallback on bad sig, fallback to legacy with warning, empty legacy token returns False).
- [x] Integration tests on `update_allocations` for all three branches.
- [x] Existing `test_migration.py` tests still pass.
- [x] Existing `test_allocation_*` / `test_regenerate_proxy_*_token` tests still pass.

### DIFF
--- a/docs/plans/2026-05-11-allocation-auth-signatures-review.md
+++ b/docs/plans/2026-05-11-allocation-auth-signatures-review.md
@@ -1,0 +1,143 @@
+# Code Review — Aleph-EIP191-V1 Allocation Auth (PR #945)
+
+Self-review of the new signature-based allocation auth, in the spirit of a
+brutally critical reviewer. Issues are ordered by severity, with mitigations.
+
+## Critical
+
+### C1. TOCTOU race on `_last_accepted_iat` — replay protection is broken under concurrency
+`_verify_aleph_signature` does:
+```python
+previous = _last_accepted_iat.get(recovered.lower(), float("-inf"))
+if iat <= previous:
+    return False
+...
+_last_accepted_iat[recovered.lower()] = iat
+```
+Between the `get` and the assignment, the function `await`s `request.read()`. Two concurrent requests with `iat=N` (same captured signed message) both read `previous < N`, both pass the gate, both write — **same-iat replay succeeds**. Worse: two concurrent legitimate calls with `iat=N` and `iat=N+1` can write in any order, leaving `previous = N`, after which `iat=N+1` can be replayed.
+
+This is the whole point of replay protection. Without serialization across the await point, you have none.
+
+**Mitigation:** wrap the check+update in an `asyncio.Lock` (one global, or one per signer). Move the `_last_accepted_iat` update *before* the `await request.read()` if you can prove the body check can't fail after the iat has been "committed" — but that's a data-corruption antipattern; just take the lock.
+
+### C2. Body-size DoS via `await request.read()` inside the verifier
+Any caller can send `Authorization: Aleph-EIP191-V1 sig=…,payload=…` plus a multi-GB body. The verifier calls `await request.read()` *unconditionally* on the way to the body-hash check, fully buffering attacker-controlled bytes into memory before any auth decision. The legacy path never read the body. This is a regression in DoS posture introduced by this PR.
+
+**Mitigation:** set aiohttp's `client_max_size` on the app (e.g., 1 MiB for control endpoints), or check `request.content_length` against a cap before calling `read()`, returning False if exceeded. Either way, **bound memory before reading**.
+
+## High
+
+### H1. Scheme comparison is case-sensitive — violates RFC 7235 §2.1
+`if scheme != ALEPH_EIP191_V1_SCHEME` rejects `aleph-eip191-v1`, but auth-scheme tokens are case-insensitive. A well-behaved client lowercasing the scheme would silently get 401s, hard to debug, and a maintainer "fixing" this later by lowercasing both sides would need to be careful about the prefix check on line 150 (`startswith(f"{ALEPH_EIP191_V1_SCHEME} ")`).
+
+**Mitigation:** `scheme.casefold() != ALEPH_EIP191_V1_SCHEME.casefold()` in the parser, and use the same comparison in the dispatcher (or extract the scheme first then dispatch).
+
+### H2. Path binding ignores query string
+`payload["path"] != request.path` — `request.path` excludes the query. If any of these 8 endpoints ever gains a query param affecting behavior, the signature won't bind to it. Footgun for future maintainers.
+
+**Mitigation:** sign `request.path_qs` (or assert `request.query_string == ""` in the verifier and document that scheduler endpoints MUST NOT accept query params).
+
+### H3. Unknown auth-params silently accepted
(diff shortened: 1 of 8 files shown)


foxpatch-aleph (approved):
This is a well-designed, carefully implemented enhancement to add per-signer EIP-191 signature authentication alongside the legacy shared-token scheme. The code is clean, the edge cases (boolean vs int iat, replay protection via monotonic iat, no legacy fallback when Authorization header is present, etc.) are properly handled, and the test coverage is thorough (30+ unit tests + integration tests). The only issues are minor: the dispatcher emits a deprecation warning for legacy requests that includes `request.remote` which may be None in certain deployment configurations, the `iat` comparison uses Python `time.time()` (float) vs an int which is safe due to the `abs(iat - now) > max_age` comparison but worth noting, and there's minor duplicated test helper code between `test_views.py` and `test_allocation_auth.py`.

**`src/aleph/vm/orchestrator/views/allocation_auth.py`** (line 91): **Minor: iat type mismatch in comparison.** `iat` is verified to be a `json` integer (Python `int`), but `time.time()` returns a `float`. The comparison `abs(iat - now) > max_age` works correctly because Python handles int-float subtraction, but `_last_accepted_iat` stores the `int` iat and compares `iat <= previous` where `previous` could be `float('-inf')` initially. This is fine in practice, but consider whether the iat should be stored/compared as a float for consistency, or whether it matters at all (it doesn't since both comparisons work correctly).

**`src/aleph/vm/orchestrator/views/allocation_auth.py`** (line 153): **Suggestion: `request.remote` may be None.** The deprecation warning logs `request.remote` which can be `None` in some aiohttp configurations (e.g., behind certain reverse proxies). Consider using `request.remote or 'unknown'` to avoid logging 'None' in the warning message.

**`src/aleph/vm/orchestrator/views/allocation_auth.py`** (line 112): **Discussion: Strict inequality `iat <= previous` prevents same-iat replays.** This is the correct choice for a monotonic counter — using `<` instead of `<=` would allow same-iat replays. However, note that if the clock precision is low or two requests are sent in the same millisecond, the second will be rejected. The PR's test `test_verify_accepts_strictly_increasing_iat` verifies the `iat+1` pattern, which is the correct client behavior. This is a reasonable tradeoff.

**`src/aleph/vm/orchestrator/views/allocation_auth.py`** (line 91): **Suggestion: Consider making the time window check inclusive vs exclusive more explicit.** `abs(iat - now) > max_age` means a request at exactly `max_age` seconds old is rejected (since `>` not `>=`). This is fine, but given that the iat is an integer (seconds) and `time.time()` is fractional, in practice the boundary is blurry anyway. A comment or using `>=` would make the intent clearer, but this is a nit.

**`tests/supervisor/test_views.py`** (line 996): **Minor: Duplicated test helper.** `_make_aleph_eip191_v1_header` in `test_views.py` duplicates `make_signed_payload` + `make_auth_header` from `test_allocation_auth.py`. Consider extracting a shared test utility to avoid drift.

**`src/aleph/vm/orchestrator/views/allocation_auth.py`** (line 58): **Observation: Module-level mutable state across endpoints is acceptable here.** `_last_accepted_iat` is a module-level dict that doesn't survive restarts. In a multi-worker deployment (multiple processes), each worker has its own `_last_accepted_iat`, so a replayed request could succeed against a different worker. The 5-minute absolute time window mitigates this — replay within 5 min is still bound by the iat floor per worker. Worth documenting in a comment or in the PR description for future readers.

foxpatch-aleph (approved):
Exceptionally well-crafted PR adding Aleph-EIP191-V1 per-signer authentication alongside the legacy shared-token path. The self-review is thorough, and every identified issue (C1 TOCTOU race via asyncio.Lock, C2 body-size DoS cap, H1 case-insensitive scheme, H3 unknown auth-param rejection, M1 WARNING-level logging, M2 boot-time deprecation notice, M3 boundary tests, M4 cross-signer iat tests, M5 request.read() side-effect docs, M6 import hoisting, L1 lazy import cleanup, L5 trailing-slash comment) has been addressed. The cryptographic design (EIP-191 + recover, body-hash binding, monotonic iat with per-signer floor, absolute time window) is sound. Test coverage is extensive with 30+ unit tests and integration tests on update_allocations. The remaining gap (no end-to-end integration tests for migration endpoints with the new scheme) is acknowledged and low-risk since all 5 migration endpoints share the same dispatch path through requires_allocation_auth.

**`tests/supervisor/views/test_migration.py`** (line 41): Consider adding at least one end-to-end test (parametrized across handlers) that fires a real Aleph-EIP191-V1 signed request at a migration endpoint. The mock_scheduler_auth fixture already uses AsyncMock(return_value=True), so the decorator chain is tested, but a real E2E test would close the H4 gap.

foxpatch-aleph (approved):
Well-structured PR adding per-signer EIP-191 signature auth to scheduler endpoints. The critical TOCTOU race on iat replay protection (C1) has been correctly fixed with an asyncio.Lock, and the DoS body-size regression (C2) is addressed with MAX_SIGNED_REQUEST_BODY_BYTES. The self-review process identified and fixed most issues before they hit review — H1 (case-insensitive scheme), H3 (unknown auth-params), M3/M4 (boundary/per-signer tests), and L1 (import hygiene) are all resolved. What remains (H4 — full integration tests for migration endpoints) is a non-blocking gap with declared follow-up scope. Code is clean, tests are thorough (30+ unit + 3 integration), and the cryptographic core (EIP-191 recover, body-hash binding, per-signer monotonic iat) is sound.

**`src/aleph/vm/orchestrator/views/__init__.py`** (line 484): `update_allocations` docstring references "scheduler-only endpoints" — but the migration endpoints (`migration.py:80,161,232,319,346`) also use `@requires_allocation_auth`. This creates an asymmetry in integration coverage. Consider adding a docstring note that integration tests for migration endpoints are tracked separately.

**`src/aleph/vm/orchestrator/views/allocation_auth.py`** (line 218): The dispatcher rejects any non-Aleph-EIP191-V1 Authorization header outright, including `Authorization: Bearer ...`, even if a valid `X-Auth-Signature` is also present. This is intentional and tested, but the error message is the generic 401 text — operators debugging a misconfigured client that sends both headers will have no indication of why the legacy path didn't kick in. Consider a debug-level log distinguishing "wrong scheme" from "scheme match but verification failed".

PR #948, opened 2026-05-26
fix: return 400 instead of 500 on invalid restore upload

odesenfans:
## Problem

The backup restore endpoint (`POST /control/machine/{ref}/restore`) returned a **500** when handed a file it can't use — e.g. uploading a `.tar` archive instead of a QCOW2 image.

## Root cause

`_do_restore` streams the upload to disk as `restore-<vm>.qcow2` *without* checking the format, then calls `verify_qemu_disk()` → `qemu-img check`. For a non-QCOW2 file, `qemu-img` exits non-zero, so `run_in_subprocess` raises `subprocess.CalledProcessError`. That is not an `HTTPException`, so it fell through to the generic `except Exception` handler → `HTTPInternalServerError`.

## Fix

Wrap the `qemu-img` validation step (`verify_qemu_disk` + `get_qemu_disk_virtual_size`) in `_do_restore` and convert a `CalledProcessError` into an `HTTPBadRequest` (`"Uploaded file is not a valid QCOW2 disk image"`), since an unusable upload is a client error. This covers both the multipart-upload and the `volume_ref` download paths.

The fix is scoped to the restore **call site** on purpose: `verify_qemu_disk` is also used elsewhere to verify a backup the node created *itself*, where a failure is genuinely server-side and must stay a 500.

## Testing

Added `test_restore_rejects_invalid_image_format` (TDD): mocks `verify_qemu_disk` to raise `CalledProcessError` and asserts a 4xx. Confirmed it fails on the pre-fix code (raises `HTTPInternalServerError`) and passes after the fix, with no regressions in `test_views.py` / `test_qemu_backup.py`.

### DIFF
--- a/src/aleph/vm/orchestrator/views/operator.py
+++ b/src/aleph/vm/orchestrator/views/operator.py
@@ -3,6 +3,7 @@ import hashlib
 import hmac
 import json
 import logging
+import subprocess
 import time
 from dataclasses import dataclass
 from datetime import timedelta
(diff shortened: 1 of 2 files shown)


foxpatch-aleph (commented):

foxpatch-aleph (approved):
A clean, well-scoped fix that correctly converts a 500 to a 400 when a non-QCOW2 file is uploaded to the restore endpoint. The try/except wraps only the qemu-img validation calls and catches `CalledProcessError` — exactly the right granularity. The existing callers of `verify_qemu_disk` are intentionally unaffected since those represent genuine internal errors. Tests are minimal but correct, covering the multipart upload path (the most common scenario). A minor nit: only the multipart path is tested, not the JSON reference path, but both benefit from the same fix.

**`tests/supervisor/test_views.py`** (line 1144): Consider also adding a test for the `_parse_restore_json` path where `verify_qemu_disk` fails on a volume reference. The fix covers both paths, but only the multipart upload is tested.

foxpatch-aleph (approved):
Clean, minimal fix that correctly converts a qemu-img validation failure from a 500 to a 400 on the restore endpoint. The change is properly scoped to the restore call site (not the backup call site where a failure is genuinely server-side). The test covers the multipart upload code path and verifies the 4xx response. The approach is correct and well-reasoned.

**`tests/supervisor/test_views.py`** (line 1096): The vm_hash `decadecadecadecadecadecadecadecadecadecadecadecadecadecadecadeca` is 62 chars — ItemHash is 64 hex chars typically. This doesn't matter since it's mocked, but for consistency with real values, using a proper 64-char hex hash would avoid confusion on future reads.

PR #952, opened 2026-05-29
feat(supervisor): Phase 0.B - Supervisor ABC + DTOs + in-process implementation

odesenfans:
## What

Phase 0.B of the aleph-vm architecture backport: introduces the Python `Supervisor` abstraction, the single agent-to-VM-management call path that a gRPC client (0.D) and eventually a Rust supervisor (Phase 2) drop into.

New package `aleph.vm.supervisor`:
- `types.py` - frozen-dataclass DTOs + enums mirroring `supervisor.proto` (no Aleph or protobuf types leak through).
- `errors.py` - closed `SupervisorError` hierarchy (one per `ErrorCode`) + `translate_exception()` mapping the scattered internal backend exceptions + a `translating_errors()` context manager. This is the wire-error vocabulary, built in Python first.
- `abc.py` - seven capability ABCs aggregated into `Supervisor` (25 async methods, 1:1 with the proto RPCs).
- `inprocess.py` - `InProcessSupervisor` wrapping today's `VmPool` / `VmExecution`.

## Scope (what is functional)

- 12 of 25 methods are real and tested against a faithful `VmPool`: `health`, `get_host_info`, `get_vm`, `list_vms`, `delete_vm`, `reboot_vm`, `reinstall_vm`, the three port-forwarding methods, and the two log methods.
- 13 methods are explicit `NotImplementedSupervisorError` stubs, deferred by design: `create_vm` (the whole create vertical, including the translator and the resources/controller decouple, is a dedicated follow-up), plus backups (6), migration (3), confidential (3).
- The abstraction is intentionally dormant: no view or `run.py` call site is rewired (that is 0.E). Production behavior is unchanged.

## Tests

71 tests across `tests/supervisor/` (DTO/enum surface, error translation, the 25-method ABC surface, a reusable conformance suite, and per-method in-process unit tests). `mypy` clean on the package; `ruff format` + `isort` clean.

Design and plan: `docs/plans/2026-05-29-phase-0b-supervisor-abc-design.md` and `...-implementation.md`. A final-review carry-forward list for 0.E is recorded in the design doc (section 12).

## Notes for reviewers

- Stacked on #951 (the hypervisor->supervisor rename). Base is `od/supervisor-rename`; will retarget to `dev` once #951 merges so the diff stays clean.
[...]

PR #953, opened 2026-05-29
feat(supervisor): message-free QEMU config build + ExecutableContent translator

odesenfans:
## What

First half of the deferred `create_vm` vertical: build a launchable QEMU instance configuration from a message-agnostic `CreateVmSpec`, and translate an Aleph message into a `CreateVmSpec`. The hypervisor side never touches `ExecutableContent`.

Key finding: `QemuVMConfiguration` (the controller JSON) is already a message-agnostic chokepoint, and the actual QEMU launch (`QemuVM.start()`) never reads the message. So this is an almost entirely **additive** decouple, reusing the existing launch path.

## Changes

- `proto/supervisor.proto`: `CreateVmRequest.ssh_authorized_keys` (field 13); bindings regenerated, drift check green.
- `CreateVmSpec.ssh_authorized_keys: list[str]` (defaulted, so existing construction sites are unaffected).
- `supervisor/qemu_build.py` (new): `build_qemu_configuration(spec, vm_id, tap_interface) -> Configuration`, mirroring `AlephQemuInstance.configure()` field-for-field but sourced from the spec + the supervisor-assigned tap interface; plus a message-free `build_cloud_init_drive(...)` that calls the existing `create_cloud_init_drive_image`.
- `supervisor/translate.py` (new): `build_create_vm_spec(vm_hash, message, *, gpus=()) -> CreateVmSpec` for QEMU non-confidential instances; runs today's `AlephQemuResources.download_all()` to materialize disks, validates (instance / qemu / non-confidential) before materializing.

## Scope

Behavior-preserving decouple, no boot. Out of scope (follow-ups): wiring `InProcessSupervisor.create_vm` + `pool.executions` (the VmExecution-from-spec question); Firecracker/program and confidential creation; GPU assignment (a translator parameter here).

## Verification

83 supervisor-package tests pass; mypy clean on the package; ruff/isort clean; proto drift clean. The full daemon suite + droplet integration run on CI. The MiB->MB memory formula and disk-role mapping (the easy-to-get-wrong spots) are pinned by tests; cloud-init argument derivation from the tap interface is asserted.

Design: `docs/plans/2026-05-29-phase-0c-create-vm-decouple-design.md`.

## Notes

- Stacked on #952 (Phase 0.B). Base is `od/supervisor-abc`; will retarget to `dev` once #952 merges.


foxpatch-aleph (approved):
Clean, well-tested decoupling of QEMU configuration from Aleph messages. The code is behavior-preserving, correctly replicates the existing memory formula, and has strong test coverage including validation-before-IO guarantees. The design cleanly separates the translation layer (message→spec) from the build layer (spec→config). Minor observations about anticipatory DATA role handling and empty Path usage are non-blocking.

**`src/aleph/vm/supervisor/qemu_build.py`** (line 107): Mount string is derived via `Path.stem` as a placeholder. The test asserts `/mnt/extra` which is correct for the current code, but the mount heuristic may need revisiting when the DATA disk role is actually produced by a translator, since those volumes may carry meaningful mount points expected by the guest.

**`src/aleph/vm/supervisor/qemu_build.py`** (line 112): `DdiskRole.DATA` is included in the set but the only current producer (translate.py) never creates DATA role disks — all volumes are mapped to EXTRA. The DATA branch is anticipatory but untested. Consider adding a test for DATA role handling when the first producer of DATA disks lands.

**`src/aleph/vm/supervisor/translate.py`** (line 85): `Path("")` resolves to `PosixPath(".")` (current directory). While this matches existing patterns and is semantically "no kernel/initrd", it could cause unexpected behavior if a downstream consumer does path operations on these values. Consider using `Path()` or leaving as `None` if a typed nullable is preferred.

**`tests/supervisor/test_supervisor_qemu_build.py`** (line 250): `test_no_rootfs_disk_raises` relies on the exception being raised before the cloud-init call, so the missing monkeypatch doesn't matter. This works correctly but the dependency on call order is implicit. Consider adding a `monkeypatch` that would fail if reached, as the translate tests do, to make the defense explicit.

odesenfans:
## Update: addressed review on memory + volume mounts

Two follow-up changes from review:

**Typed memory sizes (correct allocation, both paths).** Introduced `src/aleph/vm/sizes.py` with a typed `MemorySize` (MiB/MB/bytes + explicit conversions), and used it in BOTH the new `qemu_build` path AND the live `AlephQemuInstance.configure()` + `AlephQemuConfidentialInstance.configure()`. This fixes a latent under-allocation bug: the old formula `int(memory / 1024 / 1024 * 1000 * 1000)` turned a 2048 MiB request into `-m 1953` (QEMU `-m` is MiB), so guests ran ~5% short. They now get their full requested MiB.

NOTE: this is a production behavior change for existing QEMU/confidential instances - on next launch they receive their correctly-requested memory (slightly more than before). Both construction paths stay in parity (both correct).

**Volume mounts carried for parity.** Added `mount` to the proto `DiskConfig` (field 5) and the `DiskSpec` dataclass; the translator now carries the message's `volume.mount` through and `build_qemu_configuration` uses it, instead of the earlier derived `/mnt/{stem}` placeholder. Full parity with `download_volumes()`/`configure()`.

88 supervisor-package tests pass; mypy clean across the package + `sizes.py` + both controllers; proto drift clean.

foxpatch-aleph (approved):
Well-structured, behavior-preserving decoupling PR. New modules are clean, well-tested, and correctly implement the design in `docs/plans/2026-05-29-phase-0c-create-vm-decouple-design.md`. The memory unit fix (MemorySize round-trip replacing the buggy int-division formula) is correctly applied. Tests are thorough, covering happy paths, error paths, and the critical invariant that validation fires before I/O. No bugs or security issues found.

**`src/aleph/vm/supervisor/qemu_build.py`** (line 51): get_hostname_from_hash expects ItemHash but receives str; the type: ignore works at runtime because ItemHash extends str, but widening the parameter in cloudinit.py or adding an explicit cast here would be cleaner.

**`src/aleph/vm/supervisor/qemu_build.py`** (line 119): MemorySize.from_mebibytes(X).to_mebibytes() is a round-trip no-op. A brief comment noting this is intentional consistency with instance.py (not an actual conversion) would help future readers.

foxpatch-aleph (approved):
Clean, well-designed decouple of message types from QEMU configuration building. The code faithfully implements the design doc, tests are thorough and well-structured, and the validation-before-IO pattern in translate.py is a good defensive practice. The memory unit fix via the typed sizes module addresses a real bug. Minor points: one async test is missing the @pytest.mark.asyncio decorator that all sibling tests have, and the mem_size_mb variable naming (actually MiB) is inherited from the existing codebase.

**`tests/supervisor/test_supervisor_qemu_build.py`** (line 251): Add `@pytest.mark.asyncio` decorator for consistency with the other async tests in this file. Currently works because pytest-asyncio >=0.21 defaults to `auto` mode, but being explicit matches the file's own convention and avoids fragility.

odesenfans:
## Update: renamed `mem_size_mb` -> typed `mem_size`

The controller config field was a misnomer (it was fed to QEMU `-m`, which is MiB). Renamed to `mem_size` and made it a **typed `Bytes`** (from `aleph.vm.sizes`, the canonical unit), serialized to/from a plain integer number of bytes in the JSON via a pydantic `Annotated[Bytes, BeforeValidator, PlainSerializer]` (so the on-disk JSON stays `"mem_size": <int bytes>`, a bare int). Applied across both `QemuVMConfiguration` and `QemuConfidentialVMConfiguration`, both writers (`instance.py` x2), `qemu_build.py`, and both launchers (which now convert `mem_size.to_rounded(MiB, FLOOR)` for `-m`). A JSON round-trip test pins the bare-int shape and `Bytes` reconstruction. 100 supervisor-package tests pass; mypy clean (13 files); no `mem_size_mb` references remain.

### Migration note (for whoever deploys this line)
This changes the serialized controller-config schema (`mem_size_mb: int` -> `mem_size: int bytes`), and that JSON is read by the separate `aleph-vm-controller@` process. A controller restarting on a config written by a pre-upgrade version would fail to deserialize. It is safe on this unmerged dev-line (nothing deployed reads the new format), but before this ships it needs either regenerate-on-restart or a transient backward-compat alias. Not built here on purpose.

foxpatch-aleph (approved):
Well-structured, carefully scoped refactoring that correctly decouples QEMU configuration building from Aleph message types. The new sizes.py module prevents a class of unit-mixing bugs. Tests cover all important paths including rejection paths that verify I/O is skipped on validation failure. Minor issues: misleading docstring about memory formula, undocumented type: ignore[arg-type], and no test for multiple rootfs disks. None block merge.

**`src/aleph/vm/supervisor/qemu_build.py`** (line 94): The docstring says 'The memory formula is reproduced verbatim' but line 119 actually uses the *correct* MiB→Bytes conversion, not the buggy MiB→MB formula. Consider updating: 'Memory is passed as Bytes (correct MiB→Bytes conversion, unlike the old path which had a unit-mixing bug).'

**`src/aleph/vm/supervisor/qemu_build.py`** (line 50): The `type: ignore[arg-type]` suppresses calling `get_hostname_from_hash(vm_hash: str)` which expects `ItemHash`. A brief comment would help: '# Accepts ItemHash at runtime, only calls .encode().upper() so str is fine.'

**`tests/supervisor/test_supervisor_qemu_build.py`** (line 187): Docstring says '2048 MiB is passed to QEMU as 2048' but the test asserts mem_size bytes == 2048*1024*1024, then round-trips through floor rounding. The test is correct, consider rewording: '2048 MiB is stored as Bytes(2147483648) and round-trips to 2048 MiB via floor-rounding.'

**`tests/supervisor/test_supervisor_qemu_build.py`** (line 257): Consider adding a test for spec with multiple ROOTFS disks to document expected behavior (raise error? pick first?). Currently only the missing-ROOTFS case is tested.

foxpatch-aleph (commented):
Well-scoped decoupling PR with clean architecture and thorough tests. The new qemu_build.py and translate.py modules correctly mirror existing behavior while removing message-type dependencies from the hypervisor side. The typed size system (sizes.py) is a solid improvement over the prior unit-mixing formula. Issues found: a stale docstring referencing 'verbatim memory formula' that was actually changed, untested DiskRole.DATA path in host_volumes mapping, untested DEVELOPER_SSH_KEYS appending, the getattr-with-default in translate.py being unnecessarily defensive, and a type: ignore that could be resolved with a cast. None are blocking — all are minor observations for robustness.

**`src/aleph/vm/supervisor/qemu_build.py`** (line 94): Docstring says "The memory formula is reproduced verbatim -- do not 'fix' it; this is a deliberate decouple." But the memory formula has actually been changed in this PR (from the old MiB→MB under-allocation to the correct typed MiB→Bytes path). Update to reflect the actual intent: the transition to typed sizes is intentional, not a verbatim reproduction.

**`src/aleph/vm/supervisor/qemu_build.py`** (line 51): The `# type: ignore[arg-type]` on `get_hostname_from_hash(vm_hash)` could be resolved with a cast: `cast(ItemHash, vm_hash)`. Since `vm_hash` is `str` and `ItemHash` is a NewType over str, a cast is safe and cleaner than the ignore.

**`src/aleph/vm/supervisor/translate.py`** (line 55): The `getattr(message.environment, "trusted_execution", None)` is unnecessarily defensive — line 48 already guards that `message` is `InstanceContent`, which always has `environment: InstanceEnvironment` with a `trusted_execution` field. Prefer `message.environment.trusted_execution` directly.

**`src/aleph/vm/supervisor/qemu_build.py`** (line 112): The `DiskRole.DATA` branch in the host_volumes comprehension is untested — all test specs only use DiskRole.EXTRA. Add a test case with a DATA disk to exercise this path.

**`src/aleph/vm/supervisor/qemu_build.py`** (line 67): The `settings.USE_DEVELOPER_SSH_KEYS` / `settings.DEVELOPER_SSH_KEYS` appending path is untested. Consider adding a test (likely needs monkeypatching the settings) to assert that developer keys get appended when the setting is enabled.

**`tests/supervisor/test_supervisor_qemu_build.py`** (line 234): The `test_no_tap_interface` test only asserts `interface_name is None`, but doesn't verify that the cloud-init network parameters (ip, route, ipv6, ipv6_gateway) were passed as empty strings. Consider capturing the cloud-init call args like the happy-path test does and asserting empty network params.

**`tests/supervisor/test_supervisor_qemu_build.py`** (line 271): The `test_mem_size_json_round_trip` function accepts a `monkeypatch` fixture parameter but never uses it. Remove the unused parameter.

foxpatch-aleph (approved):
Well-structured decouple PR that introduces message-agnostic QEMU config building and message translation. Code is correct, tests are thorough (83 supervisor tests pass), design document is clear. The MiB memory type fix resolves a prior unit-mixing bug. No correctness or security issues found.

**`src/aleph/vm/supervisor/translate.py`** (line 60): `namespace=str(vm_hash)` — verify that `ItemHash.__str__()` produces the expected string form. Tests suggest it does, but worth confirming if `ItemHash` is a bytes-backed type.

**`tests/supervisor/test_supervisor_translate.py`** (line 168): The `SimpleNamespace` stand-in for the rejection test has `vcpus` and `memory` directly on `resources` rather than mirroring `MachineResources`. This works because the isinstance check fires first, but adding a brief comment explaining why the incomplete shape is safe would help future readers.

**`tests/supervisor/test_supervisor_qemu_build.py`** (line 115): The `fake_cloud_init` test double captures `**kwargs` but doesn't assert `has_gpu` or `is_confidential` derivation. Consider adding assertions for those in a follow-up, since they're derived from `spec.gpus` and `spec.backend` respectively.

PR #954, opened 2026-05-31
Message-free supervisor (1/4): spec-constructible VmExecution

odesenfans:
Part **1 of 4** of the message-free supervisor backport. Base: `dev`.

Makes `VmExecution` constructible and operable from a message-free `CreateVmSpec` (resolved on-disk paths only — no Aleph message, no download), alongside the existing message path with **no behaviour change** to it.

### Changes
- `AlephQemuResources.from_spec` — build the QEMU resources holder from a spec, no download; `message_content` made optional.
- `VmExecution.from_spec` — construct an execution with `message`/`original` = `None` and a `spec`.
- Spec-aware backend properties (`is_program`/`is_instance`/`is_confidential`/`hypervisor`) and payment-flag guards.
- Spec branches in `prepare()` (no download) and `create()` (builds `AlephQemuInstance` from spec-derived resources).
- `start(write_config=False)` to skip the message-coupled `vm.configure()`; `save()` no-ops when there is no message.
- Type-narrowing asserts document the "an execution has a spec XOR a message, never neither" invariant (keeps mypy at baseline).
- Design + implementation-plan docs under `docs/plans/`.

### Stack
1. **(this PR)** spec-constructible `VmExecution` → `dev`
2. pool create path (`create_a_vm_from_spec`, `InProcessSupervisor.create_vm`) → 1
3. route production creation through the spec (`run.py`) → 2
4. reboot-recovery from on-disk configs → 3

New tests: `tests/supervisor/test_supervisor_spec_resources.py`, `tests/supervisor/test_supervisor_spec_execution.py`.

### DIFF
--- a/docs/plans/2026-05-29-message-free-supervisor-create-plan.md
+++ b/docs/plans/2026-05-29-message-free-supervisor-create-plan.md
@@ -1,0 +1,2003 @@
+# Message-free supervisor create + reboot-recovery Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Make the supervisor *create* a VM from a message-free `CreateVmSpec` (paths only, no Aleph message, no download), wire it into `pool.executions`, route the production creation path through the spec translator, and reattach surviving VMs after a reboot from on-disk controller configs — all without the supervisor's machinery ever reading an Aleph message.
+
+**Architecture:** `VmExecution` gains a spec-driven construction path alongside the existing message path (no behavior change to the message path). `pool.create_a_vm_from_spec(spec)` mirrors `create_a_vm` but sources everything from the spec and writes the controller config via Phase 0.C's `build_qemu_configuration`, skipping the message-coupled `vm.configure()`. `InProcessSupervisor.create_vm` delegates to it. `orchestrator/run.py` (agent territory) translates eligible QEMU instances to a spec, then re-attaches the message to the created execution purely for its own consumers (operator API owner-auth, port forwarding) — the supervisor's machinery never reads it. Reboot-recovery (`pool.load_persistent_executions`) scans `*-controller.json` + systemd and reattaches via `spec_from_controller_configuration(config)` → `VmExecution.from_spec`, dropping the DB-message read entirely.
+
+**Tech Stack:** Python 3.11, asyncio, pydantic v2, pytest / pytest-asyncio. Test conventions per `tests/supervisor/`: sibling imports (no `tests/supervisor/__init__.py`), `monkeypatch`, ruff/isort/mypy green.
+
+**Scope:** PRs 1–4, stacked, each independently deployable. **Ownership is out of scope and intentionally so:** the supervisor never tracks who owns a VM. The agent API gates HTTP calls against its own ownership store; the agent's operator API may be unavailable during a reboot, which is acceptable — separating the API lifecycle from the supervisor lifecycle is the point. PR 4 therefore does not preserve any message for the agent across a reboot.
+
+**Design doc:** `docs/plans/2026-05-29-message-free-supervisor-create-reboot-design.md`
+
+**Two resolved design decisions** (called out by the controller before planning):
+1. **PR sequencing:** PR 3 (route production through the spec) lands before any reboot-recovery work, so the create path is fully wired and routed first.
+2. **Resources holder:** a `from_spec` classmethod on the existing `AlephQemuResources` (not a parallel dataclass). The `create()` path already does `assert isinstance(self.resources, AlephQemuResources)` and `pool.calculate_available_disk` reads `resources.get_disk_usage_delta()`; keeping one type avoids duck-typing and a widened annotation.
+
+---
+
+## File Structure
+
+| File | Responsibility | Change |
+|------|----------------|--------|
+| `src/aleph/vm/controllers/firecracker/executable.py` | Resources holder base | Make `message_content` nullable so a spec holder carries no message |
+| `src/aleph/vm/controllers/qemu/instance.py` | QEMU resources holder | Add `AlephQemuResources.from_spec` (no download) |
+| `src/aleph/vm/models.py` | `VmExecution` | Optional `message`/`original` + `spec`; `from_spec`; spec-aware properties; spec branches in `prepare()`/`create()`; `start(write_config=...)`; `save()` skip; `allocated_*` props |
+| `src/aleph/vm/pool.py` | Pool / lifecycle | `create_a_vm_from_spec`; nullability guards in `check_admission` + `get_executions_by_address` |
+| `src/aleph/vm/orchestrator/views/__init__.py` | Monitoring views | Guard `VmType.from_message_content` against message-less executions |
+| `src/aleph/vm/orchestrator/tasks.py` | Payment/notify tasks | Guard `execution.message.address` comprehensions |
+| `src/aleph/vm/supervisor/inprocess.py` | Supervisor boundary | Implement `create_vm` (delegate to pool) |
+| `src/aleph/vm/orchestrator/run.py` | Agent creation path | Route eligible QEMU instances through the spec; re-attach message for agent consumers |
+| `src/aleph/vm/supervisor/qemu_build.py` | Spec ⇄ config | Add `spec_from_controller_configuration` (inverse of `build_qemu_configuration`) |
+| `src/aleph/vm/pool.py` | Reboot-recovery | Rewrite `load_persistent_executions` to scan configs; config-driven `_restore_running_execution_from_config` / `_restore_network` / `_handle_dead_controller`; drop DB-message reads |
+
+---
+
+## PR 1 — `VmExecution` spec-constructible (message-free internals)
+
+### Task 1: `AlephQemuResources.from_spec` (no download)
+
+**Files:**
+- Modify: `src/aleph/vm/controllers/firecracker/executable.py:94-138` (make `message_content` nullable)
+- Modify: `src/aleph/vm/controllers/qemu/instance.py:40-96` (add `from_spec`)
+- Test: `tests/supervisor/test_supervisor_spec_resources.py`
+
+- [ ] **Step 1: Write the failing test**
+
+Create `tests/supervisor/test_supervisor_spec_resources.py`:
+
+```python
+"""Tests for AlephQemuResources.from_spec — the message-free resources holder."""
+
+from __future__ import annotations
+
+from pathlib import Path
+
+from aleph.vm.controllers.qemu.instance import AlephQemuResources
+from aleph.vm.supervisor.types import (
+    Backend,
+    CreateVmSpec,
+    DiskFormat,
+    DiskRole,
+    DiskSpec,
+    GpuSpec,
+    NetworkConfig,
+    PciAddress,
+    VmId,
+)
+
+
+def _spec() -> CreateVmSpec:
+    return CreateVmSpec(
+        vm_id=VmId("deadbeef" * 8),
+        backend=Backend.QEMU,
+        kernel_path=Path(""),
+        initrd_path=Path(""),
+        disks=[
+            DiskSpec(path=Path("/data/rootfs.qcow2"), readonly=False, format=DiskFormat.QCOW2, role=DiskRole.ROOTFS),
+            DiskSpec(
+                path=Path("/data/extra.img"),
+                readonly=True,
+                format=DiskFormat.RAW,
+                role=DiskRole.EXTRA,
+                mount="/mnt/data",
+            ),
+        ],
+        vcpus=4,
+        memory_mib=2048,
+        tee=None,
+        network=NetworkConfig(internet_access=True, requested_ipv6="", ipv6_prefix_len=0),
+        gpus=[GpuSpec(pci_host=PciAddress("0000:01:00.0"), supports_x_vga=True)],
+        numa_node=None,
+        persistent=True,
+    )
+
+
+def test_from_spec_populates_paths_without_download():
+    resources = AlephQemuResources.from_spec(_spec(), namespace="ns")
+
+    assert resources.message_content is None
+    assert resources.namespace == "ns"
+    assert resources.rootfs_path == Path("/data/rootfs.qcow2")
+    assert len(resources.volumes) == 1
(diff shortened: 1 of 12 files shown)


foxpatch-aleph (approved):
Well-structured PR that cleanly decouples the supervisor from Aleph messages by making VmExecution constructible from a spec, routing creation through a message-free path, and implementing reboot-recovery from on-disk controller configs instead of the database. The design separates supervisor concerns (paths, resources, execution) from agent concerns (message translation, owner auth, billing) with no behavior change to existing paths. Tests are thorough, mocking prevents side effects, and all existing message-path code remains untouched.

**`src/aleph/vm/models.py`** (line 151): The `assert self.message is not None` here is correct as a precondition guard — it will only fire if agent code calls `fetch_port_redirect_config_and_setup` without attaching the message first (which run.py does correctly). Consider a comment noting this is an agent-responsibility precondition.

**`src/aleph/vm/models.py`** (line 430): `from_spec` passes `snapshot_manager` and `systemd_manager` as keyword-only args, which matches how `pool.py` calls it. Good defensive API design.

foxpatch-aleph (commented):

foxpatch-aleph (commented):

foxpatch-aleph (approved):
This is a thorough and well-structured implementation plan that closely follows the design document. The code already implemented in this PR matches the plan well. The plan correctly separates concerns between supervisor (message-free, path-only) and agent (message ownership, port forwarding from user-settings). The two 'at most one of message or spec' invariant is cleanly enforced via Union type. The stacked PR structure (4 deployable PRs) is sensible. The main concern is a behavior regression in the reattach path (dropping fetch_port_redirect_config_and_setup without port_mappings-table fallback means port 22/SSH from user-settings won't be restored on fresh reboot), which should be documented as an accepted limitation or addressed. No correctness or security issues found.

**`docs/plans/2026-05-29-message-free-supervisor-create-plan.md`** (line 1): The reattach path drops `fetch_port_redirect_config_and_setup()` (which reads user-settings aggregate for port-forwarding configuration) and relies solely on `recreate_port_redirect_rules()` (which only restores pre-existing `mapped_ports` from the DB). If the `port_mappings` table is empty after a reboot (fresh host, wiped DB), the VM will reattach without SSH access (port 22) or any user-configured port forwarding. This is a behavioral regression from the current `_restore_running_execution`. Should be acknowledged as an accepted limitation of the message-free path, or addressed with a fallback mechanism.

**`docs/plans/2026-05-29-message-free-supervisor-create-plan.md`** (line 1): `_handle_dead_controller` (Task 11 Step 5) silently drops `record_usage()` compared to the existing `_handle_dead_execution`. This is technically correct since the message-free path creates no DB records, but should be explicitly noted as intentional.

**`docs/plans/2026-05-29-message-free-supervisor-create-plan.md`** (line 1): `get_executions_by_address()` (pool.py `:819-843`) accesses `execution.message.payment` and `execution.message.address` which will crash with `AttributeError` on message-free (spec) executions where `message` is `None`. The Task 6 guards cover the pool's iteration code (check_admission, get_executions_by_address iterations), but `get_executions_by_address` also reads `execution.message.payment` and `execution.message.address` inside the loop body — this needs a guard too, or specification-aware accessors. Note: the actual PR code in pool.py still has the old implementation; the plan should address this.

foxpatch-aleph (approved):
Clean, well-structured refactoring that makes VmExecution constructible from a message-free CreateVmSpec. The design cleanly introduces a discriminated union (MessageSpec | CreateVmSpec) for the spec field, extracts shared resource logic into controllers/resources.py, and adds nullability guards throughout. Tests cover the new spec path thoroughly. The one notable concern is the existing mutable default 'gpus = []' on AlephQemuResources, which is pre-existing and the spec path correctly avoids it by replacing rather than mutating. The repr crash when vm is None is also pre-existing. This is an excellent foundation for the stacked PRs to come.

**`src/aleph/vm/models.py`** (line 431): Pre-existing: __repr__ will crash with 'TypeError: type object argument must be an object, not NoneType' when accessed before create() sets self.vm. Consider accessing type(self.vm).__name__ only when self.vm is not None.

foxpatch-aleph (approved):
Well-structured, carefully tested implementation of message-free supervisor create and reboot-recovery. The code correctly distinguishes spec-driven from message-driven execution paths, maintains backward compatibility, and provides thorough test coverage. The design decisions around type narrowing via MessageSpec/CreateVmSpec, defensive guards in download methods, and message re-attachment in the agent layer are all sound.

**`src/aleph/vm/pool.py`** (line 208): Spec-built executions are skipped in the admission tally, meaning a mixed pool undercounts committed resources for these. However, spec-built VMs don't flow through `check_admission` currently, so this is safe in practice. Consider adding resource accounting for spec-built executions when the full migration is complete.

**`src/aleph/vm/models.py`** (line 557): The `write_config` parameter name is slightly ambiguous — it controls whether to skip the message-coupled `configure()` call, not just whether to write a file. A brief docstring on `start()` clarifying this would help future readers.

foxpatch-aleph (approved):
Clean architectural refactoring that makes VmExecution dual-path (message-driven or spec-driven without a message) while preserving full backward compatibility. The MessageSpec/CreateVmSpec union type cleanly encodes the invariant, all message-access paths are properly guarded, and the message-free path is comprehensively tested. No security concerns.

**`src/aleph/vm/models.py`** (line 488): The `to_dict()` method calls `**self.__dict__` which now includes a `MessageSpec | CreateVmSpec` `spec` field — neither is trivially JSON-serializable. This is invoked via `to_json()` → `record_usage()` → `save_execution_data()` during `stop()`. For spec-built executions on the current production code path this likely doesn't trigger (spec executions are always persistent and `record_usage` is called unconditionally at stop), but a non-persistent spec execution would break here. Consider either overriding `to_dict` for spec executions or adding a `__dict__` representation for `MessageSpec`/`CreateVmSpec`.

foxpatch-aleph (changes requested):
Well-structured PR with a clean design: `MessageSpec` union type correctly enforces the message-vs-spec invariant, and the accessor properties maintain backward compatibility. However, there is one blocking bug: `pool.py:_restore_network` still calls `VmType.from_message_content(execution.message)` at line 588, which will raise `TypeError` when reattaching a message-free (spec-built) execution during reboot-recovery. This needs to derive `VmType` from the spec/backend instead of the message. Additionally, `check_admission` silently skips spec-built executions (lines 206-210), which is intentional per the design but should be more clearly documented as leaving admission responsibility to the agent, since an oversubscription of spec-built VMs would bypass the pool's guard entirely.

**`src/aleph/vm/pool.py`** (line 588): `VmType.from_message_content(execution.message)` will raise `TypeError` when `execution.message` is `None` (spec-built execution during reattach). Derive `VmType` from the backend/config instead: use `VmType.instance` for QEMU, similar to the QEMU-instance value used elsewhere for reattach. The confiential QEMU executable also uses `instance`. The design doc §7 notes this line "derives `VmType` from the config/backend, not `VmType.from_message_content(execution.message)`" but the PR does not implement that change.

**`src/aleph/vm/pool.py`** (line 232): Consider a more prominent doc note on the `check_admission` guard at line 206-210. The method's docstring says it's "authoritative when called from `create_a_vm`" but the skip makes it non-authoritative for spec-built VMs. Since spec-built VMs bypass `create_a_vm` entirely, the agent must enforce its own admission before calling `create_a_vm_from_spec`. The current `# skip them in this message-based tally` comment understates this gap.

**`src/aleph/vm/orchestrator/run.py`** (line 63): This production creation path calls `pool.create_a_vm` with the message. The design says PR 3 routes eligible QEMU instances through the spec, but this file is not updated in this PR. Assuming this change lands in a stacked PR. If this is intentional deferral, consider a TODO comment here noting the planned spec path.

**`src/aleph/vm/models.py`** (line 355): The `message` and `original` properties at lines 355-362 delegate to `self.spec.message` / `self.spec.original` but `CreateVmSpec` has no `.message` or `.original` attribute. The `isinstance` guard protects this correctly (`CreateVmSpec` returns `None`), so this is safe. However, it may be worth a brief docstring note that these are safe only because of the `isinstance` check — a reader might miss the subtlety.

**`src/aleph/vm/models.py`** (line 862): The `save()` guard at line 862 correctly skips DB persistence for spec-built VMs. This means the only persistent record is the controller JSON config. Confirm that `record_usage()` at line 909 (called from `stop()`) is safe to call on message-free executions — it calls `delete_record(execution_uuid=str(self.uuid))` which would be a no-op if no record exists, and `save_execution_data` for logging. This looks safe but worth double-checking that `self.uuid` is always set (it is, in `__init__`).

**`src/aleph/vm/controllers/resources.py`** (line 61): The `disk_usage_delta` signature update to `message_content: ExecutableContent | None` with the guard at line 71 is correct and well-handled. However, for spec-built resources, the `size_mib` on volumes is set to `None` (from `AlephQemuResources.from_spec`), so the volume loop at line 79 will skip those volumes (`if not host_volume.size_mib`), correctly contributing `0`. This is the right behavior since spec-built VMs don't reserve disk through the pool.

**`tests/supervisor/test_supervisor_spec_execution.py`** (line 76): The `test_prepare_builds_resources_without_download` test is clean — monkeypatching `download_all` to a sentinel and asserting it was never called correctly validates the no-download invariant.

**`tests/supervisor/test_supervisor_spec_execution.py`** (line 113): The `test_start_skips_configure_and_save_for_spec` test at line 131 passes `write_config=False` to `start()` — this is correct for the spec path where `configure()` is message-coupled. Good that the test explicitly validates this keyword.

foxpatch-aleph (changes requested):
This is a well-structured, thoroughly documented PR that successfully introduces spec-constructible VmExecution alongside the existing message path. The MessageSpec/CreateVmSpec discriminated union pattern is clean and the tests are comprehensive. However, a resource accounting bug in check_admission is the primary concern: spec-built executions are skipped entirely from the memory/vCPU admission tally, meaning they contribute zero to the committed-resource count even though they consume host resources. This must be fixed (read from execution.allocated_memory_mib / allocated_vcpus) or explicitly documented as intentional for the spec path.

**`src/aleph/vm/pool.py`** (line 211): Spec-built executions are skipped entirely from admission accounting (`if message is None: continue`). This means a spec-built VM using 32GB RAM contributes zero to `committed_instance_memory_mib`, allowing the pool to over-admit. Either: (a) replace this block with `memory = execution.allocated_memory_mib; vcpus = execution.allocated_vcpus; if not memory and not vcpus: continue` (using the properties you already added), or (b) document explicitly in the method docstring + PR description that the spec path bypasses admission as a transitional limitation.

**`src/aleph/vm/models.py`** (line 488): `to_dict()` does `**self.__dict__`, which now includes a `spec` attribute that is either a `MessageSpec` or `CreateVmSpec` dataclass. `CreateVmSpec` contains `Path`, `Enum`, and NewType fields that are not JSON-serializable. If `to_dict()` / `to_json()` is called on a spec-built execution (e.g. monitoring endpoint), it will fail to serialize. Consider adding an `__init__` or `__getstate__` on `CreateVmSpec` or overriding `to_dict()` to handle serialization of both spec types.

**`src/aleph/vm/controllers/qemu/instance.py`** (line 65): `get_disk_usage_delta()` returns 0 for spec-built resources because `message_content is None` and volumes have `size_mib=None`. This works correctly by coincidence of the downstream logic, but it would be helpful to add a brief comment on this method noting that the return value is intentionally 0 for the spec path (no reservation-based admission).

**`src/aleph/vm/supervisor/types.py`** (line 166): `rootfs` and `require_rootfs` both perform a local import of `InvalidBackendError` to avoid circular imports. The comment explaining this is good. Consider also importing at the bottom of the module with a TYPE_CHECKING guard as an alternative pattern, which avoids the repeated local imports — though the current approach is also fine for a single accessor pair.

**`src/aleph/vm/models.py`** (line 541): `prepare_gpus()` uses `assert isinstance(self.spec, MessageSpec)` which will crash if called on a spec-built execution. This is intentional (GPUs are pre-assigned in the spec path), but a more descriptive error message would help debugging: `raise TypeError("prepare_gpus is message-only; GPU assignment is baked into CreateVmSpec")` instead of a bare AssertionError.

foxpatch-aleph (approved):
Clean, well-structured implementation of spec-constructible QEMU resources. Correctly widens message_content to Optional without affecting the existing message path, and the from_spec classmethod is a textbook example of how to add a spec-driven construction path. The test coverage on from_spec, disk_usage_delta=0, and execution-level integration is thorough.

**`src/aleph/vm/controllers/qemu/instance.py`** (line 100): consider adding an inline comment noting that `kernel_image_path` is also left as the class default (set via `settings.LINUX_PATH` in `from_spec`) for completeness, since `rootfs_path`, `volumes`, and `gpus` are explicitly assigned.

foxpatch-aleph (approved):
Clean, well-structured refactoring that makes VmExecution constructible/operable from a message-free CreateVmSpec alongside the legacy message path. The MessageSpec/CreateVmSpec union type on execution.spec is a good design that allows fully typed property dispatch. All message-only entrypoints (fetch_port_redirect_config_and_setup, prepare_gpus, watch_for_updates, save) are correctly guarded with isinstance checks. The test coverage (from_spec creation, properties, prepare-without-download, create-builds-qemu-instance, start-skips-configure) is excellent and covers all spec-path behaviours. No behaviour change to message-driven execution. The one named issue below is a minor documentation gap.

**`src/aleph/vm/models.py`** (line 665): The `start()` method's docstring should document the `write_config` parameter. Currently it's undocumented, which breaks the convention set by the rest of the class.

foxpatch-aleph (approved):
Clean decoupling of the supervisor creation path from Aleph messages. The design is sound — VmExecution gains a `spec` field that is either a `MessageSpec` (legacy) or a `CreateVmSpec` (new supervisor-owned path). Properties are correctly routed through the spec, the prepare/create/start machinery short-circuits for spec-built executions without downloads or DB persistence, and the existing message path is untouched. Tests cover all branches comprehensively. The `is_confidential` property returns `True` for `Backend.QEMU_SEV` specs despite `create()` not producing confidential instances — but `create_a_vm_from_spec()` rejects non-`Backend.QEMU`, so this is unreachable and documented as out of scope. No correctness or security issues found.

**`src/aleph/vm/models.py`** (line 470): When both `vm_spec` and `message` are passed, the spec silently wins without warning. This is safe because no caller does this, but I'd suggest logging a warning or adding a guard assertion for defensive programming.

**`src/aleph/vm/models.py`** (line 383): `is_confidential` on the spec path returns True for Backend.QEMU_SEV, but `create()` only produces non-confidential `AlephQemuInstance`. The inconsistency is gated by `create_a_vm_from_spec()` rejecting QEMU_SEV, but adding a comment referencing that gate would prevent future confusion.

PR #956, opened 2026-05-31
Message-free supervisor (3/4): route production creation through the spec

odesenfans:
Part **3 of 4** of the message-free supervisor backport. Base: `od/msgfree-2-pool-create` (#955).

> Stacked on #955 (which is stacked on #954) — review/merge those first.

Routes the production creation path through the spec for eligible workloads.

### Changes
- `orchestrator/run.py` `create_vm_execution`: for an eligible message (a non-confidential QEMU instance with no GPU requirement) it now `build_create_vm_spec(...)` → `pool.create_a_vm_from_spec(...)`, then **re-attaches** the message to the created execution purely for the agent's own consumers (operator-API owner-auth, port forwarding) — the supervisor machinery that just created the VM never read it. This is the strangler seam: the supervisor acts message-free; the agent decorates its in-process view.
- A `_is_spec_eligible` helper mirrors `build_create_vm_spec`'s validation. Everything else (programs, Firecracker, confidential, GPU instances) keeps the legacy `pool.create_a_vm` path **unchanged**.

### Stack
1. spec-constructible `VmExecution` → `dev` (#954)
2. pool create path → #954 (#955)
3. **(this PR)** route production creation through the spec → #955
4. reboot-recovery from on-disk configs → 3

New test: `test_supervisor_run_routing.py`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #955, opened 2026-05-31
Message-free supervisor (2/4): pool create path

odesenfans:
Part **2 of 4** of the message-free supervisor backport. Base: `od/msgfree-1-spec-execution` (#954).

> Stacked on #954 — review/merge that first. The diff here is only this PR's commits.

Wires the message-free create path into the pool and the supervisor boundary.

### Changes
- `pool.create_a_vm_from_spec(spec)` — the supervisor's message-free creation path: `VmExecution.from_spec` → `prepare()` (no download) → tap → controller config via `build_qemu_configuration` + `save_controller_configuration` → `create()` → `start(write_config=False)`. Holds `creation_lock`, mirrors `create_a_vm`'s teardown-on-error and `_schedule_forget_on_stop`. Port-forward fetch is intentionally omitted (agent concern).
- `InProcessSupervisor.create_vm` delegates to it (replacing the `NotImplementedSupervisorError` stub); conformance suite updated to genuinely exercise the remaining stub methods.
- A message-free execution can now live in `pool.executions`, so pool-wide iterations are guarded: new `allocated_memory_mib`/`allocated_vcpus` properties feed `check_admission`; `get_executions_by_address`, the monitoring views, and the payment/notify tasks skip/guard executions with no message.

### Stack
1. spec-constructible `VmExecution` → `dev` (#954)
2. **(this PR)** pool create path → #954
3. route production creation through the spec (`run.py`) → 2
4. reboot-recovery from on-disk configs → 3

New tests: `test_supervisor_spec_pool_guards.py`, `test_supervisor_spec_pool_create.py`, `test_supervisor_inprocess_create.py`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #957, opened 2026-05-31
Message-free supervisor: reboot-recovery from on-disk configs

odesenfans:
Part **4 of 4** of the message-free supervisor backport. Base: `od/msgfree-3-routing` (#956).

> Top of the stack — stacked on #956 → #955 → #954. Review/merge those first.

Makes reboot-recovery message-free: the supervisor reattaches surviving VMs from on-disk controller configs + systemd, dropping the DB path entirely.

### Changes
- `spec_from_controller_configuration(config)` — the inverse of `build_qemu_configuration`: reconstruct a `CreateVmSpec` from an on-disk `<hash>-controller.json`.
- Config-driven reattach helpers `_restore_running_execution_from_config` and `_handle_dead_controller`; `_restore_network` made message-free (`VmType.instance`).
- `pool.load_persistent_executions` rewritten to scan `EXECUTION_ROOT/*-controller.json` + batch systemd active-state, with the original vm_id-dedup guard preserved. Removes the DB-based path (`get_execution_records`, `ExecutionRecord`, and 4 other now-dead imports) and the old `_restore_running_execution`/`_handle_dead_execution`.
- `start_watching_for_updates`/`watch_for_updates` no-op for message-free executions instead of asserting (otherwise every reattached VM would fire an assertion when the scheduler re-issues `start_persistent_vm` after a reboot).

### By design
A reattached VM has no Aleph message, so operator-API owner-auth and Aleph update-watching are inert after a reboot until the agent re-fetches the allocation from the scheduler. This is the intended separation of the API lifecycle from the supervisor lifecycle; the changes here make those paths degrade gracefully rather than crash.

### Stack
1. spec-constructible `VmExecution` → `dev` (#954)
2. pool create path → #954 (#955)
3. route production creation through the spec → #955 (#956)
4. **(this PR)** reboot-recovery from on-disk configs → #956

New tests: `test_supervisor_spec_from_config.py`, `test_supervisor_reattach.py`; `test_vm_id_collision.py` rewritten to the config-scan path.

### DIFF
--- a/src/aleph/vm/models.py
+++ b/src/aleph/vm/models.py
@@ -851,10 +851,19 @@ class VmExecution:
             logger.info("%s stopped", self)
 
     def start_watching_for_updates(self, pubsub: PubSub):
+        if not isinstance(self.spec, MessageSpec):
+            # Message-free (spec-built / reattached) executions have no Aleph
+            # message to watch for updates. After a reboot the agent re-fetches
+            # the allocation and re-establishes watching; the supervisor's own
+            # reattach does not. No-op rather than scheduling a task that fails.
+            return
         if not self.update_task:
             self.update_task = create_task_log_exceptions(self.watch_for_updates(pubsub=pubsub))
 
     async def watch_for_updates(self, pubsub: PubSub):
+        # Message-only entrypoint. start_watching_for_updates already no-ops for
+        # spec-built / reattached executions, so reaching here with one is a
+        # programming error: fail loud rather than silently watching nothing.
         if not isinstance(self.spec, MessageSpec):
             raise TypeError("watch_for_updates is message-only; spec-built executions have no message to update")
         original = self.spec.original
--- a/src/aleph/vm/pool.py
+++ b/src/aleph/vm/pool.py
@@ -541,7 +541,7 @@ class VmPool:
         """
         try:
             config_paths = sorted(settings.EXECUTION_ROOT.glob("*-controller.json"))
-        except Exception:
+        except OSError:
             logger.warning("Failed to enumerate controller configs", exc_info=True)
             config_paths = []
 
@@ -593,7 +593,7 @@ class VmPool:
             await self.update_domain_mapping(force_update=True)
         logger.info("Loaded %d executions", len(self.executions))
 
-    async def _restore_network(self, _execution: VmExecution, vm_id: int, vm_hash: ItemHash) -> TapInterface | None:
+    async def _restore_network(self, vm_id: int, vm_hash: ItemHash) -> TapInterface | None:
         """Restore tap interface, NDP proxy, and nftables rules for a VM."""
         if not self.network:
             return None
@@ -634,7 +634,7 @@ class VmPool:
         logger.info("Loading existing mapped_ports %s", execution.mapped_ports)
 
         await execution.prepare()  # builds resources from the spec; no download
-        tap_interface = await self._restore_network(execution, vm_id, vm_hash)
+        tap_interface = await self._restore_network(vm_id, vm_hash)
 
         vm = execution.create(vm_id=vm_id, tap_interface=tap_interface, prepare=False)
         await vm.start_guest_api()
@@ -696,7 +696,7 @@ class VmPool:
         """
         try:
             config_files = list(settings.EXECUTION_ROOT.glob("*-controller.json"))
-        except Exception:
+        except OSError:
             logger.warning("Failed to enumerate controller configs", exc_info=True)
             return
 
--- a/src/aleph/vm/pool.py
+++ b/src/aleph/vm/pool.py
@@ -21,7 +21,10 @@ from aleph_message.models import (
 from pydantic import TypeAdapter
 
 from aleph.vm.conf import settings
-from aleph.vm.controllers.configuration import save_controller_configuration
+from aleph.vm.controllers.configuration import (
+    Configuration,
+    save_controller_configuration,
+)
 from aleph.vm.controllers.firecracker.snapshot_manager import SnapshotManager
 from aleph.vm.network.hostnetwork import Network, make_ipv6_allocator
 from aleph.vm.network.interfaces import TapInterface
@@ -38,7 +41,7 @@ from aleph.vm.resources import (
     get_gpu_devices,
 )
 from aleph.vm.supervisor.errors import InvalidBackendError
-from aleph.vm.supervisor.qemu_build import build_qemu_configuration
+from aleph.vm.supervisor.qemu_build import build_qemu_configuration, spec_from_controller_configuration
 from aleph.vm.supervisor.types import Backend, CreateVmSpec
 from aleph.vm.systemd import SystemDManager
 from aleph.vm.utils import get_message_executable_content
@@ -646,12 +649,13 @@ class VmPool:
         self.executions[vm_hash] = execution
         execution.record = saved_execution
 
-    async def _restore_network(self, execution: VmExecution, vm_id: int, vm_hash: ItemHash) -> TapInterface | None:
+    async def _restore_network(self, _execution: VmExecution, vm_id: int, vm_hash: ItemHash) -> TapInterface | None:
         """Restore tap interface, NDP proxy, and nftables rules for a VM."""
         if not self.network:
             return None
 
-        vm_type = VmType.from_message_content(execution.message)
+        # Reattach is QEMU-instance only; the message is gone by design.
+        vm_type = VmType.instance
         tap_interface = await self.network.prepare_tap(vm_id, vm_hash, vm_type)
 
         if not self.network.interface_exists(vm_id):
@@ -667,6 +671,54 @@ class VmPool:
 
         setup_nftables_for_vm(vm_id, interface=tap_interface)
         return tap_interface
+
+    async def _restore_running_execution_from_config(
+        self, config: Configuration, vm_id: int, vm_hash: ItemHash
+    ) -> None:
+        """Rebuild in-memory state for a VM whose controller is still active.
+
+        Sourced entirely from the on-disk controller config -- message-free.
+        """
+        spec = spec_from_controller_configuration(config)
+        execution = VmExecution.from_spec(
+            spec,
+            snapshot_manager=self.snapshot_manager,
+            systemd_manager=self.systemd_manager,
+        )
+
+        execution.mapped_ports = await get_port_mappings(vm_hash)
+        logger.info("Loading existing mapped_ports %s", execution.mapped_ports)
+
+        await execution.prepare()  # builds resources from the spec; no download
+        tap_interface = await self._restore_network(execution, vm_id, vm_hash)
+
+        vm = execution.create(vm_id=vm_id, tap_interface=tap_interface, prepare=False)
+        await vm.start_guest_api()
+        execution.ready_event.set()
+        execution.times.started_at = datetime.now(tz=timezone.utc)
+
+        self._schedule_forget_on_stop(execution)
+
+        if vm.support_snapshot and self.snapshot_manager:
+            await self.snapshot_manager.start_for(vm=execution.vm)
+
+        if execution.mapped_ports:
+            await execution.recreate_port_redirect_rules()
+
+        self.executions[vm_hash] = execution
+
+    async def _handle_dead_controller(self, config: Configuration) -> None:
+        """Stop the stale controller service for a VM that is no longer active.
+
+        The orphan controller config is removed by _cleanup_orphan_resources
+        once the VM is absent from self.executions.
+        """
+        service_name = f"aleph-vm-controller@{config.vm_hash}.service"
+        try:
+            self.systemd_manager.stop_and_disable(service_name)
+            logger.info("Stopped and disabled stale controller service %s", service_name)
+        except Exception:
+            logger.warning("Failed to stop/disable stale controller %s", service_name, exc_info=True)
 
     async def _handle_dead_execution(self, execution: VmExecution, saved_execution: ExecutionRecord) -> None:
         """Record usage for a dead execution and stop its controller service."""
--- a/src/aleph/vm/pool.py
+++ b/src/aleph/vm/pool.py
@@ -27,17 +27,14 @@ from aleph.vm.controllers.configuration import (
 from aleph.vm.controllers.firecracker.snapshot_manager import SnapshotManager
 from aleph.vm.network.hostnetwork import Network, make_ipv6_allocator
 from aleph.vm.network.interfaces import TapInterface
-from aleph.vm.orchestrator.metrics import (
-    get_port_mappings,
-)
+from aleph.vm.orchestrator.metrics import get_port_mappings
 from aleph.vm.orchestrator.utils import update_aggregate_settings
-from aleph.vm.resources import (
-    GpuDevice,
-    InsufficientResourcesError,
-    get_gpu_devices,
-)
+from aleph.vm.resources import GpuDevice, InsufficientResourcesError, get_gpu_devices
 from aleph.vm.supervisor.errors import InvalidBackendError
-from aleph.vm.supervisor.qemu_build import build_qemu_configuration, spec_from_controller_configuration
+from aleph.vm.supervisor.qemu_build import (
+    build_qemu_configuration,
+    spec_from_controller_configuration,
+)
 from aleph.vm.supervisor.types import Backend, CreateVmSpec
 from aleph.vm.systemd import SystemDManager
 from aleph.vm.vm_type import VmType
--- a/src/aleph/vm/pool.py
+++ b/src/aleph/vm/pool.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
 import asyncio
-import json
 import logging
 import pathlib
 import shutil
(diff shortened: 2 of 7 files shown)


foxpatch-aleph (approved):
Well-structured PR implementing message-free supervisor VM creation (Phase 0.C). Cleanly separates message-ful and message-free code paths with proper guards for safe reattachment. vm_id collision protection prevents tap interface conflicts during reboot recovery. All code paths handle edge cases correctly with no security issues. Minor observations about double-stops on dead controllers and redundant str() wraps are non-blocking.

**`src/aleph/vm/pool.py`** (line 564): `config.vm_hash` is already typed as `str` in the Configuration dataclass, so `str(config.vm_hash)` is redundant. Not a bug — just minor style nit.

**`src/aleph/vm/pool.py`** (line 700): Dead controller services already stopped by `_handle_dead_controller` at line 658, and now `_cleanup_orphan_controller_configs` calls `stop_and_disable` again at line 706. Harmless (systemd idempotency) but could be optimized with a skip for already-handled hashes.

foxpatch-aleph (approved):
Clean refactoring from DB-backed reboot-recovery to config-driven reattach. Correctly scans controller configs from disk, batch-queries systemd for active states, handles duplicate vm_ids by restoring only the first active config, and degrades gracefully for message-free executions (no-op update watching). Tests cover roundtrip, reattach helpers, VM ID collision scenarios, and the no-op update watch. Minor missing test coverage for exception paths in load_persistent_executions (glob failure, missing configs), but these are defensive paths and not blocking.

**`src/aleph/vm/pool.py`** (line 640): `_schedule_forget_on_stop` is called before the execution is added to `self.executions` (line 648). This works because the forget task doesn't access self.executions directly, but it's slightly inconsistent with `create_a_vm_from_spec` (line 459) which adds to executions first. Not a bug, just a style difference.

**`src/aleph/vm/pool.py`** (line 538): Missing test coverage: the `except Exception` path when `EXECUTION_ROOT.glob()` raises. Also no test for `load_controller_configuration` returning None (empty/missing config file). These are defensive paths that won't cause crashes but are worth covering in future test iterations.

foxpatch-aleph (approved):
Clean, well-structured PR that removes the DB/message dependency from reboot-recovery. The on-disk config files become the source of truth for reattaching VMs after a restart. The spec_from_controller_configuration function correctly inverts build_qemu_configuration, and the vm_id collision logic properly guards against duplicate tap interfaces. Tests are comprehensive, covering the collision scenarios, the config-to-spec round-trip, and the message-free no-op guard. A few minor observations (unused imports in tests, a missing negative test for port mappings) do not block approval.

**`tests/supervisor/test_supervisor_reattach.py`** (line 5): Unused imports: Path, Backend, DiskFormat, DiskRole, DiskSpec, NetworkConfig, VmId are only used inside _spec() which is defined in the same file. Not a blocker but adds noise.

**`tests/supervisor/test_vm_id_collision.py`** (line 10): Unused import: Path is imported but not used in this file.

**`src/aleph/vm/pool.py`** (line 628): Consider adding a test case for _restore_running_execution_from_config when get_port_mappings returns non-empty data, to exercise the recreation of port redirect rules. The current test only covers the empty case.

**`src/aleph/vm/pool.py`** (line 615): The _restore_network method receives _execution but never uses it (all info comes from vm_id/vm_hash). Consider removing the parameter in a follow-up cleanup.

foxpatch-aleph (commented):

foxpatch-aleph (approved):
Correctly eliminates the DB-dependent execution recovery path, replacing it with config-driven reattachment. The vm_id collision bug is fixed by deduplicating across on-disk configs rather than stale DB records. Batch D-Bus calls improve efficiency, and message-free executions no longer crash on watch_for_updates. Code is clean, well-tested, and properly factored.

**`src/aleph/vm/pool.py`** (line 583): Consider if _cleanup_orphan_resources should be called before or after the per-config loop in load_persistent_executions. Currently it's called after — this means stale nft rules from a dead config persist until the next execution of load_persistent_executions. This is minor and pre-existing behavior, but worth noting.

foxpatch-aleph (approved):
Clean, focused PR that rewrites reboot-recovery to be entirely message-free by loading VMs from on-disk controller configs instead of database records. The architecture is sound, the round-trip test for spec_from_controller_configuration is thorough, and the vm_id collision guard is properly preserved. The only notable risk is a potential unhandled InvalidBackendError for non-QEMU configs during recovery, but this is a known phased-approach limitation.

**`src/aleph/vm/pool.py`** (line 563): If spec_from_controller_configuration raises InvalidBackendError (e.g. for Firecracker configs), this will abort the entire recovery loop. Consider catching the exception per-config to continue scanning remaining configs.

**`src/aleph/vm/supervisor/qemu_build.py`** (line 189): 

**`tests/supervisor/test_supervisor_reattach.py`** (line 56):

foxpatch-aleph (commented):

foxpatch-aleph (approved):
The PR adds comprehensive tests for the reboot-recovery feature (reattaching VMs from on-disk configs without messages). The tests cover: restoring running/failed controller services, duplicate VM ID handling, list of services restoration, spec roundtripping from controller configurations, and no-op behavior for message-free executions. The test code is well-structured, uses appropriate mocks, and covers the key edge cases.

**`tests/supervisor/test_supervisor_reattach.py`** (line 177): The `test_restore_list_of_services` test is tightly coupled to the implementation (it asserts specific `nft` commands and internal functions). Consider testing observable behavior (e.g., checking the VM is in the correct state after reattach) for better robustness against future refactoring.

foxpatch-aleph (commented):

foxpatch-aleph (approved):
Clean, well-structured refactor that removes the DB dependency from reboot-recovery by scanning on-disk controller configs. The code is readable and well-documented, the vm_id collision dedup is handled correctly, and the tests cover the key scenarios. The only genuine concern is the missing `execution.times.starting_at` in the reattach path (noted in comments), but this doesn't affect `is_running` behavior for persistent VMs since it delegates to systemd.

**`src/aleph/vm/pool.py`** (line 644): `execution.times.starting_at` is never set for reattached VMs (the old code restored it from `saved_execution.time_prepared`). `is_running` doesn't depend on it for persistent VMs (delegates to systemd), but the operator API or other consumers of `VmExecutionTimes` may see a gap. Consider setting `execution.times.starting_at = datetime.now(tz=timezone.utc)` alongside `started_at`.

**`src/aleph/vm/pool.py`** (line 565): The comment says 'only the first active one is restored' but doesn't clarify what defines 'first' — it's the filesystem sort order of hash-named files. This is deterministic on the same host but worth noting in the comment in case someone wonders why the ordering matters.

**`src/aleph/vm/supervisor/qemu_build.py`** (line 171): `spec_from_controller_configuration` hard-codes `DiskFormat.QCOW2` for rootfs and `DiskFormat.RAW` for extra volumes. This is a lossy roundtrip — the original `DiskSpec.format` from `build_qemu_configuration` is discarded. Acceptable since the spec is only used for runtime metadata on reattach (no re-download), but worth documenting in the docstring that disk formats are best-effort.

**`tests/supervisor/test_supervisor_reattach.py`** (line 49): Minor: `_bare_pool` in this file is essentially identical to `_make_pool` in `test_vm_id_collision.py`. Consider extracting into a shared conftest fixture to reduce duplication.

foxpatch-aleph (approved):
A well-structured PR that correctly implements config-driven reboot recovery for message-free supervisor executions. The design cleanly replaces the DB-based reattach path with on-disk controller configs, properly handles duplicate vm_ids, and makes update-watching gracefully no-op for message-free executions. Tests cover the key paths including round-trip config conversion, reattach, dead controller cleanup, and vm_id collision detection.

**`src/aleph/vm/pool.py`** (line 639): `execution.prepare()` is called before `execution.create()` — same order as in the old `_restore_running_execution` and in `create_a_vm_from_spec`. Correct.

**`src/aleph/vm/pool.py`** (line 645): `execution.times.started_at` is set (line 645) as a fresh timestamp but `starting_at` and `prepared_at` are not restored from the config. This is the intended design (the old code also dropped `time_defined`), but it means the execution's timing stats are partially reset on reattach. Not a bug, just a noted behavioral change from the DB path.

**`src/aleph/vm/pool.py`** (line 548): The `except Exception` guard is good — a missing directory or permissions error won't crash the supervisor on restart. However, the broad catch makes it impossible to distinguish a transient I/O error from a permanent configuration issue. Consider whether `FileNotFoundError`/`PermissionError` could be treated differently from `OSError` in a future iteration.

**`src/aleph/vm/pool.py`** (line 553): The glob-derived `vm_hash` is extracted via substring `config_path.name[:-len("-controller.json")]`. This is correct because the pattern is `*-controller.json`, but if the hash contained a dash this would still work correctly even though the extraction is purely character-based. A minor naming readability improvement for the future: `name.removesuffix("-controller.json")` (Python 3.9+) would be more self-documenting.

**`src/aleph/vm/supervisor/qemu_build.py`** (line 171): The `spec_from_controller_configuration` function is a clean inverse of `build_qemu_configuration`. The only asymmetry is `ssh_authorized_keys` and `cloud_init_drive_path` — these are not round-tripped because cloud-init is a one-shot operation that has already run. This is correct behavior.

**`src/aleph/vm/supervisor/qemu_build.py`** (line 211): `memory_mib=vm_cfg.mem_size_mb.count` — this correctly unwraps the typed `MiB` value. Matches how `build_qemu_configuration` wraps it at line 125. Round-trip verified by test.

**`src/aleph/vm/models.py`** (line 853): The no-op guard in `start_watching_for_updates` for non-MessageSpec executions is correct and well-documented. The `watch_for_updates` method keeps its own type guard as a defense-in-depth measure — good pattern.

**`tests/supervisor/test_vm_id_collision.py`** (line 59): The collision test uses `SimpleNamespace` for config objects and patches `load_controller_configuration` with a side-effect function keyed on vm_hash. The `_write_configs` function writes empty `{}` files to `tmp_path` — these are never actually parsed since `load_controller_configuration` is mocked. This is fine (the test is about the dedup logic, not config parsing), but the dummy file writes could be misleading to future readers. Consider adding a comment that the file content is irrelevant because the loader is mocked.

**`tests/supervisor/test_vm_id_collision.py`** (line 43): The three collision scenarios (duplicate active, duplicate dead, unique) are well-chosen and cover the state machine completely. Good test design.

**`tests/supervisor/test_supervisor_spec_from_config.py`** (line 46): The round-trip test is thorough — it verifies vm_id, backend, vcpus, memory, internet_access from interface_name, both disk roles, mount, read_only, and gpus. A minor addition would be to verify the `qemu_bin_path` is not required in the spec (it's not, since the reattached VM is already running). The test implicitly confirms this is correct.

**`tests/supervisor/test_supervisor_reattach.py`** (line 69): The reattach test uses monkeypatching at the module level (`aleph.vm.pool.spec_from_controller_configuration`) and on `VmExecution` methods. This is fragile if import order changes, but the tests are in the same package as the code under test and follow existing conventions. The assertions at lines 84-89 correctly validate that the execution is registered, has a spec, no message, and ready_event is set.

foxpatch-aleph (approved):
Clean, well-structured PR that replaces the DB-based reboot-recovery path with config-driven reattachment from on-disk controller configs. The logic is correct, edge cases (vm_id collisions, dead services, no-op update watching) are handled properly, security is unaffected, and test coverage is thorough. The diff removes more code than it adds, which is a strong positive signal.

**`src/aleph/vm/pool.py`** (line 550): The `vm_hash` extraction via `config_path.name[:-len("-controller.json")]` is a bit fragile — if the suffix format ever changes, this breaks silently. Consider extracting a small helper or constant. Non-blocking for this PR.

**`src/aleph/vm/pool.py`** (line 608): Nit: the binding `vm_hash = ItemHash(str(config.vm_hash))` shadows `vm_hash` from the loop header context. The `str()` call is a no-op since `config.vm_hash` is already typed as `str`, but the intent to be explicit is clear.

**`src/aleph/vm/pool.py`** (line 656): Docstring says "orphan controller config is removed by _cleanup_orphan_resources" — this is correct (it runs at line 630), but worth a quick sanity check that no exception between here and there could skip the cleanup. The cleanup runs unconditionally after the loop, so this is fine.

foxpatch-aleph (approved):
Sound config-driven reboot recovery that replaces DB-based execution restoration with on-disk controller configs and systemd state. Correctly handles duplicate vm_ids, gracefully degrades for message-free executions, and is well-covered by unit tests.

foxpatch-aleph (approved):
This PR cleanly replaces the DB-based reattach path with config-file-based reattachment for message-free supervisor operations. The logic is correct: it scans EXECUTION_ROOT for controller configs, batch-checks systemd service states, preserves vm_id dedup protection, and gracefully degrades update-watching for message-free executions. Code quality is high with proper error handling, clear comments, and well-updated tests.

**`src/aleph/vm/pool.py`** (line 543): The try/except around glob() is appropriate defensive coding; consider also wrapping the config loading loop in case a single corrupt JSON file doesn't prevent loading others.

foxpatch-aleph (approved):
Well-structured PR that completes the message-free supervisor refactor by replacing the database-driven reboot-recovery path with config-driven reattach from on-disk controller files. The logic is correct, the roundtrip inversion of build_qemu_configuration is faithful, the vm_id collision guard is preserved, and the stale-controller cleanup is appropriately handled. Tests cover the key paths including config reconstruction, reattach helpers, and duplicate vm_id handling. The removal of fetch_port_redirect_config_and_setup from the reattach path is the intended design (agent responsibility post-reboot).

**`src/aleph/vm/pool.py`** (line 571): `str(config.vm_hash)` is a no-op since vm_hash is already typed as `str`. Consider just `ItemHash(config.vm_hash)`.

**`src/aleph/vm/pool.py`** (line 599): The `_execution` parameter is unused. Consider removing it since the reattach path doesn't need the execution for network restoration.

foxpatch-aleph (changes requested):
Clean config-driven reboot-recovery. The only blocking issue is the unhandled exception from corrupt controller config files in load_persistent_executions — a single corrupt JSON file will crash the entire pool load and prevent all VMs from reattaching. Wrap the load_controller_configuration call in a try/except. The unused _execution parameter in _restore_network is a minor style point.

**`src/aleph/vm/pool.py`** (line 556): load_controller_configuration can raise on corrupt JSON (JSONDecodeError/ValidationError), which will abort the entire pool load. Wrap in try/except that logs and continues to prevent a single corrupt file from crashing all VM reattachments.

**`src/aleph/vm/pool.py`** (line 599): _execution parameter is unused since VmType.instance is now hardcoded. Consider removing it if no future path will provide it.

foxpatch-aleph (approved):
Well-structured PR that completes the message-free supervisor refactor. Reboot-recovery now reads on-disk controller configs instead of database records. Logic is correct: scans for *-controller.json files, batch-checks systemd active states, deduplicates by vm_id (preserving existing behavior), and properly handles dead controllers via a two-phase cleanup. Tests cover roundtrip config reconstruction, reattach helpers, vm_id collision scenarios, and the update-watching no-op path. The only notable change is the loss of starting_at/prepared_at timestamps (formerly from DB fields), which is a deliberate consequence of dropping the DB path and safe since the fields are Optional.

**`src/aleph/vm/pool.py`** (line 622): The `_restore_network` method accepts `_execution` as its first parameter but no longer uses it (uses `VmType.instance` directly). The underscore prefix signals this, but the parameter could be removed entirely since all callers pass the `vm_hash` and `vm_id` explicitly.

**`src/aleph/vm/pool.py`** (line 645): `prepared_at` and `starting_at` are no longer set on `execution.times`. The old code set them from `saved_execution.time_prepared`/`time_defined` (DB fields). Since these are Optional with None defaults this is safe, but consider whether these timestamps are needed for any downstream consumers (metrics, operator API).

**`src/aleph/vm/pool.py`** (line 545): The `try/except Exception` wrapping `settings.EXECUTION_ROOT.glob(...)` is overly broad. Since `settings.EXECUTION_ROOT` is a `Path`, `glob()` can raise `OSError` (unlikely) or `PermissionError`. Consider catching only `OSError` to avoid masking programming errors.

foxpatch-aleph (approved):
Sound architectural change that decouples reboot-recovery from the database, replacing it with on-disk controller config scanning. The implementation is clean, well-tested, and correctly handles the vm_id deduplication guard against stale configs. No correctness or security issues found.

**`src/aleph/vm/pool.py`** (line 596): Consider removing the `_execution` parameter from `_restore_network` since it's no longer used (previously derived VmType from execution.message). The caller `_restore_running_execution_from_config` already has the execution reference if needed.

**`src/aleph/vm/supervisor/qemu_build.py`** (line 204): Consider adding `ssh_authorized_keys` to the reconstructed spec for completeness. Currently set via the default in `CreateVmSpec`. If `spec_from_controller_configuration` is ever used outside the reattach flow (e.g., for migration), missing SSH keys could break cloud-init.

foxpatch-aleph (approved):
Well-structured PR that cleanly removes the DB dependency from reboot-recovery by switching to on-disk controller configs. The inverse spec-from-config function is correctly implemented, the pool reattach logic handles vm_id collisions properly, and the watcher no-op for message-free executions prevents crashes. The tests cover the main paths well. Minor nits: unnecessary str() call around config.vm_hash, and the VmType.instance hardcode in _restore_network could use a brief rationale comment.

**`src/aleph/vm/pool.py`** (line 568): Nit: `config.vm_hash` is already a `str` per Configuration model definition (configuration.py:92). The `str()` call is unnecessary.

**`src/aleph/vm/pool.py`** (line 601): The `VmType.instance` hardcode is correct (only QEMU instances persist controller configs), but worth a brief comment explaining why reattach is always instance-type. A Firecracker execution calling `save_controller_configuration` would also produce a controller.json, yet this line assumes instance.

PR #958, opened 2026-05-31
refactor(controllers): untangle QEMU resources from Firecracker

odesenfans:
Follow-up to **#954**, addressing the resource-class inheritance smell that #954's message-free path exposed. Base: `od/msgfree-1-spec-execution` (#954).

> Stacked on #954 — review/merge that first. The diff here is only this PR's single commit.

### The problem
`AlephQemuResources` inherited from `AlephFirecrackerResources` purely to reuse host-resource mechanics (volumes, disk accounting, kernel path) — even though a QEMU VM is not a Firecracker VM. Because the shared base owned the single `message_content` declaration, making it optional for the message-free QEMU spec path in #954 had to widen it to `ExecutableContent | None` on the **base**, which leaked onto the Firecracker subclasses (`AlephProgramResources`, `AlephInstanceResources`) — code that *always* has a message. That added **9 `mypy` union-attr errors** at their unconditional dereference sites.

mypy rejects narrowing/widening an inherited attribute in a subclass (attributes are invariant), so the optionality could not simply be confined to the QEMU subclass — the inheritance itself had to go.

### The fix
A new hypervisor-agnostic base `VmResources` (`controllers/resources.py`) holds the shared, message-free surface — `namespace`, `volumes`, `rootfs_path`, `kernel_image_path`, `to_dict`, `download_kernel` — plus the message-coupled mechanics as free functions (`host_volumes_from_message`, `disk_usage_delta`). The two lineages now specialise it independently:

- **`AlephFirecrackerResources(VmResources)`** — `message_content` is **required** again, so the program/instance dereferences are clean (the 9 errors are gone).
- **`AlephQemuResources(VmResources)`** — owns its **optional** `message_content` honestly. The single rootfs dereference in `download_runtime` is narrowed with `assert isinstance(..., InstanceContent)` (QEMU only ever runs instances).

`HostVolume` moves to the agnostic module alongside `VmResources`; its one in-repo importer (a #953 test) is repointed.

### Result
- `mypy` across `controllers/`: **21 → 12 errors**, which is **exactly the `origin/dev` baseline** — the only remaining error just relocated from `executable.py` to `resources.py` (a pre-existing `HostVolume` arg-type). The 9 `Item "None"` errors #954 introduced are eliminated.
- **No behaviour change.** Existing message paths are untouched; the spec path keeps working (`from_spec` unchanged in behaviour).
- `ruff format` + `isort` clean; targeted supervisor tests green (14 passed).

### DIFF
--- a/src/aleph/vm/controllers/firecracker/executable.py
+++ b/src/aleph/vm/controllers/firecracker/executable.py
@@ -6,14 +6,13 @@ import asyncio
 import logging
 from dataclasses import dataclass, field
 from multiprocessing import Process, set_start_method
-from os.path import exists, isfile
+from os.path import exists
 from pathlib import Path
 from typing import Generic, TypeVar
 
 from aiohttp import ClientResponseError
 from aleph_message.models import ExecutableContent, ItemHash
 from aleph_message.models.execution.environment import MachineResources
-from aleph_message.models.execution.volume import MachineVolume, PersistentVolume
 
 from aleph.vm.conf import settings
 from aleph.vm.controllers.configuration import (
@@ -23,11 +22,16 @@ from aleph.vm.controllers.configuration import (
 )
 from aleph.vm.controllers.firecracker.snapshots import CompressedDiskVolumeSnapshot
 from aleph.vm.controllers.interface import AlephVmControllerInterface
+from aleph.vm.controllers.resources import (
+    VmResources,
+    disk_usage_delta,
+    host_volumes_from_message,
+)
 from aleph.vm.guest_api.__main__ import run_guest_api
 from aleph.vm.hypervisors.firecracker.microvm import FirecrackerConfig, MicroVM
 from aleph.vm.network.firewall import teardown_nftables_for_vm
 from aleph.vm.network.interfaces import TapInterface
-from aleph.vm.storage import chown_to_jailman, get_volume_path
+from aleph.vm.storage import chown_to_jailman
 
 try:
     import psutil  # type: ignore [no-redef]
@@ -67,14 +71,6 @@ class Volume:
 
 
 @dataclass
-class HostVolume:
-    mount: str
-    path_on_host: Path
-    read_only: bool
-    size_mib: int | None
-
-
-@dataclass
 class BaseConfiguration:
     vm_hash: ItemHash
     ip: str | None = None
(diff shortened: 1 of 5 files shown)


foxpatch-aleph (approved):
Clean refactor that breaks the problematic AlephQemuResources → AlephFirecrackerResources inheritance chain by introducing a hypervisor-agnostic VmResources base class. The new code eliminates 9 mypy errors introduced in #954, makes the message contract explicit per hypervisor (required for Firecracker, optional for QEMU), and moves shared host-resource mechanics (HostVolume, disk_usage_delta, host_volumes_from_message) into a dedicated module. No behaviour change. Well-documented, well-structured.

**`src/aleph/vm/models.py`** (line 115): Pre-existing: the type union here lists `AlephQemuConfidentialInstance` (an instance class) where it should list `AlephQemuConfidentialResources` (a resources class). Not introduced by this PR, but worth fixing.

foxpatch-aleph (approved):
Clean refactor that extracts the shared `HostVolume` dataclass from the Firecracker resource class into a new `controllers/resources.py` module. The QEMU resource pipeline already imported `HostVolume` from `controllers/resources.py` and its use in `controllers/qemu/instance.py` is unchanged. The diff is primarily deletions and import shifts — no behavior changes. Test imports are correctly updated.

**`src/aleph/vm/controllers/firecracker/executable.py`** (line 4): Great to see the import list shrink. The stale import of `AlephFirecrackerResources` from `.configuration` was correctly dead code — the class lives in this same file, not in configuration.

foxpatch-aleph (approved):
Clean refactor that extracts hypervisor-agnostic VM resource logic into a shared base class VmResources, breaking the incorrect inheritance of AlephQemuResources from AlephFirecrackerResources. The new module in controllers/resources.py correctly partitions shared mechanics (HostVolume, disk_usage_delta) from hypervisor-specific contracts. The approach fixes existing mypy errors (21→12) with zero behavior change. Code is well-structured and type-safe.

**`src/aleph/vm/controllers/resources.py`** (line 1): Consider adding `from __future__ import annotations` for consistency with newer modules in the codebase, though this is non-blocking.

**`src/aleph/vm/controllers/resources.py`** (line 90): The VmResources class attributes (kernel_image_path, rootfs_path, volumes) are declared as annotations but not initialized in __init__. This relies on subclasses setting them — consider using `None` defaults or dataclass-style field declarations for clarity. Non-blocking since existing subclasses already satisfy the contract.

**`src/aleph/vm/controllers/resources.py`** (line 61): disk_usage_delta accesses `volume.size_mib` (singular 'mib') via getattr in host_volumes_from_message but the attribute is actually `size_mib`. Let me double-check — this is indeed consistent with the PersistentVolume model field. Correct as written.

odesenfans:
Superseded by #959. This PR auto-merged (into its old base `od/msgfree-1-spec-execution`, **not** `dev`) when the untangle branch was reordered to the bottom of the stack — its head became an ancestor of its base. The untangle now targets `dev` directly in #959. No changes reached `dev` from this PR.

PR #959, opened 2026-05-31
refactor(controllers): untangle QEMU resources from Firecracker

odesenfans:
**Bottom of the message-free-supervisor stack.** Base: `dev`.

> Reordered to land *before* the spec-execution work (#954) so that no transient `message_content`-Optional pollution is ever introduced. (Supersedes the earlier #958, which auto-merged when this branch was moved below #954 in the stack.)

`AlephQemuResources` inherited from `AlephFirecrackerResources` purely to reuse host-resource mechanics (volumes, disk accounting, kernel path), even though a QEMU VM is not a Firecracker VM. That coupling forced a single shared `message_content` declaration — so any message-free path (QEMU built from a spec) would have to widen `message_content` to Optional on the shared base, polluting the Firecracker subclasses that always have a message.

### Changes
- New hypervisor-agnostic `VmResources` base (`controllers/resources.py`) holding the shared, message-free surface: `namespace`, `volumes`, `rootfs`/`kernel` paths, `to_dict`, `download_kernel`. The message-coupled mechanics become free functions (`host_volumes_from_message`, `disk_usage_delta`).
- `AlephFirecrackerResources(VmResources)` keeps `message_content` **required** — program/instance dereferences stay clean.
- `AlephQemuResources(VmResources)` owns its **optional** `message_content` (it may be built message-free); the lone rootfs dereference is narrowed with `isinstance(..., InstanceContent)` (QEMU only runs instances).
- `HostVolume` moves to the agnostic module.

### Result
No behaviour change; existing message paths untouched. `mypy src/aleph/vm/` adds zero errors over the `dev` baseline.

### Stack
1. **(this PR)** untangle QEMU resources -> `dev`
2. spec-constructible `VmExecution` -> 1 (#954)
3. pool create path -> #954 (#955)
4. route production creation through the spec -> #955 (#956)
5. reboot-recovery from on-disk configs -> #956 (#957)

### DIFF
--- a/src/aleph/vm/controllers/firecracker/executable.py
+++ b/src/aleph/vm/controllers/firecracker/executable.py
@@ -6,14 +6,13 @@ import asyncio
 import logging
 from dataclasses import dataclass, field
 from multiprocessing import Process, set_start_method
-from os.path import exists, isfile
+from os.path import exists
 from pathlib import Path
 from typing import Generic, TypeVar
 
 from aiohttp import ClientResponseError
 from aleph_message.models import ExecutableContent, ItemHash
 from aleph_message.models.execution.environment import MachineResources
-from aleph_message.models.execution.volume import MachineVolume, PersistentVolume
 
 from aleph.vm.conf import settings
 from aleph.vm.controllers.configuration import (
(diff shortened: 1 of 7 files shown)


foxpatch-aleph (approved):
Well-structured refactoring that cleanly untangles QEMU resources from Firecracker resources by extracting a hypervisor-agnostic VmResources base class. The extracted free functions (host_volumes_from_message, disk_usage_delta) preserve the original logic exactly. The type hierarchy now correctly reflects the semantic difference — Firecracker always requires a message, QEMU may be built without one. No correctness or security issues found. Tests were updated for the import path change.

**`src/aleph/vm/controllers/resources.py`** (line 71): The `hasattr(message_content, "rootfs")` check works because only InstanceContent has a rootfs attribute, but it would be slightly cleaner to use an explicit type check (`isinstance(message_content, InstanceContent)`) since the codebase already imports and uses InstanceContent from aleph_message.models. The hasattr approach is fragile — a renamed attribute or a new message type with a rootfs field would silently change behavior. This is a minor style nit inherited from the original code.

foxpatch-aleph (commented):

foxpatch-aleph (approved):
Clean extraction of a hypervisor-agnostic VmResources base class from the Firecracker-specific resource hierarchy. AlephQemuResources no longer inherits from AlephFirecrackerResources, removing the forced shared message_content coupling. All lifted functions preserve existing logic exactly. Type annotations are narrowed correctly. No behavioral change introduced.

PR #960, opened 2026-06-01
chore: add Justfile for running the test suite locally

odesenfans:
## What

Adds a `Justfile` with recipes to provision a local test environment and run the suite without root.

```
just system-deps   # apt packages + build headers (Debian/Ubuntu, sudo)
just venv          # create .testvenv and install the project + test/lint deps
just test          # run the suite (or `just test tests/supervisor`)
just mypy          # whole-package type check (the CI gate)
just lint          # ruff format + isort checks
just clean         # remove .testvenv and .test-roots
```

## Why

Running the suite locally currently has two undocumented gotchas that cost real back-and-forth with CI:

1. **Missing deps.** The test venv must include `solathon`, `aleph-superfluid` and the pinned pytest plugins (notably `pytest-mock`); without them the `mocker`-fixture tests error at *collection* rather than failing visibly, so a green local run can hide real breakage.
2. **Hardcoded roots.** The suite writes to `/var/cache/aleph` and `/var/lib/aleph`, which need root. `just test` redirects `ALEPH_VM_CACHE_ROOT` / `ALEPH_VM_EXECUTION_ROOT` to a gitignored temp dir so the bulk of the suite runs as a normal user.

## Notes

- The `venv` recipe uses `--system-site-packages` so the apt `python3-*` bindings (nftables, dbus, systemd) resolve, mirroring CI's hatch `testing` env (`system-packages = true`). Run `just system-deps` first so the `dbus-python` / `systemd-python` builds find their headers.
- A handful of tests still require root + network + qemu/firecracker (network-interface creation, VM creation, runtime downloads); those remain CI-only, as they already run under `sudo` there.
- `.testvenv/` and `.test-roots/` are gitignored.
- Validated with `just 1.36.0`: `--list` and `--fmt --check` are clean, and `just test tests/supervisor/test_resources.py` passes end-to-end (10 passed) with the roots redirected.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,10 @@
 *.pyc
 __pycache__
 
+# Local test virtualenv and writable roots created by the Justfile
+.testvenv/
+.test-roots/
+
 *.sqlite3
 *.bin
 *.ext4
--- a/Justfile
+++ b/Justfile
@@ -1,0 +1,70 @@
+# Developer tasks for running the aleph-vm test suite locally.
+#
+# Requires `just` (https://github.com/casey/just). Run `just` to list recipes.
+#
+# Quick start (Debian/Ubuntu):
+#   just system-deps   # once: apt packages + build headers (sudo)
+#   just venv          # create .testvenv, install the project + test deps
+#   just test          # run the suite without root
+#
+# `system-deps` installs the apt python3-* bindings (nftables, dbus, systemd,
+# ...) and the dev headers pip compiles dbus-python / systemd-python against,
+# mirroring .github/workflows/test-using-pytest.yml. `venv` then creates a
+# --system-site-packages virtualenv (so those bindings are visible, matching
+# CI's hatch `testing` env) and installs the rest with pip.
+#
+# A handful of tests still require root + network + qemu/firecracker (network
+# interfaces, VM creation, runtime downloads); those only pass in CI under sudo.
+# Path to the local test virtualenv.
+
+venv := ".testvenv"
+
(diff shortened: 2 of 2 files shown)


foxpatch-aleph (commented):
Well-structured Justfile that accurately mirrors CI dependencies and solves the two key pain points of local test execution (missing deps and hardcoded root dirs). The only gap is the lint recipe skipping yamlfix and pyproject-fmt checks that CI runs.

**`Justfile`** (line 65): The CI `linting:style` recipe in pyproject.toml also runs `yamlfix --check .` and `pyproject-fmt --check pyproject.toml`. The Justfile lint recipe skips these, so local lint may pass while CI fails on YAML or pyproject formatting. Consider adding them.

foxpatch-aleph (approved):
This PR adds a Justfile providing recipes for local test suite execution, addressing two key pain points: (1) missing system/test dependencies that cause silent failures at test collection, and (2) hardcoded /var/cache/aleph and /var/lib/aleph paths requiring root. The Justfile faithfully mirrors the existing CI workflow's apt packages and pip dependencies (from hatch's testing env), uses --system-site-packages to match CI's hatch config, and redirects cache/execution roots to a gitignored temp directory. The .gitignore entries for the generated directories are appropriate. The code is correct, well-commented, and the PR description accurately documents the known limitations (root-requiring tests remain CI-only).

**`Justfile`** (line 46): Consider adding a brief comment explaining why eth_typing==4.3.1 is pinned (matching the pyproject.toml comment: "Temp fix for bug in CI with 5.0.0"). This helps future maintainers understand it's not an arbitrary pin.

PR #961, opened 2026-06-02
fix: periodically re-sync haproxy domain mappings

aliel:
The domain-mapping refresh path is WS-event-driven: an aggregate update
  fires _handle_domains_aggregate, which fetches from DOMAIN_SERVICE_URL
  and rewrites the haproxy maps. If the downstream DNS indexer hasn't
  caught up yet when the CRN fetches, the fetch returns a stale list, the
  maps converge to that stale state, and no further refresh ever runs
  until the next aggregate update.

  Add a periodic background task that calls update_domain_mapping every
  DOMAIN_RESYNC_INTERVAL seconds (default 600). update_domain_mapping is
  idempotent — no haproxy writes when content matches — so the steady-
  state cost is one HTTP GET plus three map reads.

  Jittered: first sleep picks a random phase across the interval;
  subsequent sleeps use ±15% jitter. Prevents thundering-herd against
  DOMAIN_SERVICE_URL when CRNs restart together (rolling deploys).

### DIFF
--- a/src/aleph/vm/conf.py
+++ b/src/aleph/vm/conf.py
@@ -297,6 +297,14 @@ class Settings(BaseSettings):
         default=60.0,
         description="Interval in seconds between payment checks",
     )
+    DOMAIN_RESYNC_INTERVAL: float = Field(
+        default=600.0,
+        description=(
+            "Interval in seconds between periodic HAProxy domain-mapping "
+            "re-syncs against DOMAIN_SERVICE_URL. Acts as a safety net for "
+            "missed WS events or upstream indexing lag."
+        ),
+    )
     PAYMENT_RECEIVER_ADDRESS: str | None = Field(
         default=None,
         description="Address of the account receiving payments. " "Defaults to OWNER_ADDRESS if not set.",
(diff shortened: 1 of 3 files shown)


foxpatch-aleph (approved):
This PR adds periodic resync logic for HAProxy domain mappings to address a gap where downstream DNS indexing delays could cause stale domain maps to persist indefinitely until the next aggregate update. The implementation is well-structured: it introduces a new domain resync task with jittered intervals, shards DNS checks evenly across the resync window to distribute load, and ensures thread-safe idempotent merge operations. The code follows existing patterns in the codebase, avoids over-engineering, and properly integrates with the existing startup/cleanup lifecycle. All changes are functionally focused and self-contained.

**`src/aleph/vm/orchestrator/supervisor.py`** (line 323): Good call adding all three lifecycle hooks (startup + two cleanup hooks side-by-side) in the same block where WATCH_FOR_MESSAGES is gated. Keeps the conditional lifecycle registration self-contained and easy to reason about.

**`src/aleph/vm/conf.py`** (line 300): The default of 600s and the docstring clearly explain the belt-and-braces purpose. Good choice setting the default to 10 minutes — frequent enough to catch index lag without creating load.

**`src/aleph/vm/orchestrator/tasks.py`** (line 460): Nice use of `random.uniform(0, interval)` for initial phase randomization and ±15% jitter on subsequent sleeps. This prevents thundering-herd issues when multiple CRNs restart in a rolling deploy scenario.

PR #962, opened 2026-06-03
Message-free supervisor (5/N): wire the agent's create path onto the Supervisor abstraction

odesenfans:
## Summary

Phase 0 of the aleph-cvm backport: the agent's **create path** now goes through the message-agnostic `Supervisor` abstraction instead of reaching into `VmPool` directly. The supervisor creates VMs from a `CreateVmSpec` and never sees an Aleph message; the agent owns messages via a new agent-side `AgentVmRegistry` (which replaces the vestigial `pool.message_cache`).

Stacked on #957.

### What changes
- **`AgentVmRegistry`** (`orchestrator/vm_registry.py`): in-memory, agent-side cache of the messages the agent knows about (kept as a DB-backed cache conceptually; later it will be rebuilt from the scheduler plan + Aleph messages). Replaces `pool.message_cache`.
- **`create_vm_execution`** routes spec-eligible instances through `Supervisor.create_vm(spec)`, records the message in the registry, polls `get_vm` until `RUNNING`, and applies port forwards through `Supervisor.add_port_forward` (agent resolves the policy via `resolve_port_forwards`, the supervisor owns the mechanism). Tears the VM down on readiness/port-forward failure.
- **App wiring**: `setup_webapp` constructs the app-wide `InProcessSupervisor` + `AgentVmRegistry`; every create/start call site (operator reboot/reinstall/restore, allocation views, CLI) threads them through. Program paths build them locally (they always take the legacy create path, which never touches the supervisor).
- **PR-1 boundary** (documented): the operator endpoints, billing and update-watching still read owner identity off `execution.message`, so the spec-created execution is re-sourced as message-driven until those consumers move to the registry in a later PR. `start_persistent_vm` still drives the `VmExecution` for the pre-existing check / expiry-cancel.

Design doc and implementation plan are included under `docs/plans/`.

### Notable review fix
The initial create-path rewrite dropped the line that keeps the execution message-driven, which would have 500'd owner-auth on every spec-eligible instance (operator tests mock `execution.message` and never drive the create path, so it slipped through). Restored with a regression test (`31772d2a`).

## Test Plan
- [x] Full suite: only the 8 known environmental failures (root/network/binary), 552 passed
- [x] mypy union-attr gates: `src/aleph/vm/` = 2, `controllers/` = 0
- [x] mypy on touched modules: no new errors (net -4 vs base)
- [x] ruff + isort: clean (no new findings)
- [x] New unit tests: `AgentVmRegistry`, `resolve_port_forwards`, `_wait_until_running`, create-path routing (incl. timeout + port-forward-failure teardown, message-driven re-sourcing)
- [ ] Manual: create an instance and hit an owner-authed operator endpoint (reboot) to confirm auth still resolves

### DIFF
--- a/docs/plans/2026-06-01-wire-agent-onto-supervisor-design.md
+++ b/docs/plans/2026-06-01-wire-agent-onto-supervisor-design.md
@@ -1,0 +1,264 @@
+# Wiring the agent onto the Supervisor abstraction (Phase 0 completion)
+
+**Date:** 2026-06-01
+**Status:** Design, approved for planning
+**Design doc lineage:** follows `2026-05-28-aleph-vm-architecture-backport-design.md` (§6 Phase 0), `2026-05-29-phase-0b-supervisor-abc-design.md`, `2026-05-29-phase-0c-create-vm-decouple-design.md`, and the `2026-05-29-message-free-supervisor-create-reboot-design.md` plan (PRs #954–#957, merged).
+
+## 1. Goal
+
+Make the agent reach hypervisor functionality **only** through the `Supervisor`
+abstraction (`src/aleph/vm/supervisor/abc.py`, in-process impl
+`InProcessSupervisor`), starting with the **persistent-instance lifecycle**.
+Today the abstraction is fully built and conformance-tested but nothing outside
+tests instantiates it: the agent (`orchestrator/run.py`, `orchestrator/views`,
+`orchestrator/tasks.py`) still reaches straight into `VmPool` and `VmExecution`
+internals (~25 distinct touchpoints; `pool.executions` alone is read in 24
+places).
+
+This is the explicit Phase 0 exit criterion from the backport design:
+
+> *"`orchestrator/views` and `orchestrator/run.py` are migrated to call the
+> abstraction … every call into hypervisor functionality from agent code goes
+> through the abstraction."*
+
+It is also the prerequisite for Phase 1: a call's transport cannot be swapped to
+gRPC until the call goes through the boundary.
+
+## 2. Scope
+
+**In scope: persistent-instance lifecycle through the abstraction**
+
+- Create path: `create_vm_execution` (eligible-instance branch) and
+  `start_persistent_vm` in `orchestrator/run.py`.
+- Operator lifecycle endpoints: stop/delete, reboot, reinstall, logs, and
+  port-forward listing/removal in `orchestrator/views` and the deallocation
+  path in `orchestrator/tasks.py`.
+- A new **agent-side registry** of created VMs (message + original), replacing
+  the vestigial `pool.message_cache`.
+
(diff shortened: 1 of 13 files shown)


foxpatch-aleph (approved):
Cleanly wires the agent's create path onto the Supervisor abstraction with a well-defined responsibility split. AgentVmRegistry correctly replaces the vestigial message_cache, the poll loop replaces becomes_ready, and port-forward resolution is correctly agent policy vs hypervisor mechanism. Error handling is thorough (teardown on timeout/port-forward failure with secondary-error logging). Test coverage is excellent across the new helpers, routing, and failure modes. The pool.executions readback is correctly flagged as temporary.

**`src/aleph/vm/orchestrator/run.py`** (line 217): The legacy path (pool.create_a_vm) already sets execution.spec internally, but the new code doesn't need to re-source it since the message is in the registry. This is correct behavior.

foxpatch-aleph (changes requested):
The PR is architecturally sound and well-tested, but has one correctness issue: the legacy create_a_vm path records into the registry before the VM is successfully created, leaving a stale registry entry if create_a_vm raises. Also, the _wait_until_running poll should check for VmStatus.STOPPING as a terminal state to avoid hanging until deadline timeout.

**`src/aleph/vm/orchestrator/run.py`** (line 210): Legacy path asymmetry: `pool.create_a_vm()` can raise, but `registry.record(...)` is called unconditionally before. If `create_a_vm` fails (e.g. resource exhaustion, download error), the registry holds a record for a VM that was never created. The spec path (line 186) correctly wraps the record in a try/except that calls `registry.forget(vm_hash)` on failure. Fix: move `registry.record(...)` to after the successful return of `create_a_vm`, or wrap in a try/finally.

**`src/aleph/vm/orchestrator/run.py`** (line 157): Consider adding `VmStatus.STOPPING` and `VmStatus.DEFINED` to the terminal-status check in `_wait_until_running`. Currently only `STOPPED` and `FAILED` short-circuit the poll. If the VM transitions to `STOPPING` or stays `DEFINED`, the loop hangs until the deadline timeout instead of failing fast.

foxpatch-aleph (approved):
Well-structured PR that cleanly wires the agent's create path through the Supervisor abstraction. The AgentVmRegistry replaces the dead pool.message_cache, the create path polling/teardown logic is correct, and all call sites are properly threaded with supervisor/registry. The PR-1 boundary comments clearly mark the temporary pool.executions read-back. Minor nit: redundant int(vm_port) cast in resolve_port_forwards, but not a blocker. Tests cover the routing, timeouts, port-forward failures, and registry semantics thoroughly.

**`src/aleph/vm/orchestrator/run.py`** (line 125): Minor: `vm_port` is already `int` at this point (cast on line 110), so `int(vm_port)` is redundant.

PR #963, opened 2026-06-03
fix: stop event loop stalls in monitor_payments D-Bus sweep

aliel:
monitor_payments => check_payment iterates executions and reads
  execution.is_running per VM, which goes through is_service_active.
  Two layered problems made this serialize the asyncio event loop
  for tens of seconds at a time on CRNs with many VMs:

  - _ensure_connection ran ListUnits() before every operation as a "connection test", enumerating every loaded systemd unit on the host over D-Bus. Replaced with the cheap get_is_connected() bool check; real call failures still fall through to the per-method DBusException handlers.
  - Each is_running access issued its own D-Bus round-trips, synchronously, on the event loop thread. Batch the lookup once via SystemDManager.get_services_active_states (already present, one ListUnits per batch) wrapped in asyncio.to_thread, and pass the result down through get_executions_by_address.

### DIFF
--- a/src/aleph/vm/orchestrator/run.py
+++ b/src/aleph/vm/orchestrator/run.py
@@ -271,9 +271,9 @@ async def start_persistent_vm(vm_hash: ItemHash, pubsub: PubSub | None, pool: Vm
     execution: VmExecution | None = pool.executions.get(vm_hash)
     if execution:
         if execution.is_running:
-            logger.info(f"{vm_hash} is already running")
+            logger.debug(f"{vm_hash} is already running")
         elif execution.is_starting:
-            logger.info(f"{vm_hash} is already starting")
+            logger.debug(f"{vm_hash} is already starting")
         elif execution.is_stopping:
             logger.info(f"{vm_hash} is stopping, waiting for complete stop before restarting")
             await execution.stop_event.wait()
--- a/src/aleph/vm/orchestrator/tasks.py
+++ b/src/aleph/vm/orchestrator/tasks.py
@@ -334,8 +334,27 @@ async def check_payment(pool: VmPool):
             # Status is healthy — reset any previous strikes
             _terminal_strike_count.pop(str(vm_hash), None)
 
+    # Snapshot the active state of every persistent VM's controller
+    # service via a single batched D-Bus ListUnits() call, run off the
+    # event loop. Without this, get_executions_by_address would hit
+    # is_service_active per VM (each round-trip blocks the loop), and
+    # with dozens of VMs the loop stalls for tens of seconds, visible
+    # to clients as multi-second TTFB on unrelated HTTP requests.
+    persistent_services = [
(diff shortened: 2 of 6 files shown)


foxpatch-aleph (approved):
This PR fixes a real performance bug where D-Bus ListUnits() was called as a connection health-check before every operation, and per-VM is_service_active calls each issued multiple D-Bus round-trips synchronously on the event loop. The fix replaces the expensive ListUnits() health-check with get_is_connected(), batches active state lookups via get_services_active_states() in asyncio.to_thread, and passes the pre-computed dict through get_executions_by_address. The code is clean, well-commented, and the logic correctly separates persistent VMs (systemd-backed, use batch) from non-persistent VMs (use existing is_running path). Existing tests use persistent=False so they are unaffected, though new tests covering the batch path would be a welcome addition.

**`src/aleph/vm/pool.py`** (line 842): The batch path (get_services_active_states) only checks ActiveState from ListUnits(), while the old per-VM is_service_active path also checked is_service_enabled() first. A service that is loaded and active but not technically "enabled" (e.g. manually started via systemctl start) would now be included in payment checks where it was previously excluded. This is arguably more correct behavior — if the VM is running, its payments should be checked — but it is a semantic change worth noting.

**`src/aleph/vm/systemd.py`** (line 46): get_is_connected() is a local library-level flag and does not actually verify the D-Bus daemon is responsive. However, the real call failures still trigger reconnect via the per-method DBusException handlers, so this is fine in practice. Consider noting this in the docstring for clarity.

**`src/aleph/vm/orchestrator/tasks.py`** (line 343): No test coverage exists for the new batch path (persistent VMs with running_states). The existing tests all use persistent=False. Adding a test that creates a persistent VmExecution with a mocked systemd_manager.get_services_active_states would help prevent regressions.

foxpatch-aleph (approved):
This PR fixes a significant event loop stall in monitor_payments/check_payment caused by two layered issues: _ensure_connection calling ListUnits() (enumerating all systemd units) before every D-Bus operation, and each VM's is_running check issuing its own synchronous D-Bus round-trip. The fix replaces the ListUnits() health check with a cheap local get_is_connected() flag, and batches per-VM active-state lookups via get_services_active_states (one ListUnits() call) run off the event loop through asyncio.to_thread. The approach is consistent with the existing load_persistent_executions pattern. Code is well-documented, tests cover the new batching path, and logging noise is appropriately reduced.

**`src/aleph/vm/systemd.py`** (line 224): Running get_services_active_states via asyncio.to_thread exposes the shared SystemDManager instance to concurrent access from a thread pool worker while the event loop thread may also call _ensure_connection (e.g. from an HTTP handler checking is_running). The dbus-python library's SystemBus is not thread-safe. In practice this is low-risk (short-lived call, low concurrent D-Bus access likelihood), but consider adding a threading lock around the manager's D-Bus operations or at minimum documenting this assumption.

PR #964, opened 2026-06-04
fix(auth): cross-endpoint iat collisions in Aleph-EIP191-V1 replay protection

odesenfans:
## Problem

The replay check introduced in #945 keeps a **per-signer** floor of the last accepted `iat` and rejects anything not strictly greater. `iat` has integer-second resolution and the floor is keyed by signer only, so two *distinct* requests signed within the same wall-clock second collide regardless of endpoint:

1. `POST …/migration/export` signed with `iat = T` → accepted, floor becomes `T`
2. `GET …/migration/export/status` 55 ms later — different path, different signature, but `iat = T` again → `T <= T` → **401**

The endpoint binding never gets a say: the GET passes the method/path/body checks and dies on the floor. In effect the floor is a global rate limit of one request per signer per second across all endpoints — not replay protection.

## Fix

The signed payload already binds `method`, `path`, `body_sha256` and `iat`, so the only thing left to legitimately reject as a "replay" is a **byte-identical resubmission**. The per-signer monotonic floor is replaced with a per-signer map of accepted payload hashes:

- **Dedup key is the payload hash, not the signature** — ECDSA signatures are malleable (`(r, s, v) → (r, −s mod n, v′)` recovers the same signer over the same message), so a signature-keyed seen-set could be bypassed by re-encoding a captured signature.
- **Still keyed per signer** — two authorized signers can legitimately produce byte-identical payloads; a global set would reject the second as a replay of the first.
- **Memory bounded** — entries are pruned once their `iat` falls behind the absolute acceptance window (`ALLOCATION_SIGNATURE_MAX_AGE_SECONDS`), at which point a resubmission would be rejected as stale anyway.
- Same lock-serialized check-and-record, same "update last" ordering as before.

## Guarantees preserved (pinned by tests)

- Exact replay of a captured request rejected (`test_verify_rejects_replay`)
- Two concurrent copies of the same request: exactly one accepted (`test_verify_concurrent_same_iat_rejects_one`)
- No cross-signer interference (`test_verify_per_signer_dedup_is_independent`, rewritten for the new semantics)
- Absolute time-window behavior untouched (M3 boundary tests unchanged)

## New regression tests

- `test_verify_accepts_same_iat_across_endpoints` — the exact failing sequence above (verified red on `main`)
- `test_verify_accepts_same_iat_distinct_bodies_same_endpoint` — two allocations POSTed within the same second (verified red on `main`)
- `test_accepted_payloads_pruned_after_window` — seen-set memory bound (mutation-checked: fails if pruning is disabled)

`tests/supervisor/views/test_allocation_auth.py`: 59 passed. The 7 pre-existing failures in `tests/supervisor/test_views.py` (host-dependent system-usage/cert tests) are identical on unmodified `origin/main`.

### DIFF
--- a/src/aleph/vm/orchestrator/views/allocation_auth.py
+++ b/src/aleph/vm/orchestrator/views/allocation_auth.py
@@ -75,36 +75,54 @@ def _parse_auth_params(auth_header: str) -> dict[str, str]:
     return params
 
 
-_last_accepted_iat: dict[str, int] = {}
-"""Per-signer floor on accepted `iat` values. Module-level state, in-memory
-only; doesn't survive supervisor restarts (the absolute time window covers
-the post-restart gap). Single-process only: a multi-worker supervisor would
-let a captured request be replayed against a sibling worker."""
+_accepted_payloads: dict[str, dict[str, int]] = {}
+"""Per-signer map of accepted payload hashes to their `iat`, used to reject
+byte-identical resubmissions. The signed payload already binds method, path,
+body and iat, so the only thing left to call a "replay" is an exact
+duplicate — the previous design (a per-signer monotonic-iat floor) also
+rejected *distinct* same-second requests, throttling each signer to one
+request per second across all endpoints. Entries are pruned once their iat
+falls behind the absolute acceptance window (a resubmission would then be
+rejected as stale anyway), bounding memory to the signer's request rate over
+~2x the window. Module-level state, in-memory only; doesn't survive
+supervisor restarts (the absolute time window covers the post-restart gap).
+Single-process only: a multi-worker supervisor would let a captured request
+be replayed against a sibling worker."""
 
-_iat_lock: asyncio.Lock | None = None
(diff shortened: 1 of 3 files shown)


foxpatch-aleph (approved):
This PR cleanly fixes the cross-endpoint replay collision introduced by the previous per-signer monotonic-iat floor design. The replacement is a per-signer payload-hash seen-set with time-based pruning, which correctly allows distinct same-second requests while still rejecting byte-identical replays. The code is well-structured, the security rationale (signature malleability) is properly addressed, and the test coverage — including targeted regression tests that verify the fix — is thorough. No correctness or security issues found.

**`src/aleph/vm/orchestrator/views/allocation_auth.py`** (line 87): Minor: the "~2x the window" memory-bound estimate is slightly imprecise. Pruning only runs when a new request arrives for that signer, so a silent signer's entries persist indefinitely. In practice this is fine (the dict size is bounded by the number of unique payloads the signer ever submitted), but consider clarifying that it's "bounded by the signer's request rate over the window" rather than "~2x."

PR #965, opened 2026-06-05
Message-free supervisor (6/N): operator lifecycle through the Supervisor abstraction

odesenfans:
PR 2 of the agent-onto-Supervisor wiring (design: `docs/plans/2026-06-01-wire-agent-onto-supervisor-design.md` §5/§8; implementation plan: `docs/plans/2026-06-04-wire-agent-supervisor-pr2-plan.md`, first commit). Stacked on #962 — review that first; this retargets to `dev` once #962 merges.

## What moves behind the boundary

- **Operator lifecycle endpoints** — `operate_stop`, `operate_reboot`, `operate_erase`, `operate_reinstall`, the logs endpoints, and `operate_update` now reach the hypervisor only through `Supervisor` methods keyed by `VmId`. Owner-auth reads the `AgentVmRegistry` (DB fallback kept in the logs endpoints for past executions, as before).
- **Deallocation paths** — `check_payment` (terminal-status dealloc + the three insufficient-funds stops) and the `update_allocations` stop loop go through `supervisor.delete_vm`. `registry.forget` only on permanent dealloc; insufficient-funds stops keep the record (the VM may be re-paid).
- **Port forwarding** — new agent-side `reconcile_port_forwards` (aggregate → diff → `add/remove_port_forward`) drives `operate_update` and the port-forwarding aggregate handler. The DB-load half of `fetch_port_redirect_config_and_setup` is gone; persisted-mapping reload is now explicitly the creator's job (`create_a_vm`, `create_vm_from_spec`, `restart_persistent_vm`, `recreate_network`).
- **Registry persistence** — `AgentVmRecord` gains `persistent`; the registry rehydrates from `ExecutionRecord`s on startup, so owner-auth and the persistent flag survive agent restarts.

## Contract enrichment (proto + ABC + in-process)

- `DeleteVmRequest.wipe` / `delete_vm(vm_id, wipe=False)` — wipe deletes persisted port mappings (persistent VMs) + writable data volumes; rootfs stays (old `operate_erase` semantics exactly).
- `ReinstallVmRequest.wipe_volumes` / `reinstall_vm(vm_id, wipe_volumes=True)` — the in-process impl now does the real erase/re-prepare/systemd-restart mechanics (`_restart_persistent_vm` moved to `VmPool.restart_persistent_vm`). Note in the proto: the ABC default (`True`) can't be expressed in proto3; the Phase-1 gRPC server must treat unset as `True`.
- `LOG_SOURCE_STDERR` — stderr is no longer folded into stdout. `get_logs` serves journald history in-process (works for stopped VMs); `stream_logs(include_history=)` yields history then the live queue.

## Bugs fixed in passing (found while planning, see plan "Design deltas #4")

- Spec-created VMs never wrote an `ExecutionRecord` (`save()` no-op'd before the agent re-sourced the spec) — past-logs auth and rehydration would have silently missed them.
- `create_vm_from_spec` didn't preload persisted port mappings — persistent QEMU instances restarted via allocation got new host ports.
- `recreate_network` crashed (caught per-VM) on reboot-reattached executions; it now reapplies their DB mappings and skips the message-only aggregate fetch.
- `remove_port_forward` left ghost `{tcp: false, udp: false}` entries in `mapped_ports` when clearing a port's last protocol.

## Behavior notes for reviewers

- `operate_stop`/`operate_reboot` now act on `RUNNING` **or** `BOOTING` (old `is_running` missed VMs still booting).
- `operate_logs_json` timestamps are now UTC-aware (`...+00:00` suffix), derived from `timestamp_ns`.
- `operate_update` returns "VM not starting yet" for any non-RUNNING status (incl. BOOTING); config is applied on the create path anyway.

## Residuals (explicitly kept, commented in code)

`delete_port_mappings` direct DB calls on dealloc; `operate_expire`; `recreate_network`/`regenerate_proxy`; confidential/backup/restore/migration endpoints; billing decision reads (`get_executions_by_address`, `get_persistent_executions`); the program path (`run_code_on_*`). See plan "Design deltas #3".

## Tests

594 passed, 1 skipped, 3 xfailed; the 8 documented env-only failures (root/network/qemu) unchanged. Proto bindings regenerate clean; isort/ruff/format gates green; no new mypy errors.


foxpatch-aleph (approved):
Well-executed refactoring that cleanly migrates operator lifecycle endpoints behind the Supervisor abstraction. Strong test coverage, clear design docs, and good handling of edge cases. Minor issues: a latent timestamp bug in _history_chunks (double-counts microseconds), and a potential VmId string representation mismatch in journald identifier construction. Neither blocks approval.

**`src/aleph/vm/supervisor/inprocess.py`** (line 130): timestamp_ns double-counts microseconds: datetime.timestamp() already includes fractional seconds (e.g., 123.456789). The expression `int(ts.timestamp()) * 1_000_000_000 + ts.microsecond * 1_000` drops the sub-second fraction from timestamp() then adds it back from .microsecond — but only for whole microseconds, losing nanosecond precision. Should be `int(ts.timestamp() * 1_000_000_000)`.

**`tests/supervisor/test_supervisor_inprocess_logs.py`** (line 31): Potential VmId string issue: VM_ID = VmId('vm1') and _entries expects a string. If VmId is a NewType over str, str(VmId('vm1')) should be 'vm1'. But if the test ever runs against the real code path (rather than monkeypatched), journal identifiers would be wrong. Worth double-checking VmId.__str__ behavior.

foxpatch-aleph (approved):
Cleanly migrates operator lifecycle endpoints and deallocation paths onto the Supervisor abstraction. Contract enrichments (wipe flag, stderr log source, reinstall semantics) are well-motivated. Test coverage is thorough, including new edge cases for BOOTING/STOPPED states, DB fallback auth, and port-forward reconciliation. The few minor issues (timestamp precision edge case in history chunks) do not warrant blocking.

**`src/aleph/vm/supervisor/inprocess.py`** (line 138): `int(ts.timestamp()) * 1_000_000_000 + ts.microsecond * 1_000`: `ts.timestamp()` already includes the microseconds, so the `microsecond * 1_000` addition is effectively adding the sub-second part twice. Should be `int(ts.timestamp() * 1_000_000_000)` or just `ts.microsecond * 1_000` if `ts` is a naive `datetime` (but it should be timezone-aware from journald). In practice this will only cause off-by-a-few-milliseconds issues.

foxpatch-aleph (approved):
Well-structured PR that correctly migrates operator lifecycle endpoints onto the Supervisor abstraction. The contract enrichment (wipe flags, stderr log source), registry persistence with rehydration, port-forward reconciliation helper, and deallocation path updates are all correctly implemented. Tests are comprehensive and cover edge cases. Minor non-blocking observations: timestamp precision in _history_chunks uses truncation instead of full float-to-ns conversion, one test over-constrains module namespace contents, and another uses a fragile unbound-method invocation pattern. None of these warrant changes.

**`src/aleph/vm/supervisor/inprocess.py`** (line 139): Timestamp conversion: `int(ts.timestamp()) * 1_000_000_000` truncates the fractional-second part, then `+ ts.microsecond * 1_000` adds it back — but the two operations don't compose correctly for non-whole-second timestamps. Use `int(ts.timestamp() * 1_000_000_000)` instead, or keep as-is if sub-second journald precision is not expected in practice.

**`tests/supervisor/test_port_mappings.py`** (line 199): Namespace check `assert not hasattr(models_mod, "get_port_mappings")` tests a module implementation detail rather than behavioral correctness. If someone re-imports `get_port_mappings` into `models.py` for a legitimate different use, this breaks. The behavioral assertion (`update_mock.assert_awaited_once()`) below is sufficient.

**`tests/supervisor/test_supervisor_inprocess_lifecycle.py`** (line 180): Calling `VmExecution.erase_volumes(execution, ...)` as an unbound method with a `SimpleNamespace` self is fragile. If `erase_volumes` later accesses any `self.*` attribute beyond `self.resources`, the test silently passes without exercising real logic. Consider using a factory or a proper mock with the necessary attributes.

foxpatch-aleph (approved):
Correctly wires operator lifecycle (stop, reboot, erase, reinstall, logs, port-forwards) through the Supervisor abstraction and AgentVmRegistry. No regressions introduced; 594 tests pass. The implementation matches the design doc and covers auth, error handling, and edge cases thoroughly. The ghost-entry fix in remove_port_forward is a real bug fix. Timestamp computation avoids float64 precision loss.

**`src/aleph/vm/supervisor/inprocess.py`** (line 131): _history_chunks calls get_past_vm_logs which does blocking sd-journal reads on the event loop. Same behavior as today (comment says so), but worth tracking for a future async journald reader.

PR #966, opened 2026-06-05
fix(supervisor): stop destroying confidential VMs awaiting owner initialization

odesenfans:
## Problem

Reported by a user: confidential VM `c3ab0d76…` is allocated by the scheduler to a CRN (coco0.leviathan.so, 1.12.0) but never comes up. The scheduler POSTs `/control/allocations` every minute as expected, yet the VM stays "missing" forever. A second confidential VM on the same node (`e95a0b05…`) is stuck in the identical state.

## Root cause

A confidential VM execution is created with its controller service deliberately left inactive: it can only start once its owner uploads the session certificates via `/control/machine/{ref}/confidential/initialize` (the CRN cannot have the disk secret).

In that waiting state all three state properties are False:

| Property | Value | Why |
|---|---|---|
| `is_starting` | False | `times.started_at` is set by `start()` |
| `is_running` | False | persistent ⇒ systemd check; controller never started |
| `is_stopping` | False | no `stopping_at` |

Two failure loops feed each other:

1. **Node side:** every allocation POST falls into the `"unknown execution state, stopping the vm"` branch of `start_persistent_vm()` and **stops and recreates the execution that was legitimately waiting for its owner**, racing the owner's initialization handshake:

```
11:19:36 Starting instance 'c3ab0d76…'
11:19:36 c3ab0d76… unknown execution state, stopping the vm
11:19:37 <VMExecution AlephQemuConfidentialInstance c3ab0d76…> stopped
11:19:37 Starting persistent virtual machine with id: c3ab0d76…
```

2. **Scheduler side:** the scheduler reads `/about/executions/list` and only uses the keys (`executions.into_keys()` → `running_vms`; re-POSTs whenever `planned_vms != running_vms`). That list filters to *running* VMs, so the waiting VM never appears → the scheduler re-allocates every cycle and reports the VM as `missing`, forever.

## Fix

- Add `VmExecution.is_awaiting_confidential_init`: confidential + persistent + started + not stopping + controller not active.
- `start_persistent_vm()` leaves such executions untouched instead of stopping and recreating them.
- `/about/executions/list` (read by the scheduler) now includes awaiting-init confidential VMs, with an `awaiting_confidential_init` flag on all entries. This makes the scheduler observe the VM, ending the re-allocation storm and the permanent `missing` status. Entries are only added once the tap interface exists: the scheduler deserializes values into a struct with **required** `networking.ipv4`/`ipv6` fields, and one entry without them would fail the whole node poll.
- `/v2/about/executions/list` exposes `awaiting_confidential_init` as well so consumers can distinguish "waiting for its owner" from a dead VM.

Note: with this change the scheduler will count a waiting confidential VM as observed even though the guest is not booted until its owner initializes it. That is the intent — re-allocating cannot start it, only the owner can — but consoles reading scheduler state will show it as placed rather than missing until they consume the new flag.

## Tests

- `tests/supervisor/test_run.py` (new): property semantics (awaiting / running / stopping / non-confidential) and `start_persistent_vm` keeping the awaiting execution (no `stop()`, no recreate, no `forget_vm`).
- `tests/supervisor/test_views.py`: v1 list includes an awaiting-init VM (full payload incl. networking) and excludes it while the tap interface is missing; v2 list reports `awaiting_confidential_init: true`; existing list tests updated for the new field.

All tests written first and verified to fail on the previous behavior.

### DIFF
--- a/src/aleph/vm/models.py
+++ b/src/aleph/vm/models.py
@@ -339,6 +339,21 @@ class VmExecution:
         return True if getattr(self.message.environment, "trusted_execution", None) else False
 
     @property
+    def is_awaiting_confidential_init(self) -> bool:
+        """The confidential VM is created but waiting for its owner to upload the
+        session certificates and start it via /control/machine/{ref}/confidential/initialize.
+
+        The controller service is only started at that point, so the execution is
+        neither starting nor running. It must not be treated as a dead execution:
+        only the owner can start it, the orchestrator cannot."""
+        return (
+            self.is_confidential
+            and self.persistent
+            and bool(self.times.started_at and not self.times.stopping_at)
+            and not self.is_running
+        )
+
+    @property
     def hypervisor(self) -> HypervisorType:
         if self.is_program:
             return HypervisorType.firecracker
--- a/src/aleph/vm/orchestrator/run.py
+++ b/src/aleph/vm/orchestrator/run.py
@@ -278,6 +278,10 @@ async def start_persistent_vm(vm_hash: ItemHash, pubsub: PubSub | None, pool: Vm
             logger.info(f"{vm_hash} is stopping, waiting for complete stop before restarting")
             await execution.stop_event.wait()
             execution = None
+        elif execution.is_awaiting_confidential_init:
+            # Stopping and recreating the execution here would loop forever:
+            # only the owner can start it by uploading the session certificates.
+            logger.info(f"{vm_hash} is waiting for its owner to initialize the confidential session")
         else:
             logger.info(f"{vm_hash} unknown execution state, stopping the vm")
             if execution.vm:
--- a/src/aleph/vm/orchestrator/views/__init__.py
+++ b/src/aleph/vm/orchestrator/views/__init__.py
@@ -269,6 +269,9 @@ async def list_executions_v2(request: web.Request) -> web.Response:
                 ),
                 "status": execution.times,
                 "running": running_states.get(item_hash, False),
+                # Confidential VMs are only started once their owner uploads the
+                # session certificates: not running, but not dead either.
+                "awaiting_confidential_init": execution.is_awaiting_confidential_init,
                 "vm_type": VmType.from_message_content(execution.message).name,
             }
             for item_hash, execution in pool.executions.items()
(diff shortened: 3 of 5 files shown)


foxpatch-aleph (approved):
Correctly fixes an infinite loop where the scheduler repeatedly destroyed and recreated confidential VMs awaiting owner initialization. The new is_awaiting_confidential_init property properly distinguishes this state from a dead VM, and start_persistent_vm now leaves such executions untouched. Tests thoroughly cover property semantics, the allocation path, and the API exposure. Logic is sound, no security concerns, and the diff is minimal and well-structured.

**`src/aleph/vm/models.py`** (line 354): Consider the edge case where `systemd_manager` is None for a persistent VM: `is_running` falls back to checking `starting_at` (not `started_at`), which would be truthy after `start()`, making `is_awaiting_confidential_init` return False despite waiting for init. In practice persistent VMs always have `systemd_manager`, so this is defensive only — not blocking.

foxpatch-aleph (approved):
Cleanly fixes a production bug where confidential VMs awaiting owner initialization via session certificates were repeatedly stopped and recreated by the allocation loop. The fix introduces a new `is_awaiting_confidential_init` property on `VmExecution` that correctly identifies the waiting state, treats it as a known valid state in `start_persistent_vm` (instead of falling through to the 'unknown state, stop the VM' branch), and exposes the flag in both v1 and v2 execution list endpoints so the scheduler sees the VM as placed. The property logic, control flow, and API changes are all correct. Test coverage is thorough across property semantics, behavioral integration, and both API versions.

foxpatch-aleph (approved):
Correctly fixes a production bug where confidential VMs waiting for owner initialization were caught in a re-allocation loop. The fix addresses both sides: (1) prevents stop-and-recreate in start_persistent_vm for awaiting-init VMs, and (2) includes them in /about/executions/list so the scheduler stops re-allocating. The tap_interface guard prevents breaking strict v1 schema consumers. Tests cover property semantics and integration behavior thoroughly.

**`src/aleph/vm/models.py`** (line 342): `is_awaiting_confidential_init` calls `self.is_running` which makes a separate D-Bus call per execution. In the list endpoints, this adds N individual calls on top of the already-batched `_get_executions_running_states`. Consider making this property accept an optional pre-computed running state, or compute it from the batch results in the views.

**`tests/supervisor/test_run.py`** (line 93): Consider adding a test for `is_confidential=True` + `persistent=False` to explicitly verify that non-persistent confidential VMs never report as awaiting initialization, since that edge case is not currently covered.

**`tests/supervisor/test_run.py`** (line 66): Consider adding a test for when `started_at is None` (VM was never started) to ensure `is_awaiting_confidential_init` returns `False` in that case.

PR #968, opened 2026-06-08
feat(auth): source authorized allocation signers from settings aggregate

odesenfans:
## Summary

The Aleph-EIP191-V1 scheduler auth path previously read its authorized-signer allow-list only from `settings.AUTHORIZED_ALLOCATION_SIGNERS`, which defaulted to empty — so the signature path was effectively off until each operator configured it by hand.

This introduces a new resolver, `get_authorized_allocation_signers()`, that resolves signers through **three layers** (precedence: override > aggregate > default):

| Layer | Source | Wins when |
|-------|--------|-----------|
| 1. Override | `AUTHORIZED_ALLOCATION_SIGNERS` (env/config) | non-empty → used alone, nothing else consulted |
| 2. Network | settings aggregate `authorized_allocation_signers` | override empty **and** aggregate lists ≥1 signer |
| 3. Default | `DEFAULT_ALLOCATION_SIGNERS` (built-in) | override empty **and** aggregate yields nothing |

### Why
- **Zero-config day one / outage resilience** — `DEFAULT_ALLOCATION_SIGNERS` ships with the official scheduler address, so a fresh CRN trusts it without configuration and stays reachable if the aggregate is unfetchable.
- **Network rotation** — the foundation publishing a non-empty list to the aggregate replaces the default network-wide, no CRN redeploy.
- **Operator fast rotation** — the env override shadows everything instantly, independent of the aggregate's owner key being available.

### Notes
- An **empty aggregate list** is treated as "no opinion" and falls back to the default rather than authorizing nobody, so a stray empty publication can't brick scheduling globally.
- The boot-time auth summary log now reflects aggregate-sourcing (with a built-in default) when no local override is set.
- The private key matching the shipped default address must live on the scheduler.

## Test plan
- 70/70 tests pass in `tests/supervisor/views/test_allocation_auth.py`, including new coverage for all three layers, override shadowing, default fallback, empty-aggregate fallback, and end-to-end accept/reject.
- `ruff format --check`, `isort --check-only`, and `mypy` clean on the changed files.

### DIFF
--- a/src/aleph/vm/conf.py
+++ b/src/aleph/vm/conf.py
@@ -15,7 +15,7 @@ from aleph_message.models import Chain
 from aleph_message.models.execution.environment import HypervisorType
 from dotenv import load_dotenv
 from eth_utils import is_checksum_address
-from pydantic import Field, HttpUrl, field_validator
+from pydantic import Field, HttpUrl, ValidationInfo, field_validator
 from pydantic_settings import BaseSettings, SettingsConfigDict
 
 from aleph.vm.orchestrator.chain import STREAM_CHAINS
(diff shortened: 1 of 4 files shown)


foxpatch-aleph (approved):
Well-structured PR that adds a three-layer precedence (local override > settings aggregate > built-in default) for authorized allocation signers under the Aleph-EIP191-V1 auth scheme. The design handles failure modes gracefully (empty aggregate list falls back to default, aggregate unavailability falls back to default) and provides operator escape hatches (env override shadows everything immediately). Tests cover all layers comprehensively with clean fixture design.

PR #967, opened 2026-06-05
Message-free supervisor (7/N): execution list endpoints through the Supervisor

odesenfans:
PR 3 of the agent-onto-Supervisor wiring (design: `docs/plans/2026-06-01-wire-agent-onto-supervisor-design.md` §8 item 3; implementation plan: `docs/plans/2026-06-06-wire-agent-supervisor-pr3-plan.md`, first commit). Stacked on #965 — review that first; this retargets to `dev` once #965 merges.

## What moves behind the boundary

- **`/about/executions/list`** and **`/v2/about/executions/list`** now read `supervisor.list_vms()` / `get_host_info()` / `list_port_forwards()` + the `AgentVmRegistry` (for `vm_type`) — zero `pool.executions` access. Their JSON output is byte-compatible; the three pre-existing v2 byte-parity tests pass **unchanged** as the proof.
- The views' `_get_executions_running_states` batch-systemd helper is deleted; the same batching now lives behind the boundary (`InProcessSupervisor.list_vms` issues one `get_services_active_states` D-Bus call regardless of VM count).

## Contract enrichment (proto + dataclasses + in-process)

- `VmInfo` gains `ipv4_network`/`ipv6_network` (tap CIDR strings, empty until the tap exists), the 7 lifecycle timestamps (`defined/preparing/prepared/starting/started/stopping/stopped`, `uint64` unix-ns, 0 = stage not reached), and `is_instance`.
- `HostInfo` gains `host_ipv4` (hypervisor-owned `pool.network`; empty when host networking is disabled).
- Timestamps use the same lossless composition as the log chunks (`int(ts.timestamp()) * 1e9 + µs * 1000`); the agent-side reconstruction is exact at µs precision, so the v2 `"status"` datetimes serialize byte-identically.
- `is_instance` was added mid-execution: the planned registry-miss `vm_type` fallback via `backend` is lossy (an instance on the default Firecracker hypervisor reports `FIRECRACKER` and would be mislabeled `microvm` — caught by the parity tests). The hypervisor carries the signal explicitly.

## Behavior notes for reviewers

- Ghost `mapped_ports` entries (`{"tcp": false, "udp": false}`, creatable only via an all-false entry in the user's port-forwarding aggregate on the legacy create path) are intentionally no longer listed by v2: `list_port_forwards` only reports enabled protocols. Pinned by `test_v2_executions_list_omits_ghost_mapped_ports`.
[...]

PR #969, opened 2026-06-08
Message-free supervisor: agent-side expiry facility

odesenfans:
## What

Lifts idle-teardown ("expiry") of on-demand VMs off the `VmExecution` god-object into a new agent-owned `ExpiryManager`, and removes every expiry member from `VmExecution` and `VmPool`. First of the Phase-0 residual cleanups in the message-free supervisor series.

Design: `docs/plans/2026-06-08-supervisor-expiry-design.md`
Plan: `docs/plans/2026-06-08-supervisor-expiry-plan.md`

## Why

Idle teardown is agent policy, but it lived as methods on `VmExecution` (the object being split into an agent `Execution` and a hypervisor `Vm`). The pool also reached into it (`cancel_expiration` in four places). This decouples it so that when microVM creation moves behind the Supervisor in Phase 1, nothing in the idle path needs a `VmExecution`.

## How

- New `ExpiryManager` (`orchestrator/expiry.py`): timers keyed by `vm_id`, reaping via `Supervisor.delete_vm(vm_id)` (`wipe=False`, so data volumes are preserved). Built as `app["expiry"]`, threaded into the request path (`run_code_on_request`) and the reactor path (`run_code_on_event` via `Reactor`); cancelled on shutdown via `on_cleanup`.
- `run_code_on_*` cancel any pending timer when reusing a running VM and re-arm in `finally` (preserves "no expiry while serving a request").
- `start_persistent_vm` and `operate_expire` drive the manager; teardown endpoints (`operate_stop`/reboot/erase) cancel pending timers.
- `VmExecution` loses `stop_after_timeout`/`expire`/`cancel_expiration`/`expire_task`; `VmPool` loses all four `cancel_expiration` calls.

## Behavior parity

`delete_vm(wipe=False)` does not erase volumes, and `create_a_vm` already auto-forgets on stop, so the reap matches today's behavior (volumes survive, VM forgotten after idle).

## Notes / follow-ups

- `operate_expire` has a pre-existing dead route (path lacks `{timeout}` but the handler reads `match_info["timeout"]`, so it always 400s). This PR only swaps the timer mechanism and intentionally leaves the route bug and the skipped `test_operator_expire` as-is. Fixing the route is a separate follow-up.
- Update-watching (`watch_for_updates`) and the `pool.executions[vm_hash]` readback at `run.py` are the other two Phase-0 residuals, shipped separately.

## Stacking

Stacked on #967 (`od/wire-supervisor-read-views`). Base this PR on that branch; rebase onto `dev` once #967 merges.

### DIFF
--- a/docs/plans/2026-06-08-supervisor-expiry-design.md
+++ b/docs/plans/2026-06-08-supervisor-expiry-design.md
@@ -1,0 +1,195 @@
+# Wire agent onto Supervisor: agent-side expiry facility
+
+**Status:** Design / spec (approved 2026-06-08)
+**Owner:** Olivier Desenfans
+**Series:** Message-free supervisor, Phase 0 residual cleanup. First of the
+residual PRs that finish detaching the agent from `VmExecution` before the
+Phase 1 process split.
+**Stacked on:** PR 3 read views (`od/wire-supervisor-read-views`, #967), itself
+on `dev`.
+
+## 1. Context
+
+Idle teardown for on-demand programs ("expiry") is agent policy: keep a microVM
+warm for `REUSE_TIMEOUT` seconds after a request, then reap it. Today the timer
+lives on `VmExecution` (`stop_after_timeout` / `expire` / `cancel_expiration` /
+`expire_task`), the god-object Phase 0 is splitting into an agent-side
+`Execution` and a hypervisor-side `Vm`. The timer is therefore on the wrong side
+of the future boundary, and the agent reaches it by holding the `VmExecution`
+instance.
+
+The same coupling shows up on the hypervisor side: `VmPool` itself calls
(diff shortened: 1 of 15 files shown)


foxpatch-aleph (approved):
Clean extraction of idle-teardown expiry from VmExecution/VmPool into an agent-owned ExpiryManager. The design is well-researched, the implementation faithfully follows the plan, and the tests cover the new unit and the critical persistent-VM guard. One minor unused gather call in the tests, but nothing blocking.

**`tests/supervisor/test_expiry.py`** (line 55): The `asyncio.gather(first, return_exceptions=True)` here is dead code — `first.cancelled()` has already returned `True` by this point, so the gather result is never used and not assigned. Consider removing it.

PR #970, opened 2026-06-09
Message-free supervisor: agent-side update-watching + execution-free start_persistent_vm

odesenfans:
Second of the Phase-0 residual cleanups in the message-free supervisor series (after #969 expiry). Lifts **update-watching** ("redeploy a VM when its Aleph message updates") off the `VmExecution` god-object into a new agent-owned `UpdateWatcher`, and makes `start_persistent_vm` **execution-free**.

Design: `docs/plans/2026-06-09-supervisor-update-watching-design.md`
Plan: `docs/plans/2026-06-09-supervisor-update-watching-plan.md`

## Why

Update-watching is agent policy (it subscribes to the deployment's code/runtime/data/volume refs and stops the VM on update), but it lived as methods on `VmExecution` — the object being split into an agent `Execution` and a hypervisor `Vm`. This is the exact mirror of what #969 did for idle-expiry. With it gone, the idle/update paths no longer need a `VmExecution`, and `start_persistent_vm` stops reaching into `pool.executions`.

## What

- **New `UpdateWatcher`** (`orchestrator/update_watcher.py`): subscription tasks keyed by `vm_id`, reading each VM's message from the `AgentVmRegistry` and reaping via `Supervisor.delete_vm(wipe=False)` on update. Subscription-driven counterpart to `ExpiryManager`, with the same concurrency discipline (current-task identity check, `VmNotFoundError` swallow, `CancelledError` re-raise). `watch()` is **idempotent** (a live subscription is not restarted — preserves the old `if not self.update_task` guard).
- **`start_persistent_vm` is now execution-free**: the pre-existing-VM check runs through `supervisor.get_vm` (a `VmStatus` state machine: `RUNNING`→reuse, `DEFINED`/`BOOTING`→`_wait_until_running`, `STOPPING`→`_wait_until_gone`→recreate, `STOPPED`/`FAILED`→`delete_vm`→recreate, absent→create), readiness via `_wait_until_running`, update via the watcher. Returns `None` (callers already ignore the return). This kills the residual `pool.executions.get(vm_hash)` read the design doc named.
- **`VmExecution` loses** `watch_for_updates` / `start_watching_for_updates` / `update_task` / `cancel_update` (and the `cancel_update()` call in `stop()`); the ref-extraction logic relocates to `UpdateWatcher`.
- **Wiring symmetric with expiry**: `app["update_watcher"]` built in `setup_webapp`; threaded into the request path, the reactor (which now also adopts the shared `app["vm_registry"]` instead of a throwaway local one), and `start_persistent_vm`; `update_watcher.cancel` added next to every `expiry.cancel` in stop/reboot/erase; `cancel_all()` on shutdown; `cli.py` benchmark/`start_instance` updated.

## Behavior parity

- Reap-on-update is unchanged: subscribe to the same refs, then stop the VM (volumes preserved, VM forgotten — matches the old `stop()` and #969's expiry reap).
- The no-op for message-less (spec-built/reattached) VMs is preserved (the watcher no-ops when the registry has no record). Reattached persistent VMs are now *also* watched after a restart (rehydration repopulates the registry), a small improvement over the old execution-based no-op.

## Two issues caught in review (worth calling out)

1. **On-demand programs kept their watch.** `run_code_on_request` was building a throwaway local `AgentVmRegistry` for program creates, so the app-level watcher would never see the record. Fixed to use `request.app["vm_registry"]` (every caller — HTTP views + benchmark — has the app singletons), so on-demand programs remain update-watched.
2. **Sibling-timer leak on reap.** Removing `cancel_update()` from `VmExecution.stop()` broke the linkage where an expiry-reap used to also tear down the update-watch. The agent now has two idle-teardown facilities reaping the same VM, so each is given an `on_reaped` hook cross-wired to the other's `cancel` — an expired VM no longer leaks its pending `msubscribe` task, and vice-versa. Regression-tested (`test_expiry_reap_cancels_update_watch`).

## Residuals (explicitly out of scope, documented in the design doc §3)

- The `create_vm_execution` `save()` readback (`pool.executions[vm_hash]` → `execution.save()`) — the agent's `ExecutionRecord` persistence; deferred to when the rest of the agent's record-persistence leaves the execution.
- Operator owner-auth's `execution.message` reads (~10 sites) — these stay; owner-auth is agent policy and simply reads the agent's own message store. (Migrating them to read the registry is separate plumbing.)

## Tests

`UpdateWatcher` unit tests mirror `test_expiry.py` (reap-on-update, cancel, idempotent-no-restart, re-watch-after-completion, no-op-when-unrecorded, `VmNotFoundError` swallow, `cancel_all`, `on_reaped`); `start_persistent_vm` precheck-state tests (running/booting/stopping/absent/terminal); a `VmExecution`-API-gone test; the sibling-cancel regression test. Full supervisor suite shows no new failures vs the #969 base (the pre-existing `/var/cache/aleph` environment failures are identical to the dev baseline). ruff-format / isort gates green.

## Stacking

Stacked on #969 (`od/wire-supervisor-expiry`). Base this PR on that branch; retarget to `dev` once #969 merges.

### DIFF
--- a/docs/plans/2026-06-09-supervisor-update-watching-design.md
+++ b/docs/plans/2026-06-09-supervisor-update-watching-design.md
@@ -226,7 +226,7 @@ one the create path records into.
   (alongside the `supervisor`/`expiry` it already took), passes them into
   `run_code_on_event`.
 - `tasks.py` (`start_watch_for_messages_task`): construct
-  `Reactor(pubsub, pool, supervisor, app["expiry"], app["vm_registry"], app["update_watcher"])`.
+  `Reactor(pubsub, pool, supervisor, app["expiry"], app["update_watcher"], app["vm_registry"])`.
 - Lifecycle endpoints (`views/operator.py`): add `update_watcher.cancel(vm_id)`
   next to every existing `expiry.cancel(vm_id)` (stop / reboot / erase).
 - `update_allocations` / `notify_allocation` (`views/__init__.py`): pass
--- a/docs/plans/2026-06-09-supervisor-update-watching-design.md
+++ b/docs/plans/2026-06-09-supervisor-update-watching-design.md
@@ -1,0 +1,287 @@
+# Agent-side update-watching + `start_persistent_vm` execution decoupling
+
+**Status:** design
+**Date:** 2026-06-09
+**Series:** message-free supervisor (Phase-0 residual cleanup, after #969 expiry)
+**Stacks on:** `od/wire-supervisor-expiry` (#969)
+
+## 1. Goal
+
+Lift update-watching ("re-deploy on message update") off the `VmExecution`
+god-object into an agent-owned `UpdateWatcher`, and make `start_persistent_vm`
+execution-free by moving its pre-existing-VM check onto `supervisor.get_vm`.
+
+This is the second half of the design doc's future-work item *"Migrate
+expiry-cancel and update-watching off `VmExecution`"* — expiry shipped in #969;
+this PR does update-watching and the residual `start_persistent_vm`
+`pool.executions` read it named.
+
+## 2. Background
+
+The message-free-supervisor effort splits the current `VmExecution` god-object
+into an agent-side concept (knows the message, owner, billing, expiry,
+update-watching — backed by the `AgentVmRegistry` + agent DB) and a
+hypervisor-side `Vm` (process, networking, status — behind the `Supervisor`
+boundary). The agent *owns* messages; the goal is to get message-derived policy
+off the **hypervisor object**, not off the agent.
+
+Update-watching is agent policy: it subscribes to the Aleph messages referenced
+by a VM's deployment (code / runtime / data / volumes) and, when any is updated,
+stops the VM so the next request redeploys it. Today it lives as
+`VmExecution.watch_for_updates` / `start_watching_for_updates` / `update_task` /
+`cancel_update`, exactly parallel to where expiry lived before #969.
+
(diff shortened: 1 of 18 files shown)


foxpatch-aleph (approved):
Well-structured PR that cleanly extracts update-watching from VmExecution into an agent-owned UpdateWatcher, mirroring the pattern established in #969. The sibling-cancel regression fix (on_reaped hook wireup) is correct and well-tested. start_persistent_vm's new execution-free state machine is sound. No blocking issues found.

**`docs/plans/2026-06-09-supervisor-update-watching-design.md`** (line 287): Excellent design doc — clear scope boundaries and well-reasoned architectural decisions.

**`src/aleph/vm/orchestrator/update_watcher.py`** (line 50): The `not existing.done()` guard accurately preserves the old `if not self.update_task` idempotency semantics. The identity check in `_watch`'s `finally` correctly handles the race where a new task replaces the old one before its `finally` runs.

**`tests/supervisor/test_expiry.py`** (line 111): Nit: `await asyncio.sleep(0.05)` to wait for a 0.01s timer is less robust than `await asyncio.wait_for(expiry._tasks[vm_id], timeout=...)` used in the earlier expiry tests. Consider aligning patterns for CI reliability.

**`tests/supervisor/test_update_watcher.py`** (line 213): Same pattern as above — `await asyncio.sleep(0.05)` for a 0.01s expiry timer. Consider using `await asyncio.wait_for(...)` instead to avoid flakiness on loaded CI runners.

PR #971, opened 2026-06-09
Message-free supervisor: operator owner-auth reads the agent registry

odesenfans:
Third of the Phase-0 residual cleanups in the message-free supervisor series (after #969 expiry, #970 update-watching). Moves **operator owner-authorization** off the hypervisor pool object and onto the agent registry, so `execution.message` is read **nowhere** in `orchestrator/views/operator.py`.

Design: `docs/plans/2026-06-09-supervisor-owner-auth-registry-design.md`
Plan: `docs/plans/2026-06-09-supervisor-owner-auth-registry-plan.md`

## Why

Owner-auth is agent policy: a request is authorized iff the authenticated sender is the message owner or a delegate. The message it checks is the agent's own — so it should come from the agent registry (`get_agent_record_or_404(request, vm_hash).message`), not from a reach through `pool.executions` into the hypervisor object's `spec.message`. `operate_stop`/`operate_reboot`/`operate_erase` already did this; this PR finishes the migration for the remaining endpoints.

## What

Nine `execution.message` sites in `operator.py`, migrated in three shapes:

- **Delete (dead code):** `authenticate_websocket_for_vm_or_403` — zero callers repo-wide (YAGNI). Plus a stale `# TODO ... execution.message.address` comment.
- **Become execution-free** (used the execution *only* for auth): `operate_expire`, `operate_backup_status`, `operate_backup_delete` — authorize from the registry, drop the pool lookup entirely.
- **Auth-only swap** (genuinely need `execution.vm` for the hypervisor op): `operate_confidential_initialize`, `operate_confidential_measurement`, `operate_confidential_inject_secret`, `operate_backup`, `_do_restore` — keep the `get_execution_or_404` lookup, but read the auth message from the registry *before* it (incl. `_do_restore`'s two `execution.message.rootfs.size_mib` content reads → `record.message.rootfs.size_mib`).

A source-scan guard test (`test_operator_module_does_not_read_execution_message`) locks the invariant.

## Behavior parity

Authorized callers: unchanged (auth passes, the retained execution lookup still 404s a non-running VM exactly as before).

One deliberate change for **unauthorized** callers: auth now runs before the running-VM lookup, so an unauthorized request against a *known-but-stopped* VM returns **403** instead of **404**. This matches the already-migrated `operate_stop`/reboot/erase and is arguably more correct (authorization precedes VM-state disclosure). The unknown-VM path still 404s (registry miss).

## Regression caught in review

`tests/supervisor/test_views.py::test_restore_rejects_invalid_image_format` calls `_do_restore` **directly** with a hand-built `request.app` that had no `vm_registry`, so the migration broke it (`KeyError: 'vm_registry'`). Fixed by wiring a `vm_registry` mock into that test (separate commit). A tree-wide sweep confirmed it was the only direct-caller test affected; every other test routes through `setup_webapp`, which provides the registry.

## Residuals (explicitly out of scope, design §7)

- `execution.vm` / `execution.is_running` direct hypervisor reads in operator.py — Phase-1 Supervisor-boundary work, not message-coupling.
- The `create_vm_execution` `save()` readback — still deferred.
- `operate_expire`'s broken route (no `{timeout}` segment → always 400) — pre-existing dead route; its auth read was migrated but the routing bug is left as-is.

## Tests

Per migrated endpoint: unauthorized → 403, authorized → past-auth, registry-miss → 404, all sourced from a seeded `app["vm_registry"]` with an empty pool (so pre-migration code 404s — the red→green signal). Plus the four existing `confidential_initialize` tests reseeded onto the registry, the source-scan guard, and the dead-function guard. `tests/supervisor/views/test_operator.py` + `tests/supervisor/test_views.py` are green (79 passed). Remaining `tests/supervisor` failures are the pre-existing environmental set (pyroute2/netlink, Firecracker subprocess, `/var` paths) — identical to the dev baseline.

## Stacking

Stacked on #970 (`od/wire-supervisor-update-watch`). Base this PR on that branch; retarget to `dev` once #970 merges.

### DIFF
--- a/docs/plans/2026-06-09-supervisor-owner-auth-registry-design.md
+++ b/docs/plans/2026-06-09-supervisor-owner-auth-registry-design.md
@@ -1,0 +1,192 @@
+# Operator owner-auth reads the agent registry
+
+**Status:** design
+**Date:** 2026-06-09
+**Series:** message-free supervisor (Phase-0 residual cleanup, after #969 expiry, #970 update-watching)
+**Stacks on:** `od/wire-supervisor-update-watch` (#970)
+
+## 1. Goal
+
+Stop operator endpoints from reading owner identity off the hypervisor pool
+(`pool.executions[vm_hash].message`). After this PR, `execution.message` is read
+**nowhere** in `orchestrator/views/operator.py`; owner-auth consults the agent
+registry — the agent's own message store — exactly like the already-migrated
+`operate_stop` / `operate_reboot` / `operate_erase`.
+
+This is the last of the design docs' named owner-auth residual: *"Operator
+owner-auth's `execution.message` reads (~10 sites in `operator.py`) → registry."*
+
+## 2. Background
+
+The message-free-supervisor effort splits the `VmExecution` god-object into an
+agent-side concept (message, owner, billing, expiry, update-watching — backed by
+the `AgentVmRegistry` + agent DB) and a hypervisor-side `Vm` (process,
+networking, status — behind the `Supervisor` boundary). The agent *owns*
+messages; the goal is to get message-derived policy off the **hypervisor
+object**, not off the agent.
+
+Owner-authorization is agent policy: an operator request is authorized iff the
+authenticated sender is the message owner or a delegate. The message it reads is
+the agent's own — so it should come from the agent registry, not from a reach
+through `pool.executions` into the hypervisor object's `spec.message`.
+
+### What is already migrated (the template)
+
+`operate_stop`, `operate_reboot`, and `operate_erase` already read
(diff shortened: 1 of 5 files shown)


foxpatch-aleph (approved):
Clean migration of operator owner-authorization from hypervisor pool to agent registry. All 9 execution.message reads in operator.py are either deleted, replaced by registry reads, or reordered so auth happens before the pool lookup. The only behavioral change (403 vs 404 for unauthorized requests against known-but-stopped VMs) is intentional and matches already-migrated endpoints. Tests cover authorized paths, unauthorized 403 paths, registry-miss 404 paths, and include two source-scan guard tests. The test_views.py fix for the direct _do_restore caller is minimal and correct.

PR #972, opened 2026-06-10
Message-free supervisor: agent-owned record persistence; kill the create_vm_execution readback

odesenfans:
Fourth of the Phase-0 residual cleanups in the message-free supervisor series (after #969 expiry, #970 update-watching, #971 owner-auth). Makes the agent persist its **own** DB record and migrates the last agent-side readers of pool-execution messages to the `AgentVmRegistry`, then **kills the `create_vm_execution` readback**.

Design: `docs/plans/2026-06-10-agent-record-persistence-design.md`
Plan: `docs/plans/2026-06-10-agent-record-persistence-plan.md`

## Why

The supervisor's spec create path is message-free — but `create_vm_execution` immediately undid that: it reached back into `pool.executions[vm_hash]`, glued a `MessageSpec` onto the hypervisor object, and called `execution.save()` to write the agent's DB record (a block commented `TEMPORARY`). That readback is the last agent reach-in on the create path and can't exist once the supervisor is a separate process. Removing it requires (a) the agent persisting its own record, and (b) the few agent-side readers that still learned a VM's message off the pool object reading the agent registry instead.

## What

- **`persist_record`** (`vm_registry.py`) — write-side sibling of `rehydrate_registry`: the agent writes its own `ExecutionRecord` (message/original/persistent/resources; `vm_id`/`mapped_ports` left `None` as hypervisor-owned debug-only fields). Round-trip tested against `rehydrate_registry`.
- **Readback killed** (`run.py`): the spec path now `persist_record(...)` and returns `None` (signature `-> VmExecution | None`); the two program callers guard the `None` case with an explicit 400. The routing test runs against an **empty pool**, so a re-introduced readback would `KeyError`.
- **Payment grouping → registry**: `pool.get_executions_by_address` deleted; replaced by agent-side `_group_executions_by_payment(pool, registry, …)` in `tasks.py` (message from the registry, liveness from the execution).
- **`update_allocations`** payment-tier guards read `record.uses_payment_*`; the dead `VmExecution.uses_payment_stream/credit` properties are deleted.
- **Domains aggregate** reads the owner address from the registry.

## Behavior

Authorized/legacy paths unchanged. Two intended changes, both from sourcing the message via the registry (rehydrated from the agent DB) instead of the hypervisor object:

- **Spec-built and restart-restored VMs are now covered** by payment grouping, the `update_allocations` guards, and the domains-aggregate refresh — previously they carried no message on the pool object and were silently skipped.
- Spec-created executions now **never** carry a `MessageSpec`: `execution.message` is `None` for them in all code paths (a `test_pool_has_no_message_reads` source guard, plus the operator.py guard from #971, lock this in), so any new pool-message dependency fails visibly in dev rather than working-until-reboot.

## Scope (per design §3 / scheduler decision)

The new scheduler allocates PAYG and the legacy `/control/allocations/notify` flow is deprecating, so payment enforcement is a scheduler prerogative. This PR moves **where payment data comes from** (registry, not pool-message); it does **not** redesign or delete `check_payment`'s balance/credit/stream logic or the terminal-status stop loop — those deprecate later with the legacy flow. Structural `pool.executions` iteration, `execution.vm` operator reads, and the `about_executions` debug endpoint are deferred Phase-1 residuals (§7).

## Tests

`persist_record`→`rehydrate` round-trip; registry-sourced grouping (incl. the restored-VM coverage case, both diagnostic-hash skips, hold default); `update_allocations` spares-PAYG-via-registry and stops-unrecorded; domains skip-matrix; empty-pool routing regression guard; `uses_payment_*`-gone API guard; `pool.py`-has-no-message-reads source guard. Pre-existing `test_checkpayment.py` (×6) and `test_supervisor_spec_pool_guards.py` migrated to a real seeded registry. Touched test files: **125 passed**.

Remaining `tests/supervisor` failures are pre-existing on the stack and unrelated to this PR: `test_run_idle_teardown.py` (4 tests — a `KeyError: 'update_watcher'` introduced by #970's run.py change, whose `FakeRequest` was never updated; verified failing at this PR's base) plus the environmental `chown`/pyroute2/firecracker failures.

## Stacking

Stacked on #971 (`od/wire-supervisor-owner-auth`). Base this PR on that branch; retarget to `dev` once #971 merges.

### DIFF
--- a/docs/plans/2026-06-10-agent-record-persistence-design.md
+++ b/docs/plans/2026-06-10-agent-record-persistence-design.md
@@ -1,0 +1,336 @@
+# Agent-owned record persistence + message reads off the pool
+
+**Status:** design
+**Date:** 2026-06-10
+**Series:** message-free supervisor (Phase-0 residual cleanup, after #969 expiry, #970 update-watching, #971 owner-auth)
+**Stacks on:** `od/wire-supervisor-owner-auth` (#971)
+
+## 1. Goal
+
+Kill the last agent reach-in on the create path — the `create_vm_execution`
+readback (`pool.executions[vm_hash]` → `MessageSpec` attach → `execution.save()`)
+— by making the agent persist its own `ExecutionRecord`, and migrate the
+remaining agent-side readers of `pool.executions[...].message` to the agent
+registry. After this PR, nothing agent-side learns a VM's message by reaching
+through the hypervisor's pool object.
+
+## 2. Background
+
+The message-free-supervisor effort splits the `VmExecution` god-object into an
+agent-side concept (message, owner, billing, expiry, update-watching — backed by
+the `AgentVmRegistry` + agent DB) and a hypervisor-side `Vm` (process,
+networking, status). The supervisor's spec create path is already message-free —
+but `create_vm_execution` immediately undoes it (run.py:249-254):
+
+```python
+execution = pool.executions[vm_hash]                  # reach into the hypervisor's bookkeeping
+execution.spec = MessageSpec(message=content, ...)    # glue the message back on
+await execution.save()                                # have IT write the agent's DB record
+```
+
+The comment on the block says TEMPORARY. It exists for two reasons:
+
+1. **Persistence.** `VmExecution.save()` no-ops unless the spec is a
+   `MessageSpec` (models.py:821-825). The attach exists so `save()` writes the
+   message into the agent DB — which `rehydrate_registry` and past-logs
+   owner-auth read back after an agent restart. Delete `save()` with no
+   replacement and the agent forgets spec-created VMs on every reboot.
+2. **In-memory readers.** A few agent-side code paths still learn the message by
+   reading it off pool executions; the attach is what makes that work for
+   spec-created VMs during their first lifetime. (It does NOT survive a restart:
+   restored executions are spec-built with no message — these readers already
+   silently skip them today.)
+
+### The scheduler context (decision 2026-06-10)
+
+The new scheduler provides allocations for PAYG too; the
+`/control/allocations/notify` flow is legacy and will be deprecated. Payment
+enforcement is the scheduler's prerogative — agent-side payment checking is on
+the deprecation path. Consequences for this design:
+
+- `update_allocations`' reap protection is the allocation set itself; its
+  payment-tier guards are transition-window defense, not the primary mechanism.
+- `check_payment`'s logic is NOT extended or redesigned here. This PR only moves
+  **where its data comes from** (registry instead of pool-message), because the
+  transition window is long enough that the checks still matter — including for
+  restored VMs, which currently escape them entirely.
+
+## 3. Scope
+
+### In scope
+
+1. **`persist_record` (vm_registry.py)** — the write-side sibling of
+   `rehydrate_registry`: the agent writes an `ExecutionRecord` directly from an
+   `AgentVmRecord`.
+2. **`create_vm_execution` (run.py)** — spec path drops the readback / attach /
+   `save()`; calls `persist_record`; returns `None`.
+3. **Payment grouping moves to the agent** — `pool.get_executions_by_address`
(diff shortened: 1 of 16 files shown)


foxpatch-aleph (approved):
Correctly eliminates the create_vm_execution readback by making the agent persist its own ExecutionRecord and migrating all agent-side message readers (payment grouping, update_allocations guards, domains aggregate) to the AgentVmRegistry. Well-structured plan, clean commits, thorough tests including round-trip persistence and regression guards. No correctness or security issues found.

**`src/aleph/vm/orchestrator/run.py`** (line 222): docstring says `Spec-eligible messages (QEMU instances)` — the spec path also handles Firecracker instances (condition is content type, not hypervisor). Consider "Spec-eligible messages (instances)" for accuracy.

**`src/aleph/vm/orchestrator/tasks.py`** (line 329): variable name `executions_by_address` — the key is an address string, but Dict typing says `dict[Chain, list[VmExecution]]`. It's a dict that maps address -> (chain -> list). Name is fine, just noting to double-check the type annotation on the outer dict matches intent.

**`src/aleph/vm/orchestrator/tasks.py`** (line 327): the `payment` variable shadows the import from aleph_message.models. Not a bug (the import is used elsewhere in the file; the local shadows it only in this function scope), but slightly confusing on first read. Not worth blocking on.

**`tests/supervisor/test_views.py`** (line 1439): the `mocker` fixture parameter is unused in `test_update_allocations_spares_payg_via_registry`. Not a problem (pytest ignores extra params), but `mocker` is also not used in that test. Minor.

**`docs/plans/2026-06-10-agent-record-persistence-design.md`** (line 200): the `ExecutionRecord` table reference in the design doc is correct. No issues.

foxpatch-aleph (approved):
Cleanly removes the create_vm_execution readback by (1) adding persist_record to vm_registry.py, (2) migrating all agent-side message readers (payment grouping, update_allocations guards, domains aggregate) to the AgentVmRegistry, (3) deleting the dead VmExecution payment properties, and (4) adding guard tests including an empty-pool routing regression test. The implementation follows the design doc precisely, has excellent test coverage, and avoids scope creep.

**`src/aleph/vm/orchestrator/vm_registry.py`** (line 141): Consider whether `persistent` should default to `False` in `persist_record` like it does on the dataclass, or whether callers should always explicitly pass it. Not a bug — the current call site does pass it — but worth noting if more callers are added later.

**`src/aleph/vm/orchestrator/tasks.py`** (line 316): The function requires a `VmPool` but only accesses `pool.executions`. Since this is a transitional piece that will be replaced by supervisor list views in Phase 1, this is fine — no need to over-abstract.

**`src/aleph/vm/orchestrator/tasks.py`** (line 320): Nit: This line is duplicated from the old pool method. Non-blocking since the function is private, transitional, and these comments still provide useful context to readers.

**`tests/supervisor/test_tasks_registry_reads.py`** (line 135): Good edge case test for an unrecorded execution — the registry.get returns None and short-circuits before the address comparison. This validates the walrus-operator guard logic in `_handle_domains_aggregate`.

**`src/aleph/vm/orchestrator/run.py`** (line 342): The `execution` variable is used after the guard block (e.g. `execution.vm_id`) — since the guard raises for the None case, the type checker would need an `assert execution is not None` after the guard or the user would need type-narrowing. At runtime this works fine, but consider adding `assert execution is not None` after the guard for type-narrowing purposes.

**`docs/plans/2026-06-10-agent-record-persistence-design.md`** (line None): Design doc is thorough and well-reasoned. Good job documenting the scheduler context and the rationale for not touching check_payment's logic.

**`tests/supervisor/test_checkpayment.py`** (line 258): The assertion change from `registry.forget.assert_not_called()` to `assert ItemHash(hash) in registry` is more meaningful — it actually verifies the VM wasn't removed rather than just checking a mock was called.

**`src/aleph/vm/orchestrator/run.py`** (line 216): The docstring should clarify that `persistent` is only relevant for non-spec (legacy pool) paths — in the spec path it's always forced to True (instances are always persistent). This is not wrong, just implicit.

PR #974, opened 2026-06-10
fix(connectivity): replace flaky example.org DNS probe with dual-stack fallback

odesenfans:
## Problem

The droplet integration test (`curl --fail .../status/check/fastapi`) has been flaky. Root cause: the DNS connectivity probe resolves **`example.org`** — an RFC 2606 *documentation* domain that is not meant as a live, dual-stack resolution target and has had recurring DNS flakiness (notably AAAA).

Failure chain:
`status_check_fastapi` runs `check_dns` (a **required** key) → GETs `/dns` from the diagnostic VM (run from the repo's `example_fastapi` via `ALEPH_VM_FAKE_DATA_PROGRAM` in CI) → resolves `example.org` and requires **both** A and AAAA → a flaky lookup returns 503 / missing ipv6 → `check_dns` False → `status_check_fastapi` returns 503 → `curl --fail` fails the job.

## Fix

Replace the single hostname with an ordered list of reliable dual-stack hosts from two **independent** providers, tried until one resolves — mirroring the existing `CONNECTIVITY_HTTP_URLS` fallback pattern already in `conf.py`:

```python
CONNECTIVITY_DNS_HOSTNAMES = ["one.one.one.one", "dns.google"]
```

- Cloudflare `one.one.one.one` first (consistent with the existing `CONNECTIVITY_IPV4_URL = https://1.1.1.1/`), Google `dns.google` as fallback — a different failure domain, so one provider's DNS incident can't take out both.
- `host_status.check_dns_ipv4/ipv6`: first-success loop, tolerating a per-host `OSError`.
- Diagnostic VM `/dns`: loops the hosts, returns 200 once a host yields both families (else 503).
- Still env-overridable (`ALEPH_VM_` prefix) for ad-hoc repointing.

## Tests

Added node-side unit tests in `tests/supervisor/test_status.py` covering: first-host success (fallback not consulted), fallback when the first host has no address, fallback when the first host raises, and all-hosts-fail → False. `test_status.py`: 11 passed. The `/dns` endpoint loop was validated against mocked `getaddrinfo` for the dual-stack / IPv4-only-fallback / error-fallback / no-dual-stack cases.

Note: the `vm.example.org` sentinel in `haproxy.py` is a string comparison, not a DNS lookup — intentionally left unchanged.

A sibling PR applies the same change to `main`.

### DIFF
--- a/.github/workflows/build-deb-package-and-integration-tests.yml
+++ b/.github/workflows/build-deb-package-and-integration-tests.yml
@@ -207,6 +207,10 @@ jobs:
           echo ALEPH_VM_SUPERVISOR_HOST=0.0.0.0 >> supervisor.env
           echo ALEPH_VM_ALLOCATION_TOKEN_HASH=9f86d081884c7d659a2feaa0c55ad015a3bf4f1b2b0b822cd15d6c15b0f00a08 >> supervisor.env
           echo ALEPH_VM_SENTRY_DSN=${{ secrets.SENTRY_DSN }} >> supervisor.env
+          # Pin public resolvers for guest VMs: DNS auto-detection on the droplet
+          # picks up DigitalOcean's VPC-internal resolver (e.g. 10.110.15.254),
+          # which is unreachable from inside the VMs and breaks all guest DNS.
+          echo 'ALEPH_VM_DNS_NAMESERVERS=["1.1.1.1","8.8.8.8"]' >> supervisor.env
           ssh  root@${DROPLET_IPV4} mkdir -p /etc/aleph-vm/
           scp supervisor.env root@${DROPLET_IPV4}:/etc/aleph-vm/supervisor.env
           scp packaging/target/${{ matrix.os_config.package_name }} root@${DROPLET_IPV4}:/opt
@@ -354,6 +358,10 @@ jobs:
           echo ALEPH_VM_SUPERVISOR_HOST=0.0.0.0 >> supervisor.env
           echo ALEPH_VM_FAKE_DATA_PROGRAM=/opt/examples/example_fastapi >> supervisor.env
           echo ALEPH_VM_FAKE_DATA_RUNTIME=/opt/rootfs.squashfs >> supervisor.env
+          # Pin public resolvers for guest VMs: DNS auto-detection on the droplet
+          # picks up DigitalOcean's VPC-internal resolver (e.g. 10.110.15.254),
+          # which is unreachable from inside the VMs and breaks all guest DNS.
+          echo 'ALEPH_VM_DNS_NAMESERVERS=["1.1.1.1","8.8.8.8"]' >> supervisor.env
           ssh  root@${DROPLET_IPV4} mkdir -p /etc/aleph-vm/
           scp supervisor.env root@${DROPLET_IPV4}:/etc/aleph-vm/supervisor.env
           scp packaging/target/aleph-vm.debian-12.deb root@${DROPLET_IPV4}:/opt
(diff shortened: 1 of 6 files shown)


odesenfans:
## CI failure root-caused — two infra bugs, both fixed in 1273d977

**1. Guest DNS totally broken on the CI droplets (the actual red CI).**
The supervisor's startup settings dump in the failed run shows DNS auto-detection picked up `ALEPH_VM_DNS_NAMESERVERS = ['10.110.15.254']` — a DigitalOcean VPC-internal resolver. Guests can't reach it through the host NAT, so every in-VM hostname lookup fails (`socket.gaierror EAI_AGAIN`) while raw-IP egress works (`https://208.67.222.222` → `result: true` in the same run). This breaks `/dns` and `/internet` for *any* probe hostname — swapping example.org alone couldn't fix it. Fix: pin `ALEPH_VM_DNS_NAMESERVERS=["1.1.1.1","8.8.8.8"]` in both CI `supervisor.env` blocks. (The earlier "example.org flakiness" was most likely this same breakage.)

**2. `ALEPH_VM_FAKE_DATA_PROGRAM` has been dead since Oct 2023 (`c827c20d`).**
`cli.py` unconditionally wrote `args.fake_data_program` (argparse default `None`) over the settings, discarding the env var from `supervisor.env`. Proof from the failed run: the settings dump shows `ALEPH_VM_FAKE_DATA_PROGRAM = None`, and the guest traceback (`/opt/main.py` line 170, bare `getaddrinfo`) matches the January-published diagnostic VM code (`dec8f47b`), not this branch (line 242, wrapped in `try/except OSError`). Consequence: the "Test new runtime" job has been testing the *published* program and *downloaded* runtime, not the local ones (`get_runtime_path()` only honors `FAKE_DATA_RUNTIME` when `FAKE_DATA_PROGRAM` is set). Fix: only override the setting when `-f/--fake-data-program` is actually passed.

With both fixes, the "Test new runtime" job now genuinely runs this branch's `example_fastapi` (including the `/dns` fallback) on the locally built runtime.

Verified: systemd `EnvironmentFile` passes the JSON list through verbatim (tested against real systemd) and pydantic-settings parses it; unit suites pass.

foxpatch-aleph (approved):
Sound fix for a flaky DNS probe — replaces the single RFC 2606 documentation domain (example.org) with a fallback chain of reliable dual-stack hosts from independent providers, mirroring the existing CONNECTIVITY_HTTP_URLS pattern. CLI.py's FAKE_DATA_PROGRAM override condition is corrected to respect the env var. CI workflow gets the venv volume build dependency. Tests cover the first-success fallback logic comprehensively. No security concerns. Minor event-loop blocking from the pre-existing sync getaddrinfo call is noted but outside this PR's scope.

**`src/aleph/vm/orchestrator/views/host_status.py`** (line 94): `resolve_dns` calls `socket.getaddrinfo` which is blocking — this ties up the asyncio event loop. The original code had this issue and the PR doesn't introduce it, but the new loop now compounds the problem (multiple hosts = multiple event-loop blocks). Consider wrapping in `asyncio.to_thread` or `loop.run_in_executor` in a follow-up.

**`examples/example_fastapi/main.py`** (line 257): Cumulative `ipv4 = ipv4 or host_ipv4; ipv6 = ipv6 or host_ipv6` can produce a 200 response where ipv4 and ipv6 come from different providers, meaning no single host is actually dual-stack-reachable. This doesn't break the integration test (which checks ipv4 and ipv6 independently), but it's semantically imprecise for the "resolves to both families" claim in the docstring.

PR #975, opened 2026-06-10
fix(connectivity): replace flaky example.org DNS probe with dual-stack fallback

odesenfans:
Backport of #974 to `main`.

## Problem

The droplet integration test (`curl --fail .../status/check/fastapi`) is flaky because the DNS connectivity probe resolves **`example.org`** — an RFC 2606 *documentation* domain not meant as a live, dual-stack target, which has had recurring DNS flakiness (notably AAAA). The diagnostic VM's `/dns` requires **both** A and AAAA, so a flaky lookup returns 503 → `check_dns` False → `status_check_fastapi` 503 → job fails.

## Fix

Replace the single hostname with an ordered list of reliable dual-stack hosts from two independent providers, tried until one resolves — mirroring the existing `CONNECTIVITY_HTTP_URLS` fallback:

```python
CONNECTIVITY_DNS_HOSTNAMES = ["one.one.one.one", "dns.google"]
```

- Cloudflare `one.one.one.one` first (matches `CONNECTIVITY_IPV4_URL = https://1.1.1.1/`), Google `dns.google` fallback (independent failure domain).
- `host_status.check_dns_ipv4/ipv6`: first-success loop, tolerating per-host `OSError`.
- Diagnostic VM `/dns`: loops hosts, 200 once a host yields both families else 503.
- Still env-overridable.

Identical change to #974 (the `dev` PR). `vm.example.org` in `haproxy.py` is a string sentinel, not a DNS lookup — left unchanged.

## Tests

`tests/supervisor/test_status.py`: 11 passed (added first-host-success, fallback-on-no-address, fallback-on-error, all-fail cases).

PR #978, opened 2026-06-11
perf(ci): cache slow build steps and skip redundant work

odesenfans:
## Summary

Cuts PR feedback time roughly in half once caches are warm:

| Workflow | Before | Expected after |
|---|---|---|
| Build Packages and tests on Droplets | 12–18 min | ~8–10 min |
| py.test and linting | ~10 min | ~7.5 min |

Measured sinks and the corresponding changes:

- **Cache the sevctl binary** in the package-build jobs. `cargo install` at the rev pinned in `packaging/Makefile` costs ~2 min in every build job, and the builds gate all droplet tests. `target/bin/sevctl` is a Makefile file target, so the restored binary skips the cargo build entirely. Key includes the Makefile (pins the rev) and the per-OS dockerfile (pins the toolchain).
- **Cache the debootstrapped `aleph-debian-12-python` runtime** (~2 min) in both the pytest workflow and the `build_rootfs` job, keyed on the runtime build scripts plus a monthly stamp so the debootstrap/pip content cannot go stale for more than a month.
- **Reuse the `build_rootfs` artifact** in the fake-data droplet test instead of rebuilding the same runtime in-job (~2 min on the longest job, which already had a comment suggesting this).
- **Run the full `apt-get upgrade` of droplets only on pushes to main** (~1 min per droplet job). PRs skip it; main keeps catching incompatibilities with the latest distro packages before release.
- **Cancel superseded runs on new PR pushes**, which also frees the repo-global droplet concurrency locks earlier. Droplet cleanup steps are `if: always()`, so cancelled runs still delete their droplets.

The first run on a branch only populates the caches; the speedup shows from the second run on.

## Note

Stacked on #975 (`od/fix-dns-connectivity-probe-main`): the droplet tests need its DNS-pin and venv-volume fixes, and the changes touch the same lines. This PR targets main so the workflows trigger; it shows #975's commits until that one merges, after which it can be rebased.

## Possible follow-ups (not in this PR)

- Prebuild the packaging container images (each build job still spends ~1 min pulling `rust:1.79-bookworm` and apt-upgrading it).
- Per-run droplet names to stop concurrent PRs from serializing on the global droplet concurrency groups.
- `pytest-xdist` for the ~4 min unit-test step (needs care: tests create network interfaces as root).

### DIFF
--- a/.github/workflows/build-deb-package-and-integration-tests.yml
+++ b/.github/workflows/build-deb-package-and-integration-tests.yml
@@ -12,6 +12,12 @@ on:
 # This action build the package for Ubuntu and Debian
 # Then run e2e integration tests on Digital Ocean droplets (on demand VM)
 
+# Cancel superseded runs when new commits are pushed to a PR branch.
+# This also releases the global droplet concurrency groups earlier.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
 jobs:
   build_deb:
     name: "Build ${{ matrix.os }} Package"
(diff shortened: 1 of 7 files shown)


foxpatch-aleph (approved):
Adds TTL-based caching to DNS IPv4/IPv6 connectivity checks, skipping redundant DNS lookups when a recent result exists. Implementation is clean and correct: the class-level cache stores the last result and timestamp, and the check functions return cached values when within the TTL. New tests cover cache hit, cache expiry, and the 2-second jitter logic.

**`src/aleph/vm/orchestrator/views/host_status.py`** (line 23): Using class-level mutable attributes for cache is a reasonable trade-off for simplicity. Since these are module-private helpers behind the views module, there's no concern about cross-test interference in production, and the tests reset state via `ConnectivityInfo.reset()`.

**`tests/orchestrator/test_host_status.py`** (line 301): This test modifies the global `ConnectivityInfo._dns_ipv4_last_checked` which could affect other tests in the same process. Since the test suite appears to set `CONNECTIVITY_DNS_CACHE_TTL` and consistency is maintained, this is OK, but a fixture-based reset would be more robust.

PR #979, opened 2026-06-16
build(deb): mirror pyproject dependency versions in packaging Makefile

odesenfans:
## What

The Debian package pip-installs a curated subset of dependencies (the ones not provided by apt) with version pins maintained by hand in `packaging/Makefile`. These had drifted from `pyproject.toml`, so the `.deb` shipped different versions than the code is tested against:

| package | .deb shipped | pyproject (tested) |
|---|---|---|
| aleph-message | `~=1.0.1` | `~=1.1` |
| sentry-sdk | `==1.31.0` | `==2.8` |
| python-cpuid | `==0.1.0` | `==0.1.1` |
| solathon | `==1.0.2` | `==1.0.7` |

`aleph-message` is the most consequential: `~=1.0.1` excludes the `1.1` line the current code targets.

## Change

Align every shared specifier in the Makefile pip line to its `pyproject.toml` counterpart verbatim, and add a comment requiring the two to stay in sync. No packages added or removed; only version specifiers corrected/normalized. `legacy-cgi` stays (packaging-only `cgi` backfill for Python 3.13+).

### DIFF
--- a/packaging/Makefile
+++ b/packaging/Makefile
@@ -18,8 +18,10 @@ debian-package-code:
 	# Making a venv to build the wheel. to work arround a strange problem while building the wheel
 	python3 -m venv build_venv
 	build_venv/bin/pip install   --progress-bar off --upgrade pip setuptools wheel
-	# Fixing this protobuf dependency version to avoid getting CI errors as version 5.29.0 have this compilation issue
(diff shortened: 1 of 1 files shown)


foxpatch-aleph (approved):
Straightforward PR updating Debian packaging Makefile pip dependency versions to match pyproject.toml, with helpful inline comments explaining the protobuf pin and legacy-cgi inclusion. All version specifiers verified correct against pyproject.toml. No packages added or removed; only pin corrections and specifier normalization.

**`packaging/Makefile`** (line 21): Good explanatory comments covering why protobuf is pinned and why legacy-cgi is needed. Helpful for future maintainers.

PR #981, opened 2026-06-17
feat(supervisor): Phase 2 - gRPC wire protocol + two-service packaging

odesenfans:
## Phase 2: complete the gRPC wire protocol + two-service packaging

> **Stacked on #980** (Phase 1). Base branch is `od/grpc-only-supervisor`; review/merge #980 first. Diff here is Phase 2 only.

Phase 1 (#980) decoupled the agent from the pool through the `Supervisor`
interface, wired to the embedded `LocalSupervisor`. **Phase 2 makes the same
interface work over gRPC and ships the two-process split.** Design:
`docs/plans/2026-06-16-grpc-only-supervisor-design.md`; plan:
`docs/plans/2026-06-17-grpc-only-supervisor-phase2-implementation.md`.

### Part A — complete the wire protocol
- **Wired the 4 remaining client stubs** over gRPC (each had no proto RPC):
  `recreate_network`, `restore_from_image`, `run_program_code` (msgpack-framed
  scope), `reserve_resources`. No `NotImplementedError("wired in Phase 2")`
  remains (guard test enforces this).
- **Carried 8 silently-dropped fields**: `TeeConfig.firmware_path`,
  `GpuSpec.device_id`/`model`, `CreateVmSpec.owner_address`,
  `Measurement.sev_info`+`launch_measure` (new `SevInfo` message),
  `BackupInfo.checksum`/`volumes`/`source_sizes`, `StartBackup.include_volumes`.
- **Reconciled HostInfo** (hardware + reservation/inventory fields).
- **`reserve_resources` carries a message-free `ReservationRequest` DTO** — the
  agent translates the Aleph message; no `aleph_message` type crosses the wire
  (verified: zero in `proto_convert`/`grpc_server`/`grpc_client`/`types`/`abc`).
  Pool gained a shared `check_capacity` (unifying the message/spec/reserve paths)
  and a `reserve_gpus` keyed on `device_id`; orphaned message-based
  `pool.reserve_resources` removed.
- **Removed the orphan migration RPCs** (`ExportVm`/`ImportVm`/`GetMigrationStatus`).
- **Guard tests**: no stubs remain; every RPC has a server handler.

### Part B — two-service packaging
- `aleph-vm-supervisor.service` becomes the **daemon** (`python -m aleph.vm.supervisor`,
  owns the pool, serves gRPC); new `aleph-vm-agent.service` runs the HTTP
  orchestrator and dials the socket. `postinst` migrates the old single unit
[...]

PR #982, opened 2026-06-17
refactor(supervisor): retype VM identity from ItemHash to VmId

odesenfans:
## What

Removes the last `aleph_message` symbol from `src/aleph/vm/pool.py` by retyping the daemon's VM-identity from `aleph_message.models.ItemHash` to the message-free `VmId = NewType("VmId", str)` (defined in `supervisor/types.py`). `VmExecution.vm_hash` is retyped to match so the pool's `dict[VmId, VmExecution]` key type stays honest.

This is the VM-identity slice of the "remove `aleph_message` from the supervisor" follow-up flagged during Phase 2. It is intentionally scoped: `models.py` keeps its message-content imports (`ExecutableContent`/`InstanceContent`/`ProgramContent`), which drive the message path. Removing the message object from `VmExecution` entirely is a separate, larger refactor.

## Why it's safe

- `ItemHash` is a `str` subclass and `VmId` a `str` `NewType`, so stored keys and lookups stay compatible (`hash`/`==` identical). `local.py` already looked the pool up with `VmId` values; the create path now stores `spec.vm_id` directly instead of re-wrapping it as `ItemHash`.
- No daemon consumer relies on `ItemHash`-specific behavior (no `isinstance`, no `.item_hash`, no pydantic coercion); consumers either `str()` the value or use it as a plain string dict key.
- The only behavioral delta is dropping `ItemHash` format validation on the spec path. Content-consuming helpers (`cloudinit.get_hostname_from_hash`, `hostnetwork.allocate_vm_ipv6_subnet`) still assume a valid hex hash; the agent always supplies one, so a malformed id now fails inside those consumers rather than eagerly at `ItemHash()`. Out of scope, flagged for awareness.

## Verification

- `pool.py` confirmed free of `ItemHash`/`aleph_message`.
- ruff format + isort clean; mypy adds zero new errors (2 pre-existing env-noise at the same spot on the base).
- Tests: supervisor **901 passed** (7 pre-existing environment failures: jailman `chown` subprocess + pyroute2 netlink), migration+network **26 passed**.
- Reviewed by an adversarial code-quality pass: APPROVED, no defects.

## Base

Stacked on #981 (`od/grpc-only-supervisor-phase2`); `pool.py` here is already the Phase 2 pool. Merge after #981.

### DIFF
--- a/src/aleph/vm/models.py
+++ b/src/aleph/vm/models.py
@@ -7,12 +7,7 @@ from dataclasses import dataclass
 from datetime import datetime, timezone
 from enum import Enum
 
-from aleph_message.models import (
-    ExecutableContent,
-    InstanceContent,
-    ItemHash,
-    ProgramContent,
-)
+from aleph_message.models import ExecutableContent, InstanceContent, ProgramContent
 from aleph_message.models.execution.environment import (
     GpuProperties,
     HypervisorType,
@@ -61,7 +56,7 @@ from aleph.vm.orchestrator.metrics import (
     save_record,
 )
 from aleph.vm.resources import GpuDevice, HostGPU
-from aleph.vm.supervisor.types import Backend, CreateVmSpec
+from aleph.vm.supervisor.types import Backend, CreateVmSpec, VmId
 from aleph.vm.systemd import SystemDManager
 from aleph.vm.utils import dumps_for_json
 from aleph.vm.utils.aggregate import get_user_settings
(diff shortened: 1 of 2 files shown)


foxpatch-aleph (approved):
Clean, mechanically sound refactor that replaces ItemHash with VmId in pool.py and models.py, completing the removal of aleph_message from the pool. The change is type-safe because both types are str-compatible at runtime (same hash/eq semantics), leaving dict key behavior unaffected. The only behavioral delta — deferred ItemHash format validation — is correctly acknowledged in the description as out of scope.

foxpatch-aleph (approved):
This PR cleanly retypes VM identity from ItemHash to VmId in pool.py and models.py, removing the last aleph_message symbol from the pool module. The change is mechanically safe (ItemHash is a str subclass, VmId is a str NewType — dict key semantics are identical) and correctly propagates the type through all pool method signatures. The only cosmetic point is the redundant str() wrapping on VmId(str(config.vm_hash)) where config.vm_hash is already a str, which is harmless.

PR #983, opened 2026-06-17
refactor(supervisor): make VmExecution message-free; HardwareResources; delete dead AlephFirecrackerProgram

odesenfans:
## What

Makes the daemon's `VmExecution` (the execution object the VM pool owns) fully message-free, and removes the `aleph_message` value type `MachineResources` from the controller layer.

- **`VmExecution` is now spec-only.** Removed `MessageSpec` and the `spec` union, `.message`/`.original`, `prepare_gpus`, `fetch_port_redirect_config_and_setup`, the unused `hypervisor` property, the dead `run_code`, and `save()` (spec-built executions keep no DB record; the agent persists via its own `VmRegistry`). The message-driven half was dead in production: every execution is built via `from_spec`, and the agent reads messages from its registry, never from `execution.message`. `models.py` now imports zero `aleph_message`.
- **`MachineResources` -> `HardwareResources`.** New frozen `HardwareResources(vcpus, memory)` DTO in `supervisor/types.py`, swapped into the five surviving controllers. Field names match so reads are unchanged; no serialization change (the on-disk `Configuration` already uses `vcpu_count: int`). `seconds` is intentionally dropped: it was read only by the deleted program controller, and the run timeout is agent-side (passed over the gRPC run call).
- **Deleted the dead message-path program controller** `AlephFirecrackerProgram` (+ `ProgramVmConfiguration`). Production programs run via `SpecFirecrackerProgram` + the channel run path. The agent-shared symbols in `program.py` (`AlephProgramResources`, `ProgramConfiguration`, `ConfigurationResponse`, `RunCodePayload`, the `ConfigurationPayload*` chain, `Interface`, `read_input_data`, `FileTooLargeError`) are kept; the now-orphaned `get_volumes_for_program` was removed (the agent reimplements it as `build_code_and_volumes`).
- Updated `LocalSupervisor` (`local.py`) for the spec-only world (`_backend_of` derives the backend from `is_program`; removed the dead `MessageSpec`/`fetch_port_redirect` branch).

## Scope note (the controllers stay agent-side, by design)

[...]

PR #980, opened 2026-06-17
feat(supervisor): Phase 1 - decouple the agent from the VM pool

odesenfans:
## Phase 1: decouple the agent from the VM pool

This is **Phase 1** of making the two-process split the only supported
production deployment (design: `docs/plans/2026-06-16-grpc-only-supervisor-design.md`).
Phase 1 makes the agent reach VMs **exclusively** through the `Supervisor`
interface, wired to the embedded `LocalSupervisor`. Production stays
single-process; Phase 2 adds the gRPC transport + two-service packaging.

### What changed

- **`InProcessSupervisor` → `LocalSupervisor`** (`local.py`): the pool-backed
  engine, with an explicit `build_supervisor()` factory selecting embedded vs gRPC.
- **Agent is now pool-free.** Every endpoint that reached around the interface
  (`require_vm_pool()`, `app["vm_pool"]`, `.executions`, `create_a_vm`,
  `_engine_pool`) is gone. Enforced by a guard test (`test_agent_pool_free.py`).
- **Backup/restore, confidential init/measurement/inject, `recreate_network`,
  persistent programs** all route through the interface.
- **Admission + GPU reservation folded into `create_vm`** atomically
  (no cross-process TOCTOU).
- **HAProxy/L7 routing moved fully agent-side** (`haproxy_sync.py`), driven by
  `list_vms()`. The supervisor no longer knows HAProxy exists.
- **Migration** rides standard lifecycle RPCs (`stop_vm`/`start_vm`/`create_vm`/
  `delete_vm`); the directory-based flow and the `MigrationOps` ABC are deleted.
- **Confidential SEV policy** now flows from the request through to the QEMU
  `sev-guest,policy=` arg; the engine holds no default opinion.
- **Dropped the dead Firecracker-instance concept**; instances are QEMU-only,
  and the default hypervisor is now QEMU.

### Not in this PR (Phase 2)

gRPC wire protocol for the new methods/fields, two-service systemd packaging,
flipping the prod default to gRPC, removing the orphan `ExportVm`/`ImportVm`
proto RPCs.

### Testing

887 passed, ruff/isort clean, mypy at baseline. New `LocalSupervisor` engine
unit tests + agent-against-fake-supervisor tests.

PR #984, opened 2026-06-18
refactor(supervisor): drop dead in-process QEMU cloud-init path

odesenfans:
## What

Removes the now-dead in-process QEMU cloud-init `configure()` path that still read `message_content`, which is `None` on the message-free spec path (the only path used in production after the gRPC-only-supervisor refactor, #980–#983).

## Why

Production QEMU instance creation goes through `VmPool.create_vm_from_spec`, which builds cloud-init from the spec via `build_qemu_configuration` → `build_cloud_init_drive(ssh_authorized_keys=spec.ssh_authorized_keys)` and starts with `start(write_config=False)`. The in-process `AlephQemuInstance.configure()` → `CloudInitMixin._create_cloud_init_drive` path (which reads `resources.message_content`) is only reachable via `start(write_config=True)`, which production QEMU never uses. Left in place it crashes with `AttributeError: 'NoneType' object has no attribute 'authorized_keys'` if ever hit on the spec path (this is what tripped `test_qemu_instance` until #983 pointed it at the production flow).

## Changes (−193 lines)

- Delete `CloudInitMixin._create_cloud_init_drive` + the now-empty `CloudInitMixin`.
- Delete `AlephQemuInstance.configure()` / `AlephQemuConfidentialInstance.configure()` and their dead `save_controller_configuration` methods/attrs.
- Trim imports that become unused.

## Kept (verified live)

- Spec-based cloud-init in `supervisor/qemu_build.py` (`build_cloud_init_drive`, `create_cloud_init_drive_image`).
- `AlephQemuResources.message_content` + `download_*` — still used by `translate.build_create_vm_spec`.
- Firecracker `configure()` (separate class, reached via the live `start(write_config=True)` FC path).

## Verification

- mypy: no new errors (baseline-only).
- `tests/supervisor/` + `tests/migration/`: pass except known environment-only failures (`test_interfaces` pyroute2-needs-root, `test_log` journald); `test_qemu_instance` stays at its 2 expected XFAILs (no fake disk).

Pure dead-code removal; no functional change to production.

### DIFF
--- a/src/aleph/vm/controllers/qemu/cloudinit.py
+++ b/src/aleph/vm/controllers/qemu/cloudinit.py
@@ -15,16 +15,12 @@ See also the cloud-localds  man page (1)
 
 import base64
 import json
-from pathlib import Path
 from tempfile import NamedTemporaryFile
 
 import yaml
 from aleph_message.models import ItemHash
 
-from aleph.vm.conf import settings
-from aleph.vm.controllers.interface import AlephVmControllerInterface
-from aleph.vm.hypervisors.firecracker.config import Drive
-from aleph.vm.utils import is_command_available, run_in_subprocess
+from aleph.vm.utils import run_in_subprocess
 
 
 def get_hostname_from_hash(vm_hash: ItemHash) -> str:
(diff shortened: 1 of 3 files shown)


foxpatch-aleph (approved):
Pure dead-code removal of the in-process QEMU cloud-init configure() path that was unreachable in production (production uses build_qemu_configuration + start(write_config=False)). All removed imports are confirmed unused in remaining code, all references to CloudInitMixin and _create_cloud_init_drive are gone from the source, and the Firecracker configure() path is untouched. The only minor nit is stale docstring references in qemu_build.py that mention the now-deleted configure() methods.

**`src/aleph/vm/supervisor/qemu_build.py`** (line 4): Docstring references `AlephQemuInstance.configure()` which no longer exists after this PR. Consider updating to note the original method was removed (e.g., "mirrors the original in-process configure() path, removed in #984"). Non-blocking — historical context only.

Issue #27, opened 2021-06-04
Request body isn't passed to the VM code


moshemalawach:
There is two parts to consider:
- Pass the Body
- Recover it on VM side and pass it to python app and http apps.

At first we will pass it fully, but in the future having streaming will be a good addition

moshemalawach:
Works for HTTP requests now. Only need to pass it to ASGI now.

PR #985, opened 2026-06-18
fix(confidential): register agent record early, wait for awaiting-init

odesenfans:
## What

Replaces the registry poll-bridge in the confidential init endpoint with **early owner-record registration**, so owner-auth resolves immediately (matching pre-1.13's pool, which was populated at the start of create) instead of polling the registry for up to 120 s.

## Why

A confidential instance is created via the allocation/spec path: `create_vm_execution` runs `await supervisor.create_vm(spec)` (~20 s) and only then recorded the agent registry entry. The scheduler exposes placement ~6 s in, so the owner's one-shot `init-session` could arrive before the record existed — previously a 404, then mitigated by `wait_for_agent_record_or_404` polling for the record.

## Design

Separates **identity** (early) from **readiness** (inherent wait):

- `create_vm_execution` now records the owner identity **before** `create_vm` (with `registry.forget(vm_hash)` cleanup if create raises, so a failed create leaves no dangling record).
- `operate_confidential_initialize` resolves `get_agent_record_or_404` immediately (no long poll), runs owner-auth, then — because the VM only reaches `awaiting_confidential_init` when `create_vm` completes — waits on that state via `wait_for_awaiting_confidential_init` (polls `get_vm`, bounded 120 s) before calling `initialize_confidential`. Guards are made `None`-safe for the window where the supervisor hasn't registered the VM yet.

The `awaiting_confidential_init` exemption in the `vm_running` guard and the post-inject `reconcile_port_forwards` (both from #981) are unchanged.

## Verification

- mypy: no new errors (operator.py baseline multipart errors only).
- `test_operator.py` (incl. confidential) + run/registry/allocation tests pass. `test_operator_confidential_initialize` updated to drive `get_vm` in the realistic `awaiting_confidential_init` state.

## ⚠️ Needs SEV-testnet validation

The real interleaving (placement ~6 s vs. record-at-start vs. awaiting-init at ~20 s) is only fully exercised on the AMD SEV-SNP testnet. Recommend an aleph-testnets #27 run against this branch confirming `init-session` no longer needs the long poll and that initialize succeeds on the first awaiting-ready observation before merging.

### DIFF
--- a/src/aleph/vm/orchestrator/run.py
+++ b/src/aleph/vm/orchestrator/run.py
@@ -270,13 +270,26 @@ async def create_vm_execution(
         # reservation and skips other users' valid reservations during create.
         # The agent does not touch reservations at all.
         spec = await build_create_vm_spec(vm_hash, content)
-        info = await supervisor.create_vm(spec)
-        # Agent territory: record the message in the agent's own cache. This is
-        # what the message-free agent will read once owner-auth and billing move
-        # off the VmExecution (design doc section 5). The supervisor machinery
-        # that created the VM never reads it.
+        # Agent territory: record the message in the agent's own cache *before*
+        # create_vm, mirroring 1.13 where the VM entered the pool at the start of
+        # create. create_vm takes ~20s while the scheduler exposes placement
+        # earlier, so a confidential owner's one-shot init-session call can land
+        # before create returns. Recording the owner identity up front lets the
+        # operator API answer owner-auth immediately (get_agent_record_or_404)
+        # instead of polling for the record; the endpoint still waits for the VM
+        # to reach the awaiting-init state before initializing it. The supervisor
+        # machinery that creates the VM never reads this record.
         # Spec-eligible VMs are QEMU instances, which are always persistent.
         record = registry.record(vm_hash, message=content, original=original_message.content, persistent=True)
+        try:
+            info = await supervisor.create_vm(spec)
+        except Exception:
+            # create_vm failed: drop the early record so a failed create never
+            # leaves a dangling owner-identity entry behind (a record with no VM
+            # the supervisor knows about). Nothing is persisted yet, so forgetting
+            # the in-memory entry is sufficient.
+            registry.forget(vm_hash)
+            raise
         if info.awaiting_confidential_init:
             # A confidential VM is created but not started: only the owner can
             # start it, by uploading the session certificates via
--- a/src/aleph/vm/orchestrator/views/operator.py
+++ b/src/aleph/vm/orchestrator/views/operator.py
@@ -159,13 +159,15 @@ _CONFIDENTIAL_AWAITING_WAIT_SECONDS = 120
 _CONFIDENTIAL_AWAITING_POLL_INTERVAL = 2.0
 
 
-async def wait_for_awaiting_confidential_init(supervisor: Supervisor, vm_id: VmId) -> None:
-    """Block until the VM reaches awaiting_confidential_init, or time out.
+async def wait_for_awaiting_confidential_init(supervisor: Supervisor, vm_id: VmId) -> bool:
+    """Poll until the VM reaches awaiting_confidential_init. Return True once it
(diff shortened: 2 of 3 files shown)


foxpatch-aleph (commented):
The PR correctly separates owner identity registration (early, before create_vm) from readiness waiting (polling for awaiting_confidential_init). The cleanup-on-failure pattern (registry.forget in both the create_vm exception handler and the finish_instance_create exception handler) is correct and prevents dangling records. The None-safe guards handle the window where the supervisor hasn't registered the VM yet. The main gap is test coverage: the updated test mocks get_vm to return awaiting_confidential_init=True immediately, so the new wait_for_awaiting_confidential_init polling loop is never exercised, and the info is None (VmNotFoundError) path is untested. No correctness or security issues found.

**`tests/supervisor/views/test_operator.py`** (line 361): The updated test mocks get_vm to return awaiting_confidential_init=True on the first call, so the wait_for_awaiting_confidential_init loop (operator.py:450-451) is never exercised. Consider adding a test where get_vm first raises VmNotFoundError or returns a non-awaiting VmInfo, then returns awaiting_confidential_init=True on a subsequent call — this would cover the new polling logic and the info is None guard path.

**`tests/supervisor/views/test_operator.py`** (line 361): Consider adding a test for the timeout path: get_vm never returns awaiting_confidential_init=True, and the endpoint returns 404 after the 120s deadline. Use a reduced _CONFIDENTIAL_AWAITING_WAIT_SECONDS or mock asyncio.get_running_loop().time to avoid a real 120s wait.

**`src/aleph/vm/orchestrator/views/operator.py`** (line 179): The 404 body uses vm_id instead of vm_hash. Since VmId is NewType("VmId", str) and vm_id = VmId(str(vm_hash)), they produce the same string, so this is cosmetic. For consistency with other error messages in this file (e.g., line 144 uses vm_hash), consider using vm_hash here too.

foxpatch-aleph (commented):
Well-structured PR that cleanly separates owner-identity recording (early, before create_vm) from VM-readiness waiting (poll for awaiting_confidential_init). The cleanup-on-failure path (registry.forget) is correct since persist_record only runs after create_vm succeeds. Test coverage is good, covering the happy path, multi-poll wait, and timeout. One concern: when the initial get_vm returns VmNotFoundError (info=None), both the vm_running and not_confidential guards are skipped, so a VM that later appears as non-confidential or already-running will be polled for the full 120s before returning a generic 404, instead of failing fast with a descriptive 400 error. This is a latency/UX regression in the error path, not a security issue. Consider adding a fast-fail check inside wait_for_awaiting_confidential_init (or after it returns False) to distinguish 'timed out' from 'VM exists but is not confidential/awaiting'.

**`src/aleph/vm/orchestrator/views/operator.py`** (line 454): When info is None (initial VmNotFoundError), both the vm_running (line 433) and not_confidential (line 442) guards are skipped. If the VM later appears as non-confidential or already RUNNING, wait_for_awaiting_confidential_init will poll for the full 120s before returning False -> 404, instead of failing fast with a descriptive 400. Consider checking the VM state inside the wait loop (or re-checking get_vm after the wait returns False) to distinguish 'timed out' from 'VM exists but wrong state', so non-confidential/running VMs get an immediate descriptive error.

**`src/aleph/vm/orchestrator/views/operator.py`** (line 178): Consider fast-failing here when info is not None and info.confidential_mode is ConfidentialMode.NONE (or info.status is RUNNING without awaiting) — returning False immediately with a distinguishable signal, so the caller can return a 400 'not_confidential' / 'vm_running' instead of waiting 120s for a generic 404.

**`tests/supervisor/views/test_operator.py`** (line 380): Good coverage of the VmNotFoundError -> not-awaiting -> awaiting transition. Missing: a test where the initial get_vm raises VmNotFoundError and the VM that eventually appears is non-confidential (ConfidentialMode.NONE) or already RUNNING — this would expose the 120s-wait-before-404 behavior described in the operator.py comment above.

PR #987, opened 2026-06-19
fix(confidential): record owner identity before resource download

odesenfans:
## Problem

The confidential integration test (`test_confidential_instance_create_and_ssh`, aleph-testnets #27) fails at `init-session` with `VM not found` — reproducibly, on **dev** as well as feature branches. Root-caused from the CRN `tee-agent` journal:

```
11:57:30  agent starts the persistent confidential VM
11:57:36  POST .../confidential/initialize → 404   (init-session, IMMEDIATE)
11:57:46  "…waiting for its owner to initialize"   (reaches awaiting, +10s)
11:57:51→11:59:51  forgotten 1/3 → 2/3 → reaped
```

The 404 is in the **same second** as the POST — not the 120 s awaiting-init wait timing out. It fires **upstream** of that wait, at `get_agent_record_or_404` (`operator.py`), because the agent owner-record isn't in the registry yet.

Why: in `create_vm_execution` (`run.py`, spec/instance branch) the owner record was written **after** `build_create_vm_spec`, which **downloads the confidential rootfs/firmware** (several seconds) before `create_vm` (~20s more). A confidential owner's **single, one-shot** `init-session` lands somewhere in that download+create window; arriving before the record is written → `get_agent_record_or_404` 404s. The aleph CLI's `init-session` is `check=True`, so the call fails hard, the test's `finally` forgets the instance, and `check_payment` then reaps the now-genuinely-forgotten VM.

(This supersedes this PR's previous content — a reconcile-teardown skip for `awaiting_confidential_init` VMs. That was the wrong layer: the `FORGOTTEN` reap is the test's own cleanup of a VM `init-session` could never reach, so suppressing it would only leak the VM.)

## Fix

Record the owner identity **before** `build_create_vm_spec`, so owner-auth resolves the moment create begins (matching the design's stated intent — the record was meant to be written "up front", it just landed after the download). The failure guard is widened to cover the download too.

## Tests

- Regression test (`test_owner_record_recorded_before_resource_download`): asserts the owner record is present when `build_create_vm_spec` runs (fails on the old ordering).
- Local: mypy clean; `test_supervisor_run_routing` 19/19; create-path/operator suite 83/83; isort + ruff clean. No test deleted/skipped/weakened.

Full confidential validation requires the SEV testnet (#27); GitHub CI cannot exercise the SEV path.

### DIFF
--- a/src/aleph/vm/orchestrator/run.py
+++ b/src/aleph/vm/orchestrator/run.py
@@ -265,29 +265,29 @@ async def create_vm_execution(
         return None
 
     if _is_spec_eligible(content):
-        # The spec carries the owner address (build_create_vm_spec reads it from
-        # message.address), so the engine consumes this owner's own GPU
-        # reservation and skips other users' valid reservations during create.
-        # The agent does not touch reservations at all.
-        spec = await build_create_vm_spec(vm_hash, content)
-        # Agent territory: record the message in the agent's own cache *before*
-        # create_vm, mirroring 1.13 where the VM entered the pool at the start of
-        # create. create_vm takes ~20s while the scheduler exposes placement
-        # earlier, so a confidential owner's one-shot init-session call can land
-        # before create returns. Recording the owner identity up front lets the
-        # operator API answer owner-auth immediately (get_agent_record_or_404)
-        # instead of polling for the record; the endpoint still waits for the VM
-        # to reach the awaiting-init state before initializing it. The supervisor
-        # machinery that creates the VM never reads this record.
-        # Spec-eligible VMs are QEMU instances, which are always persistent.
+        # Record the owner identity up front — BEFORE build_create_vm_spec (which
+        # downloads the confidential rootfs/firmware, several seconds) and before
+        # create_vm (~20s). The scheduler exposes placement earlier, so a
+        # confidential owner's one-shot init-session call can land anywhere in that
+        # download+create window. Recording here lets the operator API answer
+        # owner-auth immediately via get_agent_record_or_404; otherwise that lookup
+        # 404s before the awaiting-init wait can even run, the owner's single call
+        # is lost, and the test forgets the instance (which the reconcile then
+        # reaps). The endpoint still waits for the VM to reach the awaiting-init
+        # state before initializing it; the supervisor machinery never reads this
+        # record. (build_create_vm_spec reads the owner address from
+        # message.address for the engine's own GPU-reservation handling; the agent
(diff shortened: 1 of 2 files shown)


foxpatch-aleph (approved):
The fix correctly excludes `awaiting_confidential_init` VMs from the terminal-status teardown loop in `check_payment`, preventing premature destruction before the owner can run `init-session`. Placing the guard before `get_message_status` avoids an unnecessary network call, and popping the strike counter defensively clears any residual strikes accumulated before the VM entered the awaiting state. The new test is well-structured, verifies both that teardown never occurs after >3 reconcile passes and that the message status is never queried, and matches existing test conventions in the file. The change is minimal and well-scoped to the described bug.

foxpatch-aleph (approved):
The diff correctly moves `registry.record` ahead of `build_create_vm_spec` (the confidential rootfs/firmware download) so the owner-identity record is queryable for the entire create window, closing the race where a one-shot `init-session` 404s during download and the instance is subsequently reaped. The `try/except` is correctly widened to cover both build and create, calling `registry.forget` on either failure to avoid a dangling in-memory record. Behavior for the happy and confidential paths is preserved (persist still happens at the end / on awaiting-init). The new regression test directly asserts the record is present inside `fake_build`, matching the fix. No security or correctness issues. Two non-blocking notes: the PR title/description describe a different fix (a `check_payment` reconcile teardown skip in `tasks.py`) than the actual diff, which the commit message does clarify; and the rewritten comment block is informative but verbose and could be trimmed.

**`src/aleph/vm/orchestrator/run.py`** (line 268): The PR title and description describe a different change (skipping `awaiting_confidential_init` VMs in `check_payment`'s terminal-status teardown loop in `tasks.py`, with a test named `test_awaiting_confidential_init_not_torn_down`), but this diff instead moves `registry.record` before `build_create_vm_spec` in `run.py` with a different test. The commit message explains the reconcile skip was \"the wrong layer,\" but the PR description has not been updated to match the actual diff. Consider updating the PR description so reviewers aren't misled about what this change does.

**`src/aleph/vm/orchestrator/run.py`** (line 281): The rewritten comment is thorough and documents non-obvious timing that was the root cause, but it's ~14 lines interleaving several concerns (timing window, owner-auth rationale, GPU reservation, persistence semantics). Consider trimming to the core rationale (record before download so owner-auth resolves during the create window; forget on either build/create failure) and letting the surrounding code/commit message carry the rest. Non-blocking style nit.

PR #989, opened 2026-06-19
refactor(boundary): PR-2 controller split (download/runtime + wire-error vocabulary)

odesenfans:
## PR-2: controller split by concern

Stacked on **#988** (PR-3). Removes the `agent -> supervisor.controllers` import edge so the import-linter contract forbids it strictly. Design: `docs/plans/2026-06-19-controller-split-by-concern-design.md`; implementation plan (audit-grounded, with verified execution notes): `docs/plans/2026-06-19-controller-split-by-concern-implementation.md`.

### What changed
- **Agent-side downloaders** (`aleph.vm.agent.vm.downloader`: `QemuDownloader`, `QemuConfidentialDownloader`, `ProgramDownloader`) — *compose* the host-resource surface (do not subclass the controller `VmResources`), hold `message_content`, own the `download_*` / `make_writable_volume` logic, and raise the **contract** error vocabulary (`supervisor_interface.errors.{ResourceDownloadError,FileTooLargeError,VmSetupError}`). `translate.py` uses them.
- **Controller runtime holders** (`AlephQemuResources` / `AlephQemuConfidentialResources` / `AlephFirecrackerResources` / `AlephProgramResources`) lose their now-unused `download_*` / `make_writable_volume` (download is agent-only); `from_spec` stays.
- **Neutral foundation modules** (importable by both sides): `aleph.vm.host_volumes`, `aleph.vm.program_config`, `aleph.vm.backup_staging`; `get_hostname_from_hash` moved to `aleph.vm.utils`. Controllers re-export for their own importers.
- **Wire-error vocabulary:** `views/__init__.py`, `run.py`, `program_client.py` catch/raise the contract `SupervisorError` set; `operator.py` uses the neutral backup-staging utils.
- **pyproject:** `aleph.vm.supervisor.controllers` added to the agent contract's `forbidden_modules`; the PR-1 documented residual is gone.

### Scope note (review focus)
The literal "two residuals" understated the coupling: the edge was also used by `run.py` and `program_client.py` (shared program-config machinery), so removing it fully required migrating those too. That is the bulk of the diff and the part most worth reviewing.

### Decisions to double-check
1. `operator.py` backup utils relocated to a **neutral util** (`aleph.vm.backup_staging`), not a boundary method — they are pure `settings`/`storage` helpers with no tenant-isolation need.
2. **Confidential/SEV runtime path** is exercised by unit tests with mocks only; the real SEV path needs the testnet #27 run before merge (same gate as the confidential-init work).
[...]

PR #990, opened 2026-06-19
refactor(supervisor): rename vm_hash to vm_id and numeric vm_id to vm_index

odesenfans:
## PR-0: supervisor identity rename (behavior-neutral)

Stacked on **#989** (PR-2). Renames the supervisor-side identity so the code says what it means. Design: `docs/plans/2026-06-19-supervisor-vm-id-rename-design.md`. (Sequenced last, on the final post-PR-1/2/3 layout, rather than first.)

### Two axes
- **numeric local VM id `vm_id: int` → `vm_index`**: `MicroVM`, `VmExecution.vm_index` property + `create()`, `network/` (`allocate_vm_ipv6_subnet`, `setup_nftables_for_vm`, `prepare_tap`, …), `hypervisors/`, controllers, `get_unique_vm_index()`.
- **string identity `vm_hash` → `vm_id`** on the supervisor objects (`VmExecution`, `VmPool` keys/params, `LocalSupervisor`) and agent reads of that attribute (`run.py`, `custom_logs.py`, `tasks.py`, `migration/`).

### Left unchanged (by design)
- the controller's own `vm_hash: ItemHash` param (controllers still receive `vm_hash=<identity>`);
- the **serialized** `Configuration.vm_id`(int)/`vm_hash`(str) fields and the recreate-dict string keys `"vm_id"`/`"vm_hash"` (changing a serialized/JSON key is out of scope, design §4) — converted at the controller boundary instead;
- the wire `CreateVmSpec.vm_id` / `VmInfo.vm_id` (already the string id);
- genuine `ItemHash` locals (agent side).

### Verification (local; no GH CI until the stack reaches dev)
- **mypy**: unchanged at the pre-rename baseline (**43 errors / 13 files**) — the key guard for the str/int collision; every conflation it flagged was fixed to baseline.
- **grep sweep** (design §3): zero `execution.vm_hash` anywhere; remaining `vm_hash` in the supervisor objects are only the protected forms above. *(This sweep caught 5 readers in `custom_logs.py`/`tasks.py`/`migration/helpers.py` that mypy and the run-tests missed — uncovered/qemu-gated paths.)*
- tests (`tests/supervisor tests/migration tests/network`): **930 passed**, 3 expected env-only `test_interfaces` (pyroute2/root) failures, 1 skipped, 2 xfailed.
- whole-tree `--collect-only`: 956 collected, 0 errors.
- import-linter: **4 kept, 0 broken**. isort + ruff clean.
- No test deleted/skipped/xfailed/weakened — test changes are identifier-retargeting only.

### DIFF
--- a/src/aleph/vm/agent/custom_logs.py
+++ b/src/aleph/vm/agent/custom_logs.py
@@ -29,7 +29,7 @@ class InjectingFilter(logging.Filter):
         if not vm_hash:
             vm_execution: VmExecution | None = ctx_current_execution.get(None)
             if vm_execution:
-                vm_hash = vm_execution.vm_hash
+                vm_hash = vm_execution.vm_id
 
         if not vm_hash:
             return False
--- a/src/aleph/vm/agent/tasks.py
+++ b/src/aleph/vm/agent/tasks.py
@@ -257,7 +257,7 @@ async def _handle_domains_aggregate(
     has_local_instance = any(
         execution.is_instance
         and execution.vm
-        and (record := registry.get(execution.vm_hash)) is not None
+        and (record := registry.get(execution.vm_id)) is not None
         and record.message.address == address
         for execution in pool.executions.values()
     )
--- a/src/aleph/vm/hypervisors/firecracker/microvm.py
+++ b/src/aleph/vm/hypervisors/firecracker/microvm.py
@@ -78,7 +78,7 @@ class RuntimeConfiguration:
 
 
 class MicroVM:
-    vm_id: int
+    vm_index: int
     use_jailer: bool
     firecracker_bin_path: Path
     jailer_bin_path: Path | None
@@ -96,15 +96,15 @@ class MicroVM:
     journal_stderr: BinaryIO | int | None = None
 
     def __repr__(self):
-        return f"<MicroVM {self.vm_id}>"
+        return f"<MicroVM {self.vm_index}>"
 
     def __str__(self):
-        return f"vm-{self.vm_id}"
+        return f"vm-{self.vm_index}"
 
     @property
     def namespace_path(self) -> str:
         firecracker_bin_name = os.path.basename(self.firecracker_bin_path)
-        return str(self.jailer_base_directory / firecracker_bin_name / str(self.vm_id))
+        return str(self.jailer_base_directory / firecracker_bin_name / str(self.vm_index))
 
     @property
     def jailer_path(self) -> str:
@@ -115,7 +115,7 @@ class MicroVM:
         if self.use_jailer:
             return f"{self.jailer_path}/run/firecracker.socket"
         else:
-            return f"/tmp/firecracker-{self.vm_id}.socket"
+            return f"/tmp/firecracker-{self.vm_index}.socket"
 
     @property
     def vsock_path(self) -> str:
@@ -126,7 +126,7 @@ class MicroVM:
 
     def __init__(
         self,
-        vm_id: int,
+        vm_index: int,
         vm_hash: ItemHash,
         firecracker_bin_path: Path,
         jailer_base_directory: Path,
@@ -135,7 +135,7 @@ class MicroVM:
         init_timeout: float = 5.0,
         enable_log: bool = True,
     ):
-        self.vm_id = vm_id
+        self.vm_index = vm_index
         self.vm_hash = vm_hash
         self.use_jailer = use_jailer
         self.jailer_base_directory = jailer_base_directory
@@ -263,7 +263,7 @@ class MicroVM:
         options = (
             str(self.jailer_bin_path),
             "--id",
-            str(self.vm_id),
+            str(self.vm_index),
             "--exec-file",
             str(self.firecracker_bin_path),
             "--uid",
@@ -433,7 +433,7 @@ class MicroVM:
             raise MicroVMFailedInitError() from error
 
     async def shutdown(self) -> None:
-        logger.debug(f"Shutdown vm={self.vm_id}")
+        logger.debug(f"Shutdown vm={self.vm_index}")
         try:
             reader, writer = await asyncio.open_unix_connection(path=self.vsock_path)
         except (
@@ -441,7 +441,7 @@ class MicroVM:
             ConnectionResetError,
             ConnectionRefusedError,
         ) as error:
-            logger.warning(f"VM={self.vm_id} cannot receive shutdown signal: {error.args}")
+            logger.warning(f"VM={self.vm_index} cannot receive shutdown signal: {error.args}")
             return
 
         try:
@@ -462,7 +462,7 @@ class MicroVM:
             if msg2 != b"STOPZ\n":
                 logger.warning(f"Unexpected response from VM: {msg2[:20]!r}")
         except ConnectionResetError as error:
-            logger.warning(f"ConnectionResetError in shutdown of {self.vm_id}: {error.args}")
+            logger.warning(f"ConnectionResetError in shutdown of {self.vm_index}: {error.args}")
 
     async def stop(self):
         if self.proc:
@@ -481,7 +481,7 @@ class MicroVM:
         try:
             await asyncio.wait_for(self.shutdown(), timeout=5)
         except asyncio.TimeoutError:
-            logger.exception(f"Timeout during VM shutdown vm={self.vm_id}")
+            logger.exception(f"Timeout during VM shutdown vm={self.vm_index}")
         logger.debug("Waiting for one second for the process to shutdown")
         await asyncio.sleep(1)
         await self.stop()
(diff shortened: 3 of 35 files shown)


foxpatch-aleph (changes requested):
A clean, well-scoped rename PR with a thorough grep sweep and identifier-only test retargeting. However, one change introduced a genuine bug: src/aleph/vm/supervisor/controllers/firecracker/executable.py:248 constructs Configuration(vm_index=self.vm_index, ...), but the Configuration Pydantic model field is vm_id: int (left unchanged by design — serialized keys are out of scope, per design §4). The two sibling QEMU call sites in qemu_build.py correctly use vm_id=vm_id; this Firecracker site must match. As written, saving a persistent Firecracker program VM's controller config will raise a Pydantic ValidationError (unexpected field vm_index, and required vm_id missing). This path is firecracker-persistent/qemu-gated and not covered by the 930 passing tests, which is why it slipped through. Fix: vm_id=self.vm_index. Two minor non-blocking nits: the firewall.py:1338 docstring says vm_index where the actual dict key is vm_id, and local.py:1077 reuses the token vm_id for both the identity and the index in one log line.

**`src/aleph/vm/supervisor/controllers/firecracker/executable.py`** (line 248): Bug: Configuration has field `vm_id: int` (serialized key unchanged by design, see configuration.py:91 and qemu_build.py:170/274 which correctly use `vm_id=vm_id`). Passing `vm_index=` here will fail Pydantic validation when saving a persistent Firecracker program VM config — the field is unexpected and the required `vm_id` is missing. This path is firecracker-persistent and not covered by the 930 supervisor/network/migration tests. Fix: `vm_id=self.vm_index`.

**`src/aleph/vm/network/firewall.py`** (line 1338): Docstring nit (non-blocking): says `vm_index: Integer ID of the VM`, but the actual dict key consumed at line 1360 is `vm_config["vm_id"]`, and the example at line 1350 also shows `"vm_id"`. The key was intentionally kept as `vm_id` (serialized-key scope rule); the docstring should say `vm_id` to match.

**`src/aleph/vm/supervisor/local.py`** (line 1077): Nit (non-blocking): `f"Found running VM {vm_id} with vm_id={execution.vm.vm_index}"` reuses the token `vm_id` for both the string identity and the numeric index in one log line, which is confusing post-rename. Suggest `with vm_index={execution.vm.vm_index}`.

foxpatch-aleph (changes requested):
The rename is mechanically sound and consistent across the supervisor, network, hypervisor, and test layers. The serialized `Configuration.vm_id`(int)/`vm_hash`(str) fields are correctly preserved, and boundary conversions in `qemu_build.py` and `controllers/__main__.py` correctly map `vm_index`→`vm_id` and `vm_id`→`vm_hash`. However, there is one genuine bug: `AlephFirecrackerExecutable.configure()` constructs `Configuration(vm_index=self.vm_index, ...)` but `Configuration` is a Pydantic BaseModel with no `vm_index` field (only `vm_id: int`). Pydantic v2 with default `extra='ignore'` silently drops the kwarg, leaving the required `vm_id` field unset, causing a `ValidationError`. This path is reachable via the firecracker spec create path (`pool.py:463` calls `execution.start()` with default `write_config=True` → `configure()`). Compare with `qemu_build.py:170` which correctly uses `vm_id=vm_index`. There is also a minor log-label inconsistency in `local.py:1077`.

**`src/aleph/vm/supervisor/controllers/firecracker/executable.py`** (line 248): Bug: `Configuration(vm_index=self.vm_index, ...)` passes `vm_index=` but the `Configuration` Pydantic model (supervisor_interface/configuration.py:90-95) has only `vm_id: int` and `vm_hash: str` — no `vm_index` field. With Pydantic v2 default `extra='ignore'`, `vm_index` is silently dropped and the required `vm_id` field is never set, raising `ValidationError: Field required [type=missing, input_value=..., input_type=dict]`. This is reachable via the firecracker spec path: `pool.py:463` calls `execution.start()` with the default `write_config=True`, which calls `self.vm.configure()` here. Fix: use `vm_id=self.vm_index` (mapping the internal numeric index to the serialized `vm_id` field), exactly as `qemu_build.py:170` does (`vm_id=vm_index`).

**`src/aleph/vm/supervisor/local.py`** (line 1077): Minor: the log label says `vm_id=` but the value logged is `execution.vm.vm_index` (the numeric index). For consistency with the rename, the label should be `vm_index=`: `f"Found running VM {vm_id} with vm_index={execution.vm.vm_index}"`.

foxpatch-aleph (approved):
A well-executed, behavior-neutral rename that splits the supervisor identity into string `vm_id` (Aleph item hash) and numeric `vm_index` (local pool id), moves shared types to neutral modules to break the agent→controllers import, and consolidates view exception handling under `SupervisorError`. The rename is mechanically consistent across models, pool, network, controllers, and tests. The agent-side downloaders correctly raise the contract error vocabulary, and the import-linter contract is tightened. No logic bugs, race conditions, or security issues found. A few minor naming inconsistencies in dict keys and docstrings are non-blocking.

**`src/aleph/vm/supervisor/local.py`** (line 1072): The dict key `"vm_id"` stores `execution.vm.vm_index` (a numeric index). The key name is misleading since `vm_id` now means the string identity elsewhere. Consider renaming the key to `"vm_index"` for consistency with the rest of the rename, or at minimum add a comment. The consumer in `recreate_network_for_vms` already reads it into a local named `vm_index`, so the key name is purely cosmetic but confusing.

**`src/aleph/vm/network/firewall.py`** (line 1350): Docstring example still uses `"vm_id": 1` as the dict key. The Args section says `vm_index: Integer ID of the VM` but the actual dict key in `local.py` is `"vm_id"`. These should all agree — either the key should be `"vm_index"` or the docstring should match the actual key name.

**`src/aleph/vm/supervisor/local.py`** (line 1077): Debug log says `vm_id={execution.vm.vm_index}` — the label says `vm_id` but the value is the numeric index. Should say `vm_index={execution.vm.vm_index}` for clarity.

**`src/aleph/vm/supervisor/controllers/firecracker/program.py`** (line 40): There are now two `FileTooLargeError` classes: this one (`Exception` subclass, kept for backward compat) and `supervisor_interface.errors.FileTooLargeError` (`SupervisorError` subclass). `read_input_data` in `program_config.py` raises the contract version. The one here is no longer raised by any code path in the PR — it's only kept for tests that import from this module. Consider whether this dead class should be removed or at least commented as a compat shim to avoid confusion about which `FileTooLargeError` is in play.

**`src/aleph/vm/agent/vm/downloader.py`** (line 132): Class-level annotations like `kernel_image_path: Path`, `rootfs_path: Path`, `code_path: Path`, `data_path: Path | None` have no defaults and are not set in `__init__`. Accessing them before the corresponding `download_*` method runs raises `AttributeError`. This matches the pre-existing pattern in `AlephProgramResources` so it's behavior-neutral, but a `# set by download_*` comment or `None` default would make the contract clearer.

foxpatch-aleph (changes requested):
A well-executed rename overall, but there is one runtime bug: `executable.py:248` passes `vm_index=self.vm_index` to `Configuration(...)`, while the `Configuration` pydantic model expects the required field `vm_id: int` (not `vm_index`). Pydantic v2 silently ignores the unknown `vm_index` kwarg and raises `ValidationError` for the missing `vm_id`. The sibling call sites in `qemu_build.py:170,274` correctly use `vm_id=vm_index`. This will crash persistent Firecracker VMs on the `start(write_config=True)` path. Fix: change `vm_index=self.vm_index` to `vm_id=self.vm_index` at executable.py:248.

**`src/aleph/vm/supervisor/controllers/firecracker/executable.py`** (line 248): Bug: `Configuration(vm_index=self.vm_index, ...)` passes the wrong keyword. The `Configuration` pydantic model (configuration.py:91) has field `vm_id: int`, not `vm_index`. Pydantic v2 default `extra="ignore"` drops `vm_index` and raises `ValidationError` for the missing required `vm_id`. Compare with the correct usage in `qemu_build.py:170`: `Configuration(vm_id=vm_index, ...)`. Fix: change to `vm_id=self.vm_index`.

foxpatch-aleph (approved):
A clean, behavior-neutral rename that correctly disambiguates the string identity (vm_hash -> vm_id: VmId) from the numeric local id (vm_id: int -> vm_index) on the supervisor objects. The critical serialized boundary (Configuration.vm_id: int / vm_hash: str) is preserved at every conversion point (executable configure, qemu_build, pool reattach, controller __main__). All remaining vm_hash references are verified to be in the deliberately-kept categories (serialized fields, controller ItemHash attribute, migration job fields, log format keys). The new regression test directly guards the most dangerous boundary (pydantic extra='ignore' silent-drop of vm_index= into Configuration) on a root-only path. Test changes are purely identifier-retargeting. No bugs or security issues found.

**`src/aleph/vm/pool.py`** (line 616): Minor comment nit: the comment says "Track claimed vm_ids to detect duplicates" but the set now tracks vm_index values (line 617 correctly says "vm_index"). Consider updating "claimed vm_ids" -> "claimed vm_indices" for consistency.

PR #992, opened 2026-06-20
refactor(boundary): remove dead controller error branches and classes

odesenfans:
Follow-up cleanup after the agent/supervisor boundary split (PR-1..PR-3). Stacked on **#990** (`od/supervisor-vm-id-rename-impl`) — retarget to `dev` once #990 merges.

With the download phase running agent-side and raising the contract vocabulary (`supervisor_interface.errors`), several controller-side exception paths are now unreachable. This removes the verified-dead ones.

### Removed
- **`run.py`** — instance-path `except MicroVMFailedInitError`. Instances are QEMU, not Firecracker microVMs, so that path never raises a MicroVM init error (local or contract). The program path still catches both forms; the import stays (used there).
- **`error_mapping.py`** — `translate_exception` branches for the controller `ResourceDownloadError` and `VmSetupError`, plus their now-unused contract imports. Every raiser of those errors is agent-side and raises the *contract* class; no supervisor-side op (the only thing `translating_errors` wraps) raises the controller class.
- **`executable.py`** — the orphaned `ResourceDownloadError` and `VmSetupError` class defs (never raised, no importers) and the now-unused `ClientResponseError` import.

### Deliberately NOT touched (needs a decision)
The controller **`program.FileTooLargeError`** class and its `error_mapping` branch are kept, because `tests/supervisor/test_supervisor_errors.py::test_translate_file_too_large` pins that controller→contract mapping. Retiring it requires editing that test, which I won't do unilaterally under the "don't denature tests" rule. Options for a follow-up:
- **(X)** retire it and change *only* the test's import to the contract `FileTooLargeError` (assertions unchanged; the test then verifies the `SupervisorError` pass-through, which is real and still passes), or
- **(Y)** leave it permanently as the single documented exception to the consolidation.

### Verification
- import-linter: **4 kept / 0 broken**
- mypy: **baseline unchanged (43 errors / 13 files)**
- full supervisor suite: **936 passed** (only the 3 env-only pyroute2 `test_interfaces` failures)
- ruff format + isort: clean

### DIFF
--- a/src/aleph/vm/agent/run.py
+++ b/src/aleph/vm/agent/run.py
@@ -353,9 +353,6 @@ async def create_vm_execution_or_raise_http_error(
     except VmSetupError as error:
         logger.exception(error)
         raise HTTPInternalServerError(reason="Error during vm initialisation") from error
-    except MicroVMFailedInitError as error:
-        logger.exception(error)
-        raise HTTPInternalServerError(reason="Error during runtime initialisation") from error
     except HostNotFoundError as error:
         logger.exception(error)
         raise HTTPInternalServerError(reason="Host did not respond to ping") from error
--- a/src/aleph/vm/supervisor/controllers/firecracker/executable.py
+++ b/src/aleph/vm/supervisor/controllers/firecracker/executable.py
@@ -10,7 +10,6 @@ from os.path import exists
 from pathlib import Path
 from typing import Generic, TypeVar
 
-from aiohttp import ClientResponseError
 from aleph_message.models import ExecutableContent, ItemHash
 
 from aleph.vm.conf import settings
@@ -49,19 +48,6 @@ except RuntimeError as error:
         raise error
 
 
-class ResourceDownloadError(ClientResponseError):
-    """An error occurred while downloading a VM resource file"""
-
-    def __init__(self, error: ClientResponseError):
-        super().__init__(
-            request_info=error.request_info,
-            history=error.history,
-            status=error.status,
-            message=error.message,
-            headers=error.headers,
-        )
-
-
 @dataclass
 class BaseConfiguration:
     vm_hash: ItemHash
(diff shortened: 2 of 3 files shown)


foxpatch-aleph (approved):
The PR correctly removes verified-dead exception branches and classes. The in-process supervisor wraps all operations in translating_errors(), which converts backend exceptions (MicroVMFailedInitError, controller ResourceDownloadError/VmSetupError) to contract SupervisorError subclasses before they reach the agent, making the removed agent-side catch branches unreachable. The controller class definitions in executable.py have no raisers and no importers (tests and agent code import the contract classes from supervisor_interface.errors). The FileTooLargeError controller class and its error_mapping branch are correctly retained, pinned by test_translate_file_too_large. The import of MicroVMFailedInitError in run.py is correctly kept as it is still used in _raise_http_for_program_error. No security or correctness issues found.

**`src/aleph/vm/agent/run.py`** (line 401): Observation (not blocking, outside this PR's scope): the MicroVMFailedInitError part of this tuple catch in _raise_http_for_program_error is also likely dead — the supervisor's translating_errors() converts it to MicroVMInitError before it reaches the agent, same reasoning as the branch removed from create_vm_execution_or_raise_http_error. The supervisor_errors.MicroVMInitError part is still reachable. A future cleanup could narrow this to just supervisor_errors.MicroVMInitError.

foxpatch-aleph (approved):
Clean, well-justified dead-code removal. All three changes were verified: (1) the removed MicroVMFailedInitError except branch in run.py was unreachable because instances are QEMU-only (build_create_vm_spec rejects non-QEMU), and MicroVMFailedInitError is a Firecracker-specific error; the import is correctly retained for the program path. (2) The removed ResourceDownloadError and VmSetupError classes in executable.py had no remaining importers — all code uses the contract equivalents from supervisor_interface.errors. (3) The removed translate_exception branches in error_mapping.py were dead because translating_errors only wraps supervisor-side operations that never raise the controller versions. No tests referenced the removed controller classes, and the FileTooLargeError mapping was correctly retained per the PR description. No unused imports remain.

PR #993, opened 2026-06-20
fix(agent): domains->HAProxy sync works in gRPC split mode

odesenfans:
## Split-mode gap 1 of 4: domains → HAProxy sync

**Problem.** In split (two-process gRPC) mode the agent has no in-process `VmPool`. `_handle_domains_aggregate` short-circuited on `pool is None`, so a `domains` aggregate change never refreshed the HAProxy domain map. Custom-domain instances stop being routed after a mapping change. (In-process mode was unaffected — and, as it turns out, the only mode our integration tests have ever exercised.)

**Fix.** Source the owner's instances from the agent registry + `supervisor.get_vm()` instead of `pool.executions`, mirroring the already-split-safe `_handle_port_forwarding_aggregate`:
- registry record's `message.address == aggregate.address` (owner match, as before),
- `supervisor.get_vm(...)` succeeding stands in for "present in the pool",
- `VmInfo.backend is Backend.QEMU` is the wire equivalent of the old `execution.is_instance`.

Works identically in-process and over gRPC. No behavior change in-process.

**Scope note.** `pool` is now unused by `watch_for_messages`. Removing that dead threading (plus the unused `Reactor` pool param) is the separate **Reactor pool-param cleanup PR**, kept out of here to stay surgical.

**Tests.** The four `_handle_domains_aggregate` tests are rewritten to drive `supervisor.get_vm` (no pool), and a new case covers an owner-matching registry record whose VM the supervisor doesn't know. `test_tasks_registry_reads.py`: 11 passed.

### DIFF
--- a/src/aleph/vm/agent/tasks.py
+++ b/src/aleph/vm/agent/tasks.py
@@ -39,7 +39,13 @@ from aleph.vm.conf import settings
 from aleph.vm.pool import VmPool
 from aleph.vm.supervisor_interface.abc import Supervisor
 from aleph.vm.supervisor_interface.errors import VmNotFoundError
-from aleph.vm.supervisor_interface.types import ConfidentialMode, VmId, VmInfo, VmStatus
+from aleph.vm.supervisor_interface.types import (
+    Backend,
+    ConfidentialMode,
+    VmId,
+    VmInfo,
+    VmStatus,
+)
 from aleph.vm.utils import create_task_log_exceptions
 
 # Terminal statuses that confirm a message is no longer valid.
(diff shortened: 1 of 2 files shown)


foxpatch-aleph (approved):
The fix correctly replaces the in-process pool enumeration with a registry + supervisor.get_vm() lookup, mirroring the already-split-safe _handle_port_forwarding_aggregate. The is_instance to info.backend is Backend.QEMU translation is exact (models.py:300). The execution.vm truthiness check is replaced by supervisor.get_vm() succeeding, which is semantically equivalent. VmNotFoundError is properly caught for stale registry entries. Tests are rewritten to exercise the split-mode path without a pool, and a new test covers owner-matching records absent from the supervisor. The deferred pool-param cleanup is appropriately scoped to a separate PR.

**`src/aleph/vm/agent/tasks.py`** (line 260): Minor pre-existing inconsistency: _handle_port_forwarding_aggregate (line 223) guards with isinstance(record.message, InstanceContent) before accessing record.message.address, while this handler does not. Both ProgramContent and InstanceContent have an address field so this is safe, but the inconsistency predates this PR and could be aligned in a future cleanup.

PR #994, opened 2026-06-20
fix(agent): graceful drain works in gRPC split mode

odesenfans:
## Split-mode gap 2 of 4: graceful drain

**Problem.** `drain_middleware` rejected new VM requests based on `pool.is_draining`, and `drain_in_flight_requests` (the `on_shutdown` hook) called `pool.drain()`. Both reach the in-process pool, which is `None` in split mode — so during a shutdown/upgrade the agent kept accepting new VM-creation requests instead of shedding them.

**Fix.** Make the draining state **agent-owned** (`app["draining"]`):
- `drain_middleware` rejects new VM requests (`/vm/*` + hostname routing) based on the flag, in both modes. The pool is no longer consulted.
- `drain_in_flight_requests` flips the flag, then — in-process only — drains the embedded pool to wait for in-flight ephemeral runs. In split mode the daemon owns the pool and keeps VMs running across its own restart (its shutdown just stops the gRPC server), so the agent flips the flag and lets aiohttp finish in-flight handlers.

**No in-process behavior change:** the flag is set on the same `on_shutdown` hook that already called `pool.drain()`, so rejection timing is identical.

**Tests.** Middleware tests now drive the flag with `pool=None` (split-realistic, proving it works with no embedded pool); added `drain_in_flight_requests` tests for both modes (split flips the flag without a pool; in-process flips the flag *and* awaits `pool.drain()`). `test_drain.py`: 14 passed.

### DIFF
--- a/src/aleph/vm/agent/supervisor.py
+++ b/src/aleph/vm/agent/supervisor.py
@@ -95,14 +95,17 @@ logger = logging.getLogger(__name__)
 
 @web.middleware
 async def drain_middleware(request, handler) -> web.Response:
-    """Reject new VM execution requests while the pool is draining.
+    """Reject new VM execution requests while the agent is draining.
 
     Only blocks the paths that trigger on-demand VM creation
     (/vm/* and hostname-based routing). Status, control, and about
     endpoints remain accessible for monitoring and operations.
+
+    The draining flag is agent-owned (``app["draining"]``), so it works in split
+    mode where the in-process pool is absent: the agent fronts every request
+    regardless of which process owns the VMs.
     """
-    pool: VmPool | None = request.app.get("_engine_pool")
-    if pool and getattr(pool, "is_draining", False):
+    if request.app.get("draining"):
         path = request.path
         is_vm_request = path.startswith("/vm/") or (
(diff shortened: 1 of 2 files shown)


foxpatch-aleph (approved):
Clean, well-scoped fix that correctly moves the drain state from the in-process pool (absent in split mode) to an agent-owned app flag. The middleware now rejects new VM requests in both modes, and drain_in_flight_requests flips the flag before optionally draining the embedded pool. Behavioral equivalence in-process is preserved — the flag is set on the same on_shutdown hook that previously called pool.drain(). Tests are improved: middleware tests now run with pool=None (split-realistic), and new shutdown-hook tests cover both modes. No correctness, security, or race condition issues found.

**`tests/supervisor/test_drain.py`** (line 115): The TestDrainShutdownHook tests use plain dicts instead of a real web.Application. This works today since drain_in_flight_requests only uses __setitem__ and .get(), but it's slightly fragile — a future change to the function (e.g., accessing app.state or a web.Application-specific API) would silently break these tests at runtime rather than at import time. Consider using setup_webapp(pool=None) for consistency with the middleware tests, though this is a non-blocking nit.

PR #995, opened 2026-06-20
fix(agent): migration reaper works in split mode (agent-side, via list_vms)

odesenfans:
## Split-mode gap: orphan migration-file reaper (corrected — agent-side)

> Redone after review. The first version moved the reaper to the daemon; that was wrong. Cold migration is **agent-side**, and a daemon-side reaper is actively unsafe — see below.

**Problem.** The orphan-migration-file reaper ran as an agent `on_startup` hook but read `pool.executions`, which is `None` for the agent in split mode, so it no-op'd and orphan `.part` / `.export.qcow2` files accumulated.

**Why it stays agent-side.** Cold migration is entirely agent-owned: `migration/runner.py` (`run_import`/`run_export`) is agent code (it imports `agent.translate`), launched as an agent background task from `agent/views/migration.py`, and it does all the file work — the `.part` download (`helpers.download_disk_from_source`), the `.export.qcow2` export, and the in-flight `import_jobs` map. It stages the disk, *then* calls `supervisor.create_vm(spec)`. The supervisor never reads or writes these files.

**Why the daemon version was unsafe.** During an active import, `.part` exists but the VM does not yet (CreateVm happens *after* the download). In split mode the daemon is an independent process — a daemon restart mid-import would see `.part` + no execution and `rmtree` the agent's **live** import, since the daemon has no view of `import_jobs`.

**Fix.** Keep the reaper agent-side; source the live-VM set from `supervisor.list_vms()` over the ABC instead of `pool.executions`. This mirrors `_handle_domains_aggregate` (#993) and the `_vm_exists` check the migration runner already uses. The set only guards against deleting a running VM's volume dir. Run at agent startup, the agent's own `import_jobs` is empty, so any `.part` is genuinely an aborted prior run. The daemon is untouched.

**Tests.** `reap_orphan_migration_files` now takes the id set (tests driven by a set, plus a new "live VM's `.part` dir is kept" case), and a new test covers the agent hook sourcing ids from `list_vms`. 5 passed.

### DIFF
--- a/src/aleph/vm/agent/supervisor.py
+++ b/src/aleph/vm/agent/supervisor.py
@@ -403,10 +403,16 @@ async def stop_program_client(app: web.Application) -> None:
 
 
 async def _run_migration_reaper(app: web.Application) -> None:
-    """on_startup hook: clean up orphan migration files left from a prior supervisor run."""
-    pool = app.get("_engine_pool")
-    if pool is not None:
-        await reap_orphan_migration_files(pool)
+    """on_startup hook: clean up orphan migration files from a prior agent run.
+
+    Cold migration staging is agent-owned, so this runs agent-side in both modes.
+    The live-VM set comes from the supervisor over the ABC (works in split mode,
+    where there is no in-process pool); it only guards against deleting a running
+    VM's volume directory.
+    """
+    supervisor = app["supervisor"]
+    known_vm_ids = {str(info.vm_id) for info in await supervisor.list_vms()}
+    await reap_orphan_migration_files(known_vm_ids)
 
 
 async def _rehydrate_vm_registry(app: web.Application) -> None:
--- a/src/aleph/vm/migration/reaper.py
+++ b/src/aleph/vm/migration/reaper.py
@@ -8,22 +8,27 @@ from aleph.vm.conf import settings
 logger = logging.getLogger(__name__)
 
 
-async def reap_orphan_migration_files(pool) -> None:
-    """Reap orphan export and partial-import files left behind by a prior supervisor run.
+async def reap_orphan_migration_files(known_vm_ids: set[str]) -> None:
+    """Reap orphan export and partial-import files left behind by a prior agent run.
+
+    Cold migration is agent-side: the agent stages the disk (the ``.part``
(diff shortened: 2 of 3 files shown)


foxpatch-aleph (approved):
Clean, well-motivated refactor that moves the orphan-migration-file reaper from an agent on_startup hook (which no-ops in split mode) to the daemon's pool startup sequence where pool.executions is available. The _prepare_pool helper is correctly factored out for testability, error handling is consistent with the existing reattach policy (log and continue), and ordering is correct (reaper runs after reattach populates pool.executions). Tests cover the happy path plus both failure modes (reaper failure and reattach failure), asserting the daemon still comes up and the reaper still runs after a failed reattach. No correctness, security, or quality issues found.

**`src/aleph/vm/supervisor/daemon.py`** (line 34): Minor: `_prepare_pool` lacks a return type annotation. Since `VmPool` is locally imported (to avoid importing the controller stack before `settings.setup()`), annotating with `-> "VmPool"` under `from __future__ import annotations` would be safe and consistent with the file's other annotations. Non-blocking.

foxpatch-aleph (approved):
Clean, well-justified fix. The reaper correctly stays agent-side (cold migration staging is agent-owned, and a daemon-side reaper would be unsafe — it has no view of import_jobs and could rmtree a live import). Sourcing the live-VM guard set from supervisor.list_vms() over the ABC mirrors the established pattern in haproxy_sync.py and the _vm_exists check in the migration runner. The set is built with str(info.vm_id), matching the str(job.vm_hash) directory naming in runner.py, so the entry.name in known_vm_ids comparison is sound. Running at agent startup is safe because import_jobs is a module-level dict reset on restart, so any .part files are genuinely aborted prior runs. Test coverage is solid: export deletion, orphan .part removal, live-VM .part preservation (new), complete-orphan retention, and the hook wiring. No security, correctness, or race-condition issues.

**`tests/migration/test_reaper.py`** (line 89): Minor nit: passing a plain dict {"supervisor": supervisor} as the `app` argument works because the hook only touches app["supervisor"], but it diverges from the real web.Application shape. Not blocking — the test is focused and correct for what it asserts.

PR #996, opened 2026-06-20
refactor(agent): drop dead pool threading from the message path

odesenfans:
## Split-mode gap 4 of 4: dead pool threading on the message path

**Problem.** `Reactor` accepted and stored a `pool` it never used. The only thing that fed it was `start_watch_for_messages_task`, which fetched the embedded pool and threaded it into both `Reactor` and `watch_for_messages`. The pool is `None` in split mode and unused in in-process mode — pure dead weight that also kept an `agent → pool` import edge alive.

**Fix.** Remove the dead thread end to end:
- `Reactor`: drop the `pool` param and field.
- `watch_for_messages`: drop the `pool` param.
- `start_watch_for_messages_task`: drop the `_engine_pool` fetch.
- Remove the now-unused `VmPool` imports in `reactor.py` and `tasks.py`.

No behavior change (the value was never read). Drops one `agent → pool` import edge.

**Stacking.** Based on **#993** (`od/split-domains-haproxy`) — that PR is what frees `watch_for_messages` from the pool (it was the last real consumer, via the domains handler). Removing the thread off `dev` directly would orphan the builder line. **Retarget to `dev` once #993 merges.**

No new tests (dead-code removal); whole suite still collects (990 tests), `test_tasks_registry_reads.py` 11 passed. Verified: import-linter 4/0, mypy baseline unchanged (43/13), ruff+isort clean.

### DIFF
--- a/src/aleph/vm/agent/reactor.py
+++ b/src/aleph/vm/agent/reactor.py
@@ -8,7 +8,6 @@ from aleph.vm.agent.expiry import ExpiryManager
 from aleph.vm.agent.update_watcher import UpdateWatcher
 from aleph.vm.agent.vm.program_client import ProgramGuestClient
 from aleph.vm.agent.vm_registry import AgentVmRegistry
-from aleph.vm.pool import VmPool
 from aleph.vm.supervisor_interface.abc import Supervisor
 from aleph.vm.utils import create_task_log_exceptions
 
@@ -45,7 +44,6 @@ def subscription_matches(subscription: Subscription, message: AlephMessage) -> b
 
 class Reactor:
     pubsub: PubSub
-    pool: VmPool
     supervisor: Supervisor
     expiry: ExpiryManager
     update_watcher: UpdateWatcher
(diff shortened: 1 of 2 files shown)


foxpatch-aleph (approved):
Clean, correct dead-code removal. Verified that `Reactor.pool` was never read anywhere in the codebase, that `Reactor` and `watch_for_messages` each have exactly one call site (both in `tasks.py`) and both match the new signatures, and that the `_engine_pool` app key is still set and consumed by other code in `supervisor.py` and tests — so removing only the `tasks.py` fetch is safe. No behavior change, no new tests needed, no security implications. The only stale references to the old signature are in historical planning docs under `docs/plans/`, which are out of scope.

foxpatch-aleph (approved):
Clean dead-code removal. The `pool` parameter was stored but never read in `Reactor`, and `watch_for_messages` never used its `pool` argument. All call sites are updated consistently, no tests reference the old signatures, and the remaining `_engine_pool` reads in `supervisor.py` are legitimate process-lifecycle paths that are correctly out of scope. No behavior change, no security concerns.

**`tests/supervisor/test_agent_pool_free.py`** (line 35): Consider adding `tasks` and `reactor` to `_MODULES` now that they are also pool-free. This would guard against regressions reintroducing a `VmPool` import or `_engine_pool` reference in those modules.

PR #998, opened 2026-06-22
fix(restore): return 400 on invalid restore image (regression from #948)

odesenfans:
## Summary

Two small fixes uncovered while gap-analysing `dev` against the 1.13 release on `main`.

### 1. Restore of an invalid image returns 500 again (regression of #948)
A malformed restore upload (e.g. a tar instead of a QCOW2) returns **HTTP 500** instead of 400.

`verify_qemu_disk()` raises `subprocess.CalledProcessError` on a bad image. Inside `LocalSupervisor.restore_from_image()` that runs under `translating_errors()`, which has no `CalledProcessError` case and falls through to `InternalSupervisorError` → the agent's generic handler → 500. The agent already maps `InvalidBackendError` → 400, so the fix is to catch the `CalledProcessError` at the validation call and raise `InvalidBackendError`.

Tightly scoped to the validation call rather than a broad `translate_exception` rule, so genuine internal subprocess failures still surface as 500. The oversized-disk path already raised `InvalidBackendError`; only the format-validation path regressed when disk validation moved into the supervisor during the split.

Adds `test_restore_from_image_rejects_invalid_qcow2` (a clone of the existing oversized-disk test).

### 2. Stale `setup_webapp` docstring
The docstring claimed backup/restore/confidential/migration/network/GPU/persistent-program endpoints "respond 501" in split mode. That predates Phase 1 — those endpoints now route through the `Supervisor` interface and work identically in split mode. Comment-only.

## Test plan
- [x] `ruff format` / `isort` clean
- [ ] CI runs the supervisor suite (couldn't run locally: env needs `libdbus-1-dev`/`libsystemd-dev`)

### DIFF
--- a/src/aleph/vm/agent/supervisor.py
+++ b/src/aleph/vm/agent/supervisor.py
@@ -214,10 +214,10 @@ def setup_webapp(pool: VmPool | None):
 
     VmPool is None in two cases: tests that won't use it, and **split mode**
     (`ALEPH_VM_SUPERVISOR_GRPC_SOCKET` set), where the supervisor daemon owns
-    the pool and the agent reaches it over gRPC. In split mode the endpoints
-    that still require the in-process pool (backups, restore, confidential,
-    migration, network recreation, GPU reservation, persistent programs)
-    respond 501.
+    the pool and the agent reaches it over gRPC. Request handlers never touch
+    the pool directly: every VM operation (backups, restore, confidential,
+    migration, network recreation, GPU reservation, persistent programs) goes
+    through the `Supervisor` interface, so it works identically in split mode.
     """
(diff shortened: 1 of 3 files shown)


foxpatch-aleph (approved):
Two well-scoped fixes. The core fix correctly catches `subprocess.CalledProcessError` at the `verify_qemu_disk` call site and re-raises as `InvalidBackendError`, which the agent maps to HTTP 400 (confirmed at `operator.py:974-978`). The catch is tightly scoped to the validation call only, so genuine internal subprocess failures still surface as 500. The regression test mirrors the existing oversized-disk test pattern and verifies both the error type/code and that the rootfs is left untouched. The docstring update is accurate and comment-only. No security issues — the error message exposes only the filename, not the full path.

PR #999, opened 2026-06-23
refactor(supervisor): drop vestigial message-built (spec=None) guards

odesenfans:
## What

`VmExecution.spec` is a required, non-optional field (`spec: CreateVmSpec`), and every construction path now goes through `VmExecution.from_spec` — `create_vm_from_spec` for new VMs and the restart-recovery path via `spec_from_controller_configuration`. No code path produces a spec-less execution, so `vm_spec` is never `None` in production. The remaining `spec is None` checks are leftovers from the now-removed message-built construction path.

This removes the unreachable defensive branches and the tests that only exercised them by faking the impossible `vm_spec=None` state via mocks.

## Changes

- **`LocalSupervisor.get_vm_spec`** — drop the `spec is None` → `NotImplementedSupervisorError` guard; return the held spec directly.
- **`LocalSupervisor.reboot_vm`** — drop the `if spec is not None:` wrapper and the "message-built ephemeral, stop only" fallback. Spec-built ephemeral VMs already reboot via `create_vm_from_spec`; that path stays and is covered by `test_reboot_ephemeral_spec_vm_recreates_from_held_spec`.
- **`VmPool._require_same_spec`** — fix the stale docstring referencing a `vm_spec=None` collision that can no longer occur.
- Remove three tests that fabricated `vm_spec=None` through mocks: the reboot stop-only test, the `get_vm_spec` NotImplemented test, and the pool-collision test.

## Why it's safe

Behaviour-neutral: the type system already guarantees `spec` is present, so the removed branches were dead. The reachable behaviour stays fully covered by the surviving tests. `NotImplementedSupervisorError` is still used by the deliberate ephemeral-lifecycle guards (stop/start/restore), so the import remains.

## Verification

- `ruff==0.4.6` format: all touched files already formatted; lint introduces no new findings (only deletes code).
- `py_compile` clean on all four files.
- Full pytest deferred to CI: the local env can't build `dbus-python`/`systemd-python` (no `libdbus-1-dev`, sudo unavailable).

### DIFF
--- a/src/aleph/vm/pool.py
+++ b/src/aleph/vm/pool.py
@@ -476,8 +476,7 @@ class VmPool:
     @staticmethod
     def _require_same_spec(current_execution: VmExecution, spec: CreateVmSpec) -> None:
         """CreateVm idempotency: a retry with the identical spec returns the
-        live VM; a different spec for a live vm_index is a conflict (also covers
-        colliding with a message-built execution, whose vm_spec is None)."""
+        live VM; a different spec for a live vm_index is a conflict."""
         if current_execution.vm_spec != spec:
             msg = f"VM {spec.vm_id} already exists with a different spec"
             raise VmAlreadyExistsError(msg)
--- a/src/aleph/vm/supervisor/local.py
+++ b/src/aleph/vm/supervisor/local.py
@@ -460,11 +460,7 @@ class LocalSupervisor(Supervisor):
     async def get_vm_spec(self, vm_id: VmId) -> CreateVmSpec:
         with translating_errors():
             execution = self._require(vm_id)
-            spec = execution.vm_spec
-            if spec is None:
-                msg = f"VM {vm_id} was created outside the spec path (message-built); no spec is held"
-                raise NotImplementedSupervisorError(msg)
-            return spec
+            return execution.vm_spec
 
     async def delete_vm(self, vm_id: VmId, wipe: bool = False) -> None:
         with translating_errors():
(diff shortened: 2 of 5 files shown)


foxpatch-aleph (approved):
Verified that VmExecution.vm_spec is a non-optional property (returns self.spec, set from required __init__ parameter) and all construction paths go through VmExecution.from_spec. The removed spec-is-None guards were genuinely dead code. The three deleted tests fabricated an impossible vm_spec=None state via mocks and tested unreachable branches. Surviving tests fully cover the reachable behavior. NotImplementedSupervisorError import remains needed for the ephemeral lifecycle guards. Clean, safe dead-code removal.

**`tests/supervisor/test_supervisor_inprocess_query.py`** (line 55): Pre-existing and out of scope, but noting for awareness: the `make_execution` helper here still sets `vm_spec=None`. It's harmless since this file only tests query/list endpoints (no calls to `get_vm_spec` or `reboot_vm`), but it could confuse future readers who expect the mock to reflect the real type contract. Consider updating to a real spec in a follow-up.

odesenfans:
Folded in the review nit (98bc7060): dropped the unused `vm_spec=None` from the `make_execution` helper in `test_supervisor_inprocess_query.py`. It was another leftover of the message-built fiction — no test in that file reads `vm_spec`, so I removed the field rather than fabricate a real spec.

foxpatch-aleph (approved):
The PR correctly removes dead defensive branches that guarded against vm_spec=None, a state that can no longer occur since VmExecution.vm_spec is a read-only property returning self.spec (typed as non-optional CreateVmSpec), and all construction paths go through VmExecution.from_spec. I verified the type definition, all construction paths (create_vm_from_spec, restart-recovery via spec_from_controller_configuration), the absence of any remaining vm_spec is None checks in src/, and that NotImplementedSupervisorError is still used by four other guards. The three removed tests fabricated an impossible state through mocks and provided no real coverage. The surviving tests fully cover the reachable behavior. The docstring update in _require_same_spec is accurate.

PR #1001, opened 2026-06-23
fix(supervisor): isolate per-VM reattach failures and protect live VMs

odesenfans:
## Problem

`load_persistent_executions` awaited each VM's restore with **no try/except** (`pool.py:642`), so a single failure aborted the entire reattach loop. The realistic trigger is a live **confidential (SEV)** instance: `spec_from_controller_configuration` rejects `QemuConfidentialVMConfiguration` with `InvalidBackendError`, and any `prepare()`/network error raises too.

Consequences on a node hosting even one such VM:
- **Split mode**: the daemon survives (its caller wraps `load_persistent_executions` in `except Exception`), but every VM after the failure in that pass is left un-reattached.
- **In-process mode (the default)**: the caller at `agent/supervisor.py:489` only catches `OSError`, so the exception crashes agent startup — a crash loop, since the live controller keeps the config "active".

## Fix

Wrap the per-VM restore in `try/except`: one VM's failure is logged (`logger.exception`) and skipped while the rest reattach.

The subtle part: a skipped VM must **not** be treated as a dead controller. Its controller is still running, so stopping the service or deleting `controller.json` would **destroy a live instance** — worse than the original crash. `_cleanup_orphan_resources()` runs right after the loop and does exactly that to any VM absent from `self.executions` (stops+disables the service, deletes the config, tears down tap/nft). So the failed VM's `vm_index`/`hash` are now passed to `_cleanup_orphan_resources` as **protected** sets, unioned into the "active" sets the sweep uses. The live VM and its on-disk artifacts are left fully intact; it's simply untracked in-memory for this cycle.

## Scope

This is the **robustness** half of the confidential-reattach problem — it stops one VM from taking down reattach for the whole node and stops the startup crash. Making confidential reattach actually *succeed* (so SEV instances are re-adopted, not just left running-but-untracked) is a separate follow-up.

## Tests

- `test_failed_reattach_does_not_abort_the_rest`: with two active configs where one restore raises, the healthy VM still reattaches, the failed VM is **not** killed as a dead controller, and its index/hash are handed to cleanup as protected.
- `test_cleanup_orphan_resources_treats_protected_as_active`: protected ids/hashes are unioned into the active sets passed to the cleanup helpers.

## Verification

`ruff==0.4.6` format + lint clean on changed files (no new findings); `py_compile` clean. Full pytest deferred to CI (local env can't build `dbus-python`).

### DIFF
--- a/src/aleph/vm/pool.py
+++ b/src/aleph/vm/pool.py
@@ -617,6 +617,13 @@ class VmPool:
         # avoid two VMs sharing a tap interface.
         claimed_vm_ids: set[int] = set()
 
+        # A controller that is alive but whose in-memory reattach failed. Its
+        # vm_index/hash are protected from the orphan sweep below: the VM is
+        # still running, so stopping its service or deleting its config would
+        # destroy a live instance -- far worse than not tracking it this cycle.
+        failed_vm_ids: set[int] = set()
+        failed_vm_hashes: set[str] = set()
+
         for config in configs:
             vm_id = VmId(str(config.vm_hash))
             vm_index = config.vm_id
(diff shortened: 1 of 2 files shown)


foxpatch-aleph (approved):
Well-reasoned, correctly scoped fix. The try/except wrapper around per-VM reattach prevents one failure (e.g. a confidential VM with unsupported config) from aborting reattach for the entire node or crashing the in-process agent startup. The protection sets for failed VMs are correctly threaded through to _cleanup_orphan_resources to prevent the orphan sweep from destroying live VMs. The edge case where a duplicate vm_index config follows a failed reattach is handled correctly (the duplicate is treated as dead, not protected). Tests cover both the integration path and the cleanup internals. No correctness or security issues found.

**`tests/supervisor/test_supervisor_reattach.py`** (line 72): Style nit: `from unittest.mock import patch` could be hoisted to the top-level imports alongside `AsyncMock, MagicMock` on line 7 for consistency.

foxpatch-aleph (approved):
A well-scoped robustness fix that correctly isolates per-VM reattach failures and protects live VMs from the orphan sweep. The try/except wrapper, the failed_vm_ids/hashes protection sets, and their union into the active sets in _cleanup_orphan_resources are all logically sound. Comments clearly explain the critical safety invariant (don't destroy a live VM). Tests cover both the isolation behavior and the protected-set union. No security issues or regressions introduced.

**`tests/supervisor/test_supervisor_reattach.py`** (line 119): Minor style nit: using `seen.update(ids=ids)` and `seen.update(hashes=hashes)` with keyword-argument keys works but is slightly opaque. A more conventional approach like `seen["ids"] = ids` / `seen["hashes"] = hashes` would be clearer at a glance. Non-blocking.

**`tests/supervisor/test_supervisor_reattach.py`** (line 103): The assertion `dead.assert_not_called()` verifies `_handle_dead_controller` is never called, which is correct for this test since both controllers are active. Consider also testing the scenario where all VMs fail reattach — the cleanup should still receive all of them as protected and none should be swept. This would strengthen coverage of the edge case, but is not blocking.

PR #1003, opened 2026-06-23
feat(supervisor): reattach confidential (SEV) VMs after a restart

odesenfans:
## Problem

`spec_from_controller_configuration` rejected `QemuConfidentialVMConfiguration` with `InvalidBackendError`, so a **running confidential (SEV) instance could not be re-adopted** on supervisor restart. This is a regression vs 1.13, which rebuilt the execution from the stored aleph message. Combined with the unguarded restore loop (#1001), one live SEV VM could take down reattach for the whole node — and crash in-process startup (#1002).

This is the **feature** half of the confidential-reattach problem: #1001/#1002 make a failure survivable; this makes confidential reattach actually *succeed*.

## Fix

Reconstruct a `TeeConfig` from the on-disk confidential config, inverting `build_qemu_confidential_configuration`:

| Spec field | ← source | Notes |
|---|---|---|
| `firmware_path` | `ovmf_path` | The **only** field reattach needs — `AlephQemuConfidentialResources.from_spec` requires just the resolved OVMF blob |
| `policy` | `hex(sev_policy)` | Matches the message path's `hex(trusted_execution.policy)`; `create()` converts back via `int(.., 0)` |
| `session_dir` | `sev_session_file.parent` | Where the owner's already-uploaded certs live |
| `backend` | `TeeBackend.SEV` | |

## Why it's safe

Reattach re-adopts an **already-running, already-measured** VM. `_restore_running_execution_from_config` calls `prepare()` + `create(prepare=False)` + `start_guest_api()` — it **never** calls `start()`, so it does not re-boot the guest, rewrite the controller config, or enable/start the systemd unit (the controller is already active). No fresh session-cert upload is required.

The SAFETY-CRITICAL "a confidential VM must never boot under a weaker configuration" invariant is untouched: this path doesn't boot anything, and a later reboot-recreate still goes through `build_qemu_confidential_configuration`, which re-validates firmware + session files and hard-stops if they're missing.

A reattached confidential VM is correctly reported as **running**, not awaiting-init: `is_awaiting_confidential_init` is False once the VM is running, and the reattach loop only restores controllers systemd reports active.

## Tests

- `test_spec_from_config_reconstructs_confidential_tee`: a confidential config rebuilds a `TeeConfig` (SEV backend, `firmware_path == ovmf_path`, `int(policy, 0)` round-trips the SEV policy, `session_dir == sev_session_file.parent`) and the core fields reconstruct as in the plain path.
- `test_spec_from_config_rejects_non_qemu` still holds — a Firecracker `VMConfiguration` is still rejected.

## Verification

`ruff==0.4.6` format + lint clean on changed files (the one new finding, `UP038`, fixed by using the `X | Y` isinstance form; remaining findings are pre-existing in `build_cloud_init_drive`/`build_qemu_confidential_configuration`); `isort` clean; `py_compile` clean. Full pytest deferred to CI (local env can't build `dbus-python`).

## Stacking

Independent of #1001 and #1002 (different files), all branched off `dev`; mergeable in any order.

### DIFF
--- a/src/aleph/vm/supervisor/qemu_build.py
+++ b/src/aleph/vm/supervisor/qemu_build.py
@@ -29,12 +29,15 @@ from aleph.vm.supervisor_interface.errors import InvalidBackendError
 from aleph.vm.supervisor_interface.types import (
     Backend,
     CreateVmSpec,
+    DirectoryPath,
     DiskFormat,
     DiskRole,
     DiskSpec,
     GpuSpec,
     NetworkConfig,
     PciAddress,
+    TeeBackend,
+    TeeConfig,
     VmId,
 )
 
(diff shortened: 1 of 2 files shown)


foxpatch-aleph (approved):
The PR correctly reconstructs a TeeConfig from an on-disk QemuConfidentialVMConfiguration, allowing running SEV VMs to be re-adopted after a supervisor restart. The inversion of build_qemu_confidential_configuration is faithful: ovmf_path maps to firmware_path, hex(sev_policy) round-trips through int(.., 0), and sev_session_file.parent maps to session_dir. The isinstance union correctly accepts both QEMU config types, the shared disk/gpu/network reconstruction path only accesses fields present on both config classes, and the isinstance check ordering (union first, then confidential-specific branch) is logically sound. The safety analysis is accurate: _restore_running_execution_from_config never calls start(), so no re-boot occurs and the measured launch is preserved; the SAFETY-CRITICAL firmware-path guard in AlephQemuConfidentialResources.from_spec is still enforced. The test coverage is adequate, verifying the TeeConfig reconstruction and core field round-trip. No correctness, security, or quality issues found.

PR #1004, opened 2026-06-24
fix(storage): don't cap large runtime/image downloads at 300s, retry on timeout

odesenfans:
## Problem

A CRN operator reported a node that wouldn't dispatch instances — it appeared "stuck downloading a runtime." Investigation of the supervisor logs showed every instance failing identically:

```
storage.py:73  chunk = await resp.content.read(65536)
  → aiohttp/helpers.py  raise asyncio.TimeoutError
TimeoutError
```

…on a strict ~5-minute (300s) cadence.

### Root cause

`download_file_in_chunks` used a bare `aiohttp.ClientSession()`, which inherits aiohttp's **default `ClientTimeout(total=300s)`** — a hard cap on the *entire* request, body included. Runtimes and instance base images are routinely several hundred MiB and are sometimes served by slow gateways (the affected node's base image was an IPFS file on `ipfs.aleph.im`).

Measured from a well-connected client, that gateway delivers a **steady ~300 KB/s** — not a stall. But the shared instance base image is **582 MiB ≈ 34 min** at that rate, so the transfer was guillotined at exactly 300s on *every* attempt and the instance could never start.

Made worse by a second bug: `download_file`'s retry loop caught `ClientConnectionError / ClientResponseError / ClientPayloadError` but **not `asyncio.TimeoutError`**, so a timed-out download got zero in-process retries.

## Fix

- **`download_file_in_chunks`**: build the session with `total=None` plus a `sock_read` stall guard (and `sock_connect`). Large-but-steady downloads now complete, while a genuine stall (no bytes for 120s) still fails fast.
- **`download_file`**: add `asyncio.TimeoutError` to the retry tuple so a timed-out transfer is retried instead of aborting on the first failure.

## Tests

TDD — added to `tests/supervisor/test_storage.py` (watched each fail first):
- `test_download_file_in_chunks_does_not_cap_total_time` — session is built with `total is None` and a `sock_read` guard.
- `test_download_file_retries_on_timeout` — a timed-out chunk download is retried, not aborted.
- `test_download_file_aborts_after_exhausting_retries_on_timeout` — the error still propagates once retries are exhausted.

All 7 tests in the file pass; `ruff check`/`ruff format` clean on the changed files.

### DIFF
--- a/src/aleph/vm/storage.py
+++ b/src/aleph/vm/storage.py
@@ -37,6 +37,12 @@ from aleph.vm.utils import fix_message_validation, run_in_subprocess
 logger = logging.getLogger(__name__)
 DEVICE_MAPPER_DIRECTORY = "/dev/mapper"
 
+# Runtimes and base images can be several hundred MiB and are sometimes served by slow
+# gateways (e.g. IPFS). We must not cap the *total* transfer time, or large-but-steady
+# downloads get killed mid-stream; instead guard against genuine stalls with sock timeouts.
+DOWNLOAD_SOCKET_CONNECT_TIMEOUT_SECONDS = 30
+DOWNLOAD_SOCKET_READ_TIMEOUT_SECONDS = 120
+
 
 class CorruptedFilesystemError(Exception):
     """Raised when a file containing a filesystem is corrupted."""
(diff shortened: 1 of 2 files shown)


foxpatch-aleph (approved):
The fix correctly addresses two bugs: (1) aiohttp's default 300s total timeout killing large-but-steady downloads, and (2) asyncio.TimeoutError not being caught by the retry loop. The ClientTimeout configuration with total=None plus sock_connect/sock_read guards is the right approach — it allows multi-hundred-MB downloads from slow gateways to complete while still failing fast on genuine stalls. The retry tuple addition is correct since aiohttp raises asyncio.TimeoutError for sock_read timeouts. The finally block properly cleans up .part files between retries. Tests are well-structured and verify the key behaviors. One minor portability note: the tests raise builtin TimeoutError rather than asyncio.TimeoutError; on Python 3.11+ these are the same class so tests pass, but on Python 3.10 (which the project nominally supports) they are distinct and the tests would not exercise the intended code path. Since CI appears to run on Python 3.11+ and the production code catches the correct exception, this is non-blocking.

**`tests/supervisor/test_storage.py`** (line 177): The mock raises builtin `TimeoutError`, but the production code catches `asyncio.TimeoutError`. These are the same class on Python 3.11+ (which CI uses), but on Python 3.10 (which `pyproject.toml` lists as supported) they are distinct classes. For strict 3.10 compatibility, the test mocks should raise `asyncio.TimeoutError` instead to match what aiohttp actually raises. Non-blocking since CI doesn't test on 3.10.

PR #1005, opened 2026-06-24
fix(storage): don't cap large runtime/image downloads at 300s, retry on timeout

odesenfans:
Backport of #1004 to `dev`, so the fix ships with the agent / supervisor split release.

Cherry-pick of the same commit; the diff is identical to #1004 (`storage.py` and `test_storage.py` are identical between `main` and `dev`, so it applied cleanly).

## Problem

A CRN operator's node wouldn't dispatch instances — it appeared "stuck downloading a runtime." Every instance failed with `asyncio.TimeoutError` at `storage.py` `resp.content.read()`, on a strict ~5-minute cadence.

### Root cause

`download_file_in_chunks` used a bare `aiohttp.ClientSession()`, inheriting aiohttp's default `ClientTimeout(total=300s)` — a hard cap on the *whole* transfer. Runtimes / instance base images are routinely several hundred MiB and are sometimes served by slow gateways (the affected base image was a 582 MiB IPFS file on `ipfs.aleph.im`, delivered at a steady ~300 KB/s ≈ 34 min). Every attempt was guillotined at 300s, so the instance never started.

Compounded by `download_file`'s retry loop not catching `asyncio.TimeoutError`, so a timed-out download got zero retries.

## Fix

- `download_file_in_chunks`: build the session with `total=None` plus `sock_read`/`sock_connect` stall guards — slow-but-steady large downloads complete; genuine stalls still fail fast.
- `download_file`: add `asyncio.TimeoutError` to the retry tuple.

## Tests

TDD, added to `tests/supervisor/test_storage.py` (each watched failing first):
- `test_download_file_in_chunks_does_not_cap_total_time`
- `test_download_file_retries_on_timeout`
- `test_download_file_aborts_after_exhausting_retries_on_timeout`

All 7 tests in the file pass; ruff clean on the changed lines.

### DIFF
--- a/src/aleph/vm/storage.py
+++ b/src/aleph/vm/storage.py
@@ -37,6 +37,12 @@ from aleph.vm.utils import fix_message_validation, run_in_subprocess
 logger = logging.getLogger(__name__)
 DEVICE_MAPPER_DIRECTORY = "/dev/mapper"
 
+# Runtimes and base images can be several hundred MiB and are sometimes served by slow
+# gateways (e.g. IPFS). We must not cap the *total* transfer time, or large-but-steady
+# downloads get killed mid-stream; instead guard against genuine stalls with sock timeouts.
+DOWNLOAD_SOCKET_CONNECT_TIMEOUT_SECONDS = 30
+DOWNLOAD_SOCKET_READ_TIMEOUT_SECONDS = 120
+
 
 class CorruptedFilesystemError(Exception):
     """Raised when a file containing a filesystem is corrupted."""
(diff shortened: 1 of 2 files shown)


foxpatch-aleph (approved):
The fix correctly addresses the root cause: aiohttp's default 300s total timeout was killing large-but-steady downloads (e.g., 582 MiB at 300 KB/s ≈ 34 min). Setting total=None with per-read sock timeouts (sock_connect=30s, sock_read=120s) is the right approach — sock_read is per-read, not cumulative, so slow-but-progressing transfers complete while genuine stalls fail fast. Adding asyncio.TimeoutError to the retry tuple is correct and complements the existing ClientConnectionError catch (which already handles aiohttp's SocketTimeoutError subclass). The tests are well-structured with a shared session mock helper and cover the three key scenarios: no total cap, retry on timeout, and abort after exhaustion. One minor portability note: the test mocks raise bare TimeoutError (builtin) while the code catches asyncio.TimeoutError; these are identical on Python 3.11+ (used in CI) but distinct on Python 3.10 (the declared minimum), so the tests wouldn't be portable to 3.10. This doesn't affect production correctness.

**`tests/supervisor/test_storage.py`** (line 177): The mock raises bare `TimeoutError` (Python builtin), but the production code catches `asyncio.TimeoutError`. On Python 3.11+ these are the same type, so the test passes in CI. However, the project declares `requires-python = \">=3.10\"`, and on Python 3.10 `asyncio.TimeoutError` is a distinct class (not a subclass of `builtins.TimeoutError`), so this test would fail there. Consider raising `asyncio.TimeoutError` instead for full portability to the declared minimum Python version.

**`tests/supervisor/test_storage.py`** (line 165): Minor coverage gap: the test asserts `sock_read is not None` but doesn't check `sock_connect`. Adding `assert timeout.sock_connect is not None` would verify the connect-timeout guard is also properly set.

PR #1007, opened 2026-06-24
fix(supervisor): re-adopt a live controller on create instead of clobbering it

odesenfans:
## Problem

After a failed reattach, #1001 leaves the VM's controller **running but untracked** (absent from `self.executions`). The agent sees the VM missing from `list_vms` and calls create for it. But `create_vm_from_spec`'s idempotency check (`pool.py:308`) only fires for VMs already in the pool — so it falls through to a **fresh create over the live controller**:

- `get_unique_vm_index` treats the old index as free;
- `prepare_tap`/`create_tap` tears down the tap the running qemu still uses;
- `enable_and_start` targets the already-active unit, with monitor/qmp socket conflicts.

It never wipes the disks (the rootfs is *referenced*, not recreated — verified: `build_qemu_configuration` sets `image_path = spec.require_rootfs().path`), but it **bounces or collides with a live instance**.

## Fix

`_readopt_live_controller`, called under `creation_lock` before the fresh-create path: if a controller config for this `vm_id` is on disk **and** its systemd unit is **active** but the VM isn't in the pool, re-adopt it through the reattach path instead of creating a duplicate.

- **Re-adopt succeeds** → return the adopted execution (the agent's create "just works", and a transient failure that has since cleared heals here).
- **Re-adopt fails** (original cause persists) → the exception propagates. The caller must **never** fall through to a fresh create over a live controller; the VM stays untracked until the next attempt — never clobbered.
- **Controller inactive** → returns `None`, so the normal create proceeds. That's a clean restart-from-disk, which is the intended "restartable without losing disks" behavior.

## Scope

QEMU instances — the reattach-eligible, persistent path. Firecracker dispatches early to `_create_firecracker_from_spec` and isn't reattach-backed.

This nudges `create_vm_from_spec`'s already-flagged `C901` complexity from 11→12 (a pre-existing, CI-tolerated finding, not a new category); I deliberately didn't refactor that large method in this focused PR.

## Tests

- `test_readopt_live_controller_readopts_when_active` — live untracked controller → re-adopts (calls the reattach path, returns the adopted execution).
- `test_readopt_live_controller_none_without_config` / `_none_when_inactive` — no config / inactive controller → returns `None`, normal create proceeds.
- `test_create_vm_from_spec_short_circuits_to_readopt` — create returns the re-adopt result and never reaches admission.

## Stacking

Step 2 of 3 (after the #1002 revert, #1006). Independent of step 3 (proactive retry + bounded escalation), which makes recovery not depend on the agent poking create. `ruff==0.4.6` format/lint clean (no new findings), `isort` clean, `py_compile` clean; full pytest in CI.

### DIFF
--- a/src/aleph/vm/pool.py
+++ b/src/aleph/vm/pool.py
@@ -309,6 +309,16 @@ class VmPool:
                 self._require_same_spec(current_execution, spec)
                 return current_execution
 
+            # The VM may already be running without a pool entry: a prior
+            # reattach failed and was isolated (#1001), leaving the live
+            # controller protected but untracked. Re-adopt it rather than
+            # building a fresh VM on top -- a fresh create reuses the index/tap
+            # and control sockets the live qemu still holds (it never wipes the
+            # disks, but it bounces or collides with a running instance).
+            readopted = await self._readopt_live_controller(vm_id)
+            if readopted is not None:
+                return readopted
+
             # Authoritative capacity admission, folded into the create path so
             # the check and the registration below are atomic under the lock.
             self.check_spec_admission(spec)
@@ -406,6 +416,35 @@ class VmPool:
 
             self._schedule_forget_on_stop(execution)
             return execution
+
+    async def _readopt_live_controller(self, vm_id: VmId) -> VmExecution | None:
+        """Re-adopt a VM whose controller is still running but is absent from the
+        pool (a reattach that failed and was isolated by load_persistent_executions).
+
+        Returns the re-adopted execution, or None when there is nothing live to
+        adopt -- no on-disk config, or the controller is not active -- in which
+        case the caller proceeds with a normal create (an inactive controller is
+        a clean restart-from-disk, which is safe).
+
+        Called under ``creation_lock``. If re-adoption itself fails (the original
+        transient cause persists), the exception propagates: the caller must NOT
+        fall through to a fresh create over the live controller. The VM stays
+        untracked until the next attempt -- never clobbered.
+        """
+        config = load_controller_configuration(str(vm_id))
+        if config is None:
+            return None
+        service_name = f"aleph-vm-controller@{vm_id}.service"
+        if not self.systemd_manager.get_services_active_states([service_name]).get(service_name, False):
+            return None
+
+        logger.warning(
+            "create requested for %s but its controller is already running and untracked "
+            "(a previous reattach failed); re-adopting it instead of creating a duplicate",
+            vm_id,
+        )
+        await self._restore_running_execution_from_config(config, config.vm_id, vm_id)
+        return self.executions[vm_id]
 
     async def _create_firecracker_from_spec(self, spec: CreateVmSpec) -> VmExecution:
         """Message-free Firecracker boot.
--- a/src/aleph/vm/pool.py
+++ b/src/aleph/vm/pool.py
@@ -434,10 +434,11 @@ class VmPool:
         restart-from-disk, which is safe).
 
         Fail-closed: when systemd cannot report a definitive unit state (a
-        D-Bus error, or a transitional activating/deactivating state), this
-        raises InternalSupervisorError so the create fails and the agent
-        retries later, instead of falling through to a fresh create over a
-        possibly live controller.
+        D-Bus error, or a transitional activating/deactivating state), or when
+        the VM's on-disk vm_index is meanwhile claimed by another tracked
+        execution, this raises InternalSupervisorError so the create fails and
+        the agent retries later, instead of falling through to a fresh create
+        over a possibly live controller.
 
         Called under ``creation_lock``. If re-adoption itself fails (the original
         transient cause persists), the exception propagates: the caller must NOT
@@ -463,6 +464,23 @@ class VmPool:
             msg = (
                 f"Cannot determine the state of controller {service_name} (ActiveState: {state}); "
                 f"refusing to create VM {vm_id} over a possibly live controller, retry later"
+            )
+            raise InternalSupervisorError(msg)
+
+        # The on-disk vm_index may have been handed to a newer VM after the
+        # failed reattach (get_unique_vm_index only sees tracked executions).
+        # Restoring on it would rewire that VM's tap/nftables (the restore
+        # path trusts the index), and a fresh create over the live controller
+        # is forbidden, so the only safe outcome is to fail.
+        claimed_by = next(
+            (other_id for other_id, execution in self.executions.items() if execution.vm_index == config.vm_id),
+            None,
+        )
+        if claimed_by is not None:
+            msg = (
+                f"Cannot re-adopt VM {vm_id}: its on-disk vm_index {config.vm_id} is now claimed by "
+                f"running VM {claimed_by}; refusing to rewire that VM's networking or to create a "
+                f"duplicate, operator intervention required"
             )
             raise InternalSupervisorError(msg)
 
--- a/src/aleph/vm/pool.py
+++ b/src/aleph/vm/pool.py
@@ -29,10 +29,12 @@ from aleph.vm.supervisor.qemu_build import (
 )
 from aleph.vm.supervisor_interface.configuration import (
     Configuration,
+    get_controller_configuration_path,
     load_controller_configuration,
     save_controller_configuration,
 )
 from aleph.vm.supervisor_interface.errors import (
+    InternalSupervisorError,
     InvalidBackendError,
     VmAlreadyExistsError,
 )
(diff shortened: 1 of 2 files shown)


foxpatch-aleph (commented):
The fix is correct, well-documented, and follows existing conventions. The re-adopt logic properly prevents clobbering a live untracked controller, and error propagation is handled correctly. The only notable gap is a missing test for the exception-propagation case — the PR description explicitly states this is a critical safety property ("the caller must NOT fall through to a fresh create over the live controller"), but no test verifies that a re-adopt failure propagates without reaching the fresh-create path. Adding that test would close the coverage gap on the most safety-critical branch.

**`tests/supervisor/test_supervisor_reattach.py`** (line 205): Missing test for the exception-propagation case. The PR description states the caller must NOT fall through to a fresh create when re-adoption fails — this is the most safety-critical branch but is untested. Consider adding a test like:

async def test_create_vm_from_spec_propagates_readopt_failure(monkeypatch):
    pool = _bare_pool()
    pool.creation_lock = asyncio.Lock()
    monkeypatch.setattr(pool, "_readopt_live_controller", AsyncMock(side_effect=RuntimeError("reattach failed")))
    pool.check_spec_admission = MagicMock(side_effect=AssertionError("must not reach fresh create"))
    with pytest.raises(RuntimeError):
        await pool.create_vm_from_spec(_spec())
    pool.check_spec_admission.assert_not_called()

odesenfans:

The fix is the right shape: probing for a live untracked controller under `creation_lock` before the fresh-create path closes the #1001 clobbering hole, and the "on re-adopt failure, propagate rather than fall through" contract is exactly right. Two new issues should be fixed before merge, and two pre-existing gaps in the shared restore path are worth follow-ups since this PR widens their exposure.

**Fix before merge**

1. **The guard fails open on D-Bus errors.** `get_services_active_states` returns all-False on `DBusException` (`src/aleph/vm/systemd.py:251-254`), so `_readopt_live_controller` returns `None` and the caller proceeds to a fresh create over the live controller, the exact clobbering the docstring forbids (`pool.py:429-432`). A momentary D-Bus hiccup while the agent retries create reproduces #1001's damage. `get_service_active_state` (systemd.py:161) returns `"unknown"` on error; use it (or equivalent) and fail closed (raise / skip create) when the state cannot be determined. This also replaces the `ListUnits` full-unit-enumeration batch helper, a synchronous D-Bus sweep executed under `creation_lock` on every create that finds an on-disk config (the file itself documents ListUnits stalls of tens of seconds on busy hosts).

2. **Missing `_require_same_spec` on the re-adopt path.** The tracked-running idempotency branch three lines above validates the spec (`pool.py:308-310`); the re-adopt branch returns the old execution unconditionally (`pool.py:318-320`). A create with changed vcpus/memory silently gets the old-spec VM back where the tracked path raises `VmAlreadyExistsError`. The re-adopted execution has `vm_spec` populated, so the same check applies directly.

**Pre-existing in the shared restore path (follow-up material, exposure widened here)**

3. **No claimed-index guard.** `_readopt_live_controller` passes `config.vm_id` as `vm_index` unchecked. If `get_unique_vm_index` (scans only `self.executions`, `pool.py:516`) handed the untracked VM's index to a new VM in the meantime, `_restore_network` sees the interface exists, skips `create_tap`, and overwrites the other VM's nftables/NDP on that index. `load_persistent_executions` has a claimed-index guard for exactly this; the create-path copy lacks it. Extracting one shared single-VM reattach step (config -> active check -> claimed-index check -> restore) used by the startup loop, this path, and #1008's retry loop would keep the three in lockstep instead of drifting.

4. **`_restore_running_execution_from_config` never populates `execution.gpus`** (fresh create sets `execution.gpus = resolved_host_gpus` explicitly for `uses_gpu` accounting). A re-adopted GPU VM's card shows as available in `get_available_gpus` and can be double-assigned (vfio bind failure or double-counted admission). Same gap applies to the startup reattach path on dev.

**Minor**

5. Every fresh create of a genuinely new VM now logs `WARNING "Controller configuration file not found"` on the happy path (`supervisor_interface/configuration.py:111-112` via the probe at `pool.py:434`). Check path existence first, or demote that log.
6. `_readopt_live_controller` relies on the restore helper's side effect of registering the execution and then re-fetches via `self.executions[vm_id]`; having `_restore_running_execution_from_config` return the execution removes the hidden coupling.
7. The `f"aleph-vm-controller@{vm_id}.service"` scheme now has a fifth inline copy in pool.py (plus models.py / garbage_collector.py); a shared `controller_service_name()` helper would make the naming single-source.

foxpatch-aleph (approved):
Well-designed fail-closed fix that re-adopts a live but untracked controller instead of clobbering it with a fresh create. The systemd state machine correctly handles all cases: active VMs are re-adopted, definitively-down VMs allow a clean restart, and indeterminate states raise to prevent unsafe creates. The spec idempotency check is properly applied to the readopted execution, matching the tracked-running branch. Tests cover the key paths including the spec-mismatch conflict. The missing test for the exception-propagation contract (readopt fails → must not fall through to create) and the pre-existing vm_index collision risk in the reattach path are minor observations, not blockers.

**`src/aleph/vm/pool.py`** (line 474): Missing vm_index collision check: `load_persistent_executions` guards against duplicate vm_indexes via `claimed_vm_ids` (line 704-711), but `_readopt_live_controller` does not. If the untracked VM's original `config.vm_id` (vm_index) was claimed by a newer VM created after the failed reattach, `_restore_running_execution_from_config` will call `_restore_network` with a conflicting index, potentially disrupting the other VM's tap/nftables. This is a pre-existing risk in `_restore_running_execution_from_config`, not introduced by this PR, but the readopt path makes it reachable at create-time (not just startup). Consider adding a guard: if `config.vm_id` is already in use by another execution in the pool, either treat the controller as dead or pick a new index.

**`tests/supervisor/test_supervisor_reattach.py`** (line 169): Consider adding a test verifying that when `_restore_running_execution_from_config` raises during readopt, the exception propagates out of `create_vm_from_spec` and does NOT fall through to the fresh-create path. The docstring explicitly states this fail-closed contract ("the caller must NOT fall through to a fresh create"), but it's not tested. A simple test: mock `_readopt_live_controller` to call a `fake_restore` that raises, then assert `create_vm_from_spec` raises and `check_spec_admission` is not called.

**`src/aleph/vm/pool.py`** (line 455): Minor: `get_service_active_state` is a synchronous D-Bus call made while holding `creation_lock`. This blocks all concurrent creates until systemd responds. In practice this only fires when a config file already exists (the common path returns `None` early), so the impact is minimal, but worth noting for the record.

odesenfans:
Both actionable review observations addressed in f08cc578:

- **vm_index collision**: `_readopt_live_controller` now fails closed (`InternalSupervisorError`) when the on-disk `config.vm_id` is claimed by another tracked execution, before any restore runs. Neither of the review's two suggested remedies was safe here: the controller is demonstrably alive (treating it as dead would orphan a running VM), and the live qemu is physically bound to its existing tap (registering it under a new index would give the pool a view that does not match reality). Raising matches the indeterminate-systemd-state contract; #1008's `get_unique_vm_index` reservation closes the underlying collision window once both merge.
- **Exception-propagation contract**: now pinned by `test_create_vm_from_spec_readopt_failure_does_not_fall_through` (re-adopt raises, `create_vm_from_spec` raises, `check_spec_admission` never runs).

The sync D-Bus call under `creation_lock` is left as noted: it only fires when an on-disk config exists, and the per-unit query already replaced the previous full `ListUnits` sweep.

foxpatch-aleph (approved):
The PR correctly addresses the problem of a live but untracked controller being clobbered by a fresh create. The _readopt_live_controller method is placed correctly within creation_lock, after the tracked-running idempotency check and before admission. The fail-closed design is sound: indeterminate systemd states and vm_index conflicts raise InternalSupervisorError (preventing a fresh create over a live controller), and re-adopt failures propagate without falling through. The config.vm_id field (which is actually the vm_index, consistent with existing load_persistent_executions code) is used correctly as the vm_index argument to _restore_running_execution_from_config. Test coverage is thorough across all branches: active re-adopt, no-config/inactive/failed/not-loaded returning None, indeterminate states raising, vm_index conflict raising, create short-circuit, spec mismatch, and failure propagation.

**`src/aleph/vm/pool.py`** (line 454): Minor: if the config file exists but load_controller_configuration returns None (e.g. corrupt JSON that fails Pydantic validation), this returns None and falls through to a fresh create over the potentially live controller. In practice this is unlikely and the fresh create would likely fail at a later stage, but it's a narrow gap in the fail-closed contract. Consider whether a corrupt-but-present config with an active controller should raise rather than return None. Non-blocking — the scenario is exceptional.

**`tests/supervisor/test_supervisor_reattach.py`** (line 287): Test coverage is comprehensive. One small gap: there's no test for the case where get_controller_configuration_path().exists() is True but load_controller_configuration returns None (line 454 in pool.py). A test asserting that this returns None (current behavior) would document the intended contract for this edge case.

PR #1008, opened 2026-06-24
feat(supervisor): retry failed reattachments, give up loudly after a bound

odesenfans:
## What

A VM whose reattach fails at startup is isolated (#1001) — its controller keeps running but the supervisor doesn't manage it. Until now, recovery only happened if the agent later called create (#1007). This adds **proactive** recovery so a VM never sits in limbo waiting to be poked.

## How

- `load_persistent_executions` queues each failed VM in `_failed_reattach`.
- A background loop, `run_reattach_retry_loop`, retries adoption every `REATTACH_RETRY_INTERVAL_SECONDS` (30s). A **transient** cause — host networking not ready yet, a brief D-Bus hiccup — clears and the VM is re-adopted **in place, no downtime, no controller bounce**.
- Retries are serialized with `create_vm_from_spec` via `creation_lock`, so the two adoption paths (proactive retry here, on-demand re-adopt in #1007) never rebuild the same VM at once; whichever adopts first wins and the other drops it from the retry set.

## Bounded escalation — Option B

After `REATTACH_RETRY_MAX_ATTEMPTS` (5) the VM is marked **exhausted**: the supervisor logs an `ERROR` and stops retrying, but **does not auto-stop the live VM**. Rationale (your call): auto-stopping would bounce a running customer instance, and a host-wide failure (e.g. missing nft base chains affecting every VM) would cascade a recoverable glitch into a mass outage. The VM keeps running; recovery then needs operator action. `unmanaged_vm_ids` exposes the give-up set for observability.

Mass-failure detection was intentionally left out (kept simple, per discussion).

## Placement

The loop runs where the pool lives and self-terminates once nothing is left to retry (a clean startup is a no-op):
- **in-process agent** — an aiohttp `on_startup` task (`start_reattach_retry_task`), cancelled on cleanup;
- **gRPC daemon** — a task in `run_daemon`, cancelled on shutdown.

## Tests

- `test_retry_readopts_a_failed_vm` — a queued VM is re-adopted and dropped from the retry set.
- `test_retry_exhausts_and_leaves_vm_running` — after MAX attempts: exhausted, kept as unmanaged, and **not** auto-stopped (`stop_and_disable` never called).
- `test_retry_drops_vm_adopted_elsewhere` — a VM already in the pool is dropped without another restore.
- `test_retry_loop_returns_immediately_without_failures` — clean startup is a no-op (doesn't hang).

## Stacking

Step 3 of 3. Independent of step 1 (#1006 revert) and step 2 (#1007 re-adopt-on-create) — all off `dev`; step 2 and step 3 touch different methods in `pool.py` and compose (on-demand + proactive adoption). `ruff==0.4.6` format/lint clean (no new findings), `isort` clean, `py_compile` clean; full pytest in CI.

### DIFF
--- a/src/aleph/vm/agent/supervisor.py
+++ b/src/aleph/vm/agent/supervisor.py
@@ -7,6 +7,7 @@ evolve in the future.
 """
 
 import asyncio
+import contextlib
 import logging
 import sys
 from pathlib import Path
@@ -367,6 +368,27 @@ async def drain_in_flight_requests(app: web.Application):
         await pool.drain()
 
 
+async def start_reattach_retry_task(app: web.Application) -> None:
+    """on_startup: retry VMs whose reattach failed at load time (in-process mode).
+
+    No-op in split mode -- the daemon owns the pool and runs its own retry loop.
+    The loop self-terminates once nothing is left to retry.
+    """
+    pool: VmPool | None = app.get("_engine_pool")
+    if pool is None:
+        return
+    app["_reattach_retry_task"] = asyncio.create_task(pool.run_reattach_retry_loop())
+
+
+async def stop_reattach_retry_task(app: web.Application) -> None:
+    """on_cleanup: cancel the reattach retry loop."""
+    task = app.get("_reattach_retry_task")
+    if task is not None:
+        task.cancel()
+        with contextlib.suppress(asyncio.CancelledError):
+            await task
+
+
 async def stop_all_vms(app: web.Application):
     pool: VmPool | None = app.get("_engine_pool")
     if pool is None:
@@ -457,7 +479,9 @@ def run():
     app.on_startup.append(_run_migration_reaper)
     app.on_startup.append(_rehydrate_vm_registry)
     app.on_startup.append(start_node_hash_discovery)
+    app.on_startup.append(start_reattach_retry_task)
     app.on_cleanup.append(stop_node_hash_discovery)
+    app.on_cleanup.append(stop_reattach_retry_task)
 
     # Store sevctl app singleton only if confidential feature is enabled
     if settings.ENABLE_CONFIDENTIAL_COMPUTING:
--- a/src/aleph/vm/agent/supervisor.py
+++ b/src/aleph/vm/agent/supervisor.py
@@ -27,6 +27,7 @@ from aleph.vm.pool import VmPool
 from aleph.vm.sevclient import SevClient
 from aleph.vm.supervisor.grpc_client import GrpcSupervisor
 from aleph.vm.supervisor.local import LocalSupervisor
+from aleph.vm.utils import create_task_log_exceptions
 from aleph.vm.version import __version__
 
 from .node_identity import (
@@ -377,7 +378,7 @@ async def start_reattach_retry_task(app: web.Application) -> None:
     pool: VmPool | None = app.get("_engine_pool")
     if pool is None:
         return
-    app["_reattach_retry_task"] = asyncio.create_task(pool.run_reattach_retry_loop())
+    app["_reattach_retry_task"] = create_task_log_exceptions(pool.run_reattach_retry_loop(), name="reattach-retry-loop")
 
 
 async def stop_reattach_retry_task(app: web.Application) -> None:
--- a/src/aleph/vm/pool.py
+++ b/src/aleph/vm/pool.py
@@ -4,7 +4,7 @@ import asyncio
 import logging
 import shutil
 from collections.abc import Iterable
-from dataclasses import replace
+from dataclasses import dataclass, replace
 from datetime import datetime, timedelta, timezone
 from typing import Any
 
@@ -60,6 +60,27 @@ from .network.interfaces import remove_orphan_tap_interfaces
 
 logger = logging.getLogger(__name__)
 
+# Background retry of VMs whose reattach failed at startup (their controller is
+# still running but the supervisor could not rebuild its in-memory state). A
+# transient cause (e.g. host networking not ready yet) heals on a later pass.
+REATTACH_RETRY_INTERVAL_SECONDS = 30
+REATTACH_RETRY_MAX_ATTEMPTS = 5
+
+
+@dataclass
+class _FailedReattach:
+    """A VM left running-but-untracked after a failed reattach, pending retry.
+
+    ``attempts`` starts at 1 (the startup attempt). Once it reaches
+    ``REATTACH_RETRY_MAX_ATTEMPTS`` the VM is ``exhausted``: the supervisor stops
+    retrying and leaves the live controller alone (operator intervention needed).
+    """
+
+    config: Configuration
+    vm_index: int
+    attempts: int = 1
+    exhausted: bool = False
+
 
 class VmPool:
     """Pool of existing VMs
@@ -79,12 +100,16 @@ class VmPool:
     """Resources reserved by an user, before launching (only GPU atm)"""
 
     _draining: bool
+    # VMs whose reattach failed and that are awaiting background retry. Keyed by
+    # vm_id; entries are removed once re-adopted, kept (exhausted) once given up.
+    _failed_reattach: dict[VmId, _FailedReattach]
 
     def __init__(self):
         self.executions = {}
         self.reservations = {}
         self.gpus = []
         self._draining = False
+        self._failed_reattach = {}
 
         self.creation_lock = asyncio.Lock()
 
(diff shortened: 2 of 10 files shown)


foxpatch-aleph (approved):
Well-structured PR that adds proactive reattach recovery for VMs whose startup reattach failed. The retry loop is correctly serialized with the on-demand create path via creation_lock, the bounded escalation (Option B — exhaust and leave the VM running) is clearly documented and tested, and the placement in both in-process and daemon modes is clean. The off-by-one analysis confirms 5 total attempts (1 startup + 4 retries). Tests cover success, exhaustion, cross-path adoption, and the no-op case. No correctness or security issues found.

**`src/aleph/vm/pool.py`** (line 724): The retry loop correctly checks `vm_id in self.executions` under the lock and drops the entry. However, `create_vm_from_spec` (the on-demand create path) never removes the entry from `_failed_reattach` itself — the entry lingers until the next retry pass notices it. This is functionally harmless (the stale entry is just skipped on the next tick), but the PR description's claim of symmetry ("whichever adopts first wins and the other drops it from the retry set") is slightly oversold: the create path doesn't proactively drop, it relies on the retry loop to notice. Consider adding a one-line `self._failed_reattach.pop(vm_id, None)` inside `create_vm_from_spec` after registering the execution, or just note the asymmetry in a comment.

**`src/aleph/vm/pool.py`** (line 713): The loop sleeps *before* the first retry, so the first retry attempt happens 30s after startup (not immediately). This is arguably intentional — the startup attempt just failed, so giving it 30s to recover before retrying is reasonable — but it's worth a brief comment so future readers don't mistake it for an off-by-one in the sleep placement.

odesenfans:

Proactive recovery with a bound and no auto-stop (Option B) is the right call, and serializing the two adoption paths on `creation_lock` is sound. But the retry loop trusts stale state in three ways that can each turn recovery into damage; all three verified against the branch.

**Fix before merge**

1. **No liveness re-check before re-adoption.** `_retry_failed_reattachments_once` calls `_restore_running_execution_from_config` directly (`pool.py:723-729`); unlike `load_persistent_executions` (active-state gate + `_handle_dead_controller`, pool.py:638/658-660) it never checks the unit is still active. A controller that dies between startup and a retry tick gets registered as a running execution anyway: QEMU's `start_guest_api` is a no-op, `ready_event` is set blindly, tap/nftables are recreated for a dead VM, and status endpoints report it running. Re-check the unit each attempt and route dead controllers to `_handle_dead_controller`.

2. **A deleted VM gets resurrected by the retry loop.** No stop/delete path touches `_failed_reattach`. Deleting a queued VM raises `VmNotFoundError` from `_require` (`local.py:454-467`), the agent swallows it and drops its registry record, and 30s later the retry re-adopts the live controller: a VM the user deleted comes back as a managed, billed execution with no agent-side record left to ever stop it. Delete/stop of an untracked-but-queued VM should at minimum dequeue it (and arguably stop the controller, since delete intent is explicit).

3. **Queued/exhausted VMs' `vm_index` is not reserved.** `get_unique_vm_index` scans only `self.executions` (`pool.py:516`), so a fresh create can be handed a queued VM's index; `create_vm_from_spec` then deletes the live VM's tap (`pool.py:393-395`), and a later retry pass restores the old VM's nftables/NDP over the new VM's interface (`pool.py:761-772`), leaving two executions sharing one index. The index-scan half pre-exists on dev, but the PR introduces the exact registry of failed indexes and doesn't feed it into allocation; excluding `_failed_reattach` entries' `vm_index` from `get_unique_vm_index` closes the escalation.

**Minor**

4. Exhausted entries are never removed (`if state.exhausted: continue` runs before the in-executions cleanup, and the loop exits once all are exhausted), so `unmanaged_vm_ids` reports a VM forever, even after it is later re-adopted or properly recreated, and pins its full `Configuration` in memory. Relatedly, `unmanaged_vm_ids` currently has no production consumer (only the unit test); either wire it into a status surface or drop the property until it has one. A pending-only dict plus a plain `set[VmId]` for the given-up ids would also collapse the three `exhausted` filters.
5. The first background retry logs "Reattach retry 2/5" (attempts counts the startup attempt and is incremented before logging), so "retry 1/5" never appears and only 4 background retries run. The total-attempts semantics are documented as intended; just make the log say "attempt n/5".
6. Both start sites use bare `asyncio.create_task` where the repo's `create_task_log_exceptions` (`utils/__init__.py:112`) is the established pattern (tasks.py, node_identity.py, reactor.py); an exception escaping the loop's bookkeeping would otherwise surface only at shutdown, if at all.
7. `creation_lock` is held across each entire restore (tap/NDP/nftables, DB reads, port redirects). The serialization is intentional, but only the check-and-claim needs the lock; with several stuck VMs on a flaky host, each 30s pass serially blocks concurrent creates for the slowest restores. Worth revisiting if create latency matters (interacts with the ListUnits cost flagged on #1007).

Note the interplay with #1007: when both merge, its `_readopt_live_controller` should also dequeue from `_failed_reattach` on success rather than relying on the next retry tick to notice.

foxpatch-aleph (approved):
Well-structured PR that adds proactive recovery for VMs whose reattach failed at startup. The retry loop is correctly serialized with the create path via creation_lock, the liveness gate properly distinguishes dead/active/uncertain controller states, and the bounded escalation (Option B) leaves the VM running while surfacing it as unmanaged. The exhaustion counting is correct (1 startup + 4 retries = 5 total). Test coverage is thorough across all key scenarios. One minor inconsistency: the retry loop's dead-controller path does not call remove_controller_configuration (unlike discard_failed_reattach), leaving the on-disk config until next restart — not a bug but the comment 'Cleaning it up' slightly oversells it.

**`src/aleph/vm/pool.py`** (line 787): Minor inconsistency: when the retry loop finds a dead controller, it calls `_handle_dead_controller` (stop+disable the service) but does NOT call `remove_controller_configuration` to delete the on-disk config. Compare with `discard_failed_reattach` (line 740) which does both. The orphan config will be cleaned up by `_cleanup_orphan_resources` on the next supervisor restart, so this is not a bug, but the comment 'Cleaning it up' on line 780 is slightly misleading — the on-disk definition survives until next restart. Consider either adding `remove_controller_configuration(str(vm_id))` here for consistency, or adjusting the comment.

**`tests/supervisor/test_supervisor_reattach.py`** (line 244): The test runs `_retry_failed_reattachments_once()` 5 times (REATTACH_RETRY_MAX_ATTEMPTS), but only 4 calls are needed to exhaust: attempts starts at 1, and each failed retry increments by 1, so after 4 calls attempts=5 and exhausted=True. The 5th call hits `if state.exhausted: continue` and is a no-op. Not wrong, but the test could assert `state.attempts == 5` to document the exact count, or loop `REATTACH_RETRY_MAX_ATTEMPTS - 1` times for the meaningful iterations.

foxpatch-aleph (approved):
Well-structured PR that adds a bounded background retry loop for VMs whose reattach failed at startup. The concurrency model is correct: both the proactive retry path and the on-demand re-adopt path serialize through `creation_lock`, with double-checked guards inside the lock. The exhaustion logic (attempts starts at 1 for the startup attempt, increments per retry, exhausts at >= MAX) is off-by-one-free. Liveness gating in the retry loop correctly distinguishes positively-dead controllers (clean up + drop) from indeterminate states (spend an attempt, retry later). The `discard_failed_reattach` delete path properly acquires `creation_lock` to avoid racing with a concurrent retry adoption. Test coverage is comprehensive across all branches. No correctness, security, or race condition issues found.

**`src/aleph/vm/pool.py`** (line 876): When a controller is found dead during retry, `_handle_dead_controller` stops/disables it but the on-disk config file is not removed here (unlike `discard_failed_reattach` which calls `remove_controller_configuration` explicitly). This is consistent with the startup path where `_cleanup_orphan_resources` handles config removal at the next restart, but it means the config file lingers until then. Not a bug given `stop_and_disable` prevents systemd revival, just an observation about the asymmetry with `discard_failed_reattach`.

**`src/aleph/vm/pool.py`** (line 842): The loop sleeps 30s *before* the first retry attempt. This is intentional (gives the host time to settle after startup), but worth noting: a VM whose reattach failed due to a transient cause that clears within seconds will still wait 30s for the first retry. If faster initial recovery is desired later, an initial shorter delay or immediate first attempt could be considered.

**`src/aleph/vm/pool.py`** (line 828): `discard_failed_reattach` stops the controller and removes the config, but does not clean up networking resources (nft chains, tap interfaces, port redirects) for the vm_index. These are left for `_cleanup_orphan_resources` at the next startup. This is consistent with the existing dead-controller handling at startup, but means a new `create` could pick the same vm_index before the next startup and encounter stale nft/tap resources. The `prepare_tap`/`create_tap` path handles existing interfaces, so this is not a crash risk, just a minor resource leak window.

**`tests/supervisor/test_supervisor_inprocess_lifecycle.py`** (line 65): The `import asyncio` inside the test body is a minor style nit — the module is already available at the top of the file via other imports. Could be hoisted to the top, but it's harmless.

foxpatch-aleph (changes requested):
Well-structured retry loop with solid locking, good test coverage, and correct off-by-one logic. However, the dead-controller path in _retry_failed_reattachments_once is missing a remove_controller_configuration call, leaving the on-disk config file to linger and potentially be revived on the next startup — inconsistent with discard_failed_reattach which does clean it up. This should be fixed before merge.

**`src/aleph/vm/pool.py`** (line 876): Bug: when the retry loop discovers the controller has died since startup, it calls _handle_dead_controller (which only does stop_and_disable) and removes the entry from _failed_reattach, but does NOT call remove_controller_configuration. The on-disk config file (<hash>-controller.json) lingers. This is inconsistent with discard_failed_reattach (line 828-829) which calls both _handle_dead_controller and remove_controller_configuration. Without removing the config, the next supervisor startup will find it in load_persistent_executions, see the controller is inactive, and call _handle_dead_controller again (redundant). Fix: add `remove_controller_configuration(str(vm_id))` after the _handle_dead_controller call, mirroring discard_failed_reattach.

**`src/aleph/vm/supervisor/local.py`** (line 477): Minor: when delete_vm handles a failed-reattach VM (the early return at line 477), the `wipe=True` path is skipped — persisted port mappings and data volumes are not cleaned up. Since the VM was never fully tracked there's no execution object to call erase_volumes() on, but port mappings could still exist on disk (they were persisted by a prior run). Consider whether delete_port_mappings(vm_id) should be called here when wipe is True, or document that wipe is a no-op for untracked VMs.

**`src/aleph/vm/pool.py`** (line 842): Style nit: the while condition `any(not state.exhausted for state in self._failed_reattach.values())` is evaluated without the creation_lock. This is safe in single-threaded asyncio (no await between the check and the sleep), but a brief comment noting that the unlocked read is intentional (no concurrent coroutine can modify _failed_reattach without an await point) would help future readers.

PR #1009, opened 2026-06-24
refactor(db): stop cross-actor DB access ahead of splitting the DB

odesenfans:
First of two PRs toward **one DB per actor**. This one removes the cross-actor DB access so the files can be split cleanly in PR2; no engine/file changes yet.

## Why

Today both tables live in `executions.sqlite3` and are touched from **both** processes — which only works because they share the file:
- the agent calls `delete_port_mappings` (a hypervisor table) in two dealloc paths;
- the supervisor calls `delete_record` (an agent table) in `record_usage`.

After the split each process opens only its own DB, so these have to go first.

## `port_mappings` → hypervisor-owned

`LocalSupervisor.delete_vm` now drops a VM's port mappings on **every** delete, not just `wipe`. (A stop keeps them so a persistent VM restarts with the same forwards; a delete is final.) That absorbs the two **residual** agent-side `delete_port_mappings` calls — `tasks.py` terminal dealloc and `views` allocation stop-loop, both already flagged in comments as "moves fully hypervisor-side with the gRPC split" — which are removed.

## `executions` → agent-owned

`record_usage` called `delete_record(execution_uuid=str(self.uuid))` from the supervisor. **It was a dead no-op:** execution records are created only agent-side with a fresh `uuid4()` (`vm_registry.persist_record`), so the supervisor's `VmExecution.uuid` never matched a row. Removed it, plus the now-unused `delete_record` function and its `delete` import.

## Behaviour

Neutral except one intended change: `delete_vm` now always clears port mappings (the agent used to do it via the residual call). The dead `delete_record` removal changes nothing.

Tests updated to pin the new ownership: dealloc paths (`test_checkpayment`, `test_views` stop-loop) assert `supervisor.delete_vm` only; `test_delete_vm_without_wipe_keeps_data` becomes `..._but_drops_port_mappings`.

## Follow-up (noted, not in scope)

Execution records now have no real GC (the deleter was the no-op) — they accumulate and are deduped on rehydrate. Proper agent-side record GC on delete is a separate change.

## Next

PR2: move `PortMapping` + CRUD to a supervisor-side module, split into two engines/DB files (`SUPERVISOR_DATABASE`), and add the one-time `port_mappings` data migration on upgrade.

## Verification

`ruff==0.4.6` format + lint clean (no new findings; remaining are pre-existing), `isort` clean, `py_compile` clean across all 8 files. Full pytest in CI.


foxpatch-aleph (approved):
Well-structured refactoring that correctly removes cross-actor DB access ahead of the DB split. The two agent-side `delete_port_mappings` calls in `tasks.py` and `views/__init__.py` are removed, with cleanup consolidated in `LocalSupervisor.delete_vm` (which runs supervisor-side). The dead `delete_record` call in `record_usage` is correctly identified as a no-op (uuid1 vs uuid4 mismatch) and removed along with the function and its `delete` import. The behavioural change — `delete_vm` now always drops port mappings, not just on `wipe` — is the right semantics for a final delete. The remaining `delete_port_mappings` call in `record_usage` (models.py:696) is correctly left in place since it runs supervisor-side via the pool. Tests are updated to pin the new ownership boundaries. One stale comment in `operator.py` references the removed `delete_record` behaviour.

**`src/aleph/vm/agent/views/operator.py`** (line 253): Stale comment: "past executions keep their record until record_usage deletes it" — `record_usage` no longer deletes execution records (the `delete_record` call was removed in this PR). The PR description acknowledges this ("Execution records now have no real GC"), so this comment should be updated to reflect that records persist until rehydrate dedup, not until `record_usage` deletes them.

odesenfans:

The boundary cleanup itself is correct: the `delete_record` no-op analysis checks out (records are created agent-side with a fresh uuid4 that never matches `VmExecution.uuid`), and moving port-mapping deletion hypervisor-side is the right ownership. But the "a delete is final" premise behind the one intended behavior change does not hold for two agent flows that use delete+recreate as a recovery/update cycle, and both now silently rotate a persistent VM's host ports. Verified against the branch and against dev's old semantics.

**Fix before merge**

1. **Crash recovery loses stable host ports.** `start_persistent_vm`'s FAILED branch calls `supervisor.delete_vm(vm_id)` (no wipe) then recreates (`agent/run.py:705-712`). On dev, a non-wipe delete kept the persistent VM's `PortMapping` rows and the recreate reloaded them (`pool.py:392-397` "Reuse persisted host ports across restarts"). With the unconditional `delete_port_mappings` in `delete_vm` (`local.py:484`), the recreate loads `{}` and `add_port_forward` assigns fresh random host ports: the owner's SSH endpoint silently changes after an automatic crash recovery.

2. **Same mechanism on message updates.** `UpdateWatcher._watch` reaps via `supervisor.delete_vm` and the VM is redeployed after (`agent/update_watcher.py:73-78`); a routine message amendment now rotates every forwarded host port where dev preserved them.

Both callers treat delete as "tear down, will recreate", not "final dealloc". Options: keep rows on non-wipe delete of persistent VMs (dev semantics, and let the two removed agent-side calls be absorbed by wipe-or-final paths only), or give `delete_vm` an explicit `final`/`keep_ports` distinction so recovery/update cycles can opt out. The current single flag conflates two intents.

**Minor**

3. When `delete_vm` raises `VmNotFoundError` in the `update_allocations` stop-loop, the except swallows it before `delete_port_mappings` is ever reached, whereas dev deleted the rows after the except (`views/__init__.py:600-602`). The window is narrow (a race with forget-on-stop), and the port picker skips active DB rows so the unique-index trip is unlikely; the real cost is a slow leak of active rows shrinking the assignable port pool. Worth a cleanup path for rows whose VM no longer exists.
4. Non-persistent deletes now soft-delete the same rows twice (`record_usage` in models.py:693-696 via stop, then `delete_vm` at local.py:484). Idempotent, but two owners of one cleanup rule; a comment cross-referencing them would prevent divergence, or let `delete_vm` be the single owner and drop the `record_usage` branch once the stop-paths-that-bypass-delete_vm question is settled.
5. Stale docstring: `operator.py:252-253` still says "past executions keep their record until record_usage deletes it", but this PR removed `delete_record` entirely and no deletion path remains on this branch (until #1010 adds `delete_records_for_vm`). Update the text here or in #1010.
6. Four tests keep a now-unused `mocker` parameter after their only patch was removed (`test_views.py:1282/1545/1674/1740`).

foxpatch-aleph (commented):

foxpatch-aleph (approved):
Clean, well-documented refactor that correctly removes cross-actor DB access ahead of the DB split. The behavioral change (delete_vm always drops port mappings unless keep_port_mappings=True) properly absorbs the removed agent-side delete_port_mappings calls, the dead delete_record removal is justified, and test coverage is thorough. The remaining cross-actor access in models.py:record_usage is explicitly scoped to PR2.

**`src/aleph/vm/supervisor/local.py`** (line 26): Note: this import of `delete_port_mappings` from `aleph.vm.agent.metrics` is itself a cross-actor DB access (supervisor calling agent-side module). It's correctly acknowledged in the PR description as PR2 scope (moving PortMapping CRUD to a supervisor-side module), but worth noting it remains a layering violation until then.

**`src/aleph/vm/models.py`** (line 696): This `delete_port_mappings` call in `record_usage` is the other remaining cross-actor DB access (supervisor-side VmExecution calling agent-side metrics). The double-delete for non-persistent VMs (here + in `local.py:489`) is idempotent since it's a soft-delete, but the comment in local.py:486-487 correctly flags the need to keep both sites in sync. This is fine for PR1 but should be addressed in PR2.

PR #1010, opened 2026-06-24
refactor(db): split the DB per actor (supervisor port mappings vs agent records)

odesenfans:
Second of two PRs toward **one DB per actor** (stacks on #1009). The agent keeps `executions.sqlite3` (executions/records); the supervisor gets its own `supervisor.sqlite3` (port mappings).

> Was stacked on #1009; now rebased directly onto `dev` after its merge. The boundary cleanup there is what makes splitting the files safe (no code crosses actors anymore).

## What moved

- **New `supervisor/networking_db.py`** owns the `PortMapping` model, its own engine + `SupervisorSessionMaker`, schema creation, the CRUD (`save`/`get`/`delete_port_mappings`, moved verbatim), and the upgrade migration. This also fixes the boundary wrinkle where `pool.py` imported port-mapping code from `aleph.vm.agent.metrics`.
- **`metrics.py`** loses `PortMapping` and its functions — the agent DB now holds only `executions`.
- **`conf.py`**: new `SUPERVISOR_DATABASE` setting (default `EXECUTION_ROOT/supervisor.sqlite3`) + `make_supervisor_db_url` / `make_supervisor_sync_db_url`.
- **`port_availability_checker.py`** (a sync `port_mappings` reader) now points at the supervisor DB.

## Engine wiring

Centralised in **`VmPool.setup()`**: the process that runs the pool (the gRPC daemon, or the agent in-process) sets up the supervisor engine, creates the schema, and runs the migration — so daemon, agent-in-process, and cli bench/test all get it without touching each entrypoint. `daemon.py` no longer sets up the agent DB (it doesn't use `executions`); the agent's `cli.py` still sets up `executions`.

## Upgrade safety

`migrate_port_mappings_from_legacy_db()` copies **active** rows from the pre-split `executions.sqlite3` into `supervisor.sqlite3` on first start, so live VMs keep their host-port forwards (otherwise the supervisor would come up blind to existing forwards → orphaned nft rules + port-reuse conflicts). Idempotent and self-skipping: same file, missing legacy DB/table, or an already-populated target all no-op. Both stores are SQLite, so rows are copied verbatim (preserving `created_at`; `id` autoincrements fresh).

The agent DB's now-vestigial `port_mappings` table (still created by the agent Alembic chain) is left in place; dropping it is a later cleanup.

## Tests

- `test_port_mappings` imports/patches the new module; new cases cover the migration (copies active rows, skips soft-deleted, idempotent, same-file / missing-legacy no-ops).
- `test_qemu_instance` sets up the supervisor engine for the real VM lifecycle (it calls `execution.stop()` → `record_usage`).
- Mock targets like `aleph.vm.pool.get_port_mappings` / `local.delete_port_mappings` still resolve (the names are imported into those namespaces), so the many tests patching by string are unaffected.

## Verification

`ruff==0.4.6` format clean, `isort` clean, `py_compile` clean across all 10 files. No new lint categories — `networking_db`'s `PLW0603` mirrors `metrics.py`'s existing `setup_engine` idiom; the other findings are pre-existing. Full pytest in CI (local env can't build `dbus-python`).

## Follow-ups (noted)

- Drop the vestigial `port_mappings` table from the agent DB via an Alembic migration.
- The execution-record GC gap from #1009.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -331,10 +331,9 @@ allow_indirect_imports = "true"
 source_modules = [ "aleph.vm.supervisor", "aleph.vm.pool", "aleph.vm.models" ]
 forbidden_modules = [ "aleph.vm.agent" ]
 # Residuals removed by the VmExecution/VmPool cleave + DB-layer extraction.
+# The DB split (per-actor databases) removed the pool/daemon/local ->
+# agent.metrics residuals; only these two remain.
 ignore_imports = [
   "aleph.vm.models -> aleph.vm.agent.metrics",
-  "aleph.vm.pool -> aleph.vm.agent.metrics",
   "aleph.vm.pool -> aleph.vm.agent.utils",
-  "aleph.vm.supervisor.daemon -> aleph.vm.agent.metrics",
-  "aleph.vm.supervisor.local -> aleph.vm.agent.metrics",
 ]
--- a/src/aleph/vm/agent/cli.py
+++ b/src/aleph/vm/agent/cli.py
@@ -285,6 +285,9 @@ async def run_instances(instances: list[ItemHash]) -> None:
     logger.info(f"Instances to run: {instances}")
     loop = asyncio.get_event_loop()
     pool = VmPool()
+    # Bind the supervisor DB engine (port mappings) and set up the network;
+    # the create path hits that DB on its first port-mapping lookup.
+    await pool.setup()
     # The main program uses a singleton pubsub instance in order to watch for updates.
     # We create another instance here since that singleton is not initialized yet.
     # Watching for updates on this instance will therefore not work.
--- a/src/aleph/vm/agent/metrics.py
+++ b/src/aleph/vm/agent/metrics.py
@@ -4,7 +4,17 @@ from pathlib import Path
 from typing import Any
 from uuid import UUID
 
-from sqlalchemy import JSON, Boolean, Column, DateTime, Float, Integer, String, select
+from sqlalchemy import (
+    JSON,
+    Boolean,
+    Column,
+    DateTime,
+    Float,
+    Integer,
+    String,
+    delete,
+    select,
+)
 from sqlalchemy.ext.asyncio import (
     AsyncEngine,
     AsyncSession,
@@ -108,3 +118,15 @@ async def get_last_record_for_vm(vm_hash) -> ExecutionRecord | None:
             select(ExecutionRecord).where(ExecutionRecord.vm_hash == vm_hash).limit(1)
         )  # Use execute for querying
         return result.scalar()
+
+
+async def delete_records_for_vm(vm_hash: str) -> None:
+    """Delete every execution record for a VM.
+
+    The on-disk sibling of AgentVmRegistry.forget: called when the agent
+    permanently drops a VM (terminal dealloc, allocation removal, operator
+    erase) so its records do not linger and get re-hydrated on the next restart.
+    """
+    async with AsyncSessionMaker() as session:
+        await session.execute(delete(ExecutionRecord).where(ExecutionRecord.vm_hash == vm_hash))
+        await session.commit()
--- a/src/aleph/vm/agent/metrics.py
+++ b/src/aleph/vm/agent/metrics.py
@@ -1,22 +1,10 @@
 import logging
 from collections.abc import Iterable
-from datetime import datetime, timezone
 from pathlib import Path
 from typing import Any
 from uuid import UUID
 
-from sqlalchemy import (
-    JSON,
-    Boolean,
-    Column,
-    DateTime,
-    Float,
-    Index,
-    Integer,
-    String,
-    select,
-    update,
-)
+from sqlalchemy import JSON, Boolean, Column, DateTime, Float, Integer, String, select
 from sqlalchemy.ext.asyncio import (
     AsyncEngine,
     AsyncSession,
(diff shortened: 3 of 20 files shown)


foxpatch-aleph (approved):
A clean, well-executed DB split. The port-mapping model, CRUD, and engine are moved verbatim to a new supervisor-owned `networking_db.py` module with its own `supervisor.sqlite3`. The migration is idempotent, uses read-only mode for the legacy DB, and handles all edge cases. The behavioral change making `delete_port_mappings` unconditional in `delete_vm` is correct (delete is final) and the double-call for non-persistent VMs is safely idempotent. Test coverage is thorough for the migration and the lifecycle behavioral changes. The `delete_record` removal from `record_usage` is also correct.

**`src/aleph/vm/supervisor/networking_db.py`** (line 185): The `Path` equality check (`legacy == target`) does a lexical comparison, not a filesystem-resolved one. If a user configures `SUPERVISOR_DATABASE` and `EXECUTION_DATABASE` to point at the same file via different path representations (e.g., one with a symlink or `..`), the migration would proceed and attempt to copy rows from the DB into itself. This is unrealistic with the default settings but worth noting.

**`src/aleph/vm/supervisor/networking_db.py`** (line 201): The migration copy is not wrapped in an explicit transaction (`BEGIN`/`COMMIT`). While `sqlite3`'s implicit transaction + the explicit `tgt.commit()` handles this correctly, wrapping the `executemany` + `commit` in a `with tgt:` context manager would make the transaction boundary more explicit and ensure rollback on partial failure.

odesenfans:
**Update — folded in one of the two follow-ups: agent-side execution-record GC** (`33954f7f`).

`metrics.delete_records_for_vm(vm_hash)` is now called at the three permanent-dealloc sites (terminal payment dealloc, allocation stop-loop, operator erase) — the on-disk sibling of `AgentVmRegistry.forget`. The `run.py` forget calls are create-time error cleanup (before `persist_record`), so they need no DB delete. Covered by a real-DB unit test + the three dealloc tests.

**The other follow-up — dropping the vestigial `port_mappings` table from the agent DB — is deliberately NOT in this PR, for correctness, not tidiness.** The supervisor's `migrate_port_mappings_from_legacy_db()` reads the agent DB's `port_mappings`. The agent's Alembic runs (`cli.py:407`) **before** `pool.setup()`'s data migration (`supervisor.run()`, line 425) in-process — and the two are unordered across processes in split mode. An Alembic `DROP port_mappings` would therefore race ahead of the copy and destroy the very rows the migration preserves. This is a two-phase migration: copy in this release, drop the table in a later release once every node has migrated.

foxpatch-aleph (approved):
Clean, well-structured refactoring that splits the database per actor. The port-mapping CRUD is moved verbatim to a new supervisor-owned `networking_db.py`, the migration is idempotent with proper edge-case handling, and the behavioral change in `delete_vm` (always cleaning port mappings on delete, not just on wipe+persistent) fixes a real gap. All references are updated, tests cover the new and changed behavior, and no stale imports remain.

**`src/aleph/vm/supervisor/local.py`** (line 484): Non-persistent VMs will have `delete_port_mappings` called twice: once from `record_usage()` (inside `stop()` → `stop_vm()`) and once here. The second call is a benign no-op since `delete_port_mappings` only soft-deletes rows where `deleted_at IS NULL`, but a brief comment noting this would help future readers understand the double-call is intentional and safe.

**`src/aleph/vm/supervisor/networking_db.py`** (line 197): The `sqlite3.OperationalError` catch handles a legacy DB that predates the `port_mappings` table, but there's no test exercising this path. The existing tests cover missing-file and same-file no-ops, but not the case where the legacy DB exists and lacks the table. A test like `test_migrate_port_mappings_noop_when_legacy_table_missing` would close the gap.

**`src/aleph/vm/supervisor/networking_db.py`** (line 191): The `file:{legacy}?mode=ro` URI will break if the path contains spaces or special characters (e.g. a `Path` with a space would produce `file:/path with space?mode=ro`). This is unlikely for a SQLite DB path in this project, but using `legacy.as_uri()` or `urllib.parse.quote` would be more robust.

odesenfans:
**Automated review (Claude, multi-agent find + verify pass; reviewed as the standalone diff on top of #1009).**

The split is cleanly executed: verbatim CRUD move, engine wiring centralized in `VmPool.setup()` so all three process shapes get it, a self-skipping legacy copy, and the `delete_records_for_vm` addition closes the record-retention gap #1009 left. One real breakage to fix, plus a handful of hardening/cleanup notes. The cross-event-loop concern one might raise about `asyncio.run(pool.setup())` was checked and refuted: it is the same proven pattern the metrics engine has always used.

**Fix before merge**

1. **`run_instances()` never calls `pool.setup()`, breaking the test/fake-instance CLI path.** `cli.py:287` creates `VmPool()` and goes straight to `start_instance` (contrast `benchmark()` at cli.py:196-197, which awaits `pool.setup()`). `SupervisorSessionMaker` is only bound in `setup_supervisor_engine()` (networking_db.py:38 is an annotation-only statement), so the first `get_port_mappings` (`pool.py:408`) raises `NameError`. This worked pre-split because the CRUD used `metrics.AsyncSessionMaker`, which cli `main()` initializes. `--run-test-instance` / `--run-fake-instance` are what the droplet CI jobs drive, so this should show up red in CI; add `await pool.setup()` there.

**Hardening (minor)**

2. The migration's done-marker is "any row exists in the target table" (networking_db.py:189-190). An upgrade -> downgrade-below-the-split -> re-upgrade sequence leaves downgrade-era mappings stranded in the legacy DB (target already has rows, so the copy never runs); aggregate reconciliation will recreate forwards but possibly on different host ports. If downgrade windows are a supported path, a real marker (or merge-copy of missing vm_hashes) closes it; if not, maybe just document it.
3. The target probe `SELECT 1 FROM port_mappings LIMIT 1` is unguarded while the legacy read is wrapped in `except OperationalError`, so the function's "idempotent and self-skipping" contract silently depends on `create_supervisor_tables` having run first (and `sqlite3.connect(target)` creates an empty file as a side effect). The one call site is ordered correctly today; guarding both sides the same way removes the latent ordering trap.
4. Agent Alembic 0004 still creates `port_mappings` in the agent DB on every fresh install, so new nodes carry a permanently empty dead table (until #1011 drops it, which is gated for phase 2). Fine as a transitional state; just noting the linkage.

**Cleanup (take or leave)**

5. `networking_db.py` duplicates the metrics engine scaffolding verbatim (annotation-only global sessionmaker, setup/create_tables, and the dead `try/except ImportError` shim for `declarative_base`, unreachable under the `sqlalchemy>=2` pin). A shared `make_engine_and_sessionmaker(url)` helper would keep future pragmas (WAL, busy_timeout) from landing in one DB only; at minimum drop the ImportError shim from the new file.
6. conf.py now has four one-line sqlite URL builders differing only in field and driver prefix; marginal, but a parameterized helper would stop the family growing with each future per-actor DB.

foxpatch-aleph (approved):
Well-structured DB split that cleanly separates supervisor-owned port mappings from agent-owned execution records. The new networking_db.py is a faithful extraction of the port-mapping CRUD with a robust, idempotent migration. The delete_vm behavioral change (always deleting port mappings unless keep_port_mappings=True) corrects the old orphan-leaving behavior, and all callers are correctly updated. Test coverage is thorough across migration edge cases, lifecycle semantics, proto bindings, and the CLI regression. The only issues are cosmetic: a slightly stale docstring in operator.py and a minor type annotation mismatch in test_expiry.py.

**`src/aleph/vm/agent/views/operator.py`** (line 253): The docstring says "nothing deletes it today (a follow-up adds explicit deletion on permanent dealloc)" but this PR already adds `delete_records_for_vm` calls at the permanent dealloc sites (tasks.py:397, views/__init__.py:604, operator.py:664). Consider updating this comment to reflect that deletion now happens on permanent dealloc.

**`tests/supervisor/test_expiry.py`** (line 16): Type annotation `list[tuple[str, bool]]` doesn't include the `keep_port_mappings` value that `delete_vm` now accepts. Functionally correct since only `(vm_id, wipe)` is appended, but the annotation is now inconsistent with the updated `test_update_watcher.py` FakeSupervisor which uses `list[tuple[str, bool, bool]]`.

foxpatch-aleph (approved):
Well-structured PR that cleanly splits the DB per actor: port mappings move to a new supervisor-owned `networking_db.py` with its own engine/session maker, while the agent keeps `executions.sqlite3`. The migration is idempotent and handles all the right edge cases (same file, missing legacy, missing table, already populated). All three entrypoints (daemon, agent in-process, CLI) now route through `VmPool.setup()` for supervisor DB initialization. Test coverage is thorough for the new migration and the `delete_records_for_vm` addition. No security issues — parameterized queries, read-only legacy DB access, no injection surface. The remaining cross-actor import (`models.py` → `agent.metrics.save_execution_data`) is pre-existing and harmless (file I/O, not DB access).

**`src/aleph/vm/supervisor/networking_db.py`** (line 203): The `tgt.executemany(...)` insert is not wrapped in a try/except. If the legacy DB contains data that violates the partial unique index on `host_port` (e.g., two active rows with the same host_port from a pre-index era), this will raise `sqlite3.IntegrityError` and crash `VmPool.setup()`, preventing the supervisor from starting at all. Consider catching `sqlite3.IntegrityError` here and logging a warning with the offending rows, so a corrupted legacy DB doesn't block supervisor startup — the live VMs' nftables rules are still in place and can be reconciled separately.

**`src/aleph/vm/supervisor/networking_db.py`** (line 29): Minor inconsistency: `metrics.py` guards the `declarative_base` import with a try/except fallback to `sqlalchemy.ext.declarative` (lines 25-28) for older SQLAlchemy compatibility, but `networking_db.py` imports it directly from `sqlalchemy.orm` only. Not a bug if the project pins a sufficiently new SQLAlchemy, but the inconsistency means the two modules could behave differently across environments.

**`src/aleph/vm/supervisor/networking_db.py`** (line 210): `len(rows)` is referenced after the `with closing(sqlite3.connect(target)) as tgt:` block exits. This works because Python doesn't block-scope `with` variables, but it's slightly fragile — if the block were later refactored into a helper function, `rows` would be out of scope. Consider moving the `logger.info` call inside the `with` block or capturing `count = len(rows)` before the block closes.

**`src/aleph/vm/agent/views/operator.py`** (line 664): If `metrics.delete_records_for_vm` raises (e.g., the agent DB engine was never set up in this process), the VM has already been erased hypervisor-side (line 658) and forgotten from the registry (line 663), but the response will be a 500 instead of 200. The record deletion failure leaves stale execution records that will be re-hydrated on restart. Consider wrapping this in a try/except that logs but doesn't fail the erase operation, since the critical actions (hypervisor erase + registry forget) have already succeeded.

foxpatch-aleph (approved):
Well-executed DB split: the supervisor gets its own SQLite file for port mappings, the agent keeps executions. The migration is idempotent and handles all edge cases. Engine wiring centralized in VmPool.setup() is clean. The new delete_records_for_vm is correctly placed at all permanent-dealloc sites with proper ordering. Test coverage is thorough across migration, engine setup, CLI regression, and the new delete function. No correctness or security issues found.

**`src/aleph/vm/supervisor/networking_db.py`** (line 195): The SELECT includes `deleted_at` in the column list but filters `WHERE deleted_at IS NULL`, so the value is always NULL. The INSERT also writes `deleted_at`. This is harmless (NULLs are inserted) but slightly redundant — you could drop `deleted_at` from both the SELECT and INSERT column lists since only active rows are copied.

**`src/aleph/vm/supervisor/networking_db.py`** (line 12): This module has `from __future__ import annotations` while the sibling `metrics.py` does not. Minor style inconsistency — not a problem since `Base: Any` and `SupervisorSessionMaker: async_sessionmaker[AsyncSession]` are module-level annotations that work fine as strings under PEP 563, and the actual values are assigned by function calls.

foxpatch-aleph (approved):
Clean, well-executed DB split that moves port-mapping persistence from the agent's executions DB to a new supervisor-owned SQLite file. The CRUD in networking_db.py is a verbatim move with only the session maker retargeted, the engine setup is correctly centralized in VmPool.setup() so all entrypoints (daemon, agent in-process, CLI) get it, and the legacy migration is idempotent with proper guards for same-file, missing-DB, missing-table, and already-populated scenarios. The new delete_records_for_vm correctly complements registry.forget at all three permanent-dealloc sites (check_payment, update_allocations, operator erase). Test coverage is thorough and mock targets are correct. The import-linter ignore cleanup is accurate — the three removed ignores correspond to imports that no longer exist.

**`src/aleph/vm/supervisor/networking_db.py`** (line 183): Minor: the `legacy == target` guard uses raw `Path` equality without `.resolve()`. If an operator configures `EXECUTION_DATABASE` and `SUPERVISOR_DATABASE` to the same file via different path strings (e.g. one containing `..`), the guard won't catch it. SQLite would handle this without corruption (the source is opened `mode=ro` and the target is a separate connection to the same file), so this is a theoretical edge case rather than a real bug. Using `legacy.resolve() == target.resolve()` would be more robust.

PR #1011, opened 2026-06-24
chore(db): drop the vestigial agent-DB port_mappings table (gated on the supervisor copy)

odesenfans:
Phase 2 of the agent/supervisor DB split (#1010, shipped in 2.0.0): drop the now-unused `port_mappings` table from the agent DB. Since the split, port mappings live in `supervisor.sqlite3`; the supervisor daemon (Python and Rust alike) copies the legacy rows out of this table once, on its first start.

## Why this no longer needs a deployment-ordering gate

The original plan was to hold this until every node had run the copy. That is not observable on a network of third-party CRNs, operators skip releases, and the copy only runs when the daemon starts cleanly while the agent's Alembic runs regardless.

The migration now checks the precondition itself, using the same condition that makes the copy a permanent no-op:

- the supervisor store holds at least one `port_mappings` row (the copy ran, or new mappings exist, and the daemon will never read the legacy table again), or
- the legacy table has no active rows left (nothing the copy would pick up).

When neither holds, `upgrade()` raises before any DDL. The agent exits non-zero, systemd restarts it after 10 s (`Restart=on-failure`), and the next start re-checks. So an agent that comes up before its daemon has done the copy waits for it instead of destroying live VMs' host-port forwards. The table is left alone when `EXECUTION_DATABASE` and `SUPERVISOR_DATABASE` point at the same file, since it is then the live store.

Net effect: safe to merge at any time, on any upgrade path, including a node jumping from 1.15 straight to the release carrying this.

Behaviour change worth knowing: on a node whose daemon cannot start at all, the agent now fails startup too (until the daemon has copied) instead of starting without forwards. A daemon-less agent cannot manage VMs anyway, and the error names the daemon.

## What it does

- Alembic revision `f1a2b3c4d5e6` on top of the `port_mappings`-creating `a1b2c3d4e5f6`: gated drop as above, `downgrade()` recreates the canonical 0004 schema.
- `tests/agent/test_drop_port_mappings_migration.py` drives the real agent Alembic runner against throwaway SQLite files: refuses on absent supervisor DB, refuses on empty supervisor table, drops once the store has rows, drops when only soft-deleted or no legacy rows remain, keeps the table on a shared file, and round-trips upgrade/downgrade to an identical `sqlite_master`.

## Merge criteria

- [x] #1010 merged and released (2.0.0)
- [x] Rebased onto `main` (2026-09-03)
- [x] Drop gated in code on the supervisor copy having happened
- [ ] Alembic head chain re-checked against other pending agent-DB migrations: #932 (rescue mode) adds revision `b3c4d5e6f7a8` with the same `down_revision = a1b2c3d4e5f6`. Whichever merges second must re-chain its `down_revision`, or `alembic upgrade head` fails at agent startup with multiple heads.
- [ ] Un-draft and merge


odesenfans:

The two-phase gating logic in the PR description is sound, and the existence-guard + downgrade are more care than most drop migrations get. But as the branch stands (cut from dev, without #1010), the migration would not actually stick, and there are two coordination hazards worth handling now since this PR will sit open for a while.

**Must address before un-drafting**

1. **Rebase onto #1010 is required for the drop to stick, not just for safety.** On this branch `PortMapping` is still registered on the agent `Base.metadata` (`metrics.py:137`), and `cli.py:405-407` runs `metrics.create_tables` (metadata.create_all) before Alembic on every startup; `supervisor/daemon.py:43` also runs `create_tables` with no migrations at all. So the first restart after `f1a2b3c4d5e6` is stamped silently recreates an empty `port_mappings` table, and since the revision stays stamped, the drop never re-runs: the vestigial table is permanently resurrected. After the #1010 rebase (model moved out of the agent Base) this resolves itself, but it means the merge criteria should include "rebased onto the post-split tree", not just "split deployed everywhere".

2. **Two-heads collision with the rescue-mode branch.** `origin/aliel-add-rescue-mode` adds a migration with `revision b3c4d5e6f7a8, down_revision a1b2c3d4e5f6`, the same parent as this one. If both land in the agent versions directory, `alembic.command.upgrade(cfg, "head")` at agent startup fails with "Multiple head revisions". Whichever merges second must re-chain its `down_revision` (that branch needs a rebase for the orchestrator -> agent rename anyway, but worth flagging in both PRs so it is not discovered on a node).

**Worth considering (altitude)**

3. The only thing standing between this migration and destroyed live port forwards is the docstring plus the HOLD label; labels do not gate merges or releases. The upgrade guard checks that the table exists but not the actual precondition (that the supervisor copy has run). Making `upgrade()` verify the precondition in code, e.g. skip (or refuse) the drop unless `settings.SUPERVISOR_DATABASE` exists and contains `port_mappings` rows for the active legacy rows being dropped, would make the migration safe to merge at any time and turn the deployment-ordering requirement into a mechanical check instead of a process promise.

**Minor**

4. `downgrade()` recreates the table via raw DDL with `AUTOINCREMENT`, a schema that never existed: 0004's `op.create_table` and the ORM model emit a plain `INTEGER NOT NULL, PRIMARY KEY (id)` (SQLAlchemy only emits AUTOINCREMENT with `sqlite_autoincrement=True`). Reusing 0004's `op.create_table` form keeps one canonical schema.
5. The `op.drop_index` before `op.drop_table` is redundant on SQLite (indexes drop with the table), and the guard checks the table but not the index, so a table-without-index state would abort mid-migration for a statement that was never needed.
6. Dead `logging` import + `logger` assignment, copied from 0004 where the logger is actually used.

One non-finding worth recording: the inline `create_engine(make_db_url())` + `Inspector.from_engine` pattern looks like it should raise `MissingGreenlet` on an async URL, but it was verified empirically that the agent runs Alembic inside `conn.run_sync(...)` (cli.py:320-323), whose greenlet context makes the sync inspection of an aiosqlite engine work. All shipped agent migrations share the pattern; it only breaks under a bare `alembic upgrade head` CLI run, a pre-existing fragility, not something this PR introduces.

PR #1012, opened 2026-07-03
feat(split)!: finish the agent/supervisor split (gRPC-only, total import separation)

odesenfans:
## What

Finishes the agent/supervisor split (items 1-3 of the roadmap). Plan: `docs/plans/2026-07-03-supervisor-split-endgame.md` (committed on this branch).

**Phase A: the supervisor no longer imports the agent.** The `models -> agent.metrics` residual turned out to be only the `EXECUTION_LOG_ENABLED` debug dump post-#1009/#1010 (supervisor state, moved into `models.py`); the aggregate-settings machinery (`update_aggregate_settings`, `AsyncTTLCache`, `RetrySession`, `get_compatible_gpus`) moved to neutral `aleph/vm/utils/`, with agent-only policy helpers staying in `agent/utils.py`.

**Phase B: gRPC is the only supervisor mode (breaking).** All eight candidate fixes on the old `od/grpc-only-supervisor*` WIP branches were verified already merged via the #981 squash (tree-identical). `build_supervisor` always returns `GrpcSupervisor`; `SUPERVISOR_GRPC_SOCKET` now defaults to `EXECUTION_ROOT/supervisor.sock` (exactly what packaging and droplet CI already pin); the agent no longer constructs a `VmPool` (`_engine_pool`, agent-side reattach hooks, in-process stop/drain branches removed). `LocalSupervisor` remains as the daemon's engine. The benchmark/fake-instance harness moved to `aleph.vm.testing.harness`, lazy-imported by the CLI.

**Phase C: total import separation.** The wire client and generated stubs moved to the contract side (`supervisor_interface/wire/`, `supervisor_interface/client.py`); `STREAM_CHAINS` moved to neutral `aleph/vm/chains.py` so `conf.py` is agent-free. import-linter now enforces three total contracts with an EMPTY ignore list: the contract layer imports neither side; `aleph.vm.agent` never imports `supervisor`/`pool`/`network`; the supervisor side (`supervisor`, `pool`, `models`, `network`, `hypervisors`) never imports `agent` or `testing`.

[...]

PR #1013, opened 2026-07-03
refactor(gpu): annotate GPU inventory agent-side; supervisor goes fully network-blind

odesenfans:
Follow-up to #1012 (now merged; rebased onto dev).

## Why

The supervisor consulted the settings aggregate for exactly one thing: annotating its lspci GPU inventory with the network's model name and `compatible` flag. Allocation never depended on it (GPU requests are matched by `device_id`; incompatible cards were inventoried and reservable all the same), so this was pure display annotation for `/about/usage/system`, and the supervisor's last remaining reason to talk to the CCN API.

## What

- **Supervisor reports raw hardware.** `parse_gpu_device_info` no longer looks up the aggregate: `model` stays unset and `compatible` false. `pool.setup()` no longer refreshes the aggregate.
- **The agent annotates.** `_gpus_from_host_info` (behind `/about/usage/system`) fills in `model` / `compatible` from the aggregate's `compatible_gpus` whitelist. Whitelist updates now apply on the agent's 60 s cache refresh instead of requiring a supervisor restart.
- **The aggregate module moves into the agent package** (`aleph.vm.utils.aggregate` → `aleph.vm.agent.aggregate`). Every consumer is agent-side now, so the existing import-linter contract ("the supervisor side does not import the agent") statically guarantees the supervisor never talks to the CCN API. The supervisor is fully network-blind.

## Behavior notes

- The wire format is unchanged: `gpu_inventory` / `available_gpus` already ride `GetHostInfo` as plain dicts; only who fills `model`/`compatible` changes.
- `VmInfo.gpus[].model` (attached-GPU info) is now empty; the agent only uses `info.gpus` as a has-GPUs gate, and QEMU passthrough only consumes `pci_host` / `x-vga`.
- Covered by the existing end-to-end test (`test_system_usage_gpu_ressources`: mocked lspci + mocked aggregate → annotated usage payload) plus new unit tests for the annotation itself (whitelisted vs unlisted device, unreachable aggregate).

### DIFF
--- a/src/aleph/vm/agent/aggregate.py
+++ b/src/aleph/vm/agent/aggregate.py
@@ -2,12 +2,22 @@ from logging import getLogger
 from typing import Any, TypedDict
 
 import aiohttp
+from pydantic import BaseModel, Field
 
 from aleph.vm.conf import settings
 from aleph.vm.utils.cache import AsyncTTLCache
 from aleph.vm.utils.http import get_session
 
 logger = getLogger(__name__)
+
+
+class CompatibleGPU(BaseModel):
+    """Shape of one ``compatible_gpus`` entry in the settings aggregate."""
+
+    vendor: str = Field(description="GPU vendor name")
+    model: str = Field(description="GPU model name")
+    name: str = Field(description="GPU full name")
+    device_id: str = Field(description="GPU device id code including vendor_id and model_id")
 
 
 class AggregateSettingsDict(TypedDict):
--- a/src/aleph/vm/agent/aggregate.py
+++ b/src/aleph/vm/agent/aggregate.py
@@ -2,7 +2,7 @@ from logging import getLogger
 from typing import Any, TypedDict
 
 import aiohttp
-from pydantic import BaseModel, Field
+from pydantic import BaseModel, Field, ValidationError
 
 from aleph.vm.conf import settings
 from aleph.vm.utils.cache import AsyncTTLCache
@@ -63,12 +63,22 @@ async def update_aggregate_settings() -> None:
     await get_aggregate_settings()
 
 
-def get_compatible_gpus() -> list[Any]:
-    """Return compatible GPUs from the cached settings aggregate."""
+def get_compatible_gpus() -> list[CompatibleGPU]:
+    """Return the validated GPU whitelist from the cached settings aggregate.
+
+    The aggregate is remote data: a malformed entry is skipped with a warning
+    rather than propagating (a raise here would take down every consumer,
+    e.g. the public usage endpoint) and older aggregates may omit the key."""
     cached = _settings_cache.get("settings")
     if not cached:
         return []
-    return cached["compatible_gpus"]
+    gpus = []
+    for entry in cached.get("compatible_gpus") or []:
+        try:
+            gpus.append(CompatibleGPU.model_validate(entry))
+        except ValidationError:
+            logger.warning(f"Skipping malformed compatible_gpus aggregate entry: {entry!r}")
+    return gpus
 
 
 async def get_user_aggregate(addr: str, keys_arg: list[str]) -> dict:
--- a/src/aleph/vm/agent/aggregate.py
+++ b/src/aleph/vm/agent/aggregate.py
@@ -21,7 +21,9 @@ class CompatibleGPU(BaseModel):
 
 
 class AggregateSettingsDict(TypedDict):
-    compatible_gpus: list[Any]
+    # Expected entry shape is CompatibleGPU; get_compatible_gpus validates
+    # each entry and skips violations, so this stays a raw-payload type.
+    compatible_gpus: list[dict[str, Any]]
     community_wallet_address: str
     community_wallet_timestamp: int
     # Optional in practice (older aggregates omit it); accessed via .get().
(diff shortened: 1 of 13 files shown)


foxpatch-aleph (approved):
Clean refactor that moves GPU network annotation (model name + compatible flag) from the supervisor to the agent, making the supervisor fully network-blind. The logic is correct: the supervisor reports raw hardware (model=None, compatible=False), and the agent annotates from the settings aggregate whitelist using a dict-merge that properly overrides the raw values. All imports are updated consistently, no stale references to the old `aleph.vm.utils.aggregate` path remain in source code, and the import-linter contract (supervisor side does not import agent) is now satisfied since pool.py and resources.py no longer pull in the aggregate. Test coverage is solid with new unit tests for whitelisted, unlisted, and missing-aggregate scenarios, plus the existing end-to-end test was updated with the correct mock path. The only minor observation is that the `CompatibleGPU` Pydantic model is dead code (defined but never used), but this is pre-existing rather than introduced by this PR.

**`src/aleph/vm/agent/aggregate.py`** (line 14): The `CompatibleGPU` model is defined but never imported or used anywhere in the codebase. It was already dead in its previous location (`resources.py`) and remains dead after the move. Consider removing it in a follow-up cleanup, though this is pre-existing and non-blocking.

foxpatch-aleph (approved):
Well-executed refactor that makes the supervisor fully network-blind by moving GPU annotation (model name + compatible flag) from the supervisor to the agent side. The logic is correct, all import paths are updated consistently, the wire format is unchanged, and the import-linter contract now statically guarantees the separation. New unit tests cover the annotation logic (whitelisted, unlisted, unreachable aggregate), and the existing end-to-end test continues to work. The only observations are minor: the `CompatibleGPU` pydantic model in `agent/aggregate.py` is dead code (never imported or used, pre-existing from the old location), and the module-level `_settings_cache` singleton is not cleared between tests (pre-existing test isolation issue, not introduced by this PR).

**`src/aleph/vm/agent/aggregate.py`** (line 14): The `CompatibleGPU` model is defined here but never imported or used anywhere in the codebase. It was also unused in its previous location (`resources.py`). Consider removing it, or if it's intended as documentation of the aggregate's shape, a comment or `TypedDict` would be lighter weight than an unused pydantic model.

foxpatch-aleph (approved):
Clean, well-executed refactor that moves network annotation responsibility from the supervisor to the agent. The module relocation (utils.aggregate → agent.aggregate) is properly propagated to all consumers, and the import-linter contract now statically guarantees the supervisor stays network-blind. The GpuDevice/AnnotatedGpuDevice split is sound: the supervisor reports raw hardware via model_dump() (no model/compatible keys), and the agent merges them in via dict union before validating as the richer subclass. The pool's zip(spec.gpus, resolved_devices, strict=True) is safe because _resolve_spec_gpus returns exactly one device per request or raises. The model echo change (request.model instead of device.model) is correct since the supervisor no longer has network-derived model names. Tests cover the key scenarios: whitelisted annotation, unlisted device, and missing aggregate fallback. No correctness, security, or quality issues found.

**`src/aleph/vm/agent/aggregate.py`** (line 14): The `CompatibleGPU` model is defined but never used anywhere in the codebase — `get_compatible_gpus()` returns `list[Any]` (raw dicts), not `list[CompatibleGPU]`. This was already dead code before the PR (previously in resources.py); consider removing it in a follow-up cleanup.

foxpatch-aleph (approved):
Well-executed refactor that cleanly moves network-dependent GPU annotation from the supervisor to the agent, making the supervisor fully network-blind. The import-linter contract now statically enforces the boundary. The code is correct: the zip with strict=True guards against index misalignment, the AnnotatedGpuDevice merge respects extra="forbid", and the async/await propagation is proper. Test coverage is good, covering whitelisted/unlisted devices, missing aggregate, and malformed entries. The only nit is that AggregateSettingsDict.compatible_gpus is still typed list[Any] when it could be tightened now that get_compatible_gpus validates and returns list[CompatibleGPU].

**`src/aleph/vm/agent/aggregate.py`** (line 24): Minor: `compatible_gpus: list[Any]` could be tightened to `list[dict[str, Any]]` now that `get_compatible_gpus` validates entries against `CompatibleGPU` before returning them. The `list[Any]` type no longer reflects how the field is consumed.

foxpatch-aleph (approved):
Clean, well-motivated refactor that makes the supervisor fully network-blind by moving aggregate annotation to the agent side. The import-linter contract now statically guarantees the supervisor never imports the agent. The `CompatibleGPU` pydantic validation with skip-on-malformed is a robustness improvement over the previous raw-dict access. The `_resolve_spec_gpus` ordering guarantee that justifies `zip(strict=True)` holds (sequential append in request order). Tests cover the new annotation logic thoroughly (whitelisted, unlisted, missing aggregate, malformed entries). The `model` field type difference between `GpuSpec` (str) and `HostGPU` (str | None) is handled correctly via `or None` coercion. No correctness, security, or testing gaps found.

**`src/aleph/vm/pool.py`** (line 393): Minor inconsistency: `GpuSpec.model` is `str` (default `""`) while `HostGPU.model` is `str | None` (default `None`). Line 393 passes `request.model` (str) to GpuSpec, line 403 coerces with `request.model or None` for HostGPU. This is correct given each type's contract, but a reader may briefly wonder why the two assignments differ. Not worth changing.

**`src/aleph/vm/agent/resources.py`** (line 135): `update_aggregate_settings()` is called on every `/about/usage/system` request. This is safe (60s TTL cache makes it a no-op when fresh), but worth noting that if the endpoint ever becomes high-traffic, the cache check is a synchronous dict lookup on `_settings_cache` — not a concern today.

PR #1014, opened 2026-07-03
fix(supervisor): survive settings renames in on-disk controller configs (1.13.0 upgrade blocker)

odesenfans:
Fixes the release blocker found in the 1.13.0 → dev gap analysis: **an in-place package upgrade of a node with live VMs crash-loops the supervisor daemon.**

## Mechanism

- Every `<vm_hash>-controller.json` embeds a full `Settings` dump from the aleph-vm version that created the VM, and `Settings` validates with `extra=forbid`.
- 1.13.0 wrote `CONNECTIVITY_DNS_HOSTNAME`; #974 renamed it to `CONNECTIVITY_DNS_HOSTNAMES` (the example.org probe was flaky, replaced by a dual-stack fallback list). The old key is now an unknown extra → `ValidationError`.
- `load_persistent_executions` parses controller configs unguarded, and the daemon intentionally aborts startup on an unparseable config (`daemon.py`: a pool that does not reflect what is running must not serve gRPC). Upgrade a 1.13.0 node with live VMs → daemon crash-loop, no gRPC socket, node unmanageable; a controller restart re-parses the same JSON and kills its VM.

## Fix

A `mode="before"` validator on `Configuration.settings` drops keys the running version does not know, logging a warning that names them. A renamed setting falls back to its current default. The guard is structural: any future settings rename/removal survives the upgrade, not just this one. Known keys with invalid values still fail loudly; leniency applies only to unknown keys.

Covered by tests: 1.13.0-style dump with the legacy key parses (and the renamed setting takes its default), an arbitrary future unknown key parses, and a corrupt value for a known key is still rejected.

## Not in this PR

- Rollback (dev → 1.13.0) remains broken for VMs reconfigured under dev: 1.13.0's parser is strict and cannot be fixed retroactively.
- A 1.13.0→candidate upgrade job in droplet CI (current CI only tests fresh installs, which is why this slipped through) is deliberately deferred.

### DIFF
--- a/src/aleph/vm/supervisor_interface/configuration.py
+++ b/src/aleph/vm/supervisor_interface/configuration.py
@@ -3,7 +3,13 @@ from enum import Enum
 from pathlib import Path
 from typing import Annotated
 
-from pydantic import BaseModel, BeforeValidator, ConfigDict, PlainSerializer
+from pydantic import (
+    BaseModel,
+    BeforeValidator,
+    ConfigDict,
+    PlainSerializer,
+    field_validator,
+)
 
 from aleph.vm.conf import Settings, settings
 from aleph.vm.sizes import MiB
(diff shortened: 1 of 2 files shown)


foxpatch-aleph (approved):
The fix is correct, well-scoped, and properly tested. The `field_validator(mode="before")` on `Configuration.settings` strips unknown keys before `Settings` (which uses pydantic-settings' default `extra=forbid`) can reject them, preventing the supervisor daemon crash-loop on upgrade. The guard is structural — it handles any future settings rename/removal, not just the `CONNECTIVITY_DNS_HOSTNAME` → `CONNECTIVITY_DNS_HOSTNAMES` case. Known keys with invalid values still fail loudly because the validator only drops keys not in `Settings.model_fields`. The three tests cover the renamed-key scenario, an arbitrary future unknown key, and the important negative case (corrupt value for a known key is still rejected). The non-dict early return is a correct safeguard for programmatic construction with an existing `Settings` instance.

**`tests/supervisor/test_controller_config_forward_compat.py`** (line 45): Minor: the real load path uses `Configuration.model_validate_json(f.read())` (JSON string), but tests use `Configuration.model_validate(dict)`. The `mode="before"` validator runs in both paths so the tests are still valid, but using `model_validate_json(json.dumps(...))` would exercise the exact deserialization path that `load_controller_configuration` uses, catching any future serialization-specific quirk.

PR #1015, opened 2026-07-03
refactor: post-split code-quality follow-ups

odesenfans:
Code-quality follow-ups for residual debt from the agent/supervisor split.

## 1. `aleph.vm.migration` moved under the agent package

The migration package is agent-side machinery (its runner drives message-free VM re-creation through `agent.run` / `agent.messages` / `agent.translate`), yet it lived outside `aleph.vm.agent` and was deliberately left out of the import-linter source lists, so its upward imports into the agent escaped the boundary checks. It now lives at `aleph.vm.agent.migration` and is covered by the existing agent contract, with zero ignore entries.

The one supervisor-side edge in the package, `helpers.graceful_shutdown` (imported `QemuVmClient` from `aleph.vm.supervisor.controllers`), turned out to be dead code with no callers anywhere; it was deleted rather than carried as a boundary exemption.

## 2. Dead cross-boundary exception branch removed

`agent/run.py` caught `MicroVMFailedInitError` from the hypervisors package. That exception is raised only inside the supervisor process and never crosses the gRPC boundary: the daemon's error mapping translates it to `ErrorCode.MICROVM_INIT_FAILED`, reconstructed client-side as `supervisor_errors.MicroVMInitError` (which the same branch already handled). The dead isinstance arm and its import are gone.

Note: this was not the sole agent import of `aleph.vm.hypervisors`; `agent.vm.program_client` still imports the `RuntimeConfiguration` leaf class. The import-linter comment now names that remaining edge precisely.

## 3. Stale dev scripts and docs

The hatch default-env scripts invoked `aleph-vm orchestrator run` / `aleph-vm orchestrator config` / `aleph-vm controller run`, subcommands the CLI does not define (`aleph.vm.agent.cli:main` is a flat argparse CLI; the daemon and controllers are module entry points). They are replaced with `agent`, `supervisor`, and `config` scripts that invoke what exists. Docs that taught the removed commands or linked to the pre-split `src/aleph/vm/orchestrator/` path were updated (README.md, tutorials/TESTING.md, the agent README including the never-existing `--print-config` flag, vm_connector/README.md, the QEMU controller notes). The agent README's run section now shows the two-process reality: start `python3 -m aleph.vm.supervisor`, then `python3 -m aleph.vm.agent`. Historical design docs in docs/plans were left untouched.

## 4. Naming drift

argparse advertised `prog="orchestrator"` and described the process as "VM Supervisor"; the installed console script is `aleph-vm` and the process is the agent. Fixed. The `vm_hash` / `vm_id` usages in the split code were audited: `vm_hash` names consistently hold `ItemHash` message keys with explicit `VmId(...)` conversions at the Supervisor-ABC boundary, so no local renames were warranted (and no field/wire renames were attempted, by design).

## 5. Fire-and-forget task convention

The reap hooks in `setup_webapp` dropped program guest state with a bare `asyncio.create_task`: a failing `ProgramGuestClient.forget` only surfaced through asyncio's GC-time "Task exception was never retrieved" handler. It now uses the `create_task_log_exceptions` convention helper, with a test asserting the failure is logged deliberately.

## 6. Event-watcher reconnect gap

The agent's supervisor event watcher reconnected after a stream drop but never compensated for events lost while disconnected: a VM that went down during the gap kept its agent-side state (idle-teardown timer, update subscription, guest API process). On reconnect it now re-lists the VMs over the existing `ListVms` RPC and applies the same idempotent drop a missed event would have triggered to every tracked VM that is now STOPPED, FAILED, or gone. A failed reconcile (daemon still down) is retried on the next attempt. Covered by tests for both the reconcile and the retry path.

**Explicitly out of scope:** the instant between the re-list and the stream re-subscribe remains uncovered (versus the entire outage before). Closing that window fully needs ordering guarantees from the supervisor (e.g. a subscribe-then-snapshot handshake or event sequence numbers on the WatchEvents RPC) and is a design decision, not a follow-up fix.


foxpatch-aleph (commented):
Well-structured code-quality follow-up covering six independent areas. The migration package move is complete with no stale references, the dead exception branch removal is correct, the event-watcher reconcile logic is sound with proper retry semantics, and the fire-and-forget task convention is properly applied with test coverage. One minor nit: a dead import remains.

**`src/aleph/vm/agent/migration/helpers.py`** (line 7): `import time` is now unused. The only function that used it (`graceful_shutdown`) was deleted in this PR. `asyncio` is still used by `compress_disk`/`rebase_overlay`/etc., but `time` has no remaining call sites in this file.

foxpatch-aleph (approved):
A well-structured code-quality PR that correctly moves the migration package under the agent tree, removes a dead cross-boundary exception branch, fixes stale dev scripts and docs, and adds a solid event-stream gap reconciliation mechanism. The reconcile logic is correct: the `stream_interrupted` flag ensures reconcile runs only after an actual interruption, the retry path when reconcile itself fails is properly handled, and the `_drop_vm_state` extraction makes both the event handler and reconcile path share the same idempotent drop. Tests cover both the happy path and the retry scenario. The only nit is a leftover unused `import time` in helpers.py after the `graceful_shutdown` deletion.

**`src/aleph/vm/agent/migration/helpers.py`** (line 7): Unused import: `time` was only used by the deleted `graceful_shutdown` function. Safe to remove.

PR #1016, opened 2026-07-03
refactor(supervisor)!: controller configs embed only the controller's settings slice

odesenfans:
Follow-up to #1014, removing the design smell it papered over: controller configs embedded a full ~200-field `Settings` dump, of which the controller process ever read **nine** (`JAILER_BASE_DIR` + the eight network fields consumed by `supervisor/controllers/__main__.py`). The full dump coupled the on-disk format to every settings rename (the 1.13.0 `CONNECTIVITY_DNS_HOSTNAME` crash-loop) and shipped ~190 stale values as dead weight in per-VM state.

## What

- **`ControllerSettings`**: a new model in `supervisor_interface/configuration.py` carrying exactly the slice the controller reads, with `extra="ignore"` and per-field defaults mirroring `conf.py`. `Configuration.settings` is now this type.
- **No file migration needed, in either direction of schema drift**:
  - configs written by ANY older version (full dumps, legacy keys like `CONNECTIVITY_DNS_HOSTNAME` included) parse by ignoring the surplus;
  - a key missing from an old dump, or dropped from the slice in the future, falls back to its default;
  - files rewrite themselves to the minimal form on the VM's next reconfigure.
- **Writers unchanged**: `from_attributes=True` lets the three `Configuration(...)` build sites keep passing the live `Settings` object; pydantic extracts the slice.
- **Controller `__main__`** no longer calls `Settings.check()` (full-host validation: KVM, firecracker binaries, fake-data paths — meaningless against a stale snapshot, redundant with the supervisor's own startup check). It validates the one field `Network` cannot default (`NETWORK_INTERFACE`) with a clear error, and `--print-settings` prints the actual slice.
- **Supersedes #1014's unknown-key validator**: the smaller model's `extra="ignore"` absorbs it; the forward-compat test file now pins both directions (1.13.0-style full dump, arbitrary future key, missing keys → defaults, corrupt known key still rejected, writer round-trip is minimal).

## Compatibility

- Reading: any old config parses (verified against a real 1.13.0-style `Settings` dump with the legacy key in tests).
- Rollback: a node rolled back to a version older than #1014 cannot parse minimal configs, same asymmetry #1014 already documented; rollback to a #1014-or-later version works (its validator drops the unknown... nothing: minimal is a strict subset, and missing keys were never rejected — the full-`Settings` model has defaults for all nine fields).
- Marked breaking (`!`) for the on-disk format change, though no operator action is required.

### DIFF
--- a/src/aleph/vm/supervisor/controllers/__main__.py
+++ b/src/aleph/vm/supervisor/controllers/__main__.py
@@ -123,9 +123,13 @@ def main():
     )
 
     if args.print_settings:
-        print(config.settings.display())
+        print(config.settings.model_dump_json(indent=4))
 
-    config.settings.check()
+    # The config predates this process (written at VM creation, possibly by
+    # an older aleph-vm): validate the one field Network cannot default.
+    if not config.settings.NETWORK_INTERFACE:
+        logger.error("Controller config %s carries no NETWORK_INTERFACE", config_path)
+        sys.exit(1)
 
     network = Network(
         vm_ipv4_address_pool_range=config.settings.IPV4_ADDRESS_POOL,
--- a/src/aleph/vm/supervisor_interface/configuration.py
+++ b/src/aleph/vm/supervisor_interface/configuration.py
@@ -152,11 +152,7 @@ def save_controller_configuration(vm_hash: str, configuration: Configuration) ->
     """Save VM configuration to be used by the controller service"""
     config_file_path = get_controller_configuration_path(vm_hash)
     with config_file_path.open("w") as controller_config_file:
-        controller_config_file.write(
-            configuration.model_dump_json(
-                by_alias=True, exclude_none=True, indent=4, exclude={"settings": {"USE_DEVELOPER_SSH_KEYS"}}
-            )
-        )
+        controller_config_file.write(configuration.model_dump_json(by_alias=True, exclude_none=True, indent=4))
     config_file_path.chmod(0o644)
     return config_file_path
 
--- a/src/aleph/vm/supervisor_interface/configuration.py
+++ b/src/aleph/vm/supervisor_interface/configuration.py
@@ -3,15 +3,9 @@ from enum import Enum
 from pathlib import Path
 from typing import Annotated
 
-from pydantic import (
-    BaseModel,
-    BeforeValidator,
-    ConfigDict,
-    PlainSerializer,
-    field_validator,
-)
+from pydantic import BaseModel, BeforeValidator, ConfigDict, PlainSerializer
 
-from aleph.vm.conf import Settings, settings
+from aleph.vm.conf import IPv6AllocationPolicy, settings
 from aleph.vm.sizes import MiB
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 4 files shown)


foxpatch-aleph (approved):
A clean, well-justified refactor that replaces the full ~200-field Settings dump in controller configs with a 9-field ControllerSettings slice. The extra="ignore" + from_attributes=True design elegantly solves both forward and backward compatibility without file migration. Defaults mirror conf.py exactly, all config.settings.* accesses in the controller are covered by the slice, and removing Settings.check() (host-level validation against a stale snapshot) in favor of a single NETWORK_INTERFACE check is sound. Test coverage pins both directions of schema drift. The only nit is a stale USE_DEVELOPER_SSH_KEYS exclude in save_controller_configuration that is now dead code (harmless but should be cleaned up).

**`src/aleph/vm/supervisor_interface/configuration.py`** (line 157): The `exclude={"settings": {"USE_DEVELOPER_SSH_KEYS"}}` is now stale dead code: `ControllerSettings` no longer has that field (it lived on the full `Settings`). Pydantic v2 silently ignores non-existent keys in `exclude`, so this doesn't break, but it should be removed for clarity since the PR's whole point is that the on-disk format now carries only the controller slice.

foxpatch-aleph (commented):
The core refactor is well-designed: ControllerSettings with extra=ignore and from_attributes=True elegantly solves both forward and backward compatibility, and the NETWORK_INTERFACE guard is a sensible replacement for the heavy Settings.check(). The forward-compat test coverage is thorough. One stale leftover needs cleanup: save_controller_configuration still passes exclude={"settings": {"USE_DEVELOPER_SSH_KEYS"}} which is now dead code since ControllerSettings has no such field — pydantic v2 silently ignores it, but it's misleading and should be removed. The new test file for __main__.py is good but could be strengthened by asserting the error message content and by adding a round-trip test through save_controller_configuration.

**`src/aleph/vm/supervisor_interface/configuration.py`** (line 157): Stale exclude: USE_DEVELOPER_SSH_KEYS no longer exists on ControllerSettings (it was on the old full Settings model). Pydantic v2 silently ignores non-existent exclude paths, so this doesn't break, but it's dead code left over from the pre-refactor model. Remove the exclude= parameter entirely (or at least drop USE_DEVELOPER_SSH_KEYS from it) to avoid confusion.

**`tests/supervisor/test_controller_main.py`** (line 38): Consider asserting the error message content, not just the exit code. For example, use caplog to verify the logger.error message mentions NETWORK_INTERFACE, so a future regression that exits(1) for a different reason doesn't pass this test silently.

**`tests/supervisor/test_controller_config_forward_compat.py`** (line 85): test_writers_pass_the_live_settings_object checks model_dump() but doesn't exercise save_controller_configuration (the actual on-disk writer with the stale exclude= parameter). Adding a round-trip through save_controller_configuration + load_controller_configuration would catch any serialization issue introduced by the exclude= leftover and confirm the minimal-form rewrite claim in the PR description.

odesenfans:
Addressed the review nit in 5eb7cdcc: removed the stale `exclude={"settings": {"USE_DEVELOPER_SSH_KEYS"}}` from `save_controller_configuration` (a no-op since `Configuration.settings` became `ControllerSettings`). The slice-only dump is already pinned by `test_writers_pass_the_live_settings_object` in `test_controller_config_forward_compat.py`.

foxpatch-aleph (approved):
This PR is a clean, well-reasoned refactor that replaces the full ~200-field Settings dump in controller configs with a minimal ControllerSettings model carrying exactly the 9 fields the controller process reads. The approach is sound: extra="ignore" absorbs surplus keys from old configs (no file migration), from_attributes=True lets the three Configuration build sites keep passing the live Settings object, and defaults mirror conf.py so missing keys fall back gracefully. The removal of Settings.check() from __main__.py is correct — those checks (KVM, firecracker binaries, fake-data paths) are meaningless against a stale snapshot and redundant with the supervisor's own startup validation; the targeted NETWORK_INTERFACE guard is the right replacement. The USE_DEVELOPER_SSH_KEYS exclusion removal is safe since that field is not in ControllerSettings. Test coverage is thorough, pinning both directions of schema drift (1.13.0-style full dump, arbitrary future keys, missing keys → defaults, corrupt known key still rejected, writer round-trip is minimal). No correctness, security, or quality issues found.

**`src/aleph/vm/supervisor/controllers/__main__.py`** (line 126): The --print-settings output format changed from the old key=value display() format to JSON (model_dump_json). This is a user-visible change for operators using --print-settings diagnostically. The JSON format is arguably better, but worth noting in the PR description or changelog if not already.

**`src/aleph/vm/supervisor/controllers/__main__.py`** (line 115): Pre-existing nit (not changed in this PR): uses builtin exit(1) instead of sys.exit(1), inconsistent with line 132 which uses sys.exit(1). Both raise SystemExit so behavior is identical, but the inconsistency could confuse readers.

foxpatch-aleph (approved):
This PR cleanly replaces the full ~200-field Settings dump in controller configs with a minimal ControllerSettings slice carrying exactly the nine fields the controller process reads. The design is sound: extra="ignore" absorbs surplus keys from old configs (no file migration), from_attributes=True lets the three existing Configuration(...) build sites keep passing the live Settings object, and defaults mirror conf.py so missing keys fall back. The controller's new NETWORK_INTERFACE guard is the right replacement for the old full Settings.check() — it validates the one field Network() cannot default, without re-running host-level checks (KVM binaries, firecracker paths) that are meaningless against a stale snapshot. Tests cover both directions of schema drift (1.13.0-style full dump, arbitrary future key, missing keys → defaults, corrupt known key still rejected, writer round-trip is minimal) and the controller entry point (missing NETWORK_INTERFACE, --print-settings, missing file). The load_persistent_executions path remains intentionally unguarded for systemic parse failures, which is the documented design. No bugs or security issues found.

**`src/aleph/vm/supervisor/controllers/__main__.py`** (line 115): Minor consistency nit (pre-existing): this uses `exit(1)` (the REPL builtin) while the new guard at line 132 uses `sys.exit(1)`. Both raise SystemExit so behavior is identical, but mixing the two in the same function is slightly inconsistent. Not blocking.

**`src/aleph/vm/supervisor/controllers/__main__.py`** (line 125): Behavioral note: `--print-settings` runs before the `NETWORK_INTERFACE` guard, so on a config with no NETWORK_INTERFACE the slice is printed AND the process exits 1. This seems intentional (print is for debugging, the config is still invalid) but worth confirming it matches operator expectations for the `--print-settings` flag.

**`src/aleph/vm/supervisor_interface/configuration.py`** (line 111): Consider documenting that these defaults must stay in sync with conf.py (or better, derive them from Settings.model_fields at class definition time). A future drift between the two would silently change controller behavior for any VM whose config omits the key. The test_writers_pass_the_live_settings_object test pins the field set but not the default values.

PR #1017, opened 2026-07-03
ci(droplets): stop apt-daily before provisioning instead of racing its lock

odesenfans:
Three droplet-CI flakes today (on #1014, #1015 and #1016) with the identical signature:

```
E: Could not get lock /var/lib/apt/lists/lock. It is held by process NNNN (apt-get)
```

on fresh Ubuntu 24.04 droplets, in the "Install Aleph-VM on the Droplet" step — despite every apt-get already passing `-o DPkg::Lock::Timeout=-1`. That option reliably guards the dpkg locks but empirically not the lists lock, and the existing `lsof` wait loop is check-then-run: the boot-time `apt-daily` timer can re-fire between the check and the next ssh command.

Fix: remove the contender instead of racing it.

- `systemctl disable --now apt-daily.timer apt-daily-upgrade.timer` and stop any in-flight `unattended-upgrades` / `apt-daily` run before the first apt command (fine on a throwaway CI droplet).
- Extend the wait loop to cover both `/var/lib/apt/lists/lock` and `/var/lib/dpkg/lock-frontend`, so a run already holding a lock is waited out.
- Apply the same treatment to the fake-instance (debian-12) job, which had no wait at all.

Fail-open by design: every mitigation command is `|| true`, so on an image without these units the behavior is exactly today's.

### DIFF
--- a/.github/workflows/build-deb-package-and-integration-tests.yml
+++ b/.github/workflows/build-deb-package-and-integration-tests.yml
@@ -223,8 +223,14 @@ jobs:
           # Wait a few seconds for DigitalOcean to setup the Droplet using apt, which conflicts with our comands:
           sleep 5
 
-          # Wait for /var/lib/apt/lists/lock to be unlocked on the remote host via SSH.
-          while ssh root@${DROPLET_IPV4} lsof /var/lib/apt/lists/lock; do sleep 1; done
+          # Fresh droplets run apt-daily / unattended-upgrades at boot, which
+          # grabs /var/lib/apt/lists/lock. DPkg::Lock::Timeout does not cover
+          # that lock reliably, and a check-then-run wait loop races (the
+          # timer can re-fire between the check and our apt-get). Stop the
+          # contender for good, then wait out any run already in flight.
+          ssh root@${DROPLET_IPV4} "systemctl disable --now apt-daily.timer apt-daily-upgrade.timer 2>/dev/null || true"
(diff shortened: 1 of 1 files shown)


foxpatch-aleph (approved):
Well-reasoned and correctly targeted fix for a real flake. Disabling the apt-daily timers eliminates the race root cause, while the extended lsof wait loop (now covering both /var/lib/apt/lists/lock and /var/lib/dpkg/lock-frontend) handles any in-flight apt process that systemctl stop won't kill. The fail-open design (|| true, 2>/dev/null) ensures no regression on images lacking these units. Comments clearly explain the rationale. The pre-existing lack of a timeout on the while loop and lsof availability assumptions are not introduced by this PR.

foxpatch-aleph (approved):
The PR correctly addresses a recurring CI flake caused by apt-daily/unattended-upgrades holding the apt lists lock on fresh Ubuntu droplets. The approach of disabling the timers and stopping in-flight services before waiting out any residual lock holders is sound and fail-open by design (all mitigation commands use `|| true`). The `||` between the two `lsof` calls is inside the remote shell quotes, so it correctly implements OR semantics: the while loop continues as long as either lock is held and only exits when both are free. Applying the same mitigation to the Debian 12 job (which previously had no wait at all) is a clear improvement. The job-level `timeout-minutes` bounds the unbounded wait loop as a safety net.

**`.github/workflows/build-deb-package-and-integration-tests.yml`** (line 224): The `sleep 5` is now largely redundant since the `systemctl` stop/disable commands and the `lsof` wait loop that follow will handle the timing. It's harmless, but could be removed in a future cleanup to shave a few seconds off every run.

PR #1019, opened 2026-07-06
Proto and dead-code cleanup ahead of the Rust supervisor port

odesenfans:
Pre-increment-1 cleanup from the Rust supervisor daemon design (docs/plans/2026-07-04-rust-supervisor-daemon-design.md): the Rust contract and port scope should be born clean, without historical holes or dead code to carry over.

## Proto: drop reserved fields, compact VmInfo numbering

There is no wire protocol to stay backward compatible with: both stubs regenerate from the same file in the same deb. All five `reserved` blocks are gone (`BACKEND_QEMU_SEV`, `DiskConfig.mount`, the VmInfo network/gateway pairs and `is_instance`) and VmInfo fields are renumbered 1-20 without gaps. `ReserveResourcesRequest.is_instance` stays: a reservation precedes the VM and carries no `Backend`, so the memory-accounting bucket cannot be derived and the client must state it.

## Delete the dead snapshot subsystem

Every concrete controller sets `support_snapshot = False` and `SNAPSHOT_FREQUENCY` defaults to 0, making the whole subsystem unreachable: `snapshot_manager.py`, `snapshots.py`, the pool/models wiring, the `SNAPSHOT_*` settings and the storage helpers. The devmapper snapshot in `storage.py` is a different mechanism and stays.

## Delete `program.py`

`AlephProgramResources` only survived in a `models.py` type union (nothing instantiates it since the spec path took over), and its legacy `FileTooLargeError` is raised nowhere, which made the `error_mapping.py` branch for it unreachable. The guest-protocol re-exports already live in `aleph.vm.program_config`.

## Validation

- `tests/supervisor`: 995 passed (the 4 failures in `test_interfaces.py`/`test_log.py` are environmental and fail identically on unmodified dev)
- import-linter: 3 contracts kept, 0 broken
- `ruff format` and `isort` clean
- proto stubs regenerated with `scripts/generate_proto.py`


foxpatch-aleph (approved):
Clean, well-scoped dead-code cleanup. All three deletion targets (snapshot subsystem, program.py, proto reserved fields) are fully removed with no dangling references in src/ or tests/. The VmInfo renumbering is gap-free (1-20), ReserveResourcesRequest.is_instance is correctly retained with an updated comment, and the wire-protocol FileTooLargeError/ERROR_CODE_FILE_TOO_LARGE are correctly preserved. The PR also silently fixes a type-union bug in models.py (AlephQemuConfidentialInstance -> AlephQemuConfidentialResources in the resources field). Test updates are consistent across all files. The only nit is the now-unused `schedule==1.2.1` dependency left in pyproject.toml.

**`pyproject.toml`** (line 60): The `schedule==1.2.1` dependency is now unused — its only consumer (`SnapshotManager`) was deleted in this PR. Consider removing it from `dependencies` to avoid carrying a dead dependency into the Rust port baseline.

PR #1020, opened 2026-07-06
Move resource reservations agent-side

odesenfans:
Second pre-increment-1 cleanup for the Rust supervisor port (after #1019): the two-bucket memory policy and the GPU reservation ledger are client policy expressed in client vocabulary (instance vs program), so they move to the agent. The supervisor keeps mechanism invariants only. The Rust daemon now never ports a reservation ledger, bucket accounting, or the vocabulary.

## Agent side (new `agent/capacity.py`)

`CapacityManager`, wired as `app["capacity"]` next to the registry:

- The two-bucket `check_capacity` ported verbatim; committed memory/vCPU sums come from `AgentVmRegistry` (each record holds the full message, rehydrated on restart). The create paths pass `exclude_vm_hash` so a VM's own early registry record (confidential owner-auth) or a leftover record on recreate does not count against its own request.
- GPU holds keyed by concrete `pci_host`, 60s TTL, lazy expiry: another user's hold blocks a card, your own hold is consumed at create. `/control/reserve_resources` keeps byte-identical HTTP behavior.
- Create paths run capacity + GPU resolution after the download (a failed download never consumes a hold) and put resolved cards (`pci_host` + `supports_x_vga`) on the spec.

## Supervisor side (mechanism backstops only)

`check_capacity`, `check_spec_admission`, `reserve_gpus`, the `Reservation` ledger and `owner_id` handling are deleted. Both create paths enforce, under `creation_lock`:

- `_check_memory_backstop`: committed + requested must fit physical memory minus `HOST_MEMORY_RESERVED_MIB`. No buckets, no vCPU check (overcommit is agent policy).
- `_validate_spec_gpus`: every `pci_host` must exist in the inventory and not be attached to any VM (nor claimed twice in one spec).

## Proto

`ReserveResources` (and with it the `is_instance` field), `VmSpec.owner_id` and `GpuConfig`'s request-side fields (`device_id`, `model`) are gone; `GpuConfig` is now always a resolved card. Stubs regenerated.

## Notes for review

- GPU resolution used to happen inside the supervisor's `creation_lock`, atomic with registration. Now there is a small window between agent-side resolve and supervisor-side attach where two concurrent creates requesting the same `device_id` could pick the same card; the supervisor backstop turns that into a clean `InsufficientResourcesError` for the loser instead of a double-attach.
- The agent computes available disk as `shutil.disk_usage(PERSISTENT_VOLUMES_DIR).free`: the pool's per-execution reserved-delta machinery always contributed 0 for spec-built VMs, so the figure is identical in practice.
- GPU flows have no droplet CI coverage (no hardware); the ledger and backstops are unit-tested (24 new tests in `test_agent_capacity.py` plus rewritten admission/GPU-create suites), but a manual check on a GPU node before the next release is advisable.

## Validation

- `tests/supervisor` + `tests/migration`: 1032 passed (baseline 995 + 24 capacity tests + rewritten suites); only the 4 known environmental failures (pyroute2 needs root, journald stub)
- import-linter: 3 contracts kept, 0 broken
- `ruff format` and `isort` clean; each of the 3 commits builds green individually


foxpatch-aleph (approved):
Well-structured refactor that cleanly moves admission policy (two-bucket memory accounting, vCPU overcommit, GPU reservation ledger) from the supervisor/pool to a new agent-side CapacityManager, leaving the supervisor with only physical-memory and GPU-double-attach backstops. The new capacity.py is well-documented and mirrors the old pool logic faithfully. The proto changes (dropping ReserveResources, owner_id, and GPU request fields) are consistently propagated through the wire types, converters, gRPC stubs, and tests. The 24 new tests in test_agent_capacity.py plus rewritten admission/GPU-create suites provide good coverage. The acknowledged race window between agent-side GPU resolution and supervisor-side attach is correctly handled by the supervisor backstop turning the loser's create into a clean InsufficientResourcesError. No correctness, security, or quality issues found.

**`src/aleph/vm/agent/capacity.py`** (line 101): check_capacity is a sync method that reads self.registry without holding self._lock, while reserve_gpus/resolve_gpus acquire the lock. In the reserve endpoint this means there's a TOCTOU between the capacity dry-run and the GPU hold. This is acceptable (the check is an early-refusal gate, the supervisor backstop is authoritative), but a comment noting that check_capacity intentionally does not need the lock would make the design intent explicit.

**`src/aleph/vm/agent/capacity.py`** (line 310): The `available` dict in the InsufficientResourcesError uses `unheld_available` (the pre-hold-filtering snapshot), so a card held by another user would still appear as "available" in the error diagnostic. This is cosmetic and matches the old pool behavior, but could be mildly confusing to an operator reading the error. Consider filtering out held cards for the diagnostic, or just note this is pre-hold inventory.

foxpatch-aleph (approved):
Well-executed refactoring that moves client-policy (two-bucket memory accounting, vCPU overcommit, GPU reservation ledger) from the supervisor/pool to a new agent-side CapacityManager, while keeping only mechanism invariants (physical-memory backstop, GPU double-attach prevention) on the supervisor. The two-bucket check_capacity is ported verbatim with the correct exclude_vm_hash logic for the early-record pattern. The race window between agent-side GPU resolution and supervisor-side attach is explicitly documented and correctly handled by the supervisor's _validate_spec_gpus backstop (turning a double-pick into a clean InsufficientResourcesError). Proto, wire layer, ABC, and tests are all consistently updated. 24 new agent-capacity tests plus rewritten admission/GPU-create suites provide thorough coverage. Only minor nits: a dead `pool.reservations = {}` in one test fixture and a slightly misleading error message for in-spec double GPU claims.

**`tests/supervisor/test_supervisor_spec_persistent_program.py`** (line 51): Leftover dead code: `pool.reservations = {}` sets an attribute that no longer exists on `VmPool` (the `reservations` dict and `Reservation` class were deleted). `test_supervisor_spec_admission.py:93` even asserts `not hasattr(pool, "reservations")`. This line is harmless but should be removed for consistency.

**`src/aleph/vm/pool.py`** (line 1080): The error message `"GPU at pci_host ... is already attached to a VM"` is also raised for the in-spec double-claim case (same pci_host appears twice in one spec, caught by the `claimed` set). Consider a distinct message like `"GPU at pci_host ... is claimed twice in this spec"` for that branch to aid debugging.

**`src/aleph/vm/agent/capacity.py`** (line 294): The `unheld_available` snapshot (used for the error `available` field) is taken before any matching, so it shows cards that were already consumed by earlier requests in the same call. This is cosmetic (error path only) but could confuse debugging — consider using the remaining `available_gpus` list instead.

foxpatch-aleph (approved):
Clean, well-documented refactoring that moves client policy (two-bucket memory accounting, vCPU overcommit, user-scoped GPU holds) from the supervisor/pool to a new agent-side CapacityManager, while the supervisor retains only mechanism invariants (physical-memory backstop and GPU attach validation). The proto changes correctly remove ReserveResources, owner_id, and GpuConfig request-side fields. The documented race window between agent-side GPU resolve and supervisor-side attach is correctly handled by the _validate_spec_gpus backstop turning a double-pick into a clean InsufficientResourcesError. Test coverage is thorough: 24 new tests in test_agent_capacity.py plus rewritten admission and GPU-create suites. The code is correct, secure, and follows existing conventions.

**`src/aleph/vm/agent/capacity.py`** (line 314): The `available` dict in the InsufficientResourcesError from `_match_requests` reports only the GPUs remaining after earlier requests in the same call have already been matched and removed from the list. For a multi-GPU request where the second fails, the error's `available.gpus` won't include cards matched by the first. This is cosmetic (the error is still raised correctly) but could be slightly misleading in logs.

**`src/aleph/vm/agent/capacity.py`** (line 191): Minor style nit: the f-string split (`f"Disk: required ... MiB, " f"available ... MiB"`) is unnecessary — a single f-string would be cleaner. This was carried over from the original pool.py code, so it's not a regression.

PR #1021, opened 2026-07-06
feat(rust): supervisor daemon increment 1 (workspace, Health, GetHostInfo)

odesenfans:
Increment 1 of the Rust supervisor port (design doc: docs/plans/2026-07-04-rust-supervisor-daemon-design.md, RPC coverage table section 7).

## What's in

- **`rust/` cargo workspace** (stable 1.90, edition 2024): `supervisor-proto` (tonic-build on the frozen `proto/supervisor.proto`) and `supervisor-daemon` (binary `aleph-vm-supervisor`).
- **Health + GetHostInfo** at field-for-field parity with `LocalSupervisor`: lspci GPU inventory ported literally from `resources.py`, same HostInfo sources, same error-trailer encoding (`aleph-supervisor-error-bin`). Every other RPC returns UNIMPLEMENTED with the trailer the Python `_abort` sends.
- **Socket lifecycle** matching the Python daemon (mkdir parent, unlink stale, serve, unlink on exit) plus day-one hardening: bind under umask 0o077 + chmod 0700, identity-checked shutdown unlink (never removes a newer instance's socket), 5s shutdown grace matching `server.stop(grace=5)`, second-signal immediate exit.
- **One Python change** (`supervisor_interface/client.py`): pin `grpc.default_authority=localhost` on the UDS channel. gRPC C-core derives `:authority` from the percent-encoded socket path; hyper/h2 rejects it and resets every stream, so without this no gRPC client of ours could talk to a Rust daemon. Both daemons ignore the value.
- **Conformance suite seed** (`tests/conformance/`, opt-in via `ALEPH_VM_CONFORMANCE=1`): boots the Rust daemon, drives it with the production Python client, asserts parity against the same host sources, SIGTERM unlink included.
- **CI**: `test-rust.yml` (fmt, clippy `-D warnings`, `cargo test --locked`, conformance run), path-filtered to `rust/**`, `proto/**`, the conformance suite, and the Python modules the contract depends on.
- **`docs/plans/rust-port-divergences.md`**: the divergence ledger, 10 entries (umask hardening, identity-checked unlink, no dotenv autoload, ported warts kept bug-for-bug, increment staging notes).

## Review and gates

Three-lens adversarial review (parity vs Python, Rust robustness, CI/conformance): 19 findings, 15 fixed, 4 dismissed into the ledger. Parity spot-proofs: error trailer byte-identical, GPU JSON key order equals `model_dump`, `is_global` matched CPython `ipaddress` on a 200k-address sweep, pydantic bool spellings reproduced.

Gates: cargo fmt/clippy clean, 33 cargo tests, conformance 4/4, Python suite 1033 passed (only the 4 known env-only failures), ruff/isort/lint-imports/yamlfix clean.

## Known gaps (by design, increment staging)

Health.vm_count is 0 and available GPUs/disk ignore surviving VMs until the world view lands (increment 2); `settings.check()` preconditions arrive with the increments that need them. The daemon is not packaged or selectable yet: no deb changes, `ALEPH_VM_SUPERVISOR_IMPL` wiring comes with packaging.


foxpatch-aleph (approved):
Well-executed increment 1 of the Rust supervisor port. Health and GetHostInfo achieve field-for-field parity with the Python LocalSupervisor, verified through careful source comparison (lspci parsing, is_global_ipv4 CIDR table, MemTotal truncation, statvfs calculation). Socket hardening (umask/chmod/identity-checked unlink) is a sound improvement with proper documentation in the divergence ledger. All unsafe blocks have correct SAFETY justifications. The idempotent SocketGuard cleanup correctly handles the race between serve-path and second-signal shutdown. The grpc.default_authority client fix is the correct workaround for the hyper/h2 authority issue. The conformance suite and CI are well-structured. No correctness, security, or testing gaps found.

**`rust/crates/supervisor-daemon/src/config.rs`** (line 53): The debug log says "skipping non-UTF8 environment variable" even when the key is valid UTF-8 but only the value is non-UTF8 (the `(Ok(key), Err(_))` case hits this arm). Minor cosmetic mismatch in the log message, not a bug.

**`rust/crates/supervisor-daemon/src/net.rs`** (line 128): The explicit `address == Ipv4Addr::BROADCAST` check is redundant — 255.255.255.255 is already covered by the `in_block([240, 0, 0, 0], 4)` check. Harmless but unnecessary; the Python ipaddress module also has this as a separate case internally, so it documents the parity intent.

**`tests/conformance/test_rust_daemon.py`** (line 103): The inlined `default_route_interface()` faithfully mirrors the Python `conf.py` version including the unguarded `parts[1]` index, which would crash on an empty line. Since `/proc/net/route` always has a multi-column header and the function is test-only, this is safe in practice. Not a bug, just noting the deliberate parity.

foxpatch-aleph (approved):
Well-executed increment 1 of the Rust supervisor port. Health and GetHostInfo are field-for-field ports of the Python LocalSupervisor, with verified parity for CPU count, memory, uname, statvfs, lspci GPU inventory, and the is_global_ipv4 table (including the 192.0.0.9/.10 exceptions). The error trailer encoding (ErrorDetail with ErrorCode.INTERNAL for UNIMPLEMENTED) matches the Python _abort path exactly. The socket lifecycle (umask hardening, identity-checked unlink, 5s shutdown grace, second-signal fast exit) is robust with proper atomic guarantees. The one Python client change (pinning grpc.default_authority=localhost) is a necessary fix for hyper/h2 compatibility and is correctly scoped. Tests are comprehensive: 33 cargo unit tests, an in-crate gRPC integration test, and a 4-test conformance suite driving the Rust daemon with the production Python client. The divergence ledger is thorough and accurately documents 10 deliberate divergences. No bugs or security issues found.

**`rust/crates/supervisor-daemon/src/config.rs`** (line 53): The `tracing::debug!` call in `from_env()` runs before the tracing subscriber is initialized (it's called from `main()` at line 75, before `tracing_subscriber::fmt().init()` at line 63), so the message is silently dropped. Consider using `eprintln!` or moving the subscriber init before `Settings::from_env()`, or just removing the debug log since it's a rare path. Non-blocking.

**`tests/conformance/test_rust_daemon.py`** (line 72): The `socket_path.exists()` check only verifies file existence, not that it's a socket. The Rust integration test (`grpc.rs:53`) correctly uses `is_socket()`. In practice this works because `UnixListener::bind` atomically creates the socket, but using `stat.S_ISSOCK` would be more robust and match the Rust test's approach.

PR #1022, opened 2026-07-06
feat(rust): supervisor daemon increment 2 (read-only world adoption)

odesenfans:
Increment 2 of the Rust supervisor port, stacked on #1021 (design doc section 7, row 2). GitHub retargets this to dev when #1021 merges.

## What's in

- **World view adoption (steps 1-4 of the design doc's boot sequence)**: scan `{EXECUTION_ROOT}/*-controller.json` (sorted, keyed on the embedded `vm_hash` like Python, duplicate-vm_index guard ported from `claimed_vm_ids`, WARN-skip on unparseable/oversized/non-regular files), systemd unit states over zbus (5s method timeout, bus-unreachable vs unit-inactive taxonomy: a transient bus failure at boot defers to live queries instead of permanently stamping VMs stopped), port mappings from `supervisor.sqlite3` via rusqlite (busy timeout for the adoption window, Python's dict-fold semantics reproduced).
- **RPCs**: GetVm, GetVmSpec, ListVms, ListPortForwards, GetLogs (journalctl subprocess, same matches/order/timestamps as Python's sd-journal reader, Rust-only 10k-line server cap), plus world-aware Health.vm_count and GetHostInfo. GetVmSpec mirrors `spec_from_controller_configuration` field-for-field, asserted in conformance against the actual Python function.
- **Two Python changes**: `save_controller_configuration` now writes atomically (temp + `os.replace`; the in-place truncate made a config unreadable exactly during the daemon-handoff window), with tests.
- **Fixtures**: controller JSONs + sqlite generated by the real pydantic/SQLAlchemy models (`scripts/generate_rust_fixtures.py`, byte-stable regeneration), committed for hermetic cargo tests.
- **Ledger** grows to 17 entries; notable: adopted VMs' GPUs are excluded from availability where restarted Python wrongly re-offers them (a real Python bug, fix-in-python-later, standalone PR coming), config-without-unit reported STOPPED where Python destroys at startup, negative-vm_index configs hidden where Python silently serves the pool's last subnet.

## Review and gates

Three-lens adversarial review: 14 confirmed findings (dbus no-timeout hang, boot-time world-view poisoning, unbounded GetLogs memory, sqlite busy drop, filename-vs-embedded-hash keying, duplicate vm_index, union shape resolution, and more), 13 fixed, 1 deliberately not ported (ledgered). Parity spot-proofs: both IPv6 allocators pinned against the Python allocators, all 20 VmInfo fields across running/dead/confidential states, sqlite fold quirks, error trailer shape.

Gates: cargo fmt/clippy clean, 79 cargo tests (`--locked`), conformance 9/9, Python suite 1036 passed (only the 4 known env-only failures), ruff/isort/lint-imports/yamlfix clean.

## Known gaps (by design, increment staging)

Adoption step 5 (idempotent nftables/ndppd reconcile) and all mutations are increment 3. `settings.check()` preconditions arrive with the increments that need them. Packaging (deb ships the binary + `supervisor-launcher`) is the next stacked PR, unlocking the read-only swap scenario on the aleph-testnets `od/aleph-vm-upgrade-checks` branch.


foxpatch-aleph (approved):
This is a high-quality, well-documented increment that faithfully ports the Python daemon's read-only world view adoption (controller config parsing, systemd unit queries, sqlite port mappings, IP derivation) and the GetVm/GetVmSpec/ListVms/ListPortForwards/GetLogs RPCs to Rust. The code demonstrates careful attention to Python parity (field-for-field verification, dict-fold semantics, shape-based union resolution, microsecond timestamp mapping), security (symlink_metadata to prevent FIFO hangs, 1MiB file size cap, read-only sqlite, no shell in journalctl subprocess, atomic config writes), and graceful degradation (bus-unreachable vs unit-inactive taxonomy, 5s D-Bus timeout, per-VM error isolation). All deliberate divergences are explicitly ledgered with clear rationale. Test coverage is thorough: 79 cargo tests, 9 conformance tests using the actual Python models as oracle, and Python tests for the atomic write change. The minor observations (conformance test not disabling networking, sequential now_ns calls, JSON whitespace) are all pre-existing patterns or explicitly ledgered divergences, not bugs.

**`tests/conformance/test_rust_daemon_read_only.py`** (line 168): The `adopted_daemon` fixture calls `start_daemon(tmp_path)` without passing `ALEPH_VM_ALLOW_VM_NETWORKING=false`. Since the Rust default is `true` (matching conf.py), the daemon will attempt to auto-detect the network interface and resolve its IPv4 at boot. This is the same pattern as the increment 1 conformance test, but could fail on CI runners without a default route. Consider explicitly disabling networking for this read-only world test: `start_daemon(tmp_path, {"ALEPH_VM_ALLOW_VM_NETWORKING": "false"})`.

**`rust/crates/supervisor-daemon/src/world.rs`** (line 224): `now_ns()` is called separately for `defined_at_ns` (line 224), `preparing_at_ns` (232), `prepared_at_ns` (233), and `started_at_ns` (293). Each call advances the clock slightly, so the timestamps are monotonically increasing but may differ by only nanoseconds. This matches the Python behavior (sequential `datetime.now()` calls during reattach) and is semantically correct, but worth noting for anyone reading the code who might expect a single adoption instant.

foxpatch-aleph (approved):
This is a high-quality increment that faithfully ports the Python supervisor's read-only world view to Rust. The adoption logic (config scanning, systemd unit queries, sqlite port mappings, IP derivation) is carefully mirrored with parity spot-proofs, and deliberate divergences are thoroughly documented in the ledger. Security is handled well: symlink_metadata before opening files prevents FIFO/symlink hangs, the 1 MiB config cap prevents memory exhaustion, D-Bus calls have a 5s timeout, journalctl output is bounded at the source, and the sqlite connection is opened read-only with a busy timeout. The atomic write change to save_controller_configuration is correct and well-tested. Test coverage is extensive across unit, integration, and conformance layers with hermetic fixtures generated from the real Python models. No blocking issues found.

**`rust/crates/supervisor-daemon/src/world.rs`** (line 439): When `prefix - pool_len` equals 64 (e.g. a /0 pool split into /64 subnets), `1u64 << 64` is UB in debug / wraps to 0 in release, making `subnet_count = 0` and the `index >= subnet_count` check always true (rejecting everything). Python's `subnets()` enumeration handles this correctly. In practice the default pool is /12 so this is unreachable, but a guard like `prefix - pool_len >= 63` returning an error would make the safety explicit.

**`rust/crates/supervisor-daemon/src/world.rs`** (line 447): If `prefix == 0` (pool_len also 0, index must be 0), `subnet_size = 1u64 << 32` which is fine in u64, but `network + 2 > broadcast` where `broadcast = 0 + 2^32 - 1` passes the check and `Ipv4Addr::from(network as u32)` truncates to 0.0.0.0. This is a degenerate edge case that Python's strict IPv4Network would also handle oddly, so it's not a practical concern — just noting for completeness.

**`rust/crates/supervisor-daemon/src/world.rs`** (line 175): Minor: `controller_units()` is called unconditionally after the `active_states` failure check, but if the bus is unreachable it will fail too and the `if let Ok(...)` silently drops the error. Consider logging the bus failure here at debug level so an operator can see both adoption-time bus problems in one place. Not blocking — the `active_states` failure already logs a WARN.

**`rust/crates/supervisor-daemon/src/service.rs`** (line 261): The `awaiting_confidential_init` formula checks `confidential.is_some() && times.started_at_ns != 0 && times.stopping_at_ns == 0 && !running`. For a confidential VM adopted under a bus failure (running == None at adoption, then live-queried as inactive), `started_at_ns` is 0 so this is correctly false. For a confidential VM adopted running then seen dead, `started_at_ns != 0` and `!running` makes it true. The test `confidential_mode_follows_the_sev_policy_bit` covers this. Good coverage.

**`rust/crates/supervisor-daemon/src/units.rs`** (line 133): The `unreachable!` after the for loop is fine since the loop returns on attempt 1's second pass, but a reader unfamiliar with the control flow might miss that `attempt` only goes 0..2. Consider a comment or restructuring to make the two-pass guarantee more obvious, e.g. `for attempt in 0..=1`. Very minor nit.

foxpatch-aleph (approved):
A meticulously crafted increment that ports the Python supervisor daemon's read-only world view to Rust with field-for-field parity. The adoption logic (disk scan, systemd unit states, sqlite port mappings, IP derivation) faithfully mirrors the Python oracle, with deliberate divergences clearly documented in the ledger. Defensive coding is excellent: non-regular files are stat-checked before opening, a FIFO can't hang boot, D-Bus has a 5s timeout, bus failures degrade gracefully instead of stamping permanent STOPPED, and GetLogs bounds subprocess output at the source. The atomic save_controller_configuration change is correct and well-tested. Test coverage spans unit tests (cargo), integration tests over a real Unix socket, and Python conformance tests asserting against the actual Python models. No security or correctness issues found.

**`rust/crates/supervisor-daemon/src/world.rs`** (line 164): When the system bus is unreachable, `active_states` already failed (line 149), yet `controller_units()` is called unconditionally here, producing a second redundant D-Bus call with a 5s timeout in production. Consider gating on `active_states.is_some()` or moving this block inside the `Ok` branch above to avoid the wasted timeout when the bus is down.

**`rust/crates/supervisor-daemon/src/service.rs`** (line 524): `get_vm_spec` holds the world read lock while calling `vm_spec_message`. This is safe today (pure computation, no I/O), but inconsistent with `get_vm` and `list_vms` which clone-then-release before external calls. If `vm_spec_message` ever gains a dependency on external state in a future increment, the held lock could become a problem. Consider cloning the entry out for consistency.

**`rust/crates/supervisor-daemon/src/service.rs`** (line 676): When `from_tail=true` with `max_lines > 0`, journalctl already returns only the last `max_lines` entries (via `-n max_lines` from `journal_tail_bound`), so the post-parse `lines.drain(..lines.len() - max_lines)` is always a no-op. A comment noting this redundancy is intentional (defense-in-depth) would help future readers.

**`rust/crates/supervisor-daemon/src/world.rs`** (line 224): `now_ns()` is called separately for `defined_at_ns`, `preparing_at_ns`, `prepared_at_ns`, and `started_at_ns` (lines 224, 232, 233, 293), so these timestamps can differ by microseconds within one VM adoption. This matches Python's per-call `time.time()` behavior and is semantically correct, but capturing a single `adopted_at = now_ns()` and reusing it would be cleaner and slightly more efficient.

foxpatch-aleph (approved):
Excellent Rust port of the read-only world view with meticulous Python parity. Every RPC handler, the IP allocator math, the systemd unit-state taxonomy, the controller config parser (including the pydantic smart-union shape resolution), the sqlite port-mapping fold, the journalctl log source, and the atomic save_controller_configuration fix are all correct. The divergences are deliberate, well-documented in the ledger, and improve on the Python oracle (GPU exclusion, bus-unreachable graceful degradation, memory-bounded GetLogs, non-destructive stopped-VM reporting). Test coverage is comprehensive: 79 cargo tests, 9 conformance tests using the actual Python `spec_from_controller_configuration` as oracle, and Python tests for the atomic write. The two minor nits (StaticUnitStates filter consistency, theoretical /0 pool edge case) are non-blocking.

**`rust/crates/supervisor-daemon/src/units.rs`** (line 215): Minor nit: `StaticUnitStates::controller_units()` filters by `CONTROLLER_UNIT_PREFIX` only, while the production `ZbusUnitStates::controller_units()` also checks `.ends_with(".service")` (line 169). Harmless since all test unit names are created by `controller_unit_name` which always appends `.service`, but the inconsistency could bite a future test that adds a non-.service unit name.

PR #1023, opened 2026-07-06
fix(supervisor): rebuild GPU attachments when reattaching running VMs

odesenfans:
After a supervisor restart, `load_persistent_executions` reattaches running VMs but never rebuilt `VmExecution.gpus` from the controller config. Two consumers were wrong afterwards:

- `get_available_gpus` re-offered cards physically passed through to running guests, so `GetHostInfo.available_gpus` lied after every daemon restart.
- `_validate_spec_gpus` was blind to adopted VMs, so a new CreateVm could be handed a card a running VM owns (real double-attach harm).

The restore path now rebuilds attachments via `_rebuild_reattached_gpus`: inventory cards keep full device metadata; a card absent from the inventory (removed, or vfio binding changed while the supervisor was down) is recorded as a bare HostGPU from the config alone, with a warning, invisible to availability and admission (which only deal in inventory cards) but still held by `uses_gpu()`.

Found while building the Rust port's world view (its adoption already implements this exclusion; `docs/plans/rust-port-divergences.md` entry 14 on the increment branches tracks this as the matching Python fix).

Known limitation, out of scope: `build_qemu_confidential_configuration` persists `QemuGPU(pci_host=...)` only, so a confidential VM's 3D-controller card that is also missing from the inventory rebuilds with `supports_x_vga=True` from the pydantic default. When the card is present, hardware truth wins.

Tests: `tests/supervisor/test_supervisor_reattach.py` round-trips real Configuration models through disk and runs the real restore; asserts exclusion from availability, rejection at admission, and the inventory-missing variant. Full suite: 1035 passed, only the 4 known env-only failures.

### DIFF
--- a/src/aleph/vm/pool.py
+++ b/src/aleph/vm/pool.py
@@ -769,6 +769,12 @@ class VmPool:
         spec = spec_from_controller_configuration(config)
         execution = VmExecution.from_spec(spec, systemd_manager=self.systemd_manager)
 
+        # Rebuild GPU attachments from the config. Without this, an adopted
+        # VM's cards look free: get_available_gpus re-offers them and
+        # _validate_spec_gpus lets a new create double-attach a card the
+        # running guest physically holds.
+        execution.gpus = self._rebuild_reattached_gpus(spec.gpus)
+
         execution.mapped_ports = await get_port_mappings(vm_id)
         logger.info("Loading existing mapped_ports %s", execution.mapped_ports)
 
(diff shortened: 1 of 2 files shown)


foxpatch-aleph (approved):
The fix correctly addresses a real bug: reattached VMs had empty `gpus` lists after a supervisor restart, causing `get_available_gpus` to re-offer cards physically held by running guests and `_validate_spec_gpus` to allow double-attach. The new `_rebuild_reattached_gpus` method mirrors the inventory-card construction in `_validate_spec_gpus`, gracefully handles cards absent from inventory (recording them from config alone, invisible to availability/admission but held by `uses_gpu()`), and the GPU rebuild runs before the execution is added to `self.executions`. Tests cover both paths with real config round-tripping. The known confidential-VM limitation is honestly scoped out.

PR #1024, opened 2026-07-06
feat(packaging): ship the Rust supervisor daemon in the deb behind a launcher

odesenfans:
Packaging step of the Rust supervisor port (design doc section 9). The deb now ships both daemons; which one runs is a one-line env change, not a packaging event.

## What's in

- **`packaging/supervisor-launcher`**: POSIX sh dispatch script. `aleph-vm-supervisor.service` execs it; it picks the implementation from `ALEPH_VM_SUPERVISOR_IMPL` in `/etc/aleph-vm/supervisor.env` (default `python`, unknown values log and fall back to python so an operator typo cannot brick a node).
- **Makefile**: new `debian-package-rust` target builds `rust/` with `cargo build --release --locked` and installs `/opt/aleph-vm/bin/aleph-vm-supervisor` plus the launcher.
- **Build containers**: rustup with the `rust-toolchain.toml` pin instead of the distro cargo (plain cargo ignores the pin; jammy's rustc predates edition 2024), protoc for tonic-build (upstream 21.12 on jammy, whose apt 3.12 predates proto3 `optional`), toolchain preinstalled in its own layer before the sources.
- **CI deb-content check** asserts both new files are in every distro's package.
- Ledger entry 14 flipped: the Python GPU reattach fix landed (#1023), the remaining gap (rebuilding `VmInfo.gpus` for adopted running VMs) moves to the Rust side in increment 3.

## Proof

- Full container package builds run locally for ubuntu-22.04 and debian-12; the debian-12 deb ships the 11 MB Rust binary and the launcher.
- The packaged binary and the launcher dispatch were exercised on a clean `debian:12` container: `--help` works, and the daemon boots for real (adopts an empty world, binds its socket, serves gRPC).
- debian-13 and ubuntu-24.04 builds are covered by this PR's own CI battery (dev-targeted, so the four-distro deb build runs here).

## What this unlocks

Testnets upgrade-check scenario B: a 1.13→2.0 upgrade followed by `ALEPH_VM_SUPERVISOR_IMPL=rust`, restart, and the read-only assertions against the Rust daemon.


foxpatch-aleph (approved):
Well-structured packaging PR that ships the Rust supervisor daemon alongside the Python one behind a POSIX sh launcher. The dispatch script correctly uses exec, handles unknown ALEPH_VM_SUPERVISOR_IMPL values by falling back to Python (operator typo safety), and the Makefile dependency chain ensures the rust target runs after the code target wipes the install tree. Dockerfiles correctly preinstall the pinned toolchain via rustup in a separate layer for cache efficiency, and the CI deb-content check now asserts both new binaries are present. No correctness or security issues found. Minor pre-existing note: Architecture: all in the control file is technically incorrect for a package containing native x86_64 binaries, but this predates the PR (firecracker, sevctl are also native).

**`packaging/Makefile`** (line 3): debian-package-code is listed both in the aggregate target and as a dependency of debian-package-rust. This is harmless (Make deduplicates), but the explicit listing in the aggregate target is redundant since debian-package-rust already depends on it.

**`packaging/aleph-vm/DEBIAN/control`** (line 3): Architecture: all is technically incorrect for a package shipping a native x86_64 Rust binary (and pre-existing native binaries like firecracker/sevctl). This is a pre-existing issue not introduced by this PR, but worth noting for future cleanup — should be amd64.

**`packaging/ubuntu-22.04.dockerfile`** (line 17): The protoc download URL is hardcoded to linux-x86_64, which is consistent with the rest of the build assuming x86_64, but means this dockerfile won't work on arm64 without modification. Not a blocker since the entire packaging pipeline is x86_64-only (sevctl also uses --target x86_64-unknown-linux-gnu).

PR #1025, opened 2026-07-06
fix(packaging): declare the deb Architecture as amd64

odesenfans:
Follow-up from #1024's review. The package has shipped native x86_64 binaries for a long time (firecracker, jailer, sevctl, and now the Rust supervisor daemon) while declaring `Architecture: all`, so dpkg would happily install it on arm64 where nothing runs. One line: `all` becomes `amd64`.

No artifact ripple: the deb filename is set explicitly by `dpkg-deb --build aleph-vm target/aleph-vm.deb` and the per-distro renames in CI, so nothing downstream keys off an architecture suffix. The only behavior change is dpkg now refusing to install on non-amd64 hosts, which is correct.

### DIFF
--- a/packaging/aleph-vm/DEBIAN/control
+++ b/packaging/aleph-vm/DEBIAN/control
@@ -1,6 +1,6 @@
 Package: aleph-vm
 Version: 0.1.8
-Architecture: all
+Architecture: amd64
 Maintainer: Aleph.im
 Description: Aleph.im VM execution engine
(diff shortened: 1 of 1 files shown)


foxpatch-aleph (approved):
Correct one-line fix changing the deb control file from Architecture: all to Architecture: amd64. The package ships native x86_64 binaries (firecracker, jailer, sevctl built with --target x86_64-unknown-linux-gnu, and vmlinux.bin), so Architecture: all was incorrect and allowed dpkg to install on non-amd64 hosts where nothing would run. The reprepro repository configs already declare Architectures: amd64 source, so this aligns the control file with the existing repo configuration. The deb filename is set explicitly via dpkg-deb --build and per-distro CI renames, so no downstream artifact naming depends on the architecture suffix. No security, correctness, or testing concerns.

PR #1026, opened 2026-07-07
feat(rust): supervisor daemon increment 3 (persistent lifecycle and nftables)

odesenfans:
Increment 3 of the Rust supervisor port (design doc section 7, row 3): the persistent QEMU lifecycle, host networking mutations, and adoption step 5.

## What's in

- **nftables engine**: firewall.py ported rule-for-rule as pure functions (ruleset in, JSON command batches out), pinned byte-for-byte against captures of the actual Python module (12 fixture scenarios incl. both-protocols ordering); `nft -j` apply behind a seam.
- **Host mutation seams**: systemd unit mutations over zbus, tap create/delete (ip(8), Python's exact error tolerance), ndppd conf rendering with the 0.5s debounced restart, port-mapping writes with SQLAlchemy-byte-identical DDL and datetime rendering, and the legacy `EXECUTION_DATABASE` port-mapping migration (a 1.13 node flipped straight to rust keeps its forwards).
- **RPCs**: CreateVm (idempotency, readopt-of-live-untracked, memory backstop, GPU validation, cloud-init seed byte-compatible with the Python builders, pydantic-identical atomic config write, ready wait with crash-loop recheck), StopVm, StartVm, RebootVm, ReinstallVm, DeleteVm, Add/RemovePortForward, RecreateNetwork, boot-time nftables/ndppd reconcile (adoption step 5), settings check slice, and the GPU rebuild for VMs adopted running (closes ledger entry 14 via #1023).
- **Integration matrix**: `tests/integration/` now drives either daemon via `ALEPH_VM_SUPERVISOR_IMPL`; the "Supervisor integration tests" CI job runs a python/rust matrix, with real-kernel port-forward and RecreateNetwork coverage on both legs, and loud failures instead of silent skips when cargo//dev/kvm are missing in CI.

## Review

Three-lens adversarial review (parity vs the Python originals, Rust robustness, CI/test quality): 43 findings consolidated into 24 fixes (each with a regression test), 12 new ledger entries (now 36), 2 refuted with evidence. Highlights: a host-port allocator race between per-VM locks that could leak a live DNAT rule pointing at the wrong guest (now serialized through one net lock, matching the Python event loop's implicit serialization), RecreateNetwork flushing chains unserialized against concurrent creates, udp-before-tcp recreation order, insertion-order enumeration parity, the legacy DB migration, and a CI gate-break where the rust integration leg would have failed on WatchEvents (increment 4 surface).

## Gates

cargo fmt/clippy clean, 148 cargo tests (`--locked`), conformance 20/20 (fixtures generated by the real pydantic/SQLAlchemy/firewall.py code, drift-guarded both directions), ruff format/isort/mypy clean on touched Python. The rust integration leg runs first in this PR's CI.

## Known staging

WatchEvents/StreamLogs, ephemeral Firecracker programs, and the reattach retry loop are increment 4; backups increment 5; confidential mutations increment 6 (ledger entry 9).


foxpatch-aleph (approved):
This is a high-quality, well-structured increment that ports the persistent QEMU lifecycle, nftables engine, host networking mutations, and port-forward management from the Python supervisor to Rust. The locking model (creation_lock → vm_lock → net_lock) is correctly ordered with no deadlock cycles, and guards are never held across long systemd waits. The nftables pure/apply separation enables thorough fixture-based parity testing without root. Security-sensitive areas are handled correctly: subprocess invocations use argument lists (no shell injection), the unsafe bind_probe properly manages fd lifetimes, atomic config writes use temp-file+fsync+rename, and the vm_hash path traversal is explicitly acknowledged as a shared wart deferred for a coordinated fix (ledger entry 34). The saturating_add fix for the memory backstop correctly prevents integer overflow from hostile requests. Test coverage is comprehensive, including concurrency tests for the host-port allocator race, panic recovery in the boot path, byte-for-byte fixture parity against the Python firewall/cloudinit/config-writer code, and real end-to-end integration tests with nftables verification on both matrix legs. The divergence ledger (now 36 entries) thoroughly documents every deliberate difference from the Python oracle with clear dispositions.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 1679): Minor observation (non-blocking): in recreate_network, between step 4 (per-VM chains rebuilt under net_lock) and step 5 (reapply port redirects outside net_lock), a concurrent stop_vm could tear down the freshly-rebuilt chains for a VM before its redirects are reapplied. The net_lock serializes the individual nft operations, but the gap between steps 4 and 5 is unlocked. This matches Python's event-loop interleaving behavior and the consequences are limited (stale rules cleaned up on next stop/recreate), so it's acceptable for parity.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 870): Minor observation (non-blocking): in update_port_redirects, if nft_add_redirect or nft_remove_redirect fails mid-way through the protocol loop, the function returns an error with partially-applied nft rules but the in-memory forwards list stays at the pre-call snapshot (correct per ledger entry 29's snapshot semantics). The nft rules themselves are not rolled back. This matches the Python daemon's behavior where the store is only written on success but nft mutations are not transactional.

foxpatch-aleph (approved):
A large, well-executed increment that ports the persistent QEMU lifecycle, nftables engine, host networking mutations, and boot-time reconcile from Python to Rust. The code maintains careful parity with the Python oracle, with deliberate divergences clearly documented in the ledger. Lock discipline is sound: creation_lock -> vm_lock -> net_lock (innermost), with the net_lock correctly serializing host-port allocation through persistence to prevent the cross-VM DNAT race. The memory backstop uses saturating_add to prevent overflow. The bind probe correctly uses raw libc without SO_REUSEARRD to match Python's plain socket semantics. The nftables pure layer is cleanly separated from the apply layer, enabling hermetic testing against captured Python fixtures. Cloud-init JSON-as-YAML is a valid subset. The known vm_id path traversal is a shared wart (ledger entry 34), not introduced here. Test coverage is extensive with 148 cargo tests plus conformance and integration matrix coverage.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 726): The untracked-VM delete path reads the config file using `vm_id` directly in `controller_config_path`. If `vm_id` contains path separators (e.g. `../../etc/passwd-controller.json`), this could resolve outside EXECUTION_ROOT. This is the known shared wart ledgered as entry 34 (fix-in-both-later), not a regression introduced by this PR, but worth noting that the delete path here is more dangerous than the read path since it calls `remove_controller_configuration` which unlinks files derived from `vm_id`.

**`rust/crates/supervisor-daemon/src/ports.rs`** (line 452): The port allocation loop `for port in start..MAX_PORT` excludes `MAX_PORT` (65535). Python's `range(start_port, 65535)` has the same off-by-one, so this is correct parity, but port 65535 is never allocated by either daemon. Just a confirmation that this matches Python.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 1342): The `catch_unwind` with `AssertUnwindSafe` is sound here because the closure's state access is either re-read under locks after the closure returns or cleaned up by the error path. The comment explains the safety reasoning well.

PR #1028, opened 2026-07-07
feat(rust): supervisor daemon increments 5 and 6 (backups, confidential)

odesenfans:
The final two increments of the Rust supervisor port (design doc section 7, rows 5 and 6). With this, every RPC the Python daemon serves has a 1:1 Rust port behind the ALEPH_VM_SUPERVISOR_IMPL flag.

## Increment 5: backups and restore

StartBackup, GetBackupStatus, ListBackups, DownloadBackup, DeleteBackup, RestoreBackup, RestoreFromImage. New backup.rs (qemu-img orchestration behind a DiskTools seam, the tar + sha256 sidecar + meta.json archive, the in-memory job registry mirroring Python's _backup_jobs/_backup_tasks/_backup_locks, background run) and qmp.rs QGA client for the best-effort fs-freeze. Restore streams a fixed archive member to a caller-fixed destination (no unpack, no path-traversal sink). Un-skips tests/integration/test_backup_restore.py on the rust leg.

## Increment 6: confidential VM surface

Scope per the design doc (row 6, "parity with current Python behavior"): reproduce Python's mixed real/QMP confidential path 1:1, not leave it UNIMPLEMENTED. InitializeConfidential writes the session files and starts the unit (no hardware); GetMeasurement/InjectSecret are QMP passthrough (query-sev, query-sev-launch-measure, sev-inject-launch-secret, cont) with the SEV data path hardware-gated (Tier 2); confidential CreateVm builds the QemuConfidentialVMConfiguration and reports awaiting_confidential_init without starting the unit, mirroring Python. The ENABLE_CONFIDENTIAL_COMPUTING settings check is ported.

## Review

Three-lens adversarial review (parity, robustness, CI/tests). Confirmed the tar-extraction path is not a traversal sink. Fixed: two backup concurrency races Python's event loop prevents (DeleteVm-vs-backup disk lock, StartBackup TOCTOU), the SEV-policy parser now matches CPython int(x,0) exactly (pinned table), QMP write timeouts and a line cap, a DownloadBackup concurrency cap, and deterministic backup-job ordering. Test depth closed: the QMP fake now asserts the exact command bytes (with inject_secret/continue_execution coverage), the restore/corrupt-archive path gains hermetic tests, and a real Python-built backup archive is committed as a drift oracle. Ledger now 53 entries.

## Gates

cargo fmt/clippy clean, 241 cargo tests (`--locked`), conformance 25/25, ruff format/isort/mypy clean on touched Python, fixtures byte-stable. The rust integration leg runs the backup cycle on real KVM.

## Remaining UNIMPLEMENTED

Only persistent Firecracker programs (the agent recreates ephemeral ones on demand) and real SEV-SNP attestation (Phase 3, aleph-cvm). Everything the current Python daemon does is ported.


foxpatch-aleph (approved):
Well-structured, thoroughly tested implementation of the backup/restore and confidential VM surfaces. Security is solid: member-streamed tar extraction prevents path traversal, backup IDs are validated against directory escape, subprocess calls avoid shell injection, and QMP/QGA sockets have bounded timeouts and line caps. Concurrency races (StartBackup TOCTOU, DeleteVm-vs-backup) are properly guarded with an admission lock and per-VM disk locks. The SEV policy parser faithfully reproduces CPython's int(x, 0) grammar. Only minor nits: one dead public function and one unused variable.

**`rust/crates/supervisor-daemon/src/confidential.rs`** (line 120): `session_dir` is dead code — never called anywhere in the codebase. It duplicates the inline session directory construction in `initialize_confidential` (line 40-44) and `confidential_config_slice` in lifecycle.rs (line 1648-1652). Remove it, or use it to DRY up those three call sites.

**`rust/crates/supervisor-daemon/src/confidential.rs`** (line 72): `let _mode = confidential_mode(&entry);` computes the confidential mode but discards it (tee_backend is always hardcoded to SEV). If the value is truly unneeded, remove the binding; if it is meant for future use, add a `#[allow(unused)]` with a forward-reference comment.

foxpatch-aleph (approved):
A thorough and well-executed port of the backup/restore and confidential VM surfaces from Python to Rust. The backup registry's admission lock correctly prevents the StartBackup TOCTOU race, the tar extraction is safe against path traversal (member-streamed, never extracted by path), the restore rollback logic is correct, and the confidential CreateVm path properly avoids starting the unit. The parse_int_base0 implementation faithfully reproduces CPython's int(x, 0) grammar with a pinned test table. Lock ordering is consistent across all paths (vm_lock → backup disk lock → net_lock). QMP/QGA clients have appropriate read/write timeouts and line caps. The only issues are minor dead code: confidential::session_dir is pub but never called, and confidential::confidential_mode is computed but its result is unused (stored in _mode).

**`rust/crates/supervisor-daemon/src/confidential.rs`** (line 72): The `_mode` variable is computed but never used — `tee_backend` is hardcoded to `Sev` on line 76. Either remove the `confidential_mode` call (and the function at line 128 if it has no other callers) or use `_mode` to set `tee_backend` if that's the intended future behavior. Currently this is dead code.

**`rust/crates/supervisor-daemon/src/confidential.rs`** (line 120): `session_dir` is `pub` but never called outside this module. The lifecycle module constructs the session directory path inline (lifecycle.rs:1648-1652) rather than calling this function. Consider removing it or making it private to avoid a misleading public API.

PR #1029, opened 2026-07-07
docs(phase3): design + implementation plan for aleph-cvm SEV-SNP + NUMA backport

odesenfans:
## Phase 3: aleph-cvm SEV-SNP + NUMA backport (design + plan)

Stack root for Phase 3, which grafts aleph-cvm's confidential-computing (SEV-SNP)
and NUMA capabilities onto the now-complete Rust supervisor without breaking the
frozen gRPC contract, the public CRN HTTP API, the existing SEV/SEV-ES flow, or
zero-downtime live-VM adoption.

**Design** (`docs/plans/2026-07-07-phase3-aleph-cvm-backport-design.md`), key decisions:
- Foundation-first: Rust takes over the persistent-QEMU launch (faithful port of the
  Python controller argv + template-unit model), then SNP and NUMA graft on as argv
  fragments. The Python controller stays the parity oracle; adoption path unchanged.
- SNP is added alongside the existing SEV/SEV-ES launch-secret flow, backend-selected,
  zero breakage. Supervisor auto-places NUMA (pack-first, port of aleph-cvm's allocator).
- Measured-boot guest image via a backported Nix flake (own PR).
- Out of scope: TDX / NVIDIA-CC, public-API changes, encrypted/compose rootfs,
  in-protocol confidential scheduling (testnets validates SNP directly over gRPC).

**Plan** (`docs/plans/2026-07-07-phase3-aleph-cvm-implementation.md`): ten stacked PRs,
SNP front-loaded so the aleph-testnets attestation loop can start as early as possible,
NUMA following. Each increment is one PR with its own reviewer gate, implemented via the
Phase-2 port loop (subagent implement, 3-lens adversarial review, fix batch with a
regression test per finding, merge on green).

Subsequent increment PRs stack on this branch.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1030, opened 2026-07-07
feat(rust): A1 - Rust aleph-vm-controller, non-confidential QEMU argv parity

odesenfans:
## A1: Rust `aleph-vm-controller`, non-confidential QEMU argv parity

First foundation increment of Phase 3 (stacks on #1029). Adds a new Rust binary crate
`aleph-vm-controller` that reads the existing `{vm_hash}-controller.json` and, for a
non-confidential persistent QEMU VM, produces a QEMU argv byte-identical to the Python
`QemuVM.start()`, spawns it, streams stdout/stderr to the systemd journal, and shuts it
down gracefully on SIGTERM. This is a faithful 1:1 port; the Python controller is the
parity oracle. Confidential/SEV is A2; packaging cutover is A3 (this PR wires nothing
into systemd yet).

**Parity is pinned by an oracle-generated conformance suite:** 14 argv fixtures are
generated by monkeypatching the real Python `QemuVM.start()` (capturing the exact argv
it would exec), and a Rust integration test asserts the Rust builder matches each
byte-for-byte. Mutation testing killed 23/23 deliberate breakages after the review batch.

**Reviewed by three adversarial lenses** (parity vs Python, Rust robustness, test
quality with mutation). Fixes applied in separate commits, each with a regression test:
- QMP reader given a wall-clock deadline + line/skip caps, and the graceful-stop QMP
  calls wrapped in a timeout so a chatty or wedged monitor can no longer hang the stop
  escalation (the qcow2-corruption path).
- Dispatch now branches on the `hypervisor` field first (matching Python precedence):
  firecracker and confidential-shaped configs are rejected in A1, not booted.
- `mem_size_mb` coercion fails closed on negative/non-finite (no silent `-m 0`).
- Fixtures added for the three falsy corners the mutation tester found unpinned
  (`qga_socket_path=None` "None"-literal wart, empty-string `interface_name` /
  `cloud_init_drive_path`). journal header protocol and escalation-budget tests added.

Deliberate divergences from Python recorded in `docs/plans/rust-port-divergences.md`
(entries 54-57): fail-open journal fallback, exit-0-on-qemu-crash (Python wart, ported),
omitted `-i` flag.

`cargo fmt`/`clippy -D warnings`/`test` green; 40 controller tests + the daemon suite.


foxpatch-aleph (approved):
A high-quality, faithful 1:1 Rust port of the Python QemuVM controller for the non-confidential persistent QEMU path. The argv builder, process lifecycle, graceful-stop escalation, and config parser all correctly reproduce the Python behavior, verified against the actual Python source. The QMP client is more robust than the Python qmp library (wall-clock deadlines, byte/event caps) while maintaining behavioral parity. Test coverage is thorough: 14 oracle-generated conformance fixtures pin the argv byte-for-byte, and QMP/config tests cover all bounded-failure modes. The deliberate divergences (fail-open journal, exit-0-on-crash, omitted -i flag) are well-documented in the divergence registry. No correctness, security, or race condition issues found.

**`rust/crates/supervisor-controller/src/config.rs`** (line 162): Minor: `sev_policy: Option<u32>` vs Python's `sev_policy: int`. The Python `QemuConfidentialVMConfiguration` field is an unbounded `int` (the `QemuConfidentialVM` subclass converts it via `hex(config.sev_policy)` for the QEMU argv). The `u32` restriction is stricter (rejects negatives and >2^32-1), which is fail-closed. Harmless for A1 since confidential configs are rejected, but for A2 parity you may want `Option<i64>` to match the Python range, or document the narrowing.

**`rust/crates/supervisor-controller/src/config.rs`** (line 112): Minor: `HostVolume.mount` has `#[serde(default)]` (defaults to empty string) while Python's `QemuVMHostVolume.mount` is a required field with no default. A config missing `mount` would fail in Python but parse in Rust. Since `mount` is never consumed (QemuVM drops it), this is purely cosmetic and already documented as 'carried for shape completeness'. No action needed for A1.

**`scripts/generate_controller_argv_fixtures.py`** (line 41): Nit: The module-level `_captured` dict is a shared mutable global. If `vm.start()` ever failed to raise `_Abort` (e.g., a future code change in QemuVM.start() before the exec), a stale argv would leak into the next test case undetected. Consider clearing `_captured` at the start of `_argv_for`, or using a return value instead of the global. Non-blocking since this is a dev tool and the `_Abort` always fires.

foxpatch-aleph (approved):
A well-executed, faithfully documented 1:1 Rust port of the Python QEMU controller for the non-confidential persistent path. The argv builder is pinned by oracle-generated conformance fixtures, the QMP client is robustly bounded against wedged monitors with three-layer protection, and the dispatch logic correctly mirrors Python's hypervisor-first precedence. Deliberate divergences (journal fail-open, exit-0-on-crash, omitted -i flag, log format) are clearly documented with rationales. The code is clean, well-commented, and thoroughly tested. The only observations are non-blocking: a theoretical float-saturation edge case in mem_size_mb coercion (not reachable via the production config writer), and a QMP command formatting pattern that is safe today but would benefit from a noted invariant. Overall this is production-ready for the A1 scope.

**`rust/crates/supervisor-controller/src/config.rs`** (line 222): Theoretical parity gap: `float as u64` saturates to `u64::MAX` for finite floats exceeding `u64::MAX`, where Python's `int(1e20)` produces a larger arbitrary-precision integer. Not practically reachable (the supervisor writer always sets mem_size_mb from integer MiB counts), but the doc comment at lines 182-198 claims faithful reproduction of `_coerce_mib` without acknowledging this edge. Consider adding a note or a saturation check for completeness.

**`rust/crates/supervisor-controller/src/qmp.rs`** (line 89): The `command` parameter is interpolated into a JSON string via `format!` rather than using `serde_json` to build the request. This is safe today because the only callers pass `&'static str` literals (`"system_powerdown"`, `"quit"`), but if a future caller passes user-controlled data, this would be a JSON injection vector. Consider adding a comment noting the `&'static str` invariant, or building the JSON with `serde_json::json!`.

**`rust/crates/supervisor-controller/src/journal.rs`** (line 70): The `identifier` is interpolated unescaped into the journald header. If `vm_hash` contained a newline, it would corrupt the 7-line header protocol. In practice `vm_hash` is a hex item hash so this is not an injection vector, but a brief comment noting this assumption would be helpful.

**`rust/crates/supervisor-controller/src/qemu.rs`** (line 185): The Rust `stop()` is awaited before `child.wait()`, which is actually better than Python's fire-and-forget `loop.create_task(execution.stop())` that may be cancelled if `process.wait()` returns first. This implicit improvement over Python is not listed in the divergences doc (entries 54-57). Consider noting it for completeness, or leave as-is since it's strictly safer.

foxpatch-aleph (approved):
Excellent faithful 1:1 Rust port of the Python aleph-vm-controller for non-confidential QEMU. I verified argv ordering, truthiness semantics, config parsing (including the mem_size_mb coercion and negative-rejection), stop escalation budgets, QMP client bounds, journal header protocol, and dispatch precedence against the Python source — all match or improve on the Python behavior with documented divergences. The conformance fixtures are oracle-generated from the real Python QemuVM.start(), mutation-tested, and guarded against drift. No bugs or security issues found.

**`rust/crates/supervisor-controller/src/qmp.rs`** (line 33): Minor observation: `READ_INACTIVITY_TIMEOUT` (10s) is always larger than `COMMAND_DEADLINE` (5s), so `remaining.min(READ_INACTIVITY_TIMEOUT)` always returns `remaining`. The 10s constant never actually fires as a per-read timeout within a single command. This is fine as defense-in-depth (if COMMAND_DEADLINE is ever raised above 10s), but worth noting it's effectively dead code in the current configuration.

**`rust/crates/supervisor-controller/src/qemu.rs`** (line 193): The exit code is always mapped to `ExitCode::SUCCESS` in `main.rs:111` (`Ok(_) => ExitCode::SUCCESS`), matching the Python exit-0 wart (divergence #55). The `code` value returned by `run()` is discarded. This is documented and intentional, but a reader might find the `Ok(code)` return misleading since the code is never used. Not a bug — just a readability note.

foxpatch-aleph (approved):
A high-quality, faithful Rust port of the Python QEMU controller for the non-confidential persistent path. The argv builder is pinned by oracle-generated conformance fixtures, the QMP client is robustly bounded against wedged/chatty monitors, and all deliberate divergences are documented. No bugs, security issues, or correctness problems found.

**`rust/crates/supervisor-controller/src/main.rs`** (line 35): Minor: The Rust `-v` flag accepts a single `-v` where Python's argparse only accepts `-vv` (two v's as a single option string). This is deliberate and documented in the comment, but the comment text "Python accepts `-v`, `-vv` and `--very-verbose`" is slightly misleading — Python only accepts `-vv` and `--very-verbose`, not `-v`. Not a blocker since production ExecStart never passes either flag.

**`rust/crates/supervisor-controller/src/qmp.rs`** (line 144): The byte-by-byte read loop is slightly inefficient for large QMP messages, but QMP responses are typically small JSON objects so the performance impact is negligible. The `BufReader` amortizes socket reads, and the per-iteration deadline check ensures correctness. No change needed.

**`rust/crates/supervisor-controller/src/qemu.rs`** (line 36): The `QMP_CALL_BUDGET` is 8s (COMMAND_DEADLINE 5s + 3s). In the worst case where both QMP calls take their full budget, the total escalation could be 8+50+8+10=76s, exceeding the 60s systemd SIGKILL deadline. In practice QMP calls complete in milliseconds, so this is not a real concern, and the behavior is strictly better than the Python code which has unbounded QMP calls. Just an observation — no change needed.

PR #1031, opened 2026-07-07
feat(rust): A2 - Rust controller, SEV/SEV-ES confidential argv parity

odesenfans:
## A2: Rust controller, SEV/SEV-ES confidential parity

Second foundation increment of Phase 3 (stacks on #1030). Extends the Rust
`aleph-vm-controller` to launch an existing SEV / SEV-ES confidential QEMU VM with a
QEMU argv byte-identical to the Python `QemuConfidentialVM.start()`. SEV-SNP is B1;
this reproduces only the current `-object sev-guest` path. Packaging is A3.

**What it adds:**
- `build_confidential_argv` (OVMF pflash, qcow2 rootfs, `--no-reboot`, paused `-S`,
  `-object sev-guest` with policy/cbitpos/reduced-phys-bits/dh-cert/session,
  `q35,confidential-guest-support=sev0`, always `host-phys-bits-limit`, NIC without
  `rombar=0`, and the confidential `-qmp`-before-qga ordering). The plain-path helpers
  (host volumes, GPU, qga) are factored out and shared; the A1 spawn/journal/graceful-
  stop path is reused verbatim (verified byte-identical, no plain-path regression).
- `cpuid.rs`: a `SevHostInfo` reader for cbitpos/reduced-phys-bits from CPUID leaf
  0x8000001F EBX, matching the Python `SecureEncryptionEbx` field layout. It fails
  closed on the SEV feature bit (a deliberate, documented safety improvement over
  Python, divergence 58). The argv builder takes an injected `SevHostInfo` so parity
  fixtures are deterministic on non-SEV CI hardware.
- Dispatch branches on the `hypervisor` field first (Python precedence): a confidential
  payload mislabelled firecracker still fails closed.

**Parity + tests:** 8 confidential fixtures generated from the real
`QemuConfidentialVM.start()` (SEV 0x1, SEV-ES 0x5, policy 0x0, GPU double
`host-phys-bits-limit`, host volume, cloud-init, and a distinct-CPUID case that makes
the injected values load-bearing). Reviewed by three adversarial lenses: no argv,
policy, bit-extraction, generator, or dispatch divergence. Fix batch pinned the one
surviving mutation (shared CPUID constants) and added prelaunch-guard tests
(SEV-platform refusal, missing godh/session).

`cargo fmt`/`clippy -D warnings`/`test` green; `ruff format` clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1032, opened 2026-07-07
feat(packaging): A3 - cut the persistent-QEMU launch over to the Rust controller

odesenfans:
## A3: packaging cutover + real-KVM integration (Rust owns the persistent-QEMU launch)

Third foundation increment of Phase 3 (stacks on #1031). Wires the Rust
`aleph-vm-controller` (A1 plain + A2 confidential) into packaging behind the existing
impl flag and validates it booting a real QEMU VM on KVM. After this, Rust owns the
persistent-QEMU launch.

**Cutover (mirrors the daemon's supervisor-launcher):**
- New `packaging/controller-launcher` (POSIX sh): dispatches `aleph-vm-controller@`
  on `ALEPH_VM_SUPERVISOR_IMPL` (rust -> the Rust binary, python -> the module,
  unknown -> python fallback that never bricks the node), forwarding `--config`.
- The controller unit now `EnvironmentFile=`s the same `/etc/aleph-vm/supervisor.env`
  the daemon reads, so one flag switches BOTH the daemon and every newly (re)started
  controller. `KillMode=mixed` / `TimeoutStopSec=60` / PYTHONPATH preserved.
- The deb builds and installs `aleph-vm-controller` + `controller-launcher` to
  `/opt/aleph-vm/bin/` (deb-contents CI check greps for both). Rollback = flip the flag.

**Real-KVM validation:** the integration harness previously always ran the python
controller even on the rust leg; it now selects the Rust controller under `impl=rust`
and fails loudly if its binary is missing. A new `assert_controller_matches_impl`
guard isolates "wrong controller ran" from "controller ran but VM did not serve", so
the rust leg cannot false-green on python. Tests covering boot -> RUNNING/serves ->
graceful stop and daemon-restart re-adoption now exercise the Rust controller.

This is the first time the Rust controller boots a real QEMU VM on KVM (A1/A2 were
argv-fixture only), so the rust integration leg here is the true validation of the
foundation. Deliberate divergence 59 recorded (the source-tree harness mirrors the
launcher dispatch rather than exec'ing `/opt/aleph-vm` paths).

`cargo fmt`/`clippy -D warnings`/`test` green; shellcheck clean on both launchers;
`ruff format` clean.


foxpatch-aleph (commented):
The PR is well-executed: the controller-launcher correctly mirrors the established supervisor-launcher pattern, the systemd unit wiring is sound (EnvironmentFile for single-source-of-truth impl selection, PYTHONPATH preserved for the python fallback), and the integration test harness adds a meaningful assert_controller_matches_impl guard that fails loudly on wrong-controller regressions. The cutover design — one env flag switching both daemon and controller — is clean and the rollback story (flip the flag) is simple. No correctness or security issues found. The main observation is that the CI shellcheck step only covers *.sh files, so neither controller-launcher nor supervisor-launcher is actually shellchecked in CI, despite the PR description claiming shellcheck coverage — this is a pre-existing gap extended by this PR. A secondary nit is the inconsistency between controller-launcher using /usr/bin/python3 (absolute) and supervisor-launcher using bare python3 (PATH).

**`packaging/controller-launcher`** (line 1): The CI `code-quality-shell` job in `.github/workflows/test-using-pytest.yml` runs `find ./ -type f -name "*.sh"`, but this file has no `.sh` extension, so shellcheck never runs on it in CI. The PR description claims "shellcheck clean on both launchers" but that's a manual claim, not CI-enforced. Consider either renaming to `controller-launcher.sh` (and `supervisor-launcher.sh`), or changing the find command to also match executable shell scripts without extensions (e.g. `find ./ -type f -name "*.sh" -o -name "*-launcher" -exec shellcheck {} \;`). This is a pre-existing gap from the supervisor-launcher that this PR extends.

**`packaging/controller-launcher`** (line 20): Minor inconsistency: this launcher uses `/usr/bin/python3` (absolute path) while the sibling `supervisor-launcher` uses bare `python3` (PATH-dependent). The absolute path here is arguably more robust; consider aligning supervisor-launcher to match for consistency, or leave as-is since both work in practice under systemd.

foxpatch-aleph (approved):
This PR cleanly wires the Rust controller into the persistent-QEMU launch path via a new controller-launcher that mirrors the existing supervisor-launcher dispatch pattern. The single-source-of-truth design (one EnvironmentFile, one flag switching both daemon and controller) is sound, the fallback-to-python-on-unknown behavior prevents operator typos from bricking nodes, and the EnvironmentFile does not clobber the unit's PYTHONPATH. The dispatch test execs the real script and covers all four branches (rust, python, unset, bogus), plus a production-defaults guard. The integration harness adds assert_controller_matches_impl as a loud guard against wrong-controller regressions. The CI gate fix (AVM_ITEST instead of CI, since sudo strips CI) is a correct and important fix. The divergence entries 59 and 60 are thorough and accurate.

**`tests/test_controller_launcher.py`** (line 43): The marker script uses `"$*"` (unquoted `$*`) which joins args with the first IFS character. This works correctly for the test's arguments (none contain spaces), but `"$@"` would be the more conventional and robust choice for preserving argument boundaries. Non-blocking since the current test arguments are space-free, but worth noting for consistency with shell best practices.

**`packaging/supervisor-launcher`** (line 15): The supervisor-launcher was modified in this PR to add the same overridable seam (`ALEPH_VM_SUPERVISOR_BIN` / `ALEPH_VM_SUPERVISOR_PYTHON`), but there is no corresponding dispatch test (unlike the new `tests/test_controller_launcher.py` for controller-launcher). The dispatch logic is identical and now shellcheck-linted, so this is a non-blocking asymmetry, but a parallel `tests/test_supervisor_launcher.py` would make the coverage symmetric and guard against future drift between the two launchers.

foxpatch-aleph (approved):
Well-executed packaging cutover that correctly wires the Rust controller into the systemd unit via a launcher script mirroring the existing supervisor-launcher pattern. The single-source-of-truth EnvironmentFile approach, the loud-failure integration guards (assert_controller_matches_impl, missing-binary pytest.fail), the thorough dispatch unit tests, and the important CI guard fix (CI -> AVM_ITEST) are all sound. No correctness, security, or testing gaps found. The two divergence entries (59, 60) are well-documented with clear rationale and latent-risk notes.

**`packaging/supervisor-launcher`** (line 16): The supervisor-launcher now also has the override seam (ALEPH_VM_SUPERVISOR_BIN / ALEPH_VM_SUPERVISOR_PYTHON) but has no corresponding dispatch test like test_controller_launcher.py. This is a pre-existing gap (the seam was added here without a test), but given the launcher is on the critical path for the production cutover, a parallel test_supervisor_launcher.py would provide the same safety guarantee the controller test does.

**`packaging/controller-launcher`** (line 20): Minor asymmetry: ALEPH_VM_CONTROLLER_PYTHON defaults to '/usr/bin/python3' (absolute) while ALEPH_VM_SUPERVISOR_PYTHON defaults to 'python3' (PATH-relative). This is correct (each matches what the systemd units previously hardcoded), but a one-line comment noting the intentional difference from supervisor-launcher's python3 would help future readers avoid 'fixing' one to match the other.

PR #1033, opened 2026-07-07
feat(rust): B0 - port aleph-tee SEV-SNP crate with AMD ARK root pinning

odesenfans:
## B0: port the aleph-tee crate (SEV-SNP report + AMD-KDS verification), hardened

First SNP increment of Phase 3 (stacks on #1032). Ports aleph-cvm's `aleph-tee`
crate into `rust/crates/aleph-tee`: the TeeBackend trait, SEV-SNP attestation report
generation and parsing, AMD-KDS certificate-chain verification, the `sev-snp-guest`
QEMU argument generator, and X.509 attestation embedding (OID 1.3.6.1.4.1.60000.1.1).
This is the shared library B1 (SNP host launch) and B2b (attest-cli) build on. The
security constants (report offsets, `SIGNED_REPORT_SIZE=0x2A0`, r/s layout, VMPL gate,
KDS URL, cbitpos/policy, OID) are byte-identical to the donor.

**Hardened beyond the donor** (reviewed by three lenses; the donor's attestation had a
forgeable trust anchor and near-zero crypto test coverage, both fixed here):
- **AMD ARK root pinning:** `verify_cert_chain` now pins the chain's ARK against AMD's
  genuine root per product (SubjectPublicKeyInfo compared to the `sev` crate's vendored
  `builtin::{milan,genoa,turin}` roots). The donor tied the root to AMD by a forgeable
  CN/O string check alone, so a fabricated self-signed ARK passed; that is now rejected.
  A wrong pin fails closed (rejects genuine reports), the safe direction.
- **Crypto now tested:** synthetic P-384 key/cert tests kill the five mutations that
  previously survived (report-signature verify, VMPL>1 gate, each cert-chain link + ARK
  identity, cbitpos=51, the OID value). A tampered signature and a forged AMD ARK are
  asserted rejected.
- **Robustness:** cert validity/expiry checking, a 256 KiB KDS response cap (anti-OOM),
  a `product` allowlist (anti path-traversal/URL-injection), and a DER-length parse that
  can no longer panic on a crafted length.

Report generation (`/dev/sev-guest`) stays hardware-gated; the live-KDS ARK-match test
is `#[ignore]`d (network). 62 tests pass. Divergences 60-62 recorded, with a
recommendation to upstream the pinning and hardening to aleph-cvm so the two do not
drift. `cargo fmt`/`clippy -D warnings`/`test` green workspace-wide.


foxpatch-aleph (approved):
A well-executed port of the aleph-tee SEV-SNP crate with meaningful security hardening beyond the donor. The AMD ARK root pinning (comparing SubjectPublicKeyInfo DER) is the critical improvement and is correctly implemented with a fail-closed posture. Test coverage is thorough — synthetic P-384 certs exercise every reject path in the chain verification, signature verification, VMPL gate, OID pinning, and DER-length overflow handling. The KDS response cap, product allowlist, and cert validity checking are all solid defensive measures. The minor issues (unused import, dead code, step renumbering, roundabout byte conversion) are non-blocking nits. The deferred basicConstraints CA:TRUE check is documented and not exploitable given the fixed-depth chain. Overall this is a high-quality security contribution.

**`rust/crates/aleph-tee/src/sev_snp/report.rs`** (line 3): `ByteParser` is imported but never used directly — `SevAttestationReport::from_bytes` is a static method that uses it internally. This import can be removed.

**`rust/crates/aleph-tee/src/sev_snp/verify.rs`** (line 105): Step numbering: this is labeled "2." but the VMPL gate on line 100 is also step "2.". Renumber to "3." (or renumber all subsequent steps) to match the documented steps in the function-level doc comment.

**`rust/crates/aleph-tee/src/sev_snp/certs.rs`** (line 250): The `first_len > data.len()` check is dead code: `der_object_length` already rejects any encoded length exceeding the buffer (line 316-321). Harmless but redundant.

**`rust/crates/aleph-tee/src/sev_snp/verify.rs`** (line 379): The `.collect::<Vec<_>>().into_iter().copied().collect()` roundtrip can be simplified to `r_bytes_le.iter().rev().copied().collect()` — same result, clearer intent.

**`rust/crates/aleph-tee/src/sev_snp/certs.rs`** (line 107): Security note (non-blocking): the on-disk cert cache has no integrity protection. A poisoned cached ARK is harmless (rejected by the pin), but a poisoned VCEK paired with a genuine cached ARK could theoretically bypass signature verification IF the attacker also controls the report. In practice the attacker needs the AMD private VCEK key to produce a valid report signature, so this is low risk. Consider adding a comment documenting that the ARK pin is what makes cache poisoning safe, so future refactors don't accidentally remove the pin while keeping the cache.

foxpatch-aleph (approved):
Well-structured SEV-SNP attestation crate with genuine security hardening over the donor. The ARK root pinning (SubjectPublicKeyInfo comparison against vendored AMD roots) is correctly implemented and defeats forged-ARK attacks that the donor's string-only check permitted. The product allowlist, KDS body cap, checked-arithmetic DER parsing, cert validity checking, and VMPL gate are all sound. Test coverage is thorough: synthetic P-384 keys/certs exercise the happy path and each rejection reason individually (broken ASK<-ARK link, broken VCEK<-ASK link, expired cert, non-self-signed ARK, pin mismatch, tampered signature, tampered body, VMPL>1). The forged-ARK-against-real-pin test directly validates the security fix. Minor observations (cosmetic step numbering, verbose Vec construction, latent ASN.1 encoding fragility in secondary metadata check) are all non-blocking nits.

**`rust/crates/aleph-tee/src/sev_snp/verify.rs`** (line 105): Comment numbering: there are two "step 2" labels (line 100 for VMPL gate, line 105 for chip_id extraction). The VMPL gate should be step 2 and chip_id extraction step 3, with subsequent steps renumbered. Cosmetic only.

**`rust/crates/aleph-tee/src/sev_snp/verify.rs`** (line 379): The `.collect::<Vec<_>>().into_iter().copied().collect()` chain can be simplified to `.iter().rev().copied().collect::<Vec<u8>>()` — the intermediate `Vec<&u8>` allocation is unnecessary. Pure style nit.

**`rust/crates/aleph-tee/src/sev_snp/certs.rs`** (line 285): Short-form DER length returns `2 + content_len` without checking it fits the buffer. This is safe because `split_der_certs` checks `first_len > data.len()` after, but `der_object_length` as a standalone function could return a length past the end for short form. Consider adding the same `total > data.len()` guard that the long-form branch has, for consistency and defense-in-depth if the function is ever called directly.

foxpatch-aleph (approved):
Well-structured, security-hardened port of the aleph-tee SEV-SNP crate. The AMD ARK root pinning (comparing SubjectPublicKeyInfo against the sev crate's vendored roots) is a genuine security improvement over the donor's forgeable CN/O string check. Crypto test coverage is excellent — the synthetic P-384 key/cert tests cover the happy path plus five mutation scenarios (signature tamper, VMPL gate, each cert-chain link, ARK pin mismatch, cert expiry). The KDS response cap, product allowlist, and DER-length overflow fix are all correctly implemented. The few items below are minor nits and defense-in-depth observations, not blockers.

**`rust/crates/aleph-tee/src/sev_snp/verify.rs`** (line 105): Duplicate step numbering: the inline comment says "// 2. Extract chip_id..." but the previous comment at line 100 already used "// 2.". The doc comment at lines 84-90 has the correct sequence (1-6). Consider renumbering this to step 3 and shifting subsequent steps.

**`rust/crates/aleph-tee/src/sev_snp/verify.rs`** (line 126): Unnecessary `vcek_der.clone()`: the original `vcek_der` is never used after the `CertChain` is constructed — line 138 reads `&chain.vcek_der`. You can move `vcek_der` into the chain (`vcek_der: vcek_der`) and use `&chain.vcek_der` on line 138, avoiding the clone of a potentially large DER blob.

**`rust/crates/aleph-tee/src/sev_snp/qemu.rs`** (line 25): Defense-in-depth: `config.tee.policy` is an arbitrary `Option<String>` interpolated directly into the QEMU `-object sev-snp-guest,...,policy={policy}` argument. A malicious value like `"0x30000,inject=on"` could inject additional QEMU options. The `product` field gets allowlist validation via `validate_product`, but `policy` does not. Consider validating it as a hex string (or at least rejecting commas/spaces) if this struct can ever be deserialized from untrusted input.

**`rust/crates/aleph-tee/src/sev_snp/verify.rs`** (line 234): Known gap (documented in divergence entry 61b): `verify_cert_chain` does not check `basicConstraints CA:TRUE` or key usage on the ASK. The risk is low because the chain is strictly ARK->ASK->VCEK with no further depth, but if openssl 0.10 ever exposes an ergonomic accessor, adding these checks would prevent a leaf-style cert from being used as an intermediate.

foxpatch-aleph (approved):
Well-structured port of the SEV-SNP attestation crate with meaningful security hardening beyond the donor. The ARK root pinning via the sev crate's vendored AMD roots is the correct defense against forged certificate chains, and it fails closed. Cert chain verification, ECDSA signature checking, VMPL gating, policy canonicalization, product allowlisting, and DER overflow-safe parsing are all correct. Test coverage is comprehensive: the five mutation killers (signature, VMPL, each chain link, ARK identity, cbitpos/OID) are all asserted, plus a forged-ARK-against-real-pin test and an ignored live-KDS divergence test. No correctness or security bugs found; the comments below are minor cleanup nits.

**`rust/crates/aleph-tee/src/sev_snp/report.rs`** (line 3): Unused import: `use sev::parser::ByteParser;` is never referenced in this file. Remove it to keep clippy clean (the PR description claims `clippy -D warnings` passes, so this may be masked by a workspace-level allow).

**`rust/crates/aleph-tee/Cargo.toml`** (line 11): `rcgen` is used only in `#[cfg(test)]` blocks in `x509.rs` (lines 177, 180, 183, 205, 208). It should be moved to `[dev-dependencies]` alongside `tokio` to avoid pulling a certificate-generation crate into production builds.

**`rust/crates/aleph-tee/src/sev_snp/verify.rs`** (line 380): The little-to-big-endian conversion does a double collect: `.collect::<Vec<_>>().into_iter().copied().collect()`. This can be simplified to `r_bytes_le.iter().rev().copied().collect::<Vec<u8>>()` — the intermediate `Vec<&u8>` allocation is unnecessary.

**`rust/crates/aleph-tee/src/sev_snp/certs.rs`** (line 214): When the KDS returns PEM, `certs[0]` is treated as ASK and `certs[1]` as ARK. The AMD KDS cert_chain endpoint returns ASK first then ARK, so this ordering is correct, but a brief comment documenting the expected order (ASK=first, ARK=second) would help future readers verify the assumption.

**`rust/crates/aleph-tee/src/sev_snp/backend.rs`** (line 12): Doc comment says "verification (stubbed for now)" — this is slightly misleading since full verification exists in `verify.rs` via `verify_sev_snp_report()`. Consider rewording to "synchronous verification is structural-only; use verify_sev_snp_report() for full async cryptographic verification" to match the method-level doc.

PR #1034, opened 2026-07-07
feat: B2a - measured-boot Nix image + hardened in-guest attest-agent

odesenfans:
## B2a: measured-boot Nix image + in-guest attest-agent (hardened)

Stacks on #1033. Two deliverables for the SEV-SNP guest side:

**1. `aleph-attest-agent` crate** (`rust/crates/aleph-attest-agent`) - the in-guest
attested-TLS sidecar: generates an ephemeral P-384 key, gets a key-bound SEV-SNP report,
embeds it in a self-signed cert via the aleph-tee OID extension, serves fresh
attestation + one-shot secret injection, and reverse-proxies to the workload. Ported
from aleph-cvm; unifies rustls onto the ring provider (aws-lc kept out of the tree).

**2. Core measured-boot Nix flake** (`nix/`) - measured OVMF + kernel + initrd (with the
attest-agent baked in) + a dm-verity rootfs, and a precomputed `sev-snp-measure`
measurement. Compose-rootfs, LUKS/encrypted, and the demo workload are scoped out
(design non-goals). **The donor's measurement was non-reproducible (random mkfs/verity
seeds); this makes it deterministic** (fixed UUID, non-zero hash_seed, SOURCE_DATE_EPOCH,
pinned verity salt/uuid), and `nix build --rebuild` of rootfs/verity/initrd/measurement
is byte-identical, resolving the top SNP risk (measurement reproducibility) before
testnets.

**Hardened beyond the donor** (three review lenses; the crypto fix is security-critical):
- **Attested-key-confusion MITM closed:** the fresh-attestation endpoint no longer places
  the raw client nonce into `report_data`. The canonical `report_data` scheme is now
  domain-separated and channel-bound (`SHA-384(DOMAIN || pubkey [|| nonce])`), placed in
  the shared `aleph-tee` crate so the agent and the B2b verifier cannot drift. A relayed
  fresh report can no longer be reused with an attacker key (attack-impossibility test).
- Secrets bind-mounted into the chroot (delivery was silently broken); secret store
  symlink-hardened (O_EXCL/O_NOFOLLOW, 0700); an nft ruleset firewalls the guest to only
  the attested port (prevents attestation bypass if the app binds 0.0.0.0); one-shot
  concurrency test; honest zeroization; hop-by-hop header stripping in the proxy.

Report generation is hardware-gated; the live-KDS test is `#[ignore]`d. Divergences
63-67 recorded, with an upstreaming recommendation for the domain-separation fix.
`cargo fmt`/`clippy -D warnings`/`test` green; Nix outputs build and rebuild reproducibly.

Known deferred (tracked, ledger 66): inject-secret is unauthenticated (owner-auth is a
later confidential-flow design step, not needed for the attest-only testnets path).


foxpatch-aleph (approved):
Well-structured PR delivering two key deliverables: the aleph-attest-agent crate (in-guest attested-TLS sidecar) and the measured-boot Nix flake. The security-critical report_data domain-separation fix (divergence 65) is correct, well-tested (domain separation, channel binding, and the specific attack-impossibility test all present), and properly placed in the shared aleph-tee crate to prevent agent/verifier drift. The SecretStore refactor from the donor's global static to injectable web::Data is a clean testability improvement with comprehensive unit tests covering one-shot semantics, validation limits, symlink hardening, and concurrency. The Nix reproducibility hardening (fixed UUID, non-zero hash_seed, SOURCE_DATE_EPOCH, fixed verity salt/uuid, ^has_journal) addresses the top SNP risk correctly. Hop-by-hop header stripping in the proxy is a good security practice. The divergences ledger (entries 63-67) is thorough and honest about deferred items (unauthenticated inject-secret, rate-limiting). Findings below are minor/comment-level and do not block merge.

**`rust/crates/aleph-attest-agent/src/proxy.rs`** (line 112): Content-Length is not stripped when forwarding request headers to the upstream. reqwest's `.body()` will set its own Content-Length from the actual body, which could conflict with the forwarded client value (duplicate or mismatched header). Similarly, on the response side (line 134-140), the forwarded Content-Length is set via `insert_header` before `resp.body(resp_body)`. Since the body is passed through verbatim the values should match, but the behavior is framework-dependent and fragile. Consider adding `content-length` to the skip list alongside Host and hop-by-hop headers, letting each framework recompute it from the actual body.

**`rust/crates/aleph-attest-agent/Cargo.toml`** (line 21): `thiserror` is declared as a dependency but not used in any source file (no `#[derive(thiserror::Error)]` or `thiserror::Error` import appears in the crate). Consider dropping it to keep the dependency tree lean, or add a `#[derive(thiserror::Error)]` if error types were intended.

**`rust/crates/aleph-attest-agent/src/main.rs`** (line 82): `reqwest::Client::new()` creates an HTTP client without an explicit rustls provider install. The workspace pins rustls with `default-features=false` (no process-default provider auto-installed), and `tls.rs` only passes the ring provider to the *server* config via `builder_with_provider`. The current upstream is plain HTTP (loopback), so no client TLS is needed today. If an HTTPS upstream is ever configured, reqwest would panic at runtime with no default provider. Consider either calling `rustls::crypto::ring::default_provider().install_default()` in main before creating the client, or documenting that only HTTP upstreams are supported.

**`rust/crates/aleph-attest-agent/src/secrets.rs`** (line 75): The `unsafe { libc::geteuid() }` block is correct (geteuid is a safe syscall with no UB preconditions) but lacks a SAFETY comment. Adding `// SAFETY: geteuid() is a safe POSIX syscall with no pointer arguments.` would satisfy the `unsafe_op_in_unsafe_fn` lint and document intent.

**`nix/init.sh`** (line 108): The guest firewall only defines an `input` chain with `policy drop`. This is correct for the current architecture (the only outbound is to loopback for the upstream proxy), but if the agent ever needs to make outbound connections to an external network (e.g. an HTTPS upstream or an external KDS endpoint), return traffic on ephemeral ports would be dropped by the stateless input filter. A comment noting that outbound connections to non-loopback destinations require either a conntrack module or an output/forward chain addition would help future maintainers.

foxpatch-aleph (commented):
Well-structured PR with a security-critical crypto fix (domain-separated, channel-bound report_data closing the attested-key-confusion MITM), thorough secret-storage hardening (O_EXCL/O_NOFOLLOW/0700/symlink-rejection/owner-check/concurrency), proper hop-by-hop header stripping in the proxy, a fail-closed guest firewall, and genuine Nix reproducibility fixes. The findings are mostly minor code-quality issues: a forwarded upstream Content-Length header that should be stripped, a Zeroizing gap where key_bytes.to_vec() creates an unwiped copy, misleading dm-crypt comments in kernel.nix for an out-of-scope LUKS path, a non-verity fallback mount that goes read-write, and minor test hygiene. None are blocking.

**`rust/crates/aleph-attest-agent/src/proxy.rs`** (line 138): The upstream `Content-Length` header is forwarded via `insert_header` but is NOT in the hop-by-hop skip list. Since `resp.body(resp_body)` is called with the fully-buffered body, actix-web computes its own `Content-Length`; `insert_header` then overrides it with the upstream's value. If they match (body is verbatim) this is safe, but a mismatched upstream `Content-Length` would cause a response desync. Consider adding `content-length` to the response-side skip list (or use a separate filter) so actix always computes it from the actual body.

**`rust/crates/aleph-attest-agent/src/tls.rs`** (line 85): The `Zeroizing` wrapper on `key_bytes` is undermined by `key_bytes.to_vec()` which creates a plain `Vec<u8>` that goes into `PrivatePkcs8KeyDer` and is never zeroized. The comment says "keep our transient copy under Zeroizing" but the actual copy rustls receives is un-wiped. Under SEV-SNP memory encryption this is acknowledged as best-effort, but the comment overstates the protection. Consider wrapping the final `Vec<u8>` in `Zeroizing` as well, or updating the comment to reflect that the `to_vec()` copy is not wiped.

**`nix/kernel.nix`** (line 30): The comment says "Crypto algorithms required by dm-crypt / LUKS (aes-xts-plain64)" but LUKS is explicitly out of scope (design section 6 non-goals). dm-verity needs SHA-256 (already at line 33), but AES/XTS/ESSIV are LUKS-specific. These are harmless to keep but the comment is misleading. Consider updating the comment to clarify these are retained for completeness or future LUKS support, not for the current dm-verity-only chain.

**`nix/init.sh`** (line 176): The non-verity fallback path does `mount "$blkdev" /mnt/root` (read-write) as a last resort if the `ro` mount fails. This means the rootfs could be mounted writable without verity protection. Since the verity-protected path is the default and the non-verity path is a "backwards compatible" fallback, consider whether a writable mount is appropriate here, or at minimum log a warning that the rootfs is not integrity-protected.

**`rust/crates/aleph-attest-agent/src/secrets.rs`** (line 75): `unsafe { libc::geteuid() }` is correct (geteuid is a trivial syscall), but a brief safety comment like `// SAFETY: geteuid() is a no-signal, no-allocation syscall` would clarify why this `unsafe` is sound.

**`rust/crates/aleph-attest-agent/src/secrets.rs`** (line 417): In `test_symlink_at_key_path_is_refused_not_followed`, the `decoy` temp dir is created via `tmpdir()` but never cleaned up if the assertion passes (only `dir` is removed at line 419). Consider adding `std::fs::remove_dir_all(&decoy).ok();` alongside the `dir` cleanup.

**`rust/crates/aleph-attest-agent/src/proxy.rs`** (line 96): URL construction does not handle a trailing slash on `state.upstream` + path starting with `/`, which would produce a double slash (e.g. `http://host:8080//path`). The default upstream has no trailing slash so this is not a live bug, but consider normalizing with a `trim_end_matches('/')` on the upstream for robustness.

foxpatch-aleph (approved):
Well-structured PR that faithfully ports the aleph-cvm attest-agent and measured-boot Nix chain with meaningful security hardening beyond the donor. The security-critical report_data module correctly closes the attested-key-confusion MITM via domain separation and channel binding, and placing it in the shared aleph-tee crate prevents agent/verifier drift. The secrets module has thorough hardening (O_EXCL|O_NOFOLLOW, owner/symlink checks, mutex-guarded one-shot with concurrency test, Zeroizing). The Nix reproducibility fixes (fixed UUID/hash_seed/verity salt, ^has_journal, SOURCE_DATE_EPOCH) address the top SNP risk. The init.sh firewall is fail-closed. Deferred items (unauthenticated inject-secret, unrate-limited attestation) are properly ledgered. Minor observations: the DHCP path in init.sh doesn't capture the gateway for the chroot resolv.conf, the proxy buffers full request/response bodies in memory (acceptable for a local-only upstream), and there's a theoretical TOCTOU in ensure_secret_dir mitigated by per-file O_NOFOLLOW. None are blockers.

**`nix/init.sh`** (line 47): When DHCP is used (the `else` branch), the `gateway` variable is never set. Later in `prepare_chroot` (line 93), `if [ -n "$gateway" ]` will be false, so the chroot's `/etc/resolv.conf` won't be written. `udhcpc` sets up the initramfs resolv.conf, but the chrooted workload won't see it. Consider capturing the DHCP gateway (e.g. from `udhcpc` output or the default route) so DNS works inside the chroot on the DHCP path.

**`rust/crates/aleph-attest-agent/src/proxy.rs`** (line 126): The proxy buffers the entire request body (`body.to_vec()`) and response body (`upstream_resp.bytes().await`) in memory. For a local-only proxy this is fine, but if the upstream or client ever sends large bodies there's no explicit size limit. Not a blocker for the current 127.0.0.1:8080 placeholder workload, but worth noting if the proxy is reused for larger workloads.

**`rust/crates/aleph-attest-agent/src/secrets.rs`** (line 82): There's a theoretical TOCTOU between `symlink_metadata` checking the directory is real/owned and the subsequent `set_permissions` call — a symlink could replace the directory in between. This is mitigated by the per-file `O_NOFOLLOW` on individual secret writes (the real defense), and the single-process agent makes this race impractical, but worth noting.

**`rust/crates/aleph-attest-agent/src/proxy.rs`** (line 149): When `upstream_resp.bytes().await` fails after headers have already been inserted into `resp`, the code correctly discards `resp` and returns a fresh `HttpResponse::BadGateway()`. This is the right behavior (don't send partial upstream headers with an error body), just confirming it's intentional.

foxpatch-aleph (approved):
Well-structured, security-conscious PR that faithfully ports the aleph-cvm attested-TLS sidecar and measured-boot Nix chain with genuine hardening beyond the donor. The security-critical report_data domain-separation fix correctly closes the attested-key-confusion MITM, with thorough tests proving the attack is impossible. Secret store hardening (O_EXCL/O_NOFOLLOW, symlink/owner checks, atomic Mutex-guarded one-shot) is solid with good test coverage including a concurrency test. The proxy correctly strips hop-by-hop headers and Content-Length to prevent request smuggling. The Nix reproducibility fixes (fixed UUID/hash_seed/verity salt, SOURCE_DATE_EPOCH, ^has_journal) address the top SNP risk. Known deferred items (unauthenticated inject-secret, report-generation rate limiting) are properly ledgered. No bugs or security issues found.

**`rust/crates/aleph-attest-agent/src/proxy.rs`** (line 126): body.to_vec() copies the entire request body into a new Vec<u8> for reqwest. For a production proxy with large uploads this doubles peak memory; consider using Bytes directly (reqwest supports .body(bytes::Bytes)). Not a concern for the placeholder busybox httpd workload.

**`rust/crates/aleph-attest-agent/src/proxy.rs`** (line 149): On the upstream body-read error path (line 151-155), the resp builder with already-forwarded response headers is dropped and a fresh HttpResponse::BadGateway() is returned. This is fine for an error case, but the client loses the upstream's response headers on this path.

**`rust/crates/aleph-attest-agent/src/secrets.rs`** (line 180): The Zeroizing wrapper only zeroes the local copy made on this line. The original String values in the HashMap (owned by actix's deserialized web::Json body) are not zeroized and persist until the handler returns. This is honestly documented in the handler doc comment (lines 226-230) as best-effort with SEV-SNP memory encryption as the backstop.

**`nix/init.sh`** (line 184): The no-verity fallback path falls back to a writable mount if read-only fails (mount without -o ro). This is clearly commented as not integrity-protected and is a backwards-compatible design choice, not a bug. In the measured-boot path (roothash present) this branch is never taken.

**`nix/rootfs.nix`** (line 64): lazy_journal_init=0 is passed alongside -O ^has_journal. The flag is harmless when there is no journal (mke2fs ignores it), so this is just a no-op redundancy, not an error.

PR #1035, opened 2026-07-07
feat(rust): B1 - SEV-SNP host launch + create-path backend selection

odesenfans:
## B1: SEV-SNP host launch + supervisor create-path backend selection

Stacks on #1034. Makes the supervisor able to launch a real SEV-SNP confidential VM.

- **Controller** (`supervisor-controller`): `build_snp_argv` emits a measured
  direct-kernel boot (`-bios` measured OVMF, `-kernel`/`-initrd`, `-append` the exact
  measured cmdline, `-cpu EPYC-v4`, `sev-snp-guest` object with `kernel-hashes=on`,
  `policy=0x30000`, `memory-backend-memfd`). No `-S`, no session/godh: SNP has no
  guest-owner launch-secret handshake; secrets arrive at runtime via the in-guest
  attest-agent (B2a). The TEE fragment is pinned byte-for-byte against the `aleph-tee`
  generator by an oracle test.
- **Daemon** (`supervisor-daemon`): `TeeBackend::SEV_SNP` selects the SNP path, writes
  the measured-boot fields into the controller JSON, and derives the kernel cmdline from
  the verity roothash sidecar so it matches B2a's `sev-snp-measure` inputs exactly.
  `HostInfo.sev_snp_supported` reads `/sys/module/kvm_amd/parameters/sev_snp`; SNP VMs
  report `ConfidentialMode::SEV_SNP` and start immediately (no session to await). The
  SEV/SEV-ES (A2) path is untouched (backend switch only, written-config bytes unchanged
  for plain/SEV).

**Measurement match** (the load-bearing property, verified by the review): the launch
presents the same OVMF/kernel/initrd/cmdline/EPYC-v4/vCPU config B2a measured. The
roothash is hex-validated (cmdline-injection barrier) and the derived cmdline is asserted
as an exact literal.

Reviewed by three lenses (parity/measurement-match, robustness, test-quality). Fixes:
tightened `is_snp()` so a partial config fails closed (no panic-loop, controller/daemon
agree), pinned the roothash guards (malformed/spaced/empty/oversized) with tests, capped
the sidecar read. Divergences 68-70 (cbitpos hardcoded to 51 matching the oracle;
vCPU-count/measurement coupling; SNP NIC matches aleph-vm's controller not the donor).

`cargo fmt`/`clippy -D warnings`/`test` green.


foxpatch-aleph (approved):
This PR adds SEV-SNP host launch and create-path backend selection to the Rust port, and it is well-executed. The byte-parity oracle test pins the controller's TEE fragment against the aleph-tee generator, the full-argv fixtures pin the complete SNP launch, and the roothash hex-validation guard with its battery of injection/malformed/oversized tests is the correct cmdline-injection barrier. The fail-closed dispatch (is_snp vs is_snp_marked) prevents both panic-loops and silent plain/SEV mislaunches, and the controller/daemon config agreements (skip-if-None for byte-identical plain/SEV configs, snp() soft-fail matching is_snp()) are consistent. The SNP path correctly skips cloud-init, starts immediately (no session handshake), and reports ConfidentialMode::SevSnp with awaiting_confidential_init=false. The measurement inputs (EPYC-v4, vCPU count, kernel/initrd/cmdline, OVMF via -bios) match the B2a sev-snp-measure contract. No correctness, security, or test-coverage issues found; the only observations are minor nits (unused _state parameter in snp_config_slice for signature symmetry, and read_to_string vs read on the roothash sidecar which is fine since roothashes are ASCII hex).

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 1756): The `_state` parameter is unused — `snp_config_slice` derives the roothash path from the spec's rootfs, not from daemon state. This is intentional (kept for signature symmetry with `confidential_config_slice` which needs `state` for the session dir), but if the parameter is never needed it could be dropped to avoid the visual noise. Non-blocking.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 1788): `read_to_string` will fail with a generic IO error on a non-UTF-8 sidecar (e.g. a corrupted binary file), rather than reaching the more specific hex-validation guard. A real dm-verity roothash is always ASCII hex so this is harmless, but `read` + `String::from_utf8` would give a more precise error path. Fail-closed is preserved either way; this is a minor robustness nit.

foxpatch-aleph (commented):
A well-executed SEV-SNP host launch increment with strong security boundaries (roothash hex validation preventing cmdline injection, fail-closed on partial configs, read-capped sidecar) and excellent test coverage including a byte-parity oracle against the aleph-tee generator. The dispatch logic correctly routes complete SNP configs to the SNP builder, refuses partial configs cleanly, and prevents silent fallback to plain/SEV. The SNP path correctly starts immediately (no -S, no session await) and skips cloud-init. One notable observation: build_snp_argv does not emit GPU args, so GPUs in an SNP spec are silently dropped by the controller — this is likely intentional for B1 scope but should be documented or explicitly guarded.

**`rust/crates/supervisor-controller/src/qemu.rs`** (line 478): build_snp_argv does not call gpu_args() — if the daemon writes a controller config with non-empty `gpus` for an SNP VM (which build_written_config does populate from spec.gpus), the controller silently drops all GPU passthrough devices. This is likely intentional for B1 (all fixtures use `gpus: []`), but consider either (a) adding a debug_assert that gpus is empty for SNP configs, or (b) documenting in the function docstring that GPU passthrough is not supported in the SNP path yet.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 2150): The `confidential` variable is `true` for SNP VMs (since `request.tee.is_some()`), but it is only consumed inside the `if !snp` block for cloud-init creation. This is functionally correct but slightly misleading — `confidential` is really `has_tee`, not `is_sev_confidential`. Consider naming it `has_tee` or adding a comment clarifying it is only used in the non-SNP branch.

foxpatch-aleph (approved):
Well-structured SEV-SNP host launch implementation with strong security posture (roothash cmdline-injection barrier, GPU fail-closed, partial-config refusal) and rigorous measurement-correctness guarantees (byte-parity oracle against aleph-tee, fixture-pinned argv, hardcoded cbitpos/reduced-phys-bits matching the measurement). The dispatcher's is_snp/is_snp_marked split cleanly separates complete, partial, and non-SNP configs. Test coverage is thorough across controller, daemon, and conformance layers. Two minor non-blocking observations: the _state parameter in snp_config_slice is unused, and require_rootfs is called twice for SNP VMs.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 1756): The `_state: &DaemonState` parameter is unused in `snp_config_slice`. The sibling `confidential_config_slice` uses it for `state.host.settings.confidential_session_directory`, but SNP has no session. This is fine for signature symmetry, but if it stays unused it could be removed to avoid confusion about whether it reads state.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 1791): `require_rootfs(spec)` is called here inside `snp_config_slice`, and again at line 1533 in `build_written_config` (`let image_path = require_rootfs(spec)?.path.clone()`). For SNP VMs this means the rootfs disk is validated twice. Harmless since the function is pure (just filters `spec.disks`), but slightly redundant.

PR #1036, opened 2026-07-07
feat(rust): B2b - attest-cli verifier, blob-derived trust + verify-before-send (SEV-SNP complete)

odesenfans:
## B2b: aleph-attest-cli client-side verifier (completes SEV-SNP)

Stacks on #1035. Ports the client verifier that connects to a confidential VM's in-guest
attest-agent (B2a), verifies the embedded SEV-SNP attestation, pins the measurement, and
injects secrets over the attested channel. After this, **SEV-SNP is complete** (B0 crypto
lib + B2a image/agent + B1 host launch + B2b verifier).

- Uses the shared, domain-separated `aleph_tee::report_data` scheme (matching the B2a
  agent), so a relayed/key-confusion report is rejected. Inherits AMD **ARK root pinning**
  from B0 automatically.

**Security-hardened beyond the donor** (three review lenses found the donor's verifier
`EXPLOITABLE` two ways; both fixed, and aleph-cvm needs the same upstream):
- **Trust is now blob-derived.** The unsigned JSON `report_data`/`measurement` copies were
  removed from `AttestationReport` entirely (it carries only the AMD-signed `data` blob),
  so gating trust on an attacker-settable copy is structurally impossible. Every check
  parses the signed blob. Previously an attacker could replay any genuine report as `data`
  and set the JSON copies to pass key-binding + measurement pinning.
- **Full AMD chain verified before any secret is sent.** `SnpCertVerifier` is now the sole
  authority: it runs the complete verification (blob-derived key-binding + measurement +
  the KDS/ARK chain via `block_in_place`) at the handshake, so the connection only
  completes for a fully-verified TEE and no secret ever reaches an unverified endpoint.
  Previously the secret POST went on the wire before the chain was checked.

Mutation-confirmed tests: a report with a borrowed-genuine blob but attacker JSON is
rejected; `inject_secret` transmits zero bytes when the chain fails (a local attested-TLS
server records receipt). 410 workspace tests pass, 1 ignored (network KDS). Divergences
71-72 recorded with an upstreaming recommendation.

Deferred (tracked): inject-secret owner-auth (ledger 66) is a separate confidential-flow
design step; this fix ensures secrets only reach a verified TEE, not that the caller is
authorized.


foxpatch-aleph (approved):
This PR correctly implements the aleph-attest-cli verifier and fixes two critical security vulnerabilities from the donor: (1) trust is now blob-derived (unsigned JSON copies removed from AttestationReport, all checks parse the AMD-signed blob), and (2) the full AMD chain is verified during the TLS handshake via block_in_place, so secrets never reach an unverified peer. The domain-separated report_data scheme is shared via aleph-tee, preventing verifier/agent drift. Test coverage is comprehensive with 16 hermetic tests including mutation-killing proofs for both fixes. All downstream files (agent, x509, backend, none backend) are correctly updated for the AttestationReport type change. Minor observations (fresh report measurement not pinned, double KDS fetch, rand version not workspace-shared) are non-blocking and acceptable for this increment.

**`rust/crates/aleph-attest-cli/src/client.rs`** (line 166): The fresh report's measurement is not pinned against `expected_measurement` (only the handshake-embedded report is). This is acceptable since the channel is already verified and measurement is static, but pinning the fresh report's measurement too would be defense-in-depth against a compromised TEE that somehow produces a different measurement for the fresh report.

**`rust/crates/aleph-attest-cli/Cargo.toml`** (line 21): `rand = "0.9"` is declared inline rather than in `[workspace.dependencies]`, introducing a separate version (0.9.4) alongside the existing 0.10.2 in the lock file. Consider promoting this to a workspace dependency if other crates may need `rand` in the future.

foxpatch-aleph (approved):
This PR implements two critical security fixes on top of the B2b attest-cli verifier port: (1) removing unsigned `report_data`/`measurement` copies from `AttestationReport` so trust is structurally impossible to gate on attacker-settable JSON, and (2) moving the full AMD certificate chain verification into the TLS handshake callback via `block_in_place` so secrets are never transmitted to an unverified peer. Both fixes are architecturally sound, the blob-derived trust model is correct, the `block_in_place` pattern is properly used with a multi-threaded tokio runtime, and the test suite is mutation-aware with recording-TLS-server ordering tests and smuggled-JSON bypass tests. The only minor concerns are `rand` version fragmentation (0.9 vs workspace) and silent duplicate-secret-key handling in the HashMap, neither of which is blocking.

**`rust/crates/aleph-attest-cli/src/client.rs`** (line 227): Duplicate secret keys are silently deduplicated by the HashMap. If the CLI user passes `--secret K=v1 --secret K=v2`, only `v2` survives with no warning. Consider detecting duplicates and bailing, or at minimum documenting the behavior.

**`rust/crates/aleph-attest-cli/Cargo.toml`** (line 21): `rand = "0.9"` is specified directly rather than via workspace dependencies, introducing a third rand version into the lock file alongside 0.10.x. The transitive `getrandom 0.3.4` -> `wasip2` -> `wit-bindgen` chain is unusual for a Linux CLI. Consider aligning with the workspace's existing rand version if possible.

**`rust/crates/aleph-attest-cli/src/verify.rs`** (line 111): A new `Arc<CryptoProvider>` is created on every `with_chain_check` call. Since `rustls::crypto::ring::default_provider()` returns a static-ish provider, this is functionally fine but slightly wasteful. Non-blocking.

foxpatch-aleph (approved):
Well-implemented client-side SEV-SNP attestation verifier with two critical security fixes over the donor: (1) removing unsigned report_data/measurement copies from AttestationReport so trust is structurally blob-derived, and (2) running the full AMD chain inside the TLS handshake gate so secrets never reach an unverified peer. The block_in_place pattern for async chain verification from a sync callback is correctly paired with the multi-threaded runtime in main.rs and proven reliable by a test exercising real IO. Test coverage is comprehensive: mutation-killing tests for the gate, smuggled-JSON attack tests proving blob-derived trust, and an end-to-end recording-TLS-server test proving zero secret transmission on chain failure. All AttestationReport construction sites across the workspace have been updated consistently. Minor observations include non-deterministic JSON key ordering from HashMap serialization and no end-to-end test for the fresh_attestation two-report flow, neither of which is blocking.

**`rust/crates/aleph-attest-cli/src/verify.rs`** (line 209): block_in_place panics on a current_thread runtime. This is safe today because main.rs uses #[tokio::main] (multi-threaded), but consider adding a doc comment on SnpCertVerifier noting the multi-thread runtime requirement, so future callers reusing this verifier outside main.rs don't hit a runtime panic.

**`rust/crates/aleph-attest-cli/src/client.rs`** (line 229): Using HashMap for secrets_map means JSON key ordering in the POST body is non-deterministic. This has no security or correctness impact, but if the server logs or hashes the request body, consider using a BTreeMap for deterministic ordering.

**`rust/crates/aleph-attest-cli/src/verify.rs`** (line 155): The TLS certificate's own validity period (not-before/not-after) is not checked — _now is ignored. The AMD chain certs are checked for expiry in verify_cert_chain, but the self-signed attestation cert is not. This is a reasonable simplification for ephemeral agent-generated certs, but worth documenting as deliberate.

PR #1037, opened 2026-07-07
feat(rust): C1 - NUMA topology + pack-first placement + cpuset pinning

odesenfans:
## C1: NUMA topology reporting + pack-first placement + cpuset pinning

Stacks on #1036. First NUMA increment: the supervisor auto-places VMs on NUMA nodes
(pack-first) and pins their vCPUs via the systemd `AllowedCPUs` property. Independent of
the QEMU argv (CPU pinning is a systemd unit property); the memory-backend NUMA binding
and hugepages are C2.

- **`numa.rs`** (ported from aleph-cvm): `NumaTopology` (sysfs parse, cpulist/MemTotal/
  hugepage counts) + `NumaAllocator` (pack-first vCPU placement ledger, cpuset formatting).
- **`HostInfo.numa_nodes`** populated from the detected topology (was empty).
- **Placement** (decision 4, supervisor auto-places): a requested `VmSpec.numa_node` is
  honored (validated), else pack-first; the effective node is reported in `VmInfo.numa_node`
  (was hard-coded `None`).
- **Pinning** via a systemd drop-in (`aleph-vm-controller@{hash}.service.d/numa.conf`,
  `AllowedCPUs={cpuset}`) written + `daemon-reload`ed before start, removed on delete.
- **Reconcile** reconstructs the placement ledger from each adopted unit's `AllowedCPUs`;
  a VM with unknown placement (pre-NUMA) stays unpinned and uncounted, never silently node 0
  (design §8).

**Placement/pinning activate only with >1 NUMA node** (a single-socket `CONFIG_NUMA` host
reports its node but writes no drop-in and reserves nothing, so non-multi-node hosts behave
as before). Empty/failed topology degrades to inert.

Reviewed by three lenses (parity, robustness, test-quality). The review found the core
placement math solid but surfaced ledger-lifecycle bugs, all fixed with regression tests:
the stop→recreate reservation leak, the hidden-VM delete over-release, the single-node
activation, per-node `from_sysfs` fault tolerance, and bounded cpulist ranges. Ledger drift
is confirmed fixed (stop→recreate 2→1; hidden-delete 0→3). No real multi-node hardware
needed (sysfs fixtures + temp unit dir). Divergences 73-74.

`cargo fmt`/`clippy -D warnings`/`test` green.


foxpatch-aleph (commented):
Well-structured NUMA placement implementation with solid pack-first logic, correct ledger lifecycle for the main paths (create/delete/stop/recreate/reconcile), and comprehensive hermetic tests. The >1-node activation gate is a clean backward-compatibility design. One edge-case bug: if readopt_live_controller's network restore step fails after reconstruct_numa_placement registers the VM in the NUMA ledger, the ledger reservation is not released (the entry is removed from world.entries but the ledger keeps counting). On a retry, readopt_live_controller calls reconstruct_numa_placement again, which registers the vCPUs a second time — a ledger double-count. This is narrow (requires a multi-NUMA host AND a network setup failure during readopt retry) but should be addressed. Minor test gap: no coverage for HostInfo.numa_nodes reporting.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 2219): NUMA ledger leak on readopt restore failure: if the `restore` closure fails here, the entry is removed from `world.entries` (line 2221) but the NUMA ledger reservation made by `reconstruct_numa_placement` at line 2165 is NOT released. When the retry loop calls `readopt_live_controller` again, `reconstruct_numa_placement` registers the vCPUs a second time, double-counting. Consider releasing the NUMA ledger reservation alongside the entry removal in this error path, e.g. by capturing `numa_node` before the `restore` closure and calling `release_numa_placement` here. Note: `reconstruct_numa_placement` itself is not idempotent (`register` just adds), so the retry would double-count even if the first registration was somehow cleaned up — the release must happen on failure.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 2165): Consider making `reconstruct_numa_placement` idempotent or guarded against double-registration for the readopt-retry path. The invariant comment in `reconcile_numa_ledger` (line 277) proves the boot-time pass and readopt are mutually exclusive, but does not cover the case where readopt itself is retried after a failure. A simple guard (e.g. checking whether the VM is already in the ledger before registering) would make this robust, though it would require the ledger to track per-VM registrations rather than just per-node counts.

**`rust/crates/supervisor-daemon/src/service.rs`** (line 287): Minor test gap: no test verifies that `GetHostInfo` correctly reports `numa_nodes` from the detected topology. The mapping is straightforward (`node.id`, `node.cpus.len()`, `node.total_ram_mb`), but a test would protect against regressions in the reporting path (independent of the placement path).

foxpatch-aleph (commented):
Solid NUMA topology + pack-first placement + cpuset pinning implementation. The core placement logic, ledger lifecycle (stop-keeps-reservation, recreate-releases-stale-first, boot-failure-cleanup, hidden-VM-delete-no-release), reconcile (unknown=unpinned), single-node inertness, and cpulist OOM guard are all correct and well-tested (11 lifecycle + 14 unit tests). Code quality is high with thorough docs and proper lock discipline (fs read before mutex). One edge-case bug found: a NUMA ledger reservation leak when readopt_live_controller's restore closure fails (the entry is removed but the ledger registration is not released), plus a minor nit on the FailingSystemd test double not overriding reload().

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 2219): NUMA ledger leak on readopt restore failure: when the `restore` closure fails, the entry is removed from `world.entries` (line 2221) but the NUMA ledger reservation registered by `reconstruct_numa_placement` (line 2165) is never released. The `entry` variable is dropped without calling `release_numa_placement(state, node, vcpus)`. This over-counts the node's vCPUs, and on a daemon restart `reconcile_numa_ledger` would re-register the same still-running VM, compounding the leak. Fix: before removing the entry, check `entry.numa_node` and call `release_numa_placement` + `remove_numa_dropin`, mirroring `delete_tracked_vm` (lines 1327-1330).

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 6099): The `FailingSystemd` test double does not override `reload()`, so it uses the default no-op `Ok(())`. This means `apply_numa_dropin`'s `state.units.reload()` succeeds silently instead of recording a `daemon-reload` action. This is fine for the `a_boot_failure_after_placement` test (which doesn't assert on daemon-reload), but worth noting for consistency with `FakeSystemd` which does override `reload()` to record the action.

foxpatch-aleph (commented):

foxpatch-aleph (approved):
Well-structured NUMA topology detection, pack-first vCPU placement, and systemd AllowedCPUs pinning implementation. The core placement math is correct, the ledger lifecycle is carefully managed across all code paths (create/stop/recreate/delete/readopt/reconcile), and every path has dedicated regression tests. The single-node-inert gate, hidden-VM-no-release, and readopt-failure-keeps-dropin behaviors are all correct and tested. No correctness or security bugs found. Minor nits include the overly broad `starts_with("node")` sysfs filter (functionally handled by parse failure) and the lack of a `vcpus == 0` defensive guard, neither of which is blocking.

**`rust/crates/supervisor-daemon/src/numa.rs`** (line 78): The filter `starts_with("node")` matches non-node entries in `/sys/devices/system/node/` like `node_has_memory`, `node_normal`, etc. These are correctly skipped by `parse_node` (the numeric ID parse fails), but it generates spurious warning logs on every boot. A tighter filter like `name.starts_with("node") && name[4..].chars().next().is_some_and(|c| c.is_ascii_digit())` would avoid the noise. Minor nit, not blocking.

**`rust/crates/supervisor-daemon/src/numa.rs`** (line 289): A `vcpus == 0` request would trivially succeed (`0 <= available` is always true) and produce a placement with the full node cpuset. The caller should reject 0-vCPU specs upstream, but a defensive `if vcpus == 0 { return Err(PlacementError::Insufficient(...)); }` here would be safer. Not blocking since 0-vCPU VMs are nonsensical.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 6147): The `FailingSystemd` wrapper re-implements all 8 trait methods just to override `get_active_state`. Consider adding a `get_active_state` override hook to `FakeSystemd` directly (e.g., a closure field), which would simplify this and future similar tests. Test-only nit.

**`rust/crates/supervisor-daemon/src/numa.rs`** (line 458): The `dropin_dir` function uses `vm_hash` directly in the path. If `vm_hash` contained path separators (e.g., `../`), this could traverse outside `systemd_unit_dir`. This is a pre-existing pattern (see `controller_config_path` in `controller_config.rs`), and systemd unit names reject such characters, but a future hardening pass could sanitize `vm_hash` to hex-only characters at the entry point. Not a new vulnerability from this PR.

foxpatch-aleph (approved):
Well-designed NUMA placement increment with thorough lifecycle integration and excellent test coverage. The pack-first allocator, cpuset drop-in mechanism, ledger reconcile, and all edge-case unwinding (stop→recreate, hidden-VM delete, boot failure, readopt failure) are correct. The single-node inertness gate, per-node fault tolerance, and OOM guard are properly implemented. No bugs or security issues found — only minor cosmetic and efficiency observations.

**`rust/crates/supervisor-daemon/src/numa.rs`** (line 466): The `\` line continuation in the `format!` string includes leading whitespace from the next line, producing 9 spaces before `[Service]` and `AllowedCPUs=` in the written INI file. Systemd strips this, so it's cosmetic, but using `concat!` or explicit `\n` joins would produce clean output.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 1295): `remove_numa_dropin` is called unconditionally in the hidden-VM delete path without an `is_placement_active()` gate. On a single-node or non-NUMA host, this triggers a wasted `Manager.Reload()` D-Bus call (the drop-in never existed). The tracked-delete path at line 1332 correctly gates on `entry.numa_node`. Consider gating this call too.

**`tests/conformance/test_rust_daemon.py`** (line 149): The assertion `sum(n.cpu_count for n in info.numa_nodes) == os.cpu_count()` could fail on hosts with offline CPUs, where `os.cpu_count()` returns the online count but the sysfs cpulist may include offline CPUs. This is unlikely in CI but could fail on developer workstations with CPU hotplug.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 5973): The tracked-delete test verifies the drop-in is removed and the ledger released, but doesn't assert that a `daemon-reload` action was recorded by `FakeSystemd` after the removal. Asserting `harness.systemd.actions()` includes `"daemon-reload"` after the delete would confirm the reload path is exercised.

foxpatch-aleph (commented):
The core NUMA topology detection, pack-first placement ledger, and systemd AllowedCPUs drop-in mechanism are well-designed and correctly implemented. The ledger lifecycle is carefully handled across all paths (create, stop, recreate, delete, boot-failure, readopt-failure, reconcile) with regression tests for each. The single-node inertness gate and empty-topology degradation are consistent. The main concern is a conformance test assertion that could fail on hosts with offline CPUs, plus a minor u32 truncation risk for very high-RAM nodes.

**`tests/conformance/test_rust_daemon.py`** (line 149): The assertion `sum(n.cpu_count for n in info.numa_nodes) == os.cpu_count()` can fail on hosts with offline CPUs. The Rust daemon reads `/sys/devices/system/node/nodeN/cpulist`, which contains only online CPUs, while Python's `os.cpu_count()` returns all configured CPUs (including offline). On a host with any offline CPU, the sum of per-node online CPU counts will be less than `os.cpu_count()`. Additionally, `os.cpu_count()` can return `None`, which would cause a `TypeError` in the equality comparison. Consider comparing against the sum of online CPUs from sysfs instead, or using `len(os.sched_getaffinity(0))` for the online count.

**`rust/crates/supervisor-daemon/src/numa.rs`** (line 168): The cast `as u32` silently truncates `total_ram_mb` for nodes with >= 4 TB of RAM (the `u64` MemTotal in kB divided by 1024 overflows `u32`'s ~4M MB max). While unlikely today, this is silent data corruption rather than a graceful error. A `u64` for `total_ram_mb` (matching the proto's `memory_mib: u64`) would avoid the truncation entirely and eliminate the cast.

**`rust/crates/supervisor-daemon/src/numa.rs`** (line 292): When `vcpus == 0`, the condition `0 <= available` is always true, so `allocate(0, None)` succeeds on node 0 and returns a full cpuset drop-in for zero vCPUs. This is not reachable from the QEMU create path (which validates vcpus > 0 earlier), but the allocator itself does not guard against it. A `if vcpus == 0 { return Err(PlacementError::Insufficient(...)); }` at the top would make the function self-documenting.

foxpatch-aleph (approved):
Well-structured NUMA topology reporting + pack-first placement + cpuset pinning implementation. The core placement math is correct, the ledger lifecycle (create, stop, recreate, delete, reconcile, readopt) is carefully handled with consistent lock ordering (world-then-ledger, never held simultaneously), and the stop→recreate double-count, hidden-VM over-release, and reconcile double-count bugs are all fixed with regression tests. The single-node inertness gate (>1 node), per-node fault tolerance in sysfs parsing, and MAX_CPULIST_RANGE OOM guard are sound. Test coverage is comprehensive (11 lifecycle tests + 12 numa unit tests). No new security attack surface (vm_hash path construction matches existing patterns). Minor observations are non-blocking.

**`rust/crates/supervisor-daemon/src/numa.rs`** (line 83): The `starts_with("node")` filter will match non-node files like `node_online` if they existed, but in practice `/sys/devices/system/node/` only has `nodeN/` directories and files like `online`, `has_memory` that don't start with `node`. The `parse_node` function's `strip_prefix("node").parse::<u32>()` would reject any false positive anyway. Not a bug, just noting the filter is slightly broader than strictly necessary.

**`rust/crates/supervisor-daemon/src/numa.rs`** (line 434): `line.contains("MemTotal:")` could theoretically match a hypothetical field like `FooMemTotal:` in a malformed meminfo. In practice, sysfs node meminfo only has `Node N MemTotal:` so this is safe. A more precise check like `line.starts_with("Node ") && line.contains(" MemTotal:")` would be more defensive but is not necessary.

**`tests/conformance/test_rust_daemon.py`** (line 149): The assertion `sum(n.cpu_count for n in info.numa_nodes) == os.cpu_count()` assumes every online CPU belongs to a NUMA node. This is true on standard Linux NUMA hosts, but could fail if CPUs are offline or if the kernel exposes CPUs outside NUMA nodes (rare/never in practice). If this ever flakes in CI, it would indicate a real topology edge worth investigating rather than a code bug.

foxpatch-aleph (approved):
Well-structured NUMA topology detection, pack-first placement, and CPU pinning implementation with comprehensive test coverage. The placement math is correct, ledger-lifecycle bugs are properly fixed with regression tests, fault tolerance is thorough (per-node sysfs skip, bounded cpulist ranges, inert degradation), and the >1-node gate makes single-node hosts behave exactly as before C1. Security is clean (no path traversal or INI injection risk). One minor theoretical edge case in create_program_vm where a stale QEMU entry's NUMA reservation could leak if the same hash were reused for a Firecracker program, but this is near-impossible in practice since hashes are content-derived.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 2746): Minor: `create_program_vm` removes a stale entry but only captures `stale_ordinal`, discarding `stale.numa_node`. If a QEMU VM with a NUMA reservation were somehow replaced by a Firecracker program with the same hash, the ledger reservation would leak. In practice this is near-impossible since VM hashes are content-derived (different backends produce different hashes), but for symmetry with the QEMU create path's `stale_numa` handling at line 2455, consider capturing and releasing the stale NUMA reservation here too, or adding a comment noting why it's safe to skip.

foxpatch-aleph (approved):
Well-structured NUMA placement + CPU pinning implementation with correct ledger lifecycle across all code paths, careful lock ordering (no nested locks, no deadlock risk), comprehensive hermetic test coverage, and proper defensive design (inert on single-node, graceful degradation on detection failure, bounded cpulist ranges). The core pack-first placement math is sound, the stop-then-recreate double-count fix is correct, and the reconcile-treats-unknown-as-unpinned design decision is the safe choice. No correctness bugs, security issues, or missing test coverage found.

**`rust/crates/supervisor-daemon/src/numa.rs`** (line 80): The `starts_with("node")` filter also matches non-node entries if any existed (e.g. a hypothetical `node_active` file). This is safe because `parse_node` validates the suffix parses as `u32`, but a more specific filter like `starts_with("node") && ends_with digit` or a regex would be more self-documenting. Minor nit, not blocking.

**`rust/crates/supervisor-daemon/src/numa.rs`** (line 439): `parse_memtotal_kb` uses `.rev().nth(1)` to get the second-from-last whitespace-delimited token (the value before "kB"). This works for the kernel's standard format `Node N MemTotal:       XXXXX kB` but would break if the unit suffix were ever absent. The kernel always includes "kB" so this is fine, but a more robust approach would find the numeric token on the line. Not blocking.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 1304): On the hidden-VM delete path, `remove_numa_dropin` has an internal `is_placement_active` guard. If the daemon restarted with a different (now single-node or failed) topology, a previously-written drop-in from a multi-node era would not be cleaned up here. This is an acceptable edge case (the controller is being stopped/disabled anyway, so the stale drop-in is harmless), and the comment at line 1287-1294 acknowledges the ledger is deliberately not touched. Just noting for awareness.

**`rust/crates/supervisor-daemon/src/numa.rs`** (line 292): The `allocate` function returns the entire node's CPU set as the cpuset (e.g. "0-3" for a 1-vCPU VM on a 4-CPU node), not a subset of individual CPUs. This is correct NUMA-aware pinning (pin to the node, let the kernel schedule within it) and matches the donor. The ledger prevents vCPU over-subscription at the node level. Just confirming this is intentional and well-designed.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 2527): The stale reservation is released before `place_vm_numa` re-allocates, but the stale drop-in is left in place (overwritten later by `apply_numa_dropin`). If `place_vm_numa` fails (returns Err), the entry is removed but the stale drop-in remains on disk. The comment at line 2524-2526 explains this is intentional (a placement failure unwinds without a boot, so no stale pin is left applied). However, the old drop-in from the *previous* placement could still be on disk. Since the VM entry is removed on failure and the controller was never started, this is a minor disk leak (cleaned up on next create or delete). Not blocking.

foxpatch-aleph (approved):
A well-structured NUMA topology detection, pack-first placement, and CPU pinning implementation. The core placement math in NumaAllocator is correct, the lifecycle integration is careful (stop→recreate leak, hidden-VM delete over-release, single-node inertness, readopt failure cleanup are all handled with regression tests), and the lock ordering is consistent (ledger and world locks are never held simultaneously). The >1-node activation gate ensures backward compatibility for single-node/non-NUMA hosts. The sysfs parsing is fault-tolerant (per-node skip on failure, bounded cpulist ranges to prevent OOM). The systemd drop-in mechanism is a sound choice for testability, restart-durability, and reconcile. The conformance test correctly validates against the same sysfs source the daemon reads. Minor nits below are non-blocking.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 237): Minor nit: `remove_numa_dropin` always calls `state.units.reload()` after `remove_cpuset_dropin` returns `Ok(())`, even when the drop-in directory was `NotFound` (idempotent success). This means a hidden-VM delete on a multi-node host issues a daemon-reload even if the VM never had a drop-in. Not incorrect (a no-op reload is harmless), but slightly wasteful. Consider checking whether the directory actually existed before reloading.

**`rust/crates/supervisor-daemon/src/numa.rs`** (line 292): Observation: `allocate` returns the FULL node cpuset (`format_cpuset(&node.cpus)`), not a per-VM subset. All VMs on the same node share the same `AllowedCPUs` value. The ledger tracks vCPU counts for capacity accounting but does not subdivide the cpuset. This is a valid design choice (node-level pinning), but it's worth noting that a VM requesting 1 vCPU on a 16-CPU node gets pinned to all 16 CPUs of that node. If per-VM CPU isolation is desired in a future increment, the allocator would need to assign specific CPU IDs.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 2843): Suggestion: there's a test for `stop_keeps_the_numa_reservation` but no test verifying that `start_vm` after stop preserves `entry.numa_node` and does not re-place. While this is implicitly correct (start_vm doesn't touch numa_node), a small assertion test would make the invariant explicit and guard against future refactors that might accidentally re-place on start.

foxpatch-aleph (approved):
Well-structured NUMA topology detection and pack-first vCPU placement with systemd AllowedCPUs drop-in pinning. The core placement math is correct, the ledger-lifecycle bugs (stop→recreate leak, hidden-VM over-release, single-node activation, readopt double-count) are all properly addressed with regression tests, and the invariant documentation on the reconcile/readopt disjointness is thorough. The single-node inertness gate is correctly designed so non-multi-node hosts behave identically to pre-C1. Test coverage is comprehensive across topology parsing, allocator behavior, drop-in helpers, and lifecycle integration including error paths. One minor non-blocking observation: an orphaned drop-in file may remain on disk if a VM is reconciled as unpinned (cpuset no longer matches any node after a topology change) and later deleted, since delete_tracked_vm only removes the drop-in when numa_node is Some. The theoretical u32 overflow in format_cpuset (range_end + 1 at line 381) is unreachable in practice due to the 4096-CPU bound in parse_cpulist.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 1345): Minor: if a VM was reconciled as unpinned (its drop-in cpuset no longer matches any node after a topology change between restarts), `entry.numa_node` is None, so `remove_numa_dropin` is skipped and the orphaned drop-in file stays on disk forever. Consider always removing the drop-in in delete_tracked_vm (outside the `if let Some(node)` guard), or having `reconstruct_numa_placement` clean up a non-matching drop-in when it returns None. Non-blocking — the orphaned file is harmless.

**`rust/crates/supervisor-daemon/src/numa.rs`** (line 381): Theoretical: `range_end + 1` would overflow if `range_end == u32::MAX`. Unreachable in practice since `parse_cpulist` bounds ranges at 4096 CPUs, but using `range_end.saturating_add(1)` or `cpu == range_end && range_end < u32::MAX` would make the intent explicit. Non-blocking.

odesenfans:
Addressing the two non-blocking review findings:

**Orphaned drop-in for a reconciled-unpinned VM (lifecycle.rs)**: fixed in 6f4617fc. The orphan was less harmless than stated: systemd applies drop-ins unconditionally at unit load, and create only overwrites the file when placement is active, so a same-hash redeploy on a host whose topology went inert would silently boot pinned to a dead topology's cpuset. Rather than gating differently in `delete_tracked_vm`, `remove_numa_dropin` is now called unconditionally on tracked delete and no longer gated on `is_placement_active` (which the suggested fix would still have missed for a multi-to-single-node reshape); `remove_cpuset_dropin` reports whether a directory was actually removed so the daemon-reload only runs when on-disk state changed, keeping every unpinned delete reload-free. Cleaning up inside `reconstruct_numa_placement` was rejected: at reconcile time the unit is loaded and running with the pin applied, and removing the file + reload would live-unpin a running adopted VM at daemon boot, against the observe-don't-mutate adoption stance.

**`range_end + 1` overflow (numa.rs)**: declining the change, with a stronger argument than the 4096 bound (which only applies to `lo-hi` ranges; the single-value parse branch accepts a bare `4294967295`). `format_cpuset` iterates a `BTreeSet<u32>`, ascending and distinct, so in the loop body `cpu > range_end` always holds and `range_end <= u32::MAX - 1` whenever `range_end + 1` is evaluated; `u32::MAX` can only be the last element and the expression is never evaluated after it. Overflow is impossible by construction, and `saturating_add` would imply a reachable saturation case that does not exist.

foxpatch-aleph (approved):
Solid NUMA topology detection and pack-first vCPU placement implementation. The core placement math is correct, the ledger lifecycle is carefully managed across all paths (create, stop, recreate, delete, reconcile, readopt, boot-failure), and the single-node inertness gate ensures backward compatibility. Fault tolerance in sysfs parsing (skip bad nodes, OOM-bounded cpulist ranges) and the drop-in-file mechanism (testable, durable, reconcilable) are well-designed. Test coverage is comprehensive with hermetic fixtures. No correctness bugs or security issues found.

**`rust/crates/supervisor-daemon/src/numa.rs`** (line 466): The `format!` with `\` line continuations embeds the leading whitespace from the source indentation into the INI body (e.g. `         [Service]`). systemd tolerates leading whitespace before keys, so this works, but it is slightly unusual. A raw string or explicit `\n` joins would make the output cleaner.

**`rust/crates/supervisor-daemon/src/numa.rs`** (line 435): `line.contains("MemTotal:")` is a substring match. A hypothetical field like `MemTotal_Cached:` would also match. The real sysfs format uses `Node N MemTotal:` so this is fine in practice, but a more precise match (e.g. `line.starts_with("Node ") && line.contains(" MemTotal:")` or splitting on whitespace and checking for `"MemTotal:"` as a token) would be more robust.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 2544): If `request.vcpus` is 0, `place_vm_numa` would still return a placement (0 <= available is always true) and write a drop-in pinning a 0-vCPU VM to the full node cpuset. This is a harmless edge case (the VM has no vCPUs to pin), but worth noting if a 0-vCPU spec is ever possible.

PR #1038, opened 2026-07-07
feat(rust): C2 - NUMA memory binding + opt-in hugepages (Phase 3 code complete)

odesenfans:
## C2: NUMA memory binding + opt-in hugepages (Phase 3 code complete)

Stacks on #1037. Final NUMA increment: binds a placed VM's memory to its node and
optionally backs it with hugepages. After this, all of Phase 3's code is complete
(foundation + SEV-SNP + NUMA).

- **Memory node binding (controller argv):** when a VM has an effective NUMA placement
  (C1 chose a node, active only on >1-node hosts), the memory-backend is bound to that
  node: plain/SEV emit `-object memory-backend-memfd,id=pcram,size={mem}M,share=on,
  host-nodes={node},policy=bind` wired onto the effective `-machine`; SNP's existing
  `ram1` memfd gains `host-nodes/policy=bind`. **Critical parity: with no placement
  (single-node/no-NUMA hosts), the argv is byte-identical to pre-C2 on every path.**
- **Hugepages (opt-in):** `ALEPH_VM_NUMA_HUGEPAGES` (default OFF). Off = regular-page
  NUMA binding only. On (and >1 node): a headroom-honoring per-node reservation
  (8 GiB/node floor, saturating math, fail-safe fallback to regular pages, never blocks
  creation) plus `hugetlb` in the backend and 1G/2M page selection.
- **SNP oracle stays consistent:** the SNP fragment's suffix (`hugetlb` then
  `host-nodes/policy`) is byte-checked against the aleph-tee generator across the
  numa/hugepage combos, so the launch/measurement path is unaffected.

Reviewed by three lenses (15/16 mutations caught; all critical parity properties pinned).
Fixes: made the hugepage pool accounting exact on release and adoption (was leaking on
delete / resetting on restart, opt-in only), pinned the opt-in selection gate with a
mutation-killing test, and tightened the invariant/doc nits. `numa_node`/`hugepage_size`
are skip-if-None in the controller JSON, so non-NUMA configs stay byte-identical to the
pydantic writer. Divergences 75-76.

`cargo fmt`/`clippy -D warnings`/`test` green (470 tests).


foxpatch-aleph (approved):
Well-structured PR implementing NUMA memory binding and opt-in hugepage backing for VMs. The critical parity property (single-node/no-NUMA hosts produce byte-identical argv to pre-C2) is correctly maintained across all three launch paths (plain, SEV/SEV-ES, SNP) via the `numa_node.is_none()` gate and `skip_serializing_if` on the written config. Hugepage pool accounting is symmetric across reserve/release/reregister using a single `pages_for` definition, preventing leaks on delete, failed-create, and daemon restart. The SNP fragment's hugetlb-then-host-nodes ordering is byte-verified against the aleph-tee generator oracle across all 4 numa/hugepage combinations. The opt-in gate (`ALEPH_VM_NUMA_HUGEPAGES`, default OFF) is properly tested with a mutation-killing test. Fail-safe behavior (per-node write failure skips to regular pages, never blocks boot) is tested. No security issues: NUMA node is a u32 formatted into QEMU argv (no shell interpolation), hugepage_size is a controlled enum, and sysfs writes are opt-in. Test coverage is comprehensive across all new code paths.

**`rust/crates/supervisor-controller/src/qemu.rs`** (line 191): Pre-existing (not introduced by this PR): the plain/SEV paths use `share=on` while the SNP path uses `share=true`. Both are accepted by QEMU, and the SNP path must match the aleph-tee generator's `share=true`. No action needed, just noting the discrepancy is intentional.

foxpatch-aleph (approved):
Well-structured NUMA memory binding + opt-in hugepages implementation with strong parity guarantees, comprehensive test coverage, and careful fail-safe semantics throughout. The byte-identical-to-pre-C2 invariant for non-NUMA hosts is properly enforced and tested. The hugepage pool accounting is exact across reserve/release/reregister cycles. One minor non-blocking observation: reregister_hugepages uses plain += instead of saturating_add, but this is practically safe given the pool is bounded by physical hugepages.

**`rust/crates/supervisor-daemon/src/numa.rs`** (line 456): Consider using saturating_add instead of plain += in reregister_hugepages, matching the saturating pattern used in release_hugepages. A corrupt config with a huge memory_mb could theoretically overflow u32 in release builds (wrap) or panic in debug. Not a real-world risk since the pool is bounded by physical hugepages, but saturating_add would be more defensive.

foxpatch-aleph (approved):
This PR implements NUMA memory binding and opt-in hugepages (C2) with careful attention to byte-identical parity for non-NUMA hosts. The implementation is correct: the memory backend suffix ordering (hugetlb before host-nodes) matches the aleph-tee generator oracle, the hugepage pool accounting is exact across create/delete/restart lifecycle paths (reserve, release, and reregister all use the same `pages_for` definition), the opt-in gate is properly tested with a mutation-killing test, and the fail-safe fallback to regular pages never blocks VM creation. The parity properties (no memory backend when numa_node is absent) are well-pinned by both existing conformance fixtures and explicit tests across all three launch paths (plain, SEV/SEV-ES, SNP). The code is thoroughly documented with clear rationale for each design decision. No correctness, security, or race condition issues were found.

**`rust/crates/supervisor-controller/src/qemu.rs`** (line 411): Minor: `memory_mb.is_multiple_of(1024)` was stabilized in Rust 1.87. No MSRV is pinned in Cargo.toml, and the PR reports tests green, so the CI toolchain supports it. If the project ever needs to support older toolchains, this would need replacing with `memory_mb % 1024 == 0`. Not a blocker since tests pass.

**`rust/crates/supervisor-controller/src/qemu.rs`** (line 966): Minor test gap: `sev_config` only accepts `numa_node`, not `hugepage_size`. The SEV + hugepages path (`build_confidential_argv` with both `numa_node` and `hugepage_size` set) is not directly tested, though the shared `memory_backend_suffix` is covered via the plain path. Adding a `sev_with_numa_and_hugepages` test would close the gap.

**`rust/crates/supervisor-daemon/src/numa.rs`** (line 460): Minor: `reregister_hugepages` uses `+=` (not `saturating_add`). If a topology change reduced a node's pool below the sum of all adopted VMs' registered pages, the pool could theoretically overflow u32. In practice this requires ~8 TiB of 2M pages or ~1 PiB of 1G pages, so it is not a real-world concern. The `reserve_hugepages` function checks `available` before allocating, so an over-capacity pool just falls back to regular pages.

**`rust/crates/supervisor-daemon/src/hugepages.rs`** (line 137): Nit: when `allocate_2m_pages_on_node` succeeds but the kernel grants fewer pages than requested (fragmentation), the topology is updated with `actual` (line 137). When the write FAILS, the node's `total_2m_hugepages` is left at its pre-existing value (0 from detection, or whatever was parsed). This means the allocator's pool for that node reflects the detected count, not the failed-reservation count. This is the correct fail-safe behavior (fall back to regular pages), but a brief comment noting that the pool stays at the detected value on failure would make the intent clearer.

foxpatch-aleph (approved):
Well-structured PR with strong parity guarantees (byte-identical argv for non-NUMA hosts), exact hugepage pool accounting across all lifecycle paths (create/delete/failed-create/reconcile/restart), and a correctly pinned opt-in gate. The fail-safe fallback to regular pages is consistently applied, and the SNP oracle test now crosses all numa/hugepage combos. The shared `memory_backend_suffix` and single `pages_for` definition ensure reserve/release/reregister all use the same page counts, preventing pool leaks. The only minor gaps are: (1) no SEV/SEV-ES hugepage test (the suffix logic is shared and tested via the plain path), (2) `compute_2m_budget` can theoretically return `u32::MAX` on overflow (fail-safe via kernel read-back), and (3) `pages_for(Size1G)` uses truncating division which is correct under the allocator's 1G-alignment invariant but would be more robust with `div_ceil` against externally-modified configs. None of these are blocking.

**`rust/crates/supervisor-daemon/src/hugepages.rs`** (line 51): Minor: `u32::try_from(budget_mb / 2).unwrap_or(u32::MAX)` can return `u32::MAX` on a theoretical overflow, which would then be written to `nr_hugepages` sysfs. The kernel grants what it can and the read-back gets the actual count, so this is fail-safe. Consider capping to the node's actual RAM in pages instead, for clarity.

**`rust/crates/supervisor-daemon/src/numa.rs`** (line 252): Latent robustness note: `Size1G` uses `memory_mb / 1024` (truncating division). This is correct because the allocator only selects 1G for 1G-aligned memory (`is_multiple_of(1024)` check in `reserve_hugepages`). However, if a config on disk is ever externally modified to have `hugepage_size: "1G"` with non-1G-aligned `mem_size_mb`, `release_hugepages`/`reregister_hugepages` would compute fewer pages than were actually reserved, leaking the pool. Using `div_ceil` here would be more defensive against such edge cases. Not a bug under normal operation.

**`rust/crates/supervisor-controller/src/qemu.rs`** (line 1000): Minor test coverage gap: `sev_config` does not accept a `hugepage_size` parameter, so `build_confidential_argv` with hugepages is untested. The `memory_backend_suffix` is shared and well-tested via the plain path, so this is non-blocking. Consider adding a `sev_config_with_hugepages` test for completeness.

foxpatch-aleph (approved):
This PR adds NUMA memory binding and opt-in hugepage backing (C2) to the Rust supervisor. The implementation is thorough and well-reasoned: byte-identical parity is maintained for all non-NUMA hosts (numa_node absent = no memory backend, no argv change), hugepages are opt-in with fail-safe fallback to regular pages on every error path, and the hugepage pool accounting is exact across create/delete/failed-create/restart lifecycle. The test coverage is comprehensive, including allocator unit tests, hugepage budget computation, fail-safe reservation, pool round-trip (no leak), adoption re-registration, and controller argv parity for all three launch paths (plain, SEV/SEV-ES, SNP) crossed with the NUMA/hugepage combinations. The SNP fragment oracle test now crosses 4 NUMA/hugepage cases against the aleph-tee generator. The daemon-side config serialization correctly uses skip_serializing_if=None so non-NUMA configs stay byte-identical to the pydantic writer. No correctness, security, or testing gaps were found.

**`rust/crates/supervisor-daemon/src/numa.rs`** (line 460): Minor: `reregister_hugepages` uses plain `+=` while `release_hugepages` uses `saturating_sub`. In practice this is safe because registered pages are bounded by the node's pool size, but using `saturating_add` would be more defensive against a topology change between restarts (where a node's hugepage count was reduced, causing `allocated` to exceed `total`). Not blocking.

**`rust/crates/supervisor-daemon/src/hugepages.rs`** (line 112): Minor: `per_node_cap` uses integer division (`effective_limit / num_nodes`), so on a host where total RAM is not evenly divisible by the node count, a small remainder of RAM is unused for the 2M reservation. This is negligible (at most `num_nodes - 1` MB) and matches the donor's approach, but worth noting for documentation.

PR #1039, opened 2026-07-07
feat(supervisor-cli): alephctl debug CLI for the supervisor gRPC socket

odesenfans:
Stacked on #1034 (B2a / `od/phase3-b2a-nix-measured-image`).

Ships `alephctl`, a Rust CLI in the aleph-vm workspace that speaks the `aleph.supervisor.v1` gRPC contract over the supervisor's Unix socket for on-node debugging and testing. Works against the Python and Rust supervisors alike (same contract, same socket).

## Commands (v1: debug essentials)

```
alephctl health
alephctl host-info
alephctl vm list
alephctl vm get <id>
alephctl vm spec <id>
alephctl vm start <id>
alephctl vm stop <id>
alephctl vm reboot <id>
alephctl vm delete <id> [--yes]     # prompts for confirmation unless --yes
alephctl logs <id> [--follow] [--tail N]
alephctl events                     # WatchEvents, streams until Ctrl-C
alephctl ports list [<id>]
```

Global flags: `--socket PATH` (default resolves via `ALEPH_VM_SUPERVISOR_GRPC_SOCKET` → `$ALEPH_VM_EXECUTION_ROOT/supervisor.sock` → `/var/lib/aleph/vm/supervisor.sock`, mirroring the daemon's config resolution), and `--json` for raw proto responses as JSON (streams emit NDJSON).

## What's in the PR

- `supervisor-proto`: `#[derive(serde::Serialize)]` on every generated type via `build.rs` `type_attribute` (generated Rust only; the wire contract is untouched). `--json` is free across all commands.
- New crate `rust/crates/supervisor-cli` (lib + thin `alephctl` bin). Command handlers take a connected `SupervisorClient` and an `impl Write`, so integration tests drive them against an in-process fake `Supervisor` tonic server on a temp-dir Unix socket — no KVM, no real supervisor.
- `packaging/Makefile`: installs `alephctl` to `/opt/aleph-vm/bin/alephctl` next to the supervisor and controller.

## Testing

- 35 tests: 16 unit (clap parsing, socket resolution, output helpers, gRPC error rendering) + 17 integration (every command against the fake supervisor, with canned data and exact output assertions) + 2 supervisor-proto JSON serialization.
- Full workspace builds (daemon + controller still build against the serde-annotated proto types). `cargo fmt --check` and `cargo clippy -p supervisor-cli --all-targets -- -D warnings` clean.

## Notes

- The fake supervisor's unimplemented RPC stubs are written out explicitly (no macro): tonic's `asyncTrait` rewrites the impl block before declarative macros expand, so macro-generated `async fn`s wouldn't be desugared — matches the convention already in `rust/crates/supervisor-daemon/src/service.rs`.
- Streaming commands rely on default SIGINT behavior (process termination); no custom Ctrl-C handler (YAGNI).
- Design doc: `docs/plans/2026-07-07-alephctl-supervisor-cli-design.md` (marked implemented). Implementation plan: `docs/plans/2026-07-07-alephctl-supervisor-cli-implementation.md`.

### DIFF
--- a/docs/plans/2026-07-07-alephctl-supervisor-cli-design.md
+++ b/docs/plans/2026-07-07-alephctl-supervisor-cli-design.md
@@ -1,0 +1,112 @@
+# alephctl: supervisor debug CLI (design)
+
+**Date:** 2026-07-07
+**Status:** implemented (see docs/plans/2026-07-07-alephctl-supervisor-cli-implementation.md)
+**Target:** Rust workspace tip (stacked on the phase3 / rust-daemon PR stack)
+
+## Purpose
+
+A Rust CLI shipped with aleph-vm that speaks the `aleph.supervisor.v1` gRPC
+contract directly over the supervisor's Unix socket, for immediate debugging
+and testing on a node. The Python and Rust supervisors serve the same
+contract on the same socket, so `alephctl` works against both without
+knowing which one is running.
+
+Primary users: node operators and developers debugging a live node, and
+developers exercising supervisor RPCs during development without going
+through the agent/orchestrator layer.
+
+## Placement
+
+- New crate: `rust/crates/supervisor-cli`
+- Binary name: `alephctl`
+- Ships as its own PR stacked on the current Rust workspace tip.
+- Depends on `supervisor-proto` for the generated tonic stubs
+  (`build_client(true)` is already enabled).
+- The UDS connector helper (dummy-URI endpoint + `UnixStream` connector,
+  same pattern as aleph-bench's `connect_uds`) lives in the CLI crate.
+  `supervisor-proto` stays a pure contract crate.
+
+## Connection resolution
+
+Socket path is resolved in this order, mirroring the daemon's own config
+resolution so the CLI finds the right socket on any correctly configured
+node with zero flags:
+
+1. `--socket PATH` flag
+2. `ALEPH_VM_SUPERVISOR_GRPC_SOCKET` environment variable (empty string =
+   unset, matching daemon behavior; the daemon reads `ALEPH_VM_*`-prefixed
+   variables, mirroring pydantic's `env_prefix`)
+3. `$ALEPH_VM_EXECUTION_ROOT/supervisor.sock`
+4. `/var/lib/aleph/vm/supervisor.sock` (default `EXECUTION_ROOT`)
+
+Connection failure produces a clear "is the supervisor running at
+\<path\>?" message on stderr, not a tonic error chain.
+
+## Commands (v1: debug essentials)
+
+```
+alephctl health
+alephctl host-info
+alephctl vm list
+alephctl vm get <id>
+alephctl vm spec <id>
+alephctl vm start <id>
+alephctl vm stop <id>
+alephctl vm reboot <id>
+alephctl vm delete <id> [--yes]     # prompts for confirmation unless --yes
+alephctl logs <id> [--follow] [--tail N]
+alephctl events                     # WatchEvents, streams until Ctrl-C
+alephctl ports list [<id>]
+```
+
+RPC mapping: `Health`, `GetHostInfo`, `ListVms`, `GetVm`, `GetVmSpec`,
+`StartVm`, `StopVm`, `RebootVm`, `DeleteVm`, `GetLogs`/`StreamLogs`,
+`WatchEvents`, `ListPortForwards`.
+
+Out of scope for v1 (explicitly deferred, clap structure leaves room for
+them as later subcommands): backups, confidential ops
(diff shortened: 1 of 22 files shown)


foxpatch-aleph (commented):

foxpatch-aleph (approved):
This is a large, well-structured PR adding the alephctl debug CLI, a Rust supervisor-controller, SEV-SNP TEE attestation/verification (aleph-tee), and an attestation agent (aleph-attest-agent). The security-sensitive code is particularly well done: domain-separated report_data construction closes a real MITM attack from the donor codebase, the cert chain verifies against a pinned AMD ARK (not a caller-supplied root), secret injection uses O_EXCL|O_NOFOLLOW with symlink/ownership/permission checks and a one-shot mutex, and the reverse proxy strips hop-by-hop headers to prevent request smuggling. Product names are allowlisted before URL interpolation. Tests are comprehensive across all new crates. The only issues found are cosmetic (a duplicated step number in a comment).

**`rust/crates/aleph-tee/src/sev_snp/verify.rs`** (line 105): Comment step numbering: line 100 says "2. Enforce the privileged-VMPL gate" and line 105 says "2. Extract chip_id and TCB version". The second "2." should be "3." (or the steps should be renumbered to match the actual sequence).

**`rust/crates/aleph-attest-agent/src/proxy.rs`** (line 96): The upstream URL is built via raw string concatenation: `format!("{}{path}", state.upstream)`. If the operator-configured `upstream` has a trailing path (e.g. `http://127.0.0.1:8080/api`), a request to `/` becomes `http://127.0.0.1:8080/api/` which is likely fine, but consider using `reqwest::Url::join` or `Url::parse` + `join` for correct RFC 3986 path resolution, especially for edge cases with `..` or `//` in the request path.

foxpatch-aleph (commented):
Well-structured, clean implementation of the alephctl debug CLI. The code is correct, the test harness with an in-process fake supervisor over a real Unix socket is solid, and the serde::Serialize derive on proto types is a clean approach for --json output. No security issues found. A few minor observations: (1) the vm start and vm reboot lifecycle commands lack integration tests (only vm stop is tested), (2) combining --json without --yes on vm delete would intermix the confirmation prompt text with JSON output on stdout — though this is an unusual edge case for a debug tool, and (3) the write_table function silently drops extra columns if a row is shorter than headers, but all current callers produce correctly-sized rows so this is not a practical issue.

**`rust/crates/supervisor-cli/tests/cli.rs`** (line 203): Integration tests cover `vm stop` but not `vm start` or `vm reboot`. These follow the same `report_transition` pattern, but a quick test for each would ensure the verb strings ("started", "rebooted") are correct and not accidentally swapped.

**`rust/crates/supervisor-cli/src/commands/vm.rs`** (line 234): When `--json` is set but `--yes` is not, the confirmation prompt ("Delete VM {vm_id}? [y/N] ") is written to `out` (stdout) before the JSON response, which would corrupt piped JSON output. Consider either skipping the prompt when `json` is true (assume the user knows what they're doing in script mode) or writing the prompt to stderr instead.

foxpatch-aleph (approved):
Well-structured PR adding alephctl, a Rust debug CLI for the supervisor gRPC socket. The code is clean, consistent with existing workspace conventions, and thoroughly tested (35 tests covering all commands, error rendering, socket resolution, and JSON output). The serde::Serialize derives on supervisor-proto are a minimal, non-invasive change that doesn't affect the wire contract. The fake supervisor test harness is well-designed and mirrors the daemon's own service.rs convention. No correctness, security, or quality issues found. A few minor observations below are non-blocking nits.

**`rust/crates/supervisor-cli/src/output.rs`** (line 24): The `render` closure zips `cells` with `widths`, so if a row has more cells than headers, the extra cells are silently dropped. This is fine for current usage (all callers construct rows to match headers), but a debug assertion or a comment noting the contract (rows must match header count) would make the implicit assumption explicit.

**`rust/crates/supervisor-cli/src/commands/vm.rs`** (line 238): The `delete` command calls `input.read_line()` (blocking) inside an async function. In `main.rs` this runs on the tokio runtime, but since no other async work is concurrent at this point (the gRPC call happens after the prompt), this is acceptable for a CLI tool. If this ever moves to a context with concurrent async work, consider `tokio::io::stdin()` instead.

**`rust/crates/supervisor-cli/src/commands/logs.rs`** (line 26): `logs --follow` uses `include_history: false`, meaning it only streams new lines (like `journalctl -f` without history). This is a deliberate design choice documented in the implementation plan. Users expecting `tail -f` semantics (which shows recent history then follows) might be surprised. Consider mentioning this in `--help` text or the `Logs` variant's doc comment.

PR #1041, opened 2026-07-07
[DO NOT MERGE] Phase 3 testnets candidate deb build (full stack A1..D1)

odesenfans:
## DO NOT MERGE - candidate deb build for Phase 3 testnets validation

This PR exists **only to trigger the deb build workflow** for the full Phase 3 stack
(branch `od/phase3-d1-snp-arch-agnostic`, which contains A1..D1: the Rust supervisor,
SEV-SNP, NUMA, and the arch-agnostic cbitpos fix), so the aleph-testnets upgrade-checks
harness can fetch `aleph-vm.debian-12.deb` as the candidate for the SNP attestation +
NUMA validation run.

The real, reviewable increments are the stacked PRs #1030-1040. **This PR must not be
merged** - it targets dev only to run `build-deb-package-and-integration-tests.yml` on
the candidate branch.


foxpatch-aleph (commented):
Large, well-structured Phase 3 implementation (SEV-SNP attestation, NUMA, hugepages, attestation agent/CLI, Nix measured-boot). Crypto design is sound (ARK pinning, VMPL gate, report_data binding, fail-closed verification). Several security issues need attention before production: unsanitized vm_id reaching filesystem paths and systemd unit names, proxy forwarding of X-Forwarded-* headers, unvalidated attestation report policy field, unbounded path_on_host in erase_volumes, and error-string information disclosure. The Nix flake.lock IS committed (reproducibility is fine). The PR is explicitly DO NOT MERGE (CI build trigger for stacked PRs), so these comments apply to the stacked increments.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 2411): SECURITY: `vm_id` from the gRPC `VmSpec` is used directly in `controller_unit_name` (line 2440), `controller_config_path` (line 2438), and D-Bus calls without any character validation. A `vm_id` containing `/`, `..`, or newlines could traverse out of `execution_root`, inject into systemd unit names, or break drop-in INI format. This is bug-for-bug parity with the Python daemon (which also doesn't validate vm_id), but it should be hardened — add a regex/charset check (e.g. `[a-f0-9]` or `[a-zA-Z0-9_-]`) before the id is used in any path or unit name.

**`rust/crates/aleph-attest-agent/src/proxy.rs`** (line 112): SECURITY: The proxy forwards all non-hop-by-hop headers verbatim, including `X-Forwarded-For`, `X-Forwarded-Host`, `X-Real-IP`, and `Forwarded`. An external client can spoof source IP/hostname as seen by the in-VM application. If the in-VM app makes authorization decisions based on `X-Forwarded-For`, this is an authentication-bypass vector. Consider stripping `X-Forwarded-*`/`Forwarded` headers and re-injecting values the agent vouches for, or documenting that the upstream must not trust these headers.

**`rust/crates/aleph-tee/src/sev_snp/verify.rs`** (line 91): SECURITY: `verify_sev_snp_report` does not validate the attestation report's `policy` field. A report from a debug-enabled guest (policy debug bit set) would pass verification. The QEMU default policy `0x30000` disables debug, but the attestation report's actual policy (AMD-signed, reflecting what the guest launched with) is never checked against the expected value. Consider adding a policy validation step, at minimum checking that the debug bit is off.

**`rust/crates/aleph-tee/src/sev_snp/verify.rs`** (line 154): SECURITY: `guest_svn` is surfaced in the result details but not compared to any minimum. An old-SVN guest (with known vulnerabilities) would still verify as `valid: true`. Consider adding a minimum-SVN check parameter or documenting that callers must enforce this.

**`rust/crates/aleph-tee/src/sev_snp/verify.rs`** (line 79): SECURITY: `verify_sev_snp_report` does not verify that `report_data` matches a caller-supplied nonce. The doc comment defers this to the caller, but a replayed genuine report (same report_data) passes `valid: true`. Consider offering a `verify_with_nonce` variant or making the replay warning more prominent in the API contract.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 1112): SECURITY: `erase_volumes` calls `std::fs::remove_file(path_on_host)` where `path_on_host` comes directly from the on-disk controller config JSON with no check that it is under `execution_root` or `persistent_volumes_dir`. A malicious or corrupted config with `path_on_host = "/etc/passwd"` would cause arbitrary file deletion. Add a path containment check (canonicalize and verify it starts with the expected prefix) before unlinking.

**`rust/crates/aleph-attest-agent/src/proxy.rs`** (line 70): SECURITY: Error responses echo internal anyhow error chains directly to external clients (e.g. `format!("attestation failed: {e:#}")`). This leaks implementation details, file paths, and TEE backend error text. Map internal errors to generic strings and log the detail server-side.

**`rust/crates/aleph-attest-agent/src/main.rs`** (line 70): SECURITY: The reqwest client used for upstream proxying is built with `Client::new()` — no connect/read/response timeouts. A slow or malicious upstream can pin worker threads indefinitely, degrading the attestation agent's availability. Add `.timeout(Duration::from_secs(30))` or similar.

**`rust/crates/aleph-attest-agent/src/proxy.rs`** (line 142): SECURITY: `upstream_resp.bytes().await` buffers the entire upstream response in memory with no size limit. A compromised or misconfigured upstream could return an arbitrarily large response, causing memory exhaustion. Consider setting a response size cap or streaming.

**`rust/crates/aleph-tee/src/sev_snp/verify.rs`** (line 276): BUG: `Asn1Time::days_from_now(0)` returns midnight of today, not the current instant. A cert that expired earlier today would compare incorrectly depending on time-of-day. Use `Asn1Time::from_unix(SystemTime::now()...)` (as the test helper already does) for accurate expiry comparison.

**`rust/crates/supervisor-controller/src/qemu.rs`** (line 56): SECURITY: Config string fields (image_path, kernel_cmdline, hugepage_size, etc.) are interpolated raw into QEMU option strings. QEMU parses commas as field separators, so a value containing a comma injects additional QEMU options. The config is trusted (written by the supervisor daemon), but `hugepage_size` is documented as '1G' or '2M' with no enforcement — validate it as an enum or hex literal. Consider QEMU-style comma escaping (doubling) for all interpolated values.

**`rust/crates/supervisor-controller/src/qmp.rs`** (line 88): SECURITY (latent): `send` builds JSON via `format!` without escaping the command string. Current callers pass only hardcoded `&'static str` literals, so this is not reachable today. Use `serde_json` to serialize `{"execute": command}` to prevent future QMP JSON injection if a non-hardcoded command is ever passed.

**`rust/crates/supervisor-daemon/src/numa.rs`** (line 485): BUG: `NumaAllocator::register` does `self.allocated_vcpus[index] += vcpus` without a capacity check. If two on-disk configs share the same `numa_node` and their combined vCPUs exceed the node's CPU count (possible after host reconfiguration or stale drop-ins), the ledger over-counts and `allocate` saturates-sub to 0 forever, silently breaking placement. Add a capacity check or clamp with a warning.

**`rust/crates/supervisor-daemon/src/hugepages.rs`** (line 48): BUG: `compute_2m_budget` does not enforce the headroom floor when `per_node_cap` is the binding constraint (explicit `global_limit_mb` smaller than `total_ram - num_nodes * headroom`). The module doc says 'every node keeps a headroom floor of regular RAM so the host is never starved,' but the math violates this guarantee for small per-node caps. Fix the formula to apply headroom after capping.

**`rust/crates/aleph-attest-agent/src/secrets.rs`** (line 23): BUG: `InjectSecretRequest` uses `HashMap<String, String>` with `#[serde(flatten)]`, so duplicate JSON keys silently collapse to the last value with no warning. The secret-injection path is the most security-sensitive operation — consider a `BTreeMap` or explicit duplicate-detection to surface conflicts.

**`rust/crates/aleph-attest-cli/src/verify.rs`** (line 209): BUG: `tokio::task::block_in_place` panics if called from a `current_thread` runtime. The CLI's `#[tokio::main]` uses the multi-thread flavor so this is fine in production, but `fresh_attestation` and `inject_secret` are `pub` and could be called from a library context. Add a doc comment or runtime-flavor assert.

**`rust/crates/supervisor-daemon/src/config.rs`** (line 251): NIT: `HOST_MEMORY_RESERVED_MIB` with a negative env value silently becomes 0 via `u64::try_from(value).unwrap_or(0)`, disabling the memory backstop. Reject negative values explicitly.

**`rust/crates/aleph-tee/src/sev_snp/certs.rs`** (line 282): NIT: `der_object_length` short-form path returns `2 + content_len` without checking it fits within `data.len()`. The caller (`split_der_certs`) does validate with `if first_len > data.len()`, so this is not exploitable, but adding the bounds check inside `der_object_length` itself would be more robust defense-in-depth.

**`rust/crates/aleph-tee/src/sev_snp/verify.rs`** (line 264): NIT: The pinned-ARK comparison uses `Vec<u8>` equality (not constant-time). These are public keys, so timing leakage is low-impact, but using `subtle::ConstantTimeEq` would be more principled for a security-critical comparison.

**`rust/crates/supervisor-controller/src/qemu.rs`** (line 764): BUG: `send_qmp` does `spawn_blocking` then `timeout(...)` — on timeout the `JoinHandle` is dropped without cancel, leaving the blocking task running for up to `COMMAND_DEADLINE` (5s). On a multi-VM host with many wedged monitors, this could exhaust the tokio blocking pool. Consider `.abort()`-ing the handle on timeout.

**`scripts/generate_controller_argv_fixtures.py`** (line 33): NIT: The docstring contains a hardcoded developer-local path `/home/olivier/git/aleph/aleph-vm/`. Replace with a relative or generic path to avoid confusion for other contributors.

foxpatch-aleph (commented):
Well-structured Phase 3 integration branch with strong security design in the attestation/crypto layer: domain-separated report_data, pinned AMD ARK roots, blob-derived trust values, and a TLS handshake gate that prevents secret leakage to unverified peers. The secret injection handler is hardened against TOCTOU, symlinks, and path traversal. Test coverage is thorough with mutation-killing and attack-scenario tests. A few non-blocking observations around proxy SSRF defense-in-depth, nonce length validation, error message information disclosure, and KDS cache TTL.

**`rust/crates/aleph-attest-agent/src/proxy.rs`** (line 96): The upstream URL is built by string-concatenating `state.upstream` with `req.uri().path()`. While the upstream is a fixed startup config (not user-controlled), consider using `url::Url::join` or validating the path to ensure it's relative, as defense-in-depth against absolute-URI request smuggling.

**`rust/crates/aleph-attest-agent/src/proxy.rs`** (line 66): The hex nonce is decoded without a length cap. `fresh_report_data` hashes it regardless so there's no security impact, but an arbitrarily large nonce body could be a minor DoS vector. Consider capping at e.g. 128 hex chars (64 bytes).

**`rust/crates/aleph-attest-agent/src/proxy.rs`** (line 154): The error response `format!("upstream unreachable: {e:#}")` could leak internal network topology to a caller that passes the attestation gate. Consider returning a generic message to the client and logging the detailed error server-side.

**`rust/crates/aleph-tee/src/sev_snp/certs.rs`** (line 107): KDS certificate cache entries never expire. A stale VCEK (after a TCB update) would fail-closed (safe), but could cause a prolonged outage. Consider adding a TTL or a cache-invalidation mechanism.

**`rust/crates/aleph-attest-agent/src/secrets.rs`** (line 75): The `unsafe { libc::geteuid() }` call is trivially safe (no args, returns uid_t), but adding a short `// SAFETY: geteuid is a read-only syscall with no preconditions` comment would satisfy the convention.

foxpatch-aleph (commented):
The D1 increment correctly makes the SNP supervisor architecture-agnostic by reading cbitpos/reduced-phys-bits from host CPUID instead of hardcoding 51/1. The key insight — that these are memory-encryption parameters, not measurement inputs — is sound, and -cpu EPYC-v4 stays fixed to preserve measurement determinism. All call sites are updated consistently, and the parity test now crosses two host pairs. The only gap is the untested `with_cbit_params` builder on `SevSnpBackend`, which has no test coverage and no callers; it's dead code today but harmless. The flake.nix attest-cli addition mirrors the attest-agent build exactly and targets the correct package.

**`rust/crates/aleph-tee/src/sev_snp/backend.rs`** (line 57): The `with_cbit_params` builder is never called by any production code or test. `SevSnpBackend::qemu_args` always uses the default 51/1 from `new()`. While the controller path doesn't use `SevSnpBackend` (it has its own `snp_tee_fragment`), this method should either have a test verifying it overrides the defaults (like the generator's parameterized test) or be removed if no consumer needs it.

PR #1042, opened 2026-07-08
[DO NOT MERGE] Phase 3 testnets candidate deb build (D2 SNP DHCP)

odesenfans:
## DO NOT MERGE - candidate deb build for Phase 3 testnets (D2: SNP guest DHCP)

Triggers `build-deb-package-and-integration-tests.yml` on `od/phase3-d2-snp-dhcp`
so the aleph-testnets upgrade-checks harness can fetch the D2 candidate deb. D2
adds per-tap DHCP serving for SEV-SNP measured VMs (the measured guest DHCPs; the
cmdline omits ip= for a host-independent measurement) plus the guest-side udhcpc
lease-application script. Real review is the stacked PRs; this PR must not merge.


foxpatch-aleph (commented):
High-quality PR with strong security architecture: domain-separated attestation report_data, ARK-pinned certificate chain verification, secrets gated behind the TLS handshake, and careful NUMA ledger management. The roothash cmdline-injection guard, secret store hardening (O_EXCL|O_NOFOLLOW, 0600), and reverse-proxy hop-by-hop header stripping are all done correctly. I found one real bug: DHCP-obtained DNS resolvers are lost in the chroot because /etc is not bind-mounted and the resolv.conf write in prepare_chroot is gated on $gateway which is only set in the static-IP branch. The cert validity check uses day-granularity comparison (Asn1Time::days_from_now(0)) which is a minor edge but acceptable for multi-year ARK/VCEK certs. Overall the code is well-tested and the design decisions are well-documented.

**`nix/init.sh`** (line 96): Bug: DNS resolvers obtained via DHCP are lost in the chroot. The `$gateway` variable is only set in the static-IP branch (line 38); in the DHCP branch it is empty, so this `if [ -n "$gateway" ]` block is skipped and `/mnt/root/etc/resolv.conf` is never written. Meanwhile udhcpc.script writes DNS to the initramfs `/etc/resolv.conf`, but `/etc` is not bind-mounted into the chroot (only /proc, /sys, /dev, /tmp/secrets are). The chrooted workload and /sbin/init will have no DNS resolution. Fix: either bind-mount /etc/resolv.conf from the initramfs into the chroot, or copy the initramfs resolv.conf into the chroot after DHCP completes.

**`nix/udhcpc.script`** (line 46): Minor: `for ns in $dns` uses unquoted `$dns` for word-splitting. This is intentional (DNS servers are space-separated by udhcpc) and the values come from the trusted host dnsmasq, so it is not an injection vector in practice. A `set -f` at the top of the script would make this safe against glob expansion if a DNS value ever contained glob characters, but that is purely defensive.

**`rust/crates/aleph-tee/src/sev_snp/verify.rs`** (line 276): Minor: `Asn1Time::days_from_now(0)` creates a time at midnight UTC of the current day, so `check_cert_validity` accepts a certificate that expires later today (the `not_after < now` comparison passes since `now` is midnight and `not_after` is later in the day). This is day-granularity comparison consistent with openssl's Asn1Time design and is acceptable for ARK/VCEK certificates with multi-year validity, but worth noting for completeness.

**`rust/crates/aleph-attest-agent/src/main.rs`** (line 76): Observation: the attest-agent binds on `0.0.0.0:8443` with no configurable bind address. The guest firewall in init.sh (nftables drop-all except tcp/8443 and loopback) is the real access control, so this is fine in the measured-boot context, but if the firewall fails to install (init.sh line 127 powers off, so this is covered) the agent would be exposed. The fail-closed poweroff on firewall failure is the right design.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 659): Good: the DHCP teardown in stop_vm_execution correctly checks `entry.config.snp().is_some()` before calling `state.dhcp.stop()`, so plain and SEV VMs never trigger a spurious dnsmasq teardown. The idempotent stop (treating 'not loaded' as success) mirrors the tap-delete tolerance correctly.

foxpatch-aleph (commented):
Well-structured Phase 3 SEV-SNP backport with solid attestation security (AMD ARK pinning, domain-separated report data, O_EXCL|O_NOFOLLOW secret writes, TLS-gated secret injection). The DHCP per-tap server is correctly integrated with idempotent teardown on all lifecycle paths. One functional bug: the SNP guest's chroot does not receive DHCP-provided DNS because prepare_chroot() references $gateway which is only set in the static-IP branch, not the DHCP branch (the SNP path). The security posture is strong overall.

**`nix/init.sh`** (line 96): When the DHCP path is taken (the SNP measured-boot case), `$gateway` is set inside the udhcpc.script subprocess, not in init's process scope. So `prepare_chroot()` writes nothing to `/mnt/root/etc/resolv.conf`, and the DHCP-provided DNS servers (written to `/etc/resolv.conf` in the initramfs by udhcpc.script) are not available in the chroot. Consider copying `/etc/resolv.conf` from the initramfs into the chroot after DHCP completes, e.g. `cp /etc/resolv.conf /mnt/root/etc/resolv.conf` in `prepare_chroot()` when `$gateway` is empty.

**`nix/udhcpc.script`** (line 46): The unquoted `$dns` in `for ns in $dns` relies on word-splitting. This is acceptable here since the values come from the trusted host dnsmasq (DHCP option 6) and are IP addresses, but consider quoting and using a different splitting mechanism for robustness if the DHCP server is ever untrusted.

PR #1043, opened 2026-07-08
feat(rust): serve DHCP to SEV-SNP measured VMs (D2)

odesenfans:
## D2: serve DHCP to SEV-SNP measured VMs

Stacked on D1 (`od/phase3-d1-snp-arch-agnostic`). Closes the last gap for the SEV-SNP measured-boot path: the guest gets its allocated IPv4.

### Why
aleph-vm assigns VM IPs statically via cloud-init; there is no DHCP server (ndppd is an NDP proxy, not a router-advertiser). But the measured SNP image deliberately omits `ip=` from the kernel cmdline (so the launch measurement stays host-independent) and its init.sh falls back to `udhcpc`. With no DHCP server on the tap, the SNP guest never gets an IP and is unreachable for attestation. This ports aleph-cvm's dnsmasq approach for the SNP path only.

### What
- **Daemon** (`supervisor-daemon/src/dhcp.rs` + `lifecycle.rs`): on SNP CreateVm, start a per-tap `dnsmasq` transient systemd unit (`aleph-vm-dhcp-<hash>.service`) with a single-address `--dhcp-range` = the allocated guest IP (network+2), `--dhcp-option=3` gateway (network+1), `--dhcp-option=6` nameservers, `--user=root`, `Type=exec`. Torn down on StopVm/DeleteVm/failed-create/failed-adoption-delete; lease file removed. Gated SNP-only (plain + SEV/SEV-ES keep cloud-init static config, byte-identical to pre-D2). `DhcpBackend` trait seam with a fake for tests. `dnsmasq-base` added to the deb.
- **Guest image** (`nix/udhcpc.script` + `init.sh`): busybox udhcpc only leases; a lease-application script (baked into the initrd, `-s /bin/udhcpc.script`) applies the IP/route/DNS. This changes the launch measurement (expected; each build is deterministic and the client pins whatever the build emits).

The measurement stays host-independent: no per-VM `ip=` in the cmdline.

### Validated end-to-end on real AMD SEV-SNP hardware
Via aleph-testnets: CreateVm a persistent SNP VM over gRPC -> the guest DHCPs its IP -> attest-agent on :8443 -> `aleph-attest-cli` reports **Attestation valid: true**, the pinned measurement matches, the AMD Genoa KDS chain verifies, and both the key-bound and fresh-nonce (Layer-3) attestations pass.

### Review (3-lens: correctness/donor-fidelity, Rust-robustness, test-quality)
Fixed in `bd3d4e9d`: dnsmasq exec-failure was masked by `Type=simple` (now `Type=exec`, so a broken dnsmasq fails CreateVm cleanly); the failed-adoption delete path leaked the DHCP unit (now stops it); lease-file leak (now removed on stop); `--user=root` to drop the missing-user failure mode; malformed-CIDR now fails closed instead of a silent /32; a create/teardown SNP-predicate guard; and test gaps closed (SEV-ES/plain skip DHCP; stderr classification).

### Deferred (see `docs/plans/rust-port-divergences.md`)
- Workload chroot resolv.conf seeding on the DHCP path (needs a `nix/init.sh` change + image rebuild; attest-agent runs in the initramfs so attestation is unaffected).
- shellcheck `nix/udhcpc.script` (SC2154 for udhcpc-injected env vars) + wiring the `*.script` CI pattern.


foxpatch-aleph (commented):
Well-architected PR with thorough documentation, comprehensive teardown coverage across all lifecycle paths, and good test isolation via the DhcpBackend trait seam. The Type=exec hardening, fail-closed CIDR validation, and lease-file cleanup are all correctly implemented. Two minor observations: (1) the udhcpc.script renew case will print a harmless 'File exists' error from `ip addr add` since deconfig is not called before renew events, and (2) there is no lifecycle-level test for dhcp.start() failure during create (the FakeDhcpBackend.fail_start injection exists but is only exercised at the unit level in dhcp.rs). Neither is blocking.

**`nix/udhcpc.script`** (line 39): On a `renew` event, busybox udhcpc does NOT call `deconfig` first, so `ip addr add` will fail with "File exists" if the address is already configured from the initial `bound`. The `ip route add` on line 42 has `2>/dev/null` suppression but `ip addr add` does not. Consider adding `2>/dev/null || true` or flushing before adding (though since the IP is fixed via the single-address range, the existing config is still correct — this is just cosmetic stderr noise).

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 2690): Test gap: there is no lifecycle-level test for `dhcp.start()` failing during create. The `FakeDhcpBackend.fail_start` method exists and is tested in dhcp.rs unit tests, but no lifecycle test exercises the path where dnsmasq fails to start (e.g., missing binary) — which should fail the create and trigger the cleanup path at line 2769. Consider adding a test that injects `fail_start` and asserts the create errors out and the DHCP server is torn down.

foxpatch-aleph (approved):
Well-structured PR that adds per-tap dnsmasq DHCP for SEV-SNP measured VMs behind a clean DhcpBackend trait seam. The Type=exec fix, SNP-only gating, idempotent teardown across all lifecycle paths (stop, delete, failed-boot, discard), lease-file cleanup, and CIDR fail-closed validation are all correct. Command arguments are passed as Vec<String> (no shell injection risk). Test coverage is thorough: unit tests for dnsmasq argv, netmask derivation, CIDR rejection, stderr classification, lease-file removal, and lifecycle tests for start-on-SNP-create, no-start-on-plain/SEV-ES, teardown-on-stop/delete/discard/failed-boot. The udhcpc.script correctly converts dotted netmask to prefix with pure shell arithmetic. The deferred follow-ups (chroot resolv.conf, shellcheck SC2154) are properly documented. Minor nits: netmask_for_prefix silently accepts prefix > 32, and ip addr add in udhcpc.script renew could be noisy without 2>/dev/null, but neither affects correctness.

**`rust/crates/supervisor-daemon/src/dhcp.rs`** (line 101): The prefix is parsed as u8, so values like 33-255 pass parsing. netmask_for_prefix(33+) clamps to 255.255.255.255 due to `prefix >= 32`. Consider adding a bounds check here (e.g. `if prefix > 32 { return Err(...) }`) to reject invalid CIDRs explicitly, matching the fail-closed philosophy.

**`nix/udhcpc.script`** (line 39): On a `renew` where the address is already assigned, `ip addr add` will print "RTNETLINK answers: File exists" to stderr. The route add on line 42 suppresses this with `2>/dev/null`, but the addr add does not. Consider adding `2>/dev/null` here for symmetry and to avoid noisy (harmless) output on renewals. The route case already handles this pattern.

foxpatch-aleph (commented):
Well-structured PR implementing per-tap DHCP for SEV-SNP measured VMs. The DhcpConfig/SystemdRunDhcp design mirrors the existing TapBackend seam pattern, the teardown paths (stop, delete, discard, failed-create) are all covered, and the test suite is thorough (dnsmasq argv construction, CIDR validation, fake backend, lease cleanup, all lifecycle paths). The Type=exec fix is load-bearing and correctly documented. The guest-side udhcpc.script is a minimal busybox sh script that correctly converts a dotted netmask to a prefix. Minor observations below are non-blocking robustness suggestions for the guest script.

**`nix/udhcpc.script`** (line 39): On `bound | renew`, `ip addr add` is not error-suppressed (unlike the route add on line 42 which uses `2>/dev/null`). If the address already exists (e.g. a `renew` without a preceding `deconfig`, or a double-`bound`), `ip addr add` prints an error to stderr but the script continues. This is probably fine since busybox udhcpc always calls `deconfig` first, but consider adding `2>/dev/null` or an `|| true` for robustness, matching the route-add pattern on line 42.

**`nix/udhcpc.script`** (line 39): The `bound` case does not flush existing addresses before `ip addr add`. While busybox udhcpc calls `deconfig` (which flushes) before `bound`, a defensive `ip addr flush dev "$interface"` before the `addr add` in the `bound | renew` branch would make the script self-contained and resilient to any udhcpc flow changes. This matches the pattern in the aleph-cvm donor's default.script.

**`rust/crates/supervisor-daemon/src/dhcp.rs`** (line 192): `netmask_for_prefix` accepts any `u8` (0-255) and clamps values > 32 to a /32 mask. The caller `DhcpConfig::for_snp` parses the prefix as `u8` but does not validate the range (0-32). A malformed CIDR like `172.16.4.0/99` would silently produce a /32 mask rather than failing. Consider validating `prefix <= 32` in `for_snp` and returning an error for out-of-range values, consistent with the fail-closed principle applied to the missing-prefix case.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 2638): The `debug_assert_eq!` guarding that create's `snp` predicate matches `config.snp()` is only checked in debug builds. In a release build, if the predicates ever diverge, a started DHCP server would leak on teardown (teardown keys on `config.snp().is_some()` while create keys on the request `snp` bool). This is a documentation guard and consistent with the codebase, but consider whether a release assert is warranted here given the leak consequence.

PR #1044, opened 2026-07-08
fix(nix): chroot DNS + secrets mount on the read-only measured rootfs (D3)

odesenfans:
## D3: give the chrooted workload DNS and a working secrets mount

Stacked on D2 (`od/phase3-d2-snp-dhcp`). Resolves the first D2-deferred follow-up (see `docs/plans/rust-port-divergences.md`).

### Why
The verity rootfs is mounted read-only, so `prepare_chroot`'s direct writes could never work on measured boots:
- the DNS `echo > /mnt/root/etc/resolv.conf` failed silently (and only ran on the static `ip=` path anyway, which the SNP image never takes), leaving the chrooted workload with no resolver;
- `mkdir -p /mnt/root/tmp/secrets` could not create the secrets bind-mount target, so injected secrets were invisible to the workload (same read-only failure, found while fixing the first).

Attestation was unaffected (the attest-agent runs in the initramfs), which is why the e2e validation passed regardless.

### What
- `nix/rootfs.nix`: ship `/etc/resolv.conf` (empty) and `/tmp/secrets` (0700, `/tmp` 1777) placeholders; a file bind-mount needs an existing target and `mkdir -p` then no-ops.
- `nix/init.sh` `prepare_chroot`: bind-mount the initramfs `/etc/resolv.conf` (DHCP option-6 nameservers written by `udhcpc.script`; gateway-seeded on the static `ip=` path) over the rootfs placeholder, with a best-effort copy fallback for legacy writable rootfs.
- `nix/udhcpc.script`: create `/etc` before writing `resolv.conf` (the minimal initramfs ships none, so the write previously failed silently).

Changes the launch measurement (expected; the client pins whatever the build emits). Verified: the built rootfs carries the placeholders (debugfs) and the initrd carries the fixed script.

### DIFF
--- a/nix/init.sh
+++ b/nix/init.sh
@@ -92,9 +92,27 @@ prepare_chroot() {
     /bin/busybox mkdir -m 0700 -p /tmp/secrets
     /bin/busybox mkdir -m 0700 -p /mnt/root/tmp/secrets
     /bin/busybox mount --bind /tmp/secrets /mnt/root/tmp/secrets
-    # DNS: use gateway as nameserver (common for VM bridges).
-    if [ -n "$gateway" ]; then
-        echo "nameserver ${gateway}" > /mnt/root/etc/resolv.conf
+    # DNS for the chrooted workload. The rootfs is mounted read-only under
+    # dm-verity, so writing /mnt/root/etc/resolv.conf directly cannot work
+    # there (the old `echo >` only ever worked on legacy writable mounts).
+    # Instead, expose the initramfs /etc/resolv.conf (written by udhcpc.script
+    # from the DHCP option-6 nameservers) via a file bind-mount over the
+    # rootfs placeholder. On the static ip= path there is no DHCP lease, so
+    # seed the initramfs copy from the gateway first (common for VM bridges).
+    if [ ! -f /etc/resolv.conf ] && [ -n "$gateway" ]; then
+        /bin/busybox mkdir -p /etc
+        echo "nameserver ${gateway}" > /etc/resolv.conf
+    fi
+    if [ -f /etc/resolv.conf ]; then
+        if [ -f /mnt/root/etc/resolv.conf ]; then
+            /bin/busybox mount --bind /etc/resolv.conf /mnt/root/etc/resolv.conf \
+                || echo "init: WARNING: resolv.conf bind-mount failed; workload DNS unavailable"
+        else
+            # Legacy rootfs without the placeholder: best-effort copy (works
+            # only if the rootfs is mounted writable).
+            /bin/busybox cp /etc/resolv.conf /mnt/root/etc/resolv.conf 2>/dev/null \
+                || echo "init: WARNING: rootfs has no /etc/resolv.conf placeholder; workload DNS unavailable"
+        fi
     fi
     echo "init: chroot environment prepared (proc, sys, dev, secrets, DNS)"
 }
--- a/nix/rootfs.nix
+++ b/nix/rootfs.nix
@@ -21,8 +21,23 @@ pkgs.runCommand "rootfs.ext4" {
   SOURCE_DATE_EPOCH = "0";
 } ''
   # Create a minimal ext4 image with /sbin/init entrypoint.
-  mkdir -p rootfs/sbin rootfs/bin rootfs/srv
+  mkdir -p rootfs/sbin rootfs/bin rootfs/srv rootfs/etc
   cp ${staticBusybox}/bin/busybox rootfs/bin/
+
+  # Empty resolv.conf placeholder: the rootfs is mounted read-only under
+  # dm-verity, so init.sh cannot create files in it at boot. It instead
+  # bind-mounts the initramfs /etc/resolv.conf (the DHCP-provided nameservers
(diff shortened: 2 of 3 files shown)


foxpatch-aleph (approved):
This PR correctly fixes two read-only rootfs issues for the SNP measured-boot path: DNS and secrets bind-mount targets. The three-part approach is sound — shipping placeholders in the rootfs image (rootfs.nix), bind-mounting the initramfs resolv.conf over the placeholder (init.sh), and ensuring /etc exists before udhcpc writes to it (udhcpc.script). The logic correctly handles both the static ip= path (seeds resolv.conf from gateway) and the DHCP path (uses udhcpc-written resolv.conf), with appropriate fallbacks for legacy writable rootfs. The empty-placeholder and 0700 secrets permissions are correct. Comments thoroughly explain the constraints and rationale. No security issues — DNS sources are trusted (kernel cmdline or host dnsmasq). No test coverage exists for the initramfs scripts, but this is consistent with the existing approach of e2e validation for the nix boot chain.

**`nix/init.sh`** (line 103): The `mkdir -p /etc` here is redundant — line 15 already creates `/etc` in the initramfs before any networking or chroot setup runs. This is a harmless defensive measure but worth noting for clarity.

**`nix/udhcpc.script`** (line 46): Similarly, `mkdir -p /etc` is redundant when called from init.sh (which creates `/etc` at line 15), but makes the script self-sufficient if ever invoked independently. Reasonable defensive choice.

**`nix/rootfs.nix`** (line 39): The `chmod 1777 rootfs/tmp` is applied after `mkdir -p rootfs/tmp/secrets` creates `rootfs/tmp` with default permissions. This works correctly (chmod overrides the default), but ordering the chmod before the secrets mkdir would be marginally cleaner to ensure the sticky bit is set before any child paths are created.

foxpatch-aleph (approved):
The PR correctly fixes two real bugs caused by the read-only dm-verity rootfs: missing DNS for the chrooted workload and a missing secrets bind-mount target. The approach of shipping placeholders in the rootfs image and using file bind-mounts is the correct solution for read-only filesystems. Both the DHCP path (udhcpc.script writes resolv.conf) and the static ip= path (gateway-seeded fallback) are handled. The legacy writable-rootfs copy fallback is a reasonable defensive measure. Comments are thorough and explain the constraints well. The redundant `mkdir -p /etc` calls in both init.sh:103 and udhcpc.script:46 are harmless (line 15 of init.sh already creates /etc), and the lack of tests is consistent with the existing codebase pattern for these VM initramfs scripts.

**`nix/init.sh`** (line 103): The `mkdir -p /etc` here is redundant — line 15 of init.sh already creates `/etc` at boot. It's harmless (idempotent), but could be removed for clarity. The same applies to the `mkdir -p /etc` added in udhcpc.script:46, though there it's slightly more defensible as a defensive guard for script reuse outside the init.sh context.

**`nix/init.sh`** (line 94): Pre-existing (not introduced by this PR): the secrets bind-mount has no error checking, unlike the new resolv.conf bind-mount which logs a warning on failure. If the rootfs lacks the `/tmp/secrets` placeholder (e.g., an older image), the `mkdir -p` on line 93 would fail silently on the read-only verity mount and the bind-mount would also fail, leaving secrets invisible to the workload with no log. Consider adding a `|| echo` warning here for parity with the DNS bind-mount error handling, but this is non-blocking since the rootfs.nix now ships the placeholder.

foxpatch-aleph (approved):
The PR correctly fixes two latent bugs on the measured-boot (dm-verity) path: the chrooted workload had no DNS (resolv.conf write silently failed on the read-only rootfs) and the secrets bind-mount target didn not exist (mkdir -p failed on the read-only rootfs). The approach of shipping placeholders in the rootfs image and bind-mounting over them is the right solution for read-only verity mounts. The logic handles both the DHCP path (udhcpc.script writes /etc/resolv.conf) and the static ip= path (gateway-seeded DNS), with a sensible fallback for legacy writable rootfs. The udhcpc.script mkdir -p /etc fix is correct and defensive. Comments are thorough and explain the why behind each change. No security issues: secrets dir maintains 0700, resolv.conf bind-mount exposes no sensitive data, and the firewall is unaffected.

**`nix/init.sh`** (line 103): Minor nit: `mkdir -p /etc` is redundant here — init.sh line 15 already does `mkdir -p /etc /tmp` early in the boot. It is harmless (defensive), but if you want to keep it, a one-line comment noting it is belt-and-suspenders would clarify intent. Not blocking.

**`nix/udhcpc.script`** (line 46): Same observation: `mkdir -p /etc` is redundant with init.sh line 15 since udhcpc is only ever invoked from init.sh after /etc is created. Harmless and defensive, but worth a note that this guards against the script being invoked in a different context.

**`nix/rootfs.nix`** (line 39): Consider whether the workload needs write access to /tmp beyond /tmp/secrets. The rootfs /tmp is now part of the read-only verity mount (chmod 1777 is preserved in the image but the mount is -o ro). A real workload that writes to /tmp would fail. This is a pre-existing limitation of the read-only rootfs, not introduced by this PR, but worth noting for future workloads that may need a writable /tmp (e.g. via a tmpfs bind-mount, similar to how /tmp/secrets is handled).

PR #1046, opened 2026-07-08
refactor(rust): read the daemon's mem_size_mb as memsizes::MiB

odesenfans:
Follow-up to the #1030 review. The daemon's `controller_config` reader carried the same lax `_coerce_mib` emulation as the A1 controller: a custom deserializer that accepted a float or string `mem_size_mb`.

No producer ever writes those forms:
- pydantic `MiB` is an `int` subclass, so `.json()` emits a bare integer;
- the Rust daemon writer holds a `u64`.

The daemon's variant was in fact the more dangerous of the two. Its float path (`number.as_f64().map(|float| float as u64)`) had no negative/non-finite guard, so a hand-corrupted config with `mem_size_mb: -2048` failed `as_u64()`, fell through to `-2048.0 as u64`, and **saturated to 0**, launching QEMU with `-m 0`.

## Change

Read the field as [`memsizes::MiB`](https://crates.io/crates/memsizes) (the aleph-rs newtype the CVM scheduler agent already uses), which serde deserializes only from a bare integer. A float, negative or string is now a parse error; the world view then skips the VM, which is the intended fail-closed behaviour. The ~15-line custom deserializer and its now-moot lax-form test are gone.

- Read struct `QemuVmConfig.mem_size_mb`: `u64` + custom deserializer → `MiB`.
- Proto/accounting edges take `.count()` at the boundary (`memory_mib` in the world spec, the `check_memory_backstop` sum).
- The write struct `WrittenQemuVmConfiguration.mem_size_mb` stays `u64`: it never coerced and its value comes from a `u64` in code, so there is no footgun to close there.
- Tests: the lax-form test becomes a strict-rejection test (float/negative/string); the standalone float-string test is subsumed and removed.

Net −11 lines; all 238 daemon tests, fmt and clippy green.

## Note

This branches off `dev` (the daemon lives there via #1026-#1028, not on the Phase 3 stack) and, like #1030, adds `memsizes` to the workspace `Cargo.toml`. Whichever of the two lands second will need a one-line rebase to drop the duplicate dependency line.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -2875,6 +2875,7 @@ dependencies = [
  "clap",
  "hyper-util",
  "libc",
+ "memsizes",
  "prost",
  "rmpv",
  "rusqlite",
--- a/rust/crates/supervisor-daemon/Cargo.toml
+++ b/rust/crates/supervisor-daemon/Cargo.toml
@@ -16,6 +16,7 @@ supervisor-proto.workspace = true
 anyhow.workspace = true
 clap.workspace = true
 libc.workspace = true
+memsizes.workspace = true
 prost.workspace = true
 rmpv.workspace = true
 rusqlite.workspace = true
--- a/rust/crates/supervisor-daemon/src/controller_config.rs
+++ b/rust/crates/supervisor-daemon/src/controller_config.rs
@@ -29,7 +29,8 @@
 
 use std::path::PathBuf;
 
-use serde::{Deserialize, Deserializer};
+use memsizes::MiB;
+use serde::Deserialize;
 
 /// A parse failure for one controller config file. The world view logs it
 /// and skips the VM; it never aborts the daemon (the Python crash-loop
@@ -152,8 +153,7 @@ pub struct QemuVmConfig {
     #[serde(default)]
     pub qga_socket_path: Option<String>,
     pub vcpu_count: u32,
-    #[serde(deserialize_with = "deserialize_mem_size_mib")]
-    pub mem_size_mb: u64,
+    pub mem_size_mb: MiB,
     #[serde(default)]
     pub interface_name: Option<String>,
     pub host_volumes: Vec<QemuHostVolume>,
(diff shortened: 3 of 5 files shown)


foxpatch-aleph (approved):
This PR correctly replaces the custom `deserialize_mem_size_mib` deserializer with the `memsizes::MiB` newtype, closing a real security footgun where a negative `mem_size_mb` like `-2048` would saturate to `0` via `float as u64`, potentially launching QEMU with `-m 0`. The `MiB` type derives serde `Deserialize` for its inner `u64`, so only bare non-negative JSON integers are accepted; floats, negatives, and strings are all rejected as parse errors, causing the world view to skip the VM (fail-closed). All boundary conversions use `.count()` correctly (memory backstop sum, proto `memory_mib` field), and the `for_program` constructor wraps at the proto edge via `MiB::from`. The write struct stays `u64` as justified — it never coerced and its value comes from code. Tests are updated appropriately: the lax-form acceptance test becomes a strict-rejection test covering float, negative, and string forms. No missed call sites or logic errors.

foxpatch-aleph (approved):
This PR correctly replaces a dangerous custom deserializer (whose float-to-u64 coercion could saturate negative values to 0, launching QEMU with -m 0) with the memsizes::MiB newtype, which deserializes only from a bare integer. All three consumption sites of the read struct's mem_size_mb are updated with .count() at the proto/accounting boundaries, the write struct correctly stays u64, the for_program constructor wraps internally with MiB::from(), and the test suite is updated from lax-acceptance to strict-rejection covering float, negative, and string inputs. The change is net -11 lines, well-documented, and closes a real security footgun.

foxpatch-aleph (approved):
Clean, well-motivated refactor that replaces a dangerous custom deserializer with the `memsizes::MiB` newtype. The old float path could saturate a negative `mem_size_mb` to 0 (launching QEMU with `-m 0`); the new type deserializes only from a bare integer, making float/negative/string a hard parse error that fail-closes by skipping the VM. All boundary accesses correctly use `.count()` to extract the raw `u64`, the write struct stays `u64` (no footgun there), and the lax-form test is correctly replaced by a strict-rejection test. No correctness, security, or coverage issues found.

foxpatch-aleph (approved):
Clean, well-motivated refactor that replaces a lax custom deserializer (which could saturate negative floats to u64=0, launching QEMU with -m 0) with the strict memsizes::MiB newtype. All four consumption sites correctly call .count() at the boundary, the write struct intentionally keeps u64 (it never coerced), and the test suite is updated to assert strict rejection of float/negative/string forms. The change mirrors the identical pattern already established in supervisor-controller (PR #1030), and the net line reduction is a bonus.

PR #1047, opened 2026-07-08
refactor(rust): controller main() as a fallible run() instead of an ExitCode ladder

odesenfans:
Addresses the #1030 review note that the `ExitCode` pattern matching in the controller is not very Rust-like.

## Why this targets the stack tip (D4), not #1030

The ladder is not unique to #1030. `main.rs` is rewritten by A2 (#1031, adds the `RunTarget` enum) and B1 (adds `RunTarget::Snp`), and the same `match { Ok(v) => v, Err(e) => { error!(); return ExitCode::FAILURE } }` shape is carried through each. `main.rs` only reaches its final three-target form at the tip.

So refactoring #1030's copy would just be overwritten when A2/B1 replay their `main.rs` edits, and basing the fix mid-stack would either conflict with those rewrites or force a rebase of the whole downstream chain (the thundering herd we are avoiding). Basing it on the tip is the one conflict-free single-shot: it operates on the final `main.rs`, touches nothing downstream (it *is* the tip), and rides down to `dev` with the chain. It keeps #1030's own diff clean, as requested.

## The change

`main()` was a ladder of `match { Ok(v) => v, Err(e) => { error!(); return ExitCode::FAILURE } }` repeated at every fallible step (config lookup, load, network validation, run-target dispatch, tap wait, runtime build, QEMU). That is the un-Rust-like shape: catching each error only to collapse it to a binary success/failure.

- Extract the fallible body into `run(&Cli) -> Result<(), ControllerError>` that uses `?`.
- `main()` initializes logging, then logs the error **once** and maps it to the exit code in a single place.
- `ControllerError` is a `thiserror` enum whose variants map one-to-one to the Python `__main__.main` `exit(1)` sites. Operator-facing messages are unchanged: the config-path / failing-operation context is carried on the variants, and `select_run_target` / `qemu::run` already render full messages (wrapped verbatim via the two `String` variants).

Behaviour is byte-for-byte identical, including the **deliberate discard of QEMU's exit code** (Python returns after `asyncio.run` regardless of the return code, so a VM that boots and later exits is still a successful controller run). Adds two tests pinning the early-return mappings (missing file → `ConfigNotFound`, unparseable file → `ConfigLoad`).

Net +96/−42 in one file; controller tests, fmt and clippy green. No effect on the SEV-SNP measurement (host-side controller binary, not part of the measured guest image).

### DIFF
--- a/rust/crates/supervisor-controller/src/main.rs
+++ b/rust/crates/supervisor-controller/src/main.rs
@@ -12,7 +12,9 @@ use std::process::ExitCode;
 use std::time::Duration;
 
 use clap::Parser;
-use supervisor_controller::config::{Configuration, HypervisorType, QemuConfig, VmConfiguration};
+use supervisor_controller::config::{
+    ConfigError, Configuration, HypervisorType, QemuConfig, VmConfiguration,
+};
 use supervisor_controller::qemu;
 
 /// The `__main__.main` network pre-check budget: wait up to 120s for the tap.
@@ -36,6 +38,36 @@ struct Cli {
     verbose: u8,
 }
 
+/// A fatal controller error. `main` logs it once and turns it into a
+/// non-zero exit; each variant is one of the `exit(1)` sites (or the
+/// uncaught-exception exit) in the Python `__main__.main`. The two `String`
+/// variants wrap helpers that already render a full, actionable message
+/// ([`select_run_target`] and [`qemu::run`]); the rest attach the context
+/// (the config path, the failing operation) that Python logged inline.
+#[derive(Debug, thiserror::Error)]
+enum ControllerError {
+    #[error("Configuration file {} not found", .0.display())]
+    ConfigNotFound(PathBuf),
+
+    #[error("cannot load {}: {source}", .path.display())]
+    ConfigLoad { path: PathBuf, source: ConfigError },
+
+    #[error("Controller config {} carries no NETWORK_INTERFACE", .0.display())]
+    NoNetworkInterface(PathBuf),
+
+    #[error("{0}")]
+    UnsupportedConfig(String),
+
+    #[error("{0}")]
+    TapUnavailable(String),
+
+    #[error("cannot start the async runtime: {0}")]
+    Runtime(#[source] std::io::Error),
+
+    #[error("{0}")]
+    Qemu(String),
+}
+
 fn main() -> ExitCode {
     let cli = Cli::parse();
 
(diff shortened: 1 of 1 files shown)


foxpatch-aleph (approved):
A clean, idiomatic refactor that extracts the fallible controller body into `run(&Cli) -> Result<(), ControllerError>` using `?`, replacing the repetitive `match { Ok(v) => v, Err(e) => { error!(); return ExitCode::FAILURE } }` ladder. The `ControllerError` enum variants map one-to-one to the Python `__main__.main` `exit(1)` and uncaught-exception sites, with `thiserror` `#[error(...)]` messages that preserve the exact operator-facing text (including `PathBuf::display()` for paths and `#[source]` chaining for `ConfigError`/`io::Error`). The deliberate QEMU exit-code discard is faithfully preserved (`result.map_err(ControllerError::Qemu)?; Ok(())`), matching Python's unguarded `asyncio.run`. The `print_settings`-before-`NETWORK_INTERFACE`-check ordering matches Python. Two new tests pin the two earliest error paths (`ConfigNotFound` for a missing file, `ConfigLoad` for unparseable JSON) with correct isolation. The dispatch comment is correctly updated to mention SNP ordering. `ConfigError` derives `thiserror::Error` (config.rs:24) so `#[source]` is valid; the `Runtime(#[source] std::io::Error)` with `{0}` Display is a valid `thiserror` idiom. `thiserror` and `tempfile` are both workspace deps. No security, correctness, or testing concerns.

foxpatch-aleph (approved):
This is a clean, idiomatic Rust refactoring that replaces a repetitive ExitCode ladder with a fallible run() function using ? and a thiserror error enum. The ControllerError variants map one-to-one to the Python __main__.main exit(1) sites and uncaught-exception paths, preserving operator-facing messages verbatim. The deliberate QEMU exit-code discard (line 149) faithfully mirrors Python's asyncio.run behavior. The two new tests pin the first two error sites with appropriate matches! assertions. No correctness, security, or documentation issues found.

foxpatch-aleph (approved):
This is a clean, idiomatic refactor that replaces the repetitive `match { Ok(v) => v, Err(e) => { error!(); return ExitCode::FAILURE } }` ladder with a `run() -> Result<(), ControllerError>` function using `?` propagation and a single error-logging/exit-code-mapping site in `main()`. The `ControllerError` enum variants map one-to-one to the Python `__main__.main` exit sites, and operator-facing messages are preserved via `thiserror` format strings that carry the same context (config path, failing operation). The deliberate discarding of QEMU's exit code is correctly preserved, matching Python's behavior. The two new tests pin the early-return error variants (missing file and unparseable config) that are unit-testable without a real VM environment. `thiserror` and `tempfile` are already dependencies, so no new deps are introduced. No correctness, security, or quality issues found.

PR #1048, opened 2026-07-09
ci(codeql): exclude non-shipped scripts/ tooling from analysis

odesenfans:
Resolves the Phase 3 CodeQL false positives (the pending "resolve CodeQL FPs" task) at the source.

## The problem

The `CodeQL` check fails on every PR that carries a fixture/codegen generator under `scripts/`. It first surfaced on a real incremental PR with #1031 (A2), whose `scripts/generate_controller_argv_fixtures.py` builds confidential argv fixtures. CodeQL taints the hardcoded dummy SEV `session`/`godh`/`dh_cert` **paths** those generators write into fixtures (e.g. `/var/lib/aleph/vm/testhash/vm_session.b64`) as clear-text secrets: `py/clear-text-storage-sensitive-data` and `py/clear-text-logging-sensitive-data`.

These are false positives on fixture data. The paths are test placeholders that do not point at real files (the fixtures exist precisely so the argv-parity tests run without real SEV material on disk), and no key bytes are stored or logged.

## The fix

`scripts/` is developer tooling: codegen, test-fixture generators (`generate_controller_argv_fixtures.py`, `generate_rust_fixtures.py`) and one-off diagnostics. Nothing in `src/aleph/` imports it and it is not packaged. So it does not belong in the security-analysis scope.

- Add `.github/codeql/codeql-config.yml` with `paths-ignore: [scripts]`.
- Wire `config-file: ./.github/codeql/codeql-config.yml` into the `codeql-action/init` step.

This removes the whole class of noise without hiding any shipped code, and it applies to every open PR through its merge commit once this lands on `dev` (both base and head then exclude `scripts/`, so the differential alert check comes back clean).

Standalone off `dev`, independent of the Rust stack (like the daemon `MiB` fix #1046).

### DIFF
--- a/.github/codeql/codeql-config.yml
+++ b/.github/codeql/codeql-config.yml
@@ -1,0 +1,14 @@
+---
+name: "aleph-vm CodeQL config"
+
+# Exclude non-shipped developer tooling from analysis. scripts/ holds codegen
+# and test-fixture generators (e.g. generate_controller_argv_fixtures.py,
+# generate_rust_fixtures.py) plus one-off diagnostics, none of which is imported
+# by the runtime package or packaged. CodeQL taints the hardcoded dummy SEV
+# session/godh *paths* those generators write into fixtures as clear-text
+# secrets (py/clear-text-{storage,logging}-sensitive-data); on fixture data that
+# is a false positive, and it fails the CodeQL check on every PR that carries a
+# generator. Dropping scripts/ from the analysis scope removes the noise without
+# hiding any shipped code.
+paths-ignore:
+  - scripts
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -49,6 +49,9 @@ jobs:
         uses: github/codeql-action/init@v3
         with:
           languages: ${{ matrix.language }}
+          # paths-ignore lives here (excludes non-shipped scripts/ tooling from
+          # the analysis scope; see the config for why).
+          config-file: ./.github/codeql/codeql-config.yml
         # If you wish to specify custom queries, you can do so here or in a config file.
         # By default, queries listed here will override any specified in a config file.
         # Prefix the list here with "+" to use these queries and those in the config file.


foxpatch-aleph (approved):
This PR correctly adds a CodeQL config file with `paths-ignore: [scripts]` and wires it into the `codeql-action/init` step. I verified that `scripts/` is developer tooling only: it is not part of the packaged wheel (pyproject.toml packages only `src/aleph`), no runtime code in `src/aleph/` imports from it, and it holds codegen/fixture generators and shell utilities. The exclusion is narrowly scoped to `scripts` only, leaving all shipped code under `src/aleph/` in the analysis scope. The config uses the correct `paths-ignore` schema key and the `config-file` path is correct. The inline comments in both the config and workflow clearly explain the rationale (CodeQL false positives on dummy SEV paths in fixture data). This is a clean, well-documented fix that removes a class of false positives without hiding any shipped code.

foxpatch-aleph (approved):
This PR correctly excludes the non-shipped scripts/ directory from CodeQL analysis by adding a well-formed config file and wiring it into the existing init step. The claim that scripts/ is developer-only tooling is verified: the wheel packages only src/aleph, no runtime code imports scripts/, and the directory contains only codegen/fixture generators and shell utilities. The paths-ignore syntax is valid, config-file is the correct action parameter, and the explanatory comments clearly document the rationale. This is a clean, minimal fix for a legitimate false-positive problem.

foxpatch-aleph (approved):
Clean, minimal, and well-documented change. The new CodeQL config correctly excludes the non-shipped scripts/ developer tooling directory from analysis, verified to contain no imports from the runtime package and to be excluded from packaging. The paths-ignore schema and config-file workflow wiring are both correct. This resolves the false-positive class at the source without hiding any shipped code.

foxpatch-aleph (approved):
This PR correctly adds a CodeQL configuration file that excludes the non-shipped `scripts/` directory from security analysis, resolving false-positive alerts on fixture-generating developer tooling. I verified that `scripts/` is not included in the wheel package (pyproject.toml packages only `src/aleph`), and that no runtime code under `src/aleph/` imports from `scripts/`. The CodeQL config uses the correct `paths-ignore` key, the YAML is well-formed, and the `config-file` parameter is properly wired into the `codeql-action/init@v3` step. The exclusion is narrowly scoped to developer tooling only, so no shipped code is hidden from analysis. Comments in both the config and workflow files clearly explain the rationale. No issues found.

PR #1049, opened 2026-07-09
feat(storage): multi-disk volume pools for VM storage

odesenfans:
## Summary

CRNs with multiple fast disks can now spread VM volumes across them. Operators declare extra pool directories via `ALEPH_VM_VOLUME_POOLS` (JSON list, optional `path=class` override); `PERSISTENT_VOLUMES_DIR` remains the implicit first pool, so unconfigured nodes are byte-for-byte unchanged.

- **New agent-side module `storage_pools.py`**: media-class detection from sysfs (`nvme`/`ssd`/`hdd`, dm/md stacks resolve to the slowest member), most-free-pool placement, sticky cross-pool lookup, `st_dev`-deduplicated accounting.
- **HDD policy**: rotational disks are refused as extra pools (they're for `CACHE_ROOT`/`BACKUP_DIRECTORY`); pool 0 on HDD warns but stays eligible so existing nodes keep working.
- **Safety**: pools carry an adoption marker plus a registry under `EXECUTION_ROOT` — an unmounted disk, a corrupted registry, or a pool removed from config while it still holds volumes are all hard startup errors instead of silent misplacement/data loss.
- **Wired call sites**: ext4/btrfs volume creation, qcow2 instance-rootfs staging, migration (reaper sweeps all pools, export dedupes by lowest-index pool matching boot-time resolution, import stages via placement).
- **Accounting**: admission (`capacity.py`, incl. a new largest-single-volume vs roomiest-pool check), pool reservations, and `/about/usage/system` all report the deduplicated aggregate; the Rust supervisor-daemon reads the same env var and sums pools for `GetHostInfo`.
- **Firecracker**: writable program volumes pin to pool 0 — the jailer silently copies (not hardlinks) across `st_dev`, which would lose guest writes on an extra pool.
- **Untouched by design**: `proto/supervisor.proto`, `supervisor-controller`, aleph-message — `DiskConfig` already carries opaque absolute paths.

Failure domain vs LVM striping: one dead SSD loses only the VMs whose volumes live on it; everything else keeps running.

Design doc: `docs/plans/2026-07-09-multi-disk-storage-pools-design.md` (local).
Operator guide: `doc/multi-disk-storage.md`.

## Test plan

- 46 new tests in `tests/supervisor/test_storage_pools.py` (detection with fake sysfs, adoption/registry, placement/lookup, accounting dedup, migration sweep, pool0 pinning) plus capacity/migration additions; focused suites 177 passed.
- Full local Python suite: 1075 passed, 4 known environment-only failures (pyroute2-needs-root, journald stub).
- Rust: `cargo test -p supervisor-daemon` 315 passed, clippy `-D warnings` clean.
- ruff/isort/mypy at baseline (no new findings).

### DIFF
--- a/doc/multi-disk-storage.md
+++ b/doc/multi-disk-storage.md
@@ -1,0 +1,44 @@
+# Using multiple disks on a CRN
+
+aleph-vm can spread VM volumes (instance rootfs overlays and persistent
+volumes) across several fast disks, and use slow disks for caches.
+
+## Fast disks (NVMe / SATA SSD): volume pools
+
+Mount each extra disk, then declare the mountpoints in
+`/etc/aleph-vm/supervisor.env`:
+
+    ALEPH_VM_VOLUME_POOLS=["/mnt/nvme1/aleph-volumes", "/mnt/sata1/aleph-volumes=ssd"]
+
+- The value is a JSON list. Each entry is a directory on the extra disk.
+- `PERSISTENT_VOLUMES_DIR` (default `/var/lib/aleph/vm/volumes/persistent`)
+  is always the first pool; existing nodes need no migration.
+- The media class (`nvme`/`ssd`/`hdd`) is detected from the backing block
+  device. Append `=nvme`/`=ssd` to override a wrong detection.
+- New volumes land on the eligible pool with the most free space. A volume
+  never spans pools. The capacity advertised to the network is the sum of
+  all pools.
+- Rotational disks (HDD) are refused as volume pools: VMs need fast storage.
+
+On first use the agent writes a `.aleph-vm-pool` marker into the pool and
+records the adoption under `EXECUTION_ROOT`. If the marker later disappears
+(typically: the disk is not mounted), the agent refuses to start rather than
+silently writing to the filesystem underneath the mountpoint.
+
+If one pooled disk fails, only the VMs whose volumes live on it are lost;
+the other pools keep serving their VMs. This is the main advantage over
+LVM/RAID0 striping, which loses the whole pool with one disk. If you prefer
+operating a single filesystem (LVM, btrfs multi-device, ZFS, mdraid), that
+still works: leave `VOLUME_POOLS` unset and mount the pooled filesystem at
+`PERSISTENT_VOLUMES_DIR`.
+
+## Slow disks (HDD): caches and backups
+
+HDDs must not hold running VMs, but they are fine for re-downloadable
+content and backups. Point the existing settings at them:
+
+    ALEPH_VM_CACHE_ROOT=/mnt/hdd0/aleph-cache
+    ALEPH_VM_BACKUP_DIRECTORY=/mnt/hdd0/aleph-backups
+
+`CACHE_ROOT` holds runtime/rootfs base images, program code archives,
+message JSON and data volumes — all safe on slow media.
--- a/doc/multi-disk-storage.md
+++ b/doc/multi-disk-storage.md
@@ -14,16 +14,40 @@ Mount each extra disk, then declare the mountpoints in
 - `PERSISTENT_VOLUMES_DIR` (default `/var/lib/aleph/vm/volumes/persistent`)
   is always the first pool; existing nodes need no migration.
 - The media class (`nvme`/`ssd`/`hdd`) is detected from the backing block
-  device. Append `=nvme`/`=ssd` to override a wrong detection.
+  device. Append `=nvme`/`=ssd` to override a wrong detection. Stacked or
+  unnamed-device setups (LVM, btrfs multi-device, mdraid, ZFS, network
+  filesystems) may detect as *undetectable*, which is a startup error: give
+  those pools an explicit `=nvme`/`=ssd` override.
 - New volumes land on the eligible pool with the most free space. A volume
   never spans pools. The capacity advertised to the network is the sum of
   all pools.
-- Rotational disks (HDD) are refused as volume pools: VMs need fast storage.
+- Rotational disks (HDD) are refused as *extra* volume pools: VMs need fast
+  storage. The first pool (`PERSISTENT_VOLUMES_DIR`) is the exception: on an
+  HDD it only logs a warning and stays eligible, so existing nodes keep
+  working unchanged.
 
 On first use the agent writes a `.aleph-vm-pool` marker into the pool and
 records the adoption under `EXECUTION_ROOT`. If the marker later disappears
 (typically: the disk is not mounted), the agent refuses to start rather than
 silently writing to the filesystem underneath the mountpoint.
+
+### Firecracker programs
+
+Writable volumes of Firecracker programs always stay on the first pool,
+regardless of free space. The Firecracker jailer hardlinks drive files into
+its chroot, and a hardlink across filesystems silently becomes a copy: guest
+writes would land in the copy and be lost. Read-only program files (runtime,
+code, data squashfs) are unaffected. Size the first pool accordingly if the
+node runs many programs with persistent volumes.
+
+### Removing a pool
+
+Once a pool has been adopted (its `.aleph-vm-pool` marker written), the agent
+refuses to start if the pool is later dropped from `VOLUME_POOLS`: the pool
+holds (or held) VM volumes, and silently forgetting it would orphan them. To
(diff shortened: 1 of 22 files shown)


foxpatch-aleph (approved):
Well-architected multi-disk storage pool feature with thorough safety mechanisms (adoption registry, marker files, HDD policy, pool0 pinning for Firecracker). The core logic is correct: media-class detection walks dm/md stacks to the slowest member, placement picks the most-free eligible pool with lowest-index tie-breaking, accounting deduplicates by st_dev, and the Firecracker jailer hardlink issue is correctly mitigated. The devmapper path correctly does NOT need pool0_only because block devices don't go through the jailer hardlink. Test coverage is comprehensive across all new code paths. No bugs or security issues found.

**`src/aleph/vm/storage_pools.py`** (line 219): Duplicate pool paths are not detected: if an operator lists the same path twice in VOLUME_POOLS, both entries are adopted and added to the pools list with different indices. This is harmless (st_dev dedup handles accounting, and lowest-index tie-breaking picks the first), but a startup error for duplicates would be a friendlier guardrail.

**`src/aleph/vm/storage_pools.py`** (line 330): The pool0 identity check `existing.parent.parent != pool0.path` uses directory path equality rather than st_dev comparison. If pool 0 and an extra pool happen to share the same filesystem (same st_dev), a Firecracker volume found on the extra pool would log a spurious ERROR about lost writes even though the hardlink would actually succeed. This is a minor false positive — the error log is alarming but the volume would work correctly.

**`src/aleph/vm/agent/resources.py`** (line 224): The `total_kB` comes from `pools_disk_usage()` (agent-side) while `available_kB` may come from `host_info.available_disk_bytes` (supervisor-side). If a pool becomes unreachable between the two calls, total could exceed the sum of available + used, producing a slightly inconsistent disk usage report. This is a minor display-only issue on the /about/usage/system endpoint.

**`src/aleph/vm/storage_pools.py`** (line 237): The orphaned-pool check compares `str(pool.path)` strings against the registry. If an operator changes a pool path's representation (e.g., trailing slash, or `//` in the config), the string comparison would treat it as a different pool and raise a false orphan error. Using `Path.resolve()` for both sides of the comparison would be more robust.

**`src/aleph/vm/agent/migration/runner.py`** (line 69): The warning message in `_collect_export_disks` logs `disk.parent` (the namespace directory, e.g. `pool/vm-x`) rather than the pool path. The message says "exporting the copy from %s" which could be slightly misleading — logging the pool path (`disk.parent.parent`) would be clearer about which pool's copy is being exported.

**`doc/multi-disk-storage.md`** (line 21): The doc says "Staked or unnamed-device setups (LVM, btrfs multi-device, mdraid, ZFS, network filesystems) may detect as *undetectable*" — typo: "Staked" should be "Stacked".

foxpatch-aleph (approved):
Well-designed multi-disk volume pool feature with thorough safety mechanisms (adoption markers, registry guards, orphan detection), correct Firecracker pool-0 pinning, faithful sysfs-based media-class detection, and comprehensive test coverage (46 new tests). The Rust daemon correctly mirrors the agent's pooled accounting. No correctness bugs or security issues found. The minor observations (devmapper ignoring pool0_only, compressed-size heuristic for import staging, duplicate iteration in export) are all either pre-existing, non-blocking, or not reachable in practice.

**`src/aleph/vm/storage.py`** (line 448): The devmapper branch (`if volume.parent:`) calls `create_devmapper(volume, namespace)` without forwarding `pool0_only`. The `create_devmapper` → `create_volume_file` → `volume_path_for` chain doesn't pass `pool0_only` either. This isn't reachable for Firecracker programs today (they use read-only runtimes from cache), but the flag silently doesn't apply to the devmapper path. Consider forwarding it for defense-in-depth, or add a comment noting why it's safe to omit.

**`src/aleph/vm/agent/migration/runner.py`** (line 274): `incoming_mib` uses the compressed export size, but the decompressed qcow2 on the destination may be larger. `select_pool` could pick a pool that can't hold the final volume. Not a regression (old code had no space check), and `check_capacity` catches it later — but worth a comment noting this is a lower bound.

**`src/aleph/vm/agent/migration/runner.py`** (line 141): `iter_namespace_dirs` is called twice for the same `vm_hash` — once on line 141 for `volume_dirs` (only used to set `job.volumes_dir`), and again inside `_collect_export_disks` on line 146. Minor redundancy; could pass `volume_dirs` into `_collect_export_disks` to avoid the double scan.

foxpatch-aleph (approved):
A well-designed and thoroughly tested implementation of multi-disk volume pools. The safety mechanisms (adoption markers, registry, unmounted-disk detection, orphaned-pool guard) are thoughtful and correctly implemented. The Firecracker pool0 pinning is properly handled across the downloader and host_volumes paths. The st_dev deduplication for accounting is consistent between the Python and Rust sides. The main observation is that create_devmapper does not thread pool0_only through to create_volume_file/volume_path_for, but this is safe because devmapper is only used for QEMU instances (not Firecracker programs). Test coverage is excellent with 46 new tests covering detection, adoption, placement, lookup, accounting, migration, and pool0 pinning.

**`src/aleph/vm/storage.py`** (line 449): The `create_devmapper` branch silently drops `pool0_only`. This is safe today because devmapper is only used for QEMU instances, but a brief comment noting "devmapper is QEMU-only, so pool0_only does not apply" would prevent a future reader from assuming it was accidentally omitted.

**`src/aleph/vm/agent/migration/runner.py`** (line 142): `job.volumes_dir` is now set to `volume_dirs[0]` but is no longer used for download resolution (the download view uses `job.export_paths`). The field is semi-vestigial; consider whether it should be removed or repurposed to avoid confusion for future maintainers.

**`src/aleph/vm/storage_pools.py`** (line 124): `_parse_pool_entry` uses `str.partition("=")` which splits on the first `=`. A pool path containing `=` (admittedly rare) would be misparsed. Since the override syntax is `path=class`, consider documenting that `=` in paths is not supported, or use `rsplit` if paths are more likely to contain `=` than class names (though neither is likely).

foxpatch-aleph (commented):
Well-designed multi-disk storage pool feature with strong safety mechanisms (adoption markers, registry, unmounted-disk detection), comprehensive test coverage, and clean Python/Rust parity. The core logic in storage_pools.py is correct — placement, lookup, accounting dedup, and pool0 pinning for Firecracker all work as documented. A few non-blocking observations: (1) _collect_export_disks uses glob *.qcow2 which also matches .export.qcow2 files — pre-existing but now scans more pools; (2) job.volumes_dir is set but never read after the migration download view was changed to use job.export_paths; (3) pool0_only is not threaded through create_devmapper → create_volume_file, which is currently safe but latent. None of these are blocking — the PR is solid.

**`src/aleph/vm/agent/migration/runner.py`** (line 63): The glob `*.qcow2` also matches `rootfs.qcow2.export.qcow2` files (since they end in `.qcow2`). The reaper cleans stale exports at startup, but if an export fails and is retried without an agent restart, a stale `.export.qcow2` could be collected as a source disk and re-exported. Consider filtering these out explicitly, e.g. `if qcow2_file.name.endswith(".export.qcow2"): continue` before adding to `seen_names`. This is pre-existing (the original code used the same glob), but the multi-pool scan increases the chance of encountering stale exports.

**`src/aleph/vm/agent/migration/runner.py`** (line 142): `job.volumes_dir` is set here but no longer read anywhere — the migration download view (migration.py:213) was updated to resolve from `job.export_paths` instead. This is a dead assignment; consider removing it (and the `volumes_dir` field on `ExportJob` in jobs.py:45) to avoid confusion.

**`src/aleph/vm/storage.py`** (line 451): `create_devmapper` is called without threading `pool0_only` through to `create_volume_file` → `volume_path_for`. This is currently safe because devmapper volumes (with `volume.parent` set) are only used for QEMU rootfs overlays, not Firecracker program volumes. But if a Firecracker program ever declares a volume with a parent, the btrfs volume file could land on an extra pool, triggering the jailer hardlink-copy data loss. Worth a comment noting this assumption, or threading `pool0_only` through for completeness.

**`tests/supervisor/test_storage_pools.py`** (line 614): Consider adding a test case for `_collect_export_disks` where a stale `rootfs.qcow2.export.qcow2` file exists alongside `rootfs.qcow2` — this would verify the glob filtering issue described above (or confirm the current behavior is intentional).

PR #1053, opened 2026-07-25
fix(systemd): retry on stale proxy after daemon-reexec

aliel:
## Summary

Fixes the ``org.freedesktop.DBus.Error.ServiceUnknown: The name :1.N was not provided by any .service files`` 500 that appears on ``notify_allocation`` after ``systemctl daemon-reexec`` (or a systemd package upgrade) rotates systemd's unique bus name. Observed on aleph-server-5:

```
POST /control/allocation/notify
500 {\"error\": \"org.freedesktop.DBus.Error.ServiceUnknown: The name :1.3 was not provided by any .service files\"}
```

The cached ``SystemDManager`` proxy was bound to systemd's old unique name; every call after the rotation went to a dead peer until the supervisor was manually restarted.

## Fix

Add ``_call_with_reconnect(work)``: on ``ServiceUnknown`` / ``NoReply`` / ``Disconnected``, call ``_connect()`` (fresh bus and manager) and re-run the callable so the retry routes through the new manager. Wrap every D-Bus-touching method — writes (enable, start, stop, restart, disable) and reads (is_service_enabled, get_service_active_state, is_service_active, get_services_active_states) — with lambdas that read ``self._get_manager()`` at call time.

Not using ``follow_name_owner_changes=True``: it internally subscribes the proxy to ``NameOwnerChanged`` signals, which requires a D-Bus main loop the asyncio supervisor does not have (RuntimeError at ``SystemDManager.__init__`` in every test and at supervisor startup). Retry-on-stale-error achieves the same self-healing at the cost of one extra round-trip on the first call after a rotation.

## Test plan
- [x] Unit tests cover happy path, non-stale reraise, all three stale error names triggering retry, work observes manager swap between attempts, retry-that-also-fails does not loop.
- [ ] Deploy to a staging CRN, verify a normal instance start works.
- [ ] Run ``sudo systemctl daemon-reexec`` on the CRN and confirm the next ``aleph instance start …`` succeeds without restarting the supervisor.
- [ ] Verify ``is_service_active`` also self-heals after the same reexec (payment monitor / status endpoints).

### DIFF
--- a/src/aleph/vm/systemd.py
+++ b/src/aleph/vm/systemd.py
@@ -23,6 +23,12 @@ _NO_SUCH_UNIT = "org.freedesktop.systemd1.NoSuchUnit"
 # name" (typically after systemctl daemon-reexec or a systemd package
 # upgrade rotates its unique bus name). Reconnecting and retrying the
 # call resolves the well-known name to the current owner.
+#
+# NoReply is included so we survive systemd being briefly unresponsive
+# during a restart, at the cost of retrying legitimate slow calls once
+# — worst case doubles that call's wall time. Acceptable tradeoff:
+# genuine NoReply is rare on a healthy host and we still fail fast if
+# the retry itself times out.
 _STALE_CONNECTION_ERRORS = frozenset(
     {
         "org.freedesktop.DBus.Error.ServiceUnknown",
@@ -92,26 +98,25 @@ class SystemDManager:
         msg = "Failed to establish D-Bus connection after multiple attempts"
         raise DBusException(msg)
 
-    def _call_with_reconnect(self, fn, *args, **kwargs):
-        """Run a manager call; on stale-connection errors, reconnect and retry once.
+    def _call_with_reconnect(self, work):
+        """Run ``work()``; on stale-connection errors, reconnect and re-run once.
 
-        ``follow_name_owner_changes=True`` on the proxy makes routine
-        systemd restarts transparent, but a call issued while systemd
-        is *mid*-restart can still hit an error before the new owner
-        registers. On the errors in ``_STALE_CONNECTION_ERRORS``, tear
-        down and rebuild the bus, then retry the same D-Bus method
-        against the freshly-built ``_manager``.
+        ``work`` is a zero-argument callable that performs the D-Bus
+        operation from scratch (typically a lambda that calls
+        ``self._get_manager().SomeMethod(...)``). Passing a callable
+        rather than a pre-resolved method matters: a pre-resolved
+        ``_ProxyMethod`` is bound to the ``_connection`` we just closed
+        via ``self._bus.close()``, so retrying it would go through the
+        dead connection. Re-invoking ``work()`` after ``_connect()``
+        reads the new ``self._manager`` and routes through the fresh
+        proxy — no need to peek at ``_method_name`` or rebind.
 
-        The retry MUST be issued through the new manager: ``fn`` here is
-        a ``dbus.proxies._ProxyMethod`` bound to a specific ``_connection``
-        that ``_connect()`` just closed via ``self._bus.close()``. Calling
-        the original ``fn`` again would go through the dead connection.
-        We rebind by looking up the same method name on the new manager
-        via ``_method_name`` (the documented ``_ProxyMethod`` attribute)
-        and ``Interface.get_dbus_method``.
+        Also handles multi-call chains (e.g. ``manager.GetUnit`` then
+        ``bus.get_object`` then ``properties.Get``): the whole sequence
+        is redone against the fresh bus and manager.
         """
         try:
-            return fn(*args, **kwargs)
+            return work()
         except DBusException as error:
             if error.get_dbus_name() not in _STALE_CONNECTION_ERRORS:
                 raise
@@ -119,14 +124,8 @@ class SystemDManager:
                 "Stale systemd D-Bus proxy (%s), reconnecting and retrying",
                 error.get_dbus_name(),
             )
-            method_name = getattr(fn, "_method_name", None)
             self._connect()
-            if method_name is None or self._manager is None:
-                # No safe way to rebind; retrying fn would call through
-                # the connection we just closed. Better to surface the
-                # original failure than to fail obscurely.
-                raise
-            return self._manager.get_dbus_method(method_name)(*args, **kwargs)
+            return work()
 
     def _ensure_connection(self) -> None:
         """Ensure D-Bus connection is active, reconnect if necessary.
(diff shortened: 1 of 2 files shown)


foxpatch-aleph (changes requested):
The core fix (follow_name_owner_changes=True) correctly addresses the reported ServiceUnknown-after-daemon-reexec 500 by keeping the proxy bound to the well-known name. However, the _call_with_reconnect retry mechanism is broken: dbus _ProxyMethod instances have neither __name__ nor __self__, so the rebinding branch at lines 110-111 is unreachable dead code, and the comment claiming it rebinds fn against the fresh manager is misleading. For ServiceUnknown the retry still happens to work (well-known-name routing), but for Disconnected/NoReply — which the PR explicitly lists in _STALE_CONNECTION_ERRORS — the retry calls the original fn whose _connection still points at the bus that _connect() just closed, so the retry fails. The fix is to use fn._method_name and self._manager.get_dbus_method(fn._method_name). Tests should be added to cover the reconnect/retry path, which would have caught this.

**`src/aleph/vm/systemd.py`** (line 110): This rebinding branch is unreachable. `fn` is a `dbus.proxies._ProxyMethod` instance (returned by `Interface.__getattr__` -> `ProxyObject.get_dbus_method`). `_ProxyMethod` does not define `__name__` or `__self__`, and `Interface.__getattr__` explicitly raises `AttributeError` for any dunder attribute. So `getattr(fn, "__self__", None)` always returns `None`, the `is not None` check is always `False`, and the block is skipped. The retry then invokes the original stale `fn` whose `_ProxyMethod._connection` still references the bus that `_connect()` just closed via `self._bus.close()`.

Net effect: for `ServiceUnknown` the retry still succeeds (because `follow_name_owner_changes=True` keeps `_named_service` as the well-known name, so D-Bus resolves the new owner at call time), but for `Disconnected`/`NoReply` — which the PR explicitly lists in `_STALE_CONNECTION_ERRORS` — the retry fails against the dead connection.

Fix: use the documented `_method_name` attribute and rebind via the Interface's own accessor, e.g.:

```python
if self._manager is not None and hasattr(fn, "_method_name"):
    fn = self._manager.get_dbus_method(fn._method_name)
return fn(*args, **kwargs)
```

A unit test that mocks `_connect` and asserts the retried call goes through the *new* manager would catch this.

**`src/aleph/vm/systemd.py`** (line 106): Comment claims callers pass "a bound method on the previous `_manager`", but `manager.EnableUnitFiles` etc. resolve through `Interface.__getattr__` and return a fresh `_ProxyMethod` per access — they are not Python bound methods and have no `__self__`. The comment should describe the actual mechanism (`_method_name` + `get_dbus_method`) once the code is fixed.

**`src/aleph/vm/systemd.py`** (line 87): No tests are added for `_call_with_reconnect` or for the `follow_name_owner_changes=True` proxy behavior. Given the rebinding bug above is silent (no exception, just dead code), a regression test is warranted — at minimum a test that stubs `_connect`, raises `ServiceUnknown` on the first call, and asserts the retry uses the freshly-built manager's method rather than the stale one.

foxpatch-aleph (approved):
A well-targeted fix for a real production issue where a cached D-Bus proxy pointed at systemd's old unique bus name after `daemon-reexec`, causing 500s on `notify_allocation`. The two-pronged approach is correct: `follow_name_owner_changes=True` handles routine owner changes transparently, and `_call_with_reconnect` covers the narrow mid-restart window by reconnecting and rebinding the method against the fresh manager via the documented `_method_name` attribute. The rebind logic is verified correct — `Interface.get_dbus_method` returns a `_ProxyMethod` bound to the new proxy/connection, so the retry does not go through the closed bus. Stale-error detection is conservative (only three specific error names, `None` safely misses the set), the retry is single-shot (no infinite loop), and the un-rebindable fallback re-raises the original error rather than failing obscurely. Tests cover the happy path, non-stale reraise, the core regression, all three error names, and the missing-`_method_name` path. No security concerns; method names come from internal `_ProxyMethod` objects and are validated by dbus-python.

**`src/aleph/vm/systemd.py`** (line 284): Non-blocking observation: `manager.ListUnits()` here is not wrapped with `_call_with_reconnect`. This is a read-only path with its own try/except that returns all-inactive on error, so it's acceptable for this PR, but it will still surface a hard error in the log (and return a potentially misleading "all inactive" result) after a `daemon-reexec` until the next `_ensure_connection`/reconnect. Worth a follow-up if you want symmetric handling.

foxpatch-aleph (commented):
The core retry-on-stale-proxy mechanism is correct and well-tested. _call_with_reconnect properly rebinds the D-Bus method against the new manager via _method_name + get_dbus_method rather than retrying through the closed connection, and the test suite covers the happy path, non-stale errors, all stale error names, and the missing-_method_name fallback. However, the PR title and description still advertise follow_name_owner_changes=True as part of the fix, but commit 6acc7ec explicitly dropped it — the metadata is now misleading and should be updated. Additionally, only the 5 state-changing methods are wrapped; read methods (is_service_enabled, get_service_active_state, is_service_active) remain unprotected and will still produce 500s on a stale proxy after daemon-reexec, which may leave the original bug partially unresolved depending on which call path notify_allocation hits.

PR title mismatch: The title says "fix(systemd): follow name-owner changes and retry on stale proxy" but commit 6acc7ec dropped follow_name_owner_changes=True. The _connect() docstring correctly explains why (requires a D-Bus main loop that the asyncio supervisor lacks), but the PR title/description were not updated to reflect this. Consider updating the title to something like "fix(systemd): retry on stale proxy after daemon-reexec" to avoid confusing future readers of the git log.

**`src/aleph/vm/systemd.py`** (line 217): is_service_enabled() is not wrapped with _call_with_reconnect. After a daemon-reexec, GetUnitFileState will raise ServiceUnknown on the stale proxy and surface as a 500 (the except branch here logs and returns False, which silently masks the stale-proxy error as "not enabled" — arguably worse than a 500 since it can cause enable_and_start to re-enable an already-enabled unit). Consider wrapping read-path manager calls too, or document why they're intentionally excluded.

**`src/aleph/vm/systemd.py`** (line 238): get_service_active_state() calls manager.GetUnit(service) without _call_with_reconnect protection. If the proxy is stale, this returns "unknown" (via the DBusException handler), which callers must retry on — but the retry will hit the same stale proxy. This is the same class of bug the PR fixes for state-changing methods.

**`src/aleph/vm/systemd.py`** (line 30): Including NoReply in _STALE_CONNECTION_ERRORS means a genuinely slow (but not stale) systemd call will be retried, doubling the effective timeout for that operation. This is a reasonable tradeoff for the daemon-reexec case, but worth a comment noting the consequence. If NoReply timeouts are tunable, the retry could be noticeable on a loaded host.

**`src/aleph/vm/systemd.py`** (line 122): Minor: getattr(fn, "_method_name", None) relies on an internal dbus-python attribute. The source comment in _ProxyMethod.__init__ ("the test suite relies on the existence of this property") gives reasonable confidence it won't disappear, but a brief note that this is a deliberate reliance on a documented-stable internal would help future maintainers understand the coupling. The existing docstring covers this well enough — just flagging it.

**`tests/supervisor/test_systemd.py`** (line 144): del plain_callable._method_name works because MagicMock auto-creates _method_name. This is slightly subtle — if MagicMock ever changes its default attribute behavior this test could silently break. Consider using a non-MagicMock callable (e.g., a plain function with side_effect set via a wrapper) to make the intent explicit. Non-blocking nit.

foxpatch-aleph (approved):
Well-reasoned fix for a real production issue (stale D-Bus proxy after systemd daemon-reexec). The retry-on-stale-error approach is correctly justified over follow_name_owner_changes (which requires a D-Bus main loop the asyncio supervisor lacks). _call_with_reconnect is simple, retries exactly once, and the callable-based design correctly ensures the retry reads the fresh self._manager rather than a pre-resolved dead _ProxyMethod. Tests cover happy path, non-stale reraise, all three stale error names, manager-swap observation, and retry-failure non-looping. The is_service_active refactor is behavior-preserving and cleaner (drops an unused Unit interface). Two minor gaps: get_services_enabled_states is not wrapped despite the PR claiming 'every D-Bus-touching method' is wrapped (currently unused, so not a live bug), and test_retry_failure_propagates has a dead broken_connect variable.

**`src/aleph/vm/systemd.py`** (line 304): get_services_enabled_states is not wrapped with _call_with_reconnect, unlike every other D-Bus-touching method in this PR. The PR description says 'Wrap every D-Bus-touching method' but this one directly calls self._get_manager() and iterates manager.GetUnitFileState(service) without retry coverage. It's currently unused in src/ so it's not a live bug, but it's an inconsistency — after a daemon-reexec this method would fail with ServiceUnknown where the others self-heal. Consider wrapping the per-service GetUnitFileState call (or the whole loop) in _call_with_reconnect for consistency.

**`src/aleph/vm/systemd.py`** (line 217): This line is quite long (91 chars). Consider extracting to a local for readability: `state = self._call_with_reconnect(lambda: self._get_manager().GetUnitFileState(service)); return state == "enabled"`. Minor nit.

**`tests/supervisor/test_systemd.py`** (line 148): broken_connect is defined but never used — the test actually relies on install(MagicMock()) on line 155 to patch _connect. This dead code is confusing; either use it (monkeypatch.setattr(SystemDManager, "_connect", broken_connect)) or remove it.

**`tests/supervisor/test_systemd.py`** (line 74): Good regression test — asserting managers_seen == [initial, new_mgr] is exactly the right invariant to pin down: the callable must observe the manager swap, which is what distinguishes the callable-based retry from a pre-resolved-method retry.

PR #1054, opened 2026-07-25
perf(systemd): run enable_and_start off the event loop

aliel:
## Summary

`SystemDManager.enable_and_start` is declared `async` but its body is entirely **synchronous D-Bus round-trips** — `is_service_enabled` → `enable` → `is_service_active` → `start`. Every VM start blocked the asyncio event loop for ~4 round-trips inline; on a busy CRN this stacks and shows up as multi-second TTFB on unrelated HTTP requests during allocation bursts. Same shape as the `monitor_payments` sweep fix in #963.

## Changes

- Extract `_enable_and_start_sync` with the existing sync body.
- Make `enable_and_start` `await asyncio.to_thread(_enable_and_start_sync)`. Callers already `await`, so no API break.
- Add `threading.Lock` around `_connect()` so the payment monitor worker (already off-thread via #963) and the `enable_and_start` worker cannot race on close+reopen of the shared `SystemBus`. Individual D-Bus calls go through dbus-python's own per-connection locking; only the reconnect path needs external serialization.
- Fast-path check in `_connect` intentionally omitted: callers reach `_connect()` precisely when they know the proxy is stale, and `get_is_connected()` is a local flag that stays True across bus-name rotations. The bounded redundant work (two threads both reconnecting once) is preferable to silently retrying against a stale proxy.

## Test plan
- [x] `test_enable_and_start_runs_off_event_loop` asserts the sync body runs on a non-main thread.
- [x] `test_enable_and_start_propagates_errors` asserts exceptions surface to the async caller.
- [ ] Deploy to staging; verify TTFB stability during a batch of `notify_allocation` calls.

## Depends on
#1053 — this branch is based on `aliel-fix-systemd-stale-proxy` and inherits its wrapped read/write methods. Rebase onto `main` after #1053 merges.

### DIFF
--- a/src/aleph/vm/systemd.py
+++ b/src/aleph/vm/systemd.py
@@ -2,7 +2,9 @@
 async SystemD Manager implementation.
 """
 
+import asyncio
 import logging
+import threading
 
 import dbus
 from dbus import DBusException, SystemBus
@@ -60,6 +62,10 @@ class SystemDManager:
     def __init__(self):
         self._bus: SystemBus | None = None
         self._manager: Interface | None = None
+        # Guards _connect(): concurrent callers (event loop + payment
+        # monitor worker + enable_and_start worker) must not race on
+        # close+reopen of self._bus.
+        self._connect_lock = threading.Lock()
         self._connect()
 
     def _connect(self, max_retries: int = 3) -> None:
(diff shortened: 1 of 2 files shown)


foxpatch-aleph (approved):
Solid performance fix with a well-reasoned threading model. The async-to-thread wrapping is API-compatible (all callers already await), the threading.Lock correctly serializes the only state-mutating path (_connect) without introducing deadlock risk, and _call_with_reconnect's callable-passing design correctly ensures retries route through the fresh manager after a stale-proxy reconnect. Tests are thorough and pass locally. The remaining race (call-in-flight vs. bus.close()) is pre-existing, explicitly documented, and mitigated by dbus-python's connection locking. Minor nits only: dead `broken_connect` in one test, and a couple of over-long lines.

**`tests/supervisor/test_systemd.py`** (line 148): `broken_connect` is defined but never used — `install(MagicMock())` is what actually patches `_connect`. Remove the dead function to avoid confusing future readers.

**`src/aleph/vm/systemd.py`** (line 232): This line is quite long. Consider extracting the lambda body to a local or wrapping for readability, e.g. `state = self._call_with_reconnect(lambda: self._get_manager().GetUnitFileState(service)); return state == "enabled"`.

foxpatch-aleph (approved):
Clean, well-documented performance fix that moves synchronous D-Bus round-trips off the asyncio event loop. The extraction of _enable_and_start_sync and wrapping with asyncio.to_thread is mechanically correct, callers already await so there's no API break, and the threading.Lock around _connect() is the right granularity to prevent close+reopen races between the event loop, payment monitor worker, and enable_and_start worker. The PR description accurately characterizes the benign TOCTOU window in _ensure_connection as bounded redundant work. Tests verify the off-thread execution guarantee and error propagation. No correctness, security, or quality concerns.

**`tests/supervisor/test_systemd.py`** (line 173): Minor nit: `import threading` is done locally inside the test, while `pytest` and `dbus.exceptions` are imported at module level. Consider hoisting `import threading` to the top-level imports for consistency. Non-blocking.

PR #1056, opened 2026-07-30
ci: drop Debian 12 from droplet integration tests

odesenfans:
DigitalOcean removed the `debian-12-x64` image when Debian 12 reached its end of standard support (June 2026), per their [image deprecation policy](https://docs.digitalocean.com/products/droplets/details/image-deprecation/). Every job creating a Debian 12 droplet now fails with:

```
Error: POST https://api.digitalocean.com/v2/droplets: 422 You specified an invalid image for Droplet creation.
```

Changes:
- Remove the Debian 12 entry from the `run_on_droplet` test matrix.
- Move the `run_new_runtime_debian_12` job (which tests the locally-built runtime, not Debian 12 itself) to a Debian 13 host: renamed to `run_new_runtime_debian_13`, uses `debian-13-x64` and the `aleph-vm.debian-13.deb` artifact.

The Debian 12 `.deb` is still built and shipped (podman build, no DigitalOcean dependency, Debian 12 is in LTS until 2028); it is just no longer integration-tested on a droplet. The `aleph-debian-12-python` guest runtime is unaffected: it is the guest rootfs built with debootstrap on the GitHub runner, independent of the droplet host OS.

PR #1057, opened 2026-07-30
ci: extend crun workaround to all podman-using workflows

odesenfans:
Follow-up to #1056. The `containers.conf` workaround for the broken `ubuntu-24.04` runner image ([actions/runner-images#14473](https://github.com/actions/runner-images/issues/14473)) only covered the `build_deb` job of the droplet workflow. Three other workflows still run podman unprotected and fail with `crun: unknown version specified` whenever a job lands on image `20260726.254`:

- `test-using-pytest.yml`: the "Build example volume" step runs `examples/volumes/build_squashfs.sh`, which prefers podman over docker when installed (bit PR #1038 today)
- `release-on-tag.yml`: all four package builds — a release tagged while the broken image is in rotation would coin-flip
- `deploy-main-on-staging.yml`: the staging package build

Same one-step fix in each: point podman at the current crun in `/usr/local/bin`. Harmless on healthy images; remove all four instances once GitHub fixes the runner image.

### DIFF
--- a/.github/workflows/deploy-main-on-staging.yml
+++ b/.github/workflows/deploy-main-on-staging.yml
@@ -30,6 +30,14 @@ jobs:
           # Fetch the whole history for all tags and branches (required for aleph.__version__)
           fetch-depth: 0
 
+      # Runner image 20260726.254 ships podman 5.8.4 but leaves a stale crun 1.14.1
+      # at /usr/bin/crun, breaking every podman run/build with "unknown version
+      # specified". Point podman at the current crun in /usr/local/bin.
+      # See https://github.com/actions/runner-images/issues/14473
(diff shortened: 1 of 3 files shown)


foxpatch-aleph (approved):
This PR correctly extends the crun workaround from PR #1056 to the three remaining podman-using workflows. The workaround block (comment + containers.conf step) is copied verbatim from the already-merged precedent in build-deb-package-and-integration-tests.yml, ensuring consistency. In all three files the step is placed before any podman invocation, and all three workflows run on ubuntu-24.04 (either explicitly or via ubuntu-latest). The fix is harmless on healthy runner images since it just points podman at /usr/local/bin/crun, and the comments clearly document the upstream issue for future cleanup. No correctness, security, or quality concerns.

PR #1058, opened 2026-07-30
feat(rust): make NUMA hugepage headroom and global limit configurable

odesenfans:
Follow-up to #1038, stacked on `od/phase3-c2-numa-memory` (retarget to `dev` happens automatically when #1038 merges).

The boot 2M hugepage reservation hardcoded `DEFAULT_HEADROOM_MB` (8192) and no global cap; the parameterized `reserve_2m_hugepages_in` was only reachable from tests. This surfaces both knobs through the settings:

- `ALEPH_VM_NUMA_HUGEPAGES_HEADROOM_MB` (default 8192): per-node floor of regular RAM kept out of the reservation
- `ALEPH_VM_NUMA_HUGEPAGES_LIMIT_MB` (default unset): global cap on the reservation, split evenly across nodes; unset keeps the total-RAM bound

Both are only read when `ALEPH_VM_NUMA_HUGEPAGES` is on; defaults leave the C2 behavior unchanged, so this is a no-op for existing hosts.

Tests: settings defaults/parse/error cases added; the budget math at different headroom/limit values was already covered by the existing `compute_2m_budget` and `reserve_2m_hugepages_in` tests. 301 crate tests pass, clippy and rustfmt clean.


foxpatch-aleph (approved):
Clean, minimal change that surfaces the previously hardcoded hugepage headroom and the previously test-only global limit as settings, with defaults that preserve existing behavior exactly. The new get_u32/get_u64 env helpers mirror the established get_int/get_i64 pattern, types align end-to-end with the existing reserve_2m_hugepages_in signature, and the call site in main.rs threads the settings through correctly. Test coverage is solid: the new test checks defaults, tuned values, and both non-integer and negative error cases, while the budget math at arbitrary headroom/limit values was already covered by the existing compute_2m_budget and reserve_2m_hugepages_in tests. The divergence doc was updated to match. No correctness, security, or quality concerns.

foxpatch-aleph (approved):
This PR cleanly surfaces two already-parameterized internal knobs (headroom and global limit for the 2M hugepage reservation) through the settings layer. The new get_u32/get_u64 EnvSlice helpers follow the established get_int/get_i64 pattern exactly, correctly rejecting negative and non-integer values via str::parse on unsigned types. The reserve_2m_hugepages function is a thin pass-through to the already-tested reserve_2m_hugepages_in, and main.rs wires the settings through the existing gate. Defaults preserve the prior behavior (DEFAULT_HEADROOM_MB and None), making this a no-op for existing hosts. The new config test covers defaults, custom values, and error cases. One minor observation: no test in hugepages.rs exercises a non-None global_limit_mb flowing through the full reserve_2m_hugepages_in path (all existing tests pass None, and compute_2m_budget tests pass a fixed per_node_cap directly), but the math is simple enough that this is not a blocking concern.

**`rust/crates/supervisor-daemon/src/hugepages.rs`** (line 260): Minor coverage gap: all existing `reserve_2m_hugepages_in` tests pass `None` for `global_limit_mb`, and the `compute_2m_budget` tests exercise the `per_node_cap` parameter in isolation. Consider adding a test that passes a non-`None` `global_limit_mb` through `reserve_2m_hugepages_in` to verify the full integration (limit / num_nodes -> per_node_cap -> compute_2m_budget). For example, with 2 nodes of 64000 MB each and a global limit of 32768, the per-node cap is 16384, so the budget per node should be min(16384, 64000 - 4096) - 1G_pages, which is far smaller than the unlimited case. This is non-blocking since the math is straightforward.

foxpatch-aleph (approved):
Clean, well-documented follow-up that surfaces the previously-hardcoded hugepage headroom and the previously-test-only global limit as operator settings. The new get_u32/get_u64 env helpers mirror the existing get_int/get_i64 pattern exactly, parsing is strict (rejecting negatives and non-integers as startup errors), and the settings are threaded through reserve_2m_hugepages correctly. The compute_2m_budget math in the new global-limit test checks out, and defaults preserve the prior C2 behavior (no-op for existing hosts). Tests cover defaults, custom values, and error cases. Minor non-blocking note: negative-integer handling differs from the nearby HOST_MEMORY_RESERVED_MIB setting (which clamps via get_i64), but the reject-as-error approach taken here is arguably the better one.

**`rust/crates/supervisor-daemon/src/config.rs`** (line 266): Non-blocking observation: HOST_MEMORY_RESERVED_MIB parses via get_i64 then clamps negatives to 0 (line 266: `u64::try_from(value).unwrap_or(0)`), while the new numa_hugepages_headroom_mb / numa_hugepages_limit_mb reject negatives as a hard error. The new behavior is stricter and arguably better, but the inconsistency is worth being aware of if you want uniformity across integer settings.

foxpatch-aleph (approved):
Clean, well-tested PR that surfaces two existing hardcoded knobs (headroom and global limit) through environment settings. The implementation correctly threads the new parameters from Settings through reserve_2m_hugepages to the already-parameterized reserve_2m_hugepages_in. The new get_u32/get_u64 env helpers follow the exact pattern of the existing get_int/get_i64 methods. Test arithmetic is verified correct for all cases, saturating math prevents underflows, edge cases (limit=0, uneven node counts) degrade gracefully, and defaults preserve the prior C2 behavior (no-op for existing hosts). The only minor observation is that defaults_mirror_conf_py doesn't assert the new field defaults, but the dedicated numa_hugepages_headroom_and_limit_are_configurable test covers this thoroughly.

**`rust/crates/supervisor-daemon/src/config.rs`** (line 593): Minor: the `defaults_mirror_conf_py` test asserts `!settings.numa_hugepages` but doesn't assert the new `numa_hugepages_headroom_mb` and `numa_hugepages_limit_mb` defaults here. The dedicated `numa_hugepages_headroom_and_limit_are_configurable` test covers them, so this is not a gap — just a consistency nit if you want all defaults in one place.

foxpatch-aleph (changes requested):
The code changes are clean and correct: the two new settings are threaded through config parsing, the production reserve_2m_hugepages call, and the tracing log with proper defaults and error handling, all mirroring existing patterns. The new test math is verified correct and covers the non-None global limit path. However, docs/plans/rust-port-divergences.md contains an UNRESOLVED GIT MERGE CONFLICT (literal <<<<<<< HEAD / ======= / >>>>>>> markers at lines 90-97), which must be fixed before merge. The conflict is between the old version of divergence entry 77 (hardcoded headroom) and the new version (configurable headroom/limit); the new version on the >>>> side should be kept and the markers removed. Entry 79 from the HEAD side is missing from the resolution side and must be preserved.

**`docs/plans/rust-port-divergences.md`** (line 90): BLOCKING: unresolved git merge conflict. Lines 90-97 contain literal <<<<<<< HEAD, =======, and >>>>>>> markers. The HEAD side has the old entry 77 (hardcoded DEFAULT_HEADROOM_MB) plus entries 78 and 79; the incoming side has the updated entry 77 (configurable headroom/limit) and entry 78, but is MISSING entry 79. The resolution must keep the updated entry 77 from the incoming side, keep entries 78 and 79 from the HEAD side, and remove all conflict markers. As-is, this file is broken.

foxpatch-aleph (approved):
Clean, focused PR that surfaces two previously hardcoded hugepage reservation knobs (headroom and global limit) through environment settings. The types are well-chosen (u32 for per-node headroom, Option<u64> for the global cap), the defaults preserve existing behavior (no-op for hosts that don't set the new vars), and the new get_u32/get_u64 helpers follow the exact pattern of the existing get_int/get_i64. The budget math in the new test is verified correct, error cases (non-integer, negative) are covered, and the divergences doc is updated. The only production caller in main.rs is correctly updated with the new parameters.

PR #1040, opened 2026-07-07
feat(rust): D1 - make the SEV-SNP launch architecture-agnostic (host-derived cbitpos)

odesenfans:
## D1: make the SEV-SNP launch architecture-agnostic (host-derived cbitpos)

Stacks on #1038. Small increment applying a design principle: **the supervisor is a
dumb launcher, agnostic of the host CPU generation; measurement verification is entirely
a client concern.**

B1 hardcoded `cbitpos=51,reduced-phys-bits=1` in the SNP fragment for byte-parity with
the aleph-cvm generator, which tied the supervisor to a specific EPYC generation. But
cbitpos/reduced-phys-bits are **host CPUID properties** (leaf 0x8000001F) and are **not
SEV-SNP measurement inputs** (they configure memory encryption, not the launch digest),
so they must be read from the host, not baked in.

- The controller's `run_snp` now reads `SevHostInfo::read()` (the A2 CPUID reader) and
  threads the host values through `build_snp_argv` -> `snp_tee_fragment`; the
  `SNP_CBITPOS`/`SNP_REDUCED_PHYS_BITS` constants are deleted. A non-SEV host is refused
  cleanly (fail-closed, same as the SEV/SEV-ES path).
- The `aleph-tee` generator `sev_snp_qemu_args` is parameterized on cbitpos/
  reduced-phys-bits and emits the passed values.
- **`-cpu EPYC-v4` stays fixed** (it *is* a measurement input via the VMSA): a fixed
  vcpu-type keeps a program's measurement host-independent, so one pinned measurement
  works on any Genoa-or-newer host. The ARK/product remains a client concern (attest-cli).

The oracle test now crosses two host pairs (51/1 and a non-51 47/2), injecting the same
pair into the controller fragment and the aleph-tee generator and asserting byte-equality
plus that the emitted `sev-snp-guest` object reflects the injected pair, so a re-hardcode
to 51 is caught. Reviewed (consolidated parity/robustness/mutation): clean, no surviving
mutations. Only the SNP path changes; daemon/proto/packaging untouched. Divergence 68
updated (recommend upstreaming the host-read to aleph-cvm).

`cargo fmt`/`clippy -D warnings`/`test` green.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1060, opened 2026-07-31
refactor(aleph-tee): drop the default SEV-SNP C-bit parameters

odesenfans:
## Kill the built-in cbitpos=51/reduced-phys-bits=1 default (D1 follow-up)

Based on dev (D1 #1040 merged). D1 made the launch path host-derived (the controller
reads CPUID leaf 0x8000001F at launch, fail-closed) but left 51/1 as a constructor
**default** in `aleph-tee`'s `SevSnpBackend` for donor API parity. These values vary by
EPYC generation and configure guest memory encryption, so a silently-applied built-in
pair is a launch footgun. This PR deletes the default and encodes "host values required"
in the API:

- `SevSnpBackend` stores `cbit_params: Option<CbitParams>` (`None` from `new()`);
  `with_cbit_params` injects the host CPUID values. The `DEFAULT_CBITPOS` /
  `DEFAULT_REDUCED_PHYS_BITS` constants are gone.
- `TeeBackend::qemu_args` becomes fallible (`Result<Vec<String>>`); the SEV-SNP impl
  bails with a CPUID-pointing message when the pair is unset, so no caller can ever
  launch with a stale hardcode. `NoTeeBackend` wraps its args in `Ok`.
- `new(product)` stays parameter-free for attestation-only users: the in-guest
  attest-agent (the only production constructor call in this repo) uses report
  retrieval/parsing/verification and never generates argv.
- Tests: unset params fail closed; an injected non-default pair (47/2) flows into the
  emitted `sev-snp-guest` object verbatim, which also closes the D1 review nit that
  `with_cbit_params` had no direct unit test.
- Ledger entry 79 (78 is D2's DHCP entry): diverges from the aleph-cvm donor (still
  hardcodes 51/1, infallible `qemu_args`); recommend upstreaming together with the
  entry-68 host-read.

The free generator `sev_snp_qemu_args` already required explicit values (D1) and is
unchanged, as are the controller argv and the byte-parity oracle. `cargo fmt` /
`clippy --workspace --all-targets --locked -D warnings` / `cargo test --workspace
--locked` (17/17 binaries) green.


foxpatch-aleph (approved):
This PR cleanly removes a launch footgun: the hardcoded cbitpos=51/reduced_phys_bits=1 default in SevSnpBackend is replaced with Option<CbitParams> (None from new()), making TeeBackend::qemu_args fallible so no caller can ever launch with a stale EPYC generation's C-bit values. The fail-closed error message points to CPUID leaf 0x8000001F and with_cbit_params(). The trait signature change is correctly propagated to all three implementors (SevSnpBackend, NoTeeBackend, two test mocks). No production caller of the trait method exists — the controller launch path already uses the free generator sev_snp_qemu_args directly, which required explicit values since D1 — so this is pure library-layer hardening with zero blast radius. Tests cover the three key paths (default pair flows through, non-default pair flows verbatim, unset params fail closed), closing the D1 review nit about with_cbit_params lacking a direct test. The CbitParams struct is well-designed with appropriate derives, and documentation is thorough.

foxpatch-aleph (approved):
Clean refactoring that removes the silently-applied C-bit parameter defaults from SevSnpBackend, replacing them with an Option<CbitParams> that fails closed in qemu_args when unset. All four TeeBackend implementations are correctly updated to the fallible signature, the free generator is appropriately left unchanged, and the three new tests directly verify the fail-closed path, the default pair pass-through, and a non-default pair pass-through. The design correctly prevents launch with stale EPYC-generation hardcoded values, and the error message guides the caller to read host CPUID and use with_cbit_params(). No correctness, security, or coverage gaps identified.

foxpatch-aleph (approved):
This PR cleanly removes the hardcoded SEV-SNP C-bit default (51/1) from SevSnpBackend, replacing it with an Option<CbitParams> that is None from new() and must be injected via with_cbit_params before qemu_args can succeed. Making TeeBackend::qemu_args fallible (Result<Vec<String>>) is the right minimal API change: it extends the controller's fail-closed contract to the library layer so no caller can ever launch with a stale EPYC generation's values. All four TeeBackend implementations (SevSnpBackend, NoTeeBackend, and two mock backends in aleph-attest-agent) are updated consistently. The free generator sev_snp_qemu_args correctly remains infallible and unchanged. Tests cover the three key paths: default pair passthrough, non-default pair (47/2) verbatim passthrough, and fail-closed on unset params with a CPUID-pointing error message. The CbitParams struct is well-encapsulated and properly re-exported. No correctness, security, or quality issues.

PR #1061, opened 2026-07-31
fix(rust): recreate the SNP per-tap DHCP server on VM restart

odesenfans:
## Stop/start cycle left SNP VMs without DHCP (D2 follow-up)

Stacks on #1043; found while reviewing the per-VM-DHCP-server design. `stop_vm_execution`
tears the per-tap dnsmasq down together with the tap and nft rules, but
`start_vm_execution` only recreated the tap/NDP/nft side. So a StopVm -> StartVm cycle
on a persistent SNP VM (or the first start after a **host reboot**, where transient
units are gone) rebooted the measured guest with no DHCP server on its tap: the guest
cmdline carries no `ip=` (measurement determinism, ledger 78), udhcpc gets no answer,
the VM never obtains its allocated IP, and attestation is unreachable.

- `start_vm_execution` now rebuilds the `DhcpConfig` and starts the server alongside
  the tap/nft recreation, gated on `config.snp()` exactly like teardown.
  `DhcpBackend::start` replaces a leftover unit (best-effort stop first), so a
  partially-failed stop cannot fail the restart.
- Regression tests: an SNP stop/start cycle re-starts the DHCP server with the same
  allocated IP on the same tap (fails without the fix: 1 start instead of 2); a plain
  VM stop/start cycle still starts none (guards the SNP gate against removal).
- Ledger entry 78 updated with the restart behavior.

Daemon-side only, no image/measurement change. `cargo fmt` / `clippy --workspace
--all-targets --locked -D warnings` / `cargo test --workspace --locked` green
(319 daemon tests incl. the two new ones; 17/17 workspace binaries).

Retarget to `dev` after #1043 merges.


foxpatch-aleph (approved):
The fix correctly addresses the missing DHCP server recreation on VM restart for SNP VMs. `start_vm_execution` now rebuilds the `DhcpConfig` and starts the dnsmasq transient unit alongside the tap/nft recreation, gated on `entry.config.snp().is_some()` exactly like the teardown path. The `SystemdRunDhcp::start` implementation handles leftover units via best-effort stop + reset-failed, so partial stop failures don't block restart. All callers of `start_vm_execution` previously called `stop_vm_execution`, maintaining teardown/recreation symmetry. Two regression tests cover the SNP restart (verifying same IP/tap config) and the plain-VM negative case. The host-reboot scenario mentioned in the PR description is also covered: after a host reboot, the VM is not `adopted_running` (the controller died), so it goes through `start_vm_execution` rather than `reconcile_boot`. A pre-existing ledger entry reference typo (77 vs 78) in dhcp.rs and the stop path comment is noted but not introduced by this PR.

**`rust/crates/supervisor-daemon/src/dhcp.rs`** (line 22): Pre-existing (not introduced by this PR): the module doc says "ledger entry 77" but the DHCP feature is ledger entry 78. Entry 77 is about NUMA hugepage backing. The same typo appears at lifecycle.rs:686 in the stop path comment.

foxpatch-aleph (approved):
A correct and well-tested fix for a real bug: stop_vm_execution tore down the per-tap DHCP server alongside the tap/nft rules, but start_vm_execution only recreated the tap/NDP/nft side, leaving a rebooting SNP VM without DHCP (and thus unreachable for attestation). The fix adds the DHCP server recreation to the start path, gated on entry.config.snp().is_some() exactly like the teardown, using the same DhcpConfig::for_snp and dhcp_lease_dir as the create path so the restarted server hands out the same allocated IP on the same tap. Error handling is fail-closed (returns Err before the unit restart), which is the correct direction for a VM that would be unreachable without DHCP. The two new tests cover the SNP restart cycle (asserting 2 DHCP starts with matching config) and the plain VM restart (asserting no DHCP server is ever started), and the ledger entry 78 documentation is updated to describe the restart behavior. No security, correctness, or quality concerns.

foxpatch-aleph (approved):
The fix is correct, minimal, and well-tested. The DHCP recreation in start_vm_execution is symmetric with the teardown in stop_vm_execution, correctly gated on config.snp().is_some(), and placed at the right scope (always within the networking block, not conditional on tap recreation). DhcpBackend::start handles leftover units gracefully. The two new tests verify the SNP restart path recreates the DHCP server with the correct allocated IP and that plain VMs never get one on restart. The only issues are pre-existing incorrect ledger entry references (77 instead of 78) in comments not touched by this PR.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 686): Pre-existing (not introduced by this PR): the comment says "ledger entry 77" but the DHCP divergence is entry 78 (entry 77 is about hugepages). Since this PR's new comment at line 931 correctly references entry 78, consider fixing this one for consistency.

**`rust/crates/supervisor-daemon/src/dhcp.rs`** (line 22): Pre-existing: module doc says "ledger entry 77" but should be "ledger entry 78" (same issue as the stop-path comment in lifecycle.rs:686).

PR #1062, opened 2026-07-31
refactor(aleph-tee): make TeeBackend attestation-only, drop qemu_args

odesenfans:
## Launch/attestation split, the compile-time version (stacks on #1060)

Follow-up to the #1060 discussion: `cbit_params: Option<CbitParams>` modeled a real
union (attestation-only users vs launchers) forced by the trait bundling both roles.
This PR makes the invalid state unrepresentable instead of failing at runtime:

- `TeeBackend::qemu_args` is **deleted**; the trait is attestation-only
  (get_report / verify_report / parse_report) and its doc points launchers at the
  explicit per-TEE argv APIs.
- `SevSnpBackend` drops `ovmf_path`, `cbit_params`, both builders, and the
  `CbitParams` struct — it now carries only the product name.
- `NoTeeBackend::qemu_args` and the attest-agent mock impls go with the trait method.
- The launch API is the already-explicit free generator
  `sev_snp_qemu_args(config, ovmf_path, cbitpos, reduced_phys_bits)` — what the
  supervisor-controller byte-parity oracle exercises. Its tests keep the non-default
  47/2 coverage from D1, so deleting the backend-level qemu_args tests loses nothing.
- Ledger entry 80: further donor divergence (aleph-cvm's node daemon calls the trait
  method); recommend upstreaming the same split together with entries 68/79.

`qemu_args` was production-dead in this repo: the controller emits its own SNP
fragment (oracle-checked against the free generator) and the in-guest agent never
builds argv. Net -160 lines.

Retarget along the chain as #1060 merges.


foxpatch-aleph (approved):
Clean refactor that makes the launch/attestation split compile-time enforceable by removing `qemu_args` from the `TeeBackend` trait. All implementors are updated consistently, the free generator `sev_snp_qemu_args` retains full test coverage (including the non-default 47/2 case), and no production code paths were relying on the trait method. The only nit is a stale doc comment on `NoTeeBackend` referencing `-cpu host` from the deleted method.

**`rust/crates/aleph-tee/src/none.rs`** (line 10): Stale doc comment: "QEMU is launched with `-cpu host` only" described the now-deleted `qemu_args` method. Since the trait no longer carries launch concerns, this sentence should be removed or reworded to say launch argv is handled elsewhere (e.g., by the supervisor-controller's own QEMU builder).

foxpatch-aleph (approved):
Clean refactor that makes the invalid state (launching with stale C-bit params) unrepresentable at compile time by removing `qemu_args` from the `TeeBackend` trait entirely. All implementors are updated consistently (SevSnpBackend, NoTeeBackend, two attest-agent mocks). The `CbitParams` struct, builders, and associated fields are correctly deleted. The free generator `sev_snp_qemu_args` and its comprehensive tests (including the non-default 47/2 coverage) are untouched, so no test coverage is lost. No stale references to any removed API remain in the Rust workspace. One minor doc nit: NoTeeBackend's doc comment still references the deleted `qemu_args` behavior.

**`rust/crates/aleph-tee/src/none.rs`** (line 10): Doc comment says "QEMU is launched with `-cpu host` only" — this described the now-deleted `NoTeeBackend::qemu_args` return value. Since the trait no longer carries launch concerns, this sentence is stale. Consider removing it or rewording to point to the launcher's responsibility.

PR #1063, opened 2026-08-03
refactor(aleph-tee): drop the verify_report stub from TeeBackend

odesenfans:
## Complete the TeeBackend role split (stacks on #1062)

Follow-up to the question of why verification is still stubbed in the agent: the answer
was that `TeeBackend::verify_report` had zero production callers and could not be
implemented honestly, so this PR deletes it, the same treatment #1062 gave `qemu_args`.

- The SEV-SNP impl did structural parsing and **always returned `valid: false`** (real
  verification is async: it fetches the VCEK/ASK chain from the AMD KDS, which cannot
  fit the sync trait signature). `NoTeeBackend` bailed. Only test mocks implemented it.
- Architecturally, the agent is the **attested party**: a verdict about its own report
  is worthless (a compromised guest would claim validity). B2b (#1036) made the
  verifying client the authoritative gate via the explicit async API
  `sev_snp::verify::verify_sev_snp_report` (ARK-pinned), driven by attest-cli's
  `SnpCertVerifier` during the TLS handshake. That path is unchanged and remains the
  only verification path.
- `TeeBackend` is now the report-producing interface only: `tee_type` +
  `get_report` + `parse_report` — exactly the role the in-guest agent has. The
  `VerificationResult` type stays (it is the real verifier's return type).
- Ledger entry 81: further donor divergence (aleph-cvm's trait keeps the stub);
  recommend upstreaming together with entries 68/79/80.

`cargo fmt` / `clippy --workspace --all-targets --locked -D warnings` /
crate tests (aleph-tee, attest-agent, attest-cli) green.

Retarget along the chain as #1062 merges.


foxpatch-aleph (approved):
Clean removal of the `verify_report` stub from `TeeBackend`, completing the role split begun in entries 79/80. The trait method had zero production callers, always returned `valid: false` on the SEV-SNP impl (real verification is async and fetches the AMD cert chain, which cannot fit the sync signature), and was architecturally wrong (the attested party's verdict about its own report is worthless). All four implementors are updated, the `VerificationResult` type is correctly retained for the real async verifier, imports are properly scoped, and the trait doc comment clearly explains why neither launch nor verification belongs. The divergence ledger entry 81 is thorough and consistent with the prior entries.

PR #1064, opened 2026-08-03
fix(storage): thread pool0_only through the devmapper volume chain

odesenfans:
Review follow-up 1/4 for #1049 (raised in three of the four review passes).

`get_volume_path` forwarded `pool0_only` on the ext4 branch but dropped it on the
devmapper branch: `create_devmapper` -> `create_volume_file` resolved the backing file
with default all-pools placement. Unreachable today (devmapper is QEMU-only), but a
hypothetical Firecracker parent volume would land off pool 0 and hit the jailer
hardlink-copy data loss the pin exists to prevent. The flag now rides the whole chain,
so the invariant is enforced rather than assumed.

Tests: devmapper branch forwards the flag; `create_volume_file` honors `pool0_only`
over a freer extra pool. 52/52 storage-pool tests pass; ruff format clean; mypy
unchanged vs base.

### DIFF
--- a/src/aleph/vm/storage.py
+++ b/src/aleph/vm/storage.py
@@ -323,10 +323,12 @@ async def create_ext4(path: Path, size_mib: int) -> bool:
     return True
 
 
-async def create_volume_file(volume: PersistentVolume | RootfsVolume, namespace: str) -> Path:
+async def create_volume_file(
+    volume: PersistentVolume | RootfsVolume, namespace: str, *, pool0_only: bool = False
+) -> Path:
     volume_name = volume.name if isinstance(volume, PersistentVolume) else "rootfs"
     # Assume that the main filesystem format is BTRFS
-    path = volume_path_for(namespace, f"{volume_name}.btrfs", volume.size_mib)
+    path = volume_path_for(namespace, f"{volume_name}.btrfs", volume.size_mib, pool0_only=pool0_only)
     if not path.is_file():
         logger.debug(f"Creating {volume.size_mib}MB volume")
         # Create an empty file the right size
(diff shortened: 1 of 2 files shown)


foxpatch-aleph (approved):
Clean, well-documented fix that threads the `pool0_only` flag through the devmapper volume chain (`get_volume_path` → `create_devmapper` → `create_volume_file` → `volume_path_for`). The keyword-only parameter with a `False` default maintains backward compatibility, and the only two call sites are both updated. Tests verify both the forwarding wiring and the actual pool 0 placement behavior. The docstring and comments clearly explain the Firecracker jailer hardlink-copy data loss risk that motivates the invariant.

foxpatch-aleph (approved):
A clean, well-scoped fix that threads the pool0_only flag through the devmapper volume chain (get_volume_path -> create_devmapper -> create_volume_file -> volume_path_for), closing a gap where the flag was dropped on the devmapper branch. The change is backward-compatible (keyword-only param with False default), both production call sites are updated, and the docstring clearly explains the rationale. Two new tests provide good coverage: one verifies the flag is forwarded from get_volume_path to create_devmapper, the other verifies create_volume_file honors pool0_only even when a non-zero pool has more free space. No correctness, security, or quality issues found.

foxpatch-aleph (approved):
The PR correctly threads the `pool0_only` flag through the devmapper volume chain (`get_volume_path` -> `create_devmapper` -> `create_volume_file` -> `volume_path_for`), closing a gap where the flag was forwarded on the ext4 branch but silently dropped on the devmapper branch. Both new parameters are keyword-only with `False` defaults, so the change is fully backward compatible. The two new tests are well-targeted: one verifies `pool0_only` is forwarded as a kwarg from `get_volume_path` into `create_devmapper`, and the other verifies `create_volume_file` honors the flag for pool placement even when a non-pool-0 pool has more free space. All 52 storage-pool tests pass. The docstring on `create_devmapper` and the inline comment at the call site clearly explain the jailer hardlink-copy rationale.

PR #1065, opened 2026-08-03
refactor(migration): drop the dead ExportJob.volumes_dir field

odesenfans:
Review follow-up 2/4 for #1049 (raised in three of the four review passes).

`job.volumes_dir` was written but never read: the download view resolves through
`job.export_paths`. Removing it also removes the `iter_namespace_dirs` scan that
existed only to feed the dead assignment, which resolves the review's separate
double-scan observation in the same stroke.

The review called it fully dead; grep verification found three test-only usages it
missed (an assertion on the dead field and two `ExportJob(volumes_dir=...)`
constructions), updated here to assert the meaningful outcome instead.

118 migration + storage-pool tests pass; ruff format clean.

### DIFF
--- a/src/aleph/vm/agent/migration/jobs.py
+++ b/src/aleph/vm/agent/migration/jobs.py
@@ -42,7 +42,6 @@ class ExportJob:
     token: str | None = None
     disk_files: list[DiskFileInfo] | None = None
     export_paths: list[Path] = field(default_factory=list)
-    volumes_dir: Path | None = None
     active_downloads: int = 0
     error: str | None = None
     task: asyncio.Task | None = None
--- a/src/aleph/vm/agent/migration/runner.py
+++ b/src/aleph/vm/agent/migration/runner.py
@@ -132,16 +132,15 @@ async def run_export(
             # until the controller's SIGTERM handler has ACPI-powered the guest
             # down and QMP-quit QEMU with a cache flush. The exported overlay is
             # therefore consistent once stop_vm returns; no separate graceful
-            # step is needed. The volumes dir is the same settings-derived path
-            # the import runner stages into.
+            # step is needed.
             await supervisor.stop_vm(job.vm_hash)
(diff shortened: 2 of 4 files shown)


foxpatch-aleph (approved):
Clean removal of a genuinely dead field. The download handler resolves export files through job.export_paths (migration.py:214), never through volumes_dir, so the field was write-only. Eliminating the redundant iter_namespace_dirs call that fed it also resolves the double-scan observation. Test updates are correct: the new assertion checks the meaningful export_paths outcome, and the volumes_dir= removals from download test fixtures are safe since those tests fail at token-check or exercise file-not-found through export_paths. One minor nit: tmp_path is now an unused fixture parameter in test_download_invalid_token and test_download_file_not_found.

**`tests/supervisor/views/test_migration.py`** (line 269): Minor nit: `tmp_path` is now unused in this test since `volumes_dir=tmp_path` was removed from the ExportJob constructor. Same for `test_download_file_not_found` at line 290. Could drop the fixture parameter to keep the signatures honest, though it's harmless.

foxpatch-aleph (approved):
Clean removal of a genuinely dead field. The `volumes_dir` attribute was written but never read — the download handler resolves through `job.export_paths`. Removing it also eliminates the redundant `iter_namespace_dirs` scan that existed only to feed the dead assignment, resolving the double-scan observation. The `iter_namespace_dirs` import is correctly retained for `_collect_export_disks`. Test changes are appropriate: the assertion in `test_local_migration_p2p.py` now checks the meaningful outcome (`export_paths`) instead of the dead field, and the two `ExportJob` constructions in `test_migration.py` correctly drop the unused parameter. Comment relocation is well-placed.

**`src/aleph/vm/agent/migration/runner.py`** (line 143): Minor nit: the blank line between the comment block and the `for` loop (line 144) is slightly unusual for this codebase's style — nearby code places comments directly above the statement they describe. Not blocking.

PR #1066, opened 2026-08-03
fix(migration): never collect stale export artifacts as source disks

odesenfans:
Review follow-up 3/4 for #1049.

`_collect_export_disks` globs `*.qcow2`, which also matches
`rootfs.qcow2.export.qcow2` (a previous attempt's artifact). The reaper only sweeps
at startup, so an export retried after a failure without an agent restart would
collect the stale artifact as a source disk and re-export an export. Pre-existing
behavior, but the multi-pool scan widens exposure.

One-line filter plus the regression test the review suggested (stale artifact next to
its source volume is skipped). 72 tests pass; ruff format clean.

### DIFF
--- a/src/aleph/vm/agent/migration/runner.py
+++ b/src/aleph/vm/agent/migration/runner.py
@@ -61,6 +61,12 @@ def _collect_export_disks(vm_hash: str) -> list[Path]:
     seen_names: set[str] = set()
     for volumes_dir in iter_namespace_dirs(vm_hash):
         for qcow2_file in sorted(volumes_dir.glob("*.qcow2")):
+            if qcow2_file.name.endswith(".export.qcow2"):
+                # A stale export artifact from an earlier failed run: the
+                # reaper only sweeps at startup, so a retry without an agent
+                # restart would otherwise collect it as a source disk and
+                # re-export an export.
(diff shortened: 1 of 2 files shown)


foxpatch-aleph (approved):
A correct, minimal fix that filters stale `.export.qcow2` artifacts from the glob results in `_collect_export_disks`, preventing a retry-without-restart from re-exporting an export. The filter matches the exact naming convention used at export creation time (`runner.py:153`) and is correctly placed before the dedup logic. The regression test is clear and follows existing test patterns.

foxpatch-aleph (approved):
The fix correctly filters out stale `.export.qcow2` artifacts from the `_collect_export_disks` glob results, preventing a retry-without-restart from re-exporting an export. The `endswith(".export.qcow2")` check is precise: export artifacts are always named `{original}.qcow2.export.qcow2` via `Path.with_suffix`, so no legitimate source disk can match. The filter is placed before the `seen_names` dedup check, which is the right order — a stale artifact must not consume a basename slot and block the real disk. The regression test places both a source and its stale export artifact in the same namespace dir and asserts only the source is collected, directly validating the scenario described in the PR.

PR #1067, opened 2026-08-03
fix(storage): normalize pool paths for adoption-registry comparison

odesenfans:
Review follow-up 4/4 for #1049.

The orphaned-pool guard and missing-marker refusal compared raw `str(path)` against
the registry. A cosmetic rewrite of a VOLUME_POOLS entry (`..` segments, doubled
separators) produced a different string for the same directory, and a mismatch is a
**hard startup failure** (false orphan), not the minor robustness issue the review
rated it as.

Two deliberate deviations from the review's suggestion:
- `os.path.normpath` instead of `Path.resolve()`: resolving follows symlinks, which
  would change pool identity across a symlink rewrite (silently merging or splitting
  pools). Lexical normalization only forgives rewrites of the same configured text.
- Normalization applies on registry **read** too, so registries written before this
  change (or edited by hand) keep matching.

Tests: a `..`-rewritten entry of an adopted pool boots cleanly; an unnormalized
registry entry still triggers the unmounted-disk refusal when the marker is gone.
52 tests pass; ruff + mypy clean.

### DIFF
--- a/src/aleph/vm/storage_pools.py
+++ b/src/aleph/vm/storage_pools.py
@@ -133,11 +133,23 @@ def _parse_pool_entry(entry: str) -> tuple[Path, MediaClass | None]:
                 "(expected nvme, ssd or hdd)"
             )
             raise StoragePoolConfigError(msg) from error
-    path = Path(path_part.strip())
-    if not path_part.strip() or not path.is_absolute():
+    stripped = path_part.strip()
+    path = Path(_canonical(stripped)) if stripped else Path("")
+    if not stripped or not path.is_absolute():
         msg = f"VOLUME_POOLS entry {entry!r} needs an absolute pool path, e.g. /mnt/nvme1/aleph-volumes"
         raise StoragePoolConfigError(msg)
     return path, override
+
+
+def _canonical(path: Path | str) -> str:
+    """The canonical string form of a pool path, for registry storage and
+    comparison: lexically normalized (redundant separators and ``..``
+    segments collapsed) but deliberately NOT symlink-resolved. Resolving
+    would change pool identity across a symlink rewrite, silently merging or
+    splitting pools; a lexical form only forgives cosmetic rewrites of the
+    same configured path. A false mismatch here is a startup failure (the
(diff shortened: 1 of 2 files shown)


foxpatch-aleph (approved):
This PR correctly and consistently applies lexical path normalization (os.path.normpath) at all four comparison/storage points in the adoption registry lifecycle: _parse_pool_entry, _load_adopted, _record_adopted, and the orphaned-pool guard in setup_pools. The choice of normpath over Path.resolve() is well-justified in the docstring — symlink resolution would change pool identity across symlink rewrites. The empty-string edge case in _parse_pool_entry is properly guarded. Two new tests cover the critical scenarios: a cosmetic path rewrite (.. segments, doubled separators) no longer triggers a false orphan, and an unnormalized registry entry still triggers the unmounted-disk refusal. Existing tests continue to pass because their clean tmp_path values are unaffected by normpath. No correctness, security, or quality issues found.

**`src/aleph/vm/storage_pools.py`** (line 182): Minor observation (non-blocking, pre-existing): `_canonical(entry)` is called on each registry entry without validating that `entry` is a string. A hand-edited registry containing a non-string element (e.g. `[123]` or `[null]`) would be silently coerced to a string by `str()` inside `_canonical` and then treated as an orphan path, producing a confusing startup error. The old code (`set(adopted)`) had the same issue, so this is not a regression. If you want to harden it, consider adding `if not isinstance(entry, str)` validation alongside the existing list-type check, but this can be deferred to a follow-up.

foxpatch-aleph (approved):
Well-reasoned, minimal fix that applies lexical path normalization (os.path.normpath) consistently at all four comparison and storage points in the adoption registry lifecycle. The deliberate choice of normpath over Path.resolve() to avoid symlink-based pool identity changes is correct and well-documented. Normalization on registry read ensures backward compatibility with pre-existing or hand-edited registries. Both new tests cover the key scenarios (cosmetic rewrite not triggering false orphan, unnormalized registry entry still triggering the missing-marker guard). All 52 tests pass; ruff and mypy are clean. No correctness, security, or quality issues found.

PR #1070, opened 2026-08-04
[DO NOT MERGE] integration: V-PROGRAM support (deb builder for testnets)

odesenfans:
Throwaway aggregation branch: current dev + #1052 (scheduler support) + #1050 (bundle tooling), merged clean, 120 combined tests green.

**DO NOT MERGE.** Exists only to build candidate `aleph-vm.<distro>.deb` artifacts for the aleph-testnets V-PROGRAM end-to-end scenario (Milestone A: CCN -> scheduler -> CRN allocation control-plane validation). Close once #1050/#1052 merge to dev, like #1041/#1042/#1068 before it.

PR #1052, opened 2026-07-10
feat: V-PROGRAM scheduler support (allocations, type threading, SNP vCPU advertising)

odesenfans:
Adds scheduler-facing support for the V-PROGRAM message type (verifiable SEV-SNP programs), per the design in docs/plans/2026-07-11-vprogram-scheduler-support-design.md. Pairs with aleph-vm-scheduler PR #193 (od/vprogram-scheduling).

## What this does

- POST /control/allocations accepts a new v_programs bucket (matching the scheduler wire model after the persistent_programs -> v_programs rename in #193). Absence from the allocation stops a running v-program: the scheduler is the single source of truth for this type, unlike credit-paid or confidential instances which the stop loop spares.
- V-PROGRAM messages parse through storage/get_message, resolve no amends (immutable by schema), and map to VmType.v_program with IPv6 type value 4 (mirrors scheduler-events VmType::ipv6_value()).
- No node-side payment checks: v-programs are excluded from the check_payment sweep. The CCN and the scheduler already enforce their credit budget.
- On-demand /vm/{item_hash} paths reject v-programs (scheduler-controlled only).
- Launch is deliberately out of scope: create fails cleanly with VmSetupError until the SNP launch increment (runtime manifest fetch, measured cmdline, verified volumes) lands on top of the Phase 3 backport.
- Capability advertising: /about/usage/system and /about/capability gain a platform-keyed properties.tee block with tee.sev_snp.supported_vcpu_types, populated by a QEMU QMP probe (query-cpu-definitions under KVM, EPYC-family models with empty unavailable-features). This lets the scheduler match a v-program's launch-measurement vcpu_type against what the node can actually launch, instead of the too-broad AuthenticAMD + sev_snp flag. A failed probe advertises nothing. check_amd_sev_snp_supported now also requires /dev/sev, like the SEV/SEV-ES probes.

## Merge gate

- [ ] aleph-message is git-pinned to the PR #158 head (bc9c0f8); swap for a released version before merging.
- [ ] aleph-vm-scheduler #193 renames its allocation bucket to v_programs.

## Testing

Canonical cross-SDK fixture (item_hash 4c319b6b...) drives new suites tests/supervisor/test_vprogram.py and test_vprogram_probe.py; full supervisor suite at the origin/dev baseline locally (only the known pyroute2/journald environment failures). The QMP probe was exercised live against this host's QEMU (reports 13 runnable EPYC models).

PR #1071, opened 2026-08-04
feat(vprogram): SEV-SNP launch path for V-PROGRAM messages

odesenfans:
Boots a V-PROGRAM as an SEV-SNP confidential guest. Replaces the "not implemented" seam left by #1052 (`run.py`) with the real launch path.

## ⚠️ Dependency (diamond)
This stacks on **#1052** (base) but also requires **#1050**'s `aleph.vm.vprogram.manifest` module (`vprogram_launch.py` imports `RuntimeManifest`). #1050 and #1052 are independent branches off dev, so the intended merge order is: **#1050 → dev, #1052 → dev, then rebase this onto dev** (both deps present, CI green). Until #1050 lands, CI here is red on the missing import — the diff is deliberately kept to the launch code only rather than vendoring #1050's module.

## What it does
On a V-PROGRAM allocation, the agent now (new `src/aleph/vm/agent/vprogram_launch.py`):
1. Fetches + strict-parses the runtime manifest (`content.runtime.ref` STORE → `RuntimeManifest`).
2. Fetches the bundle (`manifest.bundle.ref`), **verifies sha256 + size** against the manifest (fail closed), extracts the measured image into `EXECUTION_ROOT/vprogram/<vm_hash>` with `tarfile` `filter="data"` (traversal-safe).
3. Builds an SNP `CreateVmSpec` (QEMU, `tee.backend=SEV_SNP`, kernel/initrd/OVMF from the extracted members, disks `[rootfs RAW ROOTFS ro, hash_tree RAW EXTRA ro]` in that order, `persistent=True`). The daemon derives the measured cmdline from the `{rootfs}.roothash` sidecar (proto has no cmdline field); staging fails closed if a shipped sidecar disagrees with `manifest.boot.platform_roothash`, and synthesizes it from the manifest when absent.
4. `create_vm` → `_wait_until_running` → persist, mirroring the confidential-instance path (early record, capacity check after download, teardown on failure). SNP VMs never `awaiting_confidential_init`.

The `run.py` seam is small (build spec via the helper, create/wait/persist). `content.workload`/`content.volumes` are **not** attached yet — the current measured image runs the built-in `aleph.builtin/1` httpd; user-workload volumes (`workload_roothash`/`verified_volumes`) need guest-init support first (a follow-up, per #1050's design).

## Tests
`tests/supervisor/test_vprogram_launch.py` (9): manifest parse/reject, sha256+size mismatch fail-closed with nothing staged, full spec construction, roothash+verity sidecar present/synthesized/disagreement, missing member. `tests/supervisor/test_vprogram.py` updated: the obsolete "fails cleanly" test replaced with launch + build-failure-forgets-record. 21 pass on the integration branch; `ruff format`/`isort`/`mypy` clean.

## Validated end-to-end on real hardware
On aleph-testnets, a V-PROGRAM published to the CCN was scheduled to the SNP CRN, which fetched the manifest + bundle, extracted the measured image, and **booted the SEV-SNP guest** (journal: `started per-tap DHCP server for SNP VM … guest_ip=172.16.4.2`; the item hash reached RUNNING in the execution list). Aggregated deb-builder: PR #1070.

### DIFF
--- a/src/aleph/vm/agent/capacity.py
+++ b/src/aleph/vm/agent/capacity.py
@@ -19,11 +19,7 @@ from dataclasses import dataclass, field
 from datetime import datetime, timedelta, timezone
 
 import psutil
-from aleph_message.models import (
-    ExecutableContent,
-    ItemHash,
-    VerifiableProgramContent,
-)
+from aleph_message.models import ExecutableContent, ItemHash, VerifiableProgramContent
 from aleph_message.models.execution.instance import InstanceContent
 
 from aleph.vm import storage_pools
--- a/src/aleph/vm/agent/capacity.py
+++ b/src/aleph/vm/agent/capacity.py
@@ -19,7 +19,11 @@ from dataclasses import dataclass, field
 from datetime import datetime, timedelta, timezone
 
 import psutil
-from aleph_message.models import ExecutableContent, ItemHash
+from aleph_message.models import (
+    ExecutableContent,
+    ItemHash,
+    VerifiableProgramContent,
+)
 from aleph_message.models.execution.instance import InstanceContent
 
 from aleph.vm import storage_pools
@@ -217,7 +221,12 @@ class CapacityManager:
             record_vcpus = resources.vcpus
             if not memory and not record_vcpus:
                 continue
-            if isinstance(record.message, InstanceContent):
+            # V-PROGRAMs are full SNP VMs, admitted against the instance
+            # bucket (run.py passes is_instance=True), so they must also be
+            # counted there. Bucketing them as programs would both starve the
+            # small program bucket and hide their memory from instance
+            # admission (silent over-commit).
+            if isinstance(record.message, (InstanceContent, VerifiableProgramContent)):
                 committed_instance_memory_mib += memory
             else:
                 committed_program_memory_mib += memory
--- a/src/aleph/vm/agent/run.py
+++ b/src/aleph/vm/agent/run.py
@@ -29,6 +29,7 @@ from aleph.vm.agent.translate import build_create_vm_spec, build_program_create_
 from aleph.vm.agent.update_watcher import UpdateWatcher
 from aleph.vm.agent.vm.program_client import ProgramGuestClient
 from aleph.vm.agent.vm_registry import AgentVmRegistry, persist_record
+from aleph.vm.agent.vprogram_launch import build_vprogram_spec
 from aleph.vm.conf import settings
 from aleph.vm.resources import InsufficientResourcesError
 from aleph.vm.supervisor_interface import errors as supervisor_errors
@@ -344,13 +345,41 @@ async def create_vm_execution(
         return None
 
     if isinstance(content, VerifiableProgramContent):
-        # Scheduler wiring for V-Programs lands before the launch path: the
-        # allocation is accepted, but building a CreateVmSpec for an SNP guest
-        # (runtime manifest fetch, measured cmdline, verified volumes) is not
-        # implemented yet. Fail with the create-path vocabulary so
-        # update_allocations reports it per-VM and the scheduler can react.
-        msg = f"V-PROGRAM {vm_hash} accepted, but this CRN does not implement the SEV-SNP launch path yet"
-        raise VmSetupError(msg)
+        # V-PROGRAM: an auto-booting SEV-SNP VM. build_vprogram_spec fetches
+        # the runtime manifest, integrity-checks and stages the measured
+        # bundle, and returns a spec whose TeeConfig takes the SNP launch
+        # path. Mirrors the instance path above: record early, build spec,
+        # create, wait, persist; forget (and tear down) on failure. Unlike
+        # SEV instances there is no owner session to wait for: the daemon
+        # boots an SNP VM immediately (awaiting_confidential_init is never
+        # set), so the plain readiness wait applies.
+        record = registry.record(vm_hash, message=content, original=original_message.content, persistent=True)
+        try:
+            spec = await build_vprogram_spec(vm_hash, content)
+            # Agent-side admission after the download, like the instance path:
+            # a failed bundle fetch never consumes capacity.
+            capacity.check_capacity(
+                memory_mib=content.resources.memory,
+                vcpus=content.resources.vcpus,
+                disk_mib=0,
+                is_instance=True,
+                exclude_vm_hash=vm_hash,
+            )
+            info = await supervisor.create_vm(spec)
+        except Exception:
+            registry.forget(vm_hash)
+            raise
+        try:
+            await _wait_until_running(supervisor, info.vm_id)
+        except Exception:
+            registry.forget(vm_hash)
+            try:
+                await supervisor.delete_vm(info.vm_id)
+            except Exception:
+                logger.exception("Teardown of half-started V-PROGRAM %s failed", vm_hash)
+            raise
+        await persist_record(vm_hash, record)
+        return None
 
     # Every supported content type is handled above: programs through the spec
     # program path, instances (plain, GPU, confidential) through the spec path.
--- a/src/aleph/vm/agent/vprogram_launch.py
+++ b/src/aleph/vm/agent/vprogram_launch.py
@@ -1,0 +1,231 @@
+"""SEV-SNP launch path for V-PROGRAM messages: runtime bundle staging and
+CreateVmSpec construction.
+
+A V-PROGRAM pins its measured platform through a runtime manifest (a STORE
+message holding JSON parsed by ``aleph.vm.vprogram.manifest.RuntimeManifest``).
+The manifest pins the runtime bundle: ONE tar.gz holding the measured OVMF,
+kernel, initrd and the dm-verity platform rootfs plus its hash tree.
+
+This module is the agent half of the launch: fetch the manifest, fetch and
+integrity-check the bundle (sha256 pinned by the manifest), extract it into a
+per-VM staging directory, make sure the dm-verity sidecars sit where the
+supervisor daemon looks for them, and build the SNP CreateVmSpec. The daemon
+derives the measured kernel cmdline itself from the ``<rootfs>.roothash``
+sidecar next to the rootfs disk and reads the hash tree at ``<rootfs>.verity``
+(see rust/crates/supervisor-daemon/src/lifecycle.rs, snp_config_slice, and
+docs/plans/rust-port-divergences.md entry 68c): the proto deliberately has no
+cmdline field, so nothing here passes one.
+
+Every check fails closed with VmSetupError: a mismeasured or tampered bundle
+must never reach create_vm.
+"""
+
+from __future__ import annotations
+
+import hashlib
+import logging
+import shutil
+import tarfile
+from pathlib import Path
+from typing import TYPE_CHECKING
+
+from pydantic import ValidationError
+
+from aleph.vm.conf import settings
+from aleph.vm.storage import get_existing_file
+from aleph.vm.supervisor_interface.errors import VmSetupError
+from aleph.vm.supervisor_interface.types import (
+    Backend,
+    CreateVmSpec,
+    DirectoryPath,
+    DiskFormat,
+    DiskRole,
+    DiskSpec,
+    NetworkConfig,
+    TeeBackend,
+    TeeConfig,
+    VmId,
+)
+from aleph.vm.utils import get_hostname_from_hash
+from aleph.vm.vprogram.manifest import RuntimeManifest
+
+if TYPE_CHECKING:
+    from aleph_message.models import ItemHash
+    from aleph_message.models.execution.vprogram import VerifiableProgramContent
+
+logger = logging.getLogger(__name__)
+
+_SHA256_CHUNK_SIZE = 1024 * 1024
+
+
+def vprogram_staging_dir(vm_hash: ItemHash) -> Path:
+    """The per-VM directory the runtime bundle is extracted into."""
+    return Path(settings.EXECUTION_ROOT) / "vprogram" / str(vm_hash)
+
+
+async def fetch_runtime_manifest(runtime_ref: str) -> RuntimeManifest:
+    """Download and parse the runtime manifest pinned by ``runtime.ref``.
+
+    The ref is the item hash of a STORE message whose file is the manifest
+    JSON. Parsing is strict (extra keys, malformed templates and traversing
+    member paths are all rejected by the model); any failure is a
+    VmSetupError so update_allocations reports it per-VM.
+    """
+    manifest_path = await get_existing_file(runtime_ref)
+    try:
+        return RuntimeManifest.model_validate_json(manifest_path.read_bytes())
+    except (ValidationError, ValueError, OSError) as error:
+        msg = f"runtime manifest {runtime_ref} is invalid: {error}"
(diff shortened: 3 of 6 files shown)


foxpatch-aleph (approved):
Well-structured SEV-SNP launch path that correctly mirrors the existing confidential instance pattern in run.py. Security is solid: bundle integrity verified via sha256+size before extraction, tarfile filter="data" for traversal safety, defense-in-depth path containment check, and fail-closed roothash sidecar validation. Test coverage is comprehensive for the main paths. Two minor test gaps (extract failure path, hardlink-to-copy fallback) and a test isolation nit (EXECUTION_ROOT not monkeypatched to tmp_path) are non-blocking observations. The diamond dependency on #1050 is clearly documented and the diff is appropriately scoped to the launch code only.

**`src/aleph/vm/agent/vprogram_launch.py`** (line 119): If `tarfile.extractall` fails mid-extraction, the partially-populated `dest` directory is left behind. On retry this is self-healing (the `rmtree` at line 113-114 cleans it up), but consider adding `shutil.rmtree(dest, ignore_errors=True)` in the `except` block to avoid leaving stale partial state visible to concurrent observers or debugging tools between failures.

**`tests/supervisor/test_vprogram_launch.py`** (line 130): The `storage_files` fixture monkeypatches `get_existing_file` but does not isolate `settings.EXECUTION_ROOT` to `tmp_path`. The `vprogram_staging_dir` function reads `settings.EXECUTION_ROOT` at call time, so tests that check `not staging.exists()` (line 185) or compare paths against `staging` (line 209-216) depend on whatever `EXECUTION_ROOT` is configured in the test environment. Consider monkeypatching `settings.EXECUTION_ROOT` to `tmp_path` in the fixture for full isolation.

**`tests/supervisor/test_vprogram_launch.py`** (line 294): Missing test for the `_extract_bundle` failure path: a tarball that passes sha256+size verification but is corrupted internally (e.g. truncated gzip stream) should raise `VmSetupError` from the `tarfile.TarError` handler. This would cover the `except (tarfile.TarError, OSError)` branch at line 119 of vprogram_launch.py.

**`tests/supervisor/test_vprogram_launch.py`** (line 250): Missing test for the `_ensure_verity_sidecars` hardlink-to-copy fallback (line 167-169 of vprogram_launch.py). When `hardlink_to` raises `OSError` (e.g. cross-device link), the code falls back to `shutil.copyfile`. This branch is currently untested.

foxpatch-aleph (changes requested):
The V-PROGRAM SEV-SNP launch path is well-designed and security-conscious: fail-closed integrity verification, safe tar extraction with filter="data", path traversal protection, and proper roothash sidecar validation. However, there is a capacity accounting bug: V-PROGRAMs are admitted against the instance memory bucket (is_instance=True) but _committed_resources in capacity.py counts them as programs (VerifiableProgramContent falls into the else branch of the isinstance check), creating an asymmetry that can silently over-commit the program bucket. This should be fixed in _committed_resources to count V-PROGRAMs as instances, matching how they are admitted. Additionally, the test suite is missing coverage for the _wait_until_running failure → teardown path (run.py:372-380), where create_vm succeeds but the VM fails to reach RUNNING, requiring registry.forget + supervisor.delete_vm.

**`src/aleph/vm/agent/capacity.py`** (line 220): Bucket mismatch: _committed_resources uses `isinstance(record.message, InstanceContent)` to decide the memory bucket. V-PROGRAMs have VerifiableProgramContent, so they are counted in the program bucket. But the V-PROGRAM launch path (run.py:365) checks capacity with `is_instance=True`, admitting against the instance bucket. This means V-PROGRAMs consume instance-tier memory but are tracked as program-tier, allowing the program bucket to be silently over-committed. Fix: also count VerifiableProgramContent as instance-tier here, e.g. `if isinstance(record.message, (InstanceContent, VerifiableProgramContent)):`

**`src/aleph/vm/agent/run.py`** (line 372): Missing test coverage: no test exercises the path where create_vm succeeds but _wait_until_running fails, triggering the teardown (registry.forget + supervisor.delete_vm). Consider adding a test where supervisor.get_vm returns a FAILED/STOPPED status (or raises), and assert delete_vm is called and the registry record is forgotten.

**`src/aleph/vm/agent/vprogram_launch.py`** (line 197): Minor: `settings.CONFIDENTIAL_SESSION_DIRECTORY or (Path(settings.EXECUTION_ROOT) / "sessions")` is redundant — the conf model post-init (conf.py:706-707) already defaults CONFIDENTIAL_SESSION_DIRECTORY to EXECUTION_ROOT/sessions when unset, so it is never None at runtime. Not a bug, but the fallback is dead code. The existing instance path (translate.py:93) uses `settings.CONFIDENTIAL_SESSION_DIRECTORY / vm_hash` directly.

foxpatch-aleph (commented):
Well-structured implementation of the SEV-SNP launch path with strong fail-closed security posture (sha256+size verification before extraction, tarfile filter='data', path traversal prevention, roothash sidecar disagreement check). The code mirrors the existing instance launch path correctly and tests cover the key integrity gates. One concern: the `except (tarfile.TarError, OSError)` handler in `_extract_bundle` does not catch `TypeError`, which would be raised on Python 3.10/3.11 patch versions that lack the PEP 706 `filter` backport (pre-3.10.12/3.11.4). The `TypeError` would escape as an unhandled exception rather than a `VmSetupError`, producing a confusing error message — though it would still fail closed. A few minor observations on test coverage and defensive coding are noted below.

**`src/aleph/vm/agent/vprogram_launch.py`** (line 119): The `except (tarfile.TarError, OSError)` handler does not catch `TypeError`. On Python 3.10/3.11 patch versions older than 3.10.12/3.11.4 (which lack the PEP 706 `filter` parameter backport), `tar.extractall(dest, filter="data")` raises `TypeError: extractall() got an unexpected keyword argument 'filter'`, which escapes this handler as an unhandled exception. It would still fail closed (caught by `except Exception` in `create_vm_execution`), but the error would not be a `VmSetupError`, producing a confusing `TypeError` instead. Consider either adding `TypeError` to the except tuple or adding a comment noting the Python version requirement for `filter=`.

**`src/aleph/vm/agent/vprogram_launch.py`** (line 113): If `shutil.rmtree(dest)` succeeds but `dest.mkdir(parents=True)` or `tar.extractall` fails, a partially-populated staging directory is left behind. On retry, `_extract_bundle` cleans it up via `rmtree` first, so this is not a correctness issue — but there is no test covering the extraction-failure cleanup path (e.g., a tarball that passes sha256 but is corrupted and fails to extract).

**`src/aleph/vm/agent/vprogram_launch.py`** (line 197): The `or` fallback for `settings.CONFIDENTIAL_SESSION_DIRECTORY` is defensive but redundant: `conf.py` line 706-707 always populates it in `model_post_init`. The existing `translate.py:93` uses it directly without a None check. This is harmless but slightly inconsistent with the rest of the codebase. Not a blocking issue.

**`src/aleph/vm/agent/vprogram_launch.py`** (line 167): `hardlink_to` will fail if `hash_tree_path` and `verity_path` are on different filesystems (cross-device link). The `except OSError` fallback to `shutil.copyfile` handles this correctly. However, since both paths are inside the same staging directory (`bundle_dir`), they are always on the same filesystem, so the hardlink will always succeed and the copy fallback is dead code in practice. Not a problem — just an observation.

**`tests/supervisor/test_vprogram_launch.py`** (line 295): Missing test: no test covers the case where `_extract_bundle` fails (e.g., a tarball that passes sha256 verification but is not a valid gzip/tar, or contains a member that fails the `filter='data'` check). This would verify that a partially-extracted staging directory is cleaned up on retry and that the error surfaces as `VmSetupError`.

**`tests/supervisor/test_vprogram.py`** (line 230): The launch test asserts `mock_persist.assert_awaited_once()` but does not verify the arguments — specifically that `persist_record` is called with the correct `vm_hash` and a record whose `persistent` is True. Consider asserting `mock_persist.assert_awaited_once_with(vm_hash, record)` to tighten the contract.

foxpatch-aleph (approved):
Well-structured SEV-SNP launch path with strong fail-closed security: sha256+size bundle verification, tarfile filter="data" extraction, path-traversal defense-in-depth, and consistent mirroring of the existing instance launch path in run.py. The capacity bucketing fix correctly counts V-PROGRAMs against the instance bucket to match admission. Test coverage is thorough across integrity failures, sidecar synthesis/disagreement, missing members, and all three launch-path failure modes (build, create, wait). Two minor observations: _extract_bundle leaves the staging directory behind on mid-extraction failure (cleaned on next attempt by rmtree, so transient), and the module-level logger is unused (run.py handles teardown logging). Neither is blocking.

**`src/aleph/vm/agent/vprogram_launch.py`** (line 113): If extractall fails midway (e.g. disk full), the staging directory persists with partial members. Consider adding `shutil.rmtree(dest, ignore_errors=True)` in the except block before re-raising, so no stale partial content survives a failed extraction. The next attempt's top-of-function rmtree would handle it anyway, but explicit cleanup here is more defensive.

**`src/aleph/vm/agent/vprogram_launch.py`** (line 56): The module-level `logger` is defined but never used in this module. Either remove it or use it for debug logging in build_vprogram_spec (e.g. logging the fetched manifest ref or staging path).

foxpatch-aleph (approved):
Well-structured SEV-SNP launch path with strong fail-closed security (sha256+size integrity gate, PEP 706 tar filter, path traversal defense-in-depth, roothash sidecar validation). The run.py V-PROGRAM branch correctly mirrors the instance path's record-early/teardown-on-failure pattern, and the capacity bucketing change is a necessary consistency fix. Tests are thorough across manifest parsing, integrity failures, spec construction, sidecar handling, and lifecycle error paths. The staging directory is not cleaned up on VM descheduling (only on re-extraction), which is a disk leak worth addressing in a follow-up but not blocking. The code is correct, secure, and well-documented.

**`src/aleph/vm/agent/vprogram_launch.py`** (line 63): The staging directory at EXECUTION_ROOT/vprogram/<vm_hash> is never cleaned up when a V-PROGRAM is descheduled (views/__init__.py calls supervisor.delete_vm + registry.forget but not rmtree on the staging dir). _extract_bundle only cleans up on re-extraction. This is a disk leak over time for churned V-PROGRAMs. Consider adding cleanup in the stop/deschedule path as a follow-up.

**`tests/supervisor/test_vprogram_launch.py`** (line 197): test_bundle_size_mismatch_fails_closed does not assert that the staging directory was not created, unlike the sha256 mismatch test (line 185). The size check runs before extraction so it's safe, but adding the assertion would make the fail-closed guarantee symmetric and explicit.

foxpatch-aleph (approved):
Well-structured SEV-SNP launch path for V-PROGRAMs that correctly mirrors the existing confidential instance path. The integrity model is sound: sha256+size verification before extraction, tarfile `filter="data"` for safe extraction, strict manifest parsing with path traversal validation, and fail-closed behavior on every error. The capacity bucketing fix (V-PROGRAMs counted in the instance bucket) is correct and well-tested. The verity sidecar logic handles present/absent/disagreement cases properly. Tests are thorough (9 launch tests + 3 run.py integration tests covering launch, wait-failure teardown, and build-failure cleanup). The identified issues are minor: staging directory cleanup on deschedule is missing (disk leak, mitigated by re-extraction on re-allocation), no test for the capacity-check-failure path in run.py, and the sha256 comparison is not constant-time. None of these are blocking.

**`src/aleph/vm/agent/vprogram_launch.py`** (line 98): Consider using `hmac.compare_digest(actual_sha256, manifest.bundle.sha256)` instead of `!=` for the sha256 comparison. While the timing attack surface is limited here (the manifest is network-published and the bundle is content-addressed), using constant-time comparison is defense-in-depth best practice for hash verification.

**`src/aleph/vm/agent/vprogram_launch.py`** (line 191): Staging directory (`EXECUTION_ROOT/vprogram/<vm_hash>`) is not cleaned up when a V-PROGRAM is descheduled (the `delete_vm` + `registry.forget` + `delete_records_for_vm` path in `views/__init__.py` has no staging cleanup). Re-allocation handles this via `_extract_bundle`'s `shutil.rmtree`, but a permanently descheduled V-PROGRAM leaks the extracted bundle on disk. Consider adding cleanup in the stop/deschedule path or documenting that the daemon handles it.

**`src/aleph/vm/agent/run.py`** (line 361): If `capacity.check_capacity` raises after `build_vprogram_spec` has already extracted the bundle, the staging directory remains on disk (the `except Exception` at line 369 only calls `registry.forget`). This is a minor disk leak on the failure path — the extracted bundle is left behind. Not blocking since re-allocation cleans it up, but worth noting.

**`tests/supervisor/test_vprogram.py`** (line 285): Missing test for the case where `capacity.check_capacity` raises `InsufficientResourcesError` in the V-PROGRAM branch (line 361-367 of run.py). The existing build-failure test covers `build_vprogram_spec` failure, but the capacity check runs after the bundle is staged and before `create_vm` — a distinct failure path that should verify `registry.forget` is called and `supervisor.create_vm` is never reached.

**`src/aleph/vm/agent/run.py`** (line 373): When `_wait_until_running` raises `RuntimeError` (FAILED/STOPPED status), it propagates through `create_vm_execution_or_raise_http_error` as a generic `HTTPInternalServerError(reason="Unhandled error during initialisation")`. Other failure modes (VmSetupError, InsufficientResourcesError) get specific HTTP reasons. Consider wrapping the wait-failure in a `VmSetupError` or adding a dedicated catch in `create_vm_execution_or_raise_http_error` to surface a more specific error like "VM failed to start".

foxpatch-aleph (approved):
Well-structured PR that replaces the V-PROGRAM "not implemented" seam with a complete SEV-SNP launch path. The code follows the existing instance launch pattern (record early, build spec, capacity check after download, create, wait, persist, teardown on failure) with correct fail-closed behavior at every integrity gate. Security is solid: tarball extraction uses PEP 706's `filter="data"`, member paths are validated against traversal both in the manifest model and at resolution time, sha256 comparison is constant-time, and the roothash sidecar disagreement check prevents booting a mismeasured VM. The capacity bucketing fix (counting V-PROGRAMs in the instance bucket) is consistent with the `is_instance=True` admission in `run.py`. Test coverage is thorough across all failure paths. The only minor gap is a missing test for the verity-sidecar-absent hardlink/copy fallback in `_ensure_verity_sidecars`, but the code logic is straightforward and correct.

**`src/aleph/vm/agent/vprogram_launch.py`** (line 172): Minor test gap: the hardlink/copy fallback for a missing `<rootfs>.verity` sidecar (lines 172-178) is not covered by any test. All tests either ship `rootfs.ext4.verity` in the bundle (so the `if not verity_path.is_file()` branch is skipped) or remove only the roothash sidecar. A test where the verity sidecar is absent while `platform_hash_tree` points to a differently-named member would exercise this path.

**`src/aleph/vm/agent/run.py`** (line 828): For V-PROGRAMs, `create_vm_execution` already calls `_wait_until_running` internally (line 373), so `start_persistent_vm` will call it a second time here. This is a harmless no-op (the VM is already RUNNING so `get_vm` returns immediately on the first poll), and the comment at line 825 acknowledges this pattern. Not worth changing, just noting the redundancy.

**`tests/supervisor/test_vprogram.py`** (line 234): Style nit: `_failed_vm_info` is defined between two test functions (after `test_create_vm_execution_vprogram_launches_snp` and before `test_create_vm_execution_vprogram_wait_failure_tears_down`). It would be cleaner to place it next to the `_running_vm_info` helper at line 85.

PR #1072, opened 2026-08-05
feat(daemon): conditional workload_roothash in the SNP measured cmdline

odesenfans:
Part of the V-PROGRAM workload mechanism (retires the `aleph.builtin/1` busybox placeholder; fib-service now runs as a real measured user workload).

**This PR (Rust, daemon):** `snp_config_slice` appends ` workload_roothash={hex}` to the measured kernel cmdline **iff** a `{rootfs}.workload_roothash` sidecar sits next to the rootfs disk. Bare-hex only (4 KiB cap; `RpcError::InvalidBackend` on non-hex). No proto change: the daemon derives the cmdline from sidecar files, so the frozen supervisor proto stays untouched. Workload-less VMs are byte-identical to before.

Base: `dev` (the SNP daemon code is already on dev). Independent of #1050/#1052.

Stack (V-PROGRAM workload mechanism): this + #1050 (manifest tooling) + #1052 (scheduler) + #1071 (launch) + the launch-workload / nix / measurement PRs.

### DIFF
--- a/rust/crates/supervisor-daemon/src/lifecycle.rs
+++ b/rust/crates/supervisor-daemon/src/lifecycle.rs
@@ -4447,6 +4447,30 @@ mod tests {
     }
 
     #[test]
+    fn snp_config_slice_rejects_an_oversized_workload_roothash_sidecar() {
+        // Like the platform roothash, a workload sidecar larger than the read
+        // cap fails closed rather than loading unbounded into RAM. The content
+        // is all-hex, so only the size guard can reject it.
+        let harness = harness();
+        let state = &harness.state;
+        let root = state.host.settings.execution_root.clone();
+        let firmware = root.join("OVMF.fd");
+        std::fs::write(&firmware, b"ovmf").unwrap();
+
+        let vm_id = hash('n');
+        let spec = snp_spec(&vm_id, &root, &firmware.to_string_lossy());
+        let rootfs = root.join(format!("{vm_id}-rootfs.ext4"));
+        let oversized = "a".repeat(MAX_ROOTHASH_SIDECAR_BYTES as usize + 1);
+        std::fs::write(format!("{}.workload_roothash", rootfs.display()), oversized).unwrap();
+        match snp_config_slice(state, &spec) {
+            Err(RpcError::InvalidBackend(_)) => {}
+            other => panic!(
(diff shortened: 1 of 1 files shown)


foxpatch-aleph (approved):
The implementation correctly mirrors the existing platform roothash sidecar pattern with proper hex validation, size capping, and fail-closed error handling. The absent-sidecar path preserves byte-identical cmdline for workload-less VMs. Security is solid — kernel parameter injection is prevented by the hex-only check, and the size cap prevents unbounded RAM consumption. Tests cover the happy path, regression guard, and non-hex rejection. The only minor gap is a missing oversized-sidecar test for the workload path (which exists for the platform roothash), but the code path is identical and this is non-blocking.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 2160): Minor test gap: the platform roothash has a dedicated oversized-sidecar test (`snp_config_slice_rejects_an_oversized_roothash_sidecar` at line 4349), but the workload roothash path — which uses the same `MAX_ROOTHASH_SIDECAR_BYTES` size guard — does not. Adding a parallel `snp_config_slice_rejects_an_oversized_workload_roothash_sidecar` test would give full parity coverage for this new code path.

foxpatch-aleph (approved):
The PR cleanly mirrors the existing dm-verity roothash sidecar pattern to conditionally append a workload_roothash to the SNP measured kernel cmdline. The implementation is correct: the Take-based size cap, hex validation, and trim-before-validate logic all match the proven platform roothash code. Security is solid — non-hex input (including injection attempts) fails closed as InvalidBackend, NotFound is the only silently-passing error, and workload-less VMs are byte-identical to before. Tests cover all four key paths (present/absent/non-hex/oversized). The only minor observation is test coverage asymmetry with the platform roothash (which tests three malformed variants vs one for workload), but the validation logic is identical so this is non-blocking.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 4424): Minor: the existing platform roothash test (`snp_config_slice_rejects_a_malformed_roothash_so_it_cannot_reach_append`) tests three malformed variants (injection, internally-spaced, empty-but-present). This workload test only covers one non-hex case. Consider testing the empty-string and internally-spaced (e.g. `"dead beef"`) variants too, for symmetry with the platform roothash test. Non-blocking since the validation logic is identical.

PR #1073, opened 2026-08-05
feat(vprogram): attach the measured workload volume + roothash sidecar

odesenfans:
Part of the V-PROGRAM workload mechanism. **Depends on #1071** (launch path) and #1050 (manifest import); CI is red until #1050/#1052 land on dev.

**This PR (Python, agent launch):** when `content.workload` is present, download `content.workload.ref` (data) + `.hash_tree` and attach them as EXTRA disks (→ `/dev/vdc`, `/dev/vdd`), and write a `{rootfs}.workload_roothash` sidecar from `content.workload.roothash` so the daemon can bind it into the measured cmdline.

The platform hash tree is **not** attached here: the daemon force-inserts `{rootfs}.verity` as the first SNP host volume (`/dev/vdb`), so attaching it again would shift the workload disks to vdd/vde and break the guest's vdc/vdd workload-verity assumption. (This was the disk-shift bug caught on hardware.) Fail closed on a non-bare-hex roothash. Multi-volume `content.volumes` deferred.

### DIFF
--- a/src/aleph/vm/agent/vprogram_launch.py
+++ b/src/aleph/vm/agent/vprogram_launch.py
@@ -25,6 +25,7 @@ from __future__ import annotations
 import hashlib
 import hmac
 import logging
+import re
 import shutil
 import tarfile
 from pathlib import Path
@@ -57,6 +58,11 @@ if TYPE_CHECKING:
 logger = logging.getLogger(__name__)
 
 _SHA256_CHUNK_SIZE = 1024 * 1024
+
+# The workload roothash sidecar content is trusted verbatim by the daemon
+# (snp_config_slice appends it to the measured cmdline unquoted): bare hex
+# only, no length pinned here since the daemon does not pin one either.
+HEX64 = re.compile(r"\A[0-9a-fA-F]+\Z")
 
 
 def vprogram_staging_dir(vm_hash: ItemHash) -> Path:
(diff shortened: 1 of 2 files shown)


foxpatch-aleph (commented):
Solid implementation of the V-PROGRAM workload attachment mechanism. The sidecar convention correctly mirrors the existing platform roothash pattern, both layers fail closed on non-hex input preventing cmdline injection, and the disk-order rationale (removing the platform hash tree from spec disks to avoid the disk-shift bug) is well-documented. The code is correct and secure. I'm leaving a COMMENT rather than APPROVE due to a few minor nits: a misleading variable name (HEX64 doesn't enforce 64 chars), an inconsistent newline in the sidecar write, a test assertion that checks format on spec.disks[2:] instead of spec.disks[1:], and two small test coverage gaps (no oversized workload sidecar test on the Rust side, no None-workload path test on the Python side). None of these are blocking.

**`src/aleph/vm/agent/vprogram_launch.py`** (line 64): The name `HEX64` implies exactly 64 hex characters, but the regex `\A[0-9a-fA-F]+\Z` accepts any length hex string (e.g. "a" or "deadbeef"). The comment explains the intent, but the name is misleading. Consider `BARE_HEX` or `HEX_RE` to match what it actually validates.

**`src/aleph/vm/agent/vprogram_launch.py`** (line 226): The workload roothash sidecar is written without a trailing newline (`write_text(wl_roothash)`), while the platform roothash sidecar at line 166 is written with one (`write_text(platform_roothash + "\n")`). The daemon trims both so this is functionally harmless, but the inconsistency could confuse future readers.

**`tests/supervisor/test_vprogram_launch.py`** (line 356): This assertion checks `spec.disks[2:]` (only the hash tree disk) for RAW format, but line 355 checks `spec.disks[1:]` (both workload disks) for readonly. The format check should likely also be `spec.disks[1:]` to cover the workload data disk. The data disk's format is still verified indirectly by `test_build_vprogram_spec` line 253, but this test's assertion appears incomplete.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 2160): There's no Rust test for an oversized workload roothash sidecar, even though the size-check code path exists here (mirroring the platform roothash's `snp_config_slice_rejects_an_oversized_roothash_sidecar` test at line 4349). Consider adding one for parity.

**`src/aleph/vm/agent/vprogram_launch.py`** (line 217): No test exercises the `content.workload is None` branch. The fixture message always carries a workload block, and the test comment says the field is required on `VerifiableProgramContent`, which would make this check dead code. If the field is truly required, the `is not None` guard is harmless but unnecessary; if it can be absent, a test should cover the no-workload path.

foxpatch-aleph (approved):
The PR correctly implements workload volume attachment and the roothash sidecar mechanism. Disk ordering is sound — the platform hash tree is removed from the spec (daemon force-inserts as /dev/vdb) and workload data/hash tree are appended as EXTRA disks (landing at /dev/vdc, /dev/vdd). The HEX64 regex provides defense-in-depth against cmdline injection, complementing the model's stricter pattern and the daemon's own hex validation. Comments are thorough and explain the load-bearing disk ordering rationale. Test coverage is solid with happy-path and fail-closed scenarios. One minor test bug (format assertion slice off-by-one) and a misleadingly named constant are non-blocking nits.

**`tests/supervisor/test_vprogram_launch.py`** (line 384): Off-by-one in the format assertion: `spec.disks[2:]` only checks the hash tree disk (index 2), not the workload data disk (index 1). The `readonly` assertion on the line above correctly uses `spec.disks[1:]`. This should be `spec.disks[1:]` to verify both workload disks have `DiskFormat.RAW`. The format of disk index 1 is still covered by the main `test_build_vprogram_spec` test (line 259 checks all disks), so this is a coverage gap only in the dedicated workload test, not a missed bug in the production code.

**`src/aleph/vm/agent/vprogram_launch.py`** (line 65): The name `HEX64` implies a 64-character hex string, but the regex `\A[0-9a-fA-F]+\Z` does not pin the length to 64 — it accepts any length of hex characters (and allows uppercase). The comment explains this is intentional ("no length pinned here since the daemon does not pin one either"), but the constant name is misleading. Consider renaming to `BARE_HEX` or adding a comment clarifying the name refers to the expected SHA-256 hex length, not an enforced constraint.

foxpatch-aleph (approved):
Well-structured PR that correctly attaches the workload data and hash tree as EXTRA disks (vdc/vdd) while explicitly excluding the platform hash tree from the spec (the daemon force-inserts it as vdb, avoiding the disk-shift bug). The roothash re-check with VERITY_ROOTHASH_PATTERN is good defense-in-depth against model_copy/model_construct bypassing schema validation, and the fail-closed behavior is consistently tested. Comments are thorough and explain the load-bearing disk ordering. Test coverage is comprehensive: happy path verifies disk order/roles/sidecar content, and the parametrized fail-closed test covers non-hex, too-short, and uppercase roothash cases. The only minor observation is that content.volumes is silently ignored with no log, but this is explicitly deferred per the PR description.

**`src/aleph/vm/agent/vprogram_launch.py`** (line 231): If get_existing_file raises for content.workload.ref or content.workload.hash_tree (e.g. download failure), the bundle staging directory (bundle_dir) is left behind on disk. This is consistent with the pre-existing behavior for the bundle ref download at line 193, but worth noting as a potential cleanup gap if build_vprogram_spec is ever retried for the same vm_hash.

**`src/aleph/vm/agent/vprogram_launch.py`** (line 235): The sidecar is written with write_text (non-atomic). If the process crashes mid-write, a partial sidecar could be read by the daemon. Given the roothash is only 64 chars + newline, the write is effectively instantaneous on any real filesystem, and the daemon validates the content is bare hex — so a partial write of valid hex is extremely unlikely but theoretically possible. Not a blocker given the defense-in-depth validation on both sides.

PR #1074, opened 2026-08-05
feat(nix): port fib-service (static-musl actix-web demo workload)

odesenfans:
Part of the V-PROGRAM workload mechanism (retires the `aleph.builtin/1` busybox placeholder). Ported from the aleph-cvm donor.

**This PR (Nix):** a static-musl actix-web binary built with crane. `GET /fib/{n}` returns JSON `{"n":10,"result":55}`. It becomes `/sbin/init` inside the workload volume (next PR). Pure Nix; base `dev`, no #1050/#1052 dependency.

Follow-up (tracked): route the real image re-pin through #1050's `vprogram_bundle.py` instead of hand-tarring.


foxpatch-aleph (approved):
A clean, minimal demo workload: a static-musl actix-web binary exposing a saturating u64 Fibonacci endpoint and a health check. The fibonacci implementation is correct (base cases verified, saturating_add prevents overflow panics), the Nix derivation correctly mirrors the attest-agent/attest-cli pattern while omitting unneeded openssl env, and the Cargo.lock is committed for reproducibility. The only notable points are non-blocking: the service binds 127.0.0.1 (fine if the guest only needs loopback reachability, but worth confirming against the planned networking model) and doCheck=false means the in-source unit tests are never run by Nix (consistent with sibling crates, though fibonacci is pure and could be host-tested).

**`nix/fib-service/src/main.rs`** (line 44): Binding to 127.0.0.1 means the service is only reachable from inside the guest VM. If the host or any external caller needs to reach /fib/{n}, this should be 0.0.0.0. Confirm this matches the intended guest networking model; if so, no change needed.

**`nix/flake.nix`** (line 98): doCheck = false skips the three unit tests in main.rs. Unlike attest-agent/attest-cli (which need musl for SEV/openssl), fibonacci is a pure function with no platform deps, so the tests could run on the host build via a separate native derivation. Not blocking — just flagging that the existing test coverage is never exercised by Nix.

PR #1075, opened 2026-08-05
feat(nix): fib-service workload volume + dm-verity (reproducible roothash)

odesenfans:
Part of the V-PROGRAM workload mechanism. **Depends on #1074** (fib-service).

**This PR (Nix):** builds `workload.ext4` (fib binary as `/sbin/init`) plus its dm-verity hash tree, fully reproducible (fixed salt/UUID/hash_seed, `SOURCE_DATE_EPOCH=0`, `^has_journal`) so the roothash is stable: `1fe2475cb4901e40f891c8801e1c28233c1f563baa8640a42648191cafbac250`. That roothash is what the daemon measures (via #1072's sidecar) and what attestation binds.

### DIFF
--- a/nix/flake.nix
+++ b/nix/flake.nix
@@ -130,6 +130,12 @@
       };
       rootfs = pkgs.callPackage ./rootfs.nix {};
 
+      # fib-service V-PROGRAM workload volume: a content-only ext4 carrying the
+      # fib-service binary as /sbin/init, delivered to the measured guest as an
+      # extra disk (see workload.nix). Distinct from `rootfs`, which is the
+      # platform image (kernel/initrd/OS chain).
+      workloadImage = pkgs.callPackage ./workload.nix { inherit fib-service; };
+
       # dm-verity hash tree + root hash for the rootfs. The root hash is baked
       # into the kernel cmdline, binding rootfs integrity into the SEV-SNP
       # measurement.
@@ -155,6 +161,36 @@
           | grep "Root hash:" \
           | awk '{print $NF}' \
           | tr -d '\n' > $out/roothash
+      '';
+
+      # dm-verity hash tree + root hash for the fib-service workload volume.
+      # Same mechanism as `verity` above (identical fixed salt/uuid), applied to
+      # workloadImage instead of rootfs, so the workload_roothash is reproducible
+      # too and can be baked into the guest config the same way the platform
+      # rootfs roothash is.
+      workloadVerity = pkgs.runCommand "workload-verity" {
+        nativeBuildInputs = [ pkgs.cryptsetup ];
+      } ''
+        mkdir -p $out
+        veritysetup format \
+          --salt=${veritySalt} \
+          --uuid=${verityUuid} \
+          ${workloadImage} \
+          $out/hashtree \
+          | tee /dev/stderr \
+          | grep "Root hash:" \
+          | awk '{print $NF}' \
+          | tr -d '\n' > $out/roothash
+      '';
+
+      # Convenience: the workload volume + its dm-verity sidecars, staged with
+      # the .verity/.roothash suffix convention the launch path and daemon
+      # expect for extra-disk sidecars.
+      workload = pkgs.runCommand "aleph-vm-workload" {} ''
+        mkdir -p $out
+        cp ${workloadImage} $out/workload.ext4
+        cp ${workloadVerity}/hashtree $out/workload.ext4.verity
+        cp ${workloadVerity}/roothash $out/workload.ext4.roothash
       '';
 
       # Parameterized SEV-SNP launch measurement builder.
@@ -206,6 +242,9 @@
           initrd
           rootfs
           verity
+          workloadImage
+          workloadVerity
+          workload
           measurement
           image;
         default = image;
--- a/nix/flake.nix
+++ b/nix/flake.nix
@@ -188,7 +188,7 @@
       # expect for extra-disk sidecars.
       workload = pkgs.runCommand "aleph-vm-workload" {} ''
         mkdir -p $out
-        cp ${workloadImage} $out/workload.ext4
+        ln -s ${workloadImage} $out/workload.ext4
         cp ${workloadVerity}/hashtree $out/workload.ext4.verity
         cp ${workloadVerity}/roothash $out/workload.ext4.roothash
       '';
--- a/nix/workload.nix
+++ b/nix/workload.nix
@@ -1,0 +1,43 @@
+{ pkgs, fib-service, ... }:
+
+# Reproducible ext4 volume for the fib-service V-PROGRAM workload (Task 6).
+#
+# Content-only volume: the platform rootfs (rootfs.nix) already supplies the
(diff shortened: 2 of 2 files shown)


foxpatch-aleph (approved):
This PR cleanly adds a reproducible ext4 workload volume (fib-service as /sbin/init) and its dm-verity hash tree, closely mirroring the established rootfs.nix/verity patterns. The determinism levers (fixed UUID, hash_seed, SOURCE_DATE_EPOCH, ^has_journal, non-lazy init) are correctly replicated, and the workloadVerity derivation correctly reuses the shared veritySalt/verityUuid constants. The workload convenience derivation stages artifacts with the expected .verity/.roothash suffix convention. No correctness or security issues were found. The comments are thorough and explain the rationale well. Minor nits below are non-blocking.

**`nix/workload.nix`** (line 37): Minor: the comment says "non-lazy inode-table/journal init" but with `^has_journal` there is no journal to lazily initialize, so `lazy_journal_init=0` is a no-op here. The comment is inherited from rootfs.nix and is harmless, but it could be trimmed to just mention `lazy_itable_init=0` for accuracy. Not blocking since it matches the existing rootfs.nix pattern.

**`nix/flake.nix`** (line 221): Observation: the `image` convenience derivation bundles the platform artifacts (kernel, initrd, rootfs, verity sidecars, OVMF, measurement) but does NOT include the workload volume or its verity sidecars. Consumers who want a complete launchable set must also pull the separate `workload` package. This may be intentional (workload is an optional extra disk), but a brief comment in `image` noting that the workload is available as a separate `workload` output would improve discoverability.

foxpatch-aleph (approved):
This PR correctly adds a reproducible ext4 workload volume (carrying fib-service as /sbin/init) and its dm-verity hash tree, mirroring the existing rootfs.nix/verity patterns. The reproducibility levers (fixed UUID, hash_seed, SOURCE_DATE_EPOCH=0, ^has_journal, lazy init disabled) are consistently applied, and the same fixed salt/UUID are reused for the workload's veritysetup, ensuring a stable roothash. The fib-service binary is built with static musl + crt-static, confirming the kernel can exec it directly as /sbin/init without a dynamic linker. No correctness bugs or security issues found. The comments are thorough and explain the rationale well.

**`nix/flake.nix`** (line 221): The `image` convenience output's comment says "all measured-image artifacts in one directory" but it does not include workload artifacts (workload.ext4 / .verity / .roothash). Since the workload roothash is also part of the attestation measurement, consider updating the comment to clarify it covers the platform image only, or add the workload sidecars to `image` if it's meant to be comprehensive.

**`nix/workload.nix`** (line 40): `lazy_journal_init=0` is a no-op when `^has_journal` is set (there is no journal to lazily init). This is inherited from rootfs.nix for consistency, which is fine — just flagging that the comment could note it's kept for mirror-symmetry rather than functional necessity.

**`nix/flake.nix`** (line 189): The `workload` output uses `cp` for workload.ext4, which copies the entire image into a new store path. Consider `ln -s ${workloadImage} $out/workload.ext4` instead (as `image` does for rootfs at line 225) to avoid duplicating the potentially large binary in the Nix store. Only use `cp` if the consumer specifically requires a real file rather than a symlink.

**`nix/flake.nix`** (line 171): No test asserts that the workload roothash is reproducible or matches the claimed value `1fe2475cb4901e40f891c8801e1c28233c1f563baa8640a42648191cafbac250`. Given that the entire attestation security model depends on this hash being stable, a simple assertion derivation (or a CI check comparing two builds' roothashes) would add significant value. This same gap exists for the rootfs verity, so it's a pre-existing pattern — but this PR is a good opportunity to start closing it.

foxpatch-aleph (approved):
The PR correctly and consistently extends the existing dm-verity reproducibility pattern from rootfs.nix to a new fib-service workload volume. workload.nix mirrors rootfs.nix's determinism levers (fixed UUID, non-zero hash_seed, SOURCE_DATE_EPOCH=0, ^has_journal, non-lazy init), and the workloadVerity derivation in flake.nix reuses the same veritySalt/verityUuid and parsing pipeline as the existing verity derivation. The workload convenience package correctly stages the three-artifact sidecar convention expected by the downstream launch path (vprogram_launch.py) and its tests. No correctness, security, or testing gaps were found. Two minor non-blocking observations: the important non-zero hash_seed rationale from rootfs.nix is not duplicated in workload.nix, and the workload package's exclusion from the image output (architecturally correct) is undocumented.

**`nix/workload.nix`** (line 39): Consider adding the non-zero hash_seed rationale that rootfs.nix has (lines 73-77): mke2fs treats an all-zero hash_seed as "unset" and generates a random one, making the build nondeterministic. The value used here (a1e5c0de-...) is correctly non-zero, but a future maintainer changing it to all-zeros would silently break reproducibility. A one-line comment would prevent that footgun.

**`nix/flake.nix`** (line 189): The `workload` package is deliberately excluded from the `image` output, which is architecturally correct (the workload is user-delivered content, not platform infrastructure). A brief comment noting this exclusion — e.g. "not in `image`: the workload is delivered as user content via content.workload, not as part of the measured platform image" — would help future readers understand the split.

foxpatch-aleph (approved):
This PR cleanly adds a reproducible ext4 workload volume for the fib-service V-PROGRAM binary, mirroring the established rootfs.nix determinism patterns (fixed UUID/hash_seed, SOURCE_DATE_EPOCH, ^has_journal). The dm-verity hash tree and roothash derivations correctly reuse the same fixed salt/UUID constants as the platform rootfs verity. The convenience `workload` package stages artifacts with the expected sidecar suffixes. The workload is correctly kept separate from the platform `image` bundle, matching the V-PROGRAM design where the workload volume is delivered as extra disks via the message content, not the runtime bundle. No correctness or security issues found. Minor observations include redundant `lazy_journal_init=0` when `^has_journal` is already set (inherited from rootfs.nix, harmless) and no automated test asserting the roothash is stable across rebuilds.

**`nix/workload.nix`** (line 43): The `lazy_journal_init=0` option is a no-op when `^has_journal` is already set (there is no journal to lazily initialize). This is inherited from rootfs.nix where it's similarly harmless. Consider dropping it here since this is a new file and the comment on lines 35-37 already explains `^has_journal` makes the journal irrelevant — keeping `lazy_journal_init=0` muddies that explanation. Non-blocking.

**`nix/workload.nix`** (line 25): The `chmod +x` is likely redundant: `cp` preserves the source file's mode, and the crane-built `fib-service` binary is already executable. `mkfs.ext4 -d` also preserves file modes from the staging directory. This is a reasonable defensive practice, but if you want to be precise, it could be removed. Non-blocking.

**`nix/flake.nix`** (line 189): Consider adding an assertion (or a Nix check phase) that the built `workloadVerity` roothash matches the expected value `1fe2475cb4901e40f891c8801e1c28233c1f563baa8640a42648191cafbac250`. A silent roothash drift (e.g. from a nixpkgs e2fsprogs update changing mkfs defaults) would break attestation at runtime with no build-time signal. This would also serve as the reproducibility test mentioned in the review instructions. Non-blocking suggestion.

PR #1076, opened 2026-08-05
feat(nix): guest init mounts and execs the measured workload volume

odesenfans:
Part of the V-PROGRAM workload mechanism. **Depends on #1075** (workload volume).

**This PR (Nix):** the guest `init.sh` parses `workload_roothash=` from the measured cmdline; when present it `veritysetup open`s the workload (vdc data / vdd hash tree) against the measured roothash, mounts it, and execs `/sbin/init` from it. The workload-less boot path is unchanged. New platform measurement reflects the init change: `d86657015ea1edd9346c76faa636c03d1ebe6341928a3ac9ee82286b9815741698278551edea97f1baa661648725b197`.

### DIFF
--- a/nix/flake.nix
+++ b/nix/flake.nix
@@ -7,9 +7,13 @@
   #   aleph-attest-agent static-musl binary and init.sh) + a minimal
   #   dm-verity-protected rootfs + a precomputed, reproducible sev-snp-measure
   #   launch measurement.
-  # Excluded from the donor: compose-rootfs / compose-demo, the encrypted-rootfs
-  # (LUKS) mode, and the fib-service demo app (replaced by a trivial busybox
-  # httpd placeholder workload). See rust-port-divergences.
+  # Excluded from the donor: compose-rootfs / compose-demo and the
+  # encrypted-rootfs (LUKS) mode. The fib-service demo app is NOT excluded
+  # here: it now exists as the V-PROGRAM workload (see workload.nix), baked
+  # into its own measured dm-verity volume that the guest init mounts and
+  # execs when a workload_roothash is present on the cmdline. The trivial
+  # busybox httpd remains the platform rootfs's baked /sbin/init, used as the
+  # no-workload fallback. See rust-port-divergences.
 
   inputs = {
     nixpkgs.url = "github:NixOS/nixpkgs/nixos-24.11";
@@ -194,12 +198,26 @@
       '';
 
       # Parameterized SEV-SNP launch measurement builder.
-      # vcpus:    number of vCPUs (affects the launch measurement)
-      # vcpuType: QEMU CPU model ("EPYC-v4" for Genoa, "EPYC-v3" for Milan)
+      # vcpus:            number of vCPUs (affects the launch measurement)
+      # vcpuType:         QEMU CPU model ("EPYC-v4" for Genoa, "EPYC-v3" for Milan)
+      # workloadRoothash: when null (default), the cmdline is the
+      #   platform-only form `...roothash=<platform>` (workload-less parity,
+      #   what test_vm_snp and the baked `measurement` below expect). When
+      #   set to a dm-verity root hash, the cmdline is extended to the
+      #   workload form `...roothash=<platform> workload_roothash=<hex>`,
+      #   matching byte-for-byte what the daemon emits when a V-PROGRAM
+      #   workload is attached (lifecycle.rs) and the CMDLINE_TEMPLATE_EXEC_V1
+      #   manifest template (src/aleph/vm/vprogram/bundle.py). Per-workload
+      #   measurements are computed by passing this argument; they are never
+      #   baked into the platform bundle's own measurement.hex.
       # The measurement is a function of (OVMF + kernel + initrd + cmdline +
       # vCPU count + CPU type), so each configuration needs its own value.
-      measurementFor = { vcpus ? 2, vcpuType ? "EPYC-v4" }: let
-        kernelCmdline = "console=ttyS0 root=/dev/mapper/verity-root ro roothash=${builtins.readFile "${verity}/roothash"}";
+      measurementFor = { vcpus ? 2, vcpuType ? "EPYC-v4", workloadRoothash ? null }: let
+        platformRoothash = builtins.readFile "${verity}/roothash";
+        kernelCmdline =
+          if workloadRoothash == null
+          then "console=ttyS0 root=/dev/mapper/verity-root ro roothash=${platformRoothash}"
+          else "console=ttyS0 root=/dev/mapper/verity-root ro roothash=${platformRoothash} workload_roothash=${workloadRoothash}";
       in pkgs.runCommand "sev-snp-measurement-${toString vcpus}vcpus-${vcpuType}" {
         nativeBuildInputs = [ sev-snp-measure ];
       } ''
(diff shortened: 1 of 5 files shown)


foxpatch-aleph (changes requested):
The PR correctly implements the guest-side V-PROGRAM workload mounting and exec logic, and the sed regex cross-matching analysis is sound. However, there is a critical bug: `prepare_chroot` only sets up bind mounts (/proc, /sys, /dev, /tmp/secrets, /etc/resolv.conf) into `/mnt/root`, but the workload is chrooted into `/mnt/workload` — a content-only volume with no /proc, /sys, /dev, or /tmp. The chrooted workload will lack essential filesystems, breaking tokio/actix-web (no /proc for CPU discovery, no /dev/null) and rendering the secret delivery mechanism non-functional (no /tmp/secrets bind mount). Additionally, a missing /sbin/init in the workload volume is treated as a WARNING rather than FATAL, leaving the VM in a degraded state with no upstream.

**`nix/init.sh`** (line 250): Critical bug: `prepare_chroot` only sets up bind mounts into `/mnt/root` (/proc, /sys, /dev, /tmp/secrets, /etc/resolv.conf), but the workload is chrooted into `/mnt/workload` at line 262. The workload volume (per workload.nix) is content-only — it contains just `/sbin/init` and no /proc, /sys, /dev, /tmp, or /etc. The chrooted workload will have no access to pseudo-filesystems, no secret delivery (/tmp/secrets is not bind-mounted into /mnt/workload), and no DNS. `prepare_chroot` needs to either be parameterized to also set up `/mnt/workload`, or a parallel `prepare_workload_chroot` function needs to be called when `workload_roothash` is set.

**`nix/init.sh`** (line 264): When `workload_roothash` is set but `/mnt/workload/sbin/init` is missing, the script logs a WARNING and continues to start the attest-agent with no upstream on port 8080. Since verity verification already passed (the workload volume is authentic), a missing /sbin/init indicates a broken workload image. This should be FATAL (poweroff) rather than a warning, consistent with the fail-closed pattern used everywhere else in this script.

**`nix/init.sh`** (line 262): The workload is chrooted into `/mnt/workload` but the attest-agent at line 274 proxies to `http://127.0.0.1:8080`. This works because the chroot bind-mounts /dev (once prepare_chroot is fixed), and the loopback interface is shared. However, if the workload binds to a different port than 8080, the agent's `--upstream` argument is hardcoded. This is fine for the current fib-service demo but worth noting for future workloads.

**`nix/flake.nix`** (line 209): The comment references `CMDLINE_TEMPLATE_EXEC_V1` in `src/aleph/vm/vprogram/bundle.py`, but that file and constant do not exist in this repo. The daemon-side code (lifecycle.rs) also does not yet append `workload_roothash=` to the cmdline. Consider noting that this is forward-looking (the daemon-side implementation is in a future PR) to avoid confusion for readers tracking the cross-references.

**`nix/init.sh`** (line 56): The word-boundary analysis in this comment is correct — I verified that `\broothash=` does not match inside `workload_roothash=` because `_r` has no word boundary. Good defensive documentation.

foxpatch-aleph (changes requested):
The core logic is correct — the sed regex parsing is verified safe with busybox sed, the wait_for_dev refactor is clean, and the flake.nix measurement parameterization correctly keeps the default platform-only. However, prepare_chroot only bind-mounts /proc, /sys, /dev, /tmp/secrets, and /etc/resolv.conf into /mnt/root, but the V-PROGRAM workload runs chrooted into /mnt/workload which receives none of these. The workload volume (workload.nix) only contains /sbin/init with no directory structure for these mounts. Any real-world workload needing /proc, /dev, secrets, or DNS would fail. Additionally, a missing workload /sbin/init is only a WARNING while the attest-agent starts proxying to a dead upstream, and the flake.nix comment references a non-existent constant name.

**`nix/init.sh`** (line 250): prepare_chroot (line 250) only prepares /mnt/root — it bind-mounts /proc, /sys, /dev, /tmp/secrets, and /etc/resolv.conf into /mnt/root. But when a workload is present, the workload runs chrooted into /mnt/workload (line 262), which gets none of these. The workload volume (workload.nix) only contains /sbin/init — there are no /proc, /sys, /dev, /tmp, or /etc directories to even bind-mount onto. Any real-world workload that needs /proc, /dev, /tmp/secrets, or DNS will fail. At minimum, the workload chroot needs the same bind-mounts applied to /mnt/workload (and the workload.nix image needs the corresponding empty mount-point directories).

**`nix/init.sh`** (line 264): When workload_roothash is set but /mnt/workload/sbin/init is missing, only a WARNING is printed and the VM continues. The attest-agent then starts with --upstream http://127.0.0.1:8080 (line 274) but nothing is listening there, so every request proxied through the attestation agent will fail. This should be FATAL (poweroff) like the missing-device cases above, since the workload was explicitly requested but cannot run.

**`nix/flake.nix`** (line 209): The comment references CMDLINE_TEMPLATE_EXEC_V1 in src/aleph/vm/vprogram/bundle.py, but no such constant exists. The actual constant is CMDLINE_TEMPLATE_V1 (bundle.py line 124). This is a documentation cross-reference error.

**`nix/init.sh`** (line 88): wait_for_dev uses `n` without `local`, shadowing the outer `n` variable from line 66. This is not a bug today (the outer loop is done before wait_for_dev is first called), but it is a latent footgun. Consider using `local n=0` for safety.

foxpatch-aleph (commented):
Solid implementation of the V-PROGRAM workload mount/exec path in guest init.sh. The dm-verity setup for the workload volume mirrors the platform rootfs path correctly, the sed regex for parsing workload_roothash is unambiguous (verified with busybox sed), the prepare_chroot refactoring to a parameterized target is clean, and the disk ordering assumptions (/dev/vdc, /dev/vdd) match the daemon's host volume insertion order. The measurementFor parameterization correctly keeps the platform measurement workload-free. The rootfs.nix and workload.nix changes to pre-create mount-point directories are necessary and correct for the read-only verity mounts. The fail-closed behavior on missing workload /sbin/init is the right security posture. A few stale comment references should be cleaned up but are non-blocking.

**`nix/flake.nix`** (line 209): The comment references `CMDLINE_TEMPLATE_EXEC_V1` in bundle.py, but the actual constant is `CMDLINE_TEMPLATE_V1` (no `EXEC_` prefix). Additionally, the manifest template in bundle.py does NOT include a `{workload_roothash}` slot — `CMDLINE_TEMPLATE_V1` is still `"console=ttyS0 root=/dev/mapper/verity-root ro roothash={platform_roothash}"`. The daemon (lifecycle.rs) appends `workload_roothash=...` to the cmdline independently of the manifest template. The comment should reference `CMDLINE_TEMPLATE_V1` and clarify that the workload roothash is appended by the daemon, not by the manifest template.

**`src/aleph/vm/vprogram/bundle.py`** (line 122): Comment says "its init parses only roothash=" — this is now outdated since init.sh also parses `workload_roothash=`. Not changed by this PR, but worth updating in a follow-up to avoid confusion.

foxpatch-aleph (approved):
Well-structured PR that correctly extends the guest init.sh to parse `workload_roothash=` from the measured cmdline, verity-verify and mount the workload volume from /dev/vdc+vdd, and exec its /sbin/init instead of the platform fallback. The sed `` word-boundary disambiguation between `roothash=` and `workload_roothash=` is correct (verified empirically with busybox sed). The disk ordering (vda/vdb/vdc/vdd) is consistent across the Python agent spec, the Rust daemon, the QEMU argv builder, and init.sh. The cmdline format matches byte-for-byte between the daemon (lifecycle.rs), the flake's `measurementFor`, and `CMDLINE_TEMPLATE_EXEC_V1` in bundle.py — critical for attestation parity. The `prepare_chroot` parameterization is clean, all error paths fail closed with `poweroff -f`, and the firewall is installed before the workload starts. The rootfs.nix and workload.nix mount-point directory additions correctly mirror the existing pattern for read-only verity mounts. No correctness or security issues found. Minor observations: `workloadMeasurement` is exposed but unused in the build graph, and init.sh's parsing/mount logic has no automated test coverage (consistent with the pre-existing pattern).

**`nix/flake.nix`** (line 281): `workloadMeasurement` is exposed as a flake package but not consumed by any other code in the repo. The comment at lines 240-245 explains it's for sanity-checking the workload cmdline template end-to-end, which is a reasonable use, but it's effectively dead code in the build graph. Consider whether it adds enough value to justify the build-time cost of a second `sev-snp-measure` invocation, or whether a comment referencing how to compute it with `measurementFor` would suffice.

**`nix/init.sh`** (line 62): The `` word-boundary disambiguation between `roothash=` and `workload_roothash=` is correct and well-documented. The comment at lines 56-61 accurately explains why `roothash=` cannot match inside `workload_roothash=` (since `_` is a word character, there's no boundary between the `d` of `workload` and `_`, nor between `_` and `r`). Verified empirically with busybox sed: no cross-matching occurs in any combination of tokens. No change needed, just confirming the claim holds.

**`nix/init.sh`** (line 280): The workload `/sbin/init` is started with `&` (background) before the attest-agent at line 297. This matches the existing pattern for the platform `/sbin/init` at line 291. If the workload binary fails to start (e.g. immediate crash), the attest-agent will still start and proxy to a dead upstream. This is the same behavior as the platform path and is acceptable — the `wait` at line 300 will reap the dead child and PID 1 will exit, shutting down the VM. Just noting the consistency.

PR #1077, opened 2026-08-05
feat(vprogram): aleph.exec/1 workload contract + workload_roothash cmdline template

odesenfans:
Part of the V-PROGRAM workload mechanism. **Depends on #1050** (extends its `bundle.py` / manifest tooling).

**This PR (Python):** adds the `aleph.exec/1` workload contract (direct-exec a static entrypoint, not the donor's podman/compose stack) and the `CMDLINE_TEMPLATE_EXEC_V1` that carries `workload_roothash={workload_roothash}`, plus `EXEC_WORKLOAD = WorkloadSpec(contract="aleph.exec/1", upstream_port=8080)` and `make_manifest(exec_runtime=...)`. This is what the harness/manifest declares so the daemon's measured cmdline and the guest init agree.

### DIFF
--- a/src/aleph/vm/vprogram/bundle.py
+++ b/src/aleph/vm/vprogram/bundle.py
@@ -122,16 +122,39 @@ def build_bundle(image_dir: Path, out_dir: Path, source_epoch: int, source: Sour
 # its init parses only roothash=). Changing these is a runtime/format
 # evolution, not a CLI flag.
 CMDLINE_TEMPLATE_V1 = "console=ttyS0 root=/dev/mapper/verity-root ro roothash={platform_roothash}"
+# Exec-runtime flavor: the daemon measures a workload rootfs alongside the
+# platform rootfs and folds its dm-verity roothash into the cmdline (Task 1's
+# daemon emits exactly this string). Client-side launch-measurement
+# computation depends on byte-identity with what the daemon emits, so this
+# constant must not be reformatted independently of that emitter.
+CMDLINE_TEMPLATE_EXEC_V1 = (
+    "console=ttyS0 root=/dev/mapper/verity-root ro roothash={platform_roothash}"
+    " workload_roothash={workload_roothash}"
+)
 DEFAULT_CPU_MODELS = ["EPYC-v4"]
 DEFAULT_ATTESTATION = [
     AttestationProtocol(protocol="aleph.ra-tls", version="1", transport=AttestationTransport(type="tcp", port=8443))
 ]
 DEFAULT_WORKLOAD = WorkloadSpec(contract="aleph.builtin/1", upstream_port=8080)
+# Exec-runtime workload contract: a plain executable/command workload rather
+# than the builtin no-workload runtime.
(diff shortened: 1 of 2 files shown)


foxpatch-aleph (commented):
The PR is correct, secure, and well-documented. The CMDLINE_TEMPLATE_EXEC_V1 constant and EXEC_WORKLOAD spec are properly defined, the make_manifest function cleanly selects between builtin and exec flavors, and the cmdline_template validator already permits the workload_roothash placeholder. The new test covers the happy path. Two minor observations: (1) the CLI script scripts/vprogram_bundle.py does not expose the exec_runtime flag, so there is no way to produce an exec manifest from the command line yet — presumably deferred to a follow-up; (2) the non-aliasing test (test_make_manifest_does_not_alias_module_defaults) only covers the default (builtin) path, so a similar guard for EXEC_WORKLOAD would close a small coverage gap.

**`scripts/vprogram_bundle.py`** (line 103): The CLI `cmd_manifest` does not pass `exec_runtime=True` to `make_manifest`, so there is currently no way to produce an exec-runtime manifest from the command-line tool. If this is intentionally deferred to a follow-up PR, a brief comment here would help clarify that.

**`tests/vprogram/test_bundle.py`** (line 142): The non-aliasing test (`test_make_manifest_does_not_alias_module_defaults`) only covers the default builtin path. Consider adding an `exec_runtime=True` variant (or parameterizing the existing test) to verify that `EXEC_WORKLOAD` instances are also not aliased across calls.

foxpatch-aleph (approved):
A clean, well-scoped PR that adds the aleph.exec/1 workload contract and the CMDLINE_TEMPLATE_EXEC_V1 cmdline template with a keyword-only exec_runtime flag on make_manifest. The template is byte-identical to the Rust daemon's cmdline emitter (verified against lifecycle.rs:2176), the contract string matches the validated regex pattern, and the workload_roothash placeholder is in the allowlist. The new test covers the core assertions and passes. The only minor observation is that the anti-aliasing test and canonical-JSON round-trip test only cover the default (builtin) path, not the exec path, though both paths use the same model_copy(deep=True) mechanism so the risk is low.

**`tests/vprogram/test_bundle.py`** (line 152): Minor: the existing test_make_manifest_from_bundle_info (line 125) validates the canonical-JSON round-trip via `to_canonical_json()`, but this test only does `model_validate(model_dump(mode="json"))`. Consider also asserting `RuntimeManifest.model_validate_json(manifest.to_canonical_json()) == manifest` for parity, since canonical-JSON serialization is what actually gets published.

**`tests/vprogram/test_bundle.py`** (line 142): Minor: test_make_manifest_does_not_alias_module_defaults only covers the default (builtin) path. Since EXEC_WORKLOAD is also a module-level WorkloadSpec constant, consider adding an exec_runtime=True variant to confirm the exec path also avoids aliasing. Both paths use model_copy(deep=True), so the risk is low, but the test would guard against a future refactor that accidentally drops the copy.

PR #1079, opened 2026-08-06
feat(vprogram): map the attestation port to a host IPv4 port

odesenfans:
Task 1 of the V-PROGRAM CLI E2E design (docs/plans/2026-08-05-vprogram-cli-e2e-*). Reviewable 1-commit diff; already fast-forwarded into `od/vprogram-integration` (de9ed350) so the deb-builder CI picks it up.

The CRN auto-maps a running V-PROGRAM's manifest-declared `aleph.ra-tls` attestation port (:8443) to a host IPv4 port via the existing DNAT machinery, so an external client (`aleph vprogram call`) can reach the guest's attested endpoint. Manifest-driven (`build_vprogram_spec` -> `(spec, attest_port)`), idempotent on re-adopt (`reconcile_vprogram_port_forwards`), host-only DNAT so **measurement-neutral**. Surfaces in `/v2/about/executions/list` `mapped_ports`. Daemon `add_port_forward` confirmed not vm-type-gated. 26 targeted + 188 regression tests pass; mypy/ruff/isort clean.

(Base is a stable anchor at the pre-Task-1 integration tip; retarget to dev once the V-PROGRAM stack lands there.)

### DIFF
--- a/src/aleph/vm/agent/run.py
+++ b/src/aleph/vm/agent/run.py
@@ -180,9 +180,10 @@ def resolve_vprogram_port_forwards(vm_id: VmId, attest_port: int | None) -> list
 
 async def reconcile_vprogram_port_forwards(supervisor: Supervisor, vm_id: VmId, attest_port: int | None) -> None:
     """Drive the hypervisor's forwards to match the V-PROGRAM's single
-    attestation-port mapping. Reuses ``_reconcile_forwards``, so re-running
-    this on re-adoption (a fresh agent process picking the VM back up) is a
-    no-op once the mapping already exists.
+    attestation-port mapping. Reuses ``_reconcile_forwards``, so it is
+    idempotent by construction: a future re-adoption path (a fresh agent
+    process picking the VM back up) can call it safely, though today only
+    the create path does.
     """
     await _reconcile_forwards(supervisor, vm_id, resolve_vprogram_port_forwards(vm_id, attest_port))
 
@@ -413,6 +414,12 @@ async def create_vm_execution(
             # the guest's RA-TLS attestation endpoint; measurement-neutral
             # (no guest image/cmdline change) and idempotent via the same
             # reconcile machinery the instance path uses.
+            #
+            # A reconcile failure lands in the teardown branch below on
+            # purpose, mirroring the instance path: a V-PROGRAM whose
+            # attestation endpoint cannot be mapped is unreachable for its
+            # sole consumer, so failing the create loudly (and letting the
+            # scheduler retry) beats keeping an unusable VM alive.
             await reconcile_vprogram_port_forwards(supervisor, info.vm_id, attest_port)
         except Exception:
             registry.forget(vm_hash)
--- a/src/aleph/vm/agent/run.py
+++ b/src/aleph/vm/agent/run.py
@@ -132,14 +132,14 @@ async def resolve_port_forwards(vm_id: VmId, content) -> list[PortForwardSpec]:
     return forwards
 
 
-async def reconcile_port_forwards(supervisor: Supervisor, vm_id: VmId, content) -> None:
-    """Drive the hypervisor's forwards to match the aggregate settings.
-
-    Agent policy half of the old fetch_port_redirect_config_and_setup: compute
-    the desired set, diff against what the hypervisor reports, and issue
-    add/remove calls. The hypervisor owns application and persistence.
+async def _reconcile_forwards(supervisor: Supervisor, vm_id: VmId, desired_specs: list[PortForwardSpec]) -> None:
+    """Diff `desired_specs` against what the hypervisor currently reports for
+    `vm_id` and issue add/remove calls so the two converge. The hypervisor
+    owns application and persistence; this is pure agent policy, and it is
+    idempotent: calling it again with the same desired set (e.g. on
+    re-adoption) issues no calls at all.
     """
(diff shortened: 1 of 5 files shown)


foxpatch-aleph (commented):
Solid, well-structured PR that cleanly refactors the port-forward reconcile into a shared diff primitive and adds V-PROGRAM attestation port mapping on top. The code is measurement-neutral (host-only DNAT), idempotent by construction, and well-tested with both happy-path and None-port coverage. Two non-blocking observations: (1) the docstrings claim idempotency "on re-adoption" but no re-adoption code path calls reconcile_vprogram_port_forwards yet — the idempotency is real but the re-adoption scenario is future work, so the docstrings slightly oversell the current behavior; (2) a DNAT failure tears down the entire VM (via the existing except block), which is aggressive given that attestation reachability is described as non-essential for the VM to boot and run — a transient DNAT error would kill a healthy VM. Neither blocks merging.

**`src/aleph/vm/agent/run.py`** (line 184): The docstring says \"re-running this on re-adoption (a fresh agent process picking the VM back up) is a no-op,\" but there is currently no re-adoption code path that calls this function. The `_rehydrate_vm_registry` startup hook only repopulates the registry; it does not reconcile port forwards for V-PROGRAMs (unlike instances, which get reconcile_port_forwards via aggregate-update handlers). The idempotency is real, but the re-adoption scenario is future work — consider softening the docstring to say \"idempotent by construction, so a future re-adoption path can call it safely\" rather than implying it already happens.

**`src/aleph/vm/agent/run.py`** (line 416): If `reconcile_vprogram_port_forwards` raises (e.g. transient DNAT/hypervisor error), the except block at line 417 tears down the entire VM via `delete_vm`. Given the PR description states attestation reachability \"is not required for the VM itself to boot and run,\" a DNAT failure killing a healthy RUNNING VM seems overly aggressive. Consider wrapping the reconcile in its own try/except that logs but doesn't tear down, or at minimum documenting that this is the intended behavior. (The instance path has the same pattern, so this may be a deliberate consistency choice.)

**`src/aleph/vm/agent/vprogram_launch.py`** (line 185): `_select_attestation_port` returns the first `aleph.ra-tls` entry found. The manifest schema (`attestation: list[AttestationProtocol]` with `min_length=1`) does not forbid multiple `aleph.ra-tls` entries. If that ever occurs, only the first port is mapped and the rest are silently ignored. This is likely fine for now (a single RA-TLS transport is the expected case), but worth a one-line comment noting the first-match-wins policy.

foxpatch-aleph (changes requested):
The core implementation is sound: the refactoring of _reconcile_forwards is clean, the V-PROGRAM port-forward logic is correctly idempotent, measurement-neutral (host-only DNAT), and well-documented. However, two pre-existing tests were not updated for the build_vprogram_spec signature change from CreateVmSpec to tuple[CreateVmSpec, int | None]. Both test_create_vm_execution_vprogram_wait_failure_tears_down and test_create_vm_execution_vprogram_capacity_failure_forgets_record still pass a non-tuple return_value to the mock, which will cause a TypeError on the tuple unpacking at run.py:397 instead of the expected exception, making both tests fail.

**`tests/supervisor/test_vprogram.py`** (line 267): The mock return_value was not updated for the new tuple return type. build_vprogram_spec now returns (spec, attest_port), but the mock returns a bare fake_spec. The tuple unpacking at run.py:397 will raise TypeError before the VM reaches _wait_until_running, so the test will fail with TypeError instead of the expected RuntimeError. Fix: return_value=(fake_spec, 8443)

**`tests/supervisor/test_vprogram.py`** (line 329): Same issue: the mock returns a bare MagicMock() instead of a tuple. The tuple unpacking at run.py:397 will raise TypeError before check_capacity runs, so the test will fail with TypeError instead of the expected InsufficientResourcesError. Fix: return_value=(MagicMock(), 8443)

foxpatch-aleph (approved):
Clean, well-structured PR that maps V-PROGRAM attestation ports to host IPv4 ports via the existing DNAT machinery. The refactoring of reconcile_port_forwards into a shared _reconcile_forwards with thin wrappers is well-factored, the attestation port is correctly kept orthogonal to the measured launch (measurement-neutral), and the idempotency guarantee is both documented and tested. Test coverage is solid with a FakeSupervisor that genuinely exercises the reconcile diff logic. Minor non-blocking gaps: no dedicated test for port-forward-failure-triggered teardown (covered indirectly via the shared teardown code) and no test for first-match-wins with multiple ra-tls entries.

**`src/aleph/vm/agent/vprogram_launch.py`** (line 193): The `protocol.transport.type == "tcp"` check is currently always true since `AttestationTransport.type` is `Literal["tcp"]` — the schema allows nothing else. This is fine as defensive code (future-proofing if the transport union expands), but worth noting it's a no-op today.

**`src/aleph/vm/agent/run.py`** (line 836): When a V-PROGRAM is already RUNNING and `start_persistent_vm` is called (the re-adoption path after an agent restart), port forwards are not reconciled. The PR description acknowledges this ("though today only the create path does"), and `reconcile_vprogram_port_forwards` is designed for this future use. Not a gap in this PR, just noting the already-running re-adoption case is unhandled for both instances and V-PROGRAMs.

**`tests/supervisor/test_vprogram_port_map.py`** (line 100): Consider adding a test for the case where `reconcile_vprogram_port_forwards` raises (e.g., `add_port_forward` fails) inside `create_vm_execution`, verifying the VM is torn down and the registry record is forgotten. The existing `test_create_vm_execution_vprogram_wait_failure_tears_down` covers the shared teardown code, but a port-forward-specific failure test would make the intent explicit.

PR #1080, opened 2026-08-06
feat(vprogram): IPv6 reachability for confidential guests (guest-ready; host RA sender pending)

odesenfans:
The IPv6 alternative to Task 1's host-IPv4 port-map (docs/plans/2026-08-05-vprogram-cli-e2e-design.md, deferred IPv6 path). **Guest-side is ready; end-to-end IPv6 is INERT until a host-side RA sender is added** (documented, not half-implemented).

- `nix/init.sh`: obtain IPv6 via **SLAAC** (accept_ra, bounded 2s non-fatal wait) — never a measured cmdline field, so the SNP measurement stays precomputable; added the missing ICMPv6 ND/RA/PMTU accept rule (policy drop preserved; :8443 already covered v6 via the `inet` table).
- `aleph-attest-agent`: bind `[::]:8443` (dual-stack) instead of `0.0.0.0` — was IPv4-only, making the feature inert. Verified NOT v6-only (actix-web leaves IPV6_V6ONLY at Linux dual-stack default), so **the existing IPv4 path / Task 1 port-map is unaffected**.
- daemon: `VmType::VProgram = 0x4` + fixed all v-program create/adopt call sites (create-time and persisted predicates proven identical → no address drift across restart); 2 new tests.

**NOT done (documented follow-ups):** the host per-tap RA sender (radvd/RA) — required for this to actually route. **Measurement impact:** the init.sh + attest-agent changes shift the measured image → a re-pin is required downstream before this ships. This PR deliberately stays OUT of `od/vprogram-integration` for that reason.

shellcheck clean; cargo build/test/clippy clean (325 daemon tests +2). Base is the pre-Task-1 anchor for a clean diff; retarget to dev once the V-PROGRAM stack lands.

### DIFF
--- a/nix/init.sh
+++ b/nix/init.sh
@@ -49,6 +49,32 @@ else
         # per-tap dnsmasq but never configure its IP. See udhcpc.script.
         /bin/busybox udhcpc -i "$iface" -q -t 5 -A 2 -s /bin/udhcpc.script 2>&1 || echo "init: DHCP failed on ${iface}"
     fi
+
+    # IPv6: SLAAC via Router Advertisements ONLY, never a kernel-cmdline
+    # address. The assigned /124 is per-VM (CRN /64 + vm-hash), so baking it
+    # into cmdline would make the measured SNP kernel cmdline depend on the
+    # address, breaking the publisher's precompute -- the exact reason the
+    # IPv4 path above uses DHCP/static-from-gateway instead of a hardcoded
+    # guest IP. `accept_ra=2` accepts RAs even though this host never
+    # forwards (belt and suspenders: forwarding stays off here regardless).
+    # The wait below is bounded and non-fatal: with no RA source reachable
+    # on this tap the guest simply stays IPv4-only, matching Python's
+    # ipv6_forwarding_enabled=False fallback.
+    echo 0 > "/proc/sys/net/ipv6/conf/${iface}/disable_ipv6" 2>/dev/null || true
+    echo 2 > "/proc/sys/net/ipv6/conf/${iface}/accept_ra" 2>/dev/null || true
+    n=0
+    while [ "$n" -lt 20 ]; do
+        if /bin/busybox ip addr show dev "$iface" 2>/dev/null | /bin/busybox grep -q "inet6.*scope global"; then
+            break
+        fi
+        /bin/busybox sleep 0.1
+        n=$((n + 1))
+    done
+    if /bin/busybox ip addr show dev "$iface" 2>/dev/null | /bin/busybox grep -q "inet6.*scope global"; then
+        echo "init: IPv6 SLAAC address configured on ${iface}"
+    else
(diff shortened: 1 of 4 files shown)


foxpatch-aleph (approved):
Well-structured PR that correctly adds guest-side IPv6 reachability for confidential VMs while keeping the feature inert until the host RA sender lands. The SLAAC bring-up in init.sh is properly bounded (2s non-fatal wait) and never touches the kernel cmdline (preserving SNP measurement precomputability). The firewall icmpv6 rule correctly adds ND/PMTU control-plane types to prevent IPv6 from blackholing. The attest-agent bind change from 0.0.0.0 to [::] is a strict superset (Linux dual-stack default), leaving the existing IPv4 path unaffected. The VmType::VProgram variant (0x4) is consistently wired through all four call sites (create, adopt, tap_assignment, guest_ipv4) with a clean vm_type() method that mirrors the Python logic, and the create-time vs persisted predicates are provably identical. The two new tests adequately cover the VProgram hextet in both the static math and the full create path. No correctness, security, or quality issues found.

foxpatch-aleph (approved):
Well-structured PR with correct, consistent VmType::VProgram logic across create, adopt, and lazy-derivation paths. The IPv6 SLAAC bring-up is properly bounded and non-fatal, the firewall icmpv6 additions are correctly scoped to control-plane only, and the dual-stack bind is a strict superset of the old IPv4-only bind. Documentation is thorough. Two minor non-blocking observations: the VmType predicate is duplicated in build_world_view (necessary but extractable), and there's no direct test for the SNP adoption path (indirectly covered).

**`rust/crates/supervisor-daemon/src/world.rs`** (line 572): The `qemu.snp().is_some()` check here duplicates the predicate in `VmEntry::vm_type()` (line 278). This is necessary since the entry isn't constructed yet at this point, but consider extracting a `QemuVmConfig::vm_type()` method to centralize the SNP detection logic and avoid future drift. Non-blocking.

**`rust/crates/supervisor-daemon/src/world.rs`** (line 1293): Consider adding a test that verifies the adoption path (`build_world_view`) assigns the VProgram 0x4 hextet to an SNP config — this would directly test the no-address-drift-across-restart guarantee that the PR description highlights. The create path and static math are tested, but the adoption path is only indirectly covered. Non-blocking.

odesenfans:
Superseded by #1087: SLAAC cannot deliver the daemon-allocated /124 (SLAAC requires a /64 PIO; a shared-/64 RA would let the guest self-select an address outside the ndppd range and the vm-type hextet scheme). #1087 lands the same goal statically end-to-end via the existing per-tap dnsmasq (stateful single-address DHCPv6 + RA), reusing this branch's VmType::VProgram, ICMPv6 firewall, and dual-stack agent-bind pieces. Design: docs/plans/2026-08-13-vprogram-ipv6-dhcpv6-design.md on that branch.

PR #1081, opened 2026-08-06
fix(supervisor): stop mid-boot corruption of persistent VM rootfs

aliel:
## Summary

Two allocation POSTs arriving ~40 seconds apart could destroy a fresh persistent VM's BTRFS rootfs and leave it permanently unstartable, every subsequent start attempt returns 500 \`Unhandled Error\` with the real cause (BTRFS corruption) redacted from the API. Observed on crn10.leviathan.so:

```sh
POST /control/allocations (18:35:29)  → VM a512f7cd started
POST /control/allocations (18:36:08)  → same hash
  supervisor: "unknown execution state, stopping the vm"
  firecracker: "cannot receive shutdown signal: (111, 'Connection refused')"
  systemd: SIGKILL
  btrfstune: "open ctree failed: parent transid verify failed"
```

## Root cause chain

1. First POST: \`start_persistent_vm\` starts the VM. \`times.started_at\` is set optimistically, systemd controller is \`activating\`.
2. Second POST: \`start_persistent_vm\` for the same hash. Fast paths all miss because:
   - \`is_running\` → False (systemd state != \"active\")
   - \`is_starting\` → False (\`started_at\` already set)
   - \`is_stopping\` → False
   - \`is_awaiting_confidential_init\` → False (not confidential)
3. Falls into the else \`\"unknown execution state, stopping the vm\"\` branch.
4. Firecracker API socket not yet listening → graceful shutdown fails → systemd SIGKILLs → guest BTRFS write cache lost.
5. Retry to start → \`btrfstune -m\` fails, VM stuck.

## Fix (defense in depth)

**1. Don't stop VMs whose systemd state is transient** (\`run.py\`) — check the actual systemd \`ActiveState\` before assuming the VM is dead. Only stop and recreate on definitively-terminal states (\`failed\`, \`inactive\`, \`deactivating\`). Treat \`active\`, \`activating\`, and \`unknown\` as transient and return the existing execution unchanged. Removes the trigger.

**2. Auto-recover BTRFS corruption on tune** (\`storage.py\`) — on \`btrfstune -m\` failure whose stderr matches BTRFS log-tree corruption markers (\`open ctree failed\` / \`parent transid verify failed\`), run \`btrfs rescue zero-log\` and retry once. Any residual case (host crash, OOM kill, kernel bug) now self-heals on the next start instead of blocking that VM forever. Loses at most the last few seconds of guest writes.

## Test plan
- [x] Transient systemd states (\`active\` / \`activating\` / \`unknown\`) leave the VM alone.
- [x] Terminal states (\`failed\` / \`inactive\` / \`deactivating\`) still trigger stop + recreate.
- [x] Corruption stderr markers trigger \`zero-log\` + retry.
- [x] Unrelated \`btrfstune\` failures propagate without attempting recovery.
- [x] Zero-log failure does not loop into a third \`btrfstune\` attempt.
- [ ] Deploy to a staging CRN, POST /control/allocations twice in <1s for the same fresh hash, verify the VM stays up.
- [ ] Verify recovery of an already-corrupted rootfs on the next start.

### DIFF
--- a/src/aleph/vm/orchestrator/run.py
+++ b/src/aleph/vm/orchestrator/run.py
@@ -301,9 +301,14 @@ async def start_persistent_vm(vm_hash: ItemHash, pubsub: PubSub | None, pool: Vm
                     vm_hash,
                     state,
                 )
-                # Fall through: don't reset, don't recreate. Return the
-                # existing execution so callers observing "ready" state
-                # can rely on it being the same object.
+                # Return early instead of falling through to
+                # becomes_ready() / cancel_expiration() below. Unlike the
+                # is_running branch, this path is only reached from a
+                # concurrent second allocation POST — the first POST's
+                # start_persistent_vm is still in flight and is the one
+                # that will await becomes_ready(). Calling it a second
+                # time would either deadlock (the ready event isn't set
+                # yet) or double-cancel expiration timers.
                 return execution
             logger.info(
                 "%s in terminal systemd state %s, stopping and recreating",
--- a/src/aleph/vm/orchestrator/run.py
+++ b/src/aleph/vm/orchestrator/run.py
@@ -283,7 +283,33 @@ async def start_persistent_vm(vm_hash: ItemHash, pubsub: PubSub | None, pool: Vm
             # only the owner can start it by uploading the session certificates.
             logger.info(f"{vm_hash} is waiting for its owner to initialize the confidential session")
         else:
-            logger.info(f"{vm_hash} unknown execution state, stopping the vm")
+            # None of the four fast paths matched. Before assuming the VM is
+            # dead and destroying it, check what systemd actually thinks.
+            # A VM that just started is 'activating' for a moment: not yet
+            # 'active' (so is_running is False) but times.started_at was set
+            # optimistically (so is_starting is also False). Blindly stopping
+            # here caused BTRFS corruption on the ensuing hard kill, because
+            # the firecracker API socket is not yet accepting connections and
+            # graceful shutdown falls back to SIGKILL, losing the guest's
+            # unflushed writes.
+            state = "unknown"
+            if execution.systemd_manager and execution.persistent:
+                state = execution.systemd_manager.get_service_active_state(execution.controller_service)
+            if state in ("active", "activating", "unknown"):
+                logger.info(
+                    "%s in transient systemd state %s, leaving alone",
+                    vm_hash,
+                    state,
+                )
+                # Fall through: don't reset, don't recreate. Return the
+                # existing execution so callers observing "ready" state
+                # can rely on it being the same object.
+                return execution
+            logger.info(
+                "%s in terminal systemd state %s, stopping and recreating",
+                vm_hash,
+                state,
+            )
             if execution.vm:
                 await execution.stop()
             else:
--- a/src/aleph/vm/storage.py
+++ b/src/aleph/vm/storage.py
@@ -359,9 +359,38 @@ async def create_mapped_device(device_name: str, table_command: str) -> None:
     await run_in_subprocess(command, stdin_input=table_command.encode())
 
 
+_BTRFS_CORRUPTION_MARKERS = ("open ctree failed", "parent transid verify failed")
+
+
+async def _tune_with_recovery(device_path: Path) -> None:
+    """Assign a random fsid and, on log-tree corruption, self-heal.
+
+    Unclean stops of a running instance (SIGKILL before BTRFS could flush
+    its write cache) leave the guest's rootfs with a log tree that points
+    at blocks with older transids. ``btrfstune -m`` refuses to touch the
+    volume and the VM becomes permanently unstartable. Zero the log tree
+    (losing at most the last few seconds of guest writes, which for a
+    rootfs is acceptable) and retry once.
+    """
+    try:
+        await run_in_subprocess(["btrfstune", "-m", str(device_path)])
+    except CalledProcessError as error:
+        # run_in_subprocess stashes stderr in `.output` (not `.stderr`).
+        message = error.output or ""
+        if not any(marker in message for marker in _BTRFS_CORRUPTION_MARKERS):
+            raise
+        logger.warning(
+            "BTRFS corruption on %s (%s); running `btrfs rescue zero-log` and retrying",
+            device_path,
+            message.strip(),
+        )
+        await run_in_subprocess(["btrfs", "rescue", "zero-log", str(device_path)])
+        await run_in_subprocess(["btrfstune", "-m", str(device_path)])
+
+
 async def resize_and_tune_file_system(device_path: Path, mount_path: Path) -> None:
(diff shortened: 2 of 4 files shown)


foxpatch-aleph (approved):
Well-structured defense-in-depth fix for a real production BTRFS corruption bug. The systemd state guard in run.py correctly prevents destructive stops during transient states, and the BTRFS log-tree auto-recovery in storage.py provides a clean self-healing mechanism with no risk of infinite loops. Both changes follow existing codebase conventions (synchronous D-Bus calls, `.output` for stderr in CalledProcessError). Test coverage is thorough across transient/terminal states and corruption/generic/recovery-failure paths. The only minor asymmetry — returning early without `becomes_ready()` for transient states vs. falling through to it for `is_running` — is intentional and correct since callers don't use the return value and the first POST's `becomes_ready()` is still in flight.

**`src/aleph/vm/orchestrator/run.py`** (line 297): This synchronous D-Bus call could block the event loop briefly. The existing code in `models.py:620` (`wait_for_controller_ready`) does the same thing, so this is consistent with the codebase convention. If event-loop blocking becomes a concern in the future, both call sites could be wrapped in `asyncio.to_thread` like `tasks.py:347` does for the plural variant.

**`src/aleph/vm/orchestrator/run.py`** (line 307): Returning early here skips `becomes_ready()` and `cancel_expiration()` — this is correct for a second allocation POST landing on an already-starting VM (the first POST handles those). But note the asymmetry with the `is_running` branch (line 273) which falls through to `becomes_ready()`. A brief comment explaining why the transient guard returns early instead of falling through would help future readers.

**`tests/supervisor/test_run.py`** (line 155): These transient-state tests set `execution.vm = mocker.Mock()` but the early return at line 307 means `execution.vm` is never accessed. That's fine — it mirrors the existing confidential test's setup — but it means there's no test verifying that `execution.vm` is left untouched (e.g., not stopped) in the transient case. The `stop_mock.assert_not_called()` check covers the critical path, so this is just a minor observation.

foxpatch-aleph (approved):
Well-reasoned defense-in-depth fix for a real BTRFS corruption bug caused by concurrent allocation POSTs. The transient-state guard in run.py correctly checks systemd's ActiveState before stopping a VM, only stopping on terminal states (failed/inactive/deactivating) and returning early on transient states (active/activating/unknown) to avoid double-calling becomes_ready() or cancel_expiration(). The BTRFS auto-recovery in storage.py correctly detects log-tree corruption markers, runs zero-log, and retries exactly once without looping. Error handling properly accesses CalledProcessError.output (which run_in_subprocess populates with decoded stderr). Tests are thorough, covering all transient/terminal state combinations, corruption recovery, non-corruption reraise, and zero-log failure propagation. No security issues — run_in_subprocess uses exec (not shell) and device_path is internal. Minor nit: the terminal-state test could assert becomes_ready/cancel_expiration are called on the new execution, but this is pre-existing behavior.

**`tests/supervisor/test_run.py`** (line 203): Minor: the terminal-state test asserts stop + create + result but doesn't assert `new_execution.becomes_ready.assert_awaited_once()` and `new_execution.cancel_expiration.assert_called_once()`. These are already mocked (lines 192-193) but unused. Adding these assertions would verify the new execution goes through the full start path (lines 330-334), though this is pre-existing behavior rather than new logic.

foxpatch-aleph (approved):
Well-reasoned defense-in-depth fix for a real BTRFS corruption bug caused by concurrent allocation POSTs. The systemd state guard in run.py correctly checks ActiveState before stopping, treating transient states (active/activating/unknown) as non-terminal and only stopping on definitively-terminal states (failed/inactive/deactivating). The BTRFS auto-recovery in storage.py correctly detects log-tree corruption markers, runs zero-log, and retries once — with proper error propagation for non-corruption failures and no retry loops. Test coverage is thorough across all state transitions and recovery paths. The comments explaining the race window and the rationale for returning early (avoiding double becomes_ready/cancel_expiration) are excellent.

**`src/aleph/vm/orchestrator/run.py`** (line 298): Minor observation: when `state` is `"unknown"` due to a D-Bus error, the VM is left alone and the caller gets back the existing execution. This is the safe choice, but worth noting that a persistent D-Bus failure would cause the second allocation POST to silently return a potentially-stale execution without any error to the caller. If this becomes a problem operationally, consider logging at `WARNING` level (instead of `INFO`) when `state == "unknown"` to make D-Bus failures more visible in monitoring.

**`src/aleph/vm/storage.py`** (line 379): `error.output` is typed as `str | None` on `CalledProcessError`. The `or ""` fallback handles `None` correctly, but if `run_in_subprocess` ever changes to pass raw bytes (e.g. if `.decode()` is removed), the `marker in message` check would raise `TypeError: argument of type 'bytes' is not iterable`. Not a current issue since `run_in_subprocess` explicitly calls `.decode()`, just worth being aware of if that function is refactored.

foxpatch-aleph (approved):
This PR implements a well-reasoned defense-in-depth fix for a real BTRFS corruption bug caused by concurrent allocation POSTs. The systemd state guard in run.py correctly distinguishes transient states (active/activating/unknown) from terminal ones (failed/inactive/deactivating) and returns early to avoid destructive stops during the boot window. The BTRFS recovery in storage.py is narrowly scoped to log-tree corruption markers, runs zero-log + single retry without looping, and propagates all other errors unchanged. Both changes are backed by thorough parametrized tests covering transient/terminal states and corruption/generic/zero-log-failure paths. The code is consistent with existing patterns (CalledProcessError.output usage, lazy logging, systemd ActiveState values) and introduces no security issues.

foxpatch-aleph (approved):
Well-reasoned defense-in-depth fix for a real production issue (BTRFS rootfs corruption from concurrent allocation POSTs). The transient-state guard in run.py correctly checks systemd's actual ActiveState before destroying a VM, treating active/activating/unknown as transient and only stopping on terminal states. The BTRFS auto-recovery in storage.py is appropriately conservative — it only attempts zero-log recovery on specific corruption markers and retries exactly once. Tests are thorough, covering all state transitions and recovery edge cases. The code is consistent with existing patterns (synchronous D-Bus calls, error.output usage) and introduces no security issues.

foxpatch-aleph (approved):
Well-reasoned defense-in-depth fix for a real BTRFS corruption bug caused by concurrent allocation POSTs racing with systemd's activating state. The systemd ActiveState guard correctly distinguishes transient from terminal states and returns early to avoid destructive stops. The BTRFS auto-recovery via zero-log is narrowly scoped to corruption markers and retries exactly once. Tests are thorough and cover all branches. No correctness, security, or quality issues found.

**`src/aleph/vm/orchestrator/run.py`** (line 295): Minor observation: when `execution.systemd_manager` is None or `execution.persistent` is False, `state` stays `"unknown"` and the VM is left alone. This is the safe default (don't destroy what you can't inspect), and `start_persistent_vm` should only be called for persistent VMs anyway, but it's worth noting that this path changes behavior for any non-persistent VM that somehow reaches the else branch — previously it would be stopped and recreated, now it would be left alone. Not a concern in practice given the function's contract.

PR #1082, opened 2026-08-10
fix(agent): remove V-PROGRAM staging dir on teardown

odesenfans:
Follow-up to #1071 (raised in review): the V-PROGRAM runtime bundle is extracted into `EXECUTION_ROOT/vprogram/<vm_hash>`, but the launch path only clears it on *re-extraction*. A churned or permanently-descheduled V-PROGRAM therefore leaks its extracted bundle on disk over time.

## Change
Add an idempotent `remove_vprogram_staging(vm_hash)` helper (in `vprogram_launch.py`) and call it from every path that permanently removes a V-PROGRAM:

**Teardown / deschedule paths**
- **allocation deschedule** (`views/__init__.py`) — the case the reviewer cited (V-PROGRAM dropped from the scheduler allocation).
- **terminal-status stop** (`tasks.py`) — VM stopped after repeated REJECTED/FORGOTTEN/REMOVED confirmations.
- **operator erase** (`operator.py`) — `supervisor.delete_vm(wipe=True)` clears the *daemon-side* disks, but the *agent-side* staging dir is ours to remove.

**Create-time failure paths** (`run.py`, also flagged in review) — `build_vprogram_spec` extracts the bundle before agent-side capacity admission runs, so a create-path failure (capacity refused, or a post-extraction spec build error) and the wait-failure teardown both now clean the staging dir.

The helper is safe for any VM type: a non-V-PROGRAM has no such directory, so the call is a no-op (guards on existence, `rmtree(ignore_errors=True)`).

## Tests
`test_remove_vprogram_staging_is_idempotent`: removes a populated staging dir and is a no-op (never raises) when there is nothing to remove.

## Base
Stacked on `od/vprogram-launch` (#1071) since it depends on `vprogram_staging_dir`. Retarget to `dev` once #1071 merges.

### DIFF
--- a/src/aleph/vm/agent/run.py
+++ b/src/aleph/vm/agent/run.py
@@ -29,7 +29,7 @@ from aleph.vm.agent.translate import build_create_vm_spec, build_program_create_
 from aleph.vm.agent.update_watcher import UpdateWatcher
 from aleph.vm.agent.vm.program_client import ProgramGuestClient
 from aleph.vm.agent.vm_registry import AgentVmRegistry, persist_record
-from aleph.vm.agent.vprogram_launch import build_vprogram_spec
+from aleph.vm.agent.vprogram_launch import build_vprogram_spec, remove_vprogram_staging
 from aleph.vm.conf import settings
 from aleph.vm.resources import InsufficientResourcesError
 from aleph.vm.supervisor_interface import errors as supervisor_errors
@@ -368,6 +368,9 @@ async def create_vm_execution(
             info = await supervisor.create_vm(spec)
         except Exception:
             registry.forget(vm_hash)
+            # build_vprogram_spec may have already extracted the bundle (e.g.
+            # capacity admission fails after staging): do not leak it.
+            remove_vprogram_staging(vm_hash)
             raise
         try:
             await _wait_until_running(supervisor, info.vm_id)
@@ -377,6 +380,7 @@ async def create_vm_execution(
                 await supervisor.delete_vm(info.vm_id)
             except Exception:
                 logger.exception("Teardown of half-started V-PROGRAM %s failed", vm_hash)
+            remove_vprogram_staging(vm_hash)
             raise
         await persist_record(vm_hash, record)
         return None
--- a/src/aleph/vm/agent/tasks.py
+++ b/src/aleph/vm/agent/tasks.py
@@ -35,6 +35,7 @@ from aleph.vm.agent.utils import (
     is_after_community_wallet_start,
 )
 from aleph.vm.agent.vm_registry import AgentVmRegistry
+from aleph.vm.agent.vprogram_launch import remove_vprogram_staging
 from aleph.vm.conf import settings
 from aleph.vm.supervisor_interface.abc import Supervisor
 from aleph.vm.supervisor_interface.errors import VmNotFoundError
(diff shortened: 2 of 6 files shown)


foxpatch-aleph (approved):
Clean, well-scoped fix that addresses a real disk-leak: the V-PROGRAM staging directory was only cleaned on re-extraction, not on permanent teardown. The new `remove_vprogram_staging` helper is idempotent, safe for non-V-PROGRAM VMs (no-op when the directory doesn't exist), and correctly placed after `registry.forget` + `delete_records_for_vm` in all three permanent teardown paths. The insufficient-balance stops in `tasks.py` are correctly excluded — they are temporary stops where the VM stays in the registry and the staging should be reused on restart. The test covers both the removal and idempotent no-op cases. The `staging.exists()` guard is technically redundant with `ignore_errors=True` but serves to gate the debug log, which is fine.

foxpatch-aleph (approved):
Well-scoped follow-up that adds an idempotent staging-dir cleanup helper and wires it into every permanent V-PROGRAM removal path. The helper is safe for non-V-PROGRAMs (no-op when the directory doesn't exist), the TOCTOU between exists() and rmtree is handled by ignore_errors=True, and no circular imports are introduced. I verified that all other delete_vm call sites in the agent cannot reach V-PROGRAMs (payment grouping skips them, expiry/update-watcher are program-only, run_code rejects them, migration uses wipe=False, and reinstall re-creates rather than deletes). The test covers both the removal and idempotent no-op cases with proper EXECUTION_ROOT isolation.

foxpatch-aleph (approved):
The PR correctly adds an idempotent `remove_vprogram_staging` helper and calls it from all permanent V-PROGRAM removal paths: allocation deschedule, terminal-status stop, operator erase, and both create-time failure paths. The helper is safe for non-V-PROGRAM VMs (no staging dir = no-op) and handles TOCTOU races via `ignore_errors=True`. I verified that no teardown paths are missed — V-PROGRAMs are excluded from payment monitoring, are always persistent (so expiry/idle-reap doesn't apply), and `operate_stop` calls `stop_vm` (not `delete_vm`) for persistent VMs, so staging is correctly preserved for restart. The test covers both the removal and idempotent no-op cases. The `exists()` guard before `rmtree(ignore_errors=True)` is technically redundant but justified by avoiding a misleading debug log for non-V-PROGRAM VMs.

**`src/aleph/vm/agent/vprogram_launch.py`** (line 77): The `if staging.exists()` guard is technically redundant since `shutil.rmtree(staging, ignore_errors=True)` on a non-existent path is already a no-op, but it serves a useful purpose: it suppresses the debug log for non-V-PROGRAM VMs that have no staging dir. This is fine as-is.

foxpatch-aleph (approved):
The PR adds an idempotent `remove_vprogram_staging` helper and calls it from all five permanent-teardown paths for V-PROGRAMs: allocation deschedule, terminal-status stop, operator erase, and both create-time failure paths. I verified every other `delete_vm` call site in the agent and confirmed the remaining paths either cannot reach V-PROGRAMs (payment checks skip them, migration rejects confidential VMs, on-demand execution paths reject them, expiry is non-persistent only) or are recreate cycles where `_extract_bundle` already clears and re-extracts the staging dir (update watcher, start_persistent_vm FAILED recovery). The ordering is correct in all call sites: staging cleanup runs after `supervisor.delete_vm` and `registry.forget`. The helper is safe for non-V-PROGRAMs (no-op when the directory doesn't exist) and uses `rmtree(ignore_errors=True)` for concurrency safety. The unit test covers both the removal and double-call (idempotency) cases. No security concerns: `vm_hash` is a validated `ItemHash`, so the path is injection-safe.

PR #1083, opened 2026-08-10
feat(agent): specific error when a VM never reaches RUNNING

odesenfans:
Follow-up from the #1071 review (wait-failure error mapping), done as a standalone change against `dev` because it is cross-cutting, not V-PROGRAM-specific.

## Problem
`_wait_until_running` raised a bare `RuntimeError` (VM entered a terminal FAILED/STOPPED status) or `asyncio.TimeoutError` (never reached RUNNING before the deadline). Neither is caught by any specific arm of `create_vm_execution_or_raise_http_error`, so **every** VM type (program, instance, v-program) surfaced a start failure as the generic `HTTPInternalServerError(reason="Unhandled error during initialisation")` — indistinguishable from a real unexpected crash.

## Change
- Add an agent-side `VmStartupError` (in `run.py`, not the `SupervisorError` boundary vocabulary — it never crosses the supervisor gRPC boundary).
- `_wait_until_running` raises it for both the terminal-status and timeout cases.
- `create_vm_execution_or_raise_http_error` maps it to `reason="VM failed to start"`, ahead of the generic handler.

All three create paths already tear the half-started VM down (`registry.forget` + `delete_vm`) and re-raise, so teardown behaviour is unchanged; only the surfaced HTTP reason improves.

## Tests
- Updated the direct wait-loop tests (`test_supervisor_run_helpers`) and the instance (`test_supervisor_run_routing`) and v-program (`test_vprogram`) failure-path tests to expect `VmStartupError`.
- Added `test_startup_failure_maps_to_specific_http_reason` asserting the `"VM failed to start"` HTTP reason.

### DIFF
--- a/src/aleph/vm/agent/run.py
+++ b/src/aleph/vm/agent/run.py
@@ -149,6 +149,16 @@ async def reconcile_port_forwards(supervisor: Supervisor, vm_id: VmId, content) 
             await supervisor.add_port_forward(spec)
 
 
+class VmStartupError(Exception):
+    """A VM was created but never reached the running state: it entered a
+    terminal status (FAILED/STOPPED) or timed out before RUNNING.
+
+    Agent-internal (raised by ``_wait_until_running``, never crosses the
+    Supervisor boundary), so it is not part of the SupervisorError vocabulary.
+    ``create_vm_execution_or_raise_http_error`` maps it to a clear HTTP reason
+    instead of the generic "unhandled error" bucket, for every VM type."""
+
+
 async def _wait_until_running(
     supervisor: Supervisor,
     vm_id: VmId,
@@ -159,8 +169,8 @@ async def _wait_until_running(
     """Poll get_vm until the VM reports RUNNING.
 
     In-process the first poll already reports RUNNING (create_vm blocked until
-    boot); across a future gRPC boundary this does real work. Raises on a
-    terminal status or after `timeout` seconds.
+    boot); across a future gRPC boundary this does real work. Raises
+    VmStartupError on a terminal status or after `timeout` seconds.
 
     `timeout`/`interval` default to the module constants, resolved at call time
     so tests (and operators) can override them by patching the constants.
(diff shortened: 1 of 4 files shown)


foxpatch-aleph (commented):
The core change is correct and well-executed: VmStartupError properly replaces the bare RuntimeError/TimeoutError in _wait_until_running, the except arm in create_vm_execution_or_raise_http_error is correctly ordered before the generic handler, and all existing except Exception teardown blocks still catch the new error. Tests are updated comprehensively. One gap: the on-demand program path (_ensure_program_vm -> _raise_http_for_program_error) also calls _wait_until_running but does NOT have a VmStartupError branch in its error mapper, so on-demand program VMs that never reach RUNNING will still surface as the generic 'Unhandled error during initialisation' — the PR description's claim of 'every VM type' is slightly overstated. This may be intentional scoping but should be acknowledged.

**`src/aleph/vm/agent/run.py`** (line 158): The docstring says 'for every VM type', but the on-demand program path (_ensure_program_vm at line 500 -> _raise_http_for_program_error at line 464) also calls _wait_until_running and does NOT have a VmStartupError branch. A VmStartupError from that path falls through to the generic 'Unhandled error during initialisation' at line 497. Consider either adding a VmStartupError branch to _raise_http_for_program_error or adjusting the docstring to say 'for instance and v-program create paths'.

**`src/aleph/vm/agent/run.py`** (line 432): Correct placement — VmStartupError is caught before VmSetupError and the generic except Exception, so it gets the specific 'VM failed to start' reason. The logger.warning level (not .exception) is appropriate since this is an expected failure mode, not an unexpected crash.

foxpatch-aleph (approved):
This PR cleanly introduces a dedicated VmStartupError for the case where a VM is created but never reaches RUNNING (terminal status or timeout), replacing the previously bare RuntimeError/asyncio.TimeoutError. The new exception inherits from plain Exception (not SupervisorError), so it is correctly disjoint from the existing supervisor error vocabulary and is placed ahead of the generic fallback in both HTTP mappers (create_vm_execution_or_raise_http_error and _raise_http_for_program_error). All three create paths (program, instance, v-program) already tear down on except Exception and re-raise, so teardown behavior is unchanged. Tests are updated comprehensively: the two direct wait-loop tests, the instance routing teardown test, and the v-program teardown test all expect VmStartupError, and two new tests assert the specific HTTP reason mapping for both the instance and on-demand program paths. The removed asyncio imports from both test files are safe since all remaining references go through run_module.asyncio.

PR #1084, opened 2026-08-11
perf(allocations): batch systemd state lookup in update_allocations

aliel:
## Summary

\`POST /control/allocations\` reconciles the scheduler's push against the local pool: walks all persistent executions to find stop candidates, then calls \`start_persistent_vm\` for every hash in the pushed list. Every \`execution.is_running\` call is a synchronous D-Bus round-trip to systemd. On a CRN with 80+ persistent VMs and a full-list push, that's ~2N blocking round-trips on the event loop behind \`allocation_lock\`. Once the total wall time crosses aleph-admin's HTTP client timeout, the request times out; restarting the supervisor \"fixes\" it by cancelling the in-flight allocation and resetting the lock, but the underlying scaling problem returns with the next attempt.

Reported symptom (crn10.leviathan.so, 80+ instances, allocating a fresh instance):
\`\`\`
Error: crn request to https://crn10.leviathan.so/control/allocations failed: timed out
\`\`\`

## Fix

Same shape as #963 for \`monitor_payments\`:

- \`update_allocations\` snapshots systemd's ActiveState for every persistent controller in one \`ListUnits()\` call, off the event loop via \`asyncio.to_thread\`.
- Pass the dict to \`pool.get_persistent_executions\` and \`start_persistent_vm\` via a new \`running_states\` parameter. Both prefer the map lookup over \`execution.is_running\` when it is provided; fall back to the per-VM path when it isn't, so existing callers are unchanged.
- Drop the redundant \`execution.is_running\` check inside the stop loop: \`get_persistent_executions\` already filtered by running via the map, and calling it again inside the loop would be another per-VM D-Bus round-trip we now avoid.

Per-request D-Bus work drops from **~2N synchronous round-trips on the event loop** to **one \`ListUnits\` call off-thread**. On a CRN with 80 VMs the allocation call goes from \"sometimes times out\" to a few hundred milliseconds.

## Test plan
- [x] \`start_persistent_vm\`'s fast-path with a map showing \"running\" bypasses \`is_service_active\` entirely.
- [x] A missing entry in \`running_states\` falls through to #1081's transient-state guard and composes correctly.
- [x] \`pool.get_persistent_executions\` with a map never calls per-VM \`is_running\`.
- [ ] Deploy to a staging CRN, POST /control/allocations with a large list, verify sub-second response time.

## Related
- #963 (\`monitor_payments\` batched via the same pattern)
- #1081 (transient-state guard in \`start_persistent_vm\` — this PR composes with it, not against it)

### DIFF
--- a/src/aleph/vm/orchestrator/run.py
+++ b/src/aleph/vm/orchestrator/run.py
@@ -267,10 +267,27 @@ async def run_code_on_event(vm_hash: ItemHash, event, pubsub: PubSub, pool: VmPo
             await execution.stop()
 
 
-async def start_persistent_vm(vm_hash: ItemHash, pubsub: PubSub | None, pool: VmPool) -> VmExecution:
+async def start_persistent_vm(
+    vm_hash: ItemHash,
+    pubsub: PubSub | None,
+    pool: VmPool,
+    running_states: dict[str, bool] | None = None,
+) -> VmExecution:
+    """Start (or attach to) a persistent VM.
+
+    ``running_states`` is an optional pre-computed mapping of
+    controller_service name -> active state, used to replace the
+    per-VM ``is_running`` D-Bus round-trip with a dict lookup. Callers
+    reconciling many VMs in a row (``update_allocations``) should pass
+    it to avoid stalling the event loop for seconds at a time.
+    """
     execution: VmExecution | None = pool.executions.get(vm_hash)
     if execution:
-        if execution.is_running:
+        if running_states is not None and execution.persistent:
+            is_running_now = running_states.get(execution.controller_service, False)
+        else:
+            is_running_now = execution.is_running
+        if is_running_now:
             logger.debug(f"{vm_hash} is already running")
         elif execution.is_starting:
             logger.debug(f"{vm_hash} is already starting")
--- a/src/aleph/vm/orchestrator/views/__init__.py
+++ b/src/aleph/vm/orchestrator/views/__init__.py
@@ -528,14 +528,33 @@ async def update_allocations(request: web.Request):
 
     async with allocation_lock:
         logger.debug("Got allocation_lock, updating allocations")
+
+        # Snapshot systemd's ActiveState for every persistent controller
+        # in one batched D-Bus ListUnits() call off the event loop. The
+        # per-VM path (execution.is_running -> is_service_active) would
+        # otherwise issue O(N) synchronous D-Bus round-trips inside this
+        # loop; on CRNs with dozens of persistent VMs that stalls the
+        # loop long enough for the aleph-admin HTTP client to time out.
+        persistent_services = [
+            execution.controller_service for execution in pool.executions.values() if execution.persistent
+        ]
+        running_states: dict[str, bool] = {}
+        if persistent_services and pool.systemd_manager:
+            running_states = await asyncio.to_thread(
+                pool.systemd_manager.get_services_active_states,
+                persistent_services,
+            )
+
         # First, free resources from persistent programs and instances that are not scheduled anymore.
         allocations = allocation.persistent_vms | allocation.instances
         stopped_vms = []
-        # Make a copy since the pool is modified
-        for execution in list(pool.get_persistent_executions()):
+        # Make a copy since the pool is modified. get_persistent_executions
+        # already filtered by "running" via the batched map, so the extra
+        # is_running check the old code did here is both redundant and a
+        # per-VM D-Bus round-trip we now avoid.
+        for execution in list(pool.get_persistent_executions(running_states=running_states)):
             if (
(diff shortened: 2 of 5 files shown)


foxpatch-aleph (changes requested):
The core optimization is correct and follows an established pattern (#963) already used by `load_persistent_executions` and `get_executions_by_address`. The `running_states` parameter is properly optional with `None` default so existing callers are unaffected, the redundant `is_running` check removal is sound, and the new tests pin the right invariants. However, the PR breaks at least three existing tests in `test_views.py` (`test_allocation_valid_token`, `test_allocation_aleph_eip191_v1_valid`, `test_allocation_legacy_token_no_per_request_log`) because the new code in `update_allocations` accesses `pool.executions.values()` (line 539) and `pool.systemd_manager` (line 542) before reaching `get_persistent_executions`. The `FakeVmPool` (line 354) and `_FakeVmPool` (line 1150) stubs in `test_views.py` don't define `executions` or `systemd_manager` attributes, and their `get_persistent_executions` signature doesn't accept the new `running_states` keyword argument. These stubs must be updated to match the new interface, otherwise the tests will crash with `AttributeError`.

**`tests/supervisor/test_views.py`** (line 354): The `FakeVmPool` stub needs to be updated for the new `update_allocations` code path. It now accesses `pool.executions.values()` (views/__init__.py:539) and `pool.systemd_manager` (views/__init__.py:542) before calling `get_persistent_executions`. Without an `executions` attribute, the test will crash with `AttributeError` at line 539. Fix: add `executions = {}` and `systemd_manager = None` as class attributes, and update the `get_persistent_executions` signature to accept `running_states=None`.

**`tests/supervisor/test_views.py`** (line 1150): Same issue as `FakeVmPool` above: `_FakeVmPool` is used by `test_allocation_aleph_eip191_v1_valid` (line 1186) and `test_allocation_legacy_token_no_per_request_log` (line 1234), both of which POST to `/control/allocations` and expect HTTP 200. The new code will crash with `AttributeError` on `pool.executions`. Add `executions = {}` and `systemd_manager = None` class attributes, and update `get_persistent_executions` to accept `running_states=None`.

**`src/aleph/vm/orchestrator/views/__init__.py`** (line 541): Minor observation: when `persistent_services` is empty or `pool.systemd_manager` is falsy, `running_states` stays as `{}` (empty dict, not `None`). This means `get_persistent_executions(running_states={})` takes the map path and returns `False` for every service, yielding no executions. In production this is fine (systemd_manager is always set, and if there are no persistent services there's nothing to iterate), but the semantic difference between `running_states={}` (everything is not-running) and `running_states=None` (fall back to per-VM is_running) is subtle. Consider initializing `running_states` as `None` instead of `{}` so the fallback path is taken when no batch was performed, matching the behavior described in the docstring.

foxpatch-aleph (approved):
This PR is a well-executed performance optimization that batches N per-VM synchronous D-Bus round-trips into a single ListUnits() call off the event loop, directly addressing a timeout issue on CRNs with many persistent VMs. The change follows the exact pattern already established by #963 for monitor_payments, maintaining consistency across the codebase. The semantic difference between the batch path (ActiveState only) and the per-VM path (is_service_enabled + ActiveState) is immaterial because controller services are always enabled via EnableUnitFiles before StartUnit, as already documented in get_executions_by_address's docstring. The removal of the redundant execution.is_running check inside the stop loop is correct since get_persistent_executions already filters by running state. The transient-state guard from #1081 composes correctly with the new map path, as verified by the test suite. Tests pin the key invariants (map wins over D-Bus, missing entries fall through, get_persistent_executions doesn't call per-VM is_running). The TOCTOU window between snapshot and use is safe because stopped VMs are removed from the pool via forget_vm, causing pool.executions.get() to return None and bypass the map.

foxpatch-aleph (approved):
This PR correctly batches the per-VM systemd ActiveState lookups in `update_allocations` into a single off-thread `ListUnits()` call, following the same pattern already established by PR #963 for `monitor_payments`. The `running_states` map is built from persistent executions' `controller_service` names and is correctly consumed by both `get_persistent_executions` and `start_persistent_vm`, with proper `None`-sentinel handling so existing callers are unaffected. The removal of the redundant `execution.is_running` check in the stop loop is safe since `get_persistent_executions` already filters by running state. The `running_states` snapshot staleness during the start loop is not a concern given the `allocation_lock` prevents concurrent modifications, and VMs in the allocation list are never stopped by the stop loop. The semantic difference (batch path skips `is_service_enabled`) was already documented and accepted in `get_executions_by_address`. Tests are well-targeted, pinning the map-wins-over-D-Bus invariant and the missing-service fall-through path. The fake pool updates in test_views.py correctly add the `executions` and `systemd_manager` attributes that `update_allocations` now reads.

PR #1085, opened 2026-08-12
fix(agent): heal port forwards when re-adopting a running VM

odesenfans:
## The gap

Follow-up to #1079, closing the re-adoption case its review noted: only the create path ever applies port forwards (right after the VM first reaches RUNNING). A VM adopted already-created by `start_persistent_vm` (typically after an agent restart) skipped that step in the new agent's lifetime, for instances and V-PROGRAMs alike. Two concrete consequences:

- If the previous agent life crashed in the window between RUNNING and the forward setup, the VM stayed up but permanently unreachable: no SSH for an instance, no attestation endpoint for a V-PROGRAM (whose sole consumer is the CLI's attested call).
- Instances also missed any port-forwarding aggregate change published while the agent was down, past the live aggregate watcher.

## The fix

Every `start_persistent_vm` adoption branch that ends with a RUNNING VM (already running, still booting, resumed from STOPPED) now converges the hypervisor's forwards onto the desired set through the same idempotent `_reconcile_forwards` machinery #1079 introduced (which was written for exactly this call site). The recreate branches already reconcile through the create path.

Deriving the desired set cheaply, per VM type:

- **V-PROGRAM**: new `resolve_vprogram_attestation_port` re-reads the RA-TLS port from the runtime manifest alone (one message-metadata lookup; the manifest file is in the local download cache after the first create). No `build_vprogram_spec`, no bundle staging: the running VM already booted from that bundle.
- **Instance**: re-resolve the user's port-forwarding aggregate, strictly (see below).
- **Program**: no agent-side forwards, nothing to heal.

Two deliberate semantics:

- **Best-effort, unlike the create path.** The adopted VM is already up and possibly serving; a healing failure logs a warning and leaves the forwards exactly as found, rather than failing the allocation and inviting scheduler churn. The create path keeps its loud fail-and-teardown behavior.
- **Strict aggregate resolution on this path** (new `strict` keyword on `resolve_port_forwards`). The create-path fallback turns an aggregate-fetch failure into SSH-only, which is fine when there is nothing to lose; driving a *convergence* with that fallback would remove the user's declared forwards on a transient CCN error. Strict failures skip healing instead.

The `reconcile_vprogram_port_forwards` docstring loses its "though today only the create path does" caveat, which this PR makes true.

## Tests

- Adopted RUNNING V-PROGRAM with a missing forward gets it healed from the manifest; a second adoption no-ops (idempotence), no recreate.
- Manifest re-fetch failure on adoption: allocation still succeeds, existing forwards untouched.
- Missing agent record (DB loss): supervisor not touched at all.
- Adopted RUNNING instance reconciles with `strict=True`; a strict aggregate-fetch failure removes nothing and adds nothing.
- Awaiting-init confidential VMs are not healed (not RUNNING); the STOPPED resume-in-place branch is.
- `resolve_port_forwards(strict=True)` raises on fetch failure; the default fallback behavior is unchanged.

Local run of every test file importing the touched modules: identical results to the base branch plus the 9 new tests (the base's local-only failures are environmental and unchanged).

### DIFF
--- a/src/aleph/vm/agent/run.py
+++ b/src/aleph/vm/agent/run.py
@@ -29,7 +29,10 @@ from aleph.vm.agent.translate import build_create_vm_spec, build_program_create_
 from aleph.vm.agent.update_watcher import UpdateWatcher
 from aleph.vm.agent.vm.program_client import ProgramGuestClient
 from aleph.vm.agent.vm_registry import AgentVmRegistry, persist_record
-from aleph.vm.agent.vprogram_launch import build_vprogram_spec
+from aleph.vm.agent.vprogram_launch import (
+    build_vprogram_spec,
+    resolve_vprogram_attestation_port,
+)
 from aleph.vm.conf import settings
 from aleph.vm.resources import InsufficientResourcesError
 from aleph.vm.supervisor_interface import errors as supervisor_errors
@@ -97,13 +100,20 @@ def _is_spec_eligible(content) -> bool:
     return isinstance(content, InstanceContent)
 
 
-async def resolve_port_forwards(vm_id: VmId, content) -> list[PortForwardSpec]:
+async def resolve_port_forwards(vm_id: VmId, content, *, strict: bool = False) -> list[PortForwardSpec]:
     """Agent-side policy: translate the user's port-forwarding aggregate settings
     into the set of forwards the hypervisor should apply.
 
     This is the agent half of the old VmExecution.fetch_port_redirect_config_and_setup.
     Nothing here touches nftables; the caller applies each spec through
     supervisor.add_port_forward. host_port is left 0; the hypervisor assigns it.
+
+    ``strict=True`` turns an aggregate-fetch failure into an exception instead
+    of the SSH-only fallback. The fallback is right on create (worst case the
+    VM starts with SSH only), but a convergence caller reconciling an already
+    forwarded VM must not treat "could not fetch" as "the user removed every
+    port": converging on the fallback set would tear the user's forwards down
+    on a transient CCN error. Strict callers skip reconciling instead.
     """
     ports_requests: dict[int, dict[str, bool]] = {}
     try:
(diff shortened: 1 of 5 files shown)


foxpatch-aleph (approved):
Well-structured PR that closes a real gap: VMs re-adopted after an agent restart never had their port forwards healed, leaving them permanently unreachable if the previous agent life crashed between RUNNING and forward setup. The fix correctly converges hypervisor forwards for every adoption branch ending in a RUNNING VM, while deliberately using a best-effort (not loud-fail) approach and strict aggregate resolution to avoid tearing down user-declared forwards on transient CCN errors. The `resolve_vprogram_attestation_port` helper efficiently re-derives the attestation port from the manifest alone without re-staging the bundle. Test coverage is thorough (9 new tests) and covers all new branches, edge cases (DB loss, manifest failure, awaiting-init, STOPPED resume), and the strict-mode semantics. No correctness, security, or quality issues found.

foxpatch-aleph (approved):
Well-structured PR that closes a real gap in the re-adoption path. The strict/fallback distinction on resolve_port_forwards is the correct semantic for convergence vs. create callers. Best-effort healing with warning logs is the right design for already-running VMs. Control flow is correct across all branches in start_persistent_vm. Test coverage is thorough with 9 new tests covering heal success, idempotence, error paths, and noop cases. No security concerns. The only minor gap is a missing dedicated test for the DEFINED/BOOTING branch reaching heal, though the same guard logic is tested via the STOPPED branch.

**`tests/supervisor/test_supervisor_run_routing.py`** (line 604): Consider adding a test for the DEFINED/BOOTING branch reaching the heal path (e.g., `test_start_persistent_readopt_booting_heals_after_wait`). The STOPPED test covers the same guard logic, but a booting-specific test would make the intent explicit and protect against a future refactor that accidentally breaks the heal on the booting path.

PR #1086, opened 2026-08-13
feat(supervisor-cli): log output sanitization and RPC timeouts

odesenfans:
Follow-up to #1039, addressing two review findings.

## Log sanitization

Guest serial output is untrusted input to the operator's terminal: a hostile VM can embed ANSI/OSC escape sequences in its log lines (window title spoofing, screen manipulation, OSC 52 clipboard writes; same class as kubectl's CVE-2021-25743). `alephctl logs` now renders every control character except tab in escaped form (`\u{1b}]0;...` instead of a live escape) on the plain-text path. `--json` output was already safe through serde escaping.

## RPC timeouts

The CLI's main use case is poking a possibly-sick daemon, and the kernel queues UDS connections even when the process never calls accept, so every command previously hung forever on a wedged supervisor. Unary RPCs are now bounded by a global `--timeout` flag (default 10s, `--timeout 0` disables), plus a 5s connect timeout on the endpoint. Streaming commands (`logs -f`, `events`) are exempt and still run until interrupted.

## Tests

- `sanitize_line` unit tests (C0, C1, tab/UTF-8 preservation) and an end-to-end test streaming an OSC 52 payload through the fake supervisor.
- Timeout test against a fake whose `Health` handler stalls 30s: the call fails within the 200ms deadline.
- `--timeout` parsing and the streaming-command exemption are unit-tested.

### DIFF
--- a/rust/crates/supervisor-cli/src/cli.rs
+++ b/rust/crates/supervisor-cli/src/cli.rs
@@ -21,6 +21,11 @@ pub struct Cli {
     #[arg(long, global = true)]
     pub json: bool,
 
+    /// Seconds before a call to a wedged supervisor is abandoned (0 waits
+    /// forever). Streaming commands (logs -f, events) are never cut short.
+    #[arg(long, global = true, default_value_t = 10)]
+    pub timeout: u64,
+
     #[command(subcommand)]
     pub command: Command,
 }
@@ -49,6 +54,14 @@ pub enum Command {
     /// Port forwards.
     #[command(subcommand)]
     Ports(PortsCommand),
+}
+
+impl Command {
+    /// Long-lived server streams that must outlive any per-request
+    /// deadline; main.rs skips the request timeout for these.
+    pub fn is_streaming(&self) -> bool {
+        matches!(self, Command::Events | Command::Logs { follow: true, .. })
+    }
 }
 
 #[derive(Subcommand, Debug)]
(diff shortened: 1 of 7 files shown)


foxpatch-aleph (approved):
Well-executed security hardening PR addressing two review findings from #1039. Log sanitization correctly escapes all control characters (C0 and C1) except tab via char::escape_debug(), neutralizing ANSI/OSC terminal injection attacks from untrusted guest serial output while preserving readable output. The RPC timeout implementation is clean: a global --timeout flag (default 10s, 0 disables) bounds unary RPCs via tonic's endpoint.timeout(), with a separate 5s connect_timeout always applied, and streaming commands (logs -f, events) correctly exempted via is_streaming(). Test coverage is thorough across unit and integration levels, including an end-to-end OSC 52 payload test and a real timeout against a stalled fake supervisor. Code quality is high with clear doc comments explaining the security rationale.

**`rust/crates/supervisor-cli/src/output.rs`** (line 172): Consider adding a test case for embedded newlines in sanitize_line, e.g. assert that sanitize_line("line1\nline2") produces "line1\\nline2" (escaped backslash-n, not a real newline). This confirms the log-injection prevention behavior where writeln! adds the only real newline. Non-blocking — the behavior is correct by construction from escape_debug.

foxpatch-aleph (approved):
Well-executed PR addressing two real security/operability concerns. Log sanitization correctly neutralizes ANSI/OSC escape sequences from untrusted guest serial output using `escape_debug()`, which renders control characters as inert text while preserving tabs and UTF-8. The RPC timeout feature properly bounds unary calls with a configurable deadline while exempting streaming commands (`logs -f`, `events`), and adds a 5s connect timeout that protects against wedged supervisors even when `--timeout 0`. The `is_streaming()` classification is correct and well-tested. All existing tests were updated for the new `connect` signature, and new tests cover sanitization (C0/C1/tab/UTF-8), the OSC 52 end-to-end path, timeout behavior against a stalled supervisor, and `--timeout` parsing. The code is clean, idiomatic, and follows existing conventions.

**`rust/crates/supervisor-cli/src/output.rs`** (line 70): Minor observation: `\n` (U+000A) is a control character and will be escaped to `\\n` by `escape_debug()`. This is correct (it prevents a guest from injecting extra lines into the terminal), but it's a subtle behavior that could surprise users who expect newlines to be preserved in log output. Consider mentioning in the doc comment that embedded newlines are also escaped, since the proto says lines have "no trailing newline" but a hostile guest could still embed them.

**`rust/crates/supervisor-cli/tests/cli.rs`** (line 76): Minor test gap: there's a test verifying the timeout fires (positive case) and that `--timeout 0` parses correctly, but no integration test verifying that `--timeout 0` actually lets a slow call succeed (i.e., that `None` means no deadline). The logic is trivially correct by inspection, so this is non-blocking, but a test like `health_stall: Some(100ms)` with no timeout succeeding would make the round-trip explicit.

PR #1087, opened 2026-08-13
feat(vprogram): static IPv6 for confidential guests via per-tap DHCPv6 + RA

odesenfans:
Supersedes #1080 (SLAAC), which could not compose with the static /124
model: SLAAC needs a /64 PIO, so a host RA sender could never deliver the
daemon-allocated /124 address, and a shared-/64 SLAAC address would escape
the ndppd proxy range and the vm-type hextet scheme. Design:
docs/plans/2026-08-13-vprogram-ipv6-dhcpv6-design.md.

Static end-to-end instead, mirroring the shipped IPv4 SNP path:

- daemon: `VmType::VProgram` (0x4 hextet) at create and adoption (both
  paths regression-tested, including red-on-revert for the adoption
  predicate); the existing per-tap dnsmasq additionally serves a
  single-address stateful DHCPv6 range + `--enable-ra` (M=1/A=0: default
  route from the RA, no SLAAC). No new daemon, no radvd. `dnsmasq --test`
  accepts the generated v4+v6 ranges.
- guest (`nix/`): `udhcpc6` alongside `udhcpc`, both now with `-n` so a
  missing DHCP server bounds out to IPv4-only (or no-net) instead of
  hanging PID 1; new `udhcpc6.script`; ICMPv6 ND/RA/PMTU accept rule
  (policy drop preserved). No udp 68/546 rules: both clients run `-q` and
  exit before the firewall exists.
- attest-agent: bind `[::]:8443` dual-stack (Linux `IPV6_V6ONLY=0`
  default, verified for actix-web), so the IPv4 path and host port-map
  are unaffected.

**Measurement impact:** init.sh + udhcpc6.script + attest-agent shift the
measured image; a downstream measurement re-pin is required before this
ships.

**Not verified in CI (manual E2E on an SNP host):** external
`curl https://[guest_v6]:8443` reaching the agent; leased address equals
the daemon-derived one; v6 default route surviving past the RA lifetime;
upgraded daemon + old image (v4-only, no hang) and old daemon + new image
(bounded udhcpc6 timeout, IPv4-only) skew cases.

### DIFF
--- a/docs/plans/2026-08-13-vprogram-ipv6-dhcpv6-design.md
+++ b/docs/plans/2026-08-13-vprogram-ipv6-dhcpv6-design.md
@@ -59,9 +59,8 @@ guest never autoconfigures.
     allocated address" property as the v4 range;
   - `--enable-ra`: RAs on the tap. With a plain (non-`slaac`, non-`ra-only`)
     DHCPv6 range dnsmasq advertises M=1/A=0: default route yes, SLAAC no.
-  - `--dhcp-option=option6:dns-server,[...]` only if the daemon has
-    v6-capable nameservers; otherwise the guest keeps the v4 DNS servers
-    from udhcpc (acceptable).
+  - v6 DNS is not emitted (no `option6:dns-server`, deferred per section 6);
+    the guest keeps the v4 DNS servers from udhcpc (acceptable).
 - `lifecycle.rs`: no structural change. Same dnsmasq start/stop points,
   same `Type=exec` transient unit, same teardown. The tap already carries
   the host-side gateway (`create_tap` adds `host_ipv6_cidr`), so dnsmasq
@@ -72,9 +71,11 @@ guest never autoconfigures.
 
 - Enable IPv6 on the interface and `accept_ra=2` (kernel installs the
   default route from the RA; reused from the donor branch).
-- `busybox udhcpc6 -i $iface -q -t 5 -A 2 -s /bin/udhcpc6.script`, bounded
-  and non-fatal exactly like the v4 `udhcpc` call: no DHCPv6 server on the
-  tap means the guest simply stays IPv4-only.
+- `busybox udhcpc6 -i $iface -q -n -t 5 -A 2 -s /bin/udhcpc6.script`, bounded
+  and non-fatal exactly like the v4 `udhcpc` call (which also gains `-n`):
+  `-n` makes the client exit 1 once its solicit retries are exhausted
+  instead of looping forever, so `||` in init.sh always runs and no DHCPv6
+  server on the tap means the guest simply stays IPv4-only.
 - `udhcpc6.script` mirrors `udhcpc.script`: applies the leased address to
   the interface. No route handling (the RA covers it), no resolv.conf
   handling unless option6:dns-server was sent.
--- a/docs/plans/2026-08-13-vprogram-ipv6-dhcpv6-design.md
+++ b/docs/plans/2026-08-13-vprogram-ipv6-dhcpv6-design.md
@@ -1,0 +1,159 @@
+# V-PROGRAM guest IPv6: static allocation via per-tap DHCPv6 + RA
+
+Status: approved design, 2026-08-13.
+Supersedes the SLAAC approach of PR #1080 (branch `od/vprogram-ipv6`, to be
+closed). Base branch: `origin/dev`. The old branch is a parts donor only.
+
+## 1. Problem
+
+Confidential V-PROGRAM guests (SEV-SNP measured boot) need inbound IPv6
+reachability to the attestation/serving port `:8443`, without any per-VM data
+in the measured image or kernel cmdline (the publisher precomputes one
+measurement for all deployments).
+
+PR #1080 solved the guest half with SLAAC (dynamic allocation) and left "a
+host RA sender (radvd)" as a follow-up. That follow-up cannot work:
+
+1. Linux only autoconfigures from a prefix advertised as a /64. The CRN
+   model allocates a static /124 per VM out of a single host /64
+   (`StaticIPv6Allocator` vm-type hextet scheme, shared with the scheduler;
+   single-/64-per-host confirmed by the Scaleway experiments,
+   `2026-07-03-scaleway-ipv6-experiments.md`). A /124 can never be handed
+   out via SLAAC.
+2. Advertising the whole /64 on the tap instead would let the guest pick its
+   own interface ID, which never matches the daemon's derived address: the
+   published address would be fiction, the ndppd proxy range for the /124
+   would not cover the guest's real address (unreachable), and the vm-type
+   hextet scheme would be dead for v-programs.
+
+Decision (with Olivier, 2026-08-13): the CRN allocates and knows the exact
+address up front, and the guest must end up with precisely that address.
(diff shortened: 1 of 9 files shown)


foxpatch-aleph (commented):
Well-designed and implemented static IPv6 path for SNP guests. The DHCPv6 + RA approach is correct, the VmType::VProgram threading is consistent across create/adoption/runtime, and tests cover the key invariants including a red-on-revert adoption test. The firewall ICMPv6 rules are comprehensive. Two non-blocking observations: (1) the busybox udhcpc6 applet availability in nixpkgs is unverified in this PR — if missing, the feature silently no-ops (non-fatal but defeats the purpose); (2) the adoption path duplicates vm_type() logic inline, which is correct but a minor maintainability risk guarded by tests.

**`nix/initrd.nix`** (line 47): The default `pkgs.busybox` is used without verifying the `udhcpc6` applet is enabled. The design doc (section 2.4) notes a fallback `extraConfig` override with `CONFIG_UDHCPC6` if it is not present, but no such override is applied here. If the applet is absent, `udhcpc6` fails silently (caught by `||` in init.sh:70) and the guest stays IPv4-only — the feature no-ops rather than breaking, but the IPv6 reachability goal is silently unmet. Consider either adding the override preemptively or adding a CI check that `busybox --list | grep udhcpc6` succeeds in the built initrd.

**`rust/crates/supervisor-daemon/src/world.rs`** (line 572): The adoption path duplicates the `vm_type()` logic inline (`if qemu.snp().is_some() { VmType::VProgram } else { VmType::Instance }`) rather than reusing `VmEntry::vm_type()`, because the entry is not yet built at this point. This is correct, and the test `adopting_a_persisted_snp_config_gives_the_v_program_ipv6_hextet` guards against drift, but if the `vm_type()` logic ever gains another branch, this inline copy could silently diverge. A comment referencing the duplication would help future maintainers.

foxpatch-aleph (approved):
Well-designed and correctly implemented static IPv6 path for confidential V-PROGRAM guests. The dnsmasq stateful DHCPv6 + RA approach (M=1/A=0) correctly delivers the exact allocated /124-scheme address and default route without SLAAC, mirroring the shipped IPv4 path. VmType hextet values match Python exactly, the create and adoption paths are consistent, the cidr_prefix_len refactor is clean and fail-closed, and test coverage is comprehensive across both daemon and guest sides. The guest-side graceful fallback (bounded udhcpc6 with -n, IPv4-only on failure) and the dual-stack attest-agent bind are correct. Non-blocking observations: the busybox udhcpc6 applet availability relies on nixpkgs 24.11's default config without an explicit extraConfig override (acknowledged in the design doc), and the accept_ra=2 sysctl is set after interface-up but before udhcpc6 (correct since the solicit triggers the RA).

**`nix/flake.nix`** (line 130): The initrd uses `pkgs.busybox` from nixos-24.11 without an explicit `extraConfig` override to ensure `CONFIG_UDHCPC6=y`. The design doc (section 2.4) acknowledges this as an open item. While nixpkgs 24.11's default busybox config does include udhcpc6, a future nixpkgs bump could remove it. Consider adding an explicit `pkgs.busybox.override { extraConfig = "CONFIG_UDHCPC6=y"; }` for the initrd to make the dependency explicit, or at minimum a build-time assertion that the applet exists.

**`nix/udhcpc6.script`** (line 27): On a `renew` event, `ip -6 addr add` will fail with EEXIST if the address is already present (e.g., from a previous `bound`). This logs a WARNING, which could be noisy for what is a benign no-op. Consider adding `2>/dev/null || true` or checking with `ip -6 addr show` first to suppress the EEXIST case, matching the intent in the comment on line 24-25.

foxpatch-aleph (approved):
Well-designed and correctly implemented static IPv6 for SEV-SNP guests via per-tap DHCPv6 + RA. The VmType hextet derivation is consistent across create, adoption, and runtime paths. The dnsmasq args, guest init, firewall rules, and attest-agent dual-stack bind are all correct. Test coverage is thorough, including red-on-revert adoption tests. No correctness, security, or quality issues found.

odesenfans:
Re: the two review observations (both addressed, one commit each):

1. **busybox `udhcpc6` applet unverified** (nix/initrd.nix): addressed in 607cacd7, but with a build-time assertion rather than either suggested mitigation. A preemptive `extraConfig` override would force a from-source busybox rebuild (and a measurement shift) to defend against a problem the pinned nixpkgs does not have (verified: stock `pkgs.busybox --list` includes `udhcpc6`), and a CI check on the built initrd is heavier than needed. Instead the initrd now consumes busybox through a `runCommand` wrapper that fails the image derivation itself if the applet list lacks `udhcpc6`. Since nixpkgs is pinned by `flake.lock`, the only drift vector is a lock bump, which is exactly when the assertion re-runs. Negative-tested: with a bogus applet name the derivation fails with the diagnostic.

2. **Inline `vm_type()` duplication in the adoption path** (world.rs): addressed in 83c1dac0, going one step further than the suggested comment. The adoption arm binds the same `QemuVmConfig` type as `VmEntry::config`, so the predicate is now a shared helper, `VmType::of_qemu`, used by both `VmEntry::vm_type()` and the adoption path; a future branch cannot silently diverge. The create-time site in lifecycle.rs still derives from the request's TEE backend (no `QemuVmConfig` exists yet at that point) and keeps its equivalence comment. The existing red-on-revert adoption test still guards the behavior.

Also rebased onto current `dev` (the typed-error series #1088-#1103): `DhcpConfig::for_snp` now returns the typed `DhcpError`, with the `CidrNoPrefix`/`CidrBadPrefix` variants gaining a `family` field so the fail-closed prefix parsing covers both the IPv4 and IPv6 CIDRs. 339 daemon tests pass, clippy and fmt clean.

foxpatch-aleph (approved):
Well-designed and well-tested PR that adds static IPv6 for SEV-SNP guests via per-tap DHCPv6 + RA, mirroring the shipped IPv4 DHCP path. The VmType::of_qemu() predicate correctly shares the classification logic between create and adoption paths, maintaining the no-drift invariant. The bounded udhcpc6 client with -n flag handles version skew gracefully. ICMPv6 firewall rules are correctly scoped. Test coverage is comprehensive, including red-on-revert adoption tests. No correctness, security, or quality issues found.

**`rust/crates/aleph-attest-agent/src/main.rs`** (line 81): The dual-stack [::] bind relies on IPV6_V6ONLY=0 (Linux default) and the claim that actix-web does not set this sockopt. This is correct for current actix-web but could silently break if a future version sets IPV6_V6ONLY=1, making the agent IPv6-only. Consider adding a brief note or test that documents this dependency explicitly, though it is not blocking.

**`nix/initrd.nix`** (line 36): The checkedBusybox derivation copies the busybox binary into a new output path purely to assert the udhcpc6 applet exists. This works but adds an unnecessary copy step. An alternative would be using `pkgs.assertHook` or a `checkPhase` on the busybox derivation, but the current approach is clear and self-contained, so this is just a minor observation.

**`nix/udhcpc6.script`** (line 18): The deconfig action is a deliberate no-op (comment explains it would race IPv4). On a re-bind with the same address, `ip -6 addr add` returns EEXIST which is logged as a WARNING rather than silently ignored. This is slightly noisy on renews but functionally correct and matches the documented intent.

PR #1088, opened 2026-08-13
refactor(daemon): typed HugepagesError for hugepage reservation

odesenfans:
PR 1 of the typed-errors series (one PR per error type; plan: docs/plans/2026-08-13-supervisor-error-types-implementation.md, local).

Converts \`hugepages.rs\` from \`Result<_, String>\` to a strict thiserror enum:
- \`HugepagesError\` with \`Write\`/\`ReadBack\`/\`Parse\` struct variants carrying \`#[source]\` errors (io / ParseIntError)
- \`allocate_2m_pages_on_node\` returns \`Result<u32, HugepagesError>\`
- the fail-safe consumer in \`reserve_2m_hugepages_in\` logs with \`%error\`

Display strings are byte-identical to the previous \`format!\` messages; no behavior change. Pilot for the same pattern across the remaining daemon modules.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1089, opened 2026-08-13
refactor(daemon): typed ChecksError for startup preconditions

odesenfans:
PR 2 of the typed-errors series (stacked on #1088; will retarget to dev when it merges).

Converts \`checks.rs\` to a strict thiserror enum:
- \`ChecksError\` with 13 variants covering all 18 precondition-failure sites (\`FileNotFound\` and \`CommandMissing\` are shared across sites)
- \`DaemonError\` gains a transparent \`Checks(#[from] ChecksError)\` variant; the \`main.rs\` absorber becomes a bare \`?\`
- pinned message tests extended with variant \`matches!\` assertions

Display strings byte-identical (including the Python-parity trailing ", " on the two Qemu-support messages); no behavior change.

### DIFF
--- a/rust/crates/supervisor-daemon/src/checks.rs
+++ b/rust/crates/supervisor-daemon/src/checks.rs
@@ -13,9 +13,60 @@
 //! (SEV_CTL_PATH, the SEV/SEV-ES kernel modules) are ported, gated on
 //! ENABLE_CONFIDENTIAL_COMPUTING (increment 6).
 
-use std::path::Path;
+use std::path::{Path, PathBuf};
 
 use crate::config::Settings;
+
+/// Startup precondition failures from [`check`]. Display strings are
+/// identical to the pre-typed messages they replace (pinned by the Python
+/// `settings.check()` assert text and by tests below).
+#[derive(Debug, thiserror::Error)]
+pub enum ChecksError {
+    #[error("KVM not found on `/dev/kvm`.")]
+    KvmMissing,
+
+    #[error(
+        "EXECUTION_ROOT '{}' is too long for QEMU control sockets (UNIX socket paths are limited to 108 bytes)",
+        root.display()
+    )]
+    ExecutionRootTooLong { root: PathBuf },
+
+    #[error("File not found {}", path.display())]
+    FileNotFound { path: PathBuf },
+
+    #[error("Network interface is not specified")]
+    NoNetworkInterface,
+
+    #[error("Network interface {interface} does not exist")]
+    InterfaceMissing { interface: String },
+
+    #[error("invalid IPv4 pool {pool:?}")]
+    InvalidIpv4Pool { pool: String },
+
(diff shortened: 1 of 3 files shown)


foxpatch-aleph (approved):
Clean, mechanical conversion of string-based startup precondition errors to a typed thiserror enum. All 18 failure sites map correctly to 13 variants, Display strings are preserved byte-identical, the `#[from]` integration into `DaemonError` is sound, and tests are appropriately extended with `matches!` variant assertions alongside the existing Display-string checks. The only caller of `check()` is `main.rs`, so the return type change is safe. A few variants (InvalidIpv4Pool, CommandMissing, SevDisabled, SevEsDisabled, QemuDisabledConfidential, QemuDisabledGpu) lack direct test coverage, but this is expected since they require specific missing binaries or SEV hardware. Minor naming inconsistency between `ChecksError::InterfaceMissing` and the pre-existing `DaemonError::InterfaceNotFound` for the same concept, but this is non-blocking.

**`rust/crates/supervisor-daemon/src/checks.rs`** (line 41): Minor naming nit: `DaemonError` already has `InterfaceNotFound(String)` (error.rs:33) for the same concept. `ChecksError::InterfaceMissing` vs `DaemonError::InterfaceNotFound` is a slight inconsistency, though they live in separate enums so it's non-blocking.

**`rust/crates/supervisor-daemon/src/checks.rs`** (line 44): `InvalidIpv4Pool` has no test coverage. The existing networking test (`networking_requires_an_interface_and_a_sane_pool`) tests `NoNetworkInterface`, `InterfaceMissing`, and `PoolPrefixTooLong`, but doesn't exercise the invalid-pool-parse path (e.g. `ipv4_address_pool = "garbage"`). Consider adding a case to that test.

foxpatch-aleph (approved):
Clean, well-executed refactoring that converts the 18 string-based precondition failures in checks.rs into a 13-variant thiserror enum. Display strings are byte-identical to the original messages, verified against the Python conf.py source (including the deliberate trailing ", " on the two QEMU-support messages). The DaemonError::Checks(#[from] variant correctly replaces the .map_err(DaemonError::Internal) absorber in main.rs with a bare ?. Tests are extended with matches! variant assertions alongside the existing Display-string checks. The only minor gap is that InvalidIpv4Pool (the ok_or_else path for a malformed pool string) is not exercised by any test, but this is a non-blocking nit for a straightforward error-construction path.

**`rust/crates/supervisor-daemon/src/checks.rs`** (line 134): The `InvalidIpv4Pool` variant is the only ChecksError variant not exercised by any test. Consider adding a case in `networking_requires_an_interface_and_a_sane_pool` that sets `settings.ipv4_address_pool` to a malformed value (e.g. `"172.16.0.0"` without a `/`) and asserts `matches!(error, ChecksError::InvalidIpv4Pool { .. })`. Non-blocking nit.

PR #1090, opened 2026-08-13
refactor(daemon): typed WorldError for tap/IP derivation

odesenfans:
PR 3 of the typed-errors series (stacked on #1089).

Converts \`world.rs\` IP/tap derivation to a strict thiserror enum:
- \`WorldError\`, 15 variants for the 21 construction sites; the 8 near-duplicate CIDR messages collapse into 4 paired variants with a \`family\` field
- source-free by design (the original messages discard their parse errors)
- \`unique_vm_index\` / \`derive_tap_assignment\` and the private derivation helpers typed; lifecycle callers shimmed with \`.to_string()\` until the LifecycleError PR replaces the shims with \`#[from]\`

Display byte-identical (several of these strings reach the wire via RpcError::Internal); no behavior change.

### DIFF
--- a/rust/crates/supervisor-daemon/src/lifecycle.rs
+++ b/rust/crates/supervisor-daemon/src/lifecycle.rs
@@ -448,7 +448,8 @@ fn tap_assignment(state: &DaemonState, vm_id: &str) -> Result<TapAssignment, Str
         vm_id,
         vm_type,
         &mut ordinal,
-    )?;
+    )
+    .map_err(|error| error.to_string())?;
     world.ipv6_dynamic_ordinal = ordinal;
     if let Some(entry) = world.entries.get_mut(vm_id) {
         entry.ipv4 = Some(ipv4.clone());
@@ -2638,7 +2639,7 @@ fn create_vm_inner(
         });
         let vm_index = world
             .unique_vm_index(state.host.settings.start_id_index)
-            .map_err(RpcError::Internal)?;
+            .map_err(|error| RpcError::Internal(error.to_string()))?;
         let assignment = if state.host.settings.allow_vm_networking {
             let mut ordinal = world.ipv6_dynamic_ordinal;
             let pair = world::derive_tap_assignment(
@@ -2648,7 +2649,7 @@ fn create_vm_inner(
                 VmType::Instance,
                 &mut ordinal,
             )
-            .map_err(RpcError::Internal)?;
+            .map_err(|error| RpcError::Internal(error.to_string()))?;
             world.ipv6_dynamic_ordinal = ordinal;
             Some(pair)
         } else {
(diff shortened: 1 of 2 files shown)


foxpatch-aleph (approved):
A clean, well-executed refactoring that converts string-based errors in world.rs to a typed thiserror enum with 15 variants. Display strings are verified byte-identical to the pre-typed messages, including correct use of {:?} formatting for String fields that replaced &str. The family field design (using &'static str) elegantly collapses 8 near-duplicate CIDR messages into 4 paired variants. The tracing change from `error` to `%error` is correct and necessary — it ensures the Display impl is used instead of Debug now that the error type is an enum. The .to_string() shims in lifecycle.rs are appropriate as temporary bridges until the LifecycleError PR. Test improvements (matches! on specific variants) are a good upgrade from the old is_err() checks. No correctness, security, or logic issues found.

**`rust/crates/supervisor-daemon/src/world.rs`** (line 83): Minor observation: `VmHashTooShort` and `NonHexHashSlice` store `vm_hash`/`text` as `String`, but these fields are only ever used for Display output. Since the original `&str` was borrowed and the enum may outlive the borrow, `String` is the correct choice here — just noting it's intentional.

**`rust/crates/supervisor-daemon/src/world.rs`** (line 1323): Consider adding a `matches!` assertion for `IndexOutOfRange` here (as was done for `HostBitsSet` below) to strengthen the test: `assert!(matches!(ipv4_assignment("172.16.0.0/12", 24, 4096).unwrap_err(), WorldError::IndexOutOfRange { .. }))`. Not blocking — the current `is_err()` test still validates the error path.

foxpatch-aleph (approved):
Clean, well-executed refactoring that converts world.rs IP/tap derivation errors from stringly-typed Result<_, String> to a typed thiserror enum with 15 variants. Display strings are verified byte-identical to the original format! messages, including the clever family field that collapses 8 near-duplicate IPv4/IPv6 CIDR variants into 4 paired ones. The %error tracing change is necessary (uses Display instead of Debug for the new typed errors, preserving the same log output). The .to_string() shims in lifecycle.rs are correct temporary bridges to the String-based RpcError::Internal and the String-returning tap_assignment helper. All internal call sites propagate WorldError correctly via ?. Tests were appropriately updated to use matches! for variant-specific assertions where it adds value (HostBitsSet, ZeroSubnetPrefix) while leaving behavior-focused is_err() assertions unchanged. No correctness, security, or testing gaps found.

**`rust/crates/supervisor-daemon/src/world.rs`** (line 1409): Minor nit: `matches!(&result, ...)` followed by `result.unwrap_err()` works because `matches!` on a reference doesn't consume, but it means `result` is checked twice. Could destructure once with `let Err(error) = &result` then `assert!(matches!(error, WorldError::ZeroSubnetPrefix))` and `assert!(error.to_string().contains("/0"))`. Not blocking — the current code is correct.

PR #1092, opened 2026-08-13
refactor(daemon): typed PortsError for the port-mapping store

odesenfans:
PR 5 of the typed-errors series (stacked on #1091).

Converts the sqlite port-mapping store (\`ports.rs\`) to a strict thiserror enum:
- \`PortsError\`, 18 variants for the 26 construction sites, every variant keeping \`#[source] rusqlite::Error\` (no eager stringification)
- the pre-wrap raw rusqlite matches (\`QueryReturnedNoRows\`, "no such table" guards) are untouched and still fire before wrapping
- \`Exhausted\` renders the port-range message via the module consts
- lifecycle/main/world callers shimmed with \`.to_string()\` until the LifecycleError PR

Display byte-identical; no behavior change.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1093, opened 2026-08-13
refactor(daemon): typed CloudInitError for seed-image creation

odesenfans:
PR 6 of the typed-errors series (stacked on #1092).

Converts \`cloudinit.rs\` to a strict thiserror enum: \`CloudInitError\` with 5 variants (\`TempCreate\`/\`TempWrite\`/\`TempFlush\`/\`Run\` carrying \`#[source] io::Error\`, \`Failed{status, stderr}\`); \`CloudInitDrive::create_image\` returns \`Result<PathBuf, CloudInitError>\`; the single lifecycle caller is shimmed until the LifecycleError PR.

### DIFF
--- a/rust/crates/supervisor-daemon/src/cloudinit.rs
+++ b/rust/crates/supervisor-daemon/src/cloudinit.rs
@@ -17,6 +17,24 @@ use serde_json::{Value, json};
 
 use crate::tap::TapAssignment;
 
+/// Failures creating cloud-init seed images.
+#[derive(Debug, thiserror::Error)]
+pub enum CloudInitError {
+    #[error("cannot create a temp file: {source}")]
+    TempCreate { source: std::io::Error },
+    #[error("cannot write a temp file: {source}")]
+    TempWrite { source: std::io::Error },
+    #[error("cannot flush a temp file: {source}")]
+    TempFlush { source: std::io::Error },
+    #[error("cannot run cloud-localds: {source}")]
+    Run { source: std::io::Error },
+    #[error("cloud-localds failed ({status}): {stderr}")]
+    Failed {
+        status: std::process::ExitStatus,
+        stderr: String,
+    },
+}
+
 /// The qemu_build.py key merge: the spec's keys plus, when
 /// USE_DEVELOPER_SSH_KEYS is truthy, settings.DEVELOPER_SSH_KEYS.
 pub fn merged_ssh_keys(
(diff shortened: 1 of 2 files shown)


foxpatch-aleph (approved):
Clean, correct conversion from string errors to a typed thiserror enum. The 5 CloudInitError variants map 1:1 to the old format! strings, producing byte-identical Display output. The source field convention (matching sibling PRs in ports.rs and error.rs) correctly wires #[source] for error chaining. The lifecycle.rs shim (.map_err(|error| error.to_string())) preserves the exact error message until the LifecycleError PR lands. No behavior change, no security concerns. The test coverage is thin (only TempCreate is tested) but the Display strings are trivially verifiable by inspection.

**`rust/crates/supervisor-daemon/src/cloudinit.rs`** (line 414): The test only exercises the TempCreate variant. Consider adding assertions for the remaining variants (TempWrite, TempFlush, Run, Failed) — or at minimum for the Failed variant, whose Display format ({status} interpolation of ExitStatus) is the most complex. Not blocking.

**`rust/crates/supervisor-daemon/src/cloudinit.rs`** (line 417): Minor: the test creates an io::Error with message "no perms" but only uses io_error.kind(), discarding the message. Simpler would be: let kind = io::ErrorKind::PermissionDenied; let temp_create = CloudInitError::TempCreate { source: kind.into() }; assert_eq!(temp_create.to_string(), format!("cannot create a temp file: {}", kind));

foxpatch-aleph (approved):
Clean, correct refactoring from string-based errors to a typed thiserror enum. The Display output is byte-identical to the previous format! strings, the lifecycle shim via .map_err(|error| error.to_string()) is an appropriate temporary bridge, and #[source] correctly preserves the error chain. The only minor observation is that the new test covers only 1 of 5 variants and uses an indirect io_error.kind().into() pattern that is slightly confusing, but neither issue is blocking.

**`rust/crates/supervisor-daemon/src/cloudinit.rs`** (line 414): The test only covers the `TempCreate` variant. Consider adding assertions for the remaining 4 variants (`TempWrite`, `TempFlush`, `Run`, `Failed`), especially `Failed` since its format string is the most complex (uses `{status}` and `{stderr}` fields rather than `{source}`).

**`rust/crates/supervisor-daemon/src/cloudinit.rs`** (line 419): The test creates `io_error` but then uses `io_error.kind().into()` to construct the `TempCreate` variant — a different `io::Error` than the one created on line 417. This works because `Display` for `io::Error::from(ErrorKind)` matches `Display` for `ErrorKind`, but it's slightly misleading. Consider simplifying to `let io_err = io::Error::from(io::ErrorKind::PermissionDenied);` and using `io_err` directly in both the variant construction and the assertion.

PR #1094, opened 2026-08-13
refactor(daemon): typed ConfigWriteError for controller-config persistence

odesenfans:
PR 7 of the typed-errors series (stacked on #1093).

Converts the controller-config writer half to a strict thiserror enum: \`ConfigWriteError\` (TempCreate/Write/Chmod/Persist/RemoveArtifact), \`save_controller_config\` and \`remove_controller_configuration\` typed, the NotFound tolerance on artifact removal untouched, \`ConfigParseError\` untouched (revisited in the LifecycleError PR). Lifecycle callers shimmed.

### DIFF
--- a/rust/crates/supervisor-daemon/src/controller_config.rs
+++ b/rust/crates/supervisor-daemon/src/controller_config.rs
@@ -39,6 +39,24 @@ use serde::Deserialize;
 #[error("{0}")]
 pub struct ConfigParseError(pub String);
 
+/// Failures persisting the controller configuration to disk.
+#[derive(Debug, thiserror::Error)]
+pub enum ConfigWriteError {
+    #[error("cannot create a temp config file: {source}")]
+    TempCreate { source: std::io::Error },
+    #[error("cannot write the controller config: {source}")]
+    Write { source: std::io::Error },
+    #[error("cannot chmod the controller config: {source}")]
+    Chmod { source: std::io::Error },
+    #[error("cannot move the controller config in place: {source}")]
+    Persist { source: tempfile::PersistError },
+    #[error("cannot remove the VM artifact {path}: {source}")]
+    RemoveArtifact {
+        path: std::path::PathBuf,
+        source: std::io::Error,
+    },
+}
+
 /// `ControllerSettings`: the node-settings slice a controller reads.
 /// Defaults mirror conf.py so keys absent from an old dump fall back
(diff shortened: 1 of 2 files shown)


foxpatch-aleph (approved):
Clean, mechanical conversion from String errors to a typed thiserror enum. The five-variant ConfigWriteError correctly captures each failure site with proper source-error chaining. Display output is byte-identical to the old format! strings, including the PathBuf case (thiserror uses Display, matching path.display()). The NotFound tolerance in remove_controller_configuration is preserved. All three lifecycle call sites are correctly shimmed with .to_string(). The new test covers 4 of 5 variants; the Persist variant is a trivial passthrough and its omission is non-blocking.

**`rust/crates/supervisor-daemon/src/controller_config.rs`** (line 950): Minor: the Persist variant is the only ConfigWriteError case not covered by this test. Consider adding a case for it (e.g., using tempfile::PersistError::from(io::Error) or constructing one from a failed persist). Non-blocking since it's a trivial passthrough.

foxpatch-aleph (approved):
Clean, well-structured refactoring that converts string-based error handling in the controller-config writer to a typed thiserror enum. All five error variants map 1:1 to the original format! strings, confirming the claimed byte-identical Display output. The NotFound tolerance on artifact removal is correctly preserved. All three lifecycle.rs call sites are properly shimmed via .map_err(|error| error.to_string()) or .map_err(|error| RpcError::Internal(error.to_string())), leveraging the existing From<String> for RpcError impl. The test covers four of five variants with correct assertions. Minor nits: the `path.clone()` on line 570 is an unnecessary allocation (path could be moved directly into the error since it's an early return from an owning loop), and the `Persist` variant is untested.

**`rust/crates/supervisor-daemon/src/controller_config.rs`** (line 570): `path.clone()` is unnecessary here — `path` is owned by the loop iteration and this is an early return, so `path` can be moved directly into the struct: `path: path,`. The `std::fs::remove_file(&path)` call only borrows `path` for its duration and the borrow has ended by the time the block body executes. Minor nit, does not affect correctness.

**`rust/crates/supervisor-daemon/src/controller_config.rs`** (line 949): The `Persist` variant is the only one not covered by this test. Consider adding a test case for it, even if constructing a `tempfile::PersistError` requires a helper. Alternatively, a brief comment noting why it's omitted would help future readers.

PR #1095, opened 2026-08-13
refactor(daemon): typed DnsError for DNS discovery

odesenfans:
PR 8 of the typed-errors series (stacked on #1094).

Converts DNS discovery in \`net.rs\` to a strict thiserror enum:
- \`DnsError\` (Resolvectl/ResolvectlFailed/UnexpectedOutput/ReadResolvConf/NoResolver); the old nested string-wrap becomes a boxed source chain
- \`DaemonError\` gains a transparent \`Dns(#[from] DnsError)\` variant; the \`service.rs\` absorber becomes a bare \`?\`
- new test skips gracefully on hosts without \`/usr/bin/resolvectl\` (the exact path production invokes)

Display byte-identical; no behavior change.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1096, opened 2026-08-13
refactor(daemon): typed LogsError for the journalctl log source

odesenfans:
PR 9 of the typed-errors series (stacked on #1095). First trait conversion: \`LogSource::{read_history, follow}\` now return \`Result<_, LogsError>\` across \`JournalctlLogSource\`, \`StaticLogSource\`, and the test source; \`follow_child\` typed; the three \`service.rs\` absorbers keep identical wire text. New regression test for the \`NoStdout\` path.

### DIFF
--- a/rust/crates/supervisor-daemon/src/logs.rs
+++ b/rust/crates/supervisor-daemon/src/logs.rs
@@ -28,6 +28,30 @@
 
 use std::process::Command;
 
+/// Failures reading or following the journal via `journalctl`.
+#[derive(Debug, thiserror::Error)]
+pub enum LogsError {
+    #[error("failed to run journalctl: {source}")]
+    Journalctl { source: std::io::Error },
+
+    #[error("journalctl failed ({status}): {stderr}")]
+    JournalctlFailed {
+        status: std::process::ExitStatus,
+        stderr: String,
+    },
+
+    #[error("failed to run journalctl --follow: {source}")]
+    Follow { source: std::io::Error },
+
+    #[error("journalctl --follow has no stdout")]
+    NoStdout,
+
+    /// A fabricated error for test `LogSource` fakes that need to report a
+    /// specific failure without shelling out to a real `journalctl`.
+    #[error("{0}")]
+    Injected(String),
+}
+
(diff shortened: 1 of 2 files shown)


foxpatch-aleph (approved):
Clean, well-executed typed-error migration for the journalctl log source. The LogsError enum variants produce Display output that is byte-identical to the old String-based error messages (verified for all four variants). All three LogSource implementations (JournalctlLogSource, StaticLogSource, and the ControlledFollowSource test fake) are updated consistently. The three service.rs call sites correctly bridge LogsError to DaemonError::Internal via error.to_string(), preserving wire format. The new NoStdout regression test is thorough, asserting both the variant identity and the Display string. The Injected variant is unused but well-documented as a future test helper — a minor nit, non-blocking.

**`rust/crates/supervisor-daemon/src/logs.rs`** (line 52): The `Injected(String)` variant is declared but never used anywhere in the codebase. While the doc comment explains its intended purpose, adding it preemptively is mild over-engineering. Consider removing it until a test actually needs it, to keep the enum minimal.

foxpatch-aleph (approved):
A clean, correct trait error-type conversion. The `LogsError` enum's thiserror Display strings are byte-identical to the old `format!`-based `String` errors, verified across all four variants (Journalctl, JournalctlFailed, Follow, NoStdout). All three `LogSource` implementations (JournalctlLogSource, StaticLogSource, ControlledFollowSource) are updated consistently. The three service.rs absorber sites correctly flatten via `error.to_string()` into `DaemonError::Internal`, preserving wire-text equivalence. The new `NoStdout` regression test is well-targeted, asserting both the variant and the Display string. The only observation is that the `Injected(String)` variant is dead code (never constructed or matched anywhere), but this is a non-blocking nit since it's likely scaffolded for a later PR in the series.

**`rust/crates/supervisor-daemon/src/logs.rs`** (line 52): The `Injected(String)` variant is defined and documented but never constructed or matched anywhere in this codebase. Consider adding it in the PR that first uses it, or remove it here to avoid carrying dead code. Non-blocking — clearly scaffolded for future test fakes in the typed-errors series.

PR #1097, opened 2026-08-13
refactor(daemon): typed TapError for tap-device management

odesenfans:
PR 10 of the typed-errors series (stacked on #1096). \`TapBackend::{create_tap, delete_tap}\` return \`Result<(), TapError>\` (Run/Command/Injected variants) across \`IpCommand\` and \`FakeTapBackend\` (injection API unchanged); \`run_ip\` typed; the \`IpFailure\` stderr classifier untouched; lifecycle callers shimmed and warn sites log with \`%error\`.

Display byte-identical; no behavior change.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1098, opened 2026-08-13
refactor(daemon): typed NftError for the nftables executor

odesenfans:
PR 11 of the typed-errors series (stacked on #1097). \`NftExecutor::{list_ruleset, run_commands}\` and \`initialize_ipv4_commands\` return \`Result<_, NftError>\` (9 variants, io/serde_json sources kept); \`NftCli\` and \`StaticRuleset\` converted with the injection API unchanged; the pinned \`NoBaseChainFound\` struct and its pure-layer functions are untouched; log-and-continue sites use \`%error\`.

### DIFF
--- a/rust/crates/supervisor-daemon/src/lifecycle.rs
+++ b/rust/crates/supervisor-daemon/src/lifecycle.rs
@@ -542,7 +542,8 @@ fn nft_remove_redirect(
 pub fn initialize_nftables(state: &DaemonState) -> Result<(), String> {
     let prefix = chain_prefix(state);
     let ruleset = nft::fetch_ruleset(&*state.nft);
-    let commands = nft::initialize_ipv4_commands(&ruleset, prefix)?;
+    let commands =
+        nft::initialize_ipv4_commands(&ruleset, prefix).map_err(|error| error.to_string())?;
     nft::run_commands_logged(&*state.nft, &commands);
     if !state.host.settings.ipv6_forwarding_enabled {
         return Ok(());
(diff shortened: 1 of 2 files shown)


foxpatch-aleph (approved):
Clean, well-executed refactor that replaces ad-hoc String errors with a typed NftError enum. All 9 variants have Display strings that are byte-identical to the old format! outputs, preserving the no-behavior-change guarantee. The map_err patterns correctly preserve source errors via thiserror's {source} forwarding. The tracing sites correctly switch to %error (Display) from bare error (Debug), which is the idiomatic tracing syntax and produces cleaner structured logs. The new test verifies both the NftError::Injected variant and its Display string. No correctness, security, or coverage concerns.

**`rust/crates/supervisor-daemon/src/nft.rs`** (line 41): Minor pre-existing observation (not a regression): the `Run` variant covers two distinct call sites — `Command::output()` in `list_ruleset` (line 838) and `Command::spawn()` in `run_commands` (line 864). The Display message "cannot run nft: {source}" is identical in both cases, so a consumer of the error cannot distinguish which operation failed. This mirrors the old `format!("cannot run nft: {error}")` behavior, so it is not a change in semantics, but if future debugging requires distinguishing these, splitting into `RunList` and `RunApply` (or adding a context field) would help.

foxpatch-aleph (approved):
Clean, mechanical refactor that replaces stringly-typed errors with a 9-variant NftError enum. All Display strings are byte-identical to the old format! outputs. The tracing macro changes from `error` (Debug) to `%error` (Display) are correct and necessary for readable log output with typed errors. The new test covers the Injected variant. The bridge at the lifecycle.rs call site via map_err(|error| error.to_string()) preserves the existing Result<(), String> signature. No behavior change.

**`rust/crates/supervisor-daemon/src/nft.rs`** (line 71): The `Injected` variant is documented as test-only but lives in the production error enum. Since `StaticRuleset` (the only producer) is itself test-only, this is pragmatic, but consider whether a test-only executor should share the production error type or use a separate test error. Not blocking.

PR #1099, opened 2026-08-13
refactor(daemon): typed DhcpError for the SNP DHCP backend

odesenfans:
PR 12 of the typed-errors series (stacked on #1098). \`DhcpConfig::for_snp\` and \`DhcpBackend::{start, stop}\` return typed \`DhcpError\` (7 variants); \`SystemdRunDhcp\` and \`FakeDhcpBackend\` converted with the injection API unchanged; the \`is_absent_unit_error\` "unit already gone" tolerance still fires on raw stderr before wrapping; equality-based tests rewritten as \`matches!\` (no PartialEq on the enum).

### DIFF
--- a/rust/crates/supervisor-daemon/src/dhcp.rs
+++ b/rust/crates/supervisor-daemon/src/dhcp.rs
@@ -28,6 +28,8 @@
 //! (`aleph-vm-dhcp-{vm_hash}.service`) via `systemd-run`, keyed by vm_hash so
 //! it survives a daemon restart and is torn down by unit name.
 
+use std::num::ParseIntError;
+use std::path::PathBuf;
 use std::process::Command;
 
 use crate::tap::TapAssignment;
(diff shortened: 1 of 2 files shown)


foxpatch-aleph (approved):
Clean, mechanical refactor that replaces `String`-based errors in the DHCP backend with a typed `DhcpError` enum following the exact same pattern already established by `TapError` in `tap.rs`. All seven `Display` implementations are byte-identical to the old `format!` strings, the error flow through `lifecycle.rs` is preserved (DhcpError → Display → String → RpcError::Internal), the `is_absent_unit_error` tolerance still fires on raw stderr before wrapping, and the `matches!`-based test rewrites are correct since `DhcpError` deliberately doesn't derive `PartialEq`. The `%dhcp_error` tracing change from implicit `Debug` to explicit `Display` is an improvement that produces the same output. No behavior change.

**`rust/crates/supervisor-daemon/src/dhcp.rs`** (line 622): Minor: now that the error is typed, `for_snp_rejects_a_cidr_without_a_prefix` could assert the specific variant — `assert!(matches!(result, Err(DhcpError::CidrNoPrefix { .. })))` for the no-slash case and `Err(DhcpError::CidrBadPrefix { .. })` for the non-numeric prefix case. This would guard against the two branches accidentally swapping their error variants. Not a blocker since the old test also only checked `is_err()`.

foxpatch-aleph (approved):
A clean, well-structured refactoring that replaces stringly-typed errors in the DHCP module with a 7-variant `DhcpError` enum using thiserror, consistent with the crate's existing pattern. All Display impls are byte-identical to the old `format!` strings, the `is_absent_unit_error` tolerance still fires on raw stderr before wrapping, and all call sites in `lifecycle.rs` are correctly updated with appropriate error conversions (`.to_string()` for the `Result<(), String>` closure, `RpcError::Internal(error.to_string())` for the RPC boundary, and `%dhcp_error` for tracing Display). The `FakeDhcpBackend` injection API is preserved, and tests are correctly rewritten as `matches!`. No behavior change.

**`rust/crates/supervisor-daemon/src/dhcp.rs`** (line 610): Now that the error variants are typed, this test could be strengthened to assert the specific variants: `Err(DhcpError::CidrNoPrefix { .. })` for the no-slash case and `Err(DhcpError::CidrBadPrefix { .. })` for the non-numeric prefix case, rather than just `.is_err()`. Non-blocking nit.

PR #1102, opened 2026-08-13
refactor(daemon): typed FirecrackerError for the program launcher

odesenfans:
PR 15 of the typed-errors series (stacked on #1101). The 10 String-erroring launcher helpers (\`jailman staging, msgpack validation, journald stream, jail prep, config write, spawn\`) return \`Result<_, FirecrackerError>\` (19 variants); \`ProgramBootError\`/\`ChannelError\` gain thiserror derives (attribute-only; wire-visible payloads still render \`{0}\`); \`boot()\`'s trait signature is unchanged so no external ripple beyond the one lifecycle shim.

Display byte-identical (incl. the pinned Python-parity rootfs message); no behavior change.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1103, opened 2026-08-13
refactor(daemon): typed BackupError for backup/restore disk tooling

odesenfans:
PR 16 of the typed-errors series (stacked on #1102). \`DiskTools::{convert_compressed, virtual_size}\` return \`Result<_, BackupError>\` (18 variants); \`check\` deliberately keeps \`QemuImgError\` (its CheckFailed/Unavailable split maps to different wire codes) and that enum gains an attribute-only thiserror derive; the variant-collapsing site becomes a transparent \`#[from]\` fold; the \`run_backup\` sink stores \`error.to_string()\` into \`pb::BackupInfo.error_message\` unchanged.

Display byte-identical; no behavior change. Last leaf module — next PR is the LifecycleError keystone.

### DIFF
--- a/rust/crates/supervisor-daemon/src/backup.rs
+++ b/rust/crates/supervisor-daemon/src/backup.rs
@@ -52,15 +52,97 @@ pub const BACKUP_DOWNLOAD_CHUNK_BYTES: usize = 1024 * 1024;
 /// A `qemu-img check` outcome. The Python `verify_qemu_disk` raises
 /// `subprocess.CalledProcessError` when the image is invalid, which
 /// `restore_from_image` catches to map a bad upload to InvalidBackendError
-/// (a 400) while anything else (qemu-img missing) stays INTERNAL.
-#[derive(Debug)]
+/// (a 400) while anything else (qemu-img missing) stays INTERNAL. Kept as
+/// its own type (not folded into [`BackupError`]) because that variant
+/// split is load-bearing at the lifecycle call sites: `CheckFailed` ->
+/// InvalidBackend, `Unavailable` -> Internal.
+#[derive(Debug, thiserror::Error)]
 pub enum QemuImgError {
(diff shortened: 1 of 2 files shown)


foxpatch-aleph (approved):
A clean, behavior-preserving refactor that replaces `String` error returns with a typed `BackupError` enum (18 variants) across the backup disk-tooling surface, while deliberately keeping `QemuImgError` on `DiskTools::check` because its `CheckFailed`/`Unavailable` split is load-bearing at the `lifecycle.rs::restore_from_image` call site (mapping to `InvalidBackend` vs `Internal` respectively). Display strings are byte-identical to the pre-refactor `format!` messages they replace, verified by cross-checking each variant's `#[error(...)]` against the original `format!` and by the strengthened test assertion `assert_eq!(job.error_message, "fake check failed")` which exercises the `#[error(transparent)]` + `#[from]` fold path end-to-end. The `?`-based fold of `QemuImgError` into `BackupError::QemuImg` at `run_backup_locked` is equivalent to the old variant-collapsing `match` since both branches returned the inner `String`. The `start_backup` and `restore_from_image` `virtual_size` call sites correctly switch from `From<String>` to `error.to_string()` to preserve the Display. New test coverage for `ArchiveAdd` is meaningful. No correctness, security, or quality issues found.

foxpatch-aleph (approved):
A clean, well-documented typed-error refactor. BackupError's 18 variants map 1:1 to the pre-existing format! strings with byte-identical Display output. The #[from] fold on QemuImgError is correct, and the decision to keep QemuImgError separate (for the CheckFailed/Unavailable split at lifecycle call sites) is well-justified. The tracing::error! change to %error and the error.to_string() sink are both correct for the new typed error. Tests cover the ArchiveAdd variant and the transparent QemuImg fold path. No correctness, security, or quality issues found.

PR #1091, opened 2026-08-13
refactor(daemon): typed QmpError for the QMP/QGA clients

odesenfans:
PR 4 of the typed-errors series (stacked on #1090).

Converts the daemon's QMP/QGA clients (\`qmp.rs\`) to a strict thiserror enum:
- \`QmpError\`, 23 variants (12 QGA-side, 11 QMP-side), io/serde_json sources kept as \`#[source]\` fields
- the nested "no QMP greeting" string-wrap becomes a real source chain (\`NoGreeting { source: Box<QmpError> }\`)
- \`confidential.rs\` callers shimmed with \`.to_string()\` (replaced by \`#[from]\` in the LifecycleError PR); \`backup.rs\` fsfreeze warn sites log with \`%error\`

Display byte-identical (these strings reach the wire via RpcError::Internal); no behavior change.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1104, opened 2026-08-13
refactor(daemon): LifecycleError + typed RpcError; no untyped errors reach the wire

odesenfans:
PR 17 of the typed-errors series (stacked on #1103) — the keystone.

- \`LifecycleError\`: 9 own variants (byte-identical templates) + 11 transparent \`#[from]\` variants composing the leaf enums from PRs 1-16
- the 9 String-erroring lifecycle fns and 4 closures are fully typed; this series' \`.to_string()\` shims inside lifecycle.rs collapse to \`?\`
- \`RpcError\` gains its thiserror derive (\`{0}\` everywhere; the gRPC layer destructures payloads, so wire text provably cannot shift — incl. the empty-string \`MicroVmInit\` Python parity)
- **\`impl From<String> for RpcError\` is deleted**; \`git grep 'Result<.*, String>' rust/crates/supervisor-daemon/src/\` returns zero hits

No behavior change; the compiler now enforces that no untyped error reaches the wire.

### DIFF
--- a/rust/crates/supervisor-daemon/src/lifecycle.rs
+++ b/rust/crates/supervisor-daemon/src/lifecycle.rs
@@ -26,41 +26,125 @@ use crate::controller_config::{
 use crate::firecracker::{ProgramBootError, ProgramBootRequest};
 use crate::service::DaemonState;
 use crate::tap::TapAssignment;
-#[cfg(test)]
-use crate::units::UnitsError;
-use crate::units::{self, controller_unit_name};
+use crate::units::{self, UnitsError, controller_unit_name};
 use crate::world::{self, AttachedGpu, ProgramEntry, VmEntry, VmTimes, VmType, now_ns};
 use crate::{checks, cloudinit, dhcp, nft, ports};
 
 /// The closed error vocabulary slice these RPCs can produce, mapped in
 /// service.rs onto the same gRPC status codes and ErrorDetail trailers as
 /// src/aleph/vm/supervisor/grpc_server.py.
-#[derive(Debug)]
+///
+/// Every payload IS the message that reaches the client, so `Display` is
+/// `{0}` verbatim on every variant: `MicroVmInit(String::new())` renders
+/// empty, matching the Python MicroVMFailedInitError whose `str()` is empty.
+#[derive(Debug, thiserror::Error)]
 pub enum RpcError {
     /// VmNotFoundError: message is the vm_id itself.
+    #[error("{0}")]
     NotFound(String),
     /// VmAlreadyExistsError.
+    #[error("{0}")]
     AlreadyExists(String),
     /// InsufficientResourcesError.
+    #[error("{0}")]
     InsufficientResources(String),
     /// InvalidBackendError.
+    #[error("{0}")]
     InvalidBackend(String),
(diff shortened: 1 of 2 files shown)


foxpatch-aleph (approved):
Well-executed keystone PR in the typed-errors series. LifecycleError's 9 own variants produce byte-identical Display strings compared to the old format! calls, and 11 transparent #[from] variants correctly delegate to leaf error types. The From<LifecycleError> for RpcError conversion exactly replaces the deleted From<String>. The tracing macro changes from `error` to `%error` are correct and necessary (Display instead of Debug for the new typed errors). The UnitsError import correctly moves out of #[cfg(test)] since it's now used in non-test code. Remaining map_err(...to_string()) calls are all in code paths dealing with modules not yet typed in this stacked series. Existing test coverage exercises the error paths and validates message content. No behavior change, no security concerns.

foxpatch-aleph (approved):
This PR is a clean, well-executed keystone for the typed-errors series. The LifecycleError enum's 9 own variants produce byte-identical messages to the format! strings they replace (verified for all 9), the 11 transparent #[from] variants correctly compose leaf error types with #[error(transparent)] so their Display passes through unchanged, and the From<LifecycleError> for RpcError impl replicates exactly what the deleted From<String> for RpcError did. The tracing field changes from `error` to `%error` are necessary and correct — without `%`, tracing would use Debug instead of Display on the new typed errors, producing wrong output. The Numa variant's #[error("{0:#}")] correctly preserves anyhow's alternate Display chain. The ControllerStart variant's Box<LifecycleError> is sound (the inner Display is the same text the old String carried). The UnitsError import was correctly promoted from #[cfg(test)] to unconditional since LifecycleError::Units(#[from] UnitsError) requires it at compile time. The remaining .map_err(|error| RpcError::Internal(error.to_string())) calls in RPC-surface functions (start_vm_execution, reboot_vm, create_program_vm, etc.) are expected — those functions return RpcError directly and call leaf modules whose error types don't have From into RpcError; converting them is out of scope for this PR.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 3162): Not a blocker, but a note for follow-up: the `create_program_vm` boot closure still returns `Result<_, RpcError>` and uses `.map_err(|error| RpcError::Internal(error.to_string()))` for tap/ndp/dhcp errors. Converting this closure to `LifecycleError` (like the `create_vm_inner` closure was) would eliminate the last remaining `.to_string()` shims in this area, but this is correctly out of scope for this PR.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 3441): The `initialize_nftables(state).map_err(|error| RpcError::Internal(format!("Failed to initialize network: {error}")))` call is correct — it wraps the LifecycleError with an additional prefix before converting to RpcError, so using the `From` impl (which would drop the prefix) would be wrong here. Just confirming this is intentional.

odesenfans:
Heads-up: this keystone PR depends on all 13 daemon leaf PRs (#1091-#1103), so it cannot be unstacked. Its base is a holder branch (od/errors-lifecycle-base) so the diff below shows exactly this PR's two commits. Once the leaf PRs are merged into dev, this branch gets rebased onto dev and retargeted; the holder branch will be deleted then. Draft until that rebase.

PR #1105, opened 2026-08-13
refactor(daemon): anyhow at the process top level

odesenfans:
PR 18 of the typed-errors series (stacked on #1104). \`run()\` returns \`anyhow::Result<()>\`; the spawn-join absorbers carry their old inline text via \`anyhow!\`; the ports absorbers collapse to plain \`?\`. The top-level sink deliberately logs plain Display (not \`{error:#}\`): this crate's error Displays inline their sources by design, and a chain-walking render would double-print them — documented at the sink.

Rendered text byte-identical everywhere; no behavior change.

### DIFF
--- a/rust/crates/supervisor-daemon/src/main.rs
+++ b/rust/crates/supervisor-daemon/src/main.rs
@@ -76,6 +76,14 @@ fn main() -> ExitCode {
         .with_writer(std::io::stderr)
         .init();
 
+    // Plain Display, not `{error:#}`: this crate's error Displays are
+    // self-contained by design (they inline their sources to keep
+    // byte-identical text across the type-strictness series), but they also
+    // expose those same sources via `Error::source()` (thiserror auto-links
+    // any field named `source`). anyhow's alternate `{:#}` walks that chain
+    // and appends each source again, so switching this sink to `{:#}`
+    // without first stripping embedded sources from the Display templates
+    // would double-print (e.g. "cannot open X: msg: msg").
     match run(&cli) {
         Ok(()) => ExitCode::SUCCESS,
         Err(error) => {
@@ -85,7 +93,7 @@ fn main() -> ExitCode {
     }
 }
 
-fn run(cli: &Cli) -> Result<(), DaemonError> {
+fn run(cli: &Cli) -> anyhow::Result<()> {
     let mut settings = Settings::from_env()?;
     // --socket overrides the settings-derived path, after settings
     // resolution, like `args.socket or default_socket_path()` in daemon.py.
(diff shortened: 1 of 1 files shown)


foxpatch-aleph (approved):
Clean refactor that replaces manual `.map_err(|e| DaemonError::Internal(e.to_string()))` wrappers with direct `?` propagation through `anyhow::Result<()>`, preserving full error chains instead of flattening them to strings. The spawn-join absorbers correctly use `anyhow::anyhow!` for their inline context text. The top-level sink comment at lines 79-86 is thorough and accurately documents why plain Display (`%error`) is used instead of alternate Display (`{:#}`): thiserror's `source` fields are inlined in Display templates AND exposed via `Error::source()`, so `{:#}` would double-print them. All error types involved (PortsError, ChecksError, DaemonError) derive `thiserror::Error`, so they satisfy anyhow's blanket `From` impl. No behavior change; rendered text is byte-identical.

odesenfans:
Heads-up: depends on #1104 (LifecycleError) and #1092 (PortsError); stays stacked on od/errors-lifecycle. Will be rebased onto dev together with #1104 once the leaf PRs merge. Draft until then.

foxpatch-aleph (approved):
This PR correctly replaces `run()`'s return type with `anyhow::Result<()>`, collapsing the verbose `.map_err(|error| DaemonError::Internal(error.to_string()))` absorbers to plain `?` (for ports/checks errors) and `anyhow!` (for spawn-join errors that carry inline context text). The error Display output is byte-identical: direct `?` conversions preserve the domain error's Display (same as the old `error.to_string()`), and `anyhow!` preserves the `format!` text that `DaemonError::Internal` carried. The top-level sink correctly uses plain `%error` (Display) rather than `{error:#}` (alternate), with a thorough comment explaining why: thiserror auto-links fields named `source` via `Error::source()`, so anyhow's chain-walking `{:#}` would double-print sources that are already inlined in the Display templates. `DaemonError` import remains needed for `spawn_signal_task`. No behavior change, no security concerns.

foxpatch-aleph (approved):
Clean, correct refactor that changes run()'s return type from Result<(), DaemonError> to anyhow::Result<()>. All called functions return errors that derive thiserror::Error (PortsError, ChecksError, DaemonError) or std::io::Error, all of which implement std::error::Error + Send + Sync + 'static, making them compatible with anyhow's ? operator. The anyhow::anyhow! macro correctly preserves the exact Display text of the old DaemonError::Internal(format!(...)) pattern for the JoinError cases. The top-level sink uses plain %error (Display) rather than {error:#} (alternate), which would double-print sources since this crate's error Displays intentionally inline their sources. The comment at lines 79-86 thoroughly documents this decision. No behavior change, no tests needed.

PR #1106, opened 2026-08-13
refactor(controller): typed QmpError for the QMP escalation client

odesenfans:
PR 19 of the typed-errors series (stacked on #1105). The controller's QMP shutdown-escalation client goes typed: 5 fns return \`Result<_, QmpError>\` (14 variants, io/serde_json sources, the nested greeting wrap becomes a boxed source chain); the \`quit\`-tolerance match is untouched and \`qemu.rs\`'s warn interpolation needed zero edits (Display-based).

### DIFF
--- a/rust/crates/supervisor-controller/src/qmp.rs
+++ b/rust/crates/supervisor-controller/src/qmp.rs
@@ -22,6 +22,61 @@ use std::os::unix::net::UnixStream;
 use std::path::Path;
 use std::time::{Duration, Instant};
 
+/// The closed error vocabulary for the QMP escalation client. Every consumer
+/// (`qemu.rs::send_qmp`) only ever `tracing::warn!`s the rendered message, so
+/// these are display-only, but the `source` fields keep the original error
+/// typed rather than eagerly stringified.
+#[derive(Debug, thiserror::Error)]
+pub enum QmpError {
+    #[error("cannot connect to QMP: {source}")]
+    Connect { source: std::io::Error },
+
+    #[error("cannot set the QMP write timeout: {source}")]
+    WriteTimeout { source: std::io::Error },
+
+    #[error("cannot clone the QMP socket: {source}")]
+    Clone { source: std::io::Error },
+
+    #[error("no QMP greeting: {source}")]
+    NoGreeting { source: Box<QmpError> },
+
+    #[error("cannot send the QMP command {command}: {source}")]
+    Send {
+        command: String,
+        source: std::io::Error,
+    },
+
+    #[error("invalid QMP response: {source}")]
+    InvalidResponse { source: serde_json::Error },
+
+    #[error("QMP error for {command}: {error}")]
+    Protocol {
+        command: String,
+        error: serde_json::Value,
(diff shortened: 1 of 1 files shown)


foxpatch-aleph (approved):
A faithful, mechanical refactoring from String-based errors to a typed QmpError enum with 14 variants. All Display strings are byte-identical to the deleted format! templates (verified by the new test and by static inspection of each variant). The NoGreeting variant correctly boxes the recursive QmpError to break the size cycle. The quit-tolerance match is unchanged. The sole consumer (qemu.rs::send_qmp) only renders the error via tracing::warn!, so no consumer edits are needed. No behavior change, no security concerns. One trivial nit: a tautological matches! assertion in the test.

**`rust/crates/supervisor-controller/src/qmp.rs`** (line 250): The assertion `matches!(QmpError::LineTooLong, QmpError::LineTooLong)` is a tautology — it can never fail regardless of any change. It doesn't test anything. Consider removing it; the `to_string()` assertion on the next line is the meaningful one for this variant.

foxpatch-aleph (approved):
A clean, mechanical refactor from String errors to a typed QmpError enum with 14 variants. All Display outputs are byte-identical to the old format!() strings, the sole consumer (qemu.rs::send_qmp) only renders via tracing::warn! so behavior is unchanged, and the quit-tolerance match is correctly left variant-agnostic. The source fields preserve typed errors instead of eagerly stringifying them, which is the goal of the typed-errors series.

**`rust/crates/supervisor-controller/src/qmp.rs`** (line 250): The `assert!(matches!(QmpError::LineTooLong, QmpError::LineTooLong))` is a tautology — it always passes regardless of the variant's content. The `assert_eq!` on the next line is the actual test. Consider removing the tautological assertion.

foxpatch-aleph (approved):
Clean, well-executed typed-error refactor. All four error enums (QmpError, ChecksError, HugepagesError, WorldError) have Display strings byte-identical to the original String templates, verified by the format strings and pinned by tests. The NoGreeting boxing, the #[error(transparent)] + #[from] on DaemonError::Checks, the tracing %error changes, and the lifecycle.rs .to_string() boundary conversions are all correct. No behavior change, no security issues. Test coverage is adequate with both variant matching and Display assertions for the key variants.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 429): The `tap_assignment` function still returns `Result<_, String>` and stringifies `WorldError` via `.to_string()` at line 452. This is an expected transitional state in the stacked typed-errors series, but worth noting that `WorldError` is still eagerly stringified at this boundary — a future PR could type the `RpcError` variants to carry `WorldError` directly (or via `#[from]`) to preserve the source chain end-to-end.

**`rust/crates/supervisor-controller/src/qmp.rs`** (line 251): The `matches!(QmpError::LineTooLong, QmpError::LineTooLong)` assertion is a tautology — it always passes regardless of the variant's fields. It adds no test value beyond the `to_string()` assertion on the next line. Consider removing it or replacing with a more meaningful assertion on a different variant.

PR #1110, opened 2026-08-14
refactor(daemon): From impls route leaf errors into RpcError; drop the residual shims

odesenfans:
Follow-up 1/3 to the typed-errors series. Adds 12 \`From<LeafError> for RpcError\` impls (body: \`Internal(error.to_string())\` — wire text structurally unchanged) and converts all 39 residual \`.map_err(|error| RpcError::Internal(error.to_string()))\` shims to bare \`?\`. The load-bearing \`QemuImgError\` match sites and every context-adding \`format!\` site are untouched.

### DIFF
--- a/rust/crates/supervisor-daemon/src/backup.rs
+++ b/rust/crates/supervisor-daemon/src/backup.rs
@@ -966,15 +966,9 @@ pub fn start_backup(
     // space of the backup directory. InsufficientResources on shortfall.
     let mut needed: u64 = 0;
     for (_, source) in &disk_paths {
-        needed = needed.saturating_add(
-            state
-                .disk_tools
-                .virtual_size(source)
-                .map_err(|error| RpcError::Internal(error.to_string()))?,
-        );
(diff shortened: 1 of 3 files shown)


foxpatch-aleph (approved):
This is a clean, mechanical refactoring that replaces 39 `.map_err(|error| RpcError::Internal(error.to_string()))` shims with bare `?` operators, backed by 12 new `From<LeafError> for RpcError` impls whose bodies are byte-for-byte identical to the shims they replace. I verified that every leaf error type used at a converted call site has a corresponding `From` impl, that no live shims remain (only doc-comment occurrences), and that context-adding `format!` sites are correctly left untouched. The direct `From<LeafError> for RpcError` impls coexist without coherence conflicts with the existing `From<LeafError> for LifecycleError` `#[from]` impls — Rust's `?` resolves to the direct impl in `Result<_, RpcError>` functions and to the `#[from]` impl in `Result<_, LifecycleError>` functions. No wire-text changes, no security implications, no logic changes.

PR #1111, opened 2026-08-14
refactor(daemon): typed failure injection; drop the Injected variants

odesenfans:
Follow-up 2/3 to the typed-errors series (stacked on #1110).

Deletes all five \`Injected(String)\` variants: the fake backends' failure-injection API is now typed — a test injects the real variant it simulates (via a builder closure, so the error enums stay \`!Clone\` and source identity is never lossily reconstructed on a public type). Assertions follow the three-category rule from the design discussion: wire-pinning production-Display tests untouched; sentinel-propagation checks keep equivalent strength; circular fake-boundary string checks became variant assertions.

### DIFF
--- a/rust/crates/supervisor-daemon/src/dhcp.rs
+++ b/rust/crates/supervisor-daemon/src/dhcp.rs
@@ -245,11 +245,6 @@ pub enum DhcpError {
 
     #[error("systemctl stop {unit} failed: {stderr}")]
     StopFailed { unit: String, stderr: String },
-
-    /// A fabricated error for test `DhcpBackend` fakes that need to report a
-    /// specific failure without shelling out to real `systemd-run`/`systemctl`.
-    #[error("{0}")]
-    Injected(String),
 }
 
 /// Whether a `systemctl stop` stderr describes a unit systemd never knew (it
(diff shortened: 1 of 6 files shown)


foxpatch-aleph (approved):
Clean, well-documented refactor that replaces all five Injected(String) test-fake variants with typed builder closures, preserving the !Clone invariant on error enums. The Debug derive removal on FakeDhcpBackend and FakeTapBackend is correct (Box<dyn Fn> is not Debug) and no code depends on it. The as_deref-to-as_ref migration in StaticRuleset is handled correctly. All test assertions were properly migrated from string matching to variant matching. The ndppd FailingEdge correctly uses std::io::Error::other (available in edition 2024). No remaining references to Injected exist outside an unrelated comment in cpuid.rs.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 5636): Minor nit: the argv string \"tuntap add name vmtap mode tap\" is generic (no device number), while the production IpCommand would produce \"tuntap add name vmtap3 mode tap\" with the actual device name. The dedicated test in tap.rs:377 correctly uses the specific name. Not a bug since these tests don't assert on argv, but using a more representative string would improve fidelity if assertions are ever added later.

foxpatch-aleph (approved):
Clean, well-executed refactor that replaces all five Injected(String)/Injected{needle} test-only error variants with typed builder closures (Box<dyn Fn() -> ErrorType>). The approach correctly preserves the !Clone invariant on error enums (real io::Error sources stay non-reconstructed) while letting tests inject actual production variants. All call sites in dhcp.rs, tap.rs, nft.rs, ndppd.rs, and lifecycle.rs are consistently updated, no Injected references remain, the Debug derive removal is justified (Box<dyn Fn()> has no Debug), the let-chain in nft.rs run_commands correctly borrows from the MutexGuard, and test assertions are strengthened from string-only to variant+Display checks.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 5636): Minor nit: the injected argv uses "tuntap add name vmtap mode tap" (no numeric suffix), while the real IpCommand would produce e.g. "tuntap add name vmtap4 mode tap". The tests only pattern-match on TapError::Command { .. } without checking argv, so this is harmless, but matching the real format would make the fakes slightly more faithful if a future test ever asserts on argv content.

foxpatch-aleph (approved):
Clean, well-documented refactoring that replaces all five Injected(String) fake-backend error variants with typed builder closures, preserving the !Clone invariant on error types. All test sites are correctly updated with typed error injection and variant-level assertions. The Debug derive removal from fake backends is necessary and safe. No correctness, security, or coverage issues found.

**`rust/crates/supervisor-daemon/src/nft.rs`** (line 1162): ExitStatus::default() produces a zero/no-exit status rather than a realistic failure exit code. The test only asserts on the stderr portion via contains(), so it passes, but using ExitStatus::from_raw(1) (or a similar failed-status value) would be marginally more faithful to what a real nft failure looks like. Non-blocking nit.

foxpatch-aleph (approved):
Clean, well-documented refactor that replaces all five Injected(String) test-only error variants with a builder-closure pattern, producing real typed errors at each injection site. The approach correctly preserves the !Clone invariant on production error types by calling the closure fresh on each failure rather than storing a clonable value. All call sites in lifecycle.rs and the per-module unit tests are updated consistently. The as_deref() to as_ref() change in nft.rs is correct for the new tuple type. Display-string assertions are retained where they verify wire-pinning and simplified to contains() where exact matches would be brittle. No remaining references to Injected exist anywhere in the crate. The removal of Debug derives on FakeDhcpBackend and FakeTapBackend is an unavoidable and correct consequence of storing Box<dyn Fn(...)>, and no code depends on those Debug impls.

PR #1112, opened 2026-08-14
refactor(daemon): drop dead error variants and unify the panic-prefix literal

odesenfans:
Follow-up 3/3 to the typed-errors series (stacked on #1111). Drops the never-constructed \`LifecycleError::Backup\`; routes the program-path panic through \`LifecycleError::Panic\` so the "CreateVm panicked: " template exists once (wire text byte-identical, pinned by a new test); moves two last-use \`path\` clones; fixes the cloudinit Display test (real end-to-end custom-message assertion + coverage for the two-field \`Failed\` variant).

### DIFF
--- a/rust/crates/supervisor-daemon/src/cloudinit.rs
+++ b/rust/crates/supervisor-daemon/src/cloudinit.rs
@@ -413,15 +413,23 @@ mod tests {
     #[test]
     fn cloud_init_error_messages_are_correct() {
         use std::io;
-
-        let io_error = io::Error::new(io::ErrorKind::PermissionDenied, "no perms");
+        use std::os::unix::process::ExitStatusExt;
 
         let temp_create = CloudInitError::TempCreate {
-            source: io_error.kind().into(),
+            source: io::Error::new(io::ErrorKind::PermissionDenied, "no perms"),
         };
         assert_eq!(
             temp_create.to_string(),
-            format!("cannot create a temp file: {}", io_error.kind())
+            "cannot create a temp file: no perms"
+        );
+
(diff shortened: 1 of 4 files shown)


foxpatch-aleph (approved):
Clean follow-up that removes a genuinely dead error variant (LifecycleError::Backup — confirmed no references anywhere), routes the create_program_vm panic path through LifecycleError::Panic with byte-identical wire text (pinned by a new focused unit test), improves the cloudinit Display test to assert real end-to-end output instead of just io::ErrorKind, and eliminates two last-use path clones. All changes are correct, the From<BackupError> for RpcError leaf conversion is preserved for restore_backup, and the ExitStatus::from_raw(1 << 8) usage in the new test is the correct Unix raw wait status for exit code 1.

foxpatch-aleph (approved):
Clean, well-focused follow-up that drops the dead `LifecycleError::Backup` variant (never constructed, and `BackupError` already has a direct `From` impl to `RpcError`), unifies the `"CreateVm panicked: "` prefix into the single `LifecycleError::Panic` `#[error]` attribute, and eliminates two unnecessary last-use `path.clone()` calls. The cloudinit Display test is substantially improved — replacing the old `io_error.kind()` assertion (which tested the wrong thing) with a real end-to-end custom-message assertion, plus new coverage for the two-field `Failed` variant. The new `create_program_vm_panic_wire_text_is_unchanged` test is an excellent regression pin that verifies the full conversion chain produces byte-identical wire text. All changes are correct, no security concerns, and the test comments are thorough.

PR #1113, opened 2026-08-14
refactor(daemon): drop dead error variants and unify the panic-prefix literal

odesenfans:
Replaces #1112, which GitHub auto-marked as merged (and auto-deleted its branch) when a branch-pointer mistake briefly pushed its head commit onto its base branch. Nothing actually landed in dev; identical content, same branch name, still stacked on #1111 — merge #1111 first, then this.

Original description: drops the never-constructed \`LifecycleError::Backup\`; routes the program-path panic through \`LifecycleError::Panic\` so the "CreateVm panicked: " template exists once (wire text byte-identical, pinned by a new test); moves two last-use \`path\` clones; fixes the cloudinit Display test.

### DIFF
--- a/rust/crates/supervisor-daemon/src/cloudinit.rs
+++ b/rust/crates/supervisor-daemon/src/cloudinit.rs
@@ -413,15 +413,23 @@ mod tests {
     #[test]
     fn cloud_init_error_messages_are_correct() {
         use std::io;
-
-        let io_error = io::Error::new(io::ErrorKind::PermissionDenied, "no perms");
+        use std::os::unix::process::ExitStatusExt;
 
         let temp_create = CloudInitError::TempCreate {
-            source: io_error.kind().into(),
+            source: io::Error::new(io::ErrorKind::PermissionDenied, "no perms"),
         };
         assert_eq!(
             temp_create.to_string(),
-            format!("cannot create a temp file: {}", io_error.kind())
+            "cannot create a temp file: no perms"
+        );
+
+        let failed = CloudInitError::Failed {
+            status: std::process::ExitStatus::from_raw(1 << 8),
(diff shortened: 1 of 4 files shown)


foxpatch-aleph (approved):
Clean refactoring PR that removes a dead error variant (LifecycleError::Backup, never constructed), routes the create_program_vm panic path through LifecycleError::Panic to unify the "CreateVm panicked: " prefix literal, eliminates two unnecessary last-use path clones, and fixes the cloudinit Display test to use real error values instead of ErrorKind-derived ones. All changes are verified correct: the wire text is byte-identical (pinned by a new test), the direct From<BackupError> for RpcError impl remains for callers that need it, and both path moves are genuine last uses. The new cloudinit test also adds coverage for the Failed variant.

foxpatch-aleph (approved):
Clean, well-motivated refactoring PR. The dropped `LifecycleError::Backup` variant was dead code — `BackupError` already has its own direct `From` impl for `RpcError`. Routing the `create_program_vm` panic through `LifecycleError::Panic` is wire-identical to the old inline `format!` because the `thiserror` Display template is `"CreateVm panicked: {0}"` and `From<LifecycleError> for RpcError` calls `to_string()`. The new test pins this invariant. The cloudinit test fix is a genuine improvement — the old test used `ErrorKind::into()` which silently dropped the custom message; the new test uses a real `io::Error` with a message and adds coverage for the `Failed` variant. Both clone elisions are correct last-use moves. No correctness, security, or testing gaps found.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 5170): Minor nit: `message.clone()` is unnecessary here — `message` is a `String` and isn't used after this line, so `LifecycleError::Panic(message)` would suffice. Non-blocking.

PR #1117, opened 2026-08-17
fix(nix): bump guest-image nixpkgs from EOL 24.11 to nixos-26.05

odesenfans:
## What

Re-pins the measured-guest flake's nixpkgs input from `nixos-24.11` (EOL, frozen at a June 2025 rev) to current stable `nixos-26.05` and refreshes `flake.lock`.

## Why

The EOL pin could never advance again, freezing the guest kernel at 6.6.94 and edk2 below 202505. This is the enabling step for the guest-kernel CVE fixes in the stacked PRs above this one; on its own it moves the same 6.6 kernel line to 6.6.151 and edk2 to 202602 (covers the OVMF fixes the SEV-SNP audit calls for, findings E-1/E-2).

## Consequences

The launch measurement changes (kernel/OVMF bits change), `measurement.hex` is recomputed by the flake. There is no backward compatibility to preserve yet; deployed image tarballs must be rebuilt once the stack lands.

Stack: PR 1/5 (base of `od/nix-sev-snp-measure-unpin` → `od/nix-guest-kernel-6-12` → `od/nix-guest-no-ia32` → `od/nix-kernel-floor-assert`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1118, opened 2026-08-17
chore(nix): drop the manual sev-snp-measure 0.0.12 pin

odesenfans:
## What

Removes the `overridePythonAttrs` pin of sev-snp-measure to 0.0.12: nixos-26.05 ships exactly that release (the one that fixed the 0.0.11 measurement-calculation bug), so the override now refetches what the channel already provides.

## Why

Dead weight after #1117; one less hash to maintain. Measurement output is unchanged (same upstream source).

Stack: PR 2/5 on #1117.

### DIFF
--- a/nix/flake.lock
+++ b/nix/flake.lock
@@ -17,16 +17,16 @@
     },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1751274312,
-        "narHash": "sha256-/bVBlRpECLVzjV19t5KMdMFWSwKLtb5RyXdjz3LJT+g=",
+        "lastModified": 1786943417,
+        "narHash": "sha256-b4qgjdFtlz5TAZ1Hi7DFJeqX3sjaDUnrwr9OO+O1rM0=",
         "owner": "NixOS",
         "repo": "nixpkgs",
-        "rev": "50ab793786d9de88ee30ec4e4c24fb4236fc2674",
+        "rev": "0dd31db7e6dbf9ce05697c4545f6fe01accec994",
         "type": "github"
       },
       "original": {
         "owner": "NixOS",
-        "ref": "nixos-24.11",
+        "ref": "nixos-26.05",
         "repo": "nixpkgs",
         "type": "github"
       }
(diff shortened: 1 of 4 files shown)


foxpatch-aleph (approved):
Clean removal of the manual sev-snp-measure 0.0.12 override. The nixos-26.05 channel (already declared at flake.nix:19) ships exactly 0.0.12, so the override was redundant. The variable is still defined as an alias to the nixpkgs package and consumed identically, so the measurement pipeline is unchanged. The replacement comment is excellent — it preserves the historical context and warns about the reproducibility invariant for future channel bumps. The pyproject.toml pin (0.0.12) and the Python wrapper docstring remain consistent.

**`src/aleph/vm/vprogram/measurement.py`** (line 3): Minor nit (outside this diff): the docstring says "matching the nix pin in nix/flake.nix" but the flake no longer has an explicit pin — it now uses the nixpkgs default. Consider updating to something like "matching the nixpkgs-provided version" in a follow-up.

foxpatch-aleph (approved):
Clean, correct removal of a now-redundant overridePythonAttrs pin: nixos-26.05 ships sev-snp-measure 0.0.12, the same version the flake was manually pinning, so the override was dead weight. The replacement comment is more valuable than the override it replaces — it preserves the historical reason for the pin and warns that future channel bumps can silently change baked measurements. No new external inputs (one fetchFromGitHub + hash removed, reducing supply-chain surface), no logic changes, no tests needed. Minor non-blocking nit: the docstring in src/aleph/vm/vprogram/measurement.py:3 still says "matching the nix pin in nix/flake.nix" — wording is now slightly stale since the nix side no longer pins by hand, though the version-match it describes still holds.

**`src/aleph/vm/vprogram/measurement.py`** (line 3): Non-blocking nit: the docstring says "matching the nix pin in nix/flake.nix". After this PR the nix side no longer pins by hand — it uses the nixpkgs default. Consider rewording to something like "matching the nixpkgs-shipped version (nix/flake.nix)" so the comment stays accurate after the override is removed.

PR #1121, opened 2026-08-17
feat(nix): assert the SNP guest kernel security floor at evaluation

odesenfans:
## What

Makes flake evaluation fail with the CVE rationale if the resolved guest kernel drops below 6.18 (the pinned LTS line; 6.18.0 already carries every fix on the audit list). Tested both ways: 6.18.44 passes; flipping the package back to `linuxPackages_6_12` fails evaluation with the floor message.

## Why

The guest kernel regressed silently once already: an EOL nixpkgs pin froze it at 6.6.94 while the #VC-hardening floor moved to 6.9+, and nothing noticed until a security audit. Encoding the floor in the flake means a channel bump or package-line change can never lower it unnoticed. Closes the audit's "no CI check asserts the guest kernel floor" note (C-1).

Stacked on #1122 (6.18 bump), top of the stack.

### DIFF
--- a/nix/kernel.nix
+++ b/nix/kernel.nix
@@ -9,7 +9,22 @@
 # until ~end of 2027. Newer non-LTS lines (7.x) add nothing from that list
 # and would EOL under us within months, recreating the frozen-EOL-kernel
 # failure mode the nixos-26.05 re-pin fixed.
-pkgs.linuxPackages_6_18.kernel.override {
+let
+  base = pkgs.linuxPackages_6_18.kernel;
+in
+
+# Fail evaluation, not review, if a future channel or package change lowers
+# the kernel below the security floor above. 6.18.0 already carries the full
+# set, so the floor is simply the pinned LTS line itself.
+assert lib.assertMsg (lib.versionAtLeast base.version "6.18")
+  ''
+    SNP guest kernel ${base.version} is below the security floor (6.18):
+    it may lack the HECKLER/WeSee #VC-handler hardening (CVE-2024-25742/
+    -25743/-25744, Linux >= 6.9) or the pvalidate cache-line-eviction fix
+    (CVE-2025-38560). Do not ship it to confidential guests.
+  '';
+
+base.override {
   structuredExtraConfig = with lib.kernel; {
     # SEV-SNP guest support (mkForce to override base config "m" → "y")
     AMD_MEM_ENCRYPT = lib.mkForce yes;


foxpatch-aleph (approved):
The kernel security floor assertion is correct, well-structured, and serves its purpose: making flake evaluation fail (not just review) if the guest kernel drops below the 6.12.93 CVE-hardening floor. The two-branch version comparison correctly handles the semantic subtlety that 6.13+ is acceptable despite component-wise comparison quirks, with a defensive hasPrefix guard. The assertMsg pattern is idiomatic Nix, the error message is clear and actionable with CVE references, and the assertion runs at evaluation time (making it inherently self-testing). Manual testing was done both ways per the PR description. No bugs, no security issues, no missing test coverage for an evaluation-time assertion.

**`nix/kernel.nix`** (line 13): Minor comment clarity nit: the parenthetical says "a 6.13+ kernel is fine even though 6.13.0 < 6.12.93 componentwise", but component-wise comparison actually gives 6.13.0 > 6.12.93 (13 > 12 at the second component). The real subtlety is that you want to express "6.12.x with x >= 93, OR any 6.13+ line" as a single floor, and a single `versionAtLeast base.version "6.12.93"` alone would also accept 6.13+ (since 6.13 > 6.12.93), making the first branch technically redundant. The two-branch approach is still the clearest expression of intent and the `hasPrefix` guard adds defense-in-depth, so the code is correct — just the explanatory comment is slightly misleading about the componentwise comparison direction.

foxpatch-aleph (approved):
The evaluation-time kernel floor assertion is correct across all edge cases and addresses the audit finding (C-1) about missing CI checks for the guest kernel security floor. The OR-based version check with the hasPrefix guard is slightly more complex than a single versionAtLeast call would need to be, and the comment about 6.13.0 being componentwise less than 6.12.93 is imprecise (Nix compares the second component first, where 13 > 12), but neither issue affects correctness. The manual testing described in the PR (6.12.103 passes, 6.6 fails) confirms both directions work. No security concerns — the change is purely defensive against silent kernel regressions.

**`nix/kernel.nix`** (line 18): The OR structure with `hasPrefix` is correct but more complex than needed. `lib.versionAtLeast base.version "6.12.93"` alone would pass for 6.13+ kernels too (Nix compares the second component 13 > 12 before the third), so the first branch is only necessary because the `hasPrefix "6.12."` guard in branch 2 restricts it to 6.12.x. A simpler equivalent: `assert lib.assertMsg (lib.versionAtLeast base.version "6.12.93") ''...''`. Non-blocking — the current form is correct and arguably more self-documenting.

**`nix/kernel.nix`** (line 14): The comment says "6.13.0 < 6.12.93 componentwise" but Nix's `compareVersions` compares left-to-right, so 6.13.0 is actually *greater than* 6.12.93 (second component 13 > 12). The OR is needed not because of a comparison limitation but because the `hasPrefix "6.12."` guard in branch 2 excludes 6.13+ kernels. Consider rewording to clarify that the OR exists to handle the hasPrefix restriction, not a version-comparison edge case.

foxpatch-aleph (approved):
The core changes are correct and well-documented: the nixpkgs channel bump from nixos-24.11 to nixos-26.05 brings kernel 6.18 (with HECKLER/WeSee #VC-handler hardening and CVE-2025-38560 fixes), the assert at kernel.nix:17 correctly fails evaluation if the kernel drops below 6.18, the removed modules (dax.ko, libcrc32c.ko) are consistent with kernel 6.18's built-in DAX and crc32c, the IA32_EMULATION/X86_X32_ABI=n hardening is sound for a 64-bit-only guest, and the sev-snp-measure override removal is correct since nixos-26.05 ships 0.0.12 natively. The kernel.modules output path change in initrd.nix is correct for the new nixpkgs. Two non-blocking observations: the PR description is stale (references a 6.12.93 floor and 6.12.103 testing, while the code pins 6.18), and the comment at kernel.nix:42-44 about the DAX kconfig conflict is outdated since DAX is now built-in on 6.18.

**`nix/kernel.nix`** (line 42): This comment about the DM_VERITY → BLK_DEV_DM → DAX kconfig conflict is now stale: nixos-26.05 builds CONFIG_DAX=y (built-in), so DAX is no longer a module dependency and the conflict with generate-config.pl no longer applies. Consider updating or removing the DAX reference to avoid confusing future maintainers.

**`nix/kernel.nix`** (line 17): The assert correctly guards the floor. Note that the PR description references a 6.12.93 floor and testing with 6.12.103 / linuxPackages_6_6, but the actual code pins linuxPackages_6_18 and asserts >= 6.18 — the description appears to be from an earlier iteration and should be updated to match the code.

foxpatch-aleph (approved):
The PR adds a Nix `assert` using `lib.assertMsg` to fail flake evaluation if the guest kernel version drops below 6.18, with a clear CVE-rationale error message. The logic is correct: `lib.versionAtLeast base.version "6.18"` properly compares kernel version strings (e.g., "6.18.44" passes, "6.12.93" fails), `lib` is available via the function header `{ pkgs, lib, ... }` and `callPackage` injection, and the `assert ...; base.override { ... }` expression structure is syntactically valid Nix. The error message interpolates the actual version and names the specific CVEs, making it actionable. This is a sound approach to preventing silent kernel security regressions at build time.

foxpatch-aleph (approved):
Clean, idiomatic Nix assertion that enforces the SNP guest kernel security floor (6.18) at evaluation time, preventing silent regressions from channel bumps or package-line changes. The version check uses lib.versionAtLeast on the pre-override kernel package, the error message is informative with CVE references, and the placement before base.override is correct since the override only touches kconfig, not the version. Addresses audit finding C-1 directly.

Issue #158, opened 2022-03-28
Enable cgroup v2 in Firecracker Jailer


hoh:
Requires Ubuntu >= 21.10

https://rootlesscontaine.rs/getting-started/common/cgroup2/

PR #1127, opened 2026-08-19
feat: SEV-SNP confidential instances

odesenfans:
Adds support for **SEV-SNP confidential instances**: an `INSTANCE` message with `environment.trusted_execution.mode = "sev_snp"` boots a persistent QEMU SEV-SNP guest whose LUKS-encrypted user rootfs is unlocked by the owner over direct attested TLS, after the client verifies the launch measurement. This backports the aleph-cvm confidential-instance model onto the new SEV-SNP + V-PROGRAM machinery already on `dev`, but as an INSTANCE (not a new message type).

Design: `docs/plans/2026-08-18-snp-confidential-instances-design.md`. Plan: `docs/plans/2026-08-18-snp-confidential-instances-implementation.md` (both untracked on `dev`, which dropped `docs/plans/`; happy to commit them if wanted).

Related tickets: ALEPH-XXX

## Changes

Four increments, each self-contained and independently reviewed:

**1. Guest image + owner-auth (nix, aleph-tee, attest-agent, attest-cli)**
- `aleph-tee::owner_auth`: the EIP-191 owner-auth payload contract (`aleph-snp-inject-secret-v1 | sha384(server TLS key) | sha256(canonical secrets JSON)`), sign/recover/verify. Binding to the per-boot TLS key gives channel binding and replay protection in one.
- attest-agent gains `--owner`: with it, `inject-secret` requires an owner-signed envelope (`{secrets, signature}`) and allows authenticated re-injection (a typo'd passphrase is retryable); without it, today's unauthenticated one-shot flow is unchanged (v-program images).
- attest-cli gains `--owner-key` / `--owner-key-file` (the file form keeps the key out of argv); redirects are now disabled on the attested client so the signed envelope can only bind the key of the peer that actually verified.
- A dedicated **instance guest image**: LUKS `init-instance.sh` (fail-closed on a non-`luks=1 owner=` cmdline, indefinite passphrase wait, wrong-passphrase retry, no platform firewall), `cryptsetup`/`dm-crypt` in the initrd, built from a shared `init-common.sh` prologue factored out of the v-program `init.sh` (verbatim, statement-for-statement). Distinct bundle from the v-program image so neither product's TCB carries the other's code.

**2. Supervisor (proto, daemon, controller)**
- Additive proto: opaque `TeeConfig.kernel_cmdline` (the daemon passes it to QEMU verbatim, never parses it) and explicit `VmSpec.vm_type` (the daemon stops inferring V-PROGRAM from the SNP backend).
- Daemon: a second, fail-closed SNP arm. `kernel_cmdline` set = LUKS instance (writable qcow2 rootfs, no verity, no hash-tree volume, rejects a stray `.roothash` sidecar); `kernel_cmdline` empty = today's verity path byte-for-byte. Every malformed combination is an explicit `InvalidBackend`.
- Controller: `build_snp_argv` parameterizes the rootfs disk (raw+ro for verity, qcow2+writable for LUKS); a half-populated override is refused before launch.

**3. Agent (Python)**
- `snp_instance_launch.build_snp_instance_spec`: fetches + validates the `aleph-instance-runtime/1` manifest, stages the bundle (via a `snp_staging` module factored out of `vprogram_launch`), renders the measured `owner=` cmdline from the message sender, and builds the SNP spec. Rejection ladder (attestation_port set, host not SNP-capable, policy ≥ 2^32, non-EVM sender) runs before any I/O.
- `run.py` routes `mode=sev_snp` instances through this path: boots immediately (no `awaiting_confidential_init`), forwards SSH + the manifest attestation port, cleans staging on every failure path. `translate.py` gains a fail-loud guard so an SNP instance can never fall through to the SEV path.
- Example `examples/instance_confidential_snp/build_luks_rootfs.sh` for building a LUKS2 rootfs.

**4. Test fixtures + E2E**
- A dropbear test rootfs (nix) and the pin runbook (`docs/plans/2026-08-18-snp-instance-pin-runbook.md`).
- The aleph-testnets E2E scenario lives on branch `od/snp-instance-e2e` (separate repo, **not pushed** — pushing auto-triggers a paid SNP CI run before the artifact pins exist).

## How to test

Unit suites pass on the tip: Rust (`aleph-tee`, `aleph-attest-agent`, `aleph-attest-cli`, `supervisor-daemon`, `supervisor-controller`) and the Python SNP-instance / proto / v-program-model suites. End-to-end validation is the aleph-testnets scenario on the SNP host, gated on the pin uploads in the runbook.

## Notes

- **Depends on** `aleph-message` PR #158 (`od/vprogram-schema`) being merged and released; `trusted_execution.mode`/`runtime`/`measurements` ship there, currently a git-ref pin.
- **Merges into `dev`**, not `main`.
- Not hardware-validated until the aleph-testnets E2E loop runs green with the three pins uploaded (`SNP_IMAGE_HASH` refreshed, `SNP_INSTANCE_IMAGE_HASH`, `SNP_INSTANCE_TEST_ROOTFS_HASH`).
- The refreshed v-program image measurement changes (atomic-secret-write fix + shared-init refactor + kernel-bump base); the runbook covers re-pinning and re-running the existing v-program E2E.
- Known out-of-scope / deferred (tracked): pyaleph persistence + CCN measurement validation; reject empty secret values; document/reject `workload_roothash` on the opaque arm; pre-existing staging-leak on the expiry/tasks delete paths (affects v-programs too); a design-doc note on the clone-VM residual (a host can boot a measured-identical clone against a disk copy and have the owner unlock it, equivalent in power to the accepted rollback residual).
- `packaging/Makefile`: no new Python runtime deps were added, so no mirror needed; the new Rust dev-deps (`k256`, `sha3`) affect only the workspace, not the deb.


foxpatch-aleph (approved):
Well-designed and carefully implemented SEV-SNP confidential instances PR. The EIP-191 owner-auth crypto is correct, the redirect-disable fix prevents cross-instance signature replay, the atomic secret writes prevent partial-write races, and fail-closed patterns are consistently applied across daemon, controller, and agent. Test coverage is comprehensive including security-critical redirect-attack and payload-limit regressions. No blocking issues found.

**`nix/init-instance.sh`** (line 47): The zeroize_passphrase function uses `bs=1` for dd, writing one byte at a time. For typical LUKS passphrases (short strings) this is fine, but if larger secrets are ever injected through this path, consider using `bs=$size` for efficiency. Non-blocking — current code is correct, just not optimal for large values.

**`rust/crates/aleph-attest-agent/src/secrets.rs`** (line 286): In `WriteMode::Overwrite`, `rename()` atomically replaces any existing directory entry at `final_path` — including a planted symlink — without following it. This is safe on Linux, but worth a one-line comment noting that `rename()` does not follow symlinks at the target, for future readers who might worry about the missing symlink check in Overwrite mode.

**`src/aleph/vm/agent/snp_instance_launch.py`** (line 124): The policy check `int(trusted_execution.policy) >= 2**32` rejects policies above 32 bits. This is correct for the current SNP policy field size, but consider documenting that this cap exists because the proto field is a string and the daemon's `parse_sev_policy` expects a u32, so the guard prevents an opaque overflow before it reaches the daemon.

odesenfans:
Superseded by a 4-PR stack, split for reviewability (each PR is a single coherent surface, reviewed in order):

- 1/4 owner-auth (attestation crypto): #1128 -> `dev`
- 2/4 LUKS guest image + manifest (nix + Python): #1129 -> #1128
- 3/4 supervisor opaque-cmdline SNP arm (proto/daemon/controller): #1130 -> #1129
- 4/4 agent launch path (activates the feature): #1131 -> #1130

Same final tree as this PR (verified identical). Branch `od/snp-instances` kept as the integration reference.

PR #1128, opened 2026-08-19
feat(snp-instances): owner-authenticated secret injection (attestation)

odesenfans:
**Stack 1/4 for SEV-SNP confidential instances.** Base: `dev`. Reviewer lens: the attestation crypto surface, in isolation.

This PR adds the owner-authentication mechanism and nothing user-visible yet: it is entirely dormant until a guest image (PR 2) launches the agent with `--owner`.

Related tickets: ALEPH-XXX

## Changes
- `aleph-tee::owner_auth`: the EIP-191 owner-auth payload contract, defined once and consumed everywhere: `aleph-snp-inject-secret-v1 | sha384(server TLS pubkey) | sha256(canonical secrets JSON)`, with sign / recover / verify. Binding the signature to the guest's per-boot TLS key gives channel binding and replay protection in one (no nonce state).
- **attest-agent** gains `--owner`: with it, `POST /confidential/inject-secret` requires an owner-signed envelope `{secrets, signature}` and permits authenticated re-injection (a typo'd passphrase stays recoverable); without it, today's unauthenticated one-shot flow is byte-for-byte unchanged (v-program images). Secret writes are now atomic (temp file + rename) so a poll landing mid-write can never read a truncated secret. The inject-secret route's body cap is restored to 2 MiB (the `web::Json`→`web::Bytes` switch had silently dropped it to actix's 256 KiB default, which would 413 a valid ~1 MiB body on the v-program path too).
- **attest-cli** gains `--owner-key` / `--owner-key-file` (the file form keeps the key out of argv); redirects are disabled on the attested client, so a signed envelope can only ever bind the key of the peer that actually passed verification.

## How to test
`cargo test -p aleph-tee -p aleph-attest-agent -p aleph-attest-cli` (green on the tip). The redirect and owner-auth paths have dedicated adversarial tests (wrong signer, wrong server key, tampered body, a wire-level redirect-to-another-instance test).

## Notes
- First of a 4-PR stack (owner-auth → guest images → supervisor → agent), split from #1127 for reviewability. Merge in order.
- Whole feature depends on `aleph-message` #158 being released; merges into `dev`, not `main`.


foxpatch-aleph (approved):
Well-structured, security-conscious implementation of owner-authenticated secret injection. The EIP-191 personal-sign over the TLS-key-bound payload is cryptographically sound, the redirect attack mitigation (disabling follows + signing only verified keys) is correct and well-tested, and the atomic write-then-rename eliminates the partial-write race. Adversarial test coverage is thorough (wrong signer, wrong server key, tampered body, redirect-to-another-instance, chain failure). The server key capture slot is only filled on the full verification success path. Minor nits around 0x/0X case sensitivity in signature parsing and missing deny_unknown_fields on the envelope do not affect security (both fail closed).

**`rust/crates/aleph-tee/src/owner_auth.rs`** (line 54): Minor: `strip_prefix("0x")` is case-sensitive, so a signature with uppercase `0X` prefix would fail hex decoding and be rejected. Since `sign_payload` always emits lowercase `0x` this works for the first-party CLI, but third-party tools producing `0X`-prefixed signatures would be rejected. Consider also stripping `0X` (as `parse_owner_key_from` already does at lines 97-98) for consistency.

**`rust/crates/aleph-attest-agent/src/secrets.rs`** (line 36): Minor: `InjectSecretEnvelope` does not use `#[serde(deny_unknown_fields)]`, so unexpected fields are silently ignored. Not a security issue (the signature is bound to the canonical secrets JSON, not the raw body), but adding the attribute would surface client bugs earlier.

PR #1129, opened 2026-08-19
feat(snp-instances) 2/4: LUKS guest image + instance runtime manifest

odesenfans:
**Stack 2/4 for SEV-SNP confidential instances.** Base: `od/snp-inst-1-ownerauth` (PR 1). Reviewer lens: the measured guest image and its manifest, pure nix + Python, no supervisor/agent code.

Related tickets: ALEPH-XXX

## Changes
- **Shared init prologue**: the v-program `init.sh` prologue (mounts, networking, chroot helpers) is factored verbatim, statement-for-statement, into `nix/init-common.sh` so both images share it. The v-program image is unchanged in behavior (proven by an identical measurement + cpio listing across the refactor).
- **Instance guest image** (`nix/init-instance.sh`, a dedicated initrd): boots only in LUKS mode (fail-closed poweroff without `luks=1 owner=0x...`), starts the attest-agent early with `--owner`, waits indefinitely for the passphrase (heartbeat log, wrong-passphrase retry), unlocks `cryptsetup` and chroots in. No platform firewall (firewall policy belongs to the user rootfs). Carries `cryptsetup`/`dm-crypt`, drops veritysetup/nft. A distinct bundle from the v-program image so neither product's TCB carries the other's code. Includes the guest-side `[ -s ]` non-empty passphrase guard (the nix half of the passphrase-race fix whose agent half is in PR 1).
- **`aleph-instance-runtime/1` manifest** (`src/aleph/vm/vprogram/`): a TEE-agnostic manifest flavor (platform is a field), LUKS cmdline template `console=ttyS0 luks=1 owner={owner}` with a closed placeholder set, sharing the template validator with the v-program flavor so the two cannot drift. `vprogram_bundle.py --flavor instance`.
- **Test rootfs**: a dropbear test rootfs (nix) for the E2E harness.

## How to test
`nix build ./nix#instanceImage ./nix#image` (both green); `pytest tests/vprogram/`.

## Notes
- Stacks on PR 1. The instance image references the attest-agent binary whose `--owner` support lands in PR 1.
- The refreshed v-program image measurement changes (shared-init refactor + PR 1's atomic-write). The pin/re-pin procedure and the artifact-upload steps live in a separate operator runbook kept outside this PR (untracked, alongside the design and plan docs), not in the reviewed diff.

### DIFF
--- a/nix/flake.nix
+++ b/nix/flake.nix
@@ -318,6 +318,13 @@
       instanceMeasurementSmoke = instanceMeasurementFor {
         owner = "0x0000000000000000000000000000000000000000";
       };
+
+      # Test fixture for the confidential-instance E2E scenario (Task 14): a
+      # plain ext4 rootfs booting a statically linked dropbear SSH server.
+      # NOT part of the measured chain (no dm-verity, not referenced by any
+      # cmdline roothash) and NOT bit-reproducible (build-time host key). See
+      # test-rootfs.nix for the full rationale.
+      instanceTestRootfs = pkgs.callPackage ./test-rootfs.nix {};
     in {
       # Only concrete derivations are exposed as packages (a function like
       # measurementFor is not a valid flake package and would fail flake check);
@@ -340,7 +347,8 @@
           workloadMeasurement
           image
           instanceImage
-          instanceMeasurementSmoke;
+          instanceMeasurementSmoke
+          instanceTestRootfs;
         default = image;
       };
 
--- a/nix/flake.nix
+++ b/nix/flake.nix
@@ -125,6 +125,7 @@
       initrd = pkgs.callPackage ./initrd.nix {
         inherit attest-agent kernel;
         init-script = ./init.sh;
+        init-common-script = ./init-common.sh;
         udhcpc-script = ./udhcpc.script;
         udhcpc6-script = ./udhcpc6.script;
       };
--- a/nix/flake.nix
+++ b/nix/flake.nix
@@ -288,7 +288,13 @@
       # a valid flake package); `instanceMeasurementSmoke` below gives it
       # build coverage so evaluation and the sev-snp-measure invocation are
       # never dead code.
-      instanceMeasurementFor = { vcpus ? 2, vcpuType ? "EPYC-v4", owner }: let
+      instanceMeasurementFor = { vcpus ? 2, vcpuType ? "EPYC-v4", owner }:
+        # `owner` is interpolated into the measured kernel cmdline (and thus a
+        # shell argument to sev-snp-measure); only trusted nix callers reach
+        # here, but assert the EVM-address shape so a stray value fails the
+        # build rather than producing a bogus measurement or injecting tokens.
+        assert (builtins.match "0x[0-9a-fA-F]{40}" owner) != null;
+        let
         kernelCmdline = "console=ttyS0 luks=1 owner=${owner}";
       in pkgs.runCommand "snp-instance-measurement-${toString vcpus}vcpus-${vcpuType}" {
         nativeBuildInputs = [ sev-snp-measure ];
--- a/nix/flake.nix
+++ b/nix/flake.nix
@@ -7,9 +7,13 @@
   #   aleph-attest-agent static-musl binary and init.sh) + a minimal
   #   dm-verity-protected rootfs + a precomputed, reproducible sev-snp-measure
   #   launch measurement.
-  # Excluded from the donor: compose-rootfs / compose-demo and the
-  # encrypted-rootfs (LUKS) mode. The fib-service demo app is NOT excluded
-  # here: it now exists as the V-PROGRAM workload (see workload.nix), baked
+  # Excluded from the donor: compose-rootfs / compose-demo. The
+  # encrypted-rootfs (LUKS) mode returns as a second, separate image flavor
+  # for confidential instances (instanceInitrd / instanceImage below), built
+  # from the same initrd.nix with withVerity=false withNft=false withLuks=true;
+  # it does not touch the v-program image's measured initrd contents. The
+  # fib-service demo app is NOT excluded here: it now exists as the V-PROGRAM
+  # workload (see workload.nix), baked
   # into its own measured dm-verity volume that the guest init mounts and
   # execs when a workload_roothash is present on the cmdline. The trivial
   # busybox httpd remains the platform rootfs's baked /sbin/init, used as the
@@ -129,6 +133,21 @@
         udhcpc-script = ./udhcpc.script;
         udhcpc6-script = ./udhcpc6.script;
       };
+
+      # Confidential-instance initrd: LUKS encrypted rootfs, no dm-verity, no
+      # in-guest firewall (firewall policy belongs to the user rootfs on
+      # instances, design section 3 decision 6).
+      instanceInitrd = pkgs.callPackage ./initrd.nix {
+        inherit attest-agent kernel;
+        init-script = ./init-instance.sh;
+        init-common-script = ./init-common.sh;
+        udhcpc-script = ./udhcpc.script;
+        udhcpc6-script = ./udhcpc6.script;
+        withVerity = false;
+        withNft = false;
+        withLuks = true;
+      };
+
       rootfs = pkgs.callPackage ./rootfs.nix {};
 
       # fib-service V-PROGRAM workload volume: a content-only ext4 carrying the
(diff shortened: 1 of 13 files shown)


foxpatch-aleph (approved):
Well-structured PR implementing the LUKS guest image and instance runtime manifest for SEV-SNP confidential instances. The init-common.sh refactor is verbatim-equivalent to the original inline code. The instance init correctly implements fail-closed boot, indefinite passphrase wait with zeroization, and early attest-agent startup. The manifest models share a validator to prevent drift between v-program and instance formats. Test coverage is comprehensive for the Python bundle/manifest/CLI layer. No correctness or security issues found.

**`nix/flake.nix`** (line 292): The `owner` argument is interpolated directly into the Nix string `kernelCmdline`, which then becomes a shell argument to `--append`. Since this is a build-time function called by trusted daemon code and `owner` is expected to be a hex Ethereum address, the risk is minimal. However, adding a format assertion (e.g. matching `0x[0-9a-fA-F]{40}`) here would be good defense-in-depth against a future caller passing an unexpected value.

**`scripts/vprogram_bundle.py`** (line 173): The `--exec` flag is silently ignored when `--flavor instance` is used (documented in the help text). Consider either rejecting the combination with an error or printing a warning, so users don't mistakenly expect exec-runtime behavior from an instance manifest.

**`nix/init-instance.sh`** (line 87): When `/mnt/root/sbin/init` does not exist, the script prints a WARNING but continues to `wait` (line 95). The attest-agent stays up and the VM keeps running with no workload. This is intentional (the owner can debug via the attest-agent), but it contrasts with the v-program init which `poweroff -f` on a missing workload `/sbin/init`. Consider whether this asymmetry is deliberate — a misconfigured instance rootfs could hang indefinitely with no useful service running.

foxpatch-aleph (changes requested):
The PR is well-structured with a clean shared-init refactor, solid LUKS boot flow, and thorough manifest validation/tests. However, the `source.build` provenance field in `vprogram_bundle.py` is hardcoded to the v-program nix target (`nix#image`) and used unchanged for the instance flavor, so the instance bundle-info and manifest record an incorrect build command (`nix#image` instead of `nix#instanceImage`). This is a metadata correctness issue in the published manifest's `source.build` field, which is meant to document how to rebuild and audit the bundle.

**`scripts/vprogram_bundle.py`** (line 42): BUILD_COMMAND is hardcoded to `nix#image` but is used as the `source.build` provenance for BOTH flavors. When building with `--flavor instance`, the actual nix target is `nix#instanceImage` (per `_nix_target`), but the recorded `source.build` still says `nix#image`. The `SourceInfo.build` field is described as "How to rebuild and audit the bundle" — it should reflect the actual build command. Consider making BUILD_COMMAND flavor-dependent, e.g. `BUILD_COMMAND = f'nix build "git+file://$REPO?dir=nix#{_nix_target(flavor)}"'` or computing it in `cmd_build`.

**`src/aleph/vm/vprogram/manifest.py`** (line 79): The error message in `_validate_cmdline_template` hardcodes "for format version 1" regardless of which format's placeholder set is being validated. When the instance format rejects an unknown placeholder like `{platform_roothash}`, the error says "unknown cmdline placeholders for format version 1" which is slightly misleading for `aleph-instance-runtime`. Minor, but could confuse debugging. Consider passing a format name into the validator.

**`scripts/vprogram_bundle.py`** (line 178): The `--exec` flag is silently ignored when `--flavor instance` is passed (the help text documents this, but there is no runtime warning). A user who passes `--flavor instance --exec` gets an instance manifest with no indication that `--exec` was disregarded. Consider emitting a warning on stderr when both are specified.

foxpatch-aleph (changes requested):
Well-structured PR with a clean init-common.sh refactor, solid manifest model design with shared validators preventing format drift, good test coverage for the new instance flavor, and thoughtful security handling (passphrase zeroization, [ -s ] guard, fail-closed boot). However, there is a provenance bug: the BUILD_COMMAND constant in vprogram_bundle.py is hardcoded to nix#image and used unconditionally for SourceInfo.build, so instance-flavor manifests incorrectly record the v-program nix target as their build command. An auditor reproducing an instance bundle from its manifest would build the wrong image.

**`scripts/vprogram_bundle.py`** (line 42): BUILD_COMMAND is hardcoded to `'nix build "git+file://$REPO?dir=nix#image"` and used unconditionally at line 89 for `SourceInfo(build=BUILD_COMMAND)`. When `--flavor instance` is passed, the actual nix build target is `#instanceImage` (line 77, via `_nix_target`), but the recorded `source.build` still says `#image`. This means the instance manifest's `source.build` field is incorrect — an auditor trying to reproduce the instance bundle from the manifest would build the v-program image instead. Fix: make `BUILD_COMMAND` flavor-aware, e.g. `BUILD_COMMAND = f'nix build "git+file://$REPO?dir=nix#{_nix_target(flavor)}"'` or compute it in `cmd_build` using `args.flavor`.

**`scripts/vprogram_bundle.py`** (line 178): The `--exec` flag is silently ignored when `--flavor instance` is used (the instance branch at line 115 never references `args.exec_runtime`). The help text says "ignored with --flavor instance" which is technically accurate, but silently ignoring a flag that changes the output format could surprise users. Consider rejecting `--exec` with `--flavor instance` explicitly (e.g. `if args.flavor == "instance" and args.exec_runtime: parser.error("--exec is incompatible with --flavor instance")`) for a clearer UX. This is a nit, not a bug.

**`nix/init-instance.sh`** (line 88): When `/sbin/init` is missing in the LUKS rootfs, the init only prints a WARNING and falls through to `wait`, leaving the VM running with no workload but the attest-agent still serving. This is consistent with the v-program init.sh behavior (line 184), but for the instance case a missing `/sbin/init` means the user-provided rootfs is broken. Consider whether a FATAL poweroff would be more appropriate here — the VM is useless and the owner would need to rebuild and re-deploy the rootfs anyway. Not a blocker since the behavior is documented and consistent.

**`tests/vprogram/test_cli.py`** (line 131): No test covers the `--exec` + `--flavor instance` combination (verifying it is ignored or rejected). Adding a test that asserts the expected behavior (silent ignore or explicit error) would guard against future regressions if the interaction between these flags changes.

odesenfans:
Addressed the review findings in five separate commits (force-pushed):

- **`source.build` records the wrong nix target for instance bundles** → `fix(vprogram): record the flavor-correct nix target in manifest source.build`. `BUILD_COMMAND` is now derived from the flavor's actual `_nix_target`. (Metadata only: the measurement and sha256-pinned bundle were never affected.)
- **`--exec` silently ignored with `--flavor instance`** → `fix(vprogram): reject --exec with --flavor instance instead of ignoring it`. Now a hard `parser.error`, with a regression test — chose rejection over a warning to match the fail-closed style elsewhere.
- **`_validate_cmdline_template` error hardcodes "format version 1"** → `fix(vprogram): drop the misleading 'format version 1' from the placeholder error`. Now reports the offending and allowed placeholders (works for both formats).
- **`flake.nix` owner interpolation** → `fix(nix): assert the owner is an EVM address in instanceMeasurementFor`. Build-time `0x`-40-hex assertion as defense in depth.
- **Missing `/sbin/init` warned-and-continued vs the v-program init's poweroff** → `fix(nix): power off the instance VM when the unlocked rootfs has no /sbin/init`. Agreed the asymmetry was unjustified: both are malformed user payloads. Now powers off for a clean terminal STOPPED signal; the FATAL reason stays on the owner-readable guest serial.

The two nix changes shift the instance image measurement, which is fine (the image isn't pinned yet; the pin is computed at upload time).

One thing the review surfaced is broader than this PR and now tracked separately in #1137: the guest serial console (including post-`chroot` workload output) is captured to the CRN journal and returned by `/logs`, which is a confidentiality seam for a confidential instance. Pre-existing (same as the SEV path), needs a design pass, not blocking here.

foxpatch-aleph (approved):
Well-structured PR introducing the LUKS confidential-instance image with a clean shared-init prologue extraction, fail-closed boot logic, proper passphrase zeroization, and a properly typed instance manifest format with shared validators that prevent format drift. Test coverage is thorough. No security issues or bugs found; only minor nits (a test fixture with a wrong source.build value, a help text mismatch, and an unbounded owner regex on the shell side).

**`tests/vprogram/test_manifest.py`** (line 153): The MINIMAL_INSTANCE_MANIFEST fixture has source.build set to the v-program target (nix#image) rather than the instance target (nix#instanceImage). The CLI's _build_command correctly emits the right target for each flavor, but this test fixture is copy-pasted from the v-program reference and should use the instance build command for accuracy.

**`scripts/vprogram_bundle.py`** (line 183): The --exec help text says "ignored with --flavor instance" but the code calls parser.error() which rejects the combination. The help text should say "incompatible with --flavor instance" to match the actual behavior.

**`nix/init-instance.sh`** (line 19): The sed pattern for owner captures 0x[0-9a-fA-F]* (unbounded length), while the nix-side assertion enforces exactly 40 hex chars. Consider anchoring the shell regex to match the 40-character expectation so a malformed owner fails closed at boot rather than reaching the attest-agent with a truncated or over-long value.

PR #1130, opened 2026-08-19
feat(snp-instances) 3/4: supervisor opaque-cmdline SNP arm

odesenfans:
**Stack 3/4 for SEV-SNP confidential instances.** Base: `od/snp-inst-2-images` (PR 2). Reviewer lens: the supervisor contract and the daemon launch-integrity gate.

This layer is inert until an agent (PR 4) sends `kernel_cmdline`; it changes no existing launch path's behavior.

Related tickets: ALEPH-XXX

## Changes
- **Proto (additive only)**: `TeeConfig.kernel_cmdline` (field 5), an opaque measured cmdline the daemon passes to QEMU verbatim and never parses, and `VmSpec.vm_type` (field 16), so the daemon stops inferring V-PROGRAM from the SNP backend. Field numbers fixed; existing fields byte-identical.
- **Daemon**: a second, fully fail-closed SNP arm. `kernel_cmdline` set = the opaque/LUKS path (writable qcow2 rootfs, no verity, no hash-tree volume, rejects a stray `{rootfs}.roothash` sidecar, rejects `kernel_cmdline` on a non-SNP backend); `kernel_cmdline` empty = today's verity path byte-for-byte (pinned by a golden config test). `vm_type` is honored from the request with a persisted-key adoption fallback, so create and adoption can never disagree.
- **Controller**: `build_snp_argv` parameterizes the rootfs disk (raw+readonly for verity, qcow2+writable for LUKS); a half-populated override is refused before launch (`RootfsOverride::Malformed`). Divergence entry 82 records that aleph-tee emits no rootfs drive line, so our token shape is asserted directly.

## How to test
`cargo test -p supervisor-daemon -p supervisor-controller` (green; includes the fail-closed combination walk and argv parity fixtures).

## Notes
- Stacks on PR 2 (shares no files with it; based there only to keep the stack linear).


foxpatch-aleph (approved):
Well-structured PR with clean fail-closed design: two mutually-exclusive SNP arms (verity vs opaque-cmdline), a three-way RootfsOverride enum that correctly distinguishes malformed half-populated configs from absent ones, and VmType persistence with a legacy inference fallback for older configs. Proto changes are purely additive with correct field numbers. Test coverage is strong for the main paths (golden config parity, verity byte-identical pinning, opaque arm, sidecar rejection, non-SNP rejection, format validation). The only gap is the untested RootfsOverride::Writable(format, true) arm in build_snp_argv, which is a minor coverage nit. No correctness, security, or logic issues found.

**`rust/crates/supervisor-controller/src/qemu.rs`** (line 569): The `RootfsOverride::Writable(format, true)` arm (writable override with readonly=true, emitting `readonly=on`) has no test coverage. The daemon's opaque arm passes `rootfs.readonly` verbatim from the spec's DiskConfig, so an agent setting `readonly=true` on a qcow2 rootfs would hit this path. Consider adding a test case like `snp_config_with_rootfs_override(Some(("qcow2", true)))` to assert the `readonly=on` token is emitted.

**`rust/crates/supervisor-controller/src/qemu.rs`** (line 573): The `RootfsOverride::Malformed` fallback in release builds silently uses the read-only raw default instead of panicking. The `select_run_target` guard makes this unreachable in production, but the other invariant violations in this function use `.expect()`. Using `.expect()` here too (or `unreachable!`) would be more consistent for an invariant violation, though the current approach is documented and safe-by-default.

foxpatch-aleph (approved):
This PR introduces a well-structured, additive opaque-cmdline SNP arm with a consistent fail-closed design. The proto changes use new field numbers without touching existing fields. The daemon's two-arm SNP slice (verity vs opaque) is mutually exclusive by construction, with robust guards: non-SNP kernel_cmdline is rejected, dual cmdline sources (agent + roothash sidecar) are rejected, unsupported rootfs formats are rejected, and half-populated rootfs overrides are refused before dispatch. The controller's RootfsOverride enum cleanly models the three-way state (default/writable/malformed). The vm_type persistence with adoption fallback ensures create and adoption never disagree on IPv6 classification. The image_format string is safely derived from a validated proto enum (not user-supplied free text) on the daemon side. Test coverage is comprehensive, covering both arms, all fail-closed paths, backward compatibility, and conformance fixtures. No correctness, security, or testing gaps found.

**`src/aleph/vm/supervisor_interface/wire/proto_convert.py`** (line 284): `SPEC_VM_TYPE_FROM_PB[msg.vm_type]` will raise `KeyError` if a newer proto sends an unknown `vm_type` value. This is consistent with all other enum conversions in this file (same pattern for `BACKEND_FROM_PB`, `TEE_BACKEND_FROM_PB`, etc.), so it's a pre-existing convention rather than a new issue — but worth noting for future hardening if any enum ever grows forward-compat requirements.

foxpatch-aleph (approved):
Well-structured PR implementing the supervisor opaque-cmdline SNP arm with a fail-closed design. The proto changes are additive with correct field numbering. The daemon's two-arm SNP slice (verity vs opaque) is cleanly separated by the `kernel_cmdline` presence, with proper rejection of conflicting roothash sidecars, unsupported rootfs formats, and non-SNP cmdline usage. The controller's `RootfsOverride` three-way enum elegantly handles the half-populated config case. The `vm_type` persistence and adoption fallback (`of_qemu`) correctly prefers the persisted key over the legacy snp-implies-VProgram inference. Test coverage is thorough across both arms, the fail-closed combinations, the verity path byte-pinning, and the Python proto round-trip.

**`rust/crates/supervisor-daemon/src/service.rs`** (line 532): Observation (non-blocking): `vm_spec_message` always emits `kernel_cmdline: String::new()` and `vm_type: pb::VmType::Unspecified`. This means GetVmSpec does not round-trip the opaque cmdline or the vm_type the agent set at create time. This is pre-existing behavior (the function reconstructs a minimal spec from the controller config, not from `entry.spec`), but if the agent ever needs to read back the opaque cmdline via GetVmSpec, this would need to be addressed. Worth noting for a future PR if the agent requires it.

foxpatch-aleph (commented):
Well-designed additive change introducing the opaque-cmdline SEV-SNP arm with comprehensive fail-closed guards at every boundary. The verity path is proven byte-identical, the adoption round-trip correctly reconstructs the TeeConfig, and the RootfsOverride enum cleanly handles the malformed case. A few minor observations: the PR description mentions VmSpec.vm_type (field 16) which is absent from the actual proto; the Writable(format, true) arm of build_snp_argv lacks test coverage; and the variable name 'format' shadows the format! macro in qemu.rs (valid but slightly confusing). No correctness or security issues found.

**`rust/crates/supervisor-controller/src/qemu.rs`** (line 569): The `RootfsOverride::Writable(format, true)` arm (readonly qcow2/raw) is handled but has no corresponding test. The two existing tests cover `Default` and `Writable("qcow2", false)` only. Consider adding a test case like `snp_config_with_rootfs_override(Some(("qcow2", true)))` to exercise the `readonly=on` + non-raw-format path.

**`rust/crates/supervisor-controller/src/qemu.rs`** (line 563): The binding name `format` shadows the `format!` macro. This is valid Rust (macro invocation uses `!` so there's no ambiguity to the compiler), but it's slightly confusing on first read. Renaming to `fmt` or `img_format` would improve readability.

**`proto/supervisor.proto`** (line 254): PR description mentions `VmSpec.vm_type` (field 16) being added, but the proto file and diff contain no such field — VmSpec ends at field 15 (hostname). Either the description is inaccurate or the field is planned for a later PR in the stack. Worth clarifying.

PR #1131, opened 2026-08-19
feat(snp-instances) 4/4: agent launch path for mode=sev_snp

odesenfans:
**Stack 4/4 for SEV-SNP confidential instances.** Base: `od/snp-inst-3-supervisor` (PR 3). Reviewer lens: the agent wiring that activates the feature.

Related tickets: ALEPH-XXX

## Changes
- **`snp_staging`**: the bundle fetch/verify/extract helpers factored out of `vprogram_launch` (v-program behavior unchanged; its test file passes unedited), consumed by both paths.
- **`snp_instance_launch.build_snp_instance_spec`**: fetches and validates the `aleph-instance-runtime/1` manifest, stages the bundle, renders the measured `owner=` cmdline from the message sender, and builds the SNP spec. The rejection ladder (attestation_port set, host not SNP-capable, policy ≥ 2^64-unsafe, non-EVM sender) runs before any I/O, so a rejected message never leaves staging behind. The manifest cmdline template is the only source of cmdline tokens and is schema-closed to `{owner}`.
- **`run.py`** routes `mode=sev_snp` instances through this path: boots immediately (no `awaiting_confidential_init`), forwards SSH plus the manifest attestation port, cleans staging on every failure path. `translate.py` gains a fail-loud guard so an SNP instance can never fall through to the legacy SEV path.
- **`examples/instance_confidential_snp/build_luks_rootfs.sh`**: builds a LUKS2 rootfs (with the `e2fsck`-before-`resize2fs` fix). SNP-instance tests are isolated from the real `EXECUTION_ROOT`.

## How to test
`pytest tests/supervisor/test_snp_instance_launch.py tests/supervisor/test_snp_instance_run.py tests/supervisor/test_snp_staging.py` (green, hermetic).

## Notes
- Top of the 4-PR stack. The end-to-end proof is the aleph-testnets scenario (separate repo, unpushed until the artifact pins from PR 2's runbook are uploaded, so a paid SNP CI run does not fire prematurely).
- Deferred/out-of-scope (tracked): pyaleph persistence + CCN measurement validation; reject empty secret values; document/reject `workload_roothash` on the opaque arm; the pre-existing staging-leak on the expiry/tasks delete paths; a design-doc note on the clone-VM residual.

### DIFF
--- a/examples/instance_confidential_snp/build_luks_rootfs.sh
+++ b/examples/instance_confidential_snp/build_luks_rootfs.sh
@@ -21,5 +21,10 @@ printf '%s' "$PASS" | cryptsetup luksFormat --type luks2 --batch-mode "$LOOP" -
 printf '%s' "$PASS" | cryptsetup luksOpen "$LOOP" snp-rootfs-build -
 trap 'cryptsetup luksClose snp-rootfs-build; losetup -d "$LOOP"' EXIT
 dd if="$PLAIN" of=/dev/mapper/snp-rootfs-build bs=4M status=progress conv=fsync
+# resize2fs refuses to grow a filesystem whose last-check time predates its
+# last-modify time (true of any image just copied onto by dd), so force a
+# check first. e2fsck's exit codes 1 and 2 mean errors were found and fixed
+# (not a failure); only bail out above that.
+e2fsck -fp /dev/mapper/snp-rootfs-build || [ $? -le 2 ]
 resize2fs /dev/mapper/snp-rootfs-build
 echo "wrote LUKS2 rootfs to $OUT"
--- a/examples/instance_confidential_snp/build_luks_rootfs.sh
+++ b/examples/instance_confidential_snp/build_luks_rootfs.sh
@@ -1,0 +1,25 @@
+#!/bin/bash
+# Build a LUKS2-encrypted SNP instance rootfs from a plain ext4 rootfs image.
+#
+# Usage: build_luks_rootfs.sh <plain-rootfs.ext4> <output.img> [size-mib]
+# Prompts for the passphrase (or reads LUKS_PASSPHRASE from the environment).
+#
+# Contract (docs/plans/2026-08-18-snp-confidential-instances-design.md):
+# whole-device LUKS2 container, ext4 inside, an executable /sbin/init run
+# CHROOTED by the measured init (not PID 1: systemd images will not work),
+# SSH keys and all provisioning inside the image. The CRN never sees the
+# passphrase; it is injected post-attestation via aleph-attest-cli
+# inject-secret --owner-key.
+set -euo pipefail
+PLAIN="$1"; OUT="$2"; SIZE_MIB="${3:-$(( ($(stat -c%s "$PLAIN") / 1048576) + 64 ))}"
+PASS="${LUKS_PASSPHRASE:-}"
+if [ -z "$PASS" ]; then read -r -s -p "LUKS passphrase: " PASS; echo; fi
+truncate -s "${SIZE_MIB}M" "$OUT"
+LOOP=$(losetup --find --show "$OUT")
+trap 'losetup -d "$LOOP"' EXIT
+printf '%s' "$PASS" | cryptsetup luksFormat --type luks2 --batch-mode "$LOOP" -
+printf '%s' "$PASS" | cryptsetup luksOpen "$LOOP" snp-rootfs-build -
+trap 'cryptsetup luksClose snp-rootfs-build; losetup -d "$LOOP"' EXIT
+dd if="$PLAIN" of=/dev/mapper/snp-rootfs-build bs=4M status=progress conv=fsync
+resize2fs /dev/mapper/snp-rootfs-build
+echo "wrote LUKS2 rootfs to $OUT"
--- a/examples/instance_confidential_snp/build_luks_rootfs.sh
+++ b/examples/instance_confidential_snp/build_luks_rootfs.sh
@@ -27,8 +27,10 @@ trap 'cryptsetup luksClose snp-rootfs-build; losetup -d "$LOOP"' EXIT
 dd if="$PLAIN" of=/dev/mapper/snp-rootfs-build bs=4M status=progress conv=fsync
 # resize2fs refuses to grow a filesystem whose last-check time predates its
 # last-modify time (true of any image just copied onto by dd), so force a
-# check first. e2fsck's exit codes 1 and 2 mean errors were found and fixed
-# (not a failure); only bail out above that.
-e2fsck -fp /dev/mapper/snp-rootfs-build || [ $? -le 2 ]
+# check first. e2fsck's exit code is a bitwise OR: 1 = errors corrected,
+# 2 = corrected and a reboot is advised (irrelevant for a loop-mounted build
+# image), 4 = errors left UNCORRECTED, 8 and above = operational failures.
+# So anything below 4 (including 3 = 1|2) means the check succeeded.
+e2fsck -fp /dev/mapper/snp-rootfs-build || [ $? -lt 4 ]
 resize2fs /dev/mapper/snp-rootfs-build
 echo "wrote LUKS2 rootfs to $OUT"
--- a/examples/instance_confidential_snp/build_luks_rootfs.sh
+++ b/examples/instance_confidential_snp/build_luks_rootfs.sh
@@ -11,6 +11,10 @@
 # passphrase; it is injected post-attestation via aleph-attest-cli
 # inject-secret --owner-key.
 set -euo pipefail
+if [ "$#" -lt 2 ]; then
+    echo "Usage: $0 <plain-rootfs.ext4> <output.img> [size-mib]" >&2
+    exit 1
+fi
 PLAIN="$1"; OUT="$2"; SIZE_MIB="${3:-$(( ($(stat -c%s "$PLAIN") / 1048576) + 64 ))}"
 PASS="${LUKS_PASSPHRASE:-}"
 if [ -z "$PASS" ]; then read -r -s -p "LUKS passphrase: " PASS; echo; fi
--- a/src/aleph/vm/agent/run.py
+++ b/src/aleph/vm/agent/run.py
@@ -25,6 +25,11 @@ from multidict import CIMultiDict
 from aleph.vm.agent.aggregate import get_user_settings
 from aleph.vm.agent.capacity import CapacityManager, requested_gpu_ids
 from aleph.vm.agent.expiry import ExpiryManager
+from aleph.vm.agent.snp_instance_launch import (
+    build_snp_instance_spec,
+    is_snp_instance,
+    remove_snp_instance_staging,
+)
 from aleph.vm.agent.translate import build_create_vm_spec, build_program_create_vm_spec
 from aleph.vm.agent.update_watcher import UpdateWatcher
 from aleph.vm.agent.vm.program_client import ProgramGuestClient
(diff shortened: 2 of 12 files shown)


foxpatch-aleph (commented):
Well-structured PR that cleanly factors shared staging helpers into snp_staging.py, wires a dedicated SNP instance launch path with a thorough rejection ladder (all pre-I/O), and adds staging cleanup to every teardown site. The translate.py fail-loud guard is good defense-in-depth. The cmdline template injection surface is properly closed by the manifest model's placeholder validation. Tests are hermetic and cover the happy path, all rejection branches, routing, and staging cleanup. A few minor test gaps noted below, none blocking.

**`tests/supervisor/test_snp_staging.py`** (line 80): Test coverage gap: there is a test for sha256 mismatch but no test for size mismatch. The `verify_bundle` function checks size first (line 67-70 of snp_staging.py), and a size mismatch is a distinct failure path. Consider adding a test that passes a wrong `size` argument and asserts `VmSetupError` with "size mismatch" and no staging directory left behind.

**`tests/supervisor/test_snp_instance_run.py`** (line 266): Test coverage gap: `test_snp_instance_failure_cleans_staging` tests the `create_vm` failure path (first except block at run.py:348-359), but there is no test for the `finish_instance_create`/port-forward failure path (run.py:384-394), which also calls `remove_snp_instance_staging`. Consider adding a test where `finish_instance_create` or `add_port_forward` raises, and asserting staging is cleaned and the VM is deleted.

**`src/aleph/vm/agent/translate.py`** (line 89): The fail-loud guard for SNP instances is a good defense-in-depth measure. There is no direct test for this guard (an SNP instance content routed through `build_create_vm_spec` should raise `InvalidBackendError`). The existing `test_legacy_sev_instance_path_untouched` test in test_snp_instance_run.py mocks `build_snp_instance_spec` so it never hits translate.py. Consider a unit test that calls `build_create_vm_spec` directly with an SNP instance content and asserts `InvalidBackendError`.

**`examples/instance_confidential_snp/build_luks_rootfs.sh`** (line 28): The `e2fsck -fp ... || [ $? -le 2 ]` pattern is correct under `set -e` (the `||` list suppresses `set -e` on the left side). Minor note: the comment says "only bail out above that" but e2fsck exit code 3 also means errors uncorrected (the manual distinguishes 3 = errors uncorrected, 4 = operational error). The `[ $? -le 2 ]` correctly treats both 3 and 4 as failures, so the code is right — just the comment could be slightly more precise about the exit code range.

**`src/aleph/vm/agent/snp_instance_launch.py`** (line 153): The `kernel_cmdline = manifest.boot.cmdline_template.format(owner=owner)` is safe because the manifest model validates the template to only allow `{owner}` (CMDLINE_PLACEHOLDERS_LUKS_V1 in manifest.py). The `owner` value is itself validated by `_EVM_ADDRESS_PATTERN.fullmatch(owner)` on the line above. This two-layer validation (model-level template restriction + runtime address validation) is a good security pattern.

foxpatch-aleph (approved):
Well-structured PR with clean separation of concerns. The SNP instance launch path correctly validates, stages, and builds specs with a properly ordered rejection ladder that prevents staging leaks. The snp_staging module is a solid factoring of shared bundle integrity logic. Security posture is strong: tarfile filter='data' extraction, path traversal checks, constant-time hash comparison, and schema-closed cmdline templates. Tests are hermetic and cover the happy path, all rejection branches, staging cleanup, and legacy SEV non-regression. Two non-blocking gaps: the translate.py fail-loud guard has no direct test, and the second failure path (finish_instance_create/port-forward) staging cleanup is untested. The build_luks_rootfs.sh script is correct but could use basic argument validation.

**`src/aleph/vm/agent/translate.py`** (line 95): The fail-loud guard for SNP instances reaching this path is not directly tested. Consider adding a test that passes an SNP InstanceContent to `build_create_vm_spec` and asserts `InvalidBackendError` is raised — this would protect against a future routing regression where the guard is accidentally bypassed.

**`src/aleph/vm/agent/run.py`** (line 392): The second exception handler (finish_instance_create / port-forward failure) calls `remove_snp_instance_staging` but no test covers this path. `test_snp_instance_failure_cleans_staging` only tests the first except block (create_vm failure). Consider adding a test where `finish_instance_create` raises (e.g. `_wait_until_running` times out) to verify staging is cleaned on this path too.

**`tests/supervisor/test_snp_staging.py`** (line 80): Only sha256 mismatch is tested; size mismatch (the other half of `verify_bundle`) is not. A test with a wrong `size` parameter would complete the integrity-gate coverage.

**`examples/instance_confidential_snp/build_luks_rootfs.sh`** (line 14): No argument count validation. Running without arguments produces a confusing `stat` error. Consider adding `if [ $# -lt 2 ]; then echo "Usage: $0 <plain-rootfs.ext4> <output.img> [size-mib]" >&2; exit 1; fi` after the `set` line.

foxpatch-aleph (commented):
Well-structured PR with a clean security model: rejection-before-I/O, staging cleanup on every failure path, and safe tarball extraction. The shared snp_staging refactor correctly preserves v-program behavior. The main gaps are in direct test coverage for snp_staging.py's error branches (size mismatch, corrupt tarball, member_path escaping) and the second failure path in run.py (finish_instance_create failure → staging cleanup). Most staging gaps have indirect coverage through test_vprogram_launch.py since the helpers are shared, but the member_path escaping-member branch appears untested anywhere. No blocking issues found — these are non-blocking observations for future test hardening.

**`tests/supervisor/test_snp_staging.py`** (line 97): Only 2 tests here — consider adding direct tests for verify_bundle size mismatch, extract_bundle corrupt-tarball failure, and member_path (missing member + escaping member). The member_path escaping-member branch (snp_staging.py:124) does not appear to be tested anywhere in the suite, since the manifest model validates member paths upfront and the v-program tests only exercise the missing-member case indirectly.

**`tests/supervisor/test_snp_instance_launch.py`** (line 316): The warning fires on `content.authorized_keys or content.variables`, but only the authorized_keys branch is tested. Consider adding a test that passes only `variables={"KEY": "VALUE"}` to exercise the other side of the `or`.

**`tests/supervisor/test_snp_instance_run.py`** (line 293): Only the first failure path is tested (supervisor.create_vm raises → remove_snp_instance_staging). The second failure path (run.py:384-394: finish_instance_create or add_port_forward raises → remove_snp_instance_staging) is not tested. A test making add_port_forward raise after a successful create_vm would close this gap.

**`examples/instance_confidential_snp/build_luks_rootfs.sh`** (line 14): No validation of $1 and $2 — if the script is invoked with fewer than 2 args, `set -u` will abort on the unset variable, but the error message will be opaque. A usage check (`[ $# -lt 2 ] && { echo "Usage:..." >&2; exit 1; }`) would be friendlier. Not blocking.

foxpatch-aleph (approved):
Well-structured PR that cleanly factors shared SNP bundle-staging helpers, adds a dedicated SNP instance launch path with a pre-I/O rejection ladder, and wires staging cleanup into all teardown paths. Security is solid: schema-closed cmdline template, tarfile filter=data extraction, path containment checks, and explicit rejection of unmeasured host inputs. Tests are comprehensive across happy path, all rejection branches, routing, and staging cleanup. The only gaps are minor (missing tests for extract_bundle failure and finish_instance_create teardown staging cleanup), neither of which is a correctness or security concern.

**`tests/supervisor/test_snp_staging.py`** (line 79): Consider adding a test for `verify_bundle` size mismatch (e.g., pass `size=len(raw)+1`) and for `extract_bundle` failure (e.g., a corrupt tarball that passes size/sha256 but fails extraction). The vprogram tests cover size mismatch, but since `snp_staging` is now the shared module, direct coverage would be valuable if the vprogram tests ever move or change.

**`tests/supervisor/test_snp_instance_run.py`** (line 264): Missing a test for the `finish_instance_create`/port-forward failure path (run.py:384-394) that cleans staging when readiness or port-forward setup fails after a successful `create_vm`. The existing `test_snp_instance_failure_cleans_staging` only covers `create_vm` failure. A test where `finish_instance_create` raises (e.g., VM never reaches RUNNING) and asserting `remove_snp_instance_staging` is called would close this gap.

**`examples/instance_confidential_snp/build_luks_rootfs.sh`** (line 28): The `e2fsck -fp ... || [ $? -le 2 ]` pattern is correct under `set -e` (the `||` makes it a compound command, so `set -e` doesn't trigger on e2fsck's non-zero exit, and `$?` correctly captures e2fsck's exit code). The comment explains this well. No change needed, just confirming the semantics are sound.

foxpatch-aleph (commented):
Well-structured PR that cleanly routes SEV-SNP confidential instances through a dedicated launch path. The security design is solid: schema-closed cmdline templates, constant-time hash comparison, tarfile safety filters with fail-closed fallbacks, and defense-in-depth path escape checks. Staging cleanup is covered on both failure paths and all teardown routes. The rejection ladder correctly fires before any I/O. Tests are comprehensive for the main paths. A few minor test coverage gaps (attestation port fallback, corrupt-but-verified tarball extraction) and a slightly misleading assertion in the variables warning test, but nothing blocking.

**`tests/supervisor/test_snp_instance_launch.py`** (line 333): The test asserts `"authorized_keys" in record.message` but is testing the `variables` warning path. The warning message does mention both `authorized_keys/variables`, so this passes, but it would be clearer to also assert `"variables" in record.message` to distinguish this test from `test_warns_and_ignores_authorized_keys`.

**`tests/supervisor/test_snp_instance_launch.py`** (line 246): Missing test for the attestation port fallback logic: when the manifest has multiple attestation descriptors and `aleph.ra-tls` is not first, the code at snp_instance_launch.py:159-163 should prefer the `aleph.ra-tls` port. Consider adding a test variant where the first descriptor uses a non-ra-tls protocol and verifying the returned port is from the `aleph.ra-tls` descriptor.

**`tests/supervisor/test_snp_staging.py`** (line 79): Missing test for `extract_bundle` failure on a corrupt tarball that passes the size+sha256 gate but fails during extraction (e.g., truncated gzip stream). The code at snp_staging.py:96-100 handles this by cleaning up and raising VmSetupError, but no test covers that path.

**`src/aleph/vm/agent/snp_instance_launch.py`** (line 174): The `settings.CONFIDENTIAL_SESSION_DIRECTORY or (Path(settings.EXECUTION_ROOT) / "sessions")` fallback is dead code: `conf.py:701-702` always sets `CONFIDENTIAL_SESSION_DIRECTORY` to `EXECUTION_ROOT / "sessions"` when it is None. Harmless but slightly misleading — a reader might think the fallback is reachable. This matches the same pattern in `vprogram_launch.py:191` for consistency, so keeping it is fine.

foxpatch-aleph (approved):
This PR cleanly adds the SEV-SNP confidential instance launch path with well-factored shared staging helpers, thorough fail-closed validation (attestation port, host capability, policy, EVM sender), consistent staging cleanup across all teardown paths, and comprehensive hermetic tests. The refactoring of vprogram_launch.py to use snp_staging is minimal and preserves existing behavior. The translate.py fail-loud guard prevents routing regressions. Security is well-handled: the tarball integrity gate (sha256+size with constant-time compare), tarfile filter="data" safety filter, member path escape check, and schema-closed cmdline template (restricted to {owner}) provide defense in depth. No correctness, security, or logic issues found. Minor test coverage gap: no explicit test for QemuDownloader.download_all() failure within build_snp_instance_spec, but the existing exception handling in create_vm_execution correctly covers this path.

**`src/aleph/vm/agent/snp_instance_launch.py`** (line 165): Minor coverage gap: if QemuDownloader.download_all() fails after the bundle has been staged, the except block in run.py:348-359 correctly calls remove_snp_instance_staging, but no test explicitly covers this intermediate failure (staging done, rootfs download fails). The existing test_snp_instance_failure_cleans_staging covers create_vm failure, which is a later stage. Consider adding a test that patches download_all to raise after staging succeeds.

**`examples/instance_confidential_snp/build_luks_rootfs.sh`** (line 18): SIZE_MIB default arithmetic uses integer division on the byte size: $(stat -c%s "$PLAIN") / 1048576. For a rootfs whose size is not a multiple of 1 MiB, this truncates downward, then adds 64 MiB of headroom. This is fine for a LUKS container (resize2fs grows into the full size), but the +64 MiB overhead should be enough for LUKS2 metadata. Worth a comment noting the 64 MiB is the LUKS2 header padding.

PR #1132, opened 2026-08-19
feat(nix): aleph.compose/1 compose-runner runtime flavor

odesenfans:
Ports the Docker Compose runner runtime (developed in aleph-cvm) into this repo as a second V-PROGRAM runtime flavor under the reserved contract `aleph.compose/1` (named as deferred future work in the workload-mechanism design). Guest images + manifest tooling + docs only: the CRN launch path is untouched.

## What's in here

- **`nix/compose-rootfs.nix`**: podman/podman-compose/crun/fuse-overlayfs platform rootfs (full closure), deterministic ext4 (same UUID/hash_seed/no-journal/epoch-0 levers as `rootfs.nix`, fakeroot root-owned), fuse.ko from the unmodified shared kernel, and a fail-closed embedded init: every mount/insmod/layout/podman-load failure and stack exit ends in `poweroff -f` (11 fatal sites), never an attested-but-empty VM.
- **`nix/init-compose.sh`**: initrd flavor inverting the `aleph.exec/1` topology: the platform rootfs runs podman and the measured workload volume is DATA bind-mounted at `/mnt/workload` (compose file + OCI archives). A thin flavor over `init-common.sh` like `init.sh`/`init-instance.sh`, built through the reworked `initrd.nix` unchanged (the `withVerity`/`withNft` defaults are already right for compose). Both roothashes are required, all 14 fatal paths enumerated in the contract doc, guest exit powers the VM off, nft 8443-only firewall kept (it is the single-entrypoint contract); the existing `#initrd` output is bit-identical.
- **Flake outputs** `composeRootfs` / `composeVerity` / `composeInitrd` / `composeMeasurementFor` / `composeMeasurement` / `composeImage`, mirroring the existing flavor through a single cmdline-template source (`measurementFor` generalized with behavior-preserving defaults).
- **Manifest tooling**: `COMPOSE_WORKLOAD` (`aleph.compose/1`, upstream 8080) and a `--flavor compose` value on both `scripts/vprogram_bundle.py` subcommands (joining the instance flavor's axis): `build` targets `nix#composeImage` and records it as `source.build` provenance, `manifest` selects the workload-bearing cmdline template; `--exec` is rejected with any non-vprogram flavor.
- **Contract of record**: `docs/plans/2026-08-19-compose-runtime-port-design.md` (layout, compose subset as enforced by the aleph-rs CLI, fail-closed tables for both layers, sizing/RAM rules, secret injection declared out of scope for v1, image trust rationale, publish flow); divergences entry 64(a) amended (compose no longer excluded; LUKS still is).

## Verified

- `#image`, `#measurement`, `#initrd`, `#instanceInitrd`, `#instanceImage` store paths are byte-identical to `origin/dev` (checked by nix eval/build at every step): no existing flavor's published measurement can have drifted.
- Full bundle dry run: `composeImage` -> `vprogram_bundle.py build --flavor compose` (311 MB tar.gz) -> `manifest --flavor compose` parses with the Python `RuntimeManifest` AND satisfies the aleph-rs CLI's compose gate (aleph-rs PR #344), cmdline template byte-identical across flake/bundle.py/daemon.
- Compose rootfs and image reproducible (identical hashes across rebuilds).

## Notes

- No compose guest has booted yet: cgroup2/fuse/podman-compose first exercise on the SNP box; the fail-closed design means a first-boot failure powers off loudly.
- After publishing (manual `aleph file upload` x2 per the existing flow), refresh aleph-rs PR #344's manifest fixture with the real bundle.ref/sha256/roothash.


foxpatch-aleph (approved):
Well-structured port of the compose-runner runtime flavor with thorough fail-closed semantics, clean generalization of the measurement builder, and comprehensive documentation. The design doc accurately traces every fatal path and the code matches the stated contract. Test coverage is appropriate for the new CLI/manifest surface. The only notable issue is a minor inconsistency in one poweroff path (missing `exec`), which is non-blocking.

**`nix/init-compose.sh`** (line 342): This is the only fatal `poweroff -f` site in the file that does NOT use `exec` — all 13 others use `exec /bin/busybox poweroff -f`. Without `exec`, if `poweroff -f` were to fail (extremely unlikely but not impossible), the script would continue past this line into `setup_firewall` and the chroot. Consider using `exec` here too for consistency with every other fatal path: `... || { echo "init: FATAL: workload bind mount failed"; exec /bin/busybox poweroff -f; }`

**`nix/init-compose.sh`** (line 358): The attest-agent is backgrounded without its PID captured or supervised. If the agent crashes, `wait "$guest_pid"` remains blocked on the compose stack, leaving a live VM with no attested endpoint (a milder variant of the attested-but-empty problem this PR otherwise closes). This is a pre-existing pattern from `init.sh` (line 343) and not introduced by this PR, but the compose flavor's fail-closed philosophy makes it more visible. Not a blocker; noting for the record.

**`src/aleph/vm/vprogram/bundle.py`** (line 168): The Python-level `if exec_runtime and compose_runtime` check is unreachable through the CLI path (argparse's `add_mutually_exclusive_group` rejects `--exec --compose` first, producing the `"not allowed with argument"` message the test asserts on). This is correct defense-in-depth for direct library callers of `make_manifest`, but the test `test_manifest_exec_and_compose_are_mutually_exclusive` only exercises the argparse path. A unit test calling `make_manifest` directly with both flags would close the gap.

odesenfans:
Rebased onto dev after the snp-instances stack landed (base now 4b5b34ca), meshing the compose flavor with the new two-runtime flake structure:

- **init-compose.sh is now a thin flavor over init-common.sh**, like init.sh and init-instance.sh, instead of carrying a byte copy of the old prologue. Behavior is unchanged: the same 14 fatal poweroff sites (9 shared with init.sh, 5 compose deltas), with one wart fixed (the workload bind-mount fatal path now execs poweroff directly).
- **composeInitrd uses the reworked initrd.nix as-is** (init-script + init-common-script; the withVerity/withNft defaults are already right for compose), so the branch no longer patches initrd.nix at all.
- **The bundle tooling adopts the new --flavor axis**: `build --flavor compose` targets nix#composeImage and records exactly that as source.build provenance (fixing the old cosmetic gap where compose bundles carried #image provenance), and `manifest --flavor compose` replaces the separate --compose flag.

Verified after the rebase:
- `#image`, `#measurement`, `#initrd`, `#instanceInitrd`, `#instanceImage` drv paths are byte-identical to origin/dev, so neither existing flavor's published measurement can drift.
- `#composeImage` builds; the compose rootfs came straight from cache (unchanged), only the compose initrd + measurement rebuilt, and the measurement keeps its distinct store-path name.
- Full bundle dry run through the new flow: 311 MB tarball, manifest carries `aleph.compose/1`, the exec cmdline template, and `nix build ...#composeImage` provenance.
- tests/vprogram: 68 passed; shellcheck 0.11 clean on the rewritten init-compose.sh.

foxpatch-aleph (approved):
A well-structured PR that ports the Docker Compose runner runtime as a second V-PROGRAM flavor under the aleph.compose/1 contract. The fail-closed design is thorough and correct at both layers (14 poweroff sites in init-compose.sh, 11 fatal sites in the embedded compose init), the launch topology inversion is cleanly implemented, the measurementFor generalization preserves byte-identical existing outputs via behavior-preserving defaults, and the insecureAcceptAnything podman policy is properly justified by the measured-volume/no-pull-path constraints. Tests cover the new compose manifest path, mutual exclusion with --exec, and the COMPOSE_WORKLOAD constant. The only minor issue is a documentation discrepancy: the PR description claims initrd.nix gains a defaulted init-script parameter, but no default was actually added (it remains required, with all three call sites passing it explicitly), which is harmless since the code is correct.

**`nix/initrd.nix`** (line 1): The PR description says "initrd.nix gains a defaulted init-script parameter" but no default was added — `init-script` is still a required parameter (no `?` default). All three call sites (initrd, instanceInitrd, composeInitrd in flake.nix) pass it explicitly, so this is not a bug, just a minor doc mismatch.

foxpatch-aleph (approved):
Well-structured port of the Docker Compose runner runtime with thorough fail-closed design at both init layers, sound security rationale for insecureAcceptAnything (no pull path + verity-measured archives), correct generalization of measurementFor preserving existing outputs, and good test coverage for the new manifest flavor. The launch-topology inversion is cleanly implemented and the CRN launch path genuinely needs no changes. Minor nits only: an unused pre-created directory and a missing build-flavor-specific test, neither blocking.

**`nix/compose-rootfs.nix`** (line 101): The `rootfs/etc/containers/networks` directory is pre-created here but appears unused — CNI network configs are written to `rootfs/etc/cni/net.d/` later (line 206). This matches the donor and is harmless, but if the donor is ever cleaned up, this could too.

**`tests/vprogram/test_cli.py`** (line 105): The `test_manifest_compose_runtime` test runs `build` without `--flavor compose` (defaulting to `vprogram`). This works because the bundle-info format is identical, but a test that passes `--flavor compose` to `build` as well would verify the full compose-flavor CLI path end-to-end (even with `--image-dir` skipping the actual nix build).

foxpatch-aleph (approved):
Well-structured port of the Docker Compose runner runtime as an aleph.compose/1 V-PROGRAM flavor. The topology inversion (platform init as entrypoint, workload as data) is clean, fail-closed semantics are thorough at both init layers (14 poweroff sites in the outer init, 11 fatal sites in the platform init), and the security rationale for insecureAcceptAnything (no pull path + verity-measured archives) is sound. The measurementFor generalization is behavior-preserving with correct defaults, determinism is maintained via the same reproducible-mkfs recipe, and test coverage is comprehensive for the new Python/CLI paths. No correctness or security issues found.

**`nix/compose-rootfs.nix`** (line 239): The custom containers.conf does not specify conmon_path. This works because nixpkgs-built podman has the conmon path compiled in, but it creates an implicit dependency on nixpkgs' podman patching. Consider adding `conmon_path = "<nix-store-path>"` for robustness, though this is donor-identical behavior and not a blocker.

**`nix/init-compose.sh`** (line 200): The agent-start ordering (delta 6) is a subtle but correct design: starting the attest-agent before the chroot ensures `$!` captures the guest PID, not the agent PID. The header comment documents this well.

**`tests/vprogram/test_bundle.py`** (line 171): Consider adding an explicit test that `build_bundle(flavor="compose")` produces the same tarball byte layout as `flavor="vprogram"` given the same image_dir, to directly verify the claim in bundle.py's docstring that compose shares the vprogram path. The current test_build_then_manifest_compose_flavor implicitly verifies this by succeeding with a vprogram-style fixture, but an explicit byte-identity assertion would be clearer.

PR #1133, opened 2026-08-19
refactor(auth): drop the legacy X-Auth-Signature allocation auth path

odesenfans:
Removes the shared-bearer-token scheme superseded by `Aleph-EIP191-V1`: `_verify_legacy_token`, the `X-Auth-Signature` dispatch branch, `settings.ALLOCATION_TOKEN_HASH`, and the boot-time deprecation warning.

### Why now, beyond dead code

The default `ALLOCATION_TOKEN_HASH` shipped in the repo, so any CRN that never overrode it trusted a token whose hash is public — on the endpoints that start, stop and migrate VMs.

### Upgrade impact: none

`ALEPH_VM_ALLOCATION_TOKEN_HASH` left behind in an operator's `supervisor.env` is harmless. pydantic-settings ignores `ALEPH_VM_*` vars with no matching field, so the supervisor still starts; the variable simply grants nothing. `test_legacy_allocation_token_env_var_grants_no_access` pins that, driving the setting through the environment rather than through a Python attribute that no longer exists.

### Callers moved to signed requests

The deletion was the small part — CI and two docs still authenticated with the old header.

- **`scripts/sign_allocation_request.py`** (new) builds the `Authorization` header for CI, docs and manual operator checks. `test_sign_allocation_request_helper_authenticates` runs its output through the real verifier so the two encodings cannot drift apart silently.
- **CI** signs each `/control/allocations` call with a throwaway key authorized via `AUTHORIZED_ALLOCATION_SIGNERS`. Note the retry change: `curl --retry` would have resent an identical signed payload, which the supervisor rejects as a replay, so retries now re-sign in a shell loop.
- **Docs** (`doc/confidential.md`, `QEMU.md`) become working `curl` invocations instead of `.http` snippets carrying a token that no longer authenticates.

Tests that used the legacy header merely to get past auth now sign via a `scheduler_auth` fixture; the ones that existed only to prove the fallback worked are gone.

### Verification

- `tests/supervisor`: 1117 passed. The 10 failures are pre-existing environment-only ones (pyroute2 needs root, journald, dbus) — confirmed as the identical set on unmodified `dev`.
- `mypy`, `ruff format`, `isort`, `yamlfix`, and pinned `shellcheck` 0.11.0 all clean.

### DIFF
--- a/.github/scripts/test_runtime_on_droplet.sh
+++ b/.github/scripts/test_runtime_on_droplet.sh
@@ -58,9 +58,33 @@ if ! curl --retry 5 --retry-delay 10 --retry-connrefused --max-time 120 --fail "
 fi
 
 echo "==> Scheduling an instance via /control/allocations"
-curl --retry 5 --retry-delay 10 --retry-connrefused --max-time 120 --fail -X POST -H "Content-Type: application/json" \
-    -H "X-Auth-Signature: test" \
-    -d "{\"persistent_vms\": [], \"instances\": [\"${ITEM_HASH}\"]}" \
-    "http://${DROPLET_IPV4}:4020/control/allocations"
+# Throwaway CI signer, authorized via ALEPH_VM_AUTHORIZED_ALLOCATION_SIGNERS in
+# the install step. It only ever reaches a short-lived test droplet, so it is
+# checked in rather than kept as a secret. Signing happens per request: the
+# payload embeds `iat`, which the supervisor rejects once stale.
+ALLOCATION_SIGNER_KEY="0x4c0883a69102937d6231471b5dbb6204fe512961708279aeb8f1a4b0b8b0e5c1"
+ALLOCATION_BODY="{\"persistent_vms\": [], \"instances\": [\"${ITEM_HASH}\"]}"
+
+# Retry in the shell rather than with curl's --retry: the supervisor rejects a
+# replayed signed payload, so every attempt must carry a freshly signed header.
+allocation_attempt=0
+until [ $allocation_attempt -ge 5 ]; do
+    allocation_auth="$(printf '%s' "${ALLOCATION_BODY}" | \
+        python3 "$(dirname "$0")/../../scripts/sign_allocation_request.py" \
+            --key "${ALLOCATION_SIGNER_KEY}" --path /control/allocations)"
+    if curl --max-time 120 --fail -X POST -H "Content-Type: application/json" \
+        -H "Authorization: ${allocation_auth}" \
+        -d "${ALLOCATION_BODY}" \
+        "http://${DROPLET_IPV4}:4020/control/allocations"; then
+        break
+    fi
+    allocation_attempt=$((allocation_attempt + 1))
+    if [ $allocation_attempt -ge 5 ]; then
+        echo "==> ERROR: /control/allocations failed after ${allocation_attempt} attempts"
+        exit 1
+    fi
+    echo "==> /control/allocations attempt ${allocation_attempt} failed, retrying in 10s"
+    sleep 10
+done
 
 echo "==> Runtime ${ITEM_HASH} OK"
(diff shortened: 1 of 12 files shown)


foxpatch-aleph (approved):
This PR cleanly removes the insecure legacy X-Auth-Signature bearer-token auth path (whose default hash was public, allowing anyone to start/stop/migrate VMs on unconfigured CRNs) and replaces all callers with the existing Aleph-EIP191-V1 signed-request scheme. The verifier in allocation_auth.py is well-designed: replay protection is keyed by payload hash (not signature, preventing ECDSA malleability bypass), uses an asyncio.Lock for atomic check-and-record, caps body size before reading, requires Content-Length, forbids query strings, rejects unknown auth-params, and catches all exceptions to return False rather than 500. The new sign_allocation_request.py helper produces headers the real verifier accepts (pinned by test_sign_allocation_request_helper_authenticates). CI re-signs each retry in a shell loop instead of using curl --retry (which would replay identical payloads). The throwaway CI signing key is a well-known example key, deliberately committed for short-lived test droplets. ALLOCATION_TOKEN_HASH is fully removed from conf.py, supervisor.env, and all Python source; test_legacy_allocation_token_env_var_grants_no_access proves leftover env vars are harmless. Test coverage is thorough (50+ tests covering signatures, staleness, replays, concurrency, body binding, signer resolution chain). Docs updated with working curl examples. No bugs or security issues found.

**`src/aleph/vm/agent/views/allocation_auth.py`** (line 66): Minor: duplicate known auth-params (e.g. `sig=0xdead,sig=0xcafe`) are silently accepted — last value wins. Not a security issue since both values must produce a valid signature for the request to pass, but a `len(params) != len(set(...))` check would make the contract tighter and match the "reject unknowns" philosophy.

foxpatch-aleph (approved):
Clean, well-executed removal of the legacy X-Auth-Signature bearer-token auth path. The core change (dropping _verify_legacy_token, ALLOCATION_TOKEN_HASH, and the fallback dispatch) is correct and removes a real security issue: the default token hash was public, allowing anyone to control VM lifecycle endpoints. The new sign_allocation_request.py helper is well-documented and its output is verified against the real verifier in tests. The CI migration correctly re-signs per retry attempt to avoid replay rejection. The scheduler_auth fixture is a good DRY abstraction applied consistently across both test files. The test_legacy_allocation_token_env_var_grants_no_access test correctly pins that a leftover env var is harmless. Docs are updated to working curl invocations.

**`tests/supervisor/test_views.py`** (line 1257): The test patches `allocation_auth.settings` with a fresh `Settings()` but does not patch `utils.settings` (which `get_authorized_allocation_signers` uses). This happens to be fine because the request only carries `X-Auth-Signature` (no `Authorization` header), so `authenticate_api_request` returns False at the first check before ever calling `get_authorized_allocation_signers`. If the test intent is to verify the full request path rejects the legacy header, consider adding a brief comment noting this so a future maintainer doesn't add an `Authorization` header and get confused by a pass caused by the unpatched signer list.

**`scripts/sign_allocation_request.py`** (line 58): Minor: the `--body-file ""` option to sign an empty body is a nice touch for bodyless endpoints like `/control/proxy/regenerate`, but it's not exercised by any test or shown in docs. Consider adding a one-line example in the docstring, or it may go unnoticed by operators who need it.

foxpatch-aleph (approved):
Clean removal of the deprecated X-Auth-Signature / ALLOCATION_TOKEN_HASH authentication path, replaced entirely by the Aleph-EIP191-V1 signature scheme. The security motivation is sound: the default ALLOCATION_TOKEN_HASH was public, so any CRN that never overrode it trusted a known token for VM lifecycle endpoints. The new sign_allocation_request.py helper is tested against the real verifier to prevent encoding drift, CI re-signs per retry attempt to avoid replay rejection, and test_legacy_allocation_token_env_var_grants_no_access pins that leftover operator env vars grant nothing. All source, docs, and CI references are updated consistently. No bugs or security issues found.

**`tests/supervisor/test_vprogram.py`** (line 49): The `_signed_allocation` helper duplicates the payload-building logic from `_make_aleph_eip191_v1_header` in test_views.py. Consider importing and reusing that function to keep a single encoding implementation across test files, but this is a minor maintainability nit and non-blocking.

PR #1134, opened 2026-08-19
chore: dead imports, dead locals, and a broken __all__

odesenfans:
Cleanups surfaced by `ruff check --select F401,F811,F841`. The CI style step gates only `ruff format` and `isort`, so these lint rules never ran and had quietly accumulated.

### The one that was not cosmetic

`src/aleph/vm/supervisor/controllers/qemu/__init__.py` set `__all__` to a bare string rather than a list:

```python
__all__ = "AlephQemuInstance"
```

`import *` iterates `__all__`, so it walked the characters and raised `AttributeError: module 'aleph.vm.supervisor.controllers.qemu' has no attribute 'A'`. Nothing star-imports the package today, which is why it went unnoticed.

### A small test gap

`test_setup_nftables_for_vm_ipv6_disabled` patched `build_postrouting_chain_entities` and `build_masquerading_rule_entities` but asserted on neither, so half of what "only IPv4 calls - no ip6" claims to check went unverified. Asserting on them (rather than dropping the unused bindings) closes that; both do get called once, so the test passes as strengthened.

### The rest

An unused `time` import, an unused `Optional`, two dead `loop = asyncio.get_event_loop()` assignments in the test harness, four unused test imports, and a pointless assignment of a `receive_json` result on a line only ever reached by raising.

### Verification

- `tests/supervisor`: 1119 passed. The 10 failures are pre-existing environment-only ones (pyroute2 needs root, journald, dbus) — the identical set on unmodified `dev`.
- `ruff check --select F401,F811,F841` now clean across `src/`, `tests/`, `scripts/`; `mypy`, `ruff format`, `isort` clean.

### DIFF
--- a/src/aleph/vm/agent/migration/helpers.py
+++ b/src/aleph/vm/agent/migration/helpers.py
@@ -4,7 +4,6 @@ import asyncio
 import hashlib
 import logging
 import shutil
-import time
 from pathlib import Path
 
 import aiohttp
--- a/src/aleph/vm/supervisor/controllers/qemu/__init__.py
+++ b/src/aleph/vm/supervisor/controllers/qemu/__init__.py
@@ -1,3 +1,3 @@
 from .instance import AlephQemuInstance
 
-__all__ = "AlephQemuInstance"
+__all__ = ["AlephQemuInstance"]
--- a/src/aleph/vm/testing/harness.py
+++ b/src/aleph/vm/testing/harness.py
@@ -72,7 +72,6 @@ async def benchmark(runs: int):
 
     bench: list[float] = []
 
-    loop = asyncio.get_event_loop()
     pool = VmPool()
     await pool.setup()
     bench_supervisor = LocalSupervisor(pool)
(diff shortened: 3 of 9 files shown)


foxpatch-aleph (approved):
A tightly-scoped cleanup PR that removes genuinely dead imports/locals and fixes a latent `__all__` bug (bare string instead of list) in the qemu controllers package. Every removal was verified as unused in its file, and the two `loop = asyncio.get_event_loop()` deletions in the test harness have no downstream references. The firewall test change strengthens coverage by asserting on two previously-patched-but-unchecked mocks, mirroring the IPv6-enabled test; both mocks are called exactly once in the IPv4-disabled path, so the assertions pass. The `receive_json` assignment drop in test_operator preserves behavior since the `await` still executes inside `pytest.raises`. No correctness, security, or design concerns.

foxpatch-aleph (approved):
Clean, well-scoped dead-code cleanup PR. All removals are verified as genuinely unused (imports, locals, assignments). The __all__ fix in qemu/__init__.py is a real bug fix — a bare string would cause AttributeError on star-import. The firewall test strengthening is a good call: the mocks were already patched but unasserted, and adding the two assert_called_once() calls closes the gap the PR description identifies. No correctness, security, or quality concerns.

PR #1115, opened 2026-08-14
docs: current-architecture doc set

odesenfans:
Cleanup task 3, PR 1 of 2: a coherent current-architecture doc set under \`docs/architecture/\`, replacing the 45 dated plan documents (deleted in the follow-up PR).

Nine docs: \`overview\` (system map + reading order), \`process-model\`, \`wire-contract\`, \`vm-lifecycle\`, \`networking\`, \`storage\`, \`confidential\`, \`testing\`, and \`divergences\` (the rust-port registry promoted verbatim, entry numbers stable).

Method: written from the code on dev @ b2b31381 as ground truth (every doc carries a \`Verified against\` header); the old plans were mined for decision rationale only. Every doc passed an adversarial accuracy review (claims fact-checked against the code; ~10 code-contradicted claims caught and fixed across the set, including a cross-doc contradiction the final whole-set review caught), and pointer paths are mechanically verified to resolve.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1135, opened 2026-08-19
docs: analysis of removing the CRN-side payment checks

odesenfans:
Docs-only. Writes up whether the CRN's payment enforcement duplicates what happens upstream, and splits the removal into two shippable parts. No behaviour change.

### Hold and credit: already enforced upstream, removable

pyaleph removes the instance message once the owner's balance no longer covers it, so a message still `processed` is by construction one whose balance is sufficient. That is why the scheduler passes both tiers straight through (`payment_gate.rs:174`).

The part worth knowing before deleting anything: **that decision already reaches the node, but not through the balance arithmetic.** It arrives via `check_payment`'s terminal-status loop, since `MessageStatus.REMOVED` is in `_TERMINAL_STATUSES`. So the balance branches are duplicative, but the terminal-status loop is load-bearing and has to stay.

### PAYG: the genuine exception

pyaleph does not check superfluid streams, so "message exists implies paid" does not hold there. The arithmetic on the CRN and scheduler sides does match (same price endpoint, same 0.2 ratio, same split formula, same pop-last-until-covered policy — table in the doc), but the CRN cannot act on the scheduler's PAYG decisions today: the `/control/allocations` stop loop excludes stream-paid VMs.

Hence the two parts:

1. **Allocate and delete PAYG instances via `/control/allocations`** — opens that channel. Behaviour-preserving while `check_payment` still runs.
2. **Delete the payment checks** — hold/credit and superfluid branches of both `notify_allocation` and `check_payment`, keeping the terminal-status loop.

### Two PAYG items to settle before part 2

- **31 of 152 live PAYG instances carry no `node_hash`**, and the scheduler's payment gate skips those outright. 14 are recent (2025-11 → 2026-02). Part 2 would silently un-enforce them.
- **The community 80/20 split is classified by three different clocks**: scheduler by message time, `check_payment` by VM boot time, `notify_allocation` by now. This one runs in the owner's favour to fix: 21 live pre-split instances (18 on a single CRN) get stopped by the CRN on restart under a rule they predate.

Live figures are from `api3.aleph.im` on 2026-08-19 (832 INSTANCE messages, 152 superfluid, all `processed`); the doc records the exact queries so they can be re-run.

### DIFF
--- a/docs/architecture/payment-enforcement.md
+++ b/docs/architecture/payment-enforcement.md
@@ -1,0 +1,180 @@
+# Payment enforcement
+
+> Verified against: 39ec840a (2026-08-19)
+
+## What this covers
+
+Who enforces that a running VM is paid for, and where. Three components have a
+say — pyaleph, the scheduler, and the CRN itself — and they cover different
+payment tiers by different mechanisms. Also records why the CRN's own checks are
+largely duplicative, and the two-part plan for removing them.
+
+## The question this answers
+
+The CRN runs its own payment enforcement in two places. If it duplicates checks
+already made upstream, the CRN's copy is code we can delete.
+
+| Where | What it does |
+| --- | --- |
+| `notify_allocation` (`src/aleph/vm/agent/views/__init__.py:820`) | admission check, per payment tier, at allocation time |
+| `check_payment` (`src/aleph/vm/agent/tasks.py:360`) | every `PAYMENT_MONITOR_INTERVAL` (60s): stops VMs whose message went terminal, then re-checks balances and streams |
+
+Upstream there are two enforcement points, not one:
+
+- **pyaleph (CCN)** enforces hold-tier and credit balances by *removing the
+  instance message* when the owner's balance no longer covers it.
+- **the scheduler** (`aleph-vm-scheduler`, `scheduler-rs/src/actors/payg/`)
+  validates superfluid streams.
+
+## Finding 1: hold and credit are already enforced, upstream of both
+
+pyaleph auto-removes instance messages once the owner's balance is
+insufficient. So a message that is still retrievable and `processed` from
+`/api/v0/messages.json` is, by construction, a message whose balance is
+sufficient. The scheduler relies on exactly this — `payment_gate.rs:174` passes
+hold and credit VMs straight through with no balance lookup, and
+`message_watcher.rs:255` documents the credit case as "managed by CCN API, pass
+through without validation".
+
+**That enforcement already reaches the CRN, and not via the balance
+arithmetic.** `check_payment`'s *first* loop (`tasks.py:373-407`) polls each
+running VM's message status and stops it after `STOP_AFTER_CONFIRMATIONS` = 3
+consecutive terminal readings, where `_TERMINAL_STATUSES` includes
+`MessageStatus.REMOVED` — precisely what pyaleph sets. The balance branches
+that follow it (`tasks.py:412` hold, `tasks.py:437` credit) re-derive a
+conclusion the CCN has already drawn and published.
+
+So for hold and credit the CRN checks are duplicative and removable. Two
+consequences:
+
+- The terminal-status loop is **load-bearing and must stay**. It is the channel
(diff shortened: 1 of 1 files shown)


foxpatch-aleph (approved):
This is a docs-only PR adding a single analysis document (`docs/plans/2026-08-19-crn-payment-check-removal-analysis.md`, 174 lines) that examines whether CRN-side payment checks duplicate upstream enforcement. The document is well-structured and its technical claims are accurate: I verified against the actual source that `_TERMINAL_STATUSES` includes `MessageStatus.REMOVED`, the hold branch is narrowed to confidential-only at tasks.py:414, the stop loop at views/__init__.py:597 excludes stream-paid VMs, and the three-clock divergence on the community split (message time vs boot time vs now) is real. Line references are off by ~2 in two places but this is trivial. The two-part removal plan is sound: opening the `/control/allocations` channel for PAYG before deleting the local checks is the correct ordering, and the callouts on the 31 node_hash-less instances and the clock disagreement are genuinely important preconditions. No code changes, no security surface, no behaviour change.

**`docs/plans/2026-08-19-crn-payment-check-removal-analysis.md`** (line 48): Line reference `notify_allocation:866` for the hold check is off by 2 — the actual guard is at views/__init__.py:868 (`if payment_type == PaymentType.hold and (is_confidential or have_gpu):`). Similarly the stop-loop citation `584-600` is closer to 583-613. Minor, but since the doc is meant to be re-checked against source, consider noting that line numbers may drift.

**`docs/plans/2026-08-19-crn-payment-check-removal-analysis.md`** (line 68): The `>` vs `>=` buffer comparison difference is noted as differing "only at exact equality." Worth stating explicitly which side uses which: CRN `check_payment` uses `>` (tasks.py:423: `balance < (required_balance + PAYMENT_BUFFER)` is `>`-equivalent for the stop condition) and the scheduler uses `>=` (`crn_stream + buffer >= required`). At exact equality the CRN would stop while the scheduler would not — a divergence direction worth recording for part 1, since it means the CRN is currently slightly stricter.

odesenfans:
Rebased onto current `dev`. One thing changed materially beyond conflict resolution, so flagging it here:

`docs/plans/` no longer exists — it was purged in the docs cleanup (#1115/#1116), and `docs/architecture/` now holds curated, undated topic docs. Rather than resurrect the deleted directory, the file landed as **`docs/architecture/payment-enforcement.md`** and picked up the house header (`> Verified against:` / `## What this covers`). Happy to move it elsewhere if you'd rather this kind of decision record not sit alongside the subsystem references.

Line references were re-derived against `39ec840a` (they had drifted a couple of lines).

Separately, and not fixed in this rebase: finding 4's "31 of 152 **live** PAYG instances" uses message status for "live". The scheduler API shows only 11 of the 152 are actually observed running, so that phrasing overstates the population — the conclusion it drives is unchanged, and unpinned PAYG is out of scope by design either way. Say the word and I'll tighten it.

foxpatch-aleph (approved):
A well-researched, docs-only architecture analysis that correctly identifies the CRN's payment checks as largely duplicative of upstream enforcement, with PAYG as the genuine exception. All verifiable code references (line numbers, constants, status enums, buffer comparisons, clock sources) are accurate against the current tree. The two-part removal plan is logically sound: opening the scheduler-to-CRN channel for PAYG before deleting the local checks. The live-network data is dated and reproducible. No code changes, no security concerns, no testing needed.

**`docs/architecture/payment-enforcement.md`** (line 136): The reference `check_payment:460ff` points to the start of the superfluid branch; the actual `is_after_community_wallet_start(_dt_from_ns(i.started_at_ns))` call is at tasks.py:485. Consider making the reference more precise (e.g., `check_payment:485ff`) so a reader lands directly on the clock call rather than 25 lines of stream-fetch boilerplate.

**`docs/architecture/payment-enforcement.md`** (line 137): Similarly, `notify_allocation:884ff` is the start of the superfluid branch, but the `is_after_community_wallet_start()` call with no argument is at line 908. The `884ff` reference is technically correct (from 884 onward) but a reader scanning for the clock usage has to read past 24 lines to find it.

PR #1136, opened 2026-08-19
feat(agent): make PAYG instances scheduler-owned on /control/allocations

odesenfans:
**Part 1 of two** for removing the CRN-side payment checks (analysis in #1135). This one opens the channel; it deletes no check yet: `check_payment` still runs exactly as before. Rewritten from the tip of `dev-2.2` on 2026-09-15, since the rule it changes moved into `agent/allocation/teardown.py` in 2.1.

### The problem

The scheduler's payment gate validates Superfluid streams and drops unpaid PAYG instances from the plan. That decision could not reach the node: the retention rule behind the `/control/allocations` stop loop and the v2 reconciler kept every stream-paid VM, so the CRN's own `check_payment` was the only thing that could ever stop a PAYG VM.

This makes PAYG scheduler-owned, the way v-programs already are: absence from the allocation stops it. The allocation verdict follows: a stream-paid VM the plan drops is reported as `removing`, no longer retained for `payment_stream`.

### GPU and confidential PAYG are included deliberately

The GPU and confidential exclusions in `retention_reason` exist for hold-tier VMs the scheduler does not place. **5 of the 9 scheduled live PAYG instances carried a GPU** when this was measured, so leaving those exclusions applied to PAYG would put most of the fleet outside scheduler control and force part 1 to be revisited.

### The allocating half already worked

`update_allocations` starts everything in `allocation.instances` via `start_persistent_vm` with no payment-tier gating; the checks live only in `notify_allocation`. Pinned by a test rather than changed.

### Blast radius

Checked against `scheduler.api.aleph.cloud/api/v1/vms` and `/api/v1/nodes` on 2026-08-19. Of 152 live PAYG instance messages:

| | count | known to scheduler | observed running |
|---|---:|---:|---:|
| pinned to a CRN | 121 | 121 | 11 |
| unpinned | 31 | 0 | — |

The 110 pinned-but-unobserved have `observed_nodes: null`: stale messages, not live VMs.

Of the 11 observed running, 9 are `scheduled` and stay in the plan, so they are unaffected. The remaining 2 are `unschedulable` with `payment_status: validated`, both on node `31c1035d41c79d87…` which the scheduler marks `removed`, last observed 2026-06-04. So the practical blast radius is 0–2 VMs on a decommissioned node.

The 31 unpinned PAYG messages are absent from the scheduler entirely. Unpinned PAYG is not a supported configuration, so they are out of scope by design.

### Tests

The three assertions that pinned "PAYG is spared" are inverted; the rest is new.

- `test_allocation_teardown.py`: `payment_stream` leaves the retained-reason table; `test_a_stream_paid_vm_is_the_schedulers_to_stop` covers plain, GPU and confidential.
- `test_allocation_verdict.py`: `test_a_stream_paid_vm_absent_from_the_plan_is_removed` (was `..._is_retained_with_its_reason`).
- `test_allocation_reconciler.py`: `test_a_stream_paid_vm_the_plan_dropped_is_torn_down` (was `..._is_never_torn_down`).
- `views/test_allocations_retire.py`: `test_stop_loop_stops_a_payg_vm_absent_from_the_allocation` (plain, GPU, confidential) and `test_update_allocations_starts_a_payg_instance`.

### Verification

`ruff format`, `isort` and byte-compilation clean locally; the suite runs on CI.

### DIFF
--- a/src/aleph/vm/agent/views/__init__.py
+++ b/src/aleph/vm/agent/views/__init__.py
@@ -593,9 +593,17 @@ async def update_allocations(request: web.Request):
                     # v-programs: absence from the allocation stops them,
                     # even though they are credit-paid and confidential.
                     record.is_vprogram
+                    # ...and for PAYG. The scheduler's payment gate validates the
+                    # Superfluid stream and drops unpaid instances from the plan;
+                    # honouring that absence here is what makes the decision
+                    # effective. GPU-bearing and confidential PAYG instances are
+                    # included deliberately: those exclusions below exist for
+                    # hold-tier VMs the scheduler does not place, and most of the
+                    # live PAYG fleet carries a GPU, so applying them to PAYG
+                    # would leave the bulk of it outside scheduler control.
+                    or record.uses_payment_stream
                     or (
-                        not record.uses_payment_stream
-                        and not record.uses_payment_credit
+                        not record.uses_payment_credit
                         and not info.gpus
                         and info.confidential_mode is ConfidentialMode.NONE
                     )
--- a/tests/supervisor/test_views.py
+++ b/tests/supervisor/test_views.py
@@ -1465,10 +1465,12 @@ async def test_v2_executions_list_omits_ghost_mapped_ports(aiohttp_client, mocke
 
 
 @pytest.mark.asyncio
-async def test_update_allocations_spares_payg_via_registry(aiohttp_client):
-    """A stream-paid registry record must be spared by the stop loop even when absent
-    from the allocation (the restored-PAYG case).
-    Status is read from supervisor.list_vms(); payment tier from the registry."""
+async def test_update_allocations_stops_payg_via_registry(aiohttp_client, mocker):
+    """A stream-paid registry record absent from the allocation must be stopped.
+
+    Payment tier is knowable only through the registry here: spec-built and
+    restart-restored VMs carry no hypervisor-side message, so the stop loop
+    reads status from supervisor.list_vms() and the tier from the record."""
     from aleph.vm.supervisor_interface.types import (
         Backend,
         ConfidentialMode,
@@ -1521,6 +1523,7 @@ async def test_update_allocations_spares_payg_via_registry(aiohttp_client):
 
     fake_supervisor = MagicMock(delete_vm=AsyncMock(), list_vms=AsyncMock(return_value=[vm_info]))
     app["supervisor"] = fake_supervisor
+    mocker.patch("aleph.vm.agent.views.delete_records_for_vm", new_callable=AsyncMock)
 
     settings.ALLOCATION_TOKEN_HASH = "9f86d081884c7d659a2feaa0c55ad015a3bf4f1b2b0b822cd15d6c15b0f00a08"  # = "test"
     client = await aiohttp_client(app)
@@ -1532,9 +1535,9 @@ async def test_update_allocations_spares_payg_via_registry(aiohttp_client):
     )
     assert response.status == 200
     resp_json = await response.json()
-    assert vm_hash not in resp_json["stopped"]
-    fake_supervisor.delete_vm.assert_not_awaited()
-    assert ItemHash(vm_hash) in app["vm_registry"]
+    assert vm_hash in resp_json["stopped"]
+    fake_supervisor.delete_vm.assert_awaited_once_with(VmId(vm_hash))
+    assert ItemHash(vm_hash) not in app["vm_registry"]
 
 
 @pytest.mark.asyncio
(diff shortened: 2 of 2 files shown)


foxpatch-aleph (approved):
The core change correctly makes PAYG (stream-paid) instances scheduler-owned by adding `record.uses_payment_stream` to the stop-loop condition in `update_allocations`, so absence from the allocation now stops PAYG VMs regardless of GPU or confidential status. The logic is sound: v-programs remain always scheduler-owned, credit-paid and hold-tier GPU/confidential VMs keep their existing exclusions, and `notify_allocation` retains all payment checks unchanged. The four TDD tests (inverted from 'spares' to 'stops' plus a new start-path pin) provide good coverage of the behavior change. The Rust attest-agent secret injection is well-hardened. Two non-blocking observations: `ItemHash(info.vm_id)` before the registry lookup could crash on a non-item-hash VmId (pre-existing single-tenant assumption), and a non-VmNotFoundError from `delete_vm` skips registry cleanup (pre-existing pattern).

**`src/aleph/vm/agent/views/__init__.py`** (line 584): `ItemHash(info.vm_id)` is constructed before `registry.get(vm_hash)` checks for a record. If a supervisor ever returns a VmId that is not a valid item hash (e.g., a multi-tenant scenario as noted in `list_executions`), this raises `ValueError` and crashes the entire stop loop, not just that iteration. Consider wrapping in a try/except or validating before constructing, though this is a pre-existing single-tenant assumption documented elsewhere in the file.

**`src/aleph/vm/agent/views/__init__.py`** (line 614): If `supervisor.delete_vm` raises a non-`VmNotFoundError` (e.g., `InternalSupervisorError`), the exception propagates and `registry.forget` / `delete_records_for_vm` / `remove_vprogram_staging` are skipped, leaving stale registry and metrics state. On the next allocation cycle the stop loop would retry. This is a pre-existing pattern but worth noting for robustness — consider whether catching `SupervisorError` (not just `VmNotFoundError`) and still cleaning up local state would be safer.

foxpatch-aleph (approved):
The one-line logic change correctly hoists `uses_payment_stream` from a negated conjunct to a top-level disjunct, making PAYG instances scheduler-owned: absence from `/control/allocations` now stops them even when they carry a GPU or are confidential. I verified all six payment-tier × GPU/confidential combinations and confirmed hold-tier and credit-tier behavior is unchanged, so there is no regression risk for non-PAYG VMs. The start path was already ungated (no payment logic in `run.py`/`start_persistent_vm`), and the new test pins that. `check_payment` in `tasks.py` is deliberately left intact as a belt-and-suspenders stop path until part 2, so no safety net is removed prematurely. The four test mutations match the PR description: the previously-'spares' assertion is inverted, the stream-paid case is removed from the spare parametrization, a new 3-case parametrized test covers plain/GPU/confidential PAYG stopping, and a new test pins the allocating half. The renamed test correctly adds a `delete_records_for_vm` mock that the old 'spares' version didn't need (since the stop path is now actually exercised). The inline comment explaining why GPU-bearing and confidential PAYG are deliberately included is thorough and captures the non-obvious rationale. The only nit is a now-dead `registry_kwargs.get("stream")` dispatch branch in `test_stop_loop_spares_ineligible_vms` left behind after removing the stream-paid parametrize case — harmless but slightly misleading given the docstring says stream-paid is deliberately absent.

**`tests/supervisor/test_views.py`** (line 1762): Minor nit: the `if registry_kwargs.get("stream"): content = _STREAM_INSTANCE_CONTENT` branch is now dead code — the stream-paid parametrize case was removed, so no entry sets `stream: True`. The `elif registry_kwargs.get("credit")` / `else` pair covers all remaining cases. Consider dropping the `stream` branch (and the `content`/`_STREAM_INSTANCE_CONTENT` selection) to match the docstring, which already says stream-paid is deliberately absent from this test. Not blocking.

foxpatch-aleph (approved):
A small, well-placed change: folding uses_payment_stream into the v-program branch of retention_reason makes PAYG instances scheduler-owned on every allocation path at once (legacy stop loop, v2 verdict, v2 reconciler), since all three consume the shared retention rule. The logic is correct, introduces no new races (the reconciler's re-read of the desired plan before teardown is untouched), and check_payment still runs unchanged as claimed, converging on the same idempotent retire_vm(GONE). Tests are inverted and extended appropriately, including the GPU and confidential PAYG cases that motivated the change and a pin on the start path the new stop semantics now depend on. All 122 tests across the four touched files plus the payment-related suites (156 total) pass locally; the remaining suite failures are environmental (/var/lib/aleph permission paths), unrelated to this diff. The only gap is documentation: doc/scheduler-allocations.md still lists payment_stream as a possible retained reason even though the node can no longer return it.

**`doc/scheduler-allocations.md`** (line 55): This list still names `payment_stream` as a possible `retained` reason, and the following paragraph carves out only the V-PROGRAM. After this change the node never reports `payment_stream`, so the reason should be dropped here and the paragraph extended to say a stream-paid (PAYG) instance is scheduler-owned like a v-program: the plan's payment gate validates the stream, so absence from a plan stops it — GPU-bearing and confidential PAYG included, since those exclusions exist for hold-tier VMs the scheduler does not place. Worth updating in this PR (or explicitly in part 2) so the documented API contract matches the code.

PR #1138, opened 2026-08-19
feat(agent): the scheduler owns VM lifecycle; drop all node-side payment checks

odesenfans:
**Part 2 of two.** Stacked on #1136; review that first. Analysis in #1135. Rewritten from the tip of `dev-2.2` on 2026-09-15: since the August version, #1253 gave the payment sweep the consecutive-sweep confirmation and `retire_vm(GONE)` semantics, and the stop-loop rule moved into `agent/allocation/teardown.py`. All of that goes with the sweep.

The node stops deciding which VMs deserve to run and simply matches the allocation.

### What goes

- **`check_payment` and `monitor_payments` entirely**, including the terminal-status loop that polled each VM's message status and the consecutive-sweep confirmation that guarded both. Message deletion is the scheduler's signal now: it diffs a full message snapshot against `known_vms` (`message_watcher.rs`), so a forgotten or removed message becomes a delete, leaves the plan, and reaches the node as absence from the allocation.
- **`notify_allocation`'s per-tier admission checks**: hold balance, credit balance, Superfluid stream and community-wallet split, and the payment-receiver match.
- **`payment.py`**, `get_message_status` and its cache, the community-wallet helpers in `agent/utils.py`, `get_chain`, the `AgentVmRecord.uses_payment_*` properties, and the `PAYMENT_MONITOR_INTERVAL` / `PAYMENT_BUFFER` / `PAYMENT_PRICING_AGGREGATE` / `CACHE_TTL_ADDRESS_BALANCE` / `CACHE_TTL_EXECUTION_PRICE` / `CACHE_TTL_MESSAGE_STATUS` settings.
- **The `aleph-superfluid` dependency** (pyproject and the packaging Makefile mirror): only the stream check imported it.

The retention rule behind the legacy stop loop and the v2 reconciler loses its payment-tier, GPU and confidential exclusions with them, and its `info` parameter, which nothing reads any more. Absence from the allocation now stops any persistent, running, registry-known VM, and the verdict reports it as `removing`. The record check stays: a VM the agent has no record of isn't scheduler-managed and belongs to the idle-expiry path. So does `non_persistent`, the one retention reason left.

### Why this is safe

Enforcement upstream is unchanged and is what carries it. pyaleph removes instance messages whose balance no longer covers them; the scheduler's payment gate validates Superfluid streams.

Measured against the live network in August: of 180 observed-running VMs absent from the plan, **168 were already dead on the CCN** (144 forgotten, 21 removed, 3 removing). The 12 alive ones were 3 CRN self-diagnostic VMs (`CHECK_FASTAPI_VM_ID` and friends, still protected by the record/persistent checks), 7 hold-tier VMs that were in fact allocated (so `scheduling_status` is not a reliable proxy for allocation membership; treat 180 as an upper bound), and the 2 PAYG instances on a decommissioned node.

### Two consequences worth naming

- **`/status/config` no longer advertises `PAYMENT_MONITOR_INTERVAL`.** It described a monitor that no longer exists. `PAYMENT_RECEIVER_ADDRESS` and `AVAILABLE_PAYMENTS` stay; clients need them to open a stream.
- **A node that falls out of the scheduler's plan gets an empty allocation and now stops everything on it.** That is the accepted cost of making the scheduler authoritative; no node-side guard was added.

### What this does NOT fix

The `removing` gap. The scheduler polls messages with no status filter, so a message stuck in `MessageStatus.REMOVING` stays in its snapshot exactly as it stayed outside the node's old terminal set. Root cause is on the pyaleph side.

### Tests

- `test_allocation_teardown.py` and `test_allocation_verdict.py`: credit, GPU and confidential leave the retained-reason tables; `test_payment_tier_gpus_and_confidential_mode_do_not_keep_a_vm_the_plan_dropped` pins the inverse, and `non_persistent` is the one reason still reported.
- `views/test_allocation_notify.py`: `test_notify_allocation_does_not_check_payment`, no balance/price/stream mock in the way of the create call.
- Deleted `test_checkpayment.py`, `test_payment.py`, the payment-grouping tests in `test_tasks_registry_reads.py`, `test_payment_sweep_excludes_vprograms` and `test_record_payment_helpers`.

Docs: the "payment shortfall" paragraphs in `docs/architecture/storage.md` and `vm-lifecycle.md`, and the stop-guard description in `confidential.md`, now describe the deallocation path.

### Verification

`ruff format`, `isort` and byte-compilation clean locally; the suite runs on CI.


foxpatch-aleph (approved):
This PR cleanly removes the node-side payment checking infrastructure (check_payment, monitor_payments, payment.py, community wallet helpers, and related settings) and simplifies the stop-loop to trust the scheduler's allocation as the single source of truth. The core logic change in update_allocations is correct: any running, persistent, registry-known VM absent from the allocation is now stopped regardless of payment tier, GPU, or confidential mode. The notify_allocation endpoint's payment admission checks are removed and replaced with an explanatory comment. All imports and references to deleted code are clean — no orphaned dependencies. Test coverage is strong: the new parametrized test covers 8 cases across all payment tiers and hardware configurations, the ineligible-VM test is correctly reduced to its one remaining case, and the notify_allocation test cleverly uses a receiver-mismatch probe to verify the payment branch is gone without mocking internals. Two minor observations: a stale comment in local.py references the removed payment sweep/monitor_payments, and get_message_status plus its CACHE_TTL_MESSAGE_STATUS setting are now dead code (likely retained for the future removing-gap fix named in the PR description). Neither is a blocker.

**`src/aleph/vm/supervisor/local.py`** (line 455): Stale comment: references "the payment sweep" and "monitor_payments" which no longer exist after this PR. Consider updating to reflect that list_vms is now called by update_allocations and the /about/executions/list endpoints.

**`src/aleph/vm/agent/messages.py`** (line 84): get_message_status is now dead code — its only caller (check_payment's terminal-status loop in tasks.py) was removed by this PR. The supporting CACHE_TTL_MESSAGE_STATUS setting in conf.py is also unused. Likely retained for the "removing gap" fix described in the PR description, but worth noting.

foxpatch-aleph (commented):
The removal is complete and clean: no dangling references to any deleted symbol across src, tests, docs, or packaging, the single-arg retention_reason is consistently applied at all three call sites, and the legacy endpoint and v2 reconciler still share one rule so they cannot disagree. The new retention rule (persistent-only) is correctly mirrored by the verdict, and test coverage for the new behavior (teardown reason table, verdict inverse pinning, notify-without-payment, PAYG retire) is solid. Both design risks are explicitly acknowledged in the PR description: no node-side guard against an empty allocation (a scheduler outage or misconfiguration now retires every running persistent VM as GONE and purges disks under default reap, losing the three-sweep confirmation that previously protected running VMs from a single bad API answer), and the unauthenticated /control/allocation/notify endpoint now performs no payment check at admission. Neither is an implementation bug, but both deserve a follow-up guard and an operator-facing runbook note. All 146 relevant tests pass locally (two pre-existing test_vprogram.py failures are environmental: the sandbox aleph-message version lacks VerifiableProgramContent.gpu).

**`src/aleph/vm/agent/views/__init__.py`** (line 669): This is the biggest operational risk of the change, and the PR acknowledges it: a scheduler outage or misconfiguration that yields an empty allocation now retires every running persistent VM on the node as GONE, purging disks under the default VOLUME_RETENTION=reap. The old sweep required three consecutive confirmations before touching a running VM, which protected against a single bad API answer; the allocation path has no equivalent. Please consider a follow-up guard here — e.g., refuse to tear down when the incoming allocation names zero VMs while the supervisor is running more than a small threshold — or at least document the operational expectation for node operators running reap.

**`src/aleph/vm/agent/views/__init__.py`** (line 931): With the payment checks gone, this endpoint (which has no request authentication, only @cors_allow_all) now starts any instance message with no admission check beyond capacity. That is the intended design — the sweep reaps anything the plan drops — but it does shift the whole burden onto network isolation of the control interface. Worth an explicit note in the deployment docs that /control/allocation/notify must not be reachable off the management network.

**`tests/supervisor/test_allocation_verdict.py`** (line 41): The _record() helpers here and in tests/supervisor/test_allocation_reconciler.py:44 still set uses_payment_stream/uses_payment_credit and take now-dead stream/credit kwargs referencing the removed AgentVmRecord properties. Harmless at runtime (SimpleNamespace), but stale — the fake record now only needs persistent, message and is_vprogram.

**`tests/supervisor/views/test_allocations_retire.py`** (line 162): The stop-loop tests cover a plan that drops one VM, but nothing pins the empty-allocation-stops-everything behavior that this PR accepts as its cost. A test documenting the current semantics (or guarding them, per the comment on views/__init__.py:669) would make the trade-off explicit and protect against an accidental future regression in either direction.

foxpatch-aleph (commented):
Clean, complete removal of the node-side payment machinery: all call sites of the changed retention predicates are consistent, no dangling references to any removed symbol, setting, or module exist anywhere in src/tests/docs/packaging, and the tree byte-compiles. The behavioral inversion (absence from the allocation stops any persistent, registry-known VM regardless of tier/GPU/confidential mode) is a deliberate, well-documented design decision, and the rewritten tests pin it correctly. Three non-blocking observations: the now-dead to_normalized_address helper in utils, stale uses_payment_* kwargs in two test helpers that make some tests pass vacuously, and the residual billing exposure of the unauthenticated /control/allocation/notify endpoint, now gated only by scheduler sweep cadence.

**`src/aleph/vm/utils/__init__.py`** (line 248): to_normalized_address is now dead code: its only caller was the deleted agent/payment.py. The eth_typing and eth_utils imports at lines 26-27 (HexAddress, HexStr, hexstr_if_str, is_address, to_hex) now serve only this dead function. Since the PR's purpose is removing the payment machinery, this helper and those imports should go with it.

**`tests/supervisor/test_allocation_verdict.py`** (line 41): The _record helpers here and in tests/supervisor/test_allocation_reconciler.py:44 still accept stream/credit kwargs and set uses_payment_stream / uses_payment_credit — properties that no longer exist on AgentVmRecord and are never read by the code under test. Tests like test_a_stream_paid_vm_absent_from_the_plan_is_removed now pass vacuously (any record would do). Consider dropping the kwargs, or keeping one such test with a comment that the tier is deliberately irrelevant now.

**`src/aleph/vm/agent/views/__init__.py`** (line 931): Worth stating explicitly: /control/allocation/notify has no authentication, and this PR removes its last admission gate, so anyone with a valid instance message can start a VM on any node for free until the next allocation push omits it. The PR accepts this, but the residual exposure equals the scheduler's sweep interval — worth confirming that interval is short relative to what a running VM can consume in that window (bandwidth, egress).

PR #1139, opened 2026-08-19
refactor: move guest IPv6 allocation from the supervisor to the agent

odesenfans:
Moves guest **IPv6 allocation out of the supervisor and into the agent**, so the supervisor stops computing the Aleph-specific static-IPv6 scheme (and stops needing to know a VM's product type). Prerequisite cleanup: it lets the confidential-instances work drop the `vm_type` field it was about to add to the supervisor contract.

Related tickets: ALEPH-XXX

## Why

The static guest IPv6 is a pure function of `(pool, vm_hash, vm_type)` and does not depend on the supervisor-assigned `vm_index`. Today the daemon computes it, and on restart re-adoption it recomputes the address by inferring the VM's type from its persisted config. That bakes product semantics (instance vs v-program) into the supervisor. The agent already has the pool, the subnet prefix, the policy, and the exact `StaticIPv6Allocator`, so it can compute the address upfront and hand the supervisor a value to use.

## Changes

- **Agent** (`translate.py`, `vprogram_launch.py`, `network/hostnetwork.py`): computes the static guest IPv6 and passes it in the existing `NetworkConfig.requested_ipv6` (as the full `/124` CIDR). Only under the static policy; the dynamic policy leaves it empty (that scheme needs a supervisor-side ordinal the agent cannot know). Shared helper so the three spec builders do not drift.
- **Daemon** (`world.rs`, `lifecycle.rs`, `controller_config.rs`): honors a non-empty `requested_ipv6` over its own computation; IPv4/TAP stays daemon-computed from `vm_index` (mechanical, unchanged). Persists the assigned guest IPv6 in the per-VM written config (`guest_ipv6_cidr`, Optional, absent-when-unset so existing config JSON is byte-identical). Restart adoption reads the persisted value; legacy configs with no persisted field fall back to today's recompute. Persistence gates on "the address is deterministically reproducible on adoption" (`requested_ipv6` honored OR static policy), which closes a silent-address-shift path if the two processes' policy ever diverged.
- **Proto**: comment only on `NetworkConfig.requested_ipv6` (the field already existed); no wire-contract change.

## Behavior preservation

For the static policy the agent-computed address is byte-identical to what the daemon computes today (same pool hextets, type nibble, and 44 hash bits), so existing VMs are unaffected. Pinned by a shared-literal parity test on both sides (`2a01:240:2:c8:3:abcd:ef01:2340/124`) plus a daemon round-trip test. Adoption backward-compat is covered by a legacy-config (no persisted field) test.

## How to test

`cargo test -p supervisor-daemon -p supervisor-controller` (green, incl. the parity, persistence, and legacy-adoption tests); `pytest tests/supervisor -q`.

## Notes

- Merges into `dev`. After it lands, the SNP-instance supervisor PR drops its `vm_type` addition and rebases on top.
- The dynamic-policy adoption path is unchanged (out of scope).

### DIFF
--- a/proto/supervisor.proto
+++ b/proto/supervisor.proto
@@ -251,8 +251,8 @@ message TeeConfig {
 
 message NetworkConfig {
   bool internet_access = 1;
-  string requested_ipv6 = 2;         // empty = pool-assigned
-  uint32 ipv6_prefix_len = 3;        // 0 = /128
+  string requested_ipv6 = 2;         // agent-computed static /124 CIDR (network/prefix); empty = daemon-assigned
+  uint32 ipv6_prefix_len = 3;        // the requested prefix (124 for the static scheme); 0 = /128
 }
 
 message GpuConfig {
--- a/rust/crates/supervisor-daemon/src/controller_config.rs
+++ b/rust/crates/supervisor-daemon/src/controller_config.rs
@@ -211,6 +211,14 @@ pub struct QemuVmConfig {
     pub numa_node: Option<u32>,
     #[serde(default)]
     pub hugepage_size: Option<String>,
+
+    // The guest IPv6 /124 CIDR the daemon assigned at create time, Rust-only.
+    // The agent computes the Aleph static address now, so adoption reads this
+    // back rather than re-deriving it from (pool, vm_hash, vm_type). Absent on
+    // configs written before the agent took over allocation (and under the
+    // dynamic policy); adoption then falls back to the recompute.
+    #[serde(default)]
+    pub guest_ipv6_cidr: Option<String>,
 }
 
 impl QemuVmConfig {
@@ -242,6 +250,7 @@ impl QemuVmConfig {
             kernel_cmdline: None,
             numa_node: None,
             hugepage_size: None,
+            guest_ipv6_cidr: None,
         }
     }
 
@@ -481,6 +490,12 @@ pub struct WrittenQemuVmConfiguration {
     pub numa_node: Option<u32>,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub hugepage_size: Option<String>,
+    // The guest IPv6 /124 CIDR the daemon assigned, Rust-only. Set under the
+    // static policy so a daemon restart adopts the address without re-deriving
+    // it; None (and so omitted, keeping the bytes identical to the pydantic
+    // writer and to legacy configs) under the dynamic policy or with no tap.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub guest_ipv6_cidr: Option<String>,
 }
 
 /// The top-level `Configuration` as written (QEMU instances only in this
@@ -865,6 +880,7 @@ mod tests {
                 kernel_cmdline: None,
                 numa_node: None,
                 hugepage_size: None,
+                guest_ipv6_cidr: None,
             },
             hypervisor: "qemu",
         };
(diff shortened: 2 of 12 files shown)


foxpatch-aleph (changes requested):
Well-structured refactor with correct persistence and adoption logic for the QEMU instance and V-PROGRAM paths. However, build_program_create_vm_spec uses VmType.from_message_content(message) which returns VmType.persistent_program (prefix "2") for persistent Firecracker programs, while the daemon's create_program_vm always used VmType::Microvm (prefix 0x1). Since the daemon now honors the agent's address verbatim, persistent Firecracker programs will silently get a different IPv6 address, contradicting the PR's byte-identical behavior preservation claim. The Instance and V-PROGRAM paths are correct.

**`src/aleph/vm/agent/translate.py`** (line 201): Behavior break: VmType.from_message_content(message) returns VmType.persistent_program (prefix "2") for persistent programs, but the daemon's create_program_vm (lifecycle.rs:3185) always uses VmType::Microvm (prefix 0x1) for the static IPv6 computation. Since the daemon now honors the agent-supplied address verbatim, persistent Firecracker programs will get a different IPv6 address (hextet 2 instead of 1), contradicting the PR's byte-identical claim. The Rust VmType enum has no PersistentProgram variant — all Firecracker programs are Microvm. Use VmType.microvm here instead of VmType.from_message_content(message) to match the daemon's behavior. (The Instance path in build_create_vm_spec and the V-PROGRAM path in build_vprogram_spec are correct — their agent-side VmType matches the daemon's.)

**`tests/supervisor/test_ipv6_allocator.py`** (line 44): The parity tests only cover VmType.instance. Add a test with VmType.microvm (and/or persistent_program) against the same hash to pin the cross-language parity for the Firecracker program path, mirroring the Rust VmType::Microvm (prefix 0x1) computation. Without this, the behavior break in translate.py:201 goes undetected.

**`tests/supervisor/test_supervisor_translate_program.py`** (line 50): This test doesn't assert spec.network.requested_ipv6 or spec.network.ipv6_prefix_len. Add an assertion (with the static policy monkeypatched, as in test_ipv6_allocator.py) to verify the program path computes the expected address — this would have caught the persistent_program vs microvm prefix mismatch.

foxpatch-aleph (approved):
Well-structured refactor that cleanly moves IPv6 allocation to the agent. The cross-language parity is pinned by shared-literal tests, the adoption backward-compatibility is properly handled with serde defaults, and the silent-shift guard (persisting agent-supplied addresses even under a divergent daemon policy) is a thoughtful edge case. The code is correct, secure (no injection surfaces), and well-documented. Minor observations below are non-blocking.

**`src/aleph/vm/agent/guest_ipv6.py`** (line 43): Unlike `StaticIPv6Allocator.__init__` (hostnetwork.py:59), this helper does not validate that `settings.IPV6_ADDRESS_POOL` is a /56 or /64. If a misconfigured pool is used, the agent would compute an address that the legacy recompute path would reject but the agent-supplied path (ipv6_from_cidr) would accept. Not a real-world concern since pools are configured at deploy time, but worth noting the inconsistency. Consider adding a pool-length check or documenting that validation is deferred to the daemon.

**`src/aleph/vm/agent/translate.py`** (line 201): For persistent programs, `VmType.from_message_content(message)` returns `VmType.persistent_program` (prefix "2"), but the Rust `create_program_vm` (lifecycle.rs:3185) hardcodes `VmType::Microvm` (prefix "1"). Before this PR the daemon computed the address with prefix "1"; now the agent-supplied value (prefix "2") is honored verbatim. This is technically a behavior change for persistent programs under the static policy, though it's arguably more correct (the type prefix was wrong before). Since programs don't persist configs to disk, there's no adoption concern. Worth confirming this is intended.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 3185): The `create_program_vm` path now honors the agent-supplied `requested_ipv6` but has no test covering this (unlike the QEMU create path which has three dedicated tests). A test verifying that a program create with a non-empty `requested_ipv6` produces the expected `entry.ipv6` would close the coverage gap, even though programs don't persist to disk.

odesenfans:
Addressed the persistent-program IPv6 finding in ac958ba9.

The bot re-reviewed and softened this to non-blocking ("arguably more correct"), but it is a real bug in the default configuration:

- `IPV6_ALLOCATION_POLICY` defaults to `static`, and the daemon now honors the agent-supplied address verbatim.
- `build_program_create_vm_spec` derived the vm-type from the message, so a persistent program got `persistent_program` (hextet `0x2`).
- The Rust `VmType` enum and the scheduler `VmType::ipv6_value()` have no `persistent_program` variant: both fold every Firecracker program into microvm (hextet `0x1`). So `0x2` is not "more correct", it disagrees with the address the rest of the network computes.

Fix pins the program path to `VmType.microvm`, so persistent and non-persistent programs share hextet `0x1` (byte-identical with the pre-refactor daemon). Added the three suggested coverage gaps: a microvm parity case in the agent helper test, a static-policy assertion over both persistence values in the translate-program test, and a Rust test that `create_program_vm` honors a supplied `requested_ipv6`.

The pool-length validation nit is left as-is (deploy-time config; validation stays deferred to the daemon).

foxpatch-aleph (approved):
Well-executed refactor that moves guest IPv6 allocation from the supervisor daemon to the agent. The cross-language parity is carefully pinned by shared-literal tests on both sides, the persistence gating correctly closes the silent-shift path (agent static + daemon dynamic), and legacy config adoption falls back to the recompute. VM-type hextet mapping is consistent across all three paths (instance=3, microvm=1, v_program=4), and the program path correctly pins microvm to avoid a persistent-program hextet drift. Test coverage is comprehensive across both Python and Rust, covering parity, persistence, round-tripping, legacy adoption, and the silent-shift guard. No correctness, security, or quality issues found.

**`tests/supervisor/test_supervisor_translate.py`** (line 107): Minor: the instance-path translate test asserts `internet_access` but not `requested_ipv6`, while the program path (`test_supervisor_translate_program.py`) and V-PROGRAM path (`test_vprogram_launch.py`) both verify the computed IPv6. Consider adding `assert spec.network.requested_ipv6` and `assert spec.network.ipv6_prefix_len == 124` under the static policy here too, for symmetric integration coverage of `build_create_vm_spec`.

PR #1140, opened 2026-08-20
ci(nix): golden-file check of the SEV-SNP launch measurements

odesenfans:
## What

Pins the flake's four SEV-SNP launch measurements in a committed golden file and adds a dedicated workflow that rebuilds the measured boot chain and compares:

- `nix/golden-measurements.json`: `measurement` (v-program), `composeMeasurement`, `workloadMeasurement` (fib-service demo), `instanceMeasurementSmoke` (confidential-instance chain with the placeholder owner).
- `nix/check-golden-measurements.sh`: builds the four outputs and diffs against the golden file; `--update` regenerates it for intended changes.
- `.github/workflows/golden-measurements.yml`: runs the check.

## Why

Any change to the measured chain (OVMF, kernel, initrd contents including the attest-agent binary, cmdline root hashes) currently changes the launch measurement silently; the first place it surfaces is a failed attestation. With the golden file, an intended change shows up as an explicit reviewable diff in the same PR, and an unintended one fails CI as a reproducibility regression.

## Trigger surface (deliberately not every build)

A cold chain build (edk2 + guest kernel + static-musl agent) takes hours on a runner, so the workflow only runs:

- on PRs and dev/main pushes touching the measured inputs: `nix/**`, `rust/crates/aleph-attest-agent/**`, `rust/Cargo.lock`, the workflow itself;
- weekly cron (inputs are fully pinned, so any diff on cron is reproducibility drift);
- manual dispatch.

No nix store caching on purpose: the repo's 10 GB Actions cache budget is already spent on the integration-test caches and a multi-GB store would evict them.

## Testing

- Golden file seeded from a local build of all four outputs; check mode passes against it.
- Negative path verified: a tampered golden value makes the script exit 1 with the diff.
- Workflow YAML parses; script is shellcheck-clean.


foxpatch-aleph (approved):
Well-structured CI check that pins SEV-SNP launch measurements in a golden file and verifies them on relevant changes. The script correctly builds all four flake measurement outputs, manually constructs JSON matching the golden file's exact format, and uses diff for comparison with clear error messaging. The workflow trigger surface is thoughtfully scoped to measured inputs (nix/, attest-agent crate, Cargo.lock) with a weekly cron backstop for reproducibility drift. No correctness, security, or testing issues found. The manual JSON construction is acceptable given the hex-string-only values, and comments thoroughly document the rationale and trade-offs.

**`nix/check-golden-measurements.sh`** (line 43): Minor style nit: `$((${#outputs[@]} - 1))` works but the inner `$` on `${#outputs[@]}` is required for the array length expansion. This is correct as-is — mentioning only because shellcheck SC2004 sometimes flags `$` in arithmetic contexts, though in this case `${#array[@]}` is the exception that requires it.

foxpatch-aleph (commented):
Well-structured PR with correct shell script logic, proper cleanup traps, and a well-reasoned workflow trigger surface. The golden file format matches the script output exactly, all flake inputs are fully pinned, and the four measurement values are valid 96-char hex strings. The one substantive gap: the workflow path filter omits rust/crates/aleph-tee/**, which is a direct path dependency of the measured aleph-attest-agent binary (rust/crates/aleph-attest-agent/Cargo.toml:10). Changes to aleph-tee would alter the measurements but would not trigger the CI check on the PR — only the weekly cron would catch it as drift, after the change has already merged. Adding rust/crates/aleph-tee/** to both the push and pull_request path filters would close this gap.

**`.github/workflows/golden-measurements.yml`** (line 27): The path filter includes `rust/crates/aleph-attest-agent/**` and `rust/Cargo.lock`, but `aleph-attest-agent` has a direct local path dependency on `aleph-tee` (`rust/crates/aleph-attest-agent/Cargo.toml:10`: `aleph-tee = { path = "../aleph-tee" }`). Since `aleph-tee` is compiled into the attest-agent binary that is baked into the measured initrd, changes to `rust/crates/aleph-tee/**` are measured inputs too. Consider adding `rust/crates/aleph-tee/**` to both the `push` and `pull_request` path filters so PRs changing `aleph-tee` trigger the golden-measurements check instead of surfacing as drift on the weekly cron.

**`.github/workflows/golden-measurements.yml`** (line 42): No `concurrency` group is set for this job. Since it can run up to 300 minutes with no caching, multiple PRs touching `nix/**` in parallel would spin up redundant multi-hour builds. Consider adding a `concurrency` group (e.g., `concurrency: { group: golden-measurements-${{ github.ref }}, cancel-in-progress: true }`) to cancel superseded runs.

foxpatch-aleph (approved):
Well-structured PR that replaces pkgs.makeInitrd with a content-only cpio archive in initrd.nix (so the SEV-SNP launch measurement tracks file content, not nix store paths), switches busybox to pkgsStatic, adds a golden-file CI check for the four launch measurements, and includes a standalone boot-smoke helper. The implementation is correct: the cpio recipe is reproducible (fixed mtime, sorted entries, gzip -9n), the static-linking guard and store-path contract check are good defenses, the golden-file script properly handles --update and missing-file paths, and the workflow trigger surface is deliberately scoped. No bugs or security issues found.

**`.github/workflows/golden-measurements.yml`** (line 41): Consider adding a `concurrency` group to cancel superseded runs. This job takes hours on a cold build; without it, rapid pushes to `dev` or `main` could start multiple overlapping 5-hour jobs. Example:
```yaml
concurrency:
  group: golden-measurements-${{ github.ref }}
  cancel-in-progress: true
```

**`nix/boot-smoke.sh`** (line 1): This script is added in the PR but not wired into any CI workflow and isn't mentioned in the PR description. Its purpose (local manual testing of the initrd boot path) is clear from the comments, but consider noting it in the PR description so reviewers know it's intentionally a standalone helper, not an oversight.

**`nix/initrd.nix`** (line 139): The `for f in root/bin/* root/init` loop assumes `root/bin/` is non-empty. All current flavors always have files in `bin/` (busybox is always present), so this is fine today. If a future flavor ever had no `bin/` files, the glob would fail to expand, `file` would be called on the literal `root/bin/*` string, and the build would fail. This is a hypothetical concern only — noting it for awareness.

foxpatch-aleph (approved):
Well-architected PR that correctly isolates the measured guest TCB's dependency set (separate Cargo workspace/lock), fixes a real reproducibility issue (content-only initrd replacing store-closure makeInitrd), and adds a golden-file CI check with an appropriately scoped trigger surface. The initrd.nix rewrite is the key change — it makes the launch measurement depend on file content rather than nix derivation hashes, with proper static-linking and store-path contract checks as guardrails. The workspace split is correct: aleph-tee's workspace = true references resolve against the root Cargo.toml which is included in the agent's source tree. Minor nits: the missing-golden-file check runs after the expensive build, and boot-smoke.sh is not wired into CI (manual only).

**`nix/check-golden-measurements.sh`** (line 58): The missing-golden-file check runs after the expensive build loop (line 42). Consider checking for the golden file's existence before building, so a missing file doesn't waste a full chain build when --update is not passed.

**`nix/boot-smoke.sh`** (line 1): This script is added but not wired into any CI workflow. If it's intended for manual use only, consider mentioning that in the PR description or adding a comment. If it should run in CI, a workflow step or separate job should be added.

**`nix/check-golden-measurements.sh`** (line 47): Manual JSON generation works because measurement values are always hex strings, but it's fragile to format drift. Using a tool like jq or python3 -m json.tool to produce canonical JSON would be more robust and would also handle edge cases like empty values or future structural changes.

foxpatch-aleph (approved):
This PR introduces golden-file checking for SEV-SNP launch measurements, a valuable CI safeguard against silent measurement drift. The implementation is correct and well-structured: the check script builds all four measurement outputs and diffs against a committed golden file with clear error messaging on divergence. The separation of aleph-attest-agent into its own cargo workspace (with its own Cargo.lock) is the right design — it isolates the measured guest binary's dependency set from supervisor-side cargo updates. The initrd.nix refactor from pkgs.makeInitrd (store closure) to a content-only cpio archive is a significant reproducibility improvement, with good safety checks (static linking verification, store path contract check). The workflow trigger paths are deliberately scoped to avoid hours-long cold builds on every PR, with a weekly cron as a backstop. The source-rev addition to image outputs is correctly kept out of the measured chain. The aleph-tee workspace dependency inheritance from the root workspace resolves correctly when building under the agent's workspace, as confirmed by the agent's lock file. No correctness, security, or testing issues found.

**`nix/check-golden-measurements.sh`** (line 44): Minor style nit: `$((${#outputs[@]} - 1))` could be simplified to `$((${#outputs[@]}-1))` (no spaces needed inside `$(())`), but this is purely cosmetic and the current form is correct.

**`nix/check-golden-measurements.sh`** (line 38): The JSON is manually constructed with printf. This works correctly for the fixed set of hex-string values, but consider using `jq -n` or `python3 -m json.tool` for robustness if the outputs set ever grows to include non-hex values. Not a blocker since the current keys are all 96-char hex strings.

**`nix/boot-smoke.sh`** (line 28): No check for `qemu-system-x86_64` availability before invocation. Acceptable since the script is documented as a local tool (not a CI gate), but a `command -v qemu-system-x86_64 >/dev/null || { echo 'qemu-system-x86_64 not found'; exit 1; }` guard would give a clearer error message.

PR #1141, opened 2026-08-20
docs(proto): fix VmSpec field order and a stale confidential note

odesenfans:
Small proto hygiene pass, no wire change.

- **VmSpec**: `hostname` (field 15) was declared before `guest_channel` (field 14), an insertion artifact. Reordered the two declarations to run in field-number order. Field numbers are unchanged, so the wire format is untouched; the regenerated bindings just follow the new order.
- **ConfidentialMode**: dropped a stale parenthetical ("not yet emitted by the in-process backend") — SEV-SNP is now a launch path the supervisor emits.

Left `HostInfo`'s number ordering alone: its fields jump around by number on purpose (grouped logically under CPU/Memory/Topology/TEE/... headers, numbered append-only to preserve wire compat). Reordering there would break the grouping for no benefit.

Regenerated the Python bindings (`supervisor_pb2.py`/`.pyi`); the Rust prost bindings regenerate at build and are unaffected (field numbers unchanged).


foxpatch-aleph (approved):
Clean proto hygiene pass with no wire-format changes. The VmSpec field reordering correctly places guest_channel (field 14) before hostname (field 15) in declaration order, while field numbers remain unchanged. The regenerated Python bindings are consistent with the proto source. The ConfidentialMode comment update removes stale text about SEV-SNP not being emitted. HostInfo was correctly left untouched due to its intentional logical grouping. No correctness, security, or testing concerns.

PR #1142, opened 2026-08-20
[DO NOT MERGE] chore: vprogram validation aggregation 3 (#1131 + attestation port map)

odesenfans:
Deb-builder aggregation branch for the aleph-testnets validation harness, superseding od/vprogram-integration-2.

= od/snp-inst-4-agent (#1131 tip, ba690c65) + the four validation-only commits that never landed on dev:

- 905bcf2d feat(vprogram): map the attestation port to a host IPv4 port (#1079)
- 3ac19303 fix(agent): heal port forwards when re-adopting a running VM (#1085)
- 591d50be fix(agent): update_refs must not derive refs from a V-PROGRAM
- 42fbe0bc fix(integration): restore the full port-forward reconcile in agent run.py

One conflict resolved in the 905bcf2d pick: build_vprogram_spec now both computes the agent-owned requested_ipv6 (dev) and returns the (spec, attestation_port) tuple (#1079).

Close when the four commits land on dev properly. CI here exists to produce the .deb artifact the testnets harness downloads.

Context: testnets run 32361553700 failed with 'V-PROGRAM is running but no attested endpoint was resolved' because od/snp-inst-4-agent lacks #1079.

### DIFF
--- a/packaging/Makefile
+++ b/packaging/Makefile
@@ -45,7 +45,7 @@ debian-package-code:
 	# MERGE GATE: aleph-message is git-pinned to the V-PROGRAM schema ref, swap for a release with pyproject.toml.
 	# protobuf is pinned to 5.29.6: 5.29.0 has a compilation issue; 6.x needs grpcio/tools 1.71+.
 	# legacy-cgi is packaging-only: it backfills the `cgi` module removed from the Python 3.13+ stdlib.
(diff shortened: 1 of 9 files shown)


odesenfans:
All content landed on dev: #1131 was merged earlier, the four validation-only commits went in via #1148, and the packaging pin via #1147. dev's tree now matches this branch, so the aggregation branch is redundant; the testnets manifesto is being repointed to dev.

PR #1143, opened 2026-08-20
chore(vprogram): adopt measurement registers from aleph-message

odesenfans:
Adopts the register-map schema from aleph-im/aleph-message#159.

`LaunchMeasurement.digest` becomes a `registers` object, so a
multi-register TEE such as Intel TDX (MRTD plus RTMRs) fits later without a
breaking protocol change. SEV-SNP declares `{"launch": "<hex>"}`.

## Scope

Supervisor code is unaffected. Measurements are never sent to the
supervisor, and `compute_snp_measurement` returns a bare hex digest rather
than a `LaunchMeasurement`; only its docstring named the message field.

So this is the dependency pin, the fixture, and that docstring.

The fixture's `item_hash` is recomputed because `item_content` changes. The
new value matches the aleph-rs and pyaleph fixtures byte for byte, which
cross-checks that all three repos serialize the same canonical content.

## Merge gate

The pin points at aleph-im/aleph-message#159, which must merge and be
reachable on `od/vprogram-schema` before this installs.

## Test plan

`tests/supervisor`: **10 failed / 1129 passed**, an identical failure set to
`dev` on this machine (pyroute2 needs root, journald, dbus stubs). Baselined
against a clean `dev` checkout and diffed rather than eyeballed, so the
result is no regressions rather than "the failures look environmental".

Design: `docs/superpowers/specs/2026-08-20-intel-tdx-design.md` §3, §4.1

https://claude.ai/code/session_01NaLi3sSGWh8EtWcVn64yni

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,7 +36,7 @@ dependencies = [
   "aiosqlite==0.19",
   "alembic==1.13.1",
   # MERGE GATE: replace the git pin with a released aleph-message version (V-PROGRAM schema, PR #158)
-  "aleph-message @ git+https://github.com/aleph-im/aleph-message@bc9c0f86958a746faeecbd6c6f96cc6f5b4538de",
+  "aleph-message @ git+https://github.com/aleph-im/aleph-message@4ad261ff25f595a64589d49bf36f514c402e7c5e",
   "aleph-superfluid~=0.2.1",
   "dbus-python==1.3.2",
   "eth-account~=0.10",
--- a/src/aleph/vm/vprogram/measurement.py
+++ b/src/aleph/vm/vprogram/measurement.py
@@ -3,7 +3,8 @@
 TRANSIENT: wraps the sev-snp-measure pip package (0.0.12, matching the nix
 pin in nix/flake.nix). The intended home for measurement computation is
 aleph-rs (the Rust CLI); this module is the stopgap so the publisher can pin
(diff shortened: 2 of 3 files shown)


foxpatch-aleph (approved):
This is a clean, minimal schema migration: `LaunchMeasurement.digest` becomes `LaunchMeasurement.registers['launch']`, future-proofing the protocol for multi-register TEEs like Intel TDX. The changes are limited to the dependency pin, a docstring update, and the test fixture. The fixture's `item_content` and `item_hash` are correctly recomputed and verified consistent (SHA-256 match), the launch digest length is correct (96 hex chars / SHA-384), and `compute_snp_measurement` is unchanged (still returns a bare hex string). The only nit is a stale comment in test_vprogram.py referencing the old item_hash, but this is non-blocking.

**`tests/supervisor/test_vprogram.py`** (line 3): Stale comment: references the old `item_hash 4c319b6b...` but the fixture now has `item_hash 1f9df108...` after the registers schema change. Consider updating to `item_hash 1f9df108...` to keep the docstring accurate.

**`pyproject.toml`** (line 38): The MERGE GATE comment references "PR #158" but the PR description and the registers schema change come from aleph-message#159. The comment is about the V-PROGRAM schema broadly, so it may be intentionally referencing #158, but consider clarifying which PR(s) must merge before this pin can be replaced with a release.

PR #1144, opened 2026-08-20
docs: design Intel TDX support

odesenfans:
Design document only. No code.

## Driver

Roadmap / portability proof. There is no Intel TDX hardware in the fleet.
The point is to prove the `aleph-tee` abstraction generalizes beyond AMD,
and to lock the schema seams that get expensive to change once SEV-SNP
ships. aleph-im/aleph-message#159 is the first increment landing from it.

## What the design concludes

The existing split absorbs TDX better than expected. The `TeeBackend`
trait, `AttestationReport` and `report_data.rs` are all unchanged; TDX
quotes are self-certifying, and `REPORTDATA` is 64 bytes like SNP's.

Two pleasant surprises: `vcpu_type` is not a TDX measurement input (the TD's
initial vCPU state comes from the TDX module, not the QEMU CPU model), so
the mixed-fleet problem SNP has does not exist. And DCAP verification needs
no Intel API key, since the PCK chain rides inside the quote and the
remaining collateral is free and Intel-signed.

Three places where TDX genuinely does not mirror SNP, which are the actual
design work:

1. **Measurement is multi-register** (MRTD plus RTMRs), which is what
   aleph-im/aleph-message#159 makes room for.
2. **The CCN cannot recompute RTMRs.** Reproducing them means emulating
   TDVF's measurement protocol. The fix is to move the per-deployment
   binding out of the measured cmdline into `MRCONFIGID`, which keeps the
   cmdline fixed per runtime and turns CCN validation into one SHA-384
   recompute plus a manifest comparison.
3. **TDX cannot express G1's launch-TCB gate.** Quotes report the *current*
   platform TCB, so a TD that ran for months under vulnerable firmware
   silently starts passing once the operator upgrades. Two mechanisms
   answer this and they are not alternatives: node-side tracking (§7.6,
   operator-visible, explicitly **not** a trust boundary) and an `rtmr3`
   launch-TCB commitment (§6.5, extended once by the measured initrd, which
   the client can recompute and verify).

## Reviewer's attention

- §6.3 accepts a real weakening: the CCN compares `rtmr1`/`rtmr2` against
  the runtime manifest rather than deriving them.
- §6.5 rests on a Linux guest RTMR-extend interface I could not verify
  without hardware. That is spike 5; enforcement is warn-on-mismatch until
  it lands, and §5.6 says plainly that the gap reopens if the interface is
  unusable.
- §9 increment 8 (node-side launch-TCB tracking) is the only item that pays
  off **before** any TDX hardware exists, since the same tracking serves
  SEV-SNP operators today and it discharges the spec G1 §8.1 deferred.

https://claude.ai/code/session_01NaLi3sSGWh8EtWcVn64yni

### DIFF
--- a/docs/superpowers/specs/2026-08-20-intel-tdx-design.md
+++ b/docs/superpowers/specs/2026-08-20-intel-tdx-design.md
@@ -48,6 +48,12 @@ free in-flight edit. Accepted deliberately: neither `aleph-sdk` nor
 `aleph-types` is published to crates.io, so consumers resolve through git tags
 and the blast radius is a coordinated internal bump.
 
+**Spike 1 resolved, spike 2 inherits the top slot.** QEMU's `tdx-guest` object
+does expose a settable `mrconfigid`, verified through the whole host-side chain
+(6.2), so the MRCONFIGID recipe in section 6 stands. The remaining
+highest-risk unknown is whether TDVF folds anything deployment-varying into
+`rtmr1` or `rtmr2`, which would push those out of the pinned set.
+
 **Two new items, one new lever.** The guest kernel moved to LTS 6.18 with an
 eval-time floor assert (#1122), which resolves spike 5 favourably. Against
 that, PR #1169 rebuilds the guest kernel as `allnoconfig` plus an explicit
@@ -473,17 +479,32 @@ A lying CRN has no move: either it sets `MRCONFIGID` honestly, in which case the
 client's pin against the message catches any wrong value, or it sets it
 dishonestly, in which case the guest refuses to boot.
 
-**Caveat found 2026-08-28: nobody populates this field in practice.**
-`MRCONFIGID` is all-zero in **both** genuine production quotes among the 8.1
-fixtures, including the Sapphire Rapids production quote; only the synthetic v5
-samples carry a non-zero value. The field is architecturally present and
-measured, so the reasoning above stands, but the whole path from
-`qemu -object tdx-guest,mrconfigid=...` through the TDX module and into the
-quote is unexercised by any fixture we hold. That promotes spike 1 from
-background confirmation to the load-bearing question for this section: it is
-the one unknown that could still invalidate the recipe. It is answerable from
-QEMU and kernel source with no TDX hardware, so it should be settled before
-increment 7 is written up.
+**Confirmed settable, 2026-08-28.** The host-side path is real, and every link
+was checked rather than assumed:
+
+| Link | Evidence |
+|---|---|
+| `-object tdx-guest,mrconfigid=<base64>` | `mrconfigid` is a settable `string` property on the `tdx-guest` object, confirmed live via QMP `qom-list-properties` against QEMU 10.2.1. Siblings `mrowner`, `mrownerconfig`, `attributes`, `sept-ve-disable` and `quote-generation-socket` are all present, so decision 2's QGS path is settable the same way. |
+| base64, exactly 48 bytes | QEMU v10.2.0 `target/i386/kvm/tdx.c` runs `qbase64_decode`, then rejects anything whose length is not `QCRYPTO_HASH_DIGEST_LEN_SHA384`. |
+| into KVM | `memcpy(init_vm->mrconfigid, data, data_len)` then the `KVM_TDX_INIT_VM` ioctl. |
+| into the TDX module | Linux 6.18 uapi `struct kvm_tdx_init_vm { __u64 mrconfigid[6]; /* sha384 digest */ }`, landing in `struct td_params { u64 mrconfigid[6]; }`, the TD_PARAMS input to `TDH.MNG.INIT`. |
+
+Note `-object tdx-guest,help` reports "there are no options for tdx-guest";
+the properties are registered via `object_property_add_str`, so QMP
+`qom-list-properties` is the way to enumerate them. Worth knowing before
+concluding the object is unconfigurable.
+
+A useful side effect for 6.2's argument: the length check is fail-closed at
+QEMU start. A CRN that sets a malformed `mrconfigid` does not launch a
+mis-measured TD, it fails to launch at all. That is a third leg alongside the
+two in the paragraph above.
+
+**What is still unexercised.** `MRCONFIGID` is all-zero in **all four** 8.1
+fixtures. Nothing we hold demonstrates a populated value surviving into a
+quote, so the QEMU-to-quote path is confirmed by source and architecture but
+not by evidence. Closing that gap needs hardware and belongs to Tier 2. The
+recipe is safe to build against; it is the one thing to verify first when a TDX
+host does appear.
 
 **Descriptor delivery.** The descriptor must reach the guest over a channel that
 does not perturb `rtmr1` or `rtmr2`, or the recipe defeats its own purpose by
@@ -776,9 +797,19 @@ looks; it catches our parsing and chain-walk mistakes, not shared
 misinterpretations of the spec. Cross-checking against `go-tdx-guest`'s
 independent implementation is what covers that gap.
 
-Two findings from these fixtures are recorded where they bite: MRCONFIGID is
-all-zero in both genuine production quotes (6.2), and `rtmr3` is all-zero in
-every one of them (6.5).
+Two findings from these fixtures are recorded where they bite: `MRCONFIGID` is
+all-zero in all four (6.2), and so is `rtmr3` (6.5).
+
+**Parsing trap, worth stating because it was hit while checking the above.**
+The TD report body does not sit at a fixed offset. In quote v4 it begins
+directly after the 48-byte header. In v5 a *body descriptor* intervenes: a
+`u16` body type and a `u32` body size, so the body begins at offset 54, and the
+type distinguishes `TdReport10` (584 bytes) from `TdReport15` (648). Parsing a
+v5 quote at the v4 offset does not fail loudly, because everything is
+fixed-size opaque bytes; it silently shifts every register by six and yields
+plausible-looking digests. Both v5 fixtures here are `TdReport15`. Increment 2
+must dispatch on quote version before touching the body, and the parser tests
+should include a v5 quote specifically to catch this.
 
 ### 8.2 Test matrix
 
@@ -879,19 +910,18 @@ slips indefinitely, increment 8 should still ship.
 
 ## 11. Open questions (spikes, not design forks)
 
-**Status 2026-08-28:** spike 5 is resolved; spike 1 is promoted to the top of
-the list and is the next thing to investigate.
+**Status 2026-08-28:** spikes 1 and 5 are resolved, both favourably. Spike 2 is
+now the highest-risk item and the next thing to investigate.
 
-1. **[HIGHEST PRIORITY] Does QEMU's `tdx-guest` object expose `mrconfigid` as a
-   settable base64 property, and does the value survive into the quote?**
-   Believed yes, alongside `mrowner` and `mrownerconfig`, but it needs
-   confirming against the QEMU version we would target. Decision 4 depends on
-   it; if it is not settable, the fallback is per-deployment `rtmr2` plus TDVF
-   measurement emulation in the CCN, which is materially worse. Promoted
-   because the 8.1 fixtures show `MRCONFIGID` all-zero in genuine production
-   quotes (6.2), so nothing we hold exercises this path. Answerable from QEMU
-   and kernel source without hardware.
-2. **Is anything deployment-varying folded into `rtmr1` or `rtmr2`?** Under the
+1. **~~Does QEMU's `tdx-guest` object expose `mrconfigid` as a settable base64
+   property?~~ RESOLVED 2026-08-28, favourably.** It does: a settable `string`
+   property taking base64 that must decode to exactly 48 bytes, plumbed through
+   `KVM_TDX_INIT_VM` into the TDX module's TD_PARAMS. Full evidence chain in
+   6.2. Decision 4 stands and the `rtmr2` fallback is not needed. The residual
+   is only that no fixture demonstrates a populated value reaching a quote, so
+   confirming that end to end is a Tier 2 item for first hardware.
+2. **[HIGHEST PRIORITY] Is anything deployment-varying folded into `rtmr1` or
+   `rtmr2`?** Under the
    6.2 recipe the cmdline is fixed per runtime, so which of the two registers
    receives it no longer matters. What does matter is whether TDVF measures
    anything that differs between deployments of the same bundle: the descriptor
--- a/docs/superpowers/specs/2026-08-20-intel-tdx-design.md
+++ b/docs/superpowers/specs/2026-08-20-intel-tdx-design.md
@@ -1,0 +1,597 @@
+# Intel TDX support: design
+
+**Date:** 2026-08-20
+**Status:** Approved design, ready for implementation planning.
+**Branch:** `od/tdx-design` off `dev`.
+
+**Driver:** roadmap / portability proof. There is no Intel TDX hardware in the
+fleet today. The purpose of this work is to prove that the `aleph-tee`
+abstraction generalizes beyond AMD, and to lock the protocol and schema seams
+that would be expensive to change once SEV-SNP ships. Hardware bring-up is
+specified here but deliberately not built.
+
+**Related:**
+
+- [`2026-07-08-confidential-vm-protocol-design.md`](../../plans/2026-07-08-confidential-vm-protocol-design.md),
+  section 7 ("An eventual TDX backend adds a platform tag with MRTD digests")
+  and section 10 (IGVM direction).
+- [`2026-08-18-snp-confidential-instances-design.md`](../../plans/2026-08-18-snp-confidential-instances-design.md).
+- [`2026-08-12-snp-tcb-floor-design.md`](../../plans/2026-08-12-snp-tcb-floor-design.md)
+  (G1), whose client-side floor architecture this design generalizes.
+
+---
+
+## 1. Goal and scope
+
+Deliver two things:
+
+1. **The schema and trait seams**, landed now, so that adding TDX later is
+   additive rather than a breaking protocol change. The window is open: the
+   `aleph-message` TEE schema lives on the unmerged `od/vprogram-schema` branch
+   and the latest release is 1.2.0, so `LaunchMeasurement` can still change
+   shape for free.
+2. **A hardware-free TDX quote verifier** in `aleph-tee`. DCAP verification is
+   pure computation: given one captured quote plus its Intel-signed collateral,
+   the entire verifier can be implemented and fully tested with no TDX machine
+   anywhere. That is the bulk of the hard work, and it is the portability proof
+   made concrete rather than asserted.
+
+Out of scope for the buildable increment, specified in section 7: the in-guest
+backend, QEMU argument generation, the TDVF nix derivation, QGS/PCCS host
+operations, and CRN capability advertisement.
+
+---
+
+## 2. Background: what already fits, and what does not
+
+The existing split isolates TDX's differences remarkably well. Three files'
+worth of new code plus two schema changes cover it.
+
+| Layer | SEV-SNP today | TDX | Verdict |
+|---|---|---|---|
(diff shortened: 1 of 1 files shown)


foxpatch-aleph (commented):
This is a well-researched design document with accurate claims about the existing codebase (verified: TeeType::Tdx, TEE_BACKEND_TDX=3, tdx_supported=6, report_data.rs domain separation, VerificationResult.measurement shape, CONFIDENTIAL_MODE values). The security reasoning around the rtmr3 launch-TCB commitment, the fail-closed pinning discipline, and the mrconfigid injection surface all hold up. Two issues worth addressing: the TdReport10/TdReport15 byte sizes (584/648) are field-sums, not actual struct sizes, which would mislead a parser implementer; and three cross-referenced plan documents under ../../plans/ do not exist in this repository, leaving the links dead.

**`docs/superpowers/specs/2026-08-20-intel-tdx-design.md`** (line 245): The stated TdReport10 size of 584 bytes and TdReport15 of 648 bytes are sums of the listed field sizes, not the actual struct sizes a parser must handle. The real TD_REPORT structure in the Intel TDX spec is 1024 bytes (REPORTMACSTRUCT at 256B + TD_REPORT_BODY at 512B + reserved). An implementer reading "584 bytes" as a struct size would produce incorrect field offsets. Consider clarifying that 584 is the sum of the named fields, or cite the actual TD_REPORT struct size (1024B for TDX 1.0, with the 1.5 variant adding the 64B of tee_tcb_svn2 + mrservicetd within the existing 1024B frame via reserved space, not growing it to 648B).

**`docs/superpowers/specs/2026-08-20-intel-tdx-design.md`** (line 15): The three cross-referenced plan documents (2026-07-08-confidential-vm-protocol-design.md, 2026-08-18-snp-confidential-instances-design.md, 2026-08-12-snp-tcb-floor-design.md) under ../../plans/ do not exist in this repository. The links resolve to docs/plans/ which contains only a different document. If these live in another repo or on another branch, the links should note that; if they are expected to land in this tree, they should land before or with this design.

**`docs/superpowers/specs/2026-08-20-intel-tdx-design.md`** (line 686): The testing section says "table-driven over all seven statuses" for the TCB policy, but section 5.5 lists six statuses by name (UpToDate, SWHardeningNeeded, ConfigurationNeeded, Revoked, plus the implied others). Intel DCAP defines seven TCB statuses (UpToDate, SWHardeningNeeded, ConfigurationNeeded, ConfigurationAndSWHardeningNeeded, OutOfDate, Revoked, NoActionNeeded). Consider enumerating all seven in 5.5 so the test plan is self-checking against the spec.

foxpatch-aleph (commented):
A thorough, security-conscious design document that correctly identifies and handles the TDX-specific challenges. Technical claims verified against the codebase check out: the TeeBackend trait, TeeType::Tdx variant, proto placeholders, VerificationResult.measurement field, and the SNP verifier's clock problem are all as described. The rtmr3 launch-TCB commitment scheme and the MRCONFIGID boot recipe are well-reasoned. The main gaps are three dead cross-references to plans that don't exist in the repo, an unspecified DOMAIN_LAUNCH_TCB constant (critical given the module's anti-drift purpose), and an underspecified RTMR extend operation. None of these are blocking for a design document, but the dead links and the missing constant specification should be addressed before implementation begins.

**`docs/superpowers/specs/2026-08-20-intel-tdx-design.md`** (line 16): The three referenced plans (2026-07-08-confidential-vm-protocol-design.md, 2026-08-18-snp-confidential-instances-design.md, 2026-08-12-snp-tcb-floor-design.md) do not exist anywhere in the repo. These are cited as authoritative sources for protocol decisions throughout the document. Either qualify the paths (different repo?) or mark them as pending.

**`docs/superpowers/specs/2026-08-20-intel-tdx-design.md`** (line 512): The RTMR extend operation should be written explicitly as rtmr3 = SHA-384([0u8;48] || SHA-384(DOMAIN_LAUNCH_TCB || cpusvn || pcesvn || tee_tcb_svn)) to prevent an implementer from using a different construction. The TDX extend semantics (SHA-384 concat with old value) are load-bearing for client-side recomputation to work.

**`docs/superpowers/specs/2026-08-20-intel-tdx-design.md`** (line 534): DOMAIN_LAUNCH_TCB is said to belong in report_data.rs beside DOMAIN_KEY and DOMAIN_FRESH but its exact bytes are not specified. Since that module's entire purpose is preventing construct-side/verify-side drift, the constant should be pinned in the design (e.g., b"aleph-attest-launch-tcb-v1\x00") to match the existing pattern.

**`docs/superpowers/specs/2026-08-20-intel-tdx-design.md`** (line 340): The existing AttestationReport doc-comment in types.rs says "raw AMD-signed report blob (1184 bytes for SEV-SNP)" and "MUST run the full AMD certificate chain". This will need generalization when TDX lands. Worth adding to increment 2's scope as a doc-comment update.

**`docs/superpowers/specs/2026-08-20-intel-tdx-design.md`** (line 300): The unnumbered paragraph between steps 6 and 7 ("Check whether the shipped SEV-SNP verifier has the same latent problem...") reads as a call-to-action aside. Consider formatting it as a labeled Note block to distinguish it from the numbered verification steps, since it is confirmed true (check_cert_validity uses Asn1Time::days_from_now(0), which reads the system clock).

foxpatch-aleph (commented):
A thorough and well-reasoned design document for Intel TDX support. Security reasoning is sound, with the launch-TCB gap and its rtmr3 mitigation being the standout piece. All verifiable claims against the codebase are accurate (TeeType::Tdx, VerificationResult.measurement, proto placeholders, verify.rs line count, golden-measurements.json shape, report_data.rs domain separation). The main issue is that all three cross-referenced design docs in the Related section point to paths that do not exist in this repository. The document is ready to merge with those links fixed or annotated as cross-repo references.

**`docs/superpowers/specs/2026-08-20-intel-tdx-design.md`** (line 16): All three `Related` links point to `../../plans/2026-07-08-confidential-vm-protocol-design.md`, `../../plans/2026-08-18-snp-confidential-instances-design.md`, and `../../plans/2026-08-12-snp-tcb-floor-design.md` — none of these files exist in this repo. If they live in another repository (aleph-message?), the relative paths are dead. Consider noting which repo they belong to, or linking to the PR/branch where they can be found.

**`docs/superpowers/specs/2026-08-20-intel-tdx-design.md`** (line 319): Step 5 uses `SHA-256` for the QE report_data binding, while every other hash in this doc is SHA-384. This is correct per the TDX DCAP spec, but a one-line note explaining *why* it is SHA-256 here (Intel's QE report_data is 32 bytes, not 48) would prevent a future implementer from \"fixing\" it to SHA-384.

**`docs/superpowers/specs/2026-08-20-intel-tdx-design.md`** (line 364): The doc says `sev_snp/verify.rs` is 1015 lines; it is actually 1016. Trivial, but since the point is to convey cost, rounding to \"roughly 1000 lines\" or updating to 1016 would be precise.

**`docs/superpowers/specs/2026-08-20-intel-tdx-design.md`** (line 617): §6.5 says `rtmr0` and `rtmr1` are firmware-owned, but earlier in the doc (§6.1, §6.3) `rtmr1` is described as measuring \"the kernel\" and is in the pinned set that the CCN compares against the manifest. Both statements can be simultaneously true (TDVF extends rtmr1 with kernel measurement), but the term \"firmware-owned\" could be misread as \"TDVF-only\". A brief clarification that \"firmware-owned\" means \"TDVF controls the extend, including measuring guest inputs into it\" would help.

foxpatch-aleph (commented):
A thorough, well-researched design document whose codebase claims are accurate (verified against TeeBackend trait, TeeType::Tdx, VerificationResult, report_data.rs, proto placeholders, and the divergences ledger). The security reasoning is sound and honestly documents its own weaknesses. Three documentation issues: the three referenced design docs under ../../plans/ don't exist in this repo (broken links), §6.1 states a premise about TDVF measuring the cmdline into rtmr2 that §6.6 proves false without correcting §6.1 itself, and the 'seven statuses' in the test matrix are never enumerated. None are blocking, but the §6.1/§6.6 contradiction should be reconciled before implementation begins.

**`docs/superpowers/specs/2026-08-20-intel-tdx-design.md`** (line 16): The three linked documents under `../../plans/` do not exist in this repository. `docs/plans/` contains only `2026-08-19-compose-runtime-port-design.md`. The relative path `../../plans/` from `docs/superpowers/specs/` resolves to `docs/plans/`, but none of `2026-07-08-confidential-vm-protocol-design.md`, `2026-08-18-snp-confidential-instances-design.md`, or `2026-08-12-snp-tcb-floor-design.md` are present. These may live in another repo; if so, note the repo name, or move them here.

**`docs/superpowers/specs/2026-08-20-intel-tdx-design.md`** (line 471): §6.1 states as fact: "TDVF measures the kernel cmdline there" (in rtmr2). §6.6 (line 690-698) proves this is false: TDVF uses `BlobVerifierLibNull`, so the cmdline is neither measured nor verified on TDX. §6.2 (line 512) acknowledges this with an addendum ("The cmdline premise below does not survive the port"), but §6.1 itself was never corrected. A reader hitting §6.1 before §6.6 gets a factually wrong explanation for why `rtmr2` is per-deployment. Consider rewriting §6.1 to state the original assumption, then note it was invalidated by spike 2 (§6.6), so the narrative flows correctly.

**`docs/superpowers/specs/2026-08-20-intel-tdx-design.md`** (line 921): "table-driven over all seven statuses crossed with the override matrix" — the document names only five statuses (`UpToDate`, `SWHardeningNeeded`, `ConfigurationNeeded`, `Revoked`, `OutOfDate`). The remaining two (likely `InvalidAdvisory` and one other depending on spec version) should be enumerated here or in §5.5 so the implementer does not have to guess the full set.

foxpatch-aleph (approved):
This is a well-researched, internally coherent design document that correctly identifies where TDX diverges from SEV-SNP and proposes sound solutions for each divergence. I verified the document's claims against the actual codebase (TeeType::Tdx exists, TeeBackend trait shape, VerificationResult.measurement scalar, proto enum values, golden-measurements.json structure, SNP verifier clock reading) and all checked claims are accurate. The rtmr3 launch-TCB commitment scheme is cryptographically sound, the MRCONFIGID recipe's trust argument is correct, the parsing trap callout for v5 quotes is genuinely important, and the accepted weakenings (rtmr1/rtmr2 manifest comparison, warn-on-mismatch for rtmr3) are honestly documented with clear mitigation paths. The only nit is a stale preamble in section 11 that says spike 2 is the highest-risk item, when spike 2 itself is marked RESOLVED in the same section.

**`docs/superpowers/specs/2026-08-20-intel-tdx-design.md`** (line 1009): The preamble says "spikes 1 and 5 are resolved, both favourably. Spike 2 is now the highest-risk item and the next thing to investigate" — but spike 2 below (line 1019) is also marked RESOLVED 2026-08-28. The preamble text appears stale relative to the spike 2 resolution. Consider updating to reflect that spikes 1, 2, and 5 are all resolved, and the remaining open spikes are 3 and 4.

foxpatch-aleph (commented):
A thorough, well-researched design document with accurate codebase references and sound security reasoning. The fail-closed discipline, the mrconfigid injection-rejection stance, and the rtmr3 launch-TCB commitment scheme are all well-considered. One genuine internal contradiction exists around spike 2's resolution status (resolved in the revision history and spike entry, but flagged as unresolved and highest-risk in the section 11 header). Section 2.1's claim that the mixed-fleet problem 'does not exist on TDX' is more optimistic than section 6.6's revelation that vCPU count IS a measurement input. The verify.rs line count is off by one (1016, not 1015). None of these are blocking for a design document, but the spike 2 contradiction should be reconciled before implementation planning.

**`docs/superpowers/specs/2026-08-20-intel-tdx-design.md`** (line 1056): Internal contradiction: the section 11 status header says "spikes 1 and 5 are resolved" and "Spike 2 is now the highest-risk item and the next thing to investigate," but the revision history (line 56) says "Spike 2 resolved, and the design survives it," and spike 2's own entry (line 1066) is struck through and marked "RESOLVED 2026-08-28." Reconcile: if spike 2 is resolved, update this header to list all three resolved spikes and identify the actual next-highest-risk item (likely spike 3 or 4).

**`docs/superpowers/specs/2026-08-20-intel-tdx-design.md`** (line 64): After stating spike 2 is resolved at line 56, this paragraph says "The remaining highest-risk unknown is whether TDVF folds anything deployment-varying into rtmr1 or rtmr2" — which is exactly spike 2's question. If spike 2 is indeed resolved (and 6.6 answers it), this sentence should be updated to reflect that the question was answered, not that it remains unknown.

**`docs/superpowers/specs/2026-08-20-intel-tdx-design.md`** (line 129): Section 2.1 says "The mixed-fleet problem SNP has (distinct digests for EPYC-v3 and EPYC-v4) does not exist on TDX." This is true for the CPU model axis, but section 6.6 (line 752) reveals vCPU count is a TDX measurement input via ACPI table size, and calls it "the TDX analogue of the SEV-SNP vcpu_type problem." Consider softening 2.1 to note that while the CPU-model mixed-fleet problem is eliminated, a vCPU-count/machine-shape analogue exists and is addressed in 6.6.

**`docs/superpowers/specs/2026-08-20-intel-tdx-design.md`** (line 374): The file sev_snp/verify.rs is 1016 lines, not 1015. Off-by-one in an illustrative cost comparison.

foxpatch-aleph (commented):
A thorough, security-conscious design document that accurately reflects the existing codebase. All major codebase claims verified correct (TeeType::Tdx, HostInfo.tdx_supported=6, TEE_BACKEND_TDX=3, TeeBackend trait surface, AttestationReport structure, report_data.rs domain separation, VerificationResult.measurement scalar, MAX_ACCEPTED_VMPL, pinned_amd_ark_der, async SNP verifier, ConfidentialMode enum values, golden-measurements infrastructure). The design correctly identifies where TDX genuinely diverges from SNP (multi-register measurement, RTMR recomputation impracticality, launch-TCB gap) and proposes well-motivated solutions (MRCONFIGID deployment binding, rtmr3 launch-TCB commitment, node-side tracking). No correctness or security issues found. The comments below are documentation precision issues: broken relative links to non-existent referenced docs, 'unchanged' verdicts that are structurally true but overlook AMD-specific doc comments that will need updating, and a minor line-count discrepancy.

**`docs/superpowers/specs/2026-08-20-intel-tdx-design.md`** (line 15): The three referenced design docs (2026-07-08-confidential-vm-protocol-design.md, 2026-08-18-snp-confidential-instances-design.md, 2026-08-12-snp-tcb-floor-design.md) do not exist in this repository at the linked paths (docs/plans/). The links will be dead for any reader. Consider noting these are internal/cross-repo references, or linking to where they actually live.

**`docs/superpowers/specs/2026-08-20-intel-tdx-design.md`** (line 58): The table verdict 'unchanged' for AttestationReport is structurally accurate (fields don't change), but the doc comments on the struct and its `data` field are AMD-specific ('raw AMD-signed report blob', 'AMD certificate chain'). When TDX lands these must become platform-agnostic. Consider noting this so implementers don't overlook the comment updates.

**`docs/superpowers/specs/2026-08-20-intel-tdx-design.md`** (line 88): TeeConfig is called 'already backend-agnostic', but its `cpu_model` field (types.rs:114-118) has SNP-specific documentation referencing `sev_snp::qemu::DEFAULT_CPU_MODEL` and VMSA semantics. For TDX, `-cpu host` is used and vcpu_type is not a measurement input, so this field is unused. The design should explicitly state TDX ignores `cpu_model`, or the field needs generalizing.

**`docs/superpowers/specs/2026-08-20-intel-tdx-design.md`** (line 448): Minor: the doc says sev_snp/verify.rs is '1015 lines' to estimate TDX verifier cost; actual count is 1016. Trivial, but since the number supports the cost argument, worth being precise.

foxpatch-aleph (commented):
A thorough, well-structured design document that accurately characterizes the existing codebase. The TDX abstraction generalizes the SNP architecture cleanly, and the places where TDX genuinely differs (multi-register measurement, RTMR recomputation, launch-TCB gap) are identified with concrete, well-reasoned solutions. The accepted weakenings are honestly disclosed with clear paths to closure. Three issues: broken relative links to three referenced design documents that don't exist in the repo, AMD-specific doc comments on AttestationReport that the design says are unchanged but will be misleading for TDX, and a minor line count discrepancy (1016 vs 1015).

**`docs/superpowers/specs/2026-08-20-intel-tdx-design.md`** (line 15): The three referenced design documents (2026-07-08-confidential-vm-protocol-design.md, 2026-08-18-snp-confidential-instances-design.md, 2026-08-12-snp-tcb-floor-design.md) do not exist in this repo. The docs/plans/ directory contains only one unrelated file. If these live in a different repo or are not yet committed, the links should note that explicitly or use absolute references, otherwise they are dead links.

**`docs/superpowers/specs/2026-08-20-intel-tdx-design.md`** (line 388): The design says AttestationReport is "unchanged" for TDX, but the existing doc comment at types.rs:17 says "the raw AMD-signed report blob (1184 bytes for SEV-SNP)" and line 35 says "The raw AMD-signed report blob." For TDX the blob is Intel-signed and a different size. The §4.4 plan carefully carries over the VerificationResult doc comment "verbatim and applies per register" — AttestationReport's doc comment needs the same treatment: either generalize "AMD-signed" to "TEE-signed" or add a TDX note. Otherwise the comments become misleading once a TDX backend exists.

**`docs/superpowers/specs/2026-08-20-intel-tdx-design.md`** (line 271): Minor: sev_snp/verify.rs is 1016 lines, not 1015. The line count is used to estimate the TDX verifier's cost ("a cost already paid once for AMD"), so accuracy matters for the cost argument even though the difference is trivial.

foxpatch-aleph (changes requested):
An exceptionally rigorous design doc whose repo-facing claims verify almost perfectly against the codebase (proto placeholders, TeeBackend trait shape, golden-measurements cron, argv parity, domain-separated report_data all check out), and the TDX technical substance is sound and refreshingly honest about weakenings. The blocker is that section 4.4 is internally contradictory and wrong about the shipped code: the second paragraph still claims aleph_tee::types::VerificationResult carries a scalar measurement, but main already ships registers: SevSnpRegisters, which makes the increment-5 plan (the doc's own only time-sensitive item) stale. Fixing that paragraph plus the broken Related links would make this approvable.

**`docs/superpowers/specs/2026-08-20-intel-tdx-design.md`** (line 238): This paragraph contradicts the one above it and the shipped code. `aleph_tee::types::VerificationResult` on main already has `registers: SevSnpRegisters` (rust/crates/aleph-tee/src/types.rs:63); the scalar survives only as the deprecated wire-compat `measurement: Option<String>` in `VerificationResultWire`. So the scalar now exists in one shipped place, not two: the public `aleph_sdk::attest::VerificationResult`. Since the first paragraph already describes the shipped shape, this paragraph ("Both move together", "coordinated bumps in aleph-vm") should be rewritten to say the remaining move is aleph-rs-only.

**`docs/superpowers/specs/2026-08-20-intel-tdx-design.md`** (line 883): Same staleness: "`registers` on both `VerificationResult`s" and listing aleph-vm/aleph-vm-scheduler as targets implies aleph-vm's `VerificationResult` still needs the change, but it already has `registers: SevSnpRegisters` on main. Increment 5 is now the aleph-rs public type only (aleph-rs 0.18.0).

**`docs/superpowers/specs/2026-08-20-intel-tdx-design.md`** (line 15): All three Related links are broken: they resolve to `docs/superpowers/plans/...`, which does not exist (the tree has `docs/plans/` with only the compose-runtime-port design, and `docs/superpowers/` contains only `specs/`). The doc leans on these constantly ("protocol decision 3", "decision 11", "G1 decision 1/2/4", "section 10 of the protocol design"), so the references need repo-qualified links or to be corrected to wherever the protocol, SNP-instances and G1 designs actually live.

**`docs/superpowers/specs/2026-08-20-intel-tdx-design.md`** (line 32): "register-map form" contradicts section 4.4, which explicitly says a concrete struct, "not a `BTreeMap`". Presumably a leftover from the earlier revision (the revision history mentions the reshaping).

**`docs/superpowers/specs/2026-08-20-intel-tdx-design.md`** (line 890): Minor sequencing nit: the table is introduced as "In execution order", yet increment 10 is listed after 7 while its own note says it is "needed before 7 can produce a real golden entry". Either swap the rows or note that 7 lands spec-only first, with the golden entry arriving after 10.

PR #1145, opened 2026-08-20
feat(snp): launch with the measured CPU model instead of a hardcoded EPYC-v4

odesenfans:
Based on `dev`, rebased onto current `dev` (after #1131, #1146, #1079, #1085 and #1147). The V-PROGRAM spec builder now returns `(spec, attest_port)`, so the CPU-model selection sits just before that tuple is built.

## Why

An SEV-SNP guest's launch measurement is a function of the QEMU CPU model: the model determines the per-vCPU VMSA pages that enter the launch digest. The CRN hardcoded `-cpu EPYC-v4` in both argv builders (`aleph-tee/src/sev_snp/qemu.rs`, the controller's `snp_tee_fragment`) and the agent-to-supervisor proto had no field to carry a model.

So a V-PROGRAM or a `mode=sev_snp` instance whose measurements were computed for any other model launched as `EPYC-v4` anyway, produced an attestation report matching none of the message's digests, and failed at the client verifier with nothing pointing at the cause. A host that cannot launch `EPYC-v4` failed at `qemu-system-x86_64` exec time instead of at spec build.

This also completes the other half of aleph-vm-scheduler #193: that PR gates placement on `properties.tee.sev_snp.supported_vcpu_types`, which only means something if the CRN then launches the model the placement was made on.

## What

The agent picks the model, everything downstream carries it verbatim:

```
message verification.measurements[].vcpu_type      (instances: trusted_execution.measurements[])
   intersect  get_supported_snp_vcpu_types()       the QEMU query-cpu-definitions probe
   -> first match in message order
   -> CreateVmSpec.tee.cpu_model -> proto TeeConfig.cpu_model (tag 6)
   -> daemon SnpConfig -> controller config JSON "cpu_model"
   -> snp_tee_fragment / sev_snp_qemu_args -> qemu -cpu <model>
```

- **`src/aleph/vm/agent/vcpu_select.py`** (new): `requested_vcpu_types` reads the measurements in message order, deduplicated case insensitively, with an untagged measurement falling back to `EPYC-v4`. `select_snp_vcpu_type` returns the first requested model the host can launch, spelled the way QEMU reported it (`-cpu` is case sensitive), and raises `VmSetupError` naming both sets when the intersection is empty. The value that reaches argv is therefore always host-provided, never a string from the message.
- **Both spec builders** call it before returning, so no launchable model means no VM is created, rather than a VM that boots and cannot attest.
- **Absent means `EPYC-v4` at every hop.** Controller configs written before this change are re-read on restart and on adoption; they carry no `cpu_model` and produce byte-identical argv. `cpu_model` is deliberately kept out of `QemuVmConfig::snp()`'s all-fields check so an absent model can never demote an SNP config to non-SNP.
- **The byte-parity conformance test** now crosses a model axis (absent / `EPYC-v4` / `EPYC-Genoa-v2`) with the existing memory, policy, OVMF, NUMA, hugepage and CPUID axes, and asserts the emitted model against an independent literal, so the two builders' defaults drifting together still fails.
- **`docs/architecture/divergences.md` entry 68** is amended: it recorded the fixed `EPYC-v4` as deliberate. Measurements stay host-independent (the digest is a function of the declared model, not of the host); what changed is that a publisher can offer several models and the CRN is held to the one it was selected for.

## Notes for review

- Launch behaviour is unchanged for every message published today: `DEFAULT_CPU_MODELS` in `bundle.py` and the aleph-rs tooling emit `EPYC-v4` digests only. The change is that anything else now fails early instead of attesting wrong.
- Publisher-side multi-model digests and the nix golden-measurement job still cover one model; both are out of scope here.
- Version skew is worth knowing about: a new agent against an old daemon drops proto tag 6 silently and launches `EPYC-v4`. Harmless while publishers emit one model and the two ship together, not harmless after that.
- `cargo test` was not run locally (the supervisor suites need root); verified with `cargo clippy --locked --all-targets -- -D warnings` and `cargo fmt`. Python: 1168 passed, 10 pre-existing environmental failures verified identical on the base commit.


foxpatch-aleph (approved):
Well-designed change that replaces a hardcoded EPYC-v4 CPU model with a measured-selection flow: the agent intersects the message's launch-measurement vcpu_types with the host's QEMU probe results and carries the chosen model through proto, daemon, and controller to QEMU. The design fails closed (no launchable model = no VM, not a wrong-attestation VM), maintains byte-identical backward compatibility for pre-existing configs, and the value reaching QEMU argv is always host-provided (never a raw message string). Tests are comprehensive across unit and integration layers, including a model-axis expansion of the byte-parity conformance test. The only issue is a stale doc comment in build_snp_argv that still says the measurement 'pins -cpu EPYC-v4' when the model is now carried. The unrelated dev-branch fixture fix (registers= instead of digest=) is correctly scoped and necessary.

**`rust/crates/supervisor-controller/src/qemu.rs`** (line 531): Stale doc comment: still says 'which pins `-cpu EPYC-v4`' but the CPU model is now carried from the agent (via `config.cpu_model`), not pinned. The `snp_tee_fragment` doc above was updated to reflect the carried model, but this `build_snp_argv` doc was not. Suggest updating to match, e.g. 'the launch measurement (which includes `-cpu {model}`, the OVMF, kernel, initrd, cmdline and vcpu count)'.

foxpatch-aleph (approved):
Well-structured PR that carries the SEV-SNP launch CPU model from the message's measurements through the agent, proto, daemon, controller, and QEMU argv, replacing a hardcoded EPYC-v4. The selection logic (intersect requested models with host-provided QEMU probe results, first match in message order, host spelling) is correct and fail-closed. The value that reaches argv is always host-provided, preventing injection via the vcpu_type field. Backwards compatibility is preserved: absent cpu_model defaults to EPYC-v4 at every hop, and pre-existing controller configs parse and launch byte-identically. The byte-parity conformance test now crosses a cpu_model axis with an independent literal assertion. The unrelated LaunchMeasurement fixture fix (digest → registers) correctly carries to test_snp_instance_run.py via the shared fixture import. Only minor stale comments referencing EPYC-v4 as fixed in build_snp_argv's doc comments.

**`rust/crates/supervisor-controller/src/qemu.rs`** (line 504): Stale doc comment: "the exact OVMF + kernel + initrd + append + `EPYC-v4` + vcpu count" — EPYC-v4 is now parameterized via `cpu_model`. Consider updating to something like "the carried CPU model" to match the `snp_tee_fragment` doc above which was already updated.

**`rust/crates/supervisor-controller/src/qemu.rs`** (line 600): Stale inline comment: "This trio plus the vcpu count and EPYC-v4 CPU type is precisely what sev-snp-measure covered" — should reference the parameterized CPU model, not the hardcoded EPYC-v4 literal.

foxpatch-aleph (approved):
Well-executed PR that threads the SEV-SNP launch CPU model from the message's measurements through the agent, proto, daemon, and both argv builders, replacing a hardcoded EPYC-v4. The selection logic is correct: the agent intersects message-requested models with the host's QEMU probe, returns the host's spelling (safe against case-sensitivity), and fails closed with VmSetupError when no model is launchable. The value reaching QEMU is always host-provided, never a raw message string. Backwards compatibility is preserved: absent cpu_model defaults to EPYC-v4 at every hop, producing byte-identical argv for pre-existing configs. Test coverage is comprehensive across the pure function, both call sites, proto round-trip, Rust parity, and daemon config parsing. The divergence ledger update is thorough.

foxpatch-aleph (approved):
Well-executed PR that carries the QEMU CPU model from the message's launch measurements through the agent, proto, daemon, and controller to the QEMU argv, replacing the hardcoded EPYC-v4. The core selection logic (vcpu_select.py) is clean: read measurement vcpu_types in message order, deduplicate case-insensitively, intersect with the host's QEMU probe, fail closed on empty. The 'absent means EPYC-v4' default is consistently applied at every layer, preserving byte-identical argv for pre-existing configs. The byte-parity conformance test now crosses a model axis (None/EPYC-v4/EPYC-Genoa-v2) with an independent literal assertion, so the two builders' defaults cannot drift together unnoticed. Test coverage spans unit tests for the selection logic, integration tests for both spec builders (V-PROGRAM and SNP instance) covering the happy path, non-default model selection, and the fail-closed rejection, plus Rust tests for the controller config round-trip and the aleph-tee generator. The divergence doc (entry 68) is properly amended. No correctness or security issues found.

**`rust/crates/supervisor-controller/src/qemu.rs`** (line 531): The doc comment on build_snp_argv (line 531) still says "pins `-cpu EPYC-v4`" in the summary, but the CPU model is now parameterized. The comment at line 600-601 also references "EPYC-v4 CPU type" as what sev-snp-measure covered. These are stale relative to the change, though the detailed comment on snp_tee_fragment (lines 451-455) is correctly updated. Consider updating the build_snp_argv doc for consistency.

**`src/aleph/vm/agent/vcpu_select.py`** (line 24): The `_Measurement` Protocol uses `vcpu_type: str | None`, but the actual aleph_message models likely use `Optional[str]` (i.e. `str | None`). This is fine for a Protocol (structural typing), but worth noting that if the field were ever renamed on the message model, the Protocol would silently break at call time, not at import time. The TYPE_CHECKING guard mitigates this (no runtime cost), and the unit tests use a FakeMeasurement dataclass that matches, so this is an acceptable trade-off for avoiding a hard dependency on aleph_message in the module.

PR #1147, opened 2026-08-20
fix(packaging): ship the registers-schema aleph-message in the deb

odesenfans:
packaging/Makefile's pip install line must mirror pyproject.toml (its own comment says so), but #1143 moved the aleph-message git pin to 4ad261ff (measurement registers) only in pyproject. Debs built from dev since then bundle the old digest-schema models, so the CRN agent rejects any V-PROGRAM message published with the registers map:

```
2 validation errors for VerifiableProgramMessage
content.verification.measurements.0.digest: Field required
content.verification.measurements.0.registers: Extra inputs are not permitted
```

Observed live on the testnet validation run 32416035775 (aleph-cli 0.17.0-rc11 + pyaleph 0.10.3-rc9 publish/accept registers, the deb-installed agent refuses to start the VM).

One-line fix: move the Makefile pin to the same ref. Longer term the MERGE GATE comment already covers this: both pins disappear in favor of a released aleph-message.

### DIFF
--- a/packaging/Makefile
+++ b/packaging/Makefile
@@ -45,7 +45,7 @@ debian-package-code:
 	# MERGE GATE: aleph-message is git-pinned to the V-PROGRAM schema ref, swap for a release with pyproject.toml.
 	# protobuf is pinned to 5.29.6: 5.29.0 has a compilation issue; 6.x needs grpcio/tools 1.71+.
 	# legacy-cgi is packaging-only: it backfills the `cgi` module removed from the Python 3.13+ stdlib.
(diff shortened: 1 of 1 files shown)


foxpatch-aleph (approved):
One-line fix that aligns the packaging/Makefile aleph-message git pin with the pin already in pyproject.toml (both now 4ad261ff). The Makefile's own comment on line 44 states the versions MUST mirror pyproject.toml, and #1143 updated pyproject.toml without carrying the change to the Makefile, causing debs to bundle the old schema and reject V-PROGRAM messages with the registers map. The fix is correct, minimal, and the MERGE GATE comments already cover the longer-term plan of swapping both pins for a released aleph-message. No security concerns, no test coverage needed for a packaging pin alignment.

PR #1148, opened 2026-08-24
fix(agent): land the V-PROGRAM agent fixes from the validation branch

odesenfans:
Lands the four commits that so far lived only on the `od/vprogram-integration-3` validation aggregation branch (#1142), folded into three:

1. **feat(vprogram): map the attestation port to a host IPv4 port** (originally reviewed as #1079): the agent allocates a host port forwarded to the guest's :8443 attest agent and reports it on the execution, so clients can resolve an attested endpoint. Without it, `vprogram create --wait` times out with no `attested_endpoint` (aleph-testnets run 32361553700). The agent run.py reconcile (`reconcile_vprogram_port_forwards`), dropped in an earlier bad conflict resolution and restored by a separate repair commit on the aggregation branch, is squashed back into this commit where it belongs.
2. **fix(agent): heal port forwards when re-adopting a running VM** (originally reviewed as #1085): re-creates missing forwards when the agent restarts while VMs are up.
3. **fix(agent): update_refs must not derive refs from a V-PROGRAM**: the update watcher treated any non-Instance content as a program and read `.code`/`.data`/`.runtime.ref`, crashing with `'VerifiableProgramContent' object has no attribute 'data'` after every V-PROGRAM boot. V-PROGRAMs are immutable with hash-pinned refs, so it returns no refs.

Context: #1079/#1085 were reviewed and merged into the staging branch `od/vprogram-portmap-base`, which stalled and never reached dev; the third commit was written during the first live testnet run. All three are cherry-picked from `od/vprogram-integration-3`, already conflict-resolved against current dev. The resulting tree is **byte-identical** to that branch's tip, which passed the full aleph-testnets validation (runs 32368297653 and 32426136992, 28/28: SNP boot, attested exec + compose V-PROGRAM calls, port-mapped attestation endpoints).

Once merged, #1142 can close and the testnets manifesto can point at plain dev.

### DIFF
--- a/src/aleph/vm/agent/run.py
+++ b/src/aleph/vm/agent/run.py
@@ -34,7 +34,11 @@ from aleph.vm.agent.translate import build_create_vm_spec, build_program_create_
 from aleph.vm.agent.update_watcher import UpdateWatcher
 from aleph.vm.agent.vm.program_client import ProgramGuestClient
 from aleph.vm.agent.vm_registry import AgentVmRegistry, persist_record
-from aleph.vm.agent.vprogram_launch import build_vprogram_spec, remove_vprogram_staging
+from aleph.vm.agent.vprogram_launch import (
+    build_vprogram_spec,
+    remove_vprogram_staging,
+    resolve_vprogram_attestation_port,
+)
 from aleph.vm.conf import settings
 from aleph.vm.resources import InsufficientResourcesError
 from aleph.vm.supervisor_interface import errors as supervisor_errors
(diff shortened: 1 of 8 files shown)


foxpatch-aleph (approved):
Well-structured PR with three focused V-PROGRAM agent fixes: attestation port mapping, port-forward healing on re-adoption, and the update_refs crash guard. The strict-mode design for convergence callers, the best-effort healing contract, and the idempotent reconcile machinery are all correct and well-documented. Test coverage is thorough across create, re-adoption, failure, and idempotency paths. No bugs or security issues found.

**`tests/supervisor/test_supervisor_run_routing.py`** (line 602): Minor coverage gap: there is no test for the DEFINED/BOOTING re-adoption path where the VM reaches RUNNING via `_wait_until_running` and then `reconcile_adopted_port_forwards` is called. The STOPPED and RUNNING branches are tested, and the `awaiting_confidential_init` guard is tested, but the DEFINED/BOOTING -> heal path is not. This is low-risk since the logic is straightforward (info stays non-None, not awaiting init, so heal runs), but a test would close the matrix.

PR #1149, opened 2026-08-24
fix(storage)!: give the agent sole ownership of a VM's disks

odesenfans:
Fixes two defects in the delete/reinstall paths and moves storage deletion to the side that allocates it. Targeting 2.0: one is a regression against `main`, and the proto change is breaking, so this is the right window for it.

## The asymmetry

The agent allocates every volume — `qemu-img create` for the rootfs, `fallocate`/`mkfs.ext4` for persistent volumes, placed through `storage_pools` — and hands the supervisor resolved absolute paths on a `CreateVmSpec`. `VmExecution.prepare()` says so outright: *"No download (paths are resolved)"*. `storage_pools.py` says so too: *"everything here is agent policy"*.

Deletion, however, lived supervisor-side (`VmExecution.erase_volumes`, driven from `LocalSupervisor.delete_vm`/`reinstall_vm`). That inversion is the root cause of both bugs below, and of the four separate agent-side cleanup calls bolted on beside every `delete_vm`.

## The two defects

**Deleting another tenant's runtime.** The supervisor cannot tell a per-VM volume from a shared cache entry — a spec is just paths. For a Firecracker program, `resources.rootfs_path` *is* `RUNTIME_CACHE/<runtime_ref>`, shared by every program on that runtime, and `erase_volumes(include_rootfs=True)` unlinked it. On `main` this self-healed because `prepare()` re-downloaded; since the agent/supervisor split it is permanent and cross-tenant.

**Reinstall leaves the VM unbootable — a regression against `main`.** `reinstall_vm` erased the rootfs, then called `execution.prepare()` and `restart_persistent_vm`. On the spec path `prepare()` downloads nothing and `restart_persistent_vm` (`pool.py:479`) only re-registers the execution and restarts the systemd unit. Nothing recreates the qcow2 the agent originally built, so the controller restarts against a file that no longer exists. On `main`, `prepare()` was still *"Download VM required files"*, so reinstall worked. Both are mirrored in the Rust daemon (`lifecycle.rs`).

## The change

Whoever allocates, deallocates.

- **`DeleteVm` leaves the VM stopped with no handles held on its storage, and never touches a byte of it.** `wipe` is removed from `DeleteVmRequest` and its field number reserved.
- **`ReinstallVm` is removed outright.** Reinstall is agent orchestration: stop → purge → rebuild → start for plain instances and programs (keeping the execution, so the VM holds its `vm_index` and addresses), or delete → purge → create for ephemeral and confidential VMs, whose rootfs is measured and staged rather than built from the message.
- **`agent/vm/purge.py`** owns deletion, scoped to `{pool}/{vm_hash}/` across pools plus the session and staging directories. A cache entry is not merely spared by a check — it is unreachable from that code.
- **`operate_erase` now erases the rootfs.** An owner asking to "delete all data stored by a virtual machine" meant all of it; the rootfs, where most of it lives, was being kept. That was a confidentiality gap as much as a space leak.
- **`volume_path_for` gains namespace stickiness**: a volume rebuilt after a purge lands back on the pool the VM already occupies, so the paths in the running VM's spec stay valid across a reinstall.

## Behaviour changes worth reviewing

- Reinstalling a **persistent confidential VM** now goes through delete+create rather than an in-place restart, so its `vm_index` (and addresses) can change. Its measurement changes and the owner must re-attest regardless, so this seemed the honest trade rather than special-casing a staged rootfs into the in-place path.
- **Migration hand-off** (`views/migration.py`) and every other `delete_vm` caller keep their disks by construction now, rather than by passing `wipe=False`.

## Breaking

Agent and daemon must be updated together. There is no deployed skew to bridge: the gRPC boundary is new in this release.

## Testing

- Full `tests/supervisor` suite: 1119 passed, same 44 pre-existing environment failures as `dev` (pyroute2 needs root, dbus/systemd stubs, an `aleph_message` version skew in the local venv). `test_snp_instance_{launch,run}.py` fail to collect on `dev` too, for the same venv reason.
- New `tests/supervisor/test_vm_purge.py` (14 tests) pins the ownership boundary: the purge spans every pool, honours the rootfs/data-volume split, is idempotent, refuses an implausible namespace, and — the regression guards — never reaches the shared runtime cache or another VM's directory. One test covers the reinstall round trip: a purged volume is recreated on the same pool.
- `test_supervisor_inprocess_lifecycle.py` gains a guard that nothing supervisor-side can erase volumes any more, plus a test that `DeleteVm` leaves real files on disk. The Rust daemon gets the same two contract tests.
- `cargo clippy --locked --all-targets -- -D warnings` and `cargo fmt --check` clean; `ruff format`, `isort` and `mypy` clean (mypy's 7 errors in `operator.py` are pre-existing and identical on `dev`). Rust suites left to CI per the usual local-root constraint.

## Follow-up

This is the first piece of the 2.1 disk-reclamation work. Still open, by design: nothing yet deletes the disks of a VM whose message is forgotten (`tasks.py:404` and friends still delete without reclaiming), the runtime/code/data caches remain unbounded and uncapped on download, backups expire only when another backup starts, and loop devices from `create_devmapper` are never detached.


foxpatch-aleph (approved):
This PR correctly moves storage deletion from the supervisor to the agent side, fixing two real bugs: cross-tenant runtime cache deletion (a security issue where a program's shared rootfs was unlinked) and the unbootable-reinstall regression (where the rootfs was erased but never recreated on the spec path). The new purge.py module is well-designed with defense-in-depth namespace validation, scoped exclusively to {pool}/{vm_hash}/ directories that the agent itself created, making shared cache entries structurally unreachable rather than merely checked-against. The volume_path_for stickiness change ensures recreated volumes land back on the same pool. The proto changes are properly breaking (reserved field number for wipe, ReinstallVm RPC removed) and appropriate for a 2.0 release. Test coverage is comprehensive: 14 new tests in test_vm_purge.py pin the ownership boundary, the regression guards (never reaching shared cache or other VMs), and the reinstall round trip. The Rust daemon changes mirror the Python side correctly. The only issues found are cosmetic: a stale comment referencing 'reinstall' in local.py:527 and events.rs:5, and an indentation inconsistency in test_supervisor_abc.py:83.

**`src/aleph/vm/supervisor/local.py`** (line 527): Stale comment: references 'reinstall' which no longer exists. The comment says 'same trick as reinstall' but reinstall_vm was removed in this PR. Should reference 'stop_vm and the restore backup path' or similar, since restore_backup in local.py:911 uses the same stop_event trick.

**`rust/crates/supervisor-daemon/src/events.rs`** (line 5): Stale comment: lists 'reinstall' among lifecycle transitions the daemon performs, but ReinstallVm was removed from the daemon in this PR. Should read 'create/stop/start/reboot/delete'.

**`tests/supervisor/test_supervisor_abc.py`** (line 83): Cosmetic: 'run_program_code' has 20 spaces of indentation while the surrounding set members have 12. This is valid Python (inside a set literal) so ruff doesn't flag it, but it's inconsistent with the rest of the block.

**`src/aleph/vm/agent/views/operator.py`** (line 729): If start_vm fails after a successful in-place stop+purge+recreate, the VM is left stopped with fresh but unbootable volumes. This is the same failure mode as the old reinstall_vm path (where prepare or restart_persistent_vm could fail), so it's not a regression. Worth noting for the follow-up disk-reclamation work: a failed in-place reinstall leaves the VM in a state that may need manual intervention, unlike the delete+create path which would simply fail to create and return an error.

**`src/aleph/vm/agent/vm/purge.py`** (line 80): The rootfs detection uses entry.stem == ROOTFS_STEM, which means a file named 'rootfs.qcow2' has stem 'rootfs' and is correctly identified. However, a volume named with a name that happens to be 'rootfs' (e.g., a user-created persistent volume named 'rootfs') would also be treated as the rootfs. This is unlikely given the naming conventions but worth being aware of. Not a blocking issue since the downloader always names the rootfs 'rootfs' and user volumes use their declared names.

foxpatch-aleph (commented):
Architecturally sound fix that correctly moves storage deletion to the agent side. The core logic, security boundaries, and test coverage are excellent. The only issues are stale documentation references to removed APIs (ReinstallVm, erase_volumes, wipe) that should be updated.

foxpatch-aleph (commented):
The core design is sound: moving storage deletion to the agent side (which allocates the volumes) correctly fixes both the cross-tenant runtime cache deletion bug and the reinstall-leaves-VM-unbootable regression. The new purge.py is well-structured with proper namespace validation and defense-in-depth against path traversal. Test coverage is thorough. However, the documentation in docs/architecture/vm-lifecycle.md and docs/architecture/divergences.md is significantly stale, still describing the removed ReinstallVm RPC, erase_volumes, and DeleteVm(wipe) behavior. There is also a multi-pool edge case in _pool_holding_namespace where volumes spread across different pools may not be recreated on their original pool after a purge, though the common single-pool case works correctly.

**`docs/architecture/vm-lifecycle.md`** (line 269): This entire section is stale. It describes `ReinstallVm`, `erase_volumes(include_rootfs=true, ...)`, and `wipe_volumes` — all of which have been removed. The section should be rewritten to describe the agent-side reinstall flow: stop → purge_vm_volumes → recreate_vm_volumes → start (for in-place), or delete → purge → create (for ephemeral/confidential).

**`docs/architecture/vm-lifecycle.md`** (line 323): The "Erase-on-delete rules" section describes `DeleteVm(wipe)` and `erase_volumes` behavior that no longer exists. DeleteVm no longer takes a `wipe` parameter and never touches storage. The erase-on-delete semantics now live entirely in the agent's `purge_vm_storage` (called from `operate_erase`).

**`docs/architecture/divergences.md`** (line 68): Entry 51 references "the wipe erase_volumes" in DeleteVm, which has been removed. The backup disk lock is still taken in delete_tracked_vm, but the erase_volumes call it was protecting is gone. The entry should be updated to reflect that the lock now only guards the world-entry removal and backup registry reap, not a wipe.

**`docs/architecture/divergences.md`** (line 60): Entry 43 describes the old `erase_volumes` crash on programs and the `ReinstallVm(wipe_volumes=true)` behavior. Both have been removed entirely. This entry should note that the behavior is now obsolete: ReinstallVm is gone, erase_volumes is gone, and program storage is managed agent-side by purge.py which is structurally incapable of reaching the shared runtime cache.

**`src/aleph/vm/storage_pools.py`** (line 340): When a VM has volumes spread across multiple pools (e.g., rootfs on pool0, data volume on pool1), _pool_holding_namespace returns the first eligible pool for every volume, not the pool that held that specific volume. After a purge, all namespace directories remain (purge_vm_volumes only deletes files), so a data volume originally on pool1 could be recreated on pool0, producing a path that doesn't match the spec the supervisor still holds. The common case (all volumes on one pool) works correctly, but this multi-pool edge case could cause a boot failure after reinstall. Consider tracking per-volume placement or documenting this limitation.

**`src/aleph/vm/agent/views/operator.py`** (line 673): purge_vm_storage() does synchronous file I/O (unlink, shutil.rmtree, iterdir) in an async request handler. This follows the existing pattern (the old remove_vprogram_staging was also sync), but for a full purge across multiple pools it could block the event loop longer. Consider running in a thread executor if this becomes a concern at scale.

**`tests/supervisor/views/test_operator.py`** (line 960): The erase test (test_operator_erase_with_delegation) verifies delete_vm is called but does not verify that purge_vm_storage was called. Since the erase path now owns the rootfs deletion (a confidentiality-relevant change), an assertion or mock verifying purge_vm_storage is invoked would strengthen the test.

foxpatch-aleph (commented):
The core design is sound and well-executed: moving storage deletion to the agent side (which allocates the volumes) fixes both the cross-tenant runtime cache deletion and the reinstall-leaves-VM-unbootable regression. The proto changes are clean (field 2 reserved, ReinstallVm removed), the pin_layout context manager elegantly solves the pool-stickiness problem for reinstalls, and the test suite is thorough. One real issue: purge_vm_storage calls shutil.rmtree on the namespace directory after purge_vm_volumes, which bypasses the dm-held protection that purge_vm_volumes carefully provides — rmtree will unlink a dm-held .btrfs file's directory entry, leaving an orphaned loop device and pinned inode, exactly the scenario the code documents as harmful. This only affects QEMU VMs with parent-backed btrfs volumes on the operate_erase path (Firecracker tears down dm targets on stop; QEMU does not), which the PR defers, but the rmtree should at least not silently undo the protection. A few stale comments and names in the conformance test are minor.

**`src/aleph/vm/agent/vm/purge.py`** (line 170): Bug: `shutil.rmtree(volumes_dir)` bypasses the dm-held protection in `purge_vm_volumes`. When `purge_vm_volumes` skips a `.btrfs` volume whose device-mapper target is still present (line 116), `rmtree` on line 170 unlinks it anyway — the directory entry goes, but the loop device pins the inode (no space freed) and `create_devmapper` would skip the rebuild. This is exactly the scenario the code at lines 116-126 says it's avoiding. For QEMU VMs with parent-backed btrfs volumes (no dm teardown on stop, unlike Firecracker), the `operate_erase` path silently undoes the protection `purge_vm_volumes` provides. Consider checking for dm-held volumes before rmtree, or at minimum logging a warning when rmtree removes a directory that contained a dm-held file.

**`tests/conformance/test_rust_daemon_lifecycle.py`** (line 324): Stale comment: says "this one keeping the persisted port rows" but the `delete_vm` call on line 325 does not pass `keep_port_mappings=True` — it's a plain delete that removes port mappings. The original test had `wipe=True, keep_port_mappings=True` for KEPT_HASH, but now both VMs are deleted identically. The comment and the `KEPT_HASH`/`kept_row` variable names should be updated to reflect the actual test.

**`tests/supervisor/views/test_operator.py`** (line 1738): Missing coverage: there is no test for a persistent confidential VM going through the from-scratch reinstall path. The new code passes `persistent=record.persistent` to `create_vm_execution_or_raise_http_error` (operator.py line 766), which for a persistent confidential VM would be `True` — an improvement over the old code that defaulted to `False`. A test pinning `persistent=True` for a confidential persistent VM on the from-scratch path would guard this new behavior.

foxpatch-aleph (approved):
Well-executed architectural fix that moves storage deletion to the agent side, resolving two real bugs (cross-tenant runtime cache deletion and broken reinstall). The ownership model is sound, the new purge.py is properly scoped with namespace validation and device-mapper guards, pin_layout ensures volumes are recreated on the same pool, and test coverage is comprehensive across both Python and Rust daemons. The few edge cases I found (retry-ability of failed from-scratch reinstall, narrow VmNotFoundError race after purge) are pre-existing or practically unreachable.

**`src/aleph/vm/agent/vm/purge.py`** (line 169): Minor inefficiency: `iter_volume_files(namespace)` is called once per `volumes_dir` inside the loop, scanning all pools each time. For the expected 1-3 pools this is negligible, but could be hoisted above the loop and filtered per-directory instead.

**`src/aleph/vm/agent/views/operator.py`** (line 736): If `start_vm` raises `VmNotFoundError` here (after volumes are purged), the re-raise is caught by the outer `except VmNotFoundError` on line 768 and returned as 404, hiding that storage was already deleted. This is practically unreachable since `stop_vm` re-registers the execution with a fresh `stop_event`, but the semantics are slightly misleading — a 404 implies the VM was never found, when in fact it was stopped, purged, and then lost.

PR #1150, opened 2026-08-24
fix(agent): retry a failed SNP vCPU probe instead of caching it forever

odesenfans:
Spotted while assessing #1145, and independent of it: this lands cleanly on either side, and #1145 is what turns it from "node quietly drops out of the confidential pool" into "node also refuses every SNP launch".

## The problem

Two pieces that are each fine on their own:

`get_supported_snp_vcpu_types` turns every probe failure into a normal return:

```python
try:
    definitions = await query_cpu_definitions()
except Exception:
    logger.warning("QEMU vCPU probe failed, not advertising SNP guest models", exc_info=True)
    return []
```

and `@async_cache` (`utils/__init__.py:273`) keeps whatever the function returns, forever. So `[]` from "this host has no SNP" (a hardware fact, fine to cache) and `[]` from "the qemu spawn timed out at 15s during boot" (transient) are indistinguishable to the cache, and both stick for the life of the process.

The list feeds two things. It is what `/about/usage/system` advertises in `properties.tee.sev_snp.supported_vcpu_types`, which the scheduler's v-program gate reads, so a stuck `[]` silently removes the node from the confidential pool. And once #1145 lands, `select_snp_vcpu_type` fails closed on an empty host list, so the same stuck `[]` also refuses every SNP launch. Both are the correct fail-closed outcome. The recovery is a manual agent restart, and the cause was logged once at startup.

**There was a second layer.** `get_machine_properties` is `@async_cache`d too, and built the tee block *inside* that cache. Fixing the probe alone would have restored launches while the advertisement stayed frozen, which is the side the scheduler sees. `test_advertisement_recovers_after_a_failed_first_probe` is the end-to-end check for that.

## The change

| outcome of a probe | remembered for |
|---|---|
| success | the life of the process (a fact about this QEMU + kernel + silicon) |
| failure | `PROBE_RETRY_SECONDS` (60s), then tried again |

So a transient failure heals within a minute with no operator action, and a host with no working qemu costs one 15s probe per minute rather than one per request. A lock closes a small pre-existing race where the usage endpoint and a launch could both spawn a qemu at startup.

`get_machine_properties` is split: `_get_static_machine_properties` keeps the cache (it calls `get_hardware_info`, which shells out), and the tee block is re-evaluated per call, which is cheap once the probe holds its own result. `test_static_properties_stay_cached_across_calls` pins that the split doesn't cost a hardware scan per poll.

The generic `async_cache` is left alone: "cache every return" is the right semantics for its other users. Only the probe has a return value that isn't a fact about the host.

`reset_snp_vcpu_probe_cache()` exists for tests, replacing the `.__wrapped__()` trick the old tests used to bypass the decorator.

## Tests

Six new in `test_vprogram_probe.py`, plus the existing seven updated to the new reset hook:

- a successful probe is not repeated
- a failed probe is not repeated within the retry window
- a failed probe is retried after the window and recovers (the case that used to need a restart)
- concurrent first callers probe once
- the advertisement recovers after a failed first probe (end to end via `get_machine_properties`)
- the static properties stay cached across calls

Verified these discriminate rather than pass vacuously: with the old semantics injected (failure pinned as `supported = []`, tee built inside the static cache), exactly the two recovery tests fail and the other eleven still pass.

Locally: 13 passed, `ruff format` clean, `isort --profile black` clean, `mypy` clean on the three files. The one remaining `ruff check` finding in the test file (`PLC0415` at line 134) is pre-existing on `dev`. I could not build the `hatch` testing env here (`dbus-python` fails to compile on this machine), so this ran in a venv with the same pins minus that package; CI is the authority.

### DIFF
--- a/src/aleph/vm/agent/resources.py
+++ b/src/aleph/vm/agent/resources.py
@@ -179,11 +179,16 @@ async def _get_tee_properties() -> TeeProperties | None:
 
 
 @async_cache
-async def get_machine_properties() -> MachineProperties:
-    """Fetch machine properties such as architecture, CPU vendor, ...
-    These should not change while the supervisor is running.
+async def _get_static_machine_properties() -> MachineProperties:
+    """The part of the machine properties that cannot change while the agent
+    runs: architecture, vendor and the SEV feature flags. Cached because
+    ``get_hardware_info`` shells out.
 
-    In the future, some properties may have to be fetched from within a VM.
+    ``tee`` is deliberately left ``None`` here and filled in per call by
+    ``get_machine_properties``: it depends on the QEMU probe, which can fail
+    transiently and then recover, and caching it alongside the static facts
+    would freeze a failed probe's empty advertisement for the life of the
+    process.
     """
     hw = await get_hardware_info()
     cpu_info = get_cpu_info(hw)
@@ -202,8 +207,22 @@ async def get_machine_properties() -> MachineProperties:
                 )
             ),
         ),
-        tee=await _get_tee_properties(),
+        tee=None,
     )
+
+
+async def get_machine_properties() -> MachineProperties:
+    """Fetch machine properties such as architecture, CPU vendor, ...
+
+    The static part is cached; the TEE block is re-evaluated on every call so
+    that a probe which failed once and has since recovered is advertised
+    again. Cheap once the probe has succeeded, since the probe keeps its own
+    result.
+
+    In the future, some properties may have to be fetched from within a VM.
+    """
+    static = await _get_static_machine_properties()
+    return static.model_copy(update={"tee": await _get_tee_properties()})
 
 
 @async_cache
--- a/src/aleph/vm/agent/resources.py
+++ b/src/aleph/vm/agent/resources.py
@@ -167,9 +167,6 @@ async def _gpus_from_host_info(host_info) -> GpuProperties:
     )
 
 
-machine_properties_cached = None
-
-
 async def _get_tee_properties() -> TeeProperties | None:
     """TEE launch capability, absent when there is nothing provable to advertise."""
     snp_vcpu_types = await get_supported_snp_vcpu_types()
@@ -226,7 +223,12 @@ async def get_machine_properties() -> MachineProperties:
 
 
 @async_cache
-async def get_machine_capability() -> MachineCapability:
+async def _get_static_machine_capability() -> MachineCapability:
+    """The part of the machine capability that cannot change while the agent
+    runs. Cached because ``get_hardware_info`` shells out. ``tee`` is left
+    ``None`` and filled in per call by ``get_machine_capability``, for the
+    same reason as ``_get_static_machine_properties``.
+    """
     hw = await get_hardware_info()
     cpu_info = get_cpu_info(hw)
     mem_info = get_memory_info(hw)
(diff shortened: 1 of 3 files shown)


foxpatch-aleph (commented):
The core fix is correct and well-designed: splitting get_machine_properties so the TEE block is re-evaluated per call while static properties stay cached, and replacing the permanent async_cache on the probe with a retry-aware _ProbeCache that distinguishes success (cached forever) from failure (cached for 60s then retried). The lock correctly prevents concurrent QEMU spawns. Tests are thorough and discriminate properly. One observation: get_machine_capability (line 228) is still @async_cache'd and still builds tee inside that cache (line 258), so the /about/capability endpoint retains the exact bug this PR fixes for /about/usage/system. This may be a deliberate scope limitation since the scheduler reads the usage endpoint, not the capability endpoint, but it's worth noting for completeness or a follow-up. The pre-existing dead variable machine_properties_cached at line 170 is not introduced by this PR but could be cleaned up.

**`src/aleph/vm/agent/resources.py`** (line 228): get_machine_capability is still @async_cache'd and still calls _get_tee_properties() at line 258, so the /about/capability endpoint has the same bug this PR fixes for get_machine_properties: a failed probe at startup freezes tee=None in the cache forever. If this is a deliberate scope limitation (the scheduler reads the usage endpoint, not the capability endpoint), a comment noting that would help. If not, the same split pattern should be applied here.

**`src/aleph/vm/agent/resources.py`** (line 170): machine_properties_cached = None is dead code — defined but never read anywhere in the codebase. Pre-existing, not introduced by this PR, but a natural cleanup opportunity given the caching refactor in this file.

odesenfans:
Both points taken, in a separate commit rather than an amend so the delta is reviewable:

- **`/about/capability`**: same split applied. `_get_static_machine_capability` keeps the cache, `get_machine_capability` composes the tee block per call. `test_capability_advertisement_recovers_after_a_failed_first_probe` mirrors the usage-endpoint test, and fails against the previous commit (verified by reinjecting the frozen-tee version). You were right that the scheduler doesn't read it; I don't think that's a good enough reason to leave one of two identical caches broken in the same file.
- **`machine_properties_cached`**: removed. One definition, zero reads.

14 passed locally, format/isort/mypy clean.

foxpatch-aleph (approved):
This PR correctly fixes a real bug where a transient QEMU probe failure at startup was cached forever by `@async_cache`, silently removing the node from the confidential pool until manual restart. The solution is well-designed: `_ProbeCache` distinguishes successful probes (cached for process lifetime as hardware facts) from failed ones (retried after 60s), with an `asyncio.Lock` preventing duplicate concurrent probes. The `resources.py` split correctly moves TEE evaluation out of the `@async_cache`d static properties into a per-call layer using `model_copy(update={"tee": ...})`, which creates a new object without mutating the cached one. The six new tests discriminate correctly (the PR description verifies that the two recovery tests fail under old semantics while the other eleven pass). Code quality is high: docstrings explain the "why" rather than the "what", naming is clear, and the `reset_snp_vcpu_probe_cache()` hook is a clean replacement for the `.__wrapped__()` bypass trick. No correctness, security, or performance issues found.

**`src/aleph/vm/agent/vcpu_probe.py`** (line 111): The `reset()` method doesn't recreate `self.lock`. This is correct for production (where `reset()` is never called) and works in tests because Python 3.10+ `asyncio.Lock` doesn't bind to an event loop at creation time, and all tests release the lock via `async with` before the event loop closes. If a future test were to leave the lock held, a subsequent test in a new event loop could deadlock — but that would be a bug in that test, not in this code.

**`tests/supervisor/test_vprogram_probe.py`** (line 87): The `_fresh_static_properties` fixture re-wraps `__wrapped__` with a new `async_cache` per test, which correctly gives each test a clean cache. This relies on `functools.wraps` setting `__wrapped__` on the `async_cache` wrapper, which it does. One subtle point: if `async_cache` were ever changed to not use `functools.wraps`, this fixture would break silently — but that's an unlikely change and the failure would be immediate in CI.

PR #1151, opened 2026-08-24
Move backups to the agent; the supervisor keeps guest quiescence

odesenfans:
## Summary

The supervisor never allocated a VM's disks; the agent did, and since #1149 the agent also deletes them. Backups were the last storage-shaped thing the supervisor owned: it chose `BACKUP_DIRECTORY`, expired archives, checked free space and swapped rootfs files, duplicating in two languages what the agent already does for every other byte on disk. This PR finishes the rule ("whoever allocates, deallocates" and, now, copies):

- **The agent owns the archives.** `src/aleph/vm/agent/vm/backup.py` (`BackupManager`, one per webapp at `app["backups"]`) resolves a VM's disks from its namespace directories (`{pool}/{vm_hash}/rootfs.qcow2` plus the writable data volumes, a device-mapper volume read through its `/dev/mapper` device), copies them, stores, lists, streams, expires and deletes the archives, and restores from them by driving `StopVm`/`StartVm` around the rootfs swap, the shape `operate_reinstall` already has. The qemu-img and tar work moves from the QEMU controller to the neutral `aleph.vm.backup` package. The operator HTTP contract (`/control/machine/{ref}/backup*`, `/restore`) is unchanged.
- **The supervisor keeps quiescence only.** `FreezeGuest`/`ThawGuest` replace `StartBackup`, `GetBackupStatus`, `ListBackups`, `DownloadBackup`, `DeleteBackup`, `RestoreBackup` and `RestoreFromImage` on the wire (`ERROR_CODE_BACKUP_NOT_FOUND` reserved). Freeze is best effort (`frozen=false` without a guest agent or for a stopped VM), idempotent while it holds, and the supervisor thaws a guest itself after `GUEST_FREEZE_TIMEOUT` (600 s) without a `ThawGuest`, so an agent that dies mid-copy cannot leave a guest with its filesystems frozen. The Rust daemon tags each freeze with a generation so a stale timer never thaws a newer freeze.
- **Rust:** `backup.rs`, the restore paths, the download stream cap, the backup fixtures and the daemon's `sha2`/`tar` dependencies go; `quiesce.rs` comes in.
- **Docs:** `vm-lifecycle.md`, `wire-contract.md`, `overview.md`, `storage.md`, `testing.md`; divergence entries 47, 48, 51, 52 resolved with the surface they described, 44 and 53 trimmed.

`RestoreFromImage` was not in the original list of six but is the same category (the supervisor validating and swapping a rootfs file); leaving it would have kept one storage-writing RPC on the wire, so it moved with the rest.

## Test plan

- [x] `tests/supervisor`: the agent `BackupManager` suite (`test_agent_backups.py`, ported from the supervisor one, plus freeze/thaw ordering and the thaw-on-failure path), `test_supervisor_quiesce.py` (freeze/thaw, idempotency, no guest agent, stopped VM, auto-thaw timeout and its cancellation), the archive tests (`test_backup_archive.py`), the proto binding/ABC/gRPC/round-trip updates and the operator view tests. Same 46 environment-only failures as `dev`.
- [x] `cargo fmt --check`, `cargo clippy --locked --all-targets -- -D warnings`, `cargo test --workspace` (the `quiesce` tests drive a fake QGA socket).
- [x] `scripts/check_proto_clean.sh`.
- [ ] CI: the Tier 2 integration leg (`tests/integration/test_guest_quiesce.py` replaces `test_backup_restore.py`; the full backup/restore cycle on a live VM now lives above the wire and needs an agent-level integration test, not included here).


foxpatch-aleph (approved):
This PR cleanly moves backup archive ownership from the supervisor to the agent (whoever allocates, deallocates, and copies), replacing the 6+1 backup RPCs with just FreezeGuest/ThawGuest for guest quiescence. The implementation is thorough across Python, Rust, proto, docs, and tests. The agent's BackupManager correctly resolves disks from namespace directories, serializes backup/restore per-VM, and drives stop/start through the supervisor for restores. The Rust quiesce module with generation-tagged auto-thaw is well-designed to prevent stale-timer races. Security properties are preserved: path traversal validation, member-streamed tar extraction (no extractall), and namespace-scoped disk resolution. There is a minor resource leak in thaw_guest (both Python and Rust) where a VM deleted between freeze and thaw causes NotFound before the frozen record is removed — but the auto-thaw timer eventually cleans it up, and the Rust daemon's DeleteVm calls forget() which mitigates it. The start_backup check-and-register section has an await point (disk space check) that could theoretically allow a double-spawn under concurrency, but this is parity with the old Python supervisor's behavior. Test coverage is comprehensive across all new surfaces.

**`src/aleph/vm/supervisor/local.py`** (line 628): Minor resource leak: thaw_guest calls self._require(vm_id) before popping _frozen_guests. If the VM is deleted between freeze and thaw, NotFound is raised and the frozen record (plus its auto-thaw timer) leaks until the timer fires. Consider moving the pop before the _require check, or wrapping _require in a try/except that still cleans up. The Rust side mitigates this via forget() on DeleteVm, but the Python supervisor has no equivalent cleanup on its delete path.

**`rust/crates/supervisor-daemon/src/quiesce.rs`** (line 122): Same pattern as the Python side: entry_snapshot check before frozen_guests.remove. If the VM is deleted by means other than DeleteVm (which calls forget), the frozen record leaks until auto_thaw fires. Low severity since DeleteVm is the normal deletion path and it calls forget().

**`src/aleph/vm/agent/vm/backup.py`** (line 196): The check-and-register critical section in start_backup has an await point (check_disk_space_for_multiple) between the running-task check and the job insertion. Two concurrent start_backup calls for the same VM could both pass the running-task check and both spawn a run. This is parity with the old Python supervisor (same event-loop yielding), and the Rust admission lock that prevented it is now gone. Not a regression, but worth noting if the agent ever moves to a multi-threaded executor.

foxpatch-aleph (commented):
A well-executed architectural move of backup ownership from the supervisor to the agent, replacing 7 backup/restore RPCs with 2 quiescence RPCs (FreezeGuest/ThawGuest). The new BackupManager, quiesce module, and archive package are clean and well-tested. The generation-tagged auto-thaw in the Rust daemon is a nice safety mechanism. The main concern is that the second-precision intermediate-file-name collision (formerly ledger entry 48, now marked "resolved") still exists in the moved code at archive.py:83 — since the parity constraint is gone, this would have been the time to fix it with a per-disk unique name. The race condition in start_backup (no admission lock around the await point) is pre-existing in the Python implementation.

**`src/aleph/vm/backup/archive.py`** (line 83): The second-precision timestamp (`%Y%m%dT%H%M%SZ`) for the intermediate `.qcow2` file name means that when `include_volumes=True` with 2+ disks, multiple `create_qemu_disk_backup` calls within the same wall-clock second will all write to the same destination path (`{vm_hash}-{timestamp}.qcow2`), each overwriting the previous. The archive then tars whichever file won last under every member name. This was formerly ledger entry 48, marked "resolved" in this PR because the parity constraint is gone — but the bug still exists in the code. Since there is no longer a Rust daemon to keep in parity with, this could now be fixed with a per-disk unique name (e.g. using the member name or the backup id's microsecond stamp).

**`src/aleph/vm/agent/vm/backup.py`** (line 196): There is an `await` point at `check_disk_space_for_multiple` between the running-task check (line 186) and the task registration (line 216). Two concurrent `start_backup` calls for the same VM could both pass the running-task check before either registers a task, then both spawn backup tasks, with the second overwriting the first in `self._tasks[vm_hash]`. The deleted Rust code had an admission lock for this; the old Python code had the same structure, so this is pre-existing rather than a regression.

**`src/aleph/vm/agent/vm/backup.py`** (line 314): Since `download_backup` is an async generator (it contains `yield`), the `BackupNotFoundError` at line 314 will not be raised when the method is called — only when the caller starts iterating (first `__anext__`). This works in practice because the operator view checks existence via `get_backup_status` first, but it is a subtle behaviour worth being aware of: a caller who catches the exception at call time would miss it.

foxpatch-aleph (approved):
Well-executed architectural refactor that moves backup archives to the agent (who owns the disks) and reduces the supervisor's role to guest quiescence (FreezeGuest/ThawGuest). The new BackupManager, Rust quiesce module, and archive package are clean and well-tested. The auto-thaw generation tagging, thaw-on-failure paths, and idempotency semantics are all correct. The intermediate filename collision fix is a nice bonus. Minor concurrency and type-annotation nits noted but none are blocking.

**`src/aleph/vm/agent/vm/backup.py`** (line 175): The `_admission` lock is global (not per-VM) and is held across `check_disk_space_for_multiple` (line 205), which spawns `qemu-img info` subprocesses. This serializes all backup starts across all VMs during disk I/O. Consider moving the disk-space check outside the admission lock, or making the admission lock per-VM, to match the concurrency the old Rust code had.

**`src/aleph/vm/agent/vm/backup.py`** (line 319): Type annotation `-> AsyncIterator[bytes]` on an `async def` that returns an async generator (via `return _stream_file(...)`) is technically misleading — the actual return type is `Coroutine[Any, Any, AsyncIterator[bytes]]`. The caller correctly uses `async for chunk in await backups.download_backup(...)`, so this works at runtime, but the annotation could confuse type checkers.

**`src/aleph/vm/agent/vm/backup.py`** (line 284): `_freeze` catches only `SupervisorError`. If `freeze_guest` raised a non-`SupervisorError` exception (e.g., a network error from a buggy transport), the guest could be left frozen without a thaw. In practice the gRPC client only raises `SupervisorError` subclasses, so this is safe today, but a bare `except Exception` with `return False` would be more defensive.

**`src/aleph/vm/agent/vm/backup.py`** (line 402): `_swap_rootfs` checks `info.status` via `get_vm` before calling `stop_vm`, but there's a TOCTOU window between the status check and the stop. This matches the old code's pattern and the per-VM lock serializes against other backup/restore operations, but not against external stop/start. Acceptable given the design constraints.

foxpatch-aleph (approved):
This PR cleanly moves the backup archive ownership from the supervisor to the agent (who allocates the disks) while reducing the supervisor's role to guest quiescence (FreezeGuest/ThawGuest). The design is sound: the auto-thaw timeout with generation tagging prevents stale timers from thawing newer freezes; the freeze/thaw ordering correctly thaws in a `finally` block before the non-disk-I/O archive steps; the per-VM lock serializes backup and restore. The test coverage is comprehensive across both Python (agent BackupManager, supervisor quiescence, archive format, operator views) and Rust (quiesce module with fake QGA server). The proto changes properly reserve the old error code and message numbers. Documentation is thorough. A few minor nits: a stale docstring reference to 'the supervisor' in `_backup_metadata_response`, misleading variable naming in the operator view tests (`fake_sup` used for what is actually `fake_backups`), and synchronous filesystem I/O in `list_backups`/`get_backup_status`/`delete_backup` that runs on the event loop (a design trade-off, not a bug). None of these are blocking.

**`src/aleph/vm/agent/views/operator.py`** (line 775): Stale docstring: "Sourced entirely from the BackupInfo the supervisor returns" — the BackupInfo now comes from the agent's BackupManager (`app["backups"]`), not the supervisor.

**`tests/supervisor/views/test_operator.py`** (line 2489): Misleading variable name: `_seed_authorized_backup_app` returns `fake_backups` (the BackupManager mock), but the test names it `fake_sup`. This makes assertions like `fake_sup.start_backup` read as if they're testing the supervisor. Consider renaming to `fake_backups` for clarity.

**`src/aleph/vm/agent/vm/backup.py`** (line 308): Synchronous filesystem I/O: `list_backups`, `get_backup_status`, and `delete_backup` do blocking I/O (glob, stat, tarfile reads) on the event loop. The old supervisor path ran these on the supervisor's thread pool via gRPC. For a backup directory with many archives, `list_backups` could stall the event loop. Not a regression in correctness, but worth noting for production scaling.

**`src/aleph/vm/backup/archive.py`** (line 260): Minor edge case: if the `.tar.sha256` sidecar exists but is empty, `sidecar.read_text().split()[0]` raises `IndexError`. This is not caught by the `except (tarfile.TarError, OSError)` in `_info_from_tar`. Very unlikely in practice (the sidecar is always written with content), but could be made more defensive.

PR #1152, opened 2026-08-25
fix(supervisor): report FAILED for a controller unit that died

odesenfans:
`VmStatus.FAILED` is declared in the supervisor's status vocabulary but nothing ever produces it. `_status_of` derives the status from a bool ("is the controller unit active"), which collapses systemd's `ActiveState`: a unit that *failed* and one that *never started* are both "not active", so a crashed VM reports `BOOTING` forever.

This is not only a cosmetic status. `start_persistent_vm` treats `BOOTING` as "already starting" and waits in `_wait_until_running` for a VM that will never run, so its `FAILED` branch (delete + recreate, keeping the port mappings) is unreachable. The routing test covering that branch has been passing against a state production could not reach.

Prerequisite for the async-allocations work (2.1): once allocation is asynchronous, "my VM died" has to be answerable from the CRN's advertised status, not from an HTTP status code. Landing it separately because every consumer of `VmInfo` benefits, allocations or not.

## Changes

- `systemd.py`: `get_services_states()` returns the raw `ActiveState` per service from the same single `ListUnits()` call, with `"not-loaded"` for units systemd does not list. `get_services_active_states()` becomes its boolean reduction, semantics unchanged for its callers (`pool.load_persistent_executions` still uses it).
- `supervisor/local.py`: `_controller_state` / `_controller_states` replace `_is_running` / `_running_states` as the carrier. Non-persistent executions map their timestamp pair onto the same vocabulary, so there is one thing to reason about instead of a bool here and a string there. `_is_running` remains as the boolean helper for callers that only ask "is it up".
- `_status_of` maps `"failed"` to `VmStatus.FAILED`, ranked below RUNNING and above BOOTING. `stopped_at` and `stopping_at` keep precedence: a unit that fails while stopping is stopping, not crashed.
- `VmInfo.status_message`, hardcoded to `""` until now, names the failed unit. Only FAILED carries a message; on any other status it would be noise.
- Test doubles updated to mirror the real manager (state string is the source of truth, boolean derived).

Behaviour to expect: a persistent VM whose controller died after `Restart=on-failure` gave up now reports FAILED, and the next allocation push recreates it instead of waiting on it. systemd only reports `failed` once restarts are exhausted, so the state means the VM is not coming back on its own.

## How to test

`pytest tests/supervisor/test_vm_status_failed.py tests/supervisor/test_systemd_states.py`

16 new tests, written before the implementation. The batched-path test was verified to fail against `dev` with `assert [VmStatus.BOOTING] == [VmStatus.FAILED]`, which is the bug in one line.

On a real CRN: `systemctl kill --signal=SIGKILL aleph-vm-controller@<hash>.service` until systemd gives up, then `GET /v2/about/executions/list` and check the VM is no longer reported as running-but-booting.

## Notes

- The full `tests/supervisor` run has ~105 pre-existing failures in my environment (`PermissionError: /var/cache/aleph`). The FAILED set is byte-identical before and after this branch.
- No new mypy errors (the 5 in `agent/pubsub.py` are pre-existing).
- Part 1 of 2 prerequisites for the async scheduler allocations design; the other is #1153 (create-path disk admission). They are independent.

### DIFF
--- a/src/aleph/vm/supervisor/local.py
+++ b/src/aleph/vm/supervisor/local.py
@@ -69,6 +69,11 @@ if TYPE_CHECKING:
 
 logger = logging.getLogger(__name__)
 
+# systemd ActiveState values this module reasons about. The full vocabulary is
+# documented on SystemDManager.get_service_active_state.
+ACTIVE = "active"
+FAILED = "failed"
+
 
 def _backend_of(execution) -> Backend:
     """The VMM only; confidential computing is reported via confidential_mode."""
@@ -77,30 +82,46 @@ def _backend_of(execution) -> Backend:
     return Backend.QEMU
 
 
-def _is_running(execution, pool) -> bool:
+def _controller_state(execution, pool) -> str:
+    """The controller unit's systemd ActiveState.
+
+    Non-persistent executions have no unit, so their liveness (the timestamp
+    pair) is reduced to the same vocabulary: every caller then reasons about
+    one thing instead of a bool here and a string there.
+    """
     if execution.persistent and getattr(execution, "systemd_manager", None) and getattr(pool, "systemd_manager", None):
-        states = pool.systemd_manager.get_services_active_states([execution.controller_service])
-        return states.get(execution.controller_service, False)
+        states = pool.systemd_manager.get_services_states([execution.controller_service])
+        return states.get(execution.controller_service, "unknown")
     times = execution.times
-    return bool(times.starting_at and not times.stopping_at)
+    return ACTIVE if (times.starting_at and not times.stopping_at) else "inactive"
 
 
-def _status_of(execution, running: bool) -> VmStatus:
+def _is_running(execution, pool) -> bool:
+    return _controller_state(execution, pool) == ACTIVE
+
+
+def _status_of(execution, controller_state: str) -> VmStatus:
     times = execution.times
     if times.stopped_at:
         return VmStatus.STOPPED
     if times.stopping_at:
         return VmStatus.STOPPING
-    if running:
+    if controller_state == ACTIVE:
         return VmStatus.RUNNING
+    if controller_state == FAILED:
+        # systemd only reports "failed" once Restart=on-failure has given up,
+        # so the VM is not coming back on its own. Reporting BOOTING here (what
+        # the timestamps alone say) makes a crashed VM look like a starting one,
+        # and the caller waits for a VM that will never run.
+        return VmStatus.FAILED
     if times.starting_at:
         return VmStatus.BOOTING
     return VmStatus.DEFINED
 
 
-def _uptime_secs(execution, running: bool) -> int:
+def _uptime_secs(execution, controller_state: str) -> int:
     started = execution.times.started_at
-    if running and started:
+    if controller_state == ACTIVE and started:
         return int((datetime.now(tz=timezone.utc) - started).total_seconds())
     return 0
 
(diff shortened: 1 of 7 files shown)


foxpatch-aleph (approved):
Well-reasoned, correctly implemented fix that addresses a real bug: a crashed persistent VM reported BOOTING forever because the boolean "is active" collapsed systemd's failed state with never-started. The new _controller_state carries the raw ActiveState string, _status_of maps "failed" to VmStatus.FAILED with correct precedence (stopped/stopping take priority), and the batched path (_controller_states) is consistent with the per-VM path. The boolean API (get_services_active_states) is preserved as a thin reduction so existing callers in pool.py are unaffected. The stop_vm path passing "inactive" is harmless because times.stopped_at is already set. Tests are thorough (16 new, covering precedence, batched listing, unchanged states, and the systemd layer) and the test doubles correctly mirror the real manager. One stale comment reference to _running_states at line 482 is a cosmetic nit, not a blocker.

**`src/aleph/vm/supervisor/local.py`** (line 482): Stale comment: still references `_running_states`, which was renamed to `_controller_states` in this PR. Should read `_controller_states issues a single batched D-Bus ListUnits() call`.

odesenfans:
Fixed in ee34d858, and the same problem was one line up in `get_vm`: its comment still said "`_is_running` is a synchronous D-Bus round-trip" while the call below it now goes through `_controller_state`. Both comments now name the helper they sit on top of.

(`_is_running` does still exist, as the boolean helper for the four callers that only ask "is it up", so that one was misleading rather than dangling.)

foxpatch-aleph (approved):
Well-executed fix for a real bug: VmStatus.FAILED was declared but never produced, causing crashed persistent VMs to report BOOTING forever and making start_persistent_vm's recreate branch unreachable. The refactoring from bool to systemd ActiveState string is clean, the precedence chain in _status_of is correct (stopped/stopping take priority over failed), the boolean API is preserved for existing callers, and all _to_vm_info call sites are updated. 16 new tests cover the new behavior including precedence, status messages, the batched path, and the boolean reduction. No security concerns. The existing test_start_persistent_recreates_after_failed validates the end-to-end consumer.

**`tests/supervisor/test_supervisor_inprocess_query.py`** (line 66): Minor observation (not blocking): when `active={"svc": False}`, the service is silently dropped from `_states` and becomes indistinguishable from a never-mentioned service. This is functionally correct today (all callers pass `True`), but if someone later needs to test a `"failed"` state via the `active` dict API, they would need to use the `states=` parameter instead. Consider documenting this or defaulting `False` to `"inactive"` explicitly.

foxpatch-aleph (approved):
Well-reasoned, correctly implemented fix for a real bug: a crashed systemd controller unit (ActiveState=failed) was indistinguishable from a starting one because the old boolean reduction collapsed both to 'not active'. The PR preserves the raw ActiveState string from ListUnits() and maps 'failed' to VmStatus.FAILED with correct precedence (stopped/stopping take priority over failed, failed over booting). The backward-compatible get_services_active_states delegates to the new get_services_states, so pool.py's reattach path is unaffected. Tests are comprehensive and cover both the per-VM and batched code paths. The only observation is that test_views.py mocks the old get_services_active_states API while the code now calls get_services_states, but this is a pre-existing test hygiene issue (those tests are among the ~105 pre-existing failures) and the tests pass by accident via Mock's default behavior, not by design.

**`tests/supervisor/test_views.py`** (line 468): Pre-existing issue (not introduced by this PR): these tests mock `get_services_active_states` on the systemd_manager Mock, but `list_vms` now calls `get_services_states`. The tests pass because a Mock's `get_services_states` returns a Mock, and `Mock == "active"` and `Mock == "failed"` are both False, so the timestamp fallback applies — same net behavior as before. Consider adding `get_services_states=mocker.Mock(return_value={})` to these mocks for correctness, though this is non-blocking since these tests are in the pre-existing failure set.

PR #1153, opened 2026-08-25
fix(agent): admit VMs against disk before downloading their resources

odesenfans:
Every create path calls `check_capacity` with `disk_mib=0` (`agent/run.py`, three call sites), so nothing on the create side ever judges disk: only the reserve endpoint does, through `requirements_from_message`. A CRN with a full disk keeps accepting instances and fails later, on write. The per-pool "largest single volume" check (`max_volume_mib`) is skipped for the same reason.

Prerequisite for the async-allocations work (2.1), where the CRN answers "would you accept these VMs?" before downloading anything: an advisory answer that ignores disk would be strictly weaker than what the create path enforces, and a plan-time rejection could be followed by a `notify_allocation` acceptance for the same VM. Landing it separately because the gap exists today, independently of allocations.

## Changes

Filling the number in at the existing call site would have been theatre: that call sits **after** `build_*_spec`, which downloads the resources and creates the volume files (`qemu-img`, `fallocate`). By then the space is already allocated, so requiring it again would double-count it.

- New shared `_admit()` in `agent/run.py`, called **before** the spec build on all three create paths (program, instance including SNP, v-program). It is built on `requirements_from_message`, the same helper the reserve endpoint uses, so advisory and enforced admission now agree. Refusing there also means a host with no room never pays for the download.
- The existing post-build `check_capacity` calls stay as they are, at `disk_mib=0`: they re-check memory and vCPUs against anything that changed during the download, and they remain the point where GPU holds are resolved and consumed. That must stay after the build so a failed download never burns the owner's hold, which is what the current comment describes.
- `_admit` takes `is_instance` explicitly rather than reading it off the message. A V-PROGRAM belongs in the instance memory bucket (as its pre-existing post-build check already said), but it is not an `InstanceContent`, so `requirements_from_message` would have put it in the much smaller program bucket and refused it. There is a test pinning this.

Behaviour to expect: a create whose declared rootfs plus volumes exceed the free space, or whose largest single volume exceeds the roomiest pool, is refused up front with `InsufficientResourcesError` instead of proceeding and failing on write. Sizes stay best-effort for immutable volumes (they carry no `size_mib`), which the existing code already acknowledges.

## How to test

`pytest tests/supervisor/test_create_path_disk_admission.py`

5 new tests, written before the implementation: each path refuses before its build is awaited, the instance path passes the declared 10000 MiB rootfs as both total and largest-volume, and the v-program path is admitted against the instance bucket.

## Notes

- Deliberately does **not** move the memory/vCPU check or the GPU resolution: only the disk-bearing gate is added ahead of the download.
- The full `tests/supervisor` run has ~105 pre-existing failures in my environment (`PermissionError: /var/cache/aleph`). The FAILED set is byte-identical before and after this branch. No new mypy errors.
- Part 2 of 2 prerequisites for the async scheduler allocations design; the other is #1152. They are independent and can merge in either order.

### DIFF
--- a/src/aleph/vm/agent/run.py
+++ b/src/aleph/vm/agent/run.py
@@ -14,6 +14,7 @@ from aiohttp.web_exceptions import (
     HTTPServiceUnavailable,
 )
 from aleph_message.models import (
+    ExecutableContent,
     InstanceContent,
     ItemHash,
     ProgramContent,
@@ -23,7 +24,11 @@ from msgpack import UnpackValueError
 from multidict import CIMultiDict
 
 from aleph.vm.agent.aggregate import get_user_settings
-from aleph.vm.agent.capacity import CapacityManager, requested_gpu_ids
+from aleph.vm.agent.capacity import (
+    CapacityManager,
+    requested_gpu_ids,
+    requirements_from_message,
+)
 from aleph.vm.agent.expiry import ExpiryManager
 from aleph.vm.agent.snp_instance_launch import (
     build_snp_instance_spec,
@@ -329,6 +334,29 @@ async def _wait_until_gone(
         await asyncio.sleep(interval)
 
 
+def _admit(capacity: CapacityManager, content: ExecutableContent, vm_hash: ItemHash, *, is_instance: bool) -> None:
+    """Agent-side admission, ahead of any download.
+
+    Disk is judged here and only here: build_*_spec downloads the resources and
+    creates the volume files, so a disk check after it would measure space this
+    VM has already taken. Refusing here also means a host with no room never
+    pays for the download.
+
+    ``is_instance`` is passed explicitly rather than taken from the message:
+    a V-PROGRAM is an SNP VM and belongs in the instance memory bucket, but it
+    is not an InstanceContent, which is all requirements_from_message can see.
+    """
+    requirements = requirements_from_message(content)
+    capacity.check_capacity(
+        memory_mib=requirements.memory_mib,
+        vcpus=requirements.vcpus,
+        disk_mib=requirements.disk_mib,
+        max_volume_mib=requirements.max_volume_mib,
+        is_instance=is_instance,
+        exclude_vm_hash=vm_hash,
+    )
+
+
 async def create_vm_execution(
     vm_hash: ItemHash,
     *,
@@ -360,6 +388,7 @@ async def create_vm_execution(
         # on the first request through _ensure_program_vm. On-demand programs
         # are created and configured per request there too, so this branch only
         # does eager work for the persistent (scheduled) case.
+        _admit(capacity, content, vm_hash, is_instance=False)
         spec, _resources = await build_program_create_vm_spec(vm_hash, content)
         capacity.check_capacity(
             memory_mib=content.resources.memory,
@@ -400,6 +429,7 @@ async def create_vm_execution(
         snp_instance = is_snp_instance(content)
         attest_port: int | None = None
         try:
+            _admit(capacity, content, vm_hash, is_instance=True)
             if snp_instance:
                 # SEV-SNP confidential instances build through the dedicated
                 # LUKS-rootfs SNP launch path, not build_create_vm_spec (which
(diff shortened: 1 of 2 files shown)


foxpatch-aleph (approved):
The PR correctly adds a pre-download disk admission gate via `_admit()`, built on the shared `requirements_from_message` helper so advisory and enforced admission agree. The explicit `is_instance` parameter correctly handles the v-program case (SNP VM in the instance memory bucket despite not being `InstanceContent`). Error handling is sound: early registry records are cleaned up via `registry.forget` in the except blocks on the instance and v-program paths, and the program path has no record to clean up at that point. The post-build `check_capacity` calls correctly remain at `disk_mib=0` to re-check memory/vCPUs and serve as the GPU resolution point, without double-counting disk. Tests cover all three create paths, verify refusal happens before the download, assert disk values are propagated, and pin the v-program instance-bucket behavior. The type annotation `ExecutableContent` on `_admit` doesn't include `VerifiableProgramContent`, but this matches the existing convention in `requirements_from_message` and `requested_gpu_ids`.

**`tests/supervisor/test_create_path_disk_admission.py`** (line 84): Consider adding a success-path test (e.g., `capacity=capacity()` without `refuse=True`) that asserts two `check_capacity` calls happen: the pre-admit with the real `disk_mib` and the post-build with `disk_mib=0`. This would pin the two-check contract and catch any future regression that collapses them or changes the post-build disk value.

**`src/aleph/vm/agent/run.py`** (line 337): The type annotation `content: ExecutableContent` doesn't include `VerifiableProgramContent`, which is passed on the v-program path (line 531). This matches the existing pattern in `requirements_from_message` and `requested_gpu_ids` (both annotated `ExecutableContent` but called with `VerifiableProgramContent`), so it's a pre-existing convention rather than a new issue. If the `ExecutableContent` union is ever widened to include `VerifiableProgramContent` in the aleph-message pin, all three would become correct simultaneously.

odesenfans:
**On the success-path test:** added in 19974f57. The gap was real and it was the one that mattered: every test took the refusal path, so nothing pinned the post-build call staying at `disk_mib=0`, and a later change that "completed" the fix by passing the real requirement there too would have reinstated the double-counting with the suite still green. The new test asserts both calls on a successful instance create, and I verified it by mutation in both directions: making the post-build call require disk again fails on the second tuple, dropping disk from the gate fails on the first.

**On the `ExecutableContent` annotation:** the premise does not hold on the pinned aleph-message (`4ad261ff`), where the union already includes the v-program type:

```python
ExecutableContent = InstanceContent | ProgramContent | VerifiableProgramContent
```

`VerifiableProgramContent` is not a `ProgramContent` subclass (it derives from `BaseExecutableContent`), so it is a distinct member of that union rather than something the annotation covers by inheritance. The annotation on `_admit` is therefore correct as written, which is also why mypy is clean on the v-program call site at line 531. The suggested future widening has already happened; no change needed here or in `requirements_from_message` / `requested_gpu_ids`.

foxpatch-aleph (approved):
Well-reasoned change that fills a real gap: the create paths were passing disk_mib=0 to check_capacity, so a full-disk CRN kept accepting VMs and failing on write. The new _admit helper runs before the download on all three create paths, uses the same requirements_from_message as the reserve endpoint (so advisory and enforced admission agree), and the post-build check_capacity stays at disk_mib=0 as the memory/vCPU re-check and GPU resolution point. The is_instance override for V-PROGRAMs is correct (they belong in the instance memory bucket despite not being InstanceContent) and is pinned by a test. Registry records and staging are properly cleaned up when pre-build admission fails. The test coverage is thorough: all three paths are tested for pre-build refusal, the instance path verifies the declared disk is carried, the v-program path verifies the instance bucket, and the happy path pins the two-check contract (disk once before, never after).

**`tests/supervisor/test_vprogram.py`** (line 337): The docstring says "Capacity admission runs after the bundle is staged" — with this PR, _admit also runs *before* the bundle fetch. The test still passes functionally (check_capacity always raises, so the pre-build _admit fires first, build_vprogram_spec is never reached, create_vm is never reached, and the record is forgotten), but the docstring is now slightly stale. Consider updating it to acknowledge both admission gates.

PR #1154, opened 2026-08-25
proto: drop the reservations left by the removed wipe flag and backup error code

odesenfans:
Follow-up to #1151. `ERROR_CODE_BACKUP_NOT_FOUND = 12` was reserved when the backup RPCs left the wire, but no release ever shipped that number, so there is no reader to protect: the reservation is dropped and 12 is free for the next error code. Bindings regenerated; no other reference remained (Python `BackupNotFoundError` is agent-local now).

https://claude.ai/code/session_01SUXZP8HPU2TGCcaemKznnv


foxpatch-aleph (approved):
Clean removal of an unused proto reservation. The proto, generated Python bindings, and .pyi stub are all consistent. No remaining references to ERROR_CODE_BACKUP_NOT_FOUND exist in the codebase, and the agent-local BackupNotFoundError exception is correctly unaffected. Existing tests cover the enum surface and will continue to pass.

foxpatch-aleph (approved):
The PR correctly drops two proto reservations that were never shipped in a release: `reserved 2`/`reserved "wipe"` in DeleteVmRequest (renumbering `keep_port_mappings` from 3 to 2) and `reserved 12`/`reserved "ERROR_CODE_BACKUP_NOT_FOUND"` in ErrorCode. The generated Python bindings (supervisor_pb2.py and .pyi) are consistent with the .proto definition, the Rust crate compiles from the .proto at build time (no checked-in file to update), and all callers across Python and Rust use the named field `keep_port_mappings` so the renumbering is wire-safe. Docs and test docstrings are updated correctly. The relevant proto binding tests pass. No remaining references to `wipe` (as a proto field) or `ERROR_CODE_BACKUP_NOT_FOUND` exist in the codebase; `BackupNotFoundError` is correctly agent-local (Python exception only).

**`tests/supervisor/test_proto_bindings.py`** (line 294): Minor nit: since this PR's core change is renumbering `keep_port_mappings` from field 3 to field 2, consider asserting the field number here — e.g. `assert req.DESCRIPTOR.fields_by_number[2].name == "keep_port_mappings"`. The existing test verifies the field name and behavior but not the wire number, so a future accidental re-add of `reserved 2` would not be caught by the test suite.

PR #1155, opened 2026-08-25
feat(agent): disk reclamation, retire_vm and volume retention (2.1, PR A)

odesenfans:
## Summary

PR A of the 2.1 disk reclamation work. The design and implementation plan are not part of the repository; `docs/architecture/storage.md` and `vm-lifecycle.md` (included) are the reference for what the code does and why.

- **One deletion function with a mandatory reason.** `retire_vm(vm_hash, RetireReason)` with `RECREATE` / `GONE` / `ERASE` / `FAILED_CREATE`; every agent delete site (`tasks.py`, `expiry.py`, `run.py`, `views/__init__.py`, `update_watcher.py`, `operator.py`, `migration.py`) converted; a test pins that no agent module except `retire.py` calls `supervisor.delete_vm`. A failed re-create of a VM whose volumes already existed retires `RECREATE` (spec amended): owner data is never wiped by a boot timeout or an admission refusal.
- **Retention policy.** `VOLUME_RETENTION = reap | keep` (default `reap`), `VOLUME_RETENTION_BUDGET` (10% per pool). Under `keep`, a GONE VM's directories get a `.reclaimable` marker (`reclaimable_since`, `reason`, `size_bytes`, `depends_on`, `owner`); reclaimable space counts as free in all three disk figures; a create adopts a retained directory of the same hash; `ERASE` always purges and works from the marker's owner after the record is gone.
- **Reconciler.** `reconcile_storage()` at startup, every `VOLUME_RECONCILE_INTERVAL`, after every GONE and on placement pressure (`storage_pools.set_room_maker`, evicts oldest-first until the create fits, never a live hash). Passes: orphan namespaces (outside `VOLUME_CREATE_GUARD` and the `creating()` guard), `.part` files, side dirs (session, staging, empty `/mnt` mount points via `rmdir` only), retention budget, backup TTL sweep. Live set is the registry union `supervisor.list_vms()`; the startup pass refuses to purge if the supervisor cannot be asked or the registry rehydrated empty while VMs are running. Startup logs a per-pool summary before purging.
- **Transactional creates:** every create path (program, instance, V-PROGRAM, migration import) runs inside `creating()`; the migration import now records the imported VM in the registry.
- Settings: `VOLUME_RETENTION`, `VOLUME_RETENTION_BUDGET`, `VOLUME_RECONCILE_INTERVAL`, `VOLUME_CREATE_GUARD`, `CACHE_BUDGET` (consumed by PR D), `MAX_RUNTIME_ARCHIVE_SIZE` (PR D). Docs: `docs/architecture/storage.md`, `vm-lifecycle.md`.

Follow-up PRs (stacked): B admission from the message, C device-mapper teardown, D cache bounds + in-stream size cap, E `aleph-vm storage` CLI.

## Behaviour changes (upgrade notes)
- **Payment shortfalls now retire `GONE`.** On 1.x an insufficient held balance, credit balance or stream stopped the VM but kept its record and disks for a later top-up (and nothing ever reclaimed them). Now such a VM is retired like a forgotten one: record and ports dropped, volumes purged under the default `VOLUME_RETENTION=reap`, marked reclaimable (with the owner, adopted back on re-create) under `keep`. Operators who want a grace period for re-paying owners set `keep`.
- The first start on an existing node reclaims every leaked directory under `reap`; the log carries a per-pool count and byte total before it does.

## Known limitations (deliberate)
- Orphan markers written by the reconciler carry no owner, so such data can only leave via the budget.
- Erase from a marker cannot honour delegation (the security aggregate needs the message).
- The periodic pass runs with a registry-only live set when the supervisor is unreachable (startup refuses; periodic does not).
- `GONE` under `keep` queues one serialized pass per VM; not coalesced.

## Test plan
- `tests/supervisor`: 1320 passed, 48 pre-existing environment failures (identical set to `dev`).
- New: `test_storage_budget.py`, `test_reclaimable.py`, `test_retire.py`, `test_reconciler.py`, `test_agent_no_direct_delete.py`; extended capacity, resources, operator, run, migration tests.

### DIFF
--- a/docs/architecture/storage.md
+++ b/docs/architecture/storage.md
@@ -122,9 +122,13 @@ agent registered at startup (`storage_pools.set_room_maker`, wired to
 `reconciler.make_room`) so the retained directories on the target pool are
 evicted oldest-first until the create fits. See "Reclamation" below: a
 retained disk is a cache entry, not usage, so counting it as used would sell
-less capacity than the node actually has. `about_system_usage` is deliberately
-left on the raw `pools_disk_usage()` figures, since it reports what the
-filesystem currently holds, not what a create could get.
+less capacity than the node actually has. The rule holds for all three disk
+figures the agent produces, and they have to agree: the aggregate
+(`_available_disk_bytes`), the largest-single-volume check (`_check_max_volume`,
+which adds each pool's own `reclaimable_bytes(pool.path)` to that pool's free
+space) and the advertised capacity (`about_system_usage` /
+`_disk_usage_from_pools`). A node that advertises less than admission accepts
+is a node the scheduler stops sending work to.
 
 The Rust supervisor daemon does none of this pool selection or admission
 work. It parses `VOLUME_POOLS` (`config.rs`) with the same validation rules
@@ -345,11 +349,23 @@ keyed by a hash validated by `purge._checked_namespace`.
 
 Passes run at startup, every `VOLUME_RECONCILE_INTERVAL` (jittered, so a
 fleet of nodes does not sweep in lockstep), after every `GONE` under `keep`
-(`retire.set_after_gone_hook`), and on admission pressure
-(`make_room`, through `storage_pools.set_room_maker`). A pass runs in a
-worker thread, so the live-VM set is snapshotted on the event loop first
-(`live_hashes`) and handed over; `make_room` never evicts a hash in it, even
-if a stale marker says it could.
+(`retire.set_after_gone_hook`, best effort: a failing pass never breaks the
+sweep that retired the VM), and on admission pressure (`make_room`, through
+`storage_pools.set_room_maker`). A pass runs in a worker thread, so the
+live-VM set is snapshotted on the event loop first (`live_hashes`) and handed
+over; `make_room` never evicts a hash in it, even if a stale marker says it
+could. Loop-triggered passes serialize on a lock (`reconcile_now`), and
+placement runs `make_room` through `asyncio.to_thread`, so it can still
+overlap a pass: every removal therefore tolerates a directory that another
+pass took first, and counts only what it actually removed.
+
+The create guard covers more than the create paths in `run.py`: a migration
+import stages multi-GB disks into `{pool}/{vm_hash}/` for a namespace the
+registry knows nothing about yet, so `run_import` holds `creating()` across
+the whole transfer and records the imported VM in the registry (and the agent
+DB) once `CreateVm` returns. A namespace under the guard is skipped whole,
+`.part` files included: the directory mtime does not advance while a file
+inside it grows, so age is no evidence that a transfer has stopped.
 
 ### Settings
 
@@ -398,9 +414,11 @@ pass against an empty registry would call every running VM an orphan.
 - Disk admission checks aggregate free space and, separately, that the
   single largest requested volume fits the roomiest eligible pool, since no
   pool splits a volume across disks (`src/aleph/vm/agent/capacity.py`).
-- Reclaimable bytes are advertised as free, not as used, and a placement
-  that does not fit evicts them (oldest first) before it is refused
-  (`src/aleph/vm/agent/capacity.py`, `src/aleph/vm/storage_pools.py`).
+- Reclaimable bytes are advertised as free, not as used, in every disk figure
+  the agent reports (aggregate, largest single volume, advertised capacity),
+  and a placement that does not fit evicts them (oldest first) before it is
+  refused (`src/aleph/vm/agent/capacity.py`,
+  `src/aleph/vm/agent/resources.py`, `src/aleph/vm/storage_pools.py`).
 - Every agent-side deletion goes through `retire_vm` with an explicit
   reason; nothing else under `src/aleph/vm/agent/` calls
   `supervisor.delete_vm` (`src/aleph/vm/agent/vm/retire.py`).
--- a/docs/architecture/storage.md
+++ b/docs/architecture/storage.md
@@ -320,18 +320,27 @@ the volumes are reused through the existing sticky placement in
 ### The reconciler
 
 `reconcile_storage()` (`src/aleph/vm/agent/vm/reconciler.py`) treats the
-filesystem as the truth and the agent's registry as the list of owners. One
-pass does, in order:
+filesystem as the truth and asks two sources who the owners are: the agent
+registry, and the VMs the supervisor lists. The union is the live set. The
+registry alone is not enough, because it is refilled at boot from the agent
+DB and that rehydration skips any record with an empty or unparseable
+message. One pass does, in order:
 
-1. Walk `{pool}/*/`. Each directory is live (a registry record exists),
-   reclaimable (a marker is present) or an orphan (neither). Orphans are
-   purged under `reap` and marked `reason: orphan` under `keep`.
+1. Walk `{pool}/*/`. Each directory is live (the registry or the supervisor
+   knows its hash), reclaimable (a marker is present) or an orphan
+   (neither). Orphans are purged under `reap` and marked `reason: orphan`
+   under `keep`. Liveness is asked again immediately before anything is
+   marked, purged or evicted, so a create that commits while the pass walks
+   is not mistaken for an orphan.
 2. Delete `*.part` files older than the guard, in the four download caches
    and in the pools (the downloader streams volumes in place).
 3. Sweep the side directories keyed by VM hash: the confidential session
    directory, the SNP and V-PROGRAM staging directories, and
    `/mnt/{namespace}_{volume}` mount points. A directory that is still a
-   mount point is logged and left alone.
+   mount point is logged and left alone, and under `/mnt` only empty
+   mount-point directories are removed (with `rmdir`, never `rmtree`): `/mnt`
+   is the operator's directory, and an unmounted agent mount point is empty
+   by construction.
 4. Enforce the retention budget per pool: sum the markers' `size_bytes` and
    evict oldest-first (by `reclaimable_since`) until under
    `VOLUME_RETENTION_BUDGET`. Under `reap` everything marked is given back,
@@ -347,6 +356,12 @@ somehow escaped the context manager is still protected by the second.
 Everything a pass touches is under a directory the agent created and is
 keyed by a hash validated by `purge._checked_namespace`.
 
+The startup pass is stricter than the rest: it refuses to purge anything
+(it runs dry and logs why) when the supervisor did not answer `list_vms`,
+or when rehydration left the registry empty while the supervisor still runs
+at least one VM. Both states mean the live set cannot be trusted, and the
+first pass on a node is the one acting on the biggest backlog.
+
 Passes run at startup, every `VOLUME_RECONCILE_INTERVAL` (jittered, so a
 fleet of nodes does not sweep in lockstep), after every `GONE` under `keep`
 (`retire.set_after_gone_hook`, best effort: a failing pass never breaks the
@@ -354,7 +369,9 @@ sweep that retired the VM), and on admission pressure (`make_room`, through
 `storage_pools.set_room_maker`). A pass runs in a worker thread, so the
 live-VM set is snapshotted on the event loop first (`live_hashes`) and handed
 over; `make_room` never evicts a hash in it, even if a stale marker says it
-could. Loop-triggered passes serialize on a lock (`reconcile_now`), and
+could. Loop-triggered passes serialize on a lock (`reconcile_now`); they are
+serialized, not coalesced, so a sweep that retires N VMs `GONE` queues N
+passes. And
 placement runs `make_room` through `asyncio.to_thread`, so it can still
 overlap a pass: every removal therefore tolerates a directory that another
 pass took first, and counts only what it actually removed.
@@ -422,8 +439,9 @@ pass against an empty registry would call every running VM an orphan.
 - Every agent-side deletion goes through `retire_vm` with an explicit
   reason; nothing else under `src/aleph/vm/agent/` calls
   `supervisor.delete_vm` (`src/aleph/vm/agent/vm/retire.py`).
-- The reconciler never removes a directory whose hash is live, is in the
-  in-process creating set, or whose mtime is inside `VOLUME_CREATE_GUARD`
+- The reconciler never removes a directory whose hash is live in the agent
+  registry or listed by the supervisor, is in the in-process creating set,
+  or whose mtime is inside `VOLUME_CREATE_GUARD`
   (`src/aleph/vm/agent/vm/reconciler.py`).
 - The Rust supervisor daemon never selects, adopts, or writes to a pool; it
   only re-validates `VOLUME_POOLS` and sums free bytes for host-info
--- a/docs/architecture/storage.md
+++ b/docs/architecture/storage.md
@@ -115,6 +115,17 @@ happily admit, say, a 900 GiB volume onto two half-full 500 GiB disks and
 then fail at creation time. Either check failing raises
 `InsufficientResourcesError`.
 
+Admission counts retained (reclaimable) bytes as free:
+`_available_disk_bytes` adds `reclaimable_bytes()` to the pooled free sum,
+and `select_pool`, before refusing a placement, calls the room maker the
+agent registered at startup (`storage_pools.set_room_maker`, wired to
+`reconciler.make_room`) so the retained directories on the target pool are
+evicted oldest-first until the create fits. See "Reclamation" below: a
+retained disk is a cache entry, not usage, so counting it as used would sell
+less capacity than the node actually has. `about_system_usage` is deliberately
+left on the raw `pools_disk_usage()` figures, since it reports what the
+filesystem currently holds, not what a create could get.
+
 The Rust supervisor daemon does none of this pool selection or admission
 work. It parses `VOLUME_POOLS` (`config.rs`) with the same validation rules
 purely so a configuration the agent would refuse to boot on also fails
@@ -251,6 +262,122 @@ download) files, since that combination is unambiguous evidence of an
 aborted import; a namespace with only complete `.qcow2` files but no known
 VM is left alone and only logged, since a retried import can still adopt it.
 
+## Reclamation
+
+Every agent path that ends a VM goes through one function,
+`retire_vm(vm_hash, reason, ...)` (`src/aleph/vm/agent/vm/retire.py`), and
+the reason decides what happens to the disks. There is no default: a call
+site has to say what it means, which is exactly what was missing when disks
+leaked.
+
+| Reason | Volumes | Registry, ports | Session, staging, backups |
+|--------|---------|-----------------|---------------------------|
+| `RECREATE` | untouched | kept | untouched |
+| `GONE` | `reap`: purged; `keep`: marked reclaimable | dropped | removed |
+| `ERASE` | purged, whatever the policy | dropped | removed |
+| `FAILED_CREATE` | purged, whatever the policy | dropped | removed |
+
+`RECREATE` is the same VM coming straight back (message amend, crash
+recovery, idle program reap, reboot). `GONE` needs positive knowledge that
+it will not: a forgotten or removed message, a stopped payment, a scheduler
+deallocation, a completed migration hand-off. An API error or a timeout is
+not an answer and retires nothing. `ERASE` ignores `keep` on purpose:
+retention protects against administrative deletion, not against the owner's
+own request, and a `FAILED_CREATE` never ran, so it has nothing worth
+keeping. A create that fails on a VM whose volumes already existed retires
+`RECREATE` instead, since the create paths are also the re-create paths for
+a host-persistent VM.
+
+### The `.reclaimable` marker
+
+A namespace directory with no marker belongs to a live VM. Under
+`VOLUME_RETENTION=keep`, `GONE` drops the registry record and writes
+`{pool}/{vm_hash}/.reclaimable` in its place
+(`src/aleph/vm/agent/vm/reclaimable.py`):
+
+```json
+{
+  "version": 1,
+  "reclaimable_since": "<iso timestamp>",
+  "reason": "gone | orphan",
+  "size_bytes": 12345,
+  "depends_on": ["<parent image ref>", "..."]
+}
+```
+
+One marker per pool the VM spans, so each pool's budget is computed from its
+own tree. `depends_on` names the cache entries (parent images) the volumes
+still need, so the cache pass cannot evict a parent from under a retained
+overlay. Under `reap` no marker is ever written: the purge is immediate.
+A create for the same hash adopts the directory (the marker is unlinked and
+the volumes are reused through the existing sticky placement in
+`volume_path_for`), which is what `reconciler.creating()` does on entry.
+
+### The reconciler
+
+`reconcile_storage()` (`src/aleph/vm/agent/vm/reconciler.py`) treats the
+filesystem as the truth and the agent's registry as the list of owners. One
+pass does, in order:
+
+1. Walk `{pool}/*/`. Each directory is live (a registry record exists),
+   reclaimable (a marker is present) or an orphan (neither). Orphans are
+   purged under `reap` and marked `reason: orphan` under `keep`.
+2. Delete `*.part` files older than the guard, in the four download caches
+   and in the pools (the downloader streams volumes in place).
+3. Sweep the side directories keyed by VM hash: the confidential session
+   directory, the SNP and V-PROGRAM staging directories, and
+   `/mnt/{namespace}_{volume}` mount points. A directory that is still a
+   mount point is logged and left alone.
+4. Enforce the retention budget per pool: sum the markers' `size_bytes` and
+   evict oldest-first (by `reclaimable_since`) until under
+   `VOLUME_RETENTION_BUDGET`. Under `reap` everything marked is given back,
+   which is how a node switched from `keep` to `reap` drains.
+5. Sweep expired backups.
+
+Nothing a create is still building is touched. Two independent guards say
+so: a hash registered in the in-process "creating" set (every create path in
+`run.py` and `_ensure_program_vm` runs inside `reconciler.creating()`), and
+an mtime younger than `VOLUME_CREATE_GUARD`. Both matter, since a create
+that outlives the guard is only protected by the first, and a create that
+somehow escaped the context manager is still protected by the second.
+Everything a pass touches is under a directory the agent created and is
+keyed by a hash validated by `purge._checked_namespace`.
+
+Passes run at startup, every `VOLUME_RECONCILE_INTERVAL` (jittered, so a
+fleet of nodes does not sweep in lockstep), after every `GONE` under `keep`
+(`retire.set_after_gone_hook`), and on admission pressure
+(`make_room`, through `storage_pools.set_room_maker`). A pass runs in a
+worker thread, so the live-VM set is snapshotted on the event loop first
+(`live_hashes`) and handed over; `make_room` never evicts a hash in it, even
+if a stale marker says it could.
+
+### Settings
+
+| Setting | Default | What it does |
+|---------|---------|--------------|
+| `VOLUME_RETENTION` | `reap` | `reap` purges a `GONE` VM's volumes at once; `keep` retains them |
+| `VOLUME_RETENTION_BUDGET` | `10%` | Cap on reclaimable bytes per pool (percentage or absolute, e.g. `50G`) |
+| `VOLUME_RECONCILE_INTERVAL` | `3600` | Seconds between periodic passes |
+| `VOLUME_CREATE_GUARD` | `600` | Seconds a young directory or `.part` file is assumed to be an in-flight create |
+| `CACHE_BUDGET` | `20%` | Cap on each download cache (runtime, code, data, message) |
+
+`keep` is not a promise of "forever": reclaimable disks are unpaid storage,
+and an attacker does not need to stop paying to create them (create, forget,
+repeat). It means "kept as long as the budget and live demand allow, oldest
+goes first", which is the only promise a node can honestly sell.
+
+### First start on an existing node
+
+The first pass on an upgraded node finds every directory leaked by the bugs
+this machinery fixes, as an orphan. Under `reap` that is a one-shot cleanup
+of potentially a lot of data, so the startup hook runs a dry pass first and
+logs what it is about to remove, once as a total and once per pool
+(count and bytes), before the real pass runs. An operator reading the log
+can then explain why free space jumped. Under `keep` those directories
+become reclaimable with `reason: orphan` and fall under the budget instead.
+The startup hook is registered after registry rehydration on purpose: a
+pass against an empty registry would call every running VM an orphan.
+
 ## Key invariants
 
 - `PERSISTENT_VOLUMES_DIR` is always pool index 0 and is always VM-eligible
@@ -271,6 +398,15 @@ VM is left alone and only logged, since a retried import can still adopt it.
 - Disk admission checks aggregate free space and, separately, that the
   single largest requested volume fits the roomiest eligible pool, since no
   pool splits a volume across disks (`src/aleph/vm/agent/capacity.py`).
+- Reclaimable bytes are advertised as free, not as used, and a placement
+  that does not fit evicts them (oldest first) before it is refused
+  (`src/aleph/vm/agent/capacity.py`, `src/aleph/vm/storage_pools.py`).
+- Every agent-side deletion goes through `retire_vm` with an explicit
+  reason; nothing else under `src/aleph/vm/agent/` calls
+  `supervisor.delete_vm` (`src/aleph/vm/agent/vm/retire.py`).
+- The reconciler never removes a directory whose hash is live, is in the
+  in-process creating set, or whose mtime is inside `VOLUME_CREATE_GUARD`
+  (`src/aleph/vm/agent/vm/reconciler.py`).
 - The Rust supervisor daemon never selects, adopts, or writes to a pool; it
   only re-validates `VOLUME_POOLS` and sums free bytes for host-info
   reporting. All placement is agent-side
@@ -303,6 +439,13 @@ VM is left alone and only logged, since a retried import can still adopt it.
   (`_make_writable_volume`) and the `pool0_only` wiring for Firecracker.
 - `src/aleph/vm/agent/capacity.py`: `CapacityManager.check_capacity` and the
   disk-admission checks.
+- `src/aleph/vm/agent/vm/retire.py`: `retire_vm` and `RetireReason`, the one
+  way a VM's storage is released.
+- `src/aleph/vm/agent/vm/reclaimable.py`: the `.reclaimable` marker
+  (`mark_reclaimable`, `adopt`, `iter_reclaimable`, `reclaimable_bytes`).
+- `src/aleph/vm/agent/vm/reconciler.py`: `reconcile_storage`, the create
+  guard (`creating`), the evictor (`make_room`) and the startup/periodic
+  hooks the agent registers in `src/aleph/vm/agent/supervisor.py`.
 - `src/aleph/vm/agent/resources.py`: `about_system_usage`'s pooled disk
   figures.
 - `rust/crates/supervisor-daemon/src/config.rs`: `VOLUME_POOLS`/`MediaClass`
--- a/docs/architecture/storage.md
+++ b/docs/architecture/storage.md
@@ -305,14 +305,21 @@ A namespace directory with no marker belongs to a live VM. Under
   "reclaimable_since": "<iso timestamp>",
   "reason": "gone | orphan",
   "size_bytes": 12345,
-  "depends_on": ["<parent image ref>", "..."]
+  "depends_on": ["<parent image ref>", "..."],
+  "owner": "<owner address>"
 }
 ```
 
 One marker per pool the VM spans, so each pool's budget is computed from its
 own tree. `depends_on` names the cache entries (parent images) the volumes
 still need, so the cache pass cannot evict a parent from under a retained
-overlay. Under `reap` no marker is ever written: the purge is immediate.
+overlay. `owner` is the address from the VM's message, copied in at `GONE`
+because the marker outlives every other record of it (the registry record is
+forgotten and the DB rows are deleted), and it is what lets the node
+authorize the owner's own erase of retained data; it is optional, so markers
+written before the field and orphan markers for a VM whose message the node
+never held simply have none. Under `reap` no marker is ever written: the
+purge is immediate.
 A create for the same hash adopts the directory (the marker is unlinked and
 the volumes are reused through the existing sticky placement in
 `volume_path_for`), which is what `reconciler.creating()` does on entry.
--- a/docs/architecture/storage.md
+++ b/docs/architecture/storage.md
@@ -292,6 +292,15 @@ keeping. A create that fails on a VM whose volumes already existed retires
 `RECREATE` instead, since the create paths are also the re-create paths for
 a host-persistent VM.
 
+A payment shortfall is a stopped payment: when the held balance, the credit
+balance or the stream no longer covers the VMs it pays for, the payment
+monitor retires them `GONE`, so under the default `reap` their volumes are
+purged at once. This is a change from 1.x, which stopped such a VM but kept
+its record and disks for a later top-up, and then never reclaimed them. A
+node that wants to give a re-paying owner a grace period runs
+`VOLUME_RETENTION=keep`: the marker carries the owner, and a re-created VM
+adopts its retained directory untouched.
+
 ### The `.reclaimable` marker
 
 A namespace directory with no marker belongs to a live VM. Under
--- a/docs/architecture/vm-lifecycle.md
+++ b/docs/architecture/vm-lifecycle.md
@@ -355,13 +355,40 @@ shared by every program on that runtime), and a spec-only view misses the
 namespace directory itself, volumes on other pools and the staging
 directories. `DeleteVm` therefore takes no erase flag at all.
 
-Deletion lives in `src/aleph/vm/agent/vm/purge.py`, keyed by `vm_hash` and
-scoped to directories the agent created, so a shared cache entry is not
-merely spared by a check but unreachable: `purge_vm_volumes` deletes the
-files under `{pool}/{vm_hash}/` on every pool (the reinstall path);
-`purge_vm_storage` removes those directories, the confidential session
-directory and the SNP/V-PROGRAM staging directory (`operate_erase`, the
-owner's "delete all my data", which erases the rootfs too). A `.btrfs`
+Every agent path that ends a VM goes through one function,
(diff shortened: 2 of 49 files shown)


foxpatch-aleph (approved):
A well-architected disk reclamation system with a unified retirement function, filesystem-as-truth markers, and a carefully reasoned reconciler. The TOCTOU protections (registry_is_live, creating() guard, mtime guard), startup refusal logic, and pool-scoped eviction accounting are all correct and thoroughly tested. Security is solid: erase authorization covers all states (registry record, DB record, marker owner, or 403), path traversal is defended against, and operator data under /mnt is protected. The only issue found is a minor inconsistency where _sweep_side_dirs does not use the is_live re-check function unlike the namespace pass, creating a small TOCTOU window for ephemeral side directories — low impact since they are rebuilt by the next create. Code quality and documentation are exceptional, with every design decision explained in terms of why, not just what. Test coverage is comprehensive across all new modules.

Minor inconsistency: `_sweep_side_dirs` receives the snapshot `live` set but not the `is_live` re-check function, unlike `_reconcile_namespaces` (line 269). A create that commits while `_sweep_side_dirs` is walking could have its side directories removed despite being live. This is low-impact (side dirs are ephemeral and rebuilt by the next create, and the `creating()` guard still applies) but inconsistent with the namespace pass. Consider passing `is_live` to `_sweep_side_dirs` for the same TOCTOU protection.

Nit: `_creating` is a module-level `set`. The tests manage this via context managers, but a `ContextVar` (as used in `storage_pools._pinned_layouts`) would be more robust against cross-test leakage and would be consistent with the existing pattern in this codebase.

foxpatch-aleph (changes requested):
Well-designed and carefully documented overall: the single retire path, the marker system, and the reconciler are coherent, and the live-set/creating-guard/mtime-guard layering shows real thought. However, two races remain that the PR's own guards exist to prevent: _evict() ignores the in-process creating() guard, so a retained VM re-created between the eviction listing and the eviction itself has its disks purged mid-create, and the in-place reinstall rebuild runs outside creating() so its .part downloads are exposed to the sweep. Both are small, targeted fixes.

**`src/aleph/vm/agent/vm/reconciler.py`** (line 514): _evict() checks is_live but never is_creating(namespace), even though the module's other two removal guards do: _is_orphan (line 291) and _is_stale_side_dir (line 412). Race: _reclaimable_on() lists retained VM X with its marker; the owner then re-creates X, whose creating() entry calls adopt() and clears the marker while X is still absent from the registry (the record is committed only at the end of the create), so is_live(X) is False; the evictor then reaches the stale listing, _still_on_disk(X) is True (adopt clears the marker, not the directory), and purge_vm_storage() deletes the volumes the create is actively writing. The window is the whole gap between listing and eviction, which in _enforce_retention_budget / make_room loops over multi-GB directories can be minutes. The creating() docstring (line 107) identifies exactly this race as the reason adoption happens on entry; _evict needs the same one-line check next to the is_live one.

**`src/aleph/vm/agent/views/operator.py`** (line 774): This rebuild is a create path but does not run under creating(), even though the creating() docstring says every create path must be wrapped. _part_roots (reconciler.py line 368) skips only namespaces that are creating, not live ones, so a .part file from this transfer whose mtime stalls longer than VOLUME_CREATE_GUARD (600 s) gets swept mid-transfer; the registry record does not help because the part sweep never consults it. A slow multi-GB rootfs download with one long write gap qualifies. Wrap the recreate_vm_volumes call in creating(str(vm_hash)) (adopt() on entry is a harmless no-op here) or make _part_roots skip live namespaces too.

**`src/aleph/vm/agent/vm/reconciler.py`** (line 555): total is decremented before _evict() has had a chance to decline the eviction, so an entry skipped because it is live, or because its purge was refused (dm-held volumes), still reduces total, and the loop can break while those bytes remain on disk. For the dm-held case the marker survives across passes, so the under-enforcement is stable rather than transient: every pass recomputes the same total, decrements the same un-evictable entry, and stops early. Move the decrement to where the eviction is actually counted.

**`src/aleph/vm/agent/expiry.py`** (line 54): Question, not a blocker: an idle-reaped on-demand program now keeps its host-port forwards indefinitely. For a program never requested again, what releases those mappings? If nothing sweeps them, the REUSE_TIMEOUT path trades a disk leak for a port-forward leak; if they are reclaimed elsewhere, a comment saying so would help.

odesenfans:
Re: the 2026-09-02 review. All three findings were valid; fixes are on the branch (which is also rebased onto the current dev-2.1, 0d85bf41).

**_evict() ignores creating()** — fixed in a9f68bbe~3 (`fix(reconciler): never evict a namespace whose create started mid-pass`). One note: the stack had already grown this guard later, in the caches PR (#1161), which patched both _evict and the cache pass. Reviewing PR A alone was still right to flag it: the guard belongs where _evict is born. It now lives here, and #1161 keeps only the cache-pass half.

**Budget decremented before _evict can decline** — fixed in `fix(reconciler): spend the retention budget only on bytes that left the disk`. The decrement now happens only when the bytes are actually gone: a successful (or dry-run) eviction, or a directory that vanished under an overlapping pass. A declined eviction (live re-create, dm-held refusal) keeps its bytes in the excess, so younger entries make up for a held one instead of every pass stopping early at the same entry. Test pins the dm-held stable-under-enforcement case.

**Reinstall rebuild outside creating()** — fixed in `fix(agent): run the in-place reinstall rebuild under creating()`, taking the suggested wrap rather than teaching _part_roots about live namespaces (which would also shield genuinely stale .part litter of long-running VMs from the sweep).

**Port forwards of an idle-reaped program** — intentional, and now said inline (expiry.py). RECREATE keeps the forwards so the program re-woken by the next request comes back on the same ports; the reservation lasts as long as the program does. When the program actually goes away (scheduler GONE, owner erase), retire_vm deletes the records, port forwards included. A program never requested again but still paying keeps its allocation on purpose.

foxpatch-aleph (commented):

foxpatch-aleph (changes requested):
Well-built PR with a sound design (retire_vm reasons, filesystem-backed markers, careful reconciler guards) and extensive tests, but it contains one genuine capacity-accounting bug: under VOLUME_RETENTION=keep, the largest-single-volume admission check counts a pool's reclaimable bytes in its room figure, while the placement target chosen in _select_from is the free-max pool and only evicts that pool's retained directories — so a volume that fits on a non-free-max pool after eviction is admitted by the check and then repeatedly refused at placement, violating the PR's own invariant that the three disk figures agree.

**`src/aleph/vm/agent/capacity.py`** (line 254): Under VOLUME_RETENTION=keep this check can admit a volume the placement then refuses. roomiest_bytes is the max of (free + reclaimable_bytes(pool.path)) over all pools, but _select_from (storage_pools.py:330) picks the placement target by free bytes only and calls make_room on that single pool. Scenario: pool A free-max with no retained dirs, pool B with 100G free plus 300G retained within budget, and a 400G single-volume VM — _check_max_volume passes (roomiest = B, 400G), then make_room(A) frees nothing and select_pool raises InsufficientResourcesError. The scheduler retries, the create fails again, and a fresh create purges freshly allocated disks each retry (re-download). Either consult the room maker for candidate pools' room rather than only the free-max pool, or use the free-max pool's own room here. Note the new test (test_max_volume_check_counts_that_pools_reclaimable_bytes) only covers the case where the roomiest-by-room pool IS the free-max one, so the mismatch is not caught.

**`src/aleph/vm/agent/vm/reconciler.py`** (line 338): Narrow race with a GONE under keep: the retire drops the registry record, then a periodic pass walking this namespace sees not-live / no marker / old mtime and writes an "orphan" marker. If that write lands after the retire's "gone" marker (os.replace overwrites it), the marker loses its owner, so the retained VM can only leave via the budget and the owner's erase is refused with 403. Re-read the marker before writing the orphan marker (the window between _is_orphan and this write is where the retire's marker lands) to preserve the owner authorization.

**`src/aleph/vm/agent/tasks.py`** (line 425): ItemHash(last_info.vm_id) raises UnknownHashError if the supervisor lists an id that is not an item hash. The reconciler's supervisor_hashes explicitly tolerates such ids ("dropped rather than raised on"), so they are assumed possible; here the conversion sits inside the sweep loop with no guard, so one bad id breaks the whole check_payment pass. The old code (supervisor.delete_vm) accepted any string. Tolerate or guard it the way supervisor_hashes does (same at lines 449 and 507).

**`tests/supervisor/test_agent_no_direct_delete.py`** (line 21): Nit: the "migrations" exclusion silently removes the migration modules from the no-direct-delete pin. Nothing in agent/migration/ calls .delete_vm (verified), so the exclusion can be dropped to make the pin cover every agent module.

odesenfans:
Re: the 2026-09-07 review. Three of the four findings led to fixes on the branch; the fourth is a directory mixup worth recording.

**Admission/placement pool mismatch** — confirmed, the sharpest finding of this round. Fixed on the placement side rather than in the check: `_select_from`'s fallback now offers the room maker every eligible candidate, most free bytes first, instead of only the free-max pool (`fix(storage): the eviction fallback reaches every eligible pool`). Using the free-max pool's own room in `_check_max_volume` would instead under-admit: retained bytes on other pools would stop counting as free, so retention would start blocking creates, the exact failure the design exists to avoid. `make_room` on a pool with nothing retained returns immediately, so the loop costs nothing in the common case. The missing test variant (roomiest-by-room pool != free-max pool) is added.

**Orphan marker vs the gone marker's owner** — confirmed, and slightly worse than stated: the overwrite also loses `depends_on`, so the cache pass could evict parent images the retained volumes still need. Fixed one step past the suggested re-read (which narrows the window but keeps the TOCTOU): the orphan write is now create-exclusive via os.link, so it can never replace any existing marker. That is the orphan pass's intent anyway; the reverse ordering (gone over orphan) stays allowed since it only adds information.

**ItemHash in the payment sweep** — the crash is real but pre-existing: the pass's first loop and `_group_executions_by_payment` both did the bare `ItemHash(info.vm_id)` conversion before any line this PR added runs, so a non-hash id already killed `check_payment` on dev-2.1. Fixed at the root regardless: the snapshot is filtered once at the top, drop-and-warn like `supervisor_hashes`, which repairs the pre-existing fragility and covers this PR's sites together.

**The "migrations" exclusion** — two different directories: the exclusion matches `agent/migrations` (alembic's generated DB-migration scripts), while the live-migration package is `agent/migration` (singular) and is already covered by the pin. Kept the exclusion, added a comment saying which package it names.

foxpatch-aleph (commented):
Well-designed and carefully defended implementation: the retire_vm single entry point with mandatory reasons, the .reclaimable marker protocol with its exclusive os.link write, and the layered create/reconcile race guards are all correct under scrutiny, path handling is consistently hardened, and test coverage of the new modules is thorough. No correctness bugs or security issues found. The comments below are non-blocking: an event-loop performance concern on the admission hot path, a narrow eviction race make_room leaves open, an error-propagation nit in the GONE sweep, and two design-plan docs the PR description says are included but are absent from the commit.

**`src/aleph/vm/agent/capacity.py`** (line 255): Performance: reclaimable_bytes(pool.path) still walks every namespace directory on *every* pool and filters them afterward (reclaimable.py:219-225), so _check_max_volume costs a full multi-pool walk, and _available_disk_bytes (line 276) immediately does a second one — all synchronous on the event loop inside every check_capacity call. On a node with many namespaces that's O(pools × namespaces) stat+read syscalls per allocation request, on the hot create path. Consider a short-TTL cached aggregate, or at least having the per-pool variant iterate only that pool's directories rather than filter the full iter_reclaimable() stream.

**`src/aleph/vm/agent/vm/reconciler.py`** (line 596): make_room protects evictions with only the frozen live snapshot, unlike reconcile_storage, which re-asks the registry at removal time via registry_is_live. A VM whose registry record commits between the supervisor wiring's live_hashes() snapshot and _evict isn't protected here. The window is small and adopt() clears markers on the create path, but since the registered room-maker lambda already closes over the registry, threading registry_is_live(registry, live) through as the default is_live would close the race for free, matching reconcile_storage's reasoning at line 260.

**`src/aleph/vm/agent/vm/retire.py`** (line 124): mark_reclaimable errors (e.g. an OSError from write_marker's temp-file write on a full or read-only filesystem) propagate out of retire_vm(GONE), and as the docstring at line 108 notes, the GONE call sites in check_payment sweep terminal messages with no local try — so one bad pool aborts the sweep for the remaining VMs. The VM is already quiesced and its records are gone at that point, so a failure here is safe to swallow with a logged exception, like the _after_gone() hook just below does.

**`docs/architecture/storage.md`** (line 1): The PR description says the design doc (docs/plans/2026-08-24-disk-reclamation-design.md) and its implementation companion are both included, but neither is in the commit — only docs/architecture/storage.md and vm-lifecycle.md are (docs/plans/ contains only the 2026-08-19 compose-runtime one). Consider adding them, or updating the description; the architecture docs that are included are good.

foxpatch-aleph (changes requested):
Overall a very strong PR: the retire_vm reason taxonomy is applied consistently at every call site, the exclusive-marker write, startup-refusal logic and budget enforcement are all carefully reasoned, docs match the code, and the new test files all pass (local failures are pre-existing environment issues: missing pytest-aiohttp, unwritable /var/lib/aleph and /var/cache/aleph). One genuine race remains in the reconciler's orphan-purge path: the pre-purge re-check re-asks the registry (is_live) but not the creating set, so a create that enters creating() between _is_orphan() and purge_vm_storage() can have its directory purged mid-create — the same class of race _evict already guards against with its is_creating check, and the exact data-loss (host-persistent owner disks) this machinery exists to prevent. Fixing it is a two-line change (re-check is_creating in _reconcile_namespaces, and register in _creating before adopt() in creating()).

**`src/aleph/vm/agent/vm/reconciler.py`** (line 328): Race: this pre-purge re-check only consults `is_live` (registry + snapshot), not the creating set. A create that enters `creating()` after `_is_orphan()` returned True — a span that includes `_still_on_disk()` and `_dir_bytes()` filesystem walks — can have its directory purged mid-create. `_evict` gets this right (it checks `is_creating` immediately before purging, reconciler.py:519); the orphan-purge path should do the same: `if is_live(namespace) or is_creating(namespace):`. Note `test_a_hash_claimed_mid_pass_is_not_purged` only covers the registry-record variant of this race, so this gap is untested.

**`src/aleph/vm/agent/vm/reconciler.py`** (line 109): Ordering: `adopt()` clears the marker *before* `_creating.add(namespace)` registers the create, so for those instructions the directory is protected by neither the marker nor the creating set — a pass evaluating it in that window sees an unmarked, not-creating, old-mtime directory and purges it. Adding to `_creating` first closes the window, because a pass that reads `is_creating` as False will then still see the marker (adopt hasn't run yet) and treat the directory as reclaimable rather than an orphan.

**`src/aleph/vm/agent/capacity.py`** (line 255): Performance note (non-blocking): `reclaimable_bytes(pool.path)` walks every namespace directory on every pool and JSON-parses each marker, and `_available_disk_bytes` does the same — so a single admission check is O(pools × namespaces) stat+read. Fine for the periodic usage report, but during bursts of create/admission checks on a node with many namespaces this adds latency; a short-lived cache invalidated by marker mtimes would help if it shows up in practice.

**`src/aleph/vm/agent/vm/reconciler.py`** (line 458): Minor, same theme as the main finding: the side-directory sweep only consults the `live` snapshot, with no `is_live`/`is_creating` re-check before the rmtree. The VOLUME_CREATE_GUARD mtime check mostly covers it, so this is a nit, but the inconsistency is worth a comment at least.

**`src/aleph/vm/agent/vm/reclaimable.py`** (line 119): Nit: if `os.replace`/`os.link` raises (e.g. EACCES, ENOSPC) the `.tmp` file is left behind. It would be picked up by nothing — consider a try/finally unlink like the exclusive branch, or at least be aware the temp file lingers in the VM directory on failure.

odesenfans:
Re: the 2026-09-07 16:21 review. All landed:

**Orphan-purge re-check** — fixed as suggested: the pre-purge re-check now asks `is_creating` alongside `is_live`, matching `_evict`; the missing creating()-variant of the mid-pass claim test is added.

**`creating()` ordering** — sharp catch, fixed by the suggested swap: the namespace registers in `_creating` before `adopt()` clears the marker, so a pass that reads `is_creating` as False must have read it before the registration and then still sees the marker. A test pins that `adopt` runs with the guard already up.

**Side-dir sweep** — went one step past the suggested comment: the sweep now makes the same `is_live`/`is_creating` re-check immediately before each removal. This was the second review to flag the inconsistency, so retiring it beat documenting it.

**Marker temp file on a failed replace** — fixed with the same try/finally the exclusive branch has.

**Admission cost of `reclaimable_bytes`** — agreed and deferred as you suggested: recorded under the PR's known limitations (a short-lived cache keyed on marker mtimes is the shape of the fix if it shows up in practice). Namespace counts per node are small, so it stays out of this diff.

The branch is also rebased onto the current dev-2.1 (9b6a1d18).

foxpatch-aleph (commented):
High-quality implementation with careful attention to races, failure modes, and data safety. The single deletion choke point (retire_vm with a mandatory reason) is well enforced, the reconciler's two-source live set and startup refusal logic protect against the worst failure scenarios, and the test suite is thorough and targets the right things (mid-pass claims, refused purges, adoption ordering, startup refusals). Findings are all non-blocking: the room maker's live set omits supervisor-listed VMs, reclaimable_bytes() re-walks every marker per admission check, the exclusive marker write can abort a reconcile pass on filesystems without hardlink support, and the no-direct-delete guard can be evaded by aliasing. None warrant blocking the merge.

**`src/aleph/vm/agent/supervisor.py`** (line 518): The room maker is wired with live_hashes() (registry only), not the union with the supervisor's running set that _live_set() builds for reconcile passes. A VM the supervisor runs but the registry lost (the exact DB-loss scenario _startup_refusal guards the startup pass against) with a stale .reclaimable marker on its directory could have its disks evicted by admission pressure before the next periodic pass clears the marker. The make_room docstring acknowledges this ('callers should pass live ... or have reconciled first'), and the window requires both a stale marker on a live VM and a lost registry record, so it is narrow. Consider threading the supervisor's hashes into this hook (e.g., refreshed by the periodic pass or a small cache), or documenting in the wiring comment why the registry alone is deemed sufficient here.

**`src/aleph/vm/agent/capacity.py`** (line 276): reclaimable_bytes() stats and JSON-parses every marker across every pool on each call, and it now runs on every admission check (_available_disk_bytes) and every _check_max_volume evaluation. On a node with hundreds of retained namespaces this turns each capacity request into O(pools x namespaces) filesystem reads, on the request path. The PR's known-limitations note already flags a mtime-keyed cache; given this is per-request hot path, it may be worth prioritizing before large deployments rather than waiting for it to show up.

**`src/aleph/vm/agent/vm/reclaimable.py`** (line 128): In the exclusive path only FileExistsError is handled; os.link can also raise OSError (EPERM, EXDEV, ENOSYS) on filesystems without hardlink support or on a full disk. That exception propagates out of mark_reclaimable and aborts the remainder of the reconcile pass (all namespaces, .part sweep, side dirs, budget), rather than skipping just this directory. The temp file is correctly cleaned either way. Consider catching OSError more broadly and treating it as 'marker could not be written' with a warning, so one uncooperative directory does not stall the whole pass.

**`tests/supervisor/test_agent_no_direct_delete.py`** (line 28): The '.delete_vm(' substring match is evadable by importing the function directly ('from x import delete_vm; delete_vm(...)') or via an alias, and it also cannot see calls constructed dynamically. It is a fine tripwire today (and the comment documents its intent well), but a lightweight AST-based check (find Call nodes whose func resolves through an attribute named delete_vm) would be both exact and future-proof if the file keeps growing.

**`src/aleph/vm/agent/views/operator.py`** (line 689): When a registry record exists but record.message is None, the marker is never consulted (it is only loaded when record is None) and the request falls through to the 403 'No owner recorded' branch even if a retained marker with an owner exists. Rehydration skips empty-message records so this should not occur in practice, and the behavior fails closed either way, but falling back to the marker when record.message is None would be strictly more permissive without weakening auth.

foxpatch-aleph (changes requested):
High-quality PR: the retire_vm consolidation with its mandatory reason is a genuine correctness improvement, the reconciler's concurrency reasoning (creating() ordering, pre-removal liveness re-checks, vanished-directory tolerance, startup refusal on an untrustworthy live set) is unusually careful, and both docs and tests are thorough — the new test files all pass locally. I'd block only on one genuine robustness bug: ReclaimableMarker.from_json assumes the marker parses to a dict, so a marker holding valid non-object JSON raises AttributeError, which read_marker's except clause does not catch. Because iter_reclaimable() feeds reclaimable_bytes(), a single such file makes every capacity admission check raise (node refuses all creates), breaks _check_max_volume and about_system_usage, returns 500 from operate_erase, and aborts each reconcile pass before the backup sweep — permanently, since the file is never cleaned up. The remaining findings are minor races and observations that don't block.

**`src/aleph/vm/agent/vm/reclaimable.py`** (line 102): read_marker's corrupt-marker handling misses a case: from_json does `json.loads(text)` and then `data.get("owner")`, so a marker whose content is valid JSON but not an object (e.g. `[]`, `"x"`, `null`) raises AttributeError, which this `except (OSError, ValueError, KeyError, TypeError)` does not catch. That AttributeError propagates through iter_reclaimable() into reclaimable_bytes(), so one such file makes CapacityManager._available_disk_bytes raise on every admission check (the node refuses all creates), breaks _check_max_volume and about_system_usage, 500s operate_erase, and aborts each reconcile pass before the backup sweep — with no self-healing, since the file is never removed. Adding `AttributeError` (or an `isinstance(data, dict)` check in from_json) plus a test case would close it; the intent is clearly stated by the "Corrupt reclaimable marker ... ignoring it" log line just below.

**`src/aleph/vm/agent/vm/retire.py`** (line 99): Minor race: under GONE+keep, the record is forgotten here and the marker is only written later inside `asyncio.to_thread(_release_storage, ...)` (line 101). In that window the namespace is protected by neither the registry nor a marker, so a concurrent periodic reconcile pass would classify it as an orphan and purge it — defeating retention for that VM (not data loss, since the VM is gone, but the retention policy is silently skipped). Retention is documented as best-effort so this is non-blocking, but capturing the marker write before `registry.forget` (or noting the window) would make the GONE path airtight. Relatedly, an OSError from `mark_reclaimable` (ENOSPC/EACCES) propagates out of retire_vm and aborts the caller's sweep loops (check_payment) with the VM half-retired; the after_gone hook gets the no-local-try protection, storage release does not.

**`src/aleph/vm/agent/vm/reconciler.py`** (line 140): clear_marker has a check-then-unlink race: if the marker vanishes between `path.exists()` and `path.unlink()` (two adopters, or a reconcile pass clearing a stale marker while a create adopts), the FileNotFoundError propagates out of adopt() and fails the create. Everything else in this PR tolerates a directory that vanished first ("every removal tolerates a directory that another pass took first"); `path.unlink(missing_ok=True)` (returning whether it existed via `path.exists()` first is still fine for the boolean) would make clear_marker consistent with that.

**`src/aleph/vm/agent/supervisor.py`** (line 518): The room maker is wired with `live=live_hashes(app["vm_registry"])`, i.e. the registry alone, while reconcile passes deliberately use the registry ∪ supervisor-VMs union (reconciler._live_set) because rehydration can skip running VMs. make_room only evicts marked directories, so reaching this needs a stale marker on a supervisor-running VM with no registry record — a narrow edge (every create path's creating() clears the marker) — but it is an inconsistency with the reconciler's own live-set philosophy. Passing the union (or having make_room consult the supervisor, which may be too expensive on the placement path) would close it; at minimum it's worth a comment on the lambda.

**`src/aleph/vm/agent/vm/reclaimable.py`** (line 118): Nit: if `tmp.write_text()` itself fails (ENOSPC), the temp file is stranded in the VM's directory — the finally clauses only cover the link/replace step, and test_a_failed_marker_write_leaves_no_temp_file only exercises the replace failure. Harmless for the reconciler (`.tmp` isn't a plausible hash and isn't a volume file) but easy to harden with a try/finally around write_text.

odesenfans:
Re: the 19:08 and 19:52 reviews. Landed as five commits on the branch (which is also rebased onto dev-2.1 c927bdd0; #1157's `teardown_vm` survives as the allocation-plane seam and now delegates to `retire_vm(GONE)`).

**Non-object JSON escapes `read_marker`** (blocking) — confirmed, good catch: `from_json` now requires a JSON object (`ValueError` like every other corrupt shape), with tests for `[]`, `"x"`, `null`, `42`, invalid JSON and a missing key. Went one step further: a corrupt marker is *removed* on read, not ignored. Marker writes are atomic so a corrupt file is never a write in progress, and left in place it wedged the directory permanently (the exclusive orphan write backs off from any existing file, so nothing could re-mark or evict it). Removed, the directory re-enters the orphan flow on the next pass.

**`registry.forget` before the gone marker** — the window is real but benign, so the order stays. Trace what a concurrent pass does in it: under `keep` it *marks* "orphan" (it purges nothing), and retire's gone marker then lands via `os.replace`, replacing it with the owner intact; under `reap` retire purges anyway. The only theoretical loss is the budget step evicting that fresh orphan marker within the same pass, and a marker stamped *now* is the youngest, evicted last. Reordering would be actively wrong: a marker written while the registry still holds the record is exactly what `_is_orphan` treats as a stale marker on a live VM and clears. The second half of the finding stands and is fixed: a failed storage release (ENOSPC, EACCES) is logged and no longer aborts `check_payment`'s or the allocation push's sweep with the VM half-retired; the next pass picks the directory up as an orphan.

**`clear_marker` check-then-unlink** — fixed; a marker another adopter or pass removed first is not an error, like every other removal in the reclaimer. Test pins it.

**`os.link` raising beyond `FileExistsError`** — fixed: an exclusive publish the filesystem refuses (no hardlinks, no space) is a warning and "not marked"; the directory stays unmarked and the pass goes on. **Temp file stranded by a failing `write_text`** — the try/finally now covers the write too. Both tested.

**`reclaimable_bytes` on the admission path** — done now rather than deferred: the sum is cached, keyed on the pools' directory mtimes (a namespace directory appearing or going changes them) and invalidated whenever this process writes or removes a marker, with a 5 s TTL for markers another process writes. An in-process change is never stale; the common case costs one stat per pool. Removed from the known-limitations list.

**Room maker's registry-only live set** — closed: the hook now protects the supervisor's VMs as of the last reconcile pass as well (`known_live_hashes`). A listing a few minutes old can only add names to the protected set, never drop a live one.

**AST-based delete guard** — keeping the substring tripwire, with a reason: `delete_vm` is a method on the `Supervisor` handle, not a module-level function, so `from x import delete_vm` has nothing to import and every call is `something.delete_vm(...)`, which the substring catches. The one real evasion (aliasing the bound method) defeats the proposed AST check just as thoroughly.

**Marker fallback when `record.message is None`** — no change: `AgentVmRecord.message: ExecutableContent` is non-optional by construction and the registry API always supplies it, so the branch would guard a state the type forbids.

foxpatch-aleph (approved):
High-quality, carefully reasoned PR. The single retire_vm entry point with a mandatory reason, the .reclaimable marker, and the reconciler are all well-designed with defense in depth (hash-validated namespaces, creating() guard + mtime guard, registry-union-supervisor live set, per-removal liveness re-checks, dm-held purge refusal). The adopt/evict and pass-overlap races are handled thoughtfully, the startup dry-run refusal logic is conservative in the right direction, and the test coverage (including the no-direct-delete guard test) is excellent. All new and migration tests pass; remaining suite failures are pre-existing environment issues. Only minor observations below, none blocking.

**`src/aleph/vm/agent/vm/reconciler.py`** (line 115): Minor: `_creating` is a plain set, not a refcount. Two overlapping creates for the same hash (e.g. a scheduler push racing an operator-triggered reboot/create) means the first exit's `finally` discards the guard while the second create is still in flight. For a young directory the mtime guard still protects, but an adopted (old-mtime) retained directory would only be protected by `creating()`. In practice the program paths serialize per vm_id via `creation_lock`, so the window is theoretical — but a counter (or a set of counts) would remove the class of bug entirely.

**`src/aleph/vm/agent/vm/reconciler.py`** (line 550): Tiny TOCTOU worth documenting: between `_evict`'s `is_creating(namespace)` check (line 550) and `purge_vm_storage` (line 561), a create can enter `creating()` and adopt the directory — clearing the marker — and its disks are then purged mid-adopt. The window is microseconds and requires the pool to be over retention budget at the same instant, so it's not a blocker, but a comment noting the residual window (or re-checking `is_creating` inside `_evict` right before the purge) would harden it.

**`src/aleph/vm/agent/vm/purge.py`** (line 89): `vm_has_volumes` counts any regular file, including stale `*.part` leftovers from a previous failed attempt. Such a partial makes `had_volumes` True, so a subsequent create failure retires RECREATE and keeps a phantom registry record for a VM whose only 'volume' is a dead partial. Conservative and self-healing (the parts sweep removes the `.part`; the allocation cycle eventually retires the record), but worth a line in the docstring if this is unintended.

**`src/aleph/vm/storage_budget.py`** (line 19): Lint: with the project's own ruff config this file has 3x EM102 (f-string in exception, lines 19/25/29) and 1x PLR2004 (magic `100`, line 24); `reconciler.py:471` adds a PLR0913. If CI enforces `ruff check` on src, these will fail — worth fixing before merge.

**`src/aleph/vm/agent/vm/reconciler.py`** (line 397): Nit: `_sweep_parts` skips namespaces held by `creating()` but not namespaces that are live-but-not-creating; this is only sound because every download path (create, reinstall's recreate_vm_volumes, migration import) now runs under the guard. That invariant is stated in the module docstring, but a pointer here would help future readers adding a new download path that runs against a running VM.

odesenfans:
Round 7 (the approval's observations), on `d475ee77`:

- **`_creating` is a set, not a refcount**: fixed in `ed08514f`. It is now a namespace-to-count dict; the guard drops when the last overlapping `creating()` exits. Test: `test_overlapping_creates_of_one_hash_stay_guarded_until_the_last_exits`. You are right that only the program path serializes per hash: an operator reinstall and a scheduler push take no common lock.
- **TOCTOU between the `is_creating` check and the purge in `_evict`**: documented in `a798d5ec` rather than narrowed. The window exists only on the periodic pass (worker thread); the room maker's evictions run on the event loop where `creating()` cannot interleave. Closing it would need a lock shared with `creating()` held across the rmtree, which would stall every create for the length of a purge.
- **`vm_has_volumes` counts stale `.part` files**: fixed in `bca2076c`. `.part` and `.tmp` are skipped there only; `iter_volume_files` still yields them so the purges remove them. Test: `test_vm_has_volumes_ignores_a_partial_download`.
- **Lint**: `c3640430` clears the EM102 and PLR2004 findings in `storage_budget.py`. CI runs `ruff format --diff` only and `ruff check src` reports 173 pre-existing errors on `dev-2.1`; the PLR0913 on the keyword-only reconciler signatures stays with the 12 existing ones.
- **Parts sweep pointer**: `d475ee77` adds to `_part_roots` why consulting the creating set alone is sound and what a new download path must do.

Stack B to E re-cascaded on top (one fixup on D for two tests that patched the guard as a set).

foxpatch-aleph (changes requested):
High-quality, carefully reasoned PR: the retire_vm funnel with mandatory reasons is enforced by a pinning test, the reconciler's live-set reasoning (registry ∪ supervisor, re-checked immediately before removals) and the create-guard belt-and-braces are sound, and all new/changed suites pass (remaining full-suite failures are pre-existing environment issues, verified identical on base main). Blocking items are all in new failure paths: (1) an OSError while writing an orphan marker aborts the reconcile pass, and at startup propagates out of the on_startup hook so a full pool prevents the agent from booting — the very state the reconciler exists to fix; (2) adopt() runs outside the try in creating(), so a raise permanently leaks the create guard; (3) the insufficient-balance stop now retires GONE (record dropped, volumes purged under default reap) where the old code deliberately kept the record for re-payment — a data-loss semantic change on a transient wallet shortfall that deserves explicit confirmation; (4) make_room evicts every retained directory on a pool when the create can never fit there, and _select_from repeats this per pool. Everything else (path validation, marker atomicity, erase authorization from the marker's owner, /mnt rmdir-only) checks out.

**`src/aleph/vm/agent/vm/reconciler.py`** (line 371): A marker-write failure here aborts the whole pass. write_marker's exclusive path deliberately swallows os.link errors ("raising here would abort the whole reconcile pass for one directory"), but tmp.write_text at reclaimable.py:137 can still raise OSError (ENOSPC/EACCES) in any mode, and mark_reclaimable does not catch it. periodic_reconcile and the after-gone hook swallow it, but reconcile_at_startup does not: an OSError during the non-dry startup pass propagates out of the on_startup hook and the agent fails to boot — on a full pool, the exact condition the reconciler exists to relieve, turning it into a boot loop the operator can only break by hand. _sweep_parts and _sweep_side_dirs already take the per-item try/except approach; the namespace pass should do the same (and arguably reconcile_at_startup should not let any reconcile failure prevent boot).

**`src/aleph/vm/agent/vm/reconciler.py`** (line 120): adopt(namespace) runs before the try, so if it raises the _creating counter is never decremented and the namespace is exempt from every future reconcile pass (guarded forever, marker or not). The increment is correctly before the adopt for the documented ordering, but the adopt itself should be inside the try: try: adopt(namespace); yield / finally: ... — one line, closes the leak.

**`src/aleph/vm/agent/tasks.py`** (line 436): Behavioral change to confirm explicitly: the old test asserted `ItemHash(hash) in registry` with the comment "(VM may be re-paid and restarted)" — the insufficient-balance stop deliberately kept the record. GONE now drops the record, the DB rows and the port mappings, and under the default VOLUME_RETENTION=reap purges the host-persistent volumes. A wallet balance dipping below required + PAYMENT_BUFFER is a transient shortfall, not the "stopped payment" the docs list as GONE knowledge; users who intended to top up now lose their data where it previously survived. If this is intended, it should be called out in the PR description / upgrade notes; if not, this site (and the credit-balance one below) may warrant different treatment.

**`src/aleph/vm/agent/vm/reconciler.py`** (line 655): make_room's only bounds are "pool fits now" and "freed >= needed". When needed_bytes exceeds everything the pool could ever give back (free + all reclaimable), the loop evicts every retained directory on the pool and returns 0, and _select_from (storage_pools.py:339) then repeats this on each remaining pool — so a multi-volume VM that passes aggregate admission (reclaimable counted as free) but cannot be placed wipes retained data on every pool before failing anyway. _check_max_volume covers the oversized-single-volume case; consider a third bound here, e.g. break out when _pool_free(pool) + reclaimable_bytes(pool.path) < needed_bytes.

**`src/aleph/vm/agent/vm/reclaimable.py`** (line 135): Minor: the cache is invalidated before the change is published, so a concurrent reclaimable_bytes() computed between invalidate and os.replace caches the pre-change sum; a marker write does not change the pool directory's mtime, so the fingerprint will not invalidate it and the stale value survives up to the 5s TTL despite the "never stale" claim in the module comment. Invalidating after the publish too (or in a finally) closes the window.

**`src/aleph/vm/agent/migration/runner.py`** (line 366): The registry record is created before finish_instance_create, and a failure there leaves a half-configured VM running and registered, where create_vm_execution retires the half-started VM for the same failure. The dest_dir cleanup in the except block also skips it in that case (because _vm_exists is true). Consider mirroring the create path's teardown for a finish_instance_create failure, or noting why the import path deliberately differs.

**`docs/architecture/storage.md`** (line 1): The PR description states the design and implementation plan (docs/plans/2026-08-24-disk-reclamation-design.md and -implementation.md) are "both included", but neither is in the repo, and many comments cite "spec section 1/3/6" and "spec S1" that reviewers cannot consult. Please either add the plan documents or fix the description.

odesenfans:
Round 8, on `447bb1cc` (the stack is also rebased onto the current `dev-2.1`, which absorbed #1158 and #1159):

- **Marker write failure aborts the pass / the boot**: fixed in `8e6d67b5`, in three layers. `write_marker` treats any OSError on the temp write or the publish as "not marked" with a warning; the namespace pass skips a directory whose handling raises, like the parts and side-dir sweeps; `reconcile_at_startup` logs a failed pass instead of propagating out of the on_startup hook. Tests for all three.
- **`adopt()` outside the try in `creating()`**: fixed in `992d657f`, adopt runs inside the try. Test: an adopt that raises leaves the guard down.
- **Payment shortfalls retire GONE**: intended, and now stated. The design lists a stopped payment under GONE, and the alternative (stop and keep the record) has no exit: the allocation push and the terminal-status check both iterate the supervisor's running VMs, so a stopped-unpaid VM with a kept record would hold its disks and its committed capacity forever, the leak this PR closes. `docs/architecture/storage.md` and `vm-lifecycle.md` (`a2608378`) and the PR description now say so, including the 1.x difference and the `VOLUME_RETENTION=keep` route for a grace period (the marker carries the owner and a re-create adopts the disks back).
- **`make_room` evicts everything when the create can never fit**: fixed in `38226c80`. A third bound skips the pool when free plus its retained bytes is short of the need, so placement fails without wiping anything, on that pool or the next. Test: `test_make_room_evicts_nothing_when_the_create_could_never_fit`.
- **Cache invalidated before the publish only**: fixed in `447bb1cc`, `write_marker` and `clear_marker` invalidate again once the change is on disk. Test reads mid-write and after.
- **Migration import failure after `create_vm`**: fixed in `3789bf54`. The import now retires the VM `FAILED_CREATE`, the teardown a fresh create gets when it fails after `create_vm`. Not RECREATE: a failed import already discards its staged disks, so a kept record would describe disks that are gone.
- **Design citations**: gone from src and tests (`b5d9127c` on A, one commit each on B and D), rationale inlined where the citation carried it. The PR description no longer claims the plan documents are included. The two "design doc" references in `vm_registry.py` predate this PR.

foxpatch-aleph (commented):
This is a large, unusually careful PR: the single retire_vm(reason) funnel (with a test pinning that nothing else calls supervisor.delete_vm), the .reclaimable marker with exclusive/atomic writes, and the reconciler are all well designed, and the hard parts (startup refusal states, liveness re-checks immediately before each removal, purge-refusal accounting, the creating()-adopts-before-registering ordering) are both documented and tested. I could not execute the test suite (the sandbox has the pre-existing web3/eth_typing import breakage), so this is a static review. I found no blocking bugs: the residual race windows are known and documented, and the mtime guard is sound against in-flight downloads because sock_read=120s means an active .part can never outlive the 600s guard. My comments are robustness nits in error paths (an unguarded iterdir in _sweep_side_dirs, the corrupt-marker unlink able to raise into every admission check, delete_records_for_vm sitting outside the swallow-all try in retire_vm), one inaccurate comment (the room maker runs in a worker thread, not on the event loop, so the creating() race window applies to it too), and one product-level flag: under the default VOLUME_RETENTION=reap, a transient payment shortfall now immediately purges host-persistent volumes, converting a temporary lapse into permanent data loss unless operators opt into keep — deliberate and documented, but worth explicit sign-off.

**`src/aleph/vm/agent/vm/reconciler.py`** (line 607): This last sentence is inaccurate: the room maker is only reached through _select_from, and every production caller of volume_path_for/select_pool goes through asyncio.to_thread (storage.py:340/506, downloader.py:73, migration/runner.py), so make_room runs in a worker thread where creating() CAN interleave. The code is still safe because _evict re-checks is_creating before purging, but the comment asserts a guarantee the code does not have and weakens the reasoning for why the residual window is acceptable.

**`src/aleph/vm/agent/vm/reconciler.py`** (line 528): list(root.iterdir()) is not OSError-guarded, unlike the identical operations in _sweep_parts and storage_pools.iter_namespace_dirs. A PermissionError on /mnt (or a staging root the agent user cannot read) would abort the pass before _enforce_retention_budget and the backup sweep run, on every cycle, forever. Worth the same try/except OSError -> continue the neighbours use.

**`src/aleph/vm/agent/vm/reclaimable.py`** (line 119): unlink(missing_ok=True) only swallows FileNotFoundError; an EACCES/EPERM/EBUSY here propagates out of read_marker, through iter_reclaimable() into reclaimable_bytes(), i.e. into every admission check and capacity report. Every other destructive/cleanup operation in this module deliberately swallows OSError; this is the one hole, and a read-only pool directory would break VM creation.

**`src/aleph/vm/agent/vm/retire.py`** (line 100): delete_records_for_vm is awaited outside the swallow-all try around _release_storage, yet the surrounding comments argue a storage/DB failure must not abort the caller's sweep (check_payment, allocation push). A DB error here leaves the VM half-retired — supervisor-forgotten and registry-forgotten but storage unreleased — and stops the rest of the sweep, which is exactly the outcome the try/except below exists to prevent.

**`src/aleph/vm/agent/vm/reconciler.py`** (line 398): mark_reclaimable's return value (and its per-directory list of written markers) is ignored: on ENOSPC/EACCES it returns [] and logs, but the namespace is still appended to marked_orphans, so the startup preview and the pass summary can report a mark that never landed. Consider only appending when something was actually written.

**`src/aleph/vm/agent/vm/reclaimable.py`** (line 88): Only MARKER_NAME is excluded here, so a .reclaimable.tmp/.reclaimable.x.tmp left behind by a writer that crashed between write_text and the publish inflates every subsequently recorded size_bytes for that directory (the write_marker finally only cleans up within a live process). Excluding the temp names too would keep the budget figures honest after a crash.

**`src/aleph/vm/agent/capacity.py`** (line 485): Cosmetic: the message says 'roomiest pool has X MiB' but X now includes reclaimable bytes (free + reclaimable), so an operator reading a refusal sees a number larger than the pool's actual free space. Maybe 'roomiest pool can free X MiB' or similar.

**`src/aleph/vm/agent/vm/reconciler.py`** (line 640): If _pool_total() fails it returns 0, and a percentage budget then parses to 0, so every retained directory on that pool is evicted on the next pass. An unreachable pool is already logged elsewhere; falling back to something non-zero (or skipping enforcement for that pool) would be gentler.

**`src/aleph/vm/agent/tasks.py`** (line 436): Flagging the flagship behaviour change rather than a bug: a payment shortfall now retires GONE, and under the default VOLUME_RETENTION=reap this immediately purges host-persistent volumes — a transient payment lapse becomes permanent data loss unless the operator opts into keep. This is clearly documented in the PR description and both architecture docs, but given reap is the default it deserves explicit product-level sign-off before merge.

foxpatch-aleph (approved):
Solid, carefully-reasoned design. The single retire_vm entry point with a mandatory reason, the re-create vs failed-create split (had_volumes snapshot before allocation), and the .reclaimable marker scheme are all coherent and well-tested. Security posture is strong: all delete paths go through hash-validated namespaces, /mnt is only ever rmdir'd, ownerless markers refuse erase (403, fail-closed), and the startup pass refuses to purge when the live set can't be established. Two minor documentation findings, neither blocking: the _evict comment mis-states the room maker's threading model (it runs in a worker thread via asyncio.to_thread, sharing the same residual race window as the periodic pass, as make_room's own docstring says), and periodic_reconcile caches the reconcile interval for the task's lifetime.

**`src/aleph/vm/agent/vm/reconciler.py`** (line 608): This comment is incorrect: the room maker's evictions do NOT run on the event loop. make_room is reached via storage_pools._select_from, whose callers wrap select_pool/volume_path_for in asyncio.to_thread (storage.py:339, storage.py:505, downloader.py:73, migration/runner.py:295), so it runs in a worker thread exactly like the periodic pass and shares the same residual window described in the preceding lines. make_room's own docstring ("Synchronous, and called from placement: its callers wrap it in asyncio.to_thread") says as much. Since this comment is the safety argument for the residual race, it should describe both paths accurately.

**`src/aleph/vm/agent/vm/reconciler.py`** (line 836): periodic_reconcile reads settings.VOLUME_RECONCILE_INTERVAL once, on task start, so a runtime change to the setting is never picked up by the running task. Minor, but reading it inside the loop (or documenting that the interval is startup-time only) would make the behavior less surprising.

PR #1156, opened 2026-08-25
feat(agent): admit VMs from the message before allocating (2.1, PR B)

odesenfans:
## Summary

PR B of the 2.1 disk reclamation work (spec Section 2, "Admission before allocation"). Stacked on #1155.

- `CapacityManager.check_message(content, *, exclude_vm_hash=None)`: admission from the message (disk, memory, vCPUs) before a byte is downloaded or allocated. One admission path: `_admit` is gone, and the four post-download `check_capacity(disk_mib=0)` blocks that #1153 had left on the create paths are removed.
- `_ensure_program_vm` (on-demand programs) and the cold migration import now admit before their downloads; #1153 had not covered either. The reserve endpoint uses the same path.
- Re-creates are not charged for what they already hold: the discount is per declared volume, matched by on-disk filename and capped at the declared size; stale files never discount; one file discounts at most one volume; the held bytes of the largest volume are credited to the pool that holds its file rather than shrinking the requirement.

## Known follow-ups
- Volume-name-to-filename derivation is a convention duplicated between capacity, storage and downloader; one shared function would remove the duplication.
- A persistent volume literally named `rootfs` next to a boot disk never discounts (fail-closed).
- A volume matches its stem under every suffix its kind can take, so a file left in another on-disk format still discounts a volume the create allocates afresh; bounded by that volume's declared size, and placement still checks real free space. The exact filename belongs with the shared naming helper above.

## Test plan
- `tests/supervisor`: 1336 passed, 48 pre-existing environment failures (identical set to base).
- New/extended: `test_agent_capacity.py` (per-volume discount, stale files, colliding names, per-pool credit, counter-examples), `test_create_path_disk_admission.py` (stubs expose only `check_message`, so a reintroduced direct `check_capacity` fails loudly), `test_run_program_path.py`.

### DIFF
--- a/docs/architecture/vm-lifecycle.md
+++ b/docs/architecture/vm-lifecycle.md
@@ -107,35 +107,44 @@ never survives on disk, so there is nothing to reattach to.
 Admission happens in two layers that never overlap in what they check:
 
 - **Agent-side policy** (`src/aleph/vm/agent/capacity.py`,
-  `CapacityManager.check_capacity`) runs before every `create_vm` call, after
-  the spec is built (so a failed download or bundle fetch never consumes
-  capacity). It enforces two-bucket memory accounting: instances (and
-  V-PROGRAMs, which are full SNP VMs admitted against the instance bucket)
-  share `physical - HOST_MEMORY_RESERVED_MIB - PROGRAM_MEMORY_RESERVED_MIB`;
+  `CapacityManager.check_message`) runs once per create, from the message
+  and before any download or allocation, so a host without room refuses
+  before the downloader has written anything. It is the single admission
+  path of every create-path caller in `src/aleph/vm/agent/run.py`
+  (`create_vm_execution`, which covers instances, programs and V-PROGRAMs,
+  and `_ensure_program_vm`, the on-demand program path), of the cold
+  migration import (`src/aleph/vm/agent/migration/runner.py`) and of the
+  `/control/reserve_resources` dry-run endpoint (`operate_reserve_resources`
+  in `src/aleph/vm/agent/views/__init__.py`), which reports capacity ahead
+  of a scheduler placement decision without creating anything. It enforces
+  two-bucket memory accounting: instances (and V-PROGRAMs, which are full
+  SNP VMs admitted against the instance bucket) share
+  `physical - HOST_MEMORY_RESERVED_MIB - PROGRAM_MEMORY_RESERVED_MIB`;
   programs share `PROGRAM_MEMORY_RESERVED_MIB` alone. vCPUs are capped at
-  `physical_cores * VCPU_OVERCOMMIT_FACTOR` across both buckets. Every real
-  create-path caller in `src/aleph/vm/agent/run.py` (`create_vm_execution`,
-  which covers instances, programs and V-PROGRAMs, and `_ensure_program_vm`,
-  the on-demand program path) passes `disk_mib=0` and no `max_volume_mib`,
-  so `check_capacity`'s disk checks no-op on create: disk is not actually
-  gated at create time. `check_capacity` also implements an
-  aggregate-free-space check and a single-largest-volume-fits-the-roomiest-
-  pool check (`_check_max_volume`), but the only caller that exercises them
-  with real figures is the separate `/control/reserve_resources` dry-run
-  endpoint (`operate_reserve_resources` in
-  `src/aleph/vm/agent/views/__init__.py`), which reports capacity ahead of a
-  scheduler placement decision without creating anything. The VM's own
-  early registry record (recorded before the spec build, to make owner-auth
-  answerable during a slow confidential download) is excluded from the
-  committed memory/vCPU sums via `exclude_vm_hash`, or a create would count
-  its own request against itself. GPU admission is a separate reservation
-  ledger (`CapacityManager.holds`, keyed by concrete `pci_host`) with a
-  short-lived hold/resolve two-step: `reserve_gpus` (used by the same
+  `physical_cores * VCPU_OVERCOMMIT_FACTOR` across both buckets. Disk is
+  judged from the sizes the message declares (the instance rootfs and every
+  `size_mib` volume): the total must fit the pools' free space plus what the
+  reclaimer could free, and the single largest volume must fit one pool
+  (`_check_max_volume`). A re-create is not charged for what it already
+  holds: each declared volume is matched to its existing file by name inside
+  `{pool}/{vm_hash}/`, the credit is capped at the declared size, one file
+  discounts at most one volume, and the largest volume's held bytes are
+  credited to the pool holding its file rather than shrinking the
+  requirement. Files that back no declared volume never discount. Since
+  `creating()` adopts a retained directory before admission runs, held
+  bytes are neither counted as reclaimable nor discounted twice. The VM's
+  own early registry record (recorded before the spec build, to make
+  owner-auth answerable during a slow confidential download) is excluded
+  from the committed memory/vCPU sums via `exclude_vm_hash`, or a create
+  would count its own request against itself. GPU admission is a separate
+  reservation ledger (`CapacityManager.holds`, keyed by concrete `pci_host`)
+  with a short-lived hold/resolve two-step: `reserve_gpus` (used by the same
   dry-run endpoint) holds a card for a user for `RESERVATION_TTL_SECONDS`,
-  `resolve_gpus` (called from the create path) consumes the caller's own
-  holds and skips cards held by another user. So create-time admission is,
-  in practice, memory plus vCPUs plus GPU holds; disk admission is a
-  dry-run-only feature of the same `CapacityManager`.
+  `resolve_gpus` (called from the create path after the spec is built)
+  consumes the caller's own holds and skips cards held by another user.
+  `check_capacity` remains underneath as the scalar seam that `simulate`
+  (batch placement advice) uses; it cannot apply the held-volume discount,
+  so an advisory answer is at most more conservative than the enforced one.
 - **Supervisor mechanism backstops** (`check_memory_backstop`,
   `validate_spec_gpus` in `lifecycle.rs`) run under `creation_lock` inside
   `create_vm_inner` itself: committed memory (summed from every tracked
--- a/src/aleph/vm/agent/capacity.py
+++ b/src/aleph/vm/agent/capacity.py
@@ -15,16 +15,23 @@ from __future__ import annotations
 
 import asyncio
 import logging
+import re
 from collections.abc import Collection
 from dataclasses import dataclass, field
 from datetime import datetime, timedelta, timezone
+from pathlib import Path
 
 import psutil
 from aleph_message.models import ExecutableContent, ItemHash, VerifiableProgramContent
 from aleph_message.models.execution.instance import InstanceContent
 
 from aleph.vm import storage_pools
-from aleph.vm.agent.vm.reclaimable import directory_size_bytes, reclaimable_bytes
+from aleph.vm.agent.vm.purge import ROOTFS_STEM, _checked_namespace
+from aleph.vm.agent.vm.reclaimable import (
+    MARKER_NAME,
+    file_size_bytes,
+    reclaimable_bytes,
+)
 from aleph.vm.agent.vm_registry import AgentVmRegistry
 from aleph.vm.conf import settings
 from aleph.vm.resources import GpuDevice, InsufficientResourcesError
@@ -78,27 +85,123 @@ def is_instance_bucket(content: ExecutableContent) -> bool:
     return isinstance(content, (InstanceContent, VerifiableProgramContent))
 
 
+@dataclass(frozen=True)
+class DeclaredVolume:
+    """One volume a message asks the node to allocate.
+
+    ``stems`` are the filename stems the volume's file can carry inside
+    ``{pool}/{vm_hash}/``: "rootfs" for the boot disk, the volume's name for
+    an extra persistent volume (see ``purge.ROOTFS_STEM`` and the naming
+    convention it documents). Both spellings of a name are listed because
+    ``storage.get_volume_path`` sanitizes a name that is not already
+    ``[\\w\\-_/]+`` while ``downloader._make_writable_volume`` does not.
+    """
+
+    stems: tuple[str, ...]
+    size_mib: int
+
+
+@dataclass(frozen=True)
+class HeldVolume:
+    """An existing file that already backs a declared volume.
+
+    ``size_bytes`` is what the file allocates on disk, capped at the size the
+    volume declares: a file cannot discount more than the message asks for.
+    """
+
+    path: Path
+    size_bytes: int
+
+    @property
+    def pool_path(self) -> Path:
+        """The pool holding the file (``{pool}/{vm_hash}/{name.ext}``)."""
+        return self.path.parent.parent
+
+
+def _volume_stems(name: str) -> tuple[str, ...]:
+    """The filename stems a declared volume's file can carry on disk."""
+    sanitized = name if re.match(r"^[\w\-_/]+$", name) else re.sub(r"[^\w\-_]", "_", name)
+    return (name,) if sanitized == name else (name, sanitized)
+
+
+def declared_volumes(content: ExecutableContent) -> list[DeclaredVolume]:
+    """The per-VM volumes a message asks the node to allocate.
+
+    Immutable volumes are absent on purpose: they resolve to a shared cache
+    entry, not to a file this VM owns.
+    """
+    volumes: list[DeclaredVolume] = []
+    if isinstance(content, InstanceContent) and content.rootfs:
+        volumes.append(DeclaredVolume(stems=(ROOTFS_STEM,), size_mib=content.rootfs.size_mib))
+    for volume in content.volumes or []:
+        size_mib = getattr(volume, "size_mib", 0) or 0
+        name = getattr(volume, "name", "") or ""
+        volumes.append(DeclaredVolume(stems=_volume_stems(name) if name else (), size_mib=size_mib))
+    return volumes
+
+
 def requirements_from_message(content: ExecutableContent) -> ResourceRequirements:
     """Extract the resources a message requests into a message-free DTO."""
-    is_instance = is_instance_bucket(content)
-    volume_sizes_mib: list[int] = []
-    if isinstance(content, InstanceContent) and content.rootfs:
-        volume_sizes_mib.append(content.rootfs.size_mib)
-    for volume in content.volumes or []:
-        volume_sizes_mib.append(getattr(volume, "size_mib", 0) or 0)
+    volume_sizes_mib = [volume.size_mib for volume in declared_volumes(content)]
     return ResourceRequirements(
         vcpus=content.resources.vcpus,
         memory_mib=content.resources.memory,
         disk_mib=sum(volume_sizes_mib),
         max_volume_mib=max(volume_sizes_mib, default=0),
-        is_instance=is_instance,
+        is_instance=is_instance_bucket(content),
         gpu_device_ids=requested_gpu_ids(content),
     )
 
 
-def allocated_bytes_for(vm_hash: ItemHash | str) -> int:
-    """Bytes already sitting under this VM's directories on every pool."""
-    return sum(directory_size_bytes(directory) for directory in storage_pools.iter_namespace_dirs(str(vm_hash)))
+def existing_volume_files(vm_hash: ItemHash | str) -> dict[str, Path]:
+    """``{stem: path}`` of the VM's existing volume files, on every pool.
+
+    The namespace is validated before it reaches a path join, like every
+    other caller of ``iter_namespace_dirs``: ``pool.path / namespace`` is a
+    bare join, so an unchecked hash lets "../.." resolve and be counted as
+    space the VM already holds, and an inflated discount relaxes admission.
+
+    Symlinks and the marker are skipped, and pools are walked in order with
+    the first match winning, so the result does not depend on iteration luck.
+    """
+    namespace = _checked_namespace(vm_hash)
+    files: dict[str, Path] = {}
+    for directory in storage_pools.iter_namespace_dirs(namespace):
+        try:
+            entries = sorted(directory.iterdir())
+        except OSError:
+            logger.warning("Volume directory %s not readable, not discounting it", directory)
+            continue
+        for entry in entries:
+            if entry.name == MARKER_NAME or entry.is_symlink() or not entry.is_file():
+                continue
+            files.setdefault(entry.stem, entry)
+    return files
+
+
+def held_volumes(vm_hash: ItemHash | str, volumes: list[DeclaredVolume]) -> list[HeldVolume | None]:
+    """What the VM already holds for each declared volume, positionally.
+
+    The discount is per declared volume, never per directory: a file that
+    backs no declared volume (a volume the updated message renamed or
+    dropped, left behind on a node the VM is re-scheduled to) discounts
+    nothing. Summing the directory instead would let such a leftover cancel
+    the space a genuinely new volume still needs, turning the disk check off
+    for exactly the create that needs it.
+
+    Each hit is capped at the size its volume declares, so a file that grew
+    past its declaration cannot credit the difference either.
+    """
+    existing = existing_volume_files(vm_hash)
+    held: list[HeldVolume | None] = []
+    for volume in volumes:
+        path = next((existing[stem] for stem in volume.stems if stem in existing), None)
+        if path is None:
+            held.append(None)
+            continue
+        size_bytes = min(file_size_bytes(path), volume.size_mib * 1024 * 1024)
+        held.append(HeldVolume(path=path, size_bytes=size_bytes))
+    return held
 
 
 def requested_gpu_ids(content: ExecutableContent) -> list[str]:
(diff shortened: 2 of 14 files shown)


foxpatch-aleph (approved):
This PR cleanly moves disk admission from post-download (where it double-counts space the VM just allocated) to pre-download via `check_message`, consolidating four separate `check_capacity(disk_mib=0)` call sites into a single message-driven admission path. The per-volume discount logic is well-designed: files are matched to declared volumes by filename (stem + suffix), consumed on match (one file discounts one volume), capped at the declared size, and credited only to the pool that holds the file. The stale-file counter-example is handled correctly — a leftover file from a renamed/dropped volume never discounts a new one. Security is solid: namespace validation prevents path traversal, symlinks are skipped, and the marker file is excluded. Test coverage is thorough, including end-to-end tests through the real `check_message` → `check_capacity` path, counter-examples for stale files and colliding names, and stubs that deliberately expose only `check_message` to catch regressions. Two minor conservative edge cases exist (tie-breaking in `largest` volume selection may lose a credit when the first max-size volume has no held file, and floor division in `held_mib` may lose sub-MiB discounts), but both under-admit rather than over-admit. The migration runner follow-up is explicitly acknowledged.

**`src/aleph/vm/agent/capacity.py`** (line 302): When multiple declared volumes share the same maximum `size_mib`, `max(range(...), key=...)` returns the first index. If that volume has no held file but a later one (same size) does, `max_volume_credit` is `None` and the per-pool check doesn't credit the pool that holds the later volume's file. This is conservative (under-admit) and unlikely in practice — a recreate typically holds files for all previously-declared volumes — but could cause a false rejection in a contrived edge case. Not blocking.

**`src/aleph/vm/agent/capacity.py`** (line 295): The floor division `// (1024 * 1024)` on the summed held bytes means sub-MiB fractional discounts are truncated. With many small held volumes, the cumulative truncation could lose a meaningful discount. Conservative (under-admit), not blocking, but worth noting if small volumes ever become common.

**`src/aleph/vm/agent/vm/reclaimable.py`** (line 87): Minor: `path.is_symlink()` does a second `lstat` after `path.lstat()` on line 84 already stored the result in `st`. The `st` could be used directly via `stat.S_ISLNK(st.st_mode)` to avoid the redundant syscall. Not a bug, just a minor inefficiency.

foxpatch-aleph (approved):
This PR cleanly replaces the post-download `check_capacity` admission gate (which passed `disk_mib=0` and thus never actually checked disk) with a single message-driven `check_message` call that runs before any allocation. The per-volume discount logic is well-designed: it matches files to declared volumes by filename (stem + suffix), caps discounts at declared sizes, consumes matched files so one file can't discount two volumes, and credits the per-pool check only to the pool that actually holds the file. The stale-file counter-example and colliding-stem edge cases are tested end-to-end through `check_capacity`. Path traversal via `../..` in the VM hash is prevented by `_checked_namespace`, and symlinks are correctly excluded from discount calculations. The test stubs deliberately omit `check_capacity` to catch regressions where `run.py` calls it directly. No correctness, security, or coverage issues found.

**`src/aleph/vm/agent/capacity.py`** (line 110): The regex `^[\w\-/]+$` allows `/` in volume names, matching `storage.py:493`, but `existing_volume_files` only scans direct children of `{pool}/{vm_hash}/`. A volume name containing `/` would produce a file in a subdirectory that the discount scan never finds — a missed discount (fail-closed), not an over-admission. This is noted in the PR's known follow-ups (duplicated naming convention), so no action needed here.

**`src/aleph/vm/agent/capacity.py`** (line 295): The floor division `// (1024 * 1024)` means a held file smaller than 1 MiB contributes 0 to the discount. This is conservative (under-discounts), which is the safe direction for admission, but worth being aware of if someone debugs a recreate that's refused by a fraction of a MiB.

foxpatch-aleph (approved):
This PR is well-executed. The unification of admission into check_message before the download is correct across all four create paths (including the _ensure_program_vm gap), and the re-create discount is carefully constructed: per-volume matching by filename (never a directory sum), each hit capped at the declared size, one file consumed per volume so nothing discounts twice, and the per-pool credit for the largest-volume check going only to the pool that holds the file. I verified the filename-derivation conventions against every allocator (qcow2/btrfs/ext4, sanitized and raw name spellings, the rootfs-stem collision rule) and they match; the namespace is validated before any path join; symlinks and markers are excluded so a discount cannot be inflated; and adoption ordering (creating() clears the marker before check_message) keeps _available_disk_bytes and the discount consistent. The counter-example tests (stale file, colliding stem, cross-pool credit, fail-closed rootfs) are exactly the right ones, and all changed test files pass locally. Remaining findings are nits, none blocking.

**`src/aleph/vm/agent/capacity.py`** (line 293): `exclude_vm_hash` now does double duty: it is both the registry-exclusion key and the discount-lookup namespace. They are semantically distinct concerns ("skip my own committed record" vs "read my directory"); today every caller wants both or neither, so this is fine, but a future caller wanting one without the other will be stuck. Consider at least documenting the dual role on `check_message` itself, or splitting the parameters when the next caller appears.

**`src/aleph/vm/agent/capacity.py`** (line 28): Importing the private `_checked_namespace` from purge across modules works, but note its ValueError message ("Refusing to purge storage for an implausible VM hash") is confusing when it fires from an admission path. A tiny public wrapper or a reworded message would help debuggability; the same package-private reuse argument applies to `_checked_namespace` vs `purge._ITEM_HASH_PATTERN`.

**`src/aleph/vm/agent/vm/reclaimable.py`** (line 87): Minor: `path.is_symlink() or not path.is_file()` issues two extra syscalls after the lstat already ran (plus a TOCTOU window). `stat.S_ISREG(st.st_mode)` on the lstat result is race-free and cheaper. Behavior is unchanged either way, so this is purely a nit.

**`src/aleph/vm/agent/capacity.py`** (line 224): Micro-optimization: `held_volumes` unconditionally walks the pools via `existing_volume_files` even when `volumes` is empty (a program with no persistent volumes — the common on-demand case). An early return for an empty `volumes` list would skip the per-pool `is_dir()` stats on every cold start.

foxpatch-aleph (approved):
PR B correctly consolidates admission into `check_message`, which admits disk, memory and vCPUs from the message before any byte is allocated, and removes the four post-download disk_mib=0 blocks #1153 left on the create paths. The per-volume discount is capped at the declared size, matched by on-disk filename (suffix included) and consumed so one file discounts at most one volume; stale files never discount; and the largest volume's held bytes are credited only to the pool that holds its file. The `_checked_namespace` guard before the bare path join closes the ../.. discount inflation, and the V-PROGRAM instance bucketing matches `_committed_resources`. Tests pass (65 in the new/extended files; 3 setup errors pre-existing on base). The migration runner still admits after its own overlay download and the volume-name-to-filename convention is duplicated across three modules, both documented as known follow-ups.

**`src/aleph/vm/agent/capacity.py`** (line 295): `held_mib` floors the *total* held bytes to MiB, which keeps slightly more than flooring each per-volume cap would (floor of a sum >= sum of floors). That makes the discount slightly larger than per-volume flooring, though still sub-MiB and conservative relative to the declared total. If you want strictly fail-closed rounding, floor each `HeldVolume.size_bytes` to MiB in `held_volumes` instead.

**`src/aleph/vm/agent/capacity.py`** (line 302): When two declared volumes share the same largest size, `max(...)` picks the first index, so only that volume's held file gets the per-pool credit and the other equal-sized held volume gets none. This is conservative (under-admits, never over-admits), but worth noting since the first match wins by message order rather than by which file is larger.

foxpatch-aleph (approved):
Sound PR: one message-driven admission path (check_message) replaces _admit and the four post-build disk_mib=0 re-checks, with a per-volume discount matched by on-disk filename and capped at the declared size, and the largest-volume credit credited only to the pool holding the file. The dual accounting is physically consistent (aggregate subtraction vs per-pool credit), fail-closed on stale files, colliding stems and unpausible hashes. Verified naming conventions against storage/downloader, bucketing against _committed_resources, and acknowledged migration follow-up. Tests pass (only pre-existing environment failures).

**`src/aleph/vm/agent/capacity.py`** (line 110): A volume name containing `/` (allowed by the match regex, mirroring get_volume_path) yields filenames like "a/b.ext4", whose file sits in a subdirectory that existing_volume_files never sees — so such a volume never discounts. Fail-closed and conservative, but worth a note alongside the rootfs follow-up in the known follow-ups.

foxpatch-aleph (approved):
This PR is correct and unusually well-defended: the per-volume filename matching, one-file-one-volume consumption, declared-size capping, and per-pool credit (rather than globally shrinking max_volume_mib) all hold up under the adversarial scenarios the tests encode, and I verified the duplicated name-sanitization logic in _volume_stems matches storage.get_volume_path exactly today. Namespace validation before the path join, symlink/marker exclusion, and keying matched files by full name (suffix included) close the security-relevant holes; a discount can never exceed what the message declares, so admission cannot be relaxed beyond the declared requirement. All create paths now admit once before any download, the migration runner remains compatible with the extended check_capacity signature, and the tests for the touched files pass locally. Two minor observations follow as comments; neither blocks.

**`src/aleph/vm/agent/capacity.py`** (line 66): Non-blocking observation: because BOOT_DISK_SUFFIXES spans both formats, a recreate whose message (or INSTANCE_DEFAULT_HYPERVISOR setting) switches hypervisor will match the old-format rootfs.btrfs / rootfs.qcow2 and discount it, even though the new create allocates a fresh file in the other format while the old one is retained (retire RECREATE keeps disks). The under-admission is bounded by one rootfs on a nearly-full host, and the effective hypervisor is derivable from content.environment.hypervisor or settings if you ever want to narrow the match — but as a fail-closed-leaning edge case it seems fine to leave with the other naming-convention follow-ups.

**`src/aleph/vm/agent/capacity.py`** (line 450): This comment is now stale after the PR: run.py no longer passes is_instance=True explicitly — check_message derives it via isinstance on InstanceContent | VerifiableProgramContent. The reasoning is still right, but the parenthetical could be updated to reference check_message so a future reader does not go looking for a caller that passes the flag.

**`src/aleph/vm/agent/capacity.py`** (line 193): Nit: existing_volume_files does synchronous directory walks on the event loop, while the neighbouring vm_has_volumes snapshot is wrapped in asyncio.to_thread. The scan is small (one namespace dir per pool) and check_capacity already blocks on statvfs plus the reclaimable walk, so this is marginal — but to_thread would be more consistent if this ever grows.

foxpatch-aleph (commented):
This PR is correct and carefully engineered: the single `check_message` admission path replaces the split pre/post-download checks on all four create paths, and the per-volume discount is fail-closed everywhere it needs to be (per-volume matching rather than directory sums, credit capped at declared size, one file discounts one volume, `_checked_namespace` blocks traversal, credit confined to the pool holding the file, admission ordered after `creating()` adopts the retained directory). The volume-filename derivation is faithful to the storage layer, the `is_instance` bucketing matches both what run.py previously passed explicitly and what `_committed_resources` counts, and the tests cover the right counter-examples end to end. The only real gap is documentation: `docs/architecture/vm-lifecycle.md` still describes the pre-PR behavior this PR deletes. Two non-blocking observations on the per-pool credit corner and the sanitization duplication are noted below.

**`docs/architecture/vm-lifecycle.md`** (line 110): The "Spec admission and capacity" section is now stale and contradicts this PR in several specific, load-bearing claims: (1) line 110 says admission "runs before every `create_vm` call, after the spec is built" — it now runs before the build, ahead of any download; (2) lines 119-121 say every create-path caller "passes `disk_mib=0` and no `max_volume_mib`, so `check_capacity`'s disk checks no-op on create: disk is not actually gated at create time" — disk is now gated at create time, from the message; (3) lines ~124-127 say the only caller exercising the disk checks with real figures is the `/control/reserve_resources` dry run, and line ~133 concludes "disk admission is a dry-run-only feature of the same `CapacityManager`" — now every create path exercises them via `check_message`. Since the whole point of this PR is flipping that invariant, the doc should be updated in the same change; it also still names `check_capacity` as the entry point where `check_message` is now the single admission path.

**`src/aleph/vm/agent/capacity.py`** (line 487): Corner-case observation, not a bug: the per-pool check judges the largest *declared* volume with its own credit, so when that volume is held, its credit can make its pool the "roomiest" and let a fresh, slightly-smaller volume escape per-pool scrutiny. Example: pool A has 5 GiB free and holds the 20 GiB file of a fully-held volume, pool B has 15 GiB free, and the message also declares a fresh 18 GiB volume — pool A's room is counted as 25 GiB, the check passes, and the fresh volume only fails later at `select_pool` (after the download, which is what the PR tries to save). A tie between the largest held volume and a fresh volume has the same shape. No over-commit results — `select_pool` enforces real free space — so this is a lost-early-refusal corner, worth a comment at most. If you ever tighten it, the natural fix is to run the per-pool check over the largest *unheld* volume (still passing the held one's credit when it is the largest).

**`src/aleph/vm/agent/capacity.py`** (line 110): The sanitization regex in `_volume_stems` is a faithful but unlinked copy of `storage.get_volume_path`'s (and this module also reaches into `purge` for the private `_checked_namespace`). If either copy drifts, a volume silently stops discounting (fail-closed, but a recreate gets refused for space it already holds) or starts discounting a file it shouldn't. The PR description already lists a shared volume-name-to-filename helper as a follow-up — agreed, and it would also absorb the suffix tables (`BOOT_DISK_SUFFIXES` / `VOLUME_SUFFIXES`) that are likewise duplicated from the storage layer.

foxpatch-aleph (approved):
This PR is careful, well-documented, and well-tested. The single-admission-path consolidation is correct: all four create paths now admit from the message before any download, and the removal of the post-download disk_mib=0 re-checks is safe because memory/vCPU/GPU semantics are preserved (GPU resolution intentionally stays after the build). The re-create discount is sound: I verified against the actual create paths (create_ext4, create_volume_file, _make_writable_volume) that existing files are reused rather than re-allocated, that per-volume filename matching (including suffixes and both sanitization spellings) matches the real naming conventions, and that the per-pool credit is consistent with sticky placement (both existing_volume_files and find_existing_volume prefer the first pool in order). Security posture is good: _checked_namespace blocks path traversal into the discount, symlinks are skipped and re-checked in file_size_bytes, discounts are capped at declared size, and one file discounts at most one volume. Tests are thorough, including the reviewer counter-examples (stale files, colliding stems, per-pool credit). All findings below are non-blocking comment-level observations; the only behavior gap (unnamed persistent volumes never discount) is fail-closed and availability-only.

**`src/aleph/vm/agent/capacity.py`** (line 143): An unnamed persistent volume never discounts its file: PersistentVolume.name is optional (None), and host_volumes_from_message fills unnamed_volume_{i} only during the build, i.e. after admission, so declared_volumes sees name == "" and stems == (). A recreate of such a VM is charged the full declared size even though its unnamed_volume_N.ext4 file exists and will be reused -- the exact over-refusal this PR fixes, in a corner. Fail-closed (conservative refusal, never over-admission), same class as the documented volume-named-rootfs case, but worth a docstring note in declared_volumes so the next reader does not rediscover it.

**`src/aleph/vm/agent/capacity.py`** (line 108): _volume_stems duplicates the sanitize regex (^[\w\-_/]+$ and re.sub(r"[^\w\-_]", "_")) from storage.get_volume_path. This is already called out as a known follow-up in the PR description; flagging it because a silent divergence between the two spellings would break matching exactly here, and a shared helper would remove the drift risk. Same for the .qcow2/.ext4/.btrfs suffix tuples duplicated from downloader/storage conventions.

**`src/aleph/vm/agent/capacity.py`** (line 293): check_message now does blocking filesystem I/O (iter_namespace_dirs + iterdir + lstat per entry) synchronously in the async create paths, while the adjacent vm_has_volumes call is wrapped in asyncio.to_thread. Bounded by files-per-directory so it is cheap in practice, but worth a note or a to_thread wrap for consistency.

**`src/aleph/vm/agent/capacity.py`** (line 294): Minor: declared_mib and volumes[largest].size_mib duplicate requirements.disk_mib and requirements.max_volume_mib, which requirements_from_message(content, volumes) already computed. Reading them off `requirements` would keep a single source of truth for the derived figures.

foxpatch-aleph (approved):
This PR is correct and carefully engineered. The single admission path (check_message) is sound: I verified the discount arithmetic against the actual filesystem state (creating() adopts the retained directory before admission, so held bytes are neither free nor reclaimable, and the per-volume discount correctly avoids double-counting), confirmed the filename conventions in BOOT_DISK_SUFFIXES/VOLUME_SUFFIXES against every actual volume creator (downloader._make_writable_volume, storage.get_volume_path, create_volume_file, including the sanitized/unsanitized name split), and traced the per-pool credit logic (one file discounts one volume, capped at declared size, credited only to the pool holding it). No over-admission vector exists via message-controlled data, and the namespace is validated before the bare path join. The test suite encodes the adversarial counter-examples (stale files, colliding stems, per-pool credit) and the stubs deliberately fail loudly if a direct check_capacity call reappears. Remaining items are acknowledged follow-ups or nits, none blocking.

**`src/aleph/vm/agent/capacity.py`** (line 138): This sanitization regex is a duplicate of the one in storage.get_volume_path (storage.py:493-496). If the storage-side sanitization ever changes, admission's discount silently diverges from the actual on-disk name. You acknowledge this in the PR description as a follow-up — agreed, but worth tracking with an issue since it is a security-adjacent invariant (discount correctness depends on matching the real filename).

**`src/aleph/vm/agent/capacity.py`** (line 29): Importing _checked_namespace, a private symbol, across modules is a small smell — worth promoting it (and ROOTFS_STEM) to a non-underscore name in purge.py, or moving it to a shared location, given capacity now depends on it as a security check.

**`src/aleph/vm/agent/capacity.py`** (line 103): Worth a line in the DeclaredVolume docstring: a persistent volume with no name never receives a discount, because host_volumes_from_message only assigns the unnamed_volume_{i} name at download time, after admission. Fail-closed, so not a bug, but it is one more convention a reader would otherwise have to rediscover.

**`src/aleph/vm/agent/capacity.py`** (line 197): existing_volume_files runs blocking filesystem I/O synchronously inside check_message, which every async create path calls on the event loop. This is consistent with check_capacity's existing pools_disk_usage/reclaimable_bytes calls, so not blocking — but note run.py uses asyncio.to_thread for the adjacent vm_has_volumes scan, so the codebase is split on this pattern.

**`src/aleph/vm/agent/migration/runner.py`** (line 349): Acknowledge in the PR description, but flagging for visibility: the migration import still calls check_capacity with disk_mib=0 after its own overlay download — the exact shape this PR eliminates elsewhere, so a migrated-in VM skips disk admission entirely. Worth an issue so it is not forgotten.

foxpatch-aleph (approved):
Correct and carefully engineered. Admission now happens from the message before any download, replacing the four post-download no-op checks without losing coverage. The per-volume discount is conservative on every edge I checked: matching is per volume-by-volume (never a directory sum, so stale/renamed files discount nothing), each file is consumed once, hits are capped at the declared size, and the largest-volume credit is only applied to the pool actually holding the file. The discount's premise holds (the create path returns early on existing volume files rather than re-allocating), and there is no double-credit with reclaimable_bytes because creating() adopts the retained directory before check_message runs inside the guard. Path injection into the discount is closed via _checked_namespace before the pool-path join. Tests are thorough (59 in test_agent_capacity.py, all touched suites green) and the remaining full-suite failures are pre-existing root/directory-permission environment issues. Only nits: a now-inaccurate _check_against docstring, an over-long rewrapped docstring line in run.py, and a minor missed early-return in held_volumes.

**`src/aleph/vm/agent/capacity.py`** (line 402): Docstring nit: the claim that one shared implementation means "an advisory answer can never be stronger or weaker than the enforced one" is no longer accurate. Since simulate() never passes max_volume_credit (nor the held-volume discount), the advisory answer can now be strictly weaker — more conservative — than the enforced one for a recreate whose largest volume file already exists. That's the safe direction, but the sentence should be softened or the simulate() gap noted.

**`src/aleph/vm/agent/run.py`** (line 452): Nit: this rewrapped docstring line runs well past the file's usual wrap width; worth a re-wrap pass.

**`src/aleph/vm/agent/capacity.py`** (line 227): Nit: held_volumes walks every pool directory via existing_volume_files() even when the message declares no volumes (the common on-demand program path), only to build an empty list. An early return for an empty volumes list would skip the filesystem walk on a hot path.

foxpatch-aleph (approved):
This PR is correct and rigorously tested. The admission logic moves to a single message-driven path (check_message) before any download, and the recreate discount is fail-closed in every direction I probed: per-volume (not per-directory) matching, one file discounts at most one volume via pop(), each hit capped at the declared size, stale files and symlinks never discount, and the per-pool credit goes only to the pool holding the file. The filename conventions (BOOT_DISK_SUFFIXES, VOLUME_SUFFIXES, _volume_stems sanitization) accurately mirror storage.py and downloader.py, and ImmutableVolume's lack of name/size_mib is handled by the getattr defaults. The creating()-adopts-before-check ordering that prevents double-crediting retained bytes is pinned by tests. All modified test files pass locally. Remaining items are minor: the migration import path still performs no disk admission (acknowledged follow-up), check_message does blocking filesystem I/O on the event loop where adjacent calls use asyncio.to_thread, and check_message recomputes disk_mib/max_volume_mib that requirements_from_message just derived.

**`src/aleph/vm/agent/migration/runner.py`** (line 349): This remains the one create path that downloads first and never checks disk at all (disk_mib=0 after the overlay is already on disk, so even the aggregate and per-pool guards are skipped). The PR description lists it as a known follow-up, and check_capacity's docstring now preserves the scalar seam for exactly this caller — but since the overlay download here is preceded by nothing, calling capacity.check_message(message.content, exclude_vm_hash=job.vm_hash) before build_create_vm_spec would close the gap with the same shape as the fixed paths.

**`src/aleph/vm/agent/capacity.py`** (line 314): held_volumes() walks the VM's namespace directory across every pool synchronously on the event loop (check_message is called directly by the async create paths), while the adjacent vm_has_volumes calls at these same call sites are wrapped in asyncio.to_thread. The directory is small so this is a nit, but the to_thread convention exists at the same call sites for the same kind of I/O.

**`src/aleph/vm/agent/capacity.py`** (line 325): Minor redundancy: requirements_from_message(content, volumes) just computed disk_mib and max_volume_mib from the same volumes list, and this call ignores those two fields and recomputes them (declared_mib - held_mib, volumes[largest]). Harmless today since both derive from the same list, but a future change to one derivation could silently diverge from the other. Using requirements.disk_mib as the base for the subtraction would remove the duplication.

**`src/aleph/vm/agent/capacity.py`** (line 481): Note that simulate() cannot apply the held-volume discount (candidates are judged from ResourceRequirements scalars, and it has no vm_hash-scoped filesystem lookup), so an advisory verdict is now strictly stricter than the enforced path for a recreate whose volumes are already on disk. That is consistent with the stated rule that an advisory answer never be stronger than the enforced one, but it means a plan could refuse a recreate the create path would admit — worth keeping in mind if/when simulate() gains production callers.

foxpatch-aleph (approved):
Solid, well-tested PR: admission now happens once, from the message, before any download, on all four create paths, with a per-volume discount that is matched (not summed), capped at declared sizes, and credited only to the pool holding the file. I verified the suffix tables against the actual allocation paths (_make_writable_volume, create_volume_file, get_volume_path), confirmed the creating()-adoption ordering prevents double-counting held bytes as reclaimable-free, validated the discount logic with a standalone script against real files, and ran the full admission-related test sweep (156 passed; only pre-existing environment failures remain). The findings below are non-blocking edge cases, two of which the PR description already discloses as follow-ups.

**`src/aleph/vm/agent/capacity.py`** (line 248): The 'stale files never discount' invariant holds for renamed/dropped volumes but not retyped ones: a declared volume matches any of its suffix spellings, so a leftover data.qcow2 from an old code path still discounts a create that will allocate data.ext4 fresh, relaxing admission by up to that volume's declared size. Since the message hash pins its content, the realistic trigger is a node upgrade changing which suffix a given message's volume takes (or the migration importer staging differently-named files). Bounded and narrow, but the held_volumes/DeclaredVolume docstrings could state this residual case explicitly.

**`src/aleph/vm/agent/capacity.py`** (line 253): Capping the credit at allocated bytes means a recreate of a sparse volume (e.g. a fresh qcow2 overlay with few allocated blocks) can be refused by _check_max_volume on a nearly-full pool even though the create would reuse the existing file and allocate nothing. Fail-closed, so fine, but a sentence noting the credit is deliberately allocated-bytes (rather than declared size, which the reuse would justify) would help future readers wondering why.

**`src/aleph/vm/agent/migration/runner.py`** (line 349): Acknowledged in the PR description as a known follow-up, but worth tracking: this is the one remaining caller that admits after its own download with disk_mib=0, i.e. exactly the shape this PR fixes on the create paths. The migration download can consume disk the admission never checked.

foxpatch-aleph (approved):
High-quality PR that cleanly unifies admission into a single message-driven check_message path. The discount logic is careful and internally consistent (per-volume matching with file consumption, declared-size caps, pool-specific credit for the largest volume), the exclude_vm_hash registry coupling and path validation are sound, and every create path including the migration import and reserve endpoint is covered. The removed post-build re-checks closed no real race (the program path never had a registry claim before admission anyway), so the simplification is safe. The test suite is exemplary: it pins the derived scalars, exercises the adversarial counter-examples the docstrings warn about, and makes stubs fail loudly on a reintroduced direct check_capacity call. Remaining nits are comment-level.

**`src/aleph/vm/agent/capacity.py`** (line 330): Minor API-clarity nit: exclude_vm_hash is now doing double duty — it excludes the VM's own registry record from the committed sums AND serves as the namespace key for the on-disk discount lookup in held_volumes. Every current caller passes the same hash for both meanings, but the coupling is implicit; a future caller passing a hash purely for record-exclusion (e.g. a phantom-record retry) would silently get a filesystem discount keyed on that hash. A one-line note in the docstring ("the same hash identifies the directory whose files may discount") would make the contract explicit.

**`src/aleph/vm/agent/capacity.py`** (line 145): This sanitization regex duplicates storage.get_volume_path's; the two must stay in lock-step or a volume silently stops matching its file (fail-closed, but a discount silently lost). Already flagged as a known follow-up in the PR description — just confirming the shared-helper extraction is worth prioritizing.

**`src/aleph/vm/agent/capacity.py`** (line 715): Worth a note here: the credit only ever applies to eligible pools, so a pure recreate whose largest volume's file sits on a pool that is no longer vm_eligible (or is unreachable) is refused by _check_max_volume even though nothing new would be allocated. Fail-closed is the right default, but it is currently undocumented; a sentence in the docstring would help.

**`src/aleph/vm/agent/capacity.py`** (line 221): existing_volume_files does synchronous directory I/O inline on the event loop in the async create paths, where neighboring work (vm_has_volumes, select_pool) is wrapped in asyncio.to_thread. The walk is bounded ({pool}/{vm_hash} on a few pools) and consistent with the pre-existing sync reclaimable_bytes() in check_capacity, so not a blocker — just something to keep in mind if pool counts grow.

PR #1157, opened 2026-08-25
refactor(agent): extract the allocation teardown and its removability rule

odesenfans:
First of a stack implementing the async scheduler allocations design (`docs/plans/2026-08-20-async-allocations-design.md`). Behaviour-neutral on its own.

The stop half of `update_allocations` is an inline composite (supervisor delete, registry forget, DB records, two staging removers) guarded by a predicate whose v-program branch inverts every other rule: a v-program is stopped when the plan drops it *because* the scheduler is its single source of truth, even though it is credit-paid and confidential. The v2 reconciler (later in this stack) needs exactly that behaviour, so it becomes one tested module rather than a copy left to drift.

## Changes

- `agent/allocation/teardown.py`: `is_removable_by_allocation(record, info)` and `teardown_vm(vm_hash, *, supervisor, registry)`. The teardown is idempotent: a VM the supervisor has already forgotten still gets its agent-side state cleaned, because that state (registry entry, DB rows, staging dirs) is ours and would otherwise leak.
- `update_allocations` calls both; the four imports that moved with the code are dropped from `views/__init__.py`.
- Three tests patched `aleph.vm.agent.views.delete_records_for_vm`; the symbol moved, so their patch targets follow it. The assertions are unchanged.

## How to test

`pytest tests/supervisor/test_allocation_teardown.py` (9 new tests, written before the implementation).

Behaviour-neutrality is the real check: the whole `tests/supervisor` FAILED set is identical to `dev`.

## Notes

- ~105 pre-existing failures in my environment (`PermissionError: /var/cache/aleph`); the FAILED set is diffed against a `dev` worktree rather than counted.
- mypy unchanged at its 48-error baseline.
- Stack: this -> #verify -> #simulate -> plan/verdict -> reconciler -> v2 endpoints -> status surface.

### DIFF
--- a/src/aleph/vm/agent/allocation/__init__.py
+++ b/src/aleph/vm/agent/allocation/__init__.py
@@ -1,0 +1,1 @@
+"""Scheduler allocation: the plan the CRN was handed, and converging to it."""
--- a/src/aleph/vm/agent/allocation/teardown.py
+++ b/src/aleph/vm/agent/allocation/teardown.py
@@ -1,0 +1,56 @@
+"""Stopping a VM because the scheduler's plan no longer lists it.
+
+Removability is not uniform. A VM the user pays for directly (a payment stream
+or credits), one holding GPUs, or a confidential one is retained: the scheduler
+is not the authority on those. A v-program inverts that, because the scheduler
+IS its single source of truth, and it is credit-paid and confidential by
+construction.
+
+Tearing down is a composite: the supervisor owns the VM, the agent owns the
+registry record, the DB rows and the staging directories, so all of them have
+to go.
+"""
+
+import logging
+
+from aleph_message.models import ItemHash
+
+from aleph.vm.agent.metrics import delete_records_for_vm
+from aleph.vm.agent.snp_instance_launch import remove_snp_instance_staging
+from aleph.vm.agent.vm_registry import AgentVmRecord, AgentVmRegistry
+from aleph.vm.agent.vprogram_launch import remove_vprogram_staging
+from aleph.vm.supervisor_interface.abc import Supervisor
+from aleph.vm.supervisor_interface.errors import VmNotFoundError
+from aleph.vm.supervisor_interface.types import ConfidentialMode, VmId, VmInfo
+
+logger = logging.getLogger(__name__)
+
+
+def is_removable_by_allocation(record: AgentVmRecord, info: VmInfo) -> bool:
+    """Whether an allocation push may stop this VM when the plan drops it."""
+    if not record.persistent:
+        return False
+    if record.is_vprogram:
+        return True
+    return (
+        not record.uses_payment_stream
+        and not record.uses_payment_credit
+        and not info.gpus
+        and info.confidential_mode is ConfidentialMode.NONE
+    )
+
+
+async def teardown_vm(vm_hash: ItemHash, *, supervisor: Supervisor, registry: AgentVmRegistry) -> None:
+    """Delete the VM and every piece of agent-side state that belongs to it.
+
+    Idempotent: a VM the supervisor has already forgotten still has its agent
+    state cleaned, since that state is ours and would otherwise leak.
+    """
+    try:
+        await supervisor.delete_vm(VmId(str(vm_hash)))
+    except VmNotFoundError:
+        logger.info("Supervisor no longer knows %s; cleaning agent state anyway", vm_hash)
+    registry.forget(vm_hash)
+    await delete_records_for_vm(str(vm_hash))
+    remove_vprogram_staging(vm_hash)
+    remove_snp_instance_staging(vm_hash)
--- a/src/aleph/vm/agent/views/__init__.py
+++ b/src/aleph/vm/agent/views/__init__.py
@@ -22,11 +22,11 @@ from pydantic import ValidationError
 from aleph.vm import haproxy
 from aleph.vm.agent import payment, status
 from aleph.vm.agent.aggregate import update_aggregate_settings
+from aleph.vm.agent.allocation.teardown import is_removable_by_allocation, teardown_vm
 from aleph.vm.agent.capacity import CapacityManager, requirements_from_message
 from aleph.vm.agent.custom_logs import set_vm_for_logging
 from aleph.vm.agent.haproxy_sync import sync_domain_mappings
 from aleph.vm.agent.messages import try_get_message
-from aleph.vm.agent.allocation.teardown import is_removable_by_allocation, teardown_vm
 from aleph.vm.agent.metrics import get_execution_records
 from aleph.vm.agent.node_identity import NodeIdentity
 from aleph.vm.agent.payment import (
@@ -72,12 +72,7 @@ from aleph.vm.supervisor_interface.errors import (
     SupervisorError,
     VmNotFoundError,
 )
-from aleph.vm.supervisor_interface.types import (
-    PortForwardInfo,
-    VmId,
-    VmInfo,
-    VmStatus,
-)
+from aleph.vm.supervisor_interface.types import PortForwardInfo, VmId, VmInfo, VmStatus
 from aleph.vm.utils import (
     HostNotFoundError,
     b32_to_b16,
--- a/src/aleph/vm/agent/views/__init__.py
+++ b/src/aleph/vm/agent/views/__init__.py
@@ -26,7 +26,8 @@ from aleph.vm.agent.capacity import CapacityManager, requirements_from_message
 from aleph.vm.agent.custom_logs import set_vm_for_logging
 from aleph.vm.agent.haproxy_sync import sync_domain_mappings
 from aleph.vm.agent.messages import try_get_message
-from aleph.vm.agent.metrics import delete_records_for_vm, get_execution_records
+from aleph.vm.agent.allocation.teardown import is_removable_by_allocation, teardown_vm
+from aleph.vm.agent.metrics import get_execution_records
 from aleph.vm.agent.node_identity import NodeIdentity
 from aleph.vm.agent.payment import (
     InvalidAddressError,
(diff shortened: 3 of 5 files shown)


foxpatch-aleph (approved):
A clean, behaviour-neutral extraction of the allocation teardown logic into a dedicated module. The removability predicate and teardown composite are faithfully lifted from the inline code in update_allocations, with the v-program inversion preserved and well-documented. The only behavioural change is logging VmNotFoundError at INFO instead of silently swallowing it, which is an improvement. Test coverage is thorough: 7 predicate tests cover every branch including the v-program inversion, and 2 teardown tests verify both the happy path and the idempotent supervisor-not-found path. The three existing test files correctly updated their patch targets to follow the moved symbol. No correctness, security, or quality concerns.

**`src/aleph/vm/agent/allocation/teardown.py`** (line 52): Minor improvement over the original inline code, which used a bare `except VmNotFoundError: pass`. Logging at INFO here is the right call — a supervisor that has already forgotten a VM is an expected but noteworthy condition during reconciliation. Worth calling out in the PR description as an intentional (minor) behavioural change in observability, even though the logic is neutral.

**`tests/supervisor/test_allocation_teardown.py`** (line 22): Using `SimpleNamespace` instead of the real `AgentVmRecord` and `VmInfo` dataclasses means these tests won't catch type signature drift (e.g., if `is_vprogram` were renamed on the real dataclass). This is a common trade-off for unit tests and is acceptable here since the integration tests in test_views.py and test_vprogram.py exercise the real types through the endpoint. Just noting for awareness.

foxpatch-aleph (approved):
A clean, genuinely behaviour-neutral extraction. The removability predicate in is_removable_by_allocation is logically identical to the original inline guard (including the deliberate v-program inversion), teardown_vm preserves the original cleanup sequence and exception handling with only a benign added info-log in the VmNotFoundError branch, and all moved symbols/imports were verified to match their real signatures with no leftover references in views/__init__.py. The 9 new tests are well-targeted (I re-verified both async teardown paths directly since the sandbox's outdated pytest-asyncio skipped them; test_vprogram.py passes 15/15 including the re-pointed patch target), and the 40 remaining supervisor-suite failures are all environmental (/var/lib/aleph unwritable) in files this PR does not touch. Two trivial test nits below do not block.

**`tests/supervisor/test_allocation_teardown.py`** (line 31): The `status` parameter of `_info()` is never varied by any test — it's dead test surface. Either drop it or add a case exercising a non-RUNNING VM (e.g., `test_a_stopped_vm_is_not_touched`), which would mirror how the caller guards on `VmStatus.RUNNING`.

**`tests/supervisor/test_allocation_teardown.py`** (line 94): The docstring says "the registry entry, the DB rows and the staging dirs are ours", but the test only asserts `registry.forget`. Keeping references to the mocked `delete_records_for_vm` / staging removers and asserting them (as the first test does) would make the assertions match the stated intent — otherwise a regression dropping those three calls would still pass here.

foxpatch-aleph (approved):
A clean, faithful extraction. The predicate is truth-table identical to the inline condition it replaces, and teardown_vm preserves the exact teardown sequence with only one benign addition (an INFO log on the previously silent VmNotFoundError path). No stale imports, patch targets, or import-cycle risks remain; the agent/supervisor import boundary is respected. All 9 new tests pass and correctly document the v-program inversion that motivated naming the predicate. Minor observations only: tasks.py:404-408 still holds a byte-for-byte duplicate of the composite (a natural follow-up consumer once the stack lands, and it lacks the VmNotFoundError tolerance teardown_vm now has), the PR description's 'three tests patched' claim overstates it (only tests/supervisor/test_vprogram.py:171 needed retargeting), and the SimpleNamespace test doubles bypass the real AgentVmRecord property computations, which is acceptable isolation for a predicate test.

**`src/aleph/vm/agent/tasks.py`** (line 404): This composite is byte-for-byte the same teardown teardown_vm now encapsulates, except it lacks the VmNotFoundError tolerance — delete_vm raising here would strand the registry record, DB rows and staging dirs. Worth folding into teardown_vm as a follow-up once the rest of the stack lands, so the two cannot drift.

**`tests/supervisor/test_allocation_teardown.py`** (line 21): Note: the SimpleNamespace doubles bypass the real AgentVmRecord properties (uses_payment_stream/is_vprogram inspect message shape), so these tests pin the predicate but not the property computations. That's fine isolation for a predicate test — the real record is exercised end-to-end by test_vprogram.py's stops/spares pair — just flagging the trade-off.

**`src/aleph/vm/agent/allocation/teardown.py`** (line 52): Nice improvement over the silent `pass` this replaces. One nit: the original swallowed the exception entirely, so this is the only observable behavior delta in the PR — worth a line in the PR description's behaviour-neutrality claim, since it is a log-emission difference rather than a state one.

foxpatch-aleph (approved):
A clean, faithful extraction. The predicate is logically identical to the inline condition it replaces (the persistent-check reordering is safe, all conditions being pure), teardown_vm preserves the original teardown sequence with only an added INFO log on VmNotFoundError, and the moved import targets in the three affected tests were correctly repointed with no stale patch targets remaining. Signatures at every call site were verified against the real modules, no circular imports exist, and all 9 new tests pass (the aiohttp_client errors in test_vprogram.py are environmental - pytest-aiohttp is missing from this sandbox). The only nits are cosmetic: the removability tests use SimpleNamespace stand-ins rather than real AgentVmRecord/VmInfo types, so they would not catch attribute-name drift on the real classes, and the pyproject.toml comment realignment is unrelated to this change.

**`tests/supervisor/test_allocation_teardown.py`** (line 21): Minor: `_record`/`_info` use SimpleNamespace stand-ins, so these tests won't catch attribute-name drift on the real `AgentVmRecord`/`VmInfo` (e.g. a rename of `uses_payment_credit`). Constructing real types in the helper would make the predicate's contract harder to break; fine as a trade-off for a pure predicate, though.

**`pyproject.toml`** (line 41): This whitespace-only realignment of the grpcio comment is unrelated to the teardown extraction (rebase artifact). Harmless, but if the PR is meant to stay behaviour-neutral and focused you might drop it.

foxpatch-aleph (approved):
A clean, faithful extraction. The predicate and composite teardown in src/aleph/vm/agent/allocation/teardown.py are logically identical to the code they replace in update_allocations, including operation order and the VmNotFoundError swallow; the only delta is a new INFO log on the already-swallowed path, which is an improvement. The module docstring does a good job explaining the non-obvious v-program inversion. The 9 new tests pass and assert the right things (I ran them); the retargeted patch in test_vprogram.py is correct and no dangling patch targets remain repo-wide. Import pruning in views is correct. Non-blocking nits: teardown tests assert each collaborator is called but not their order, though remove_vprogram_staging's contract requires it to run after supervisor delete and registry.forget; no test that a non-VmNotFoundError from delete_vm propagates; and the PR description's 'three tests' claim doesn't match the single retargeted patch in the diff.

**`src/aleph/vm/agent/allocation/teardown.py`** (line 52): This INFO log is the only observable behavior change in an otherwise behaviour-neutral PR (the original was a silent `pass`). It's a good change — just flagging it since the PR description promises neutrality.

**`tests/supervisor/test_allocation_teardown.py`** (line 76): The two teardown tests assert each collaborator fires but not the order, while `remove_vprogram_staging`'s docstring (vprogram_launch.py:79) explicitly requires staging removal to run "after the supervisor delete and registry.forget". Appending to a shared list in each mock and asserting the sequence would protect that contract. Also worth a test that a non-VmNotFoundError from delete_vm propagates rather than being swallowed.

**`tests/supervisor/test_allocation_teardown.py`** (line 31): Minor: `_info`/`_record` build SimpleNamespaces where the real `VmInfo`/`AgentVmRecord` dataclasses might be usable, which would keep the tests honest if the shapes change. Fine as-is given `VmInfo`'s required fields, just noting the duplication between the two TestTeardown tests could fold into a small fixture.

foxpatch-aleph (approved):
A clean, genuinely behaviour-neutral extraction. I verified the predicate is boolean-equivalent to the original inline condition (all operands are side-effect-free property accesses, so the reordered and-chain is safe), the teardown composite preserves the exact call sequence with matching signatures, the four imports removed from views/__init__.py are truly unused there, and no import cycles are introduced. All 9 new tests pass, and the test_vprogram.py failures are pre-existing environment issues (installed pytest 7.2.1/pytest-asyncio 0.20.3 vs pinned 8.2.1/0.23.7) — I confirmed the identical failure set at the parent commit in a worktree. The only behaviour delta is a benign INFO log replacing a silent `pass` on the VmNotFoundError path. Remaining items are non-blocking: a pre-existing duplicate of the teardown composite in tasks.py that has already drifted (missing the VmNotFoundError guard), an untested-but-load-bearing edge in the predicate (non-persistent v-program), and an design doc referenced in the description that is absent from the branch.

**`src/aleph/vm/agent/tasks.py`** (line 404): This block is a pre-existing duplicate of the exact teardown composite extracted here — and it has already drifted: unlike the original update_allocations inline block, its `delete_vm` is not guarded by `VmNotFoundError`. Since the PR's stated goal is preventing the composite from drifting, this is the obvious consolidation candidate for a later PR in the stack. Note that swapping in `teardown_vm` is not behaviour-neutral for this call site (it swallows VmNotFoundError and adds a log), so the swap needs a deliberate decision about which semantics tasks.py should have.

**`tests/supervisor/test_allocation_teardown.py`** (line 58): The predicate checks `persistent` before `is_vprogram`, so a non-persistent v-program is not removable — that precedence is load-bearing, but no test pins it. Consider adding a `_record(vprogram=True, persistent=False)` case so a future reordering of the two early returns (which would silently change who gets stopped) cannot slip through.

**`tests/supervisor/test_allocation_teardown.py`** (line 32): Minor: `_info()` sets a `vm_id` field nothing reads — the same dead surface the follow-up commit removed when it dropped `status`.

**`src/aleph/vm/agent/allocation/teardown.py`** (line 52): For the record: this INFO log (replacing the original silent `pass`) is the only behaviour delta in the PR. It is an improvement — worth keeping in mind if a strict-neutrality claim is needed for the stack.

PR #1158, opened 2026-08-25
feat(agent): verify messages the scheduler embeds in a plan

odesenfans:
Stacked on #1157. Second of the async-allocations stack: the trust boundary for messages the scheduler will embed in a v2 plan.

Two gates, only meaningful together:

1. `aleph_message`'s own validators enforce `item_hash == sha256(item_content)` and `content == json(item_content)` for inline messages, so parsing binds the content to the hash.
2. The sender's signature covers `chain\nsender\ntype\nitem_hash`, so a valid signature proves the sender authorized exactly that content.

Net effect: the scheduler keeps the authority it already has (choosing *which* user messages run here) and gains none over *what* they contain.

## Changes

- `agent/allocation/verify.py`: `verify_entry(entry) -> (outcome, verified, reason)` with three outcomes. **REJECTED** (bad signature, hash mismatch, non-executable type) is never retried and never falls back to fetching: a silent fallback would turn a tampering attempt into a successful boot of the real message and we would never see it. **UNVERIFIABLE** (no embedded message, non-inline `item_type`, non-EVM chain) degrades to today's fetch path, so exotic senders keep working and only lose the fast path.
- Only EVM signatures are checked. All 200 most recent INSTANCE messages on the network are ETH, so implementing Solana now would be speculation; those entries simply take the fetch path.

## Notes: two things found while writing this

- **The design's `amended_message` field is dropped.** `update_message()` (`agent/messages.py:54`) does not resolve a message-level amend: it refreshes the `use_latest` refs *inside* volumes, rootfs and code to their latest STORE hashes (the amend field on `InstanceContent` is `replaces`, not `ref`). A refreshed ref is a bare hash the agent has nothing to check against, so an embedded amend cannot be verified at all, and trusting one would let the scheduler point a VM's rootfs at content of its choosing. Decision: ignore message-level amends for programs, as the legacy path already does, and keep following `use_latest` for volumes. The question does not arise for instances or v-programs by design. Design doc updated.
- **`tests/supervisor/fixtures/vprogram_message.json` carries a signature that does not verify.** Nothing depends on it today, but any future test pushing it through this module must re-sign it.

## How to test

`pytest tests/supervisor/test_allocation_verify.py` (8 tests, written first). The verification buffer format was confirmed by recovering the signer of five live INSTANCE messages from api3.aleph.im before any code was written.

### DIFF
--- a/src/aleph/vm/agent/allocation/verify.py
+++ b/src/aleph/vm/agent/allocation/verify.py
@@ -2,11 +2,25 @@
 
 Two gates, only meaningful together:
 
-1. aleph_message's own validators enforce item_hash == sha256(item_content)
-   and content == json(item_content) for inline messages, so parsing binds the
-   content to the hash.
+1. aleph_message's validators enforce item_hash == sha256(item_content), so
+   the hash binds the JSON the user actually signed.
 2. The sender's signature covers "chain\\nsender\\ntype\\nitem_hash", so a
-   valid signature proves the sender authorized exactly that content.
+   valid signature proves the sender authorized exactly that JSON.
+
+Those two gates bind item_content, and nothing else. A message also carries a
+separate ``content`` field, and that is the one a caller reads. aleph_message
+does not tie the two together for every type: ProgramMessage and
+VerifiableProgramMessage have a check_content validator, InstanceMessage does
+not. Left alone, a scheduler could therefore hand us a message carrying the
+user's genuine item_content and signature next to a ``content`` of its own
+choosing (more vCPUs, a different rootfs) and it would verify.
+
+So we do not read the ``content`` the scheduler sent: it is replaced with the
+JSON parsed out of item_content before the message is built. A divergent
+``content`` is ignored rather than rejected, because comparing a parsed
+pydantic model against raw JSON invites false rejections over defaults and
+type coercion, while re-deriving cannot be wrong: what comes out is the signed
+bytes or nothing.
 
 The result is that the scheduler picks WHICH user messages run here and never
 WHAT they contain, which is the authority it already had.
@@ -23,6 +37,7 @@ let it point a rootfs at content of its choosing, which is exactly the
 authority this module exists to deny. The agent resolves those refs itself.
 """
 
+import json
 import logging
 from dataclasses import dataclass
 from enum import Enum
@@ -72,10 +87,18 @@ def _signature_matches(message: ExecutableMessage) -> bool:
 def _parse(raw: dict) -> ExecutableMessage:
     """Parse and narrow to an executable message.
 
+    The content is re-derived from item_content rather than taken from the
+    ``content`` the sender of the plan supplied, so the parsed message can only
+    ever hold what the signature covers.
+
     parse_message returns the whole message union, so the type check is also
     what makes the narrowed return type honest.
     """
-    message = parse_message(raw)
+    item_content = raw.get("item_content")
+    if not isinstance(item_content, str):
+        msg = "inline message without item_content"
+        raise ValueError(msg)
+    message = parse_message({**raw, "content": json.loads(item_content)})
     if not isinstance(message, (InstanceMessage, ProgramMessage, VerifiableProgramMessage)):
         msg = f"message type {message.type} is not executable"
         raise ValueError(msg)
--- a/src/aleph/vm/agent/allocation/verify.py
+++ b/src/aleph/vm/agent/allocation/verify.py
@@ -39,6 +39,7 @@ authority this module exists to deny. The agent resolves those refs itself.
 
 import json
 import logging
+from copy import deepcopy
 from dataclasses import dataclass
 from enum import Enum
 
@@ -55,7 +56,11 @@ from eth_account.messages import encode_defunct
 
 logger = logging.getLogger(__name__)
 
-EVM_CHAINS = {"ETH", "AVAX", "BASE", "BSC"}
+# Chains whose signatures we currently recover, not every EVM chain aleph
+# supports. The enum also carries OP, ARB, POL, LINEA and more; a message on
+# one of those is UNVERIFIABLE, which costs a CCN fetch and nothing else, so
+# this set can grow whenever a chain is worth the fast path.
+VERIFIABLE_CHAINS = {"ETH", "AVAX", "BASE", "BSC"}
 
 
 class VerificationOutcome(Enum):
@@ -66,6 +71,13 @@ class VerificationOutcome(Enum):
 
 @dataclass(frozen=True)
 class VerifiedMessage:
+    """A message and an untouched copy of it.
+
+    ``original`` is a deep copy, not the same object: a caller that resolves
+    use_latest refs does so with update_message, which mutates the message in
+    place, and an aliased original would be rewritten along with it.
+    """
+
     message: ExecutableMessage
     original: ExecutableMessage
 
@@ -114,9 +126,14 @@ def verify_entry(entry: dict) -> tuple[VerificationOutcome, VerifiedMessage | No
     raw = entry.get("message")
     if not raw:
         return VerificationOutcome.UNVERIFIABLE, None, ""
+    if not isinstance(raw, dict):
+        # Everything else malformed answers REJECTED; a non-dict must not be
+        # the one shape that escapes with an AttributeError instead.
+        logger.warning("Refusing embedded message for %s: not an object", entry.get("item_hash"))
+        return VerificationOutcome.REJECTED, None, "message is not an object"
     if raw.get("item_type") != ItemType.inline.value:
         return VerificationOutcome.UNVERIFIABLE, None, ""
-    if raw.get("chain") not in EVM_CHAINS:
+    if raw.get("chain") not in VERIFIABLE_CHAINS:
         return VerificationOutcome.UNVERIFIABLE, None, ""
 
     try:
@@ -133,4 +150,4 @@ def verify_entry(entry: dict) -> tuple[VerificationOutcome, VerifiedMessage | No
         logger.warning("Embedded message %s is not signed by its sender", message.item_hash)
         return VerificationOutcome.REJECTED, None, "signature does not match the sender"
 
-    return VerificationOutcome.VERIFIED, VerifiedMessage(message=message, original=message), ""
+    return VerificationOutcome.VERIFIED, VerifiedMessage(message=message, original=deepcopy(message)), ""
--- a/src/aleph/vm/agent/allocation/verify.py
+++ b/src/aleph/vm/agent/allocation/verify.py
@@ -1,0 +1,113 @@
+"""Local verification of a message the scheduler embedded in its plan.
+
+Two gates, only meaningful together:
+
+1. aleph_message's own validators enforce item_hash == sha256(item_content)
+   and content == json(item_content) for inline messages, so parsing binds the
+   content to the hash.
+2. The sender's signature covers "chain\\nsender\\ntype\\nitem_hash", so a
+   valid signature proves the sender authorized exactly that content.
+
+The result is that the scheduler picks WHICH user messages run here and never
+WHAT they contain, which is the authority it already had.
+
+Only EVM signatures are checked. A chain we cannot verify, or a message with
+no inline content, is UNVERIFIABLE rather than REJECTED: the caller falls back
+to fetching it from the CCN, exactly as before this feature existed.
+
+There is deliberately no "amended message" in the payload. What
+``update_message`` resolves is not a message-level amend but the ``use_latest``
+refs inside volumes, rootfs and code, and a resolved ref is just a hash: the
+agent has nothing to check it against. Accepting one from the scheduler would
+let it point a rootfs at content of its choosing, which is exactly the
+authority this module exists to deny. The agent resolves those refs itself.
+"""
+
+import logging
+from dataclasses import dataclass
+from enum import Enum
+
+from aleph_message import parse_message
+from aleph_message.models import (
+    ExecutableMessage,
+    InstanceMessage,
+    ItemType,
+    ProgramMessage,
+    VerifiableProgramMessage,
+)
+from eth_account import Account
+from eth_account.messages import encode_defunct
+
+logger = logging.getLogger(__name__)
+
+EVM_CHAINS = {"ETH", "AVAX", "BASE", "BSC"}
+
+
+class VerificationOutcome(Enum):
+    VERIFIED = "verified"
+    REJECTED = "rejected"
+    UNVERIFIABLE = "unverifiable"
+
+
+@dataclass(frozen=True)
+class VerifiedMessage:
+    message: ExecutableMessage
+    original: ExecutableMessage
+
+
+def verification_buffer(message: ExecutableMessage) -> bytes:
+    """The bytes an aleph sender signs. Confirmed against live network messages."""
+    return f"{message.chain.value}\n{message.sender}\n{message.type.value}\n{message.item_hash}".encode()
+
+
+def _signature_matches(message: ExecutableMessage) -> bool:
+    try:
+        recovered = Account.recover_message(encode_defunct(verification_buffer(message)), signature=message.signature)
+    except Exception:
+        logger.warning("Could not recover the signer of %s", message.item_hash, exc_info=True)
+        return False
+    return recovered.lower() == message.sender.lower()
+
+
+def _parse(raw: dict) -> ExecutableMessage:
+    """Parse and narrow to an executable message.
+
+    parse_message returns the whole message union, so the type check is also
+    what makes the narrowed return type honest.
+    """
+    message = parse_message(raw)
+    if not isinstance(message, (InstanceMessage, ProgramMessage, VerifiableProgramMessage)):
+        msg = f"message type {message.type} is not executable"
+        raise ValueError(msg)
+    return message
+
+
+def verify_entry(entry: dict) -> tuple[VerificationOutcome, VerifiedMessage | None, str]:
+    """Verify one plan entry.
+
+    Returns (outcome, verified, reason). ``verified`` is set only for VERIFIED;
+    ``reason`` is a safe, caller-facing string for REJECTED.
+    """
+    raw = entry.get("message")
+    if not raw:
+        return VerificationOutcome.UNVERIFIABLE, None, ""
+    if raw.get("item_type") != ItemType.inline.value:
+        return VerificationOutcome.UNVERIFIABLE, None, ""
+    if raw.get("chain") not in EVM_CHAINS:
+        return VerificationOutcome.UNVERIFIABLE, None, ""
+
+    try:
+        message = _parse(raw)
+    except Exception as error:
+        logger.warning("Refusing embedded message for %s: %s", entry.get("item_hash"), error)
+        return VerificationOutcome.REJECTED, None, "message failed validation"
+
+    if str(message.item_hash) != str(entry.get("item_hash")):
+        logger.warning("Embedded message claims %s, plan lists %s", message.item_hash, entry.get("item_hash"))
+        return VerificationOutcome.REJECTED, None, "message does not match the requested item_hash"
+
+    if not _signature_matches(message):
+        logger.warning("Embedded message %s is not signed by its sender", message.item_hash)
+        return VerificationOutcome.REJECTED, None, "signature does not match the sender"
+
+    return VerificationOutcome.VERIFIED, VerifiedMessage(message=message, original=message), ""
--- a/src/aleph/vm/agent/allocation/verify.py
+++ b/src/aleph/vm/agent/allocation/verify.py
@@ -135,7 +135,10 @@ def verify_entry(entry: dict) -> tuple[VerificationOutcome, VerifiedMessage | No
         return VerificationOutcome.REJECTED, None, "message is not an object"
     if raw.get("item_type") != ItemType.inline.value:
         return VerificationOutcome.UNVERIFIABLE, None, ""
-    if raw.get("chain") not in VERIFIABLE_CHAINS:
+    # str() because an unhashable chain (a list, a dict) would raise out of the
+    # set membership test rather than answer, and every shape we do not handle
+    # belongs on the fetch path, not in a traceback.
+    if str(raw.get("chain")) not in VERIFIABLE_CHAINS:
         return VerificationOutcome.UNVERIFIABLE, None, ""
 
     try:
@@ -154,6 +157,6 @@ def verify_entry(entry: dict) -> tuple[VerificationOutcome, VerifiedMessage | No
 
     if not _signature_matches(message):
         logger.warning("Embedded message %s is not signed by its sender", message.item_hash)
-        return VerificationOutcome.REJECTED, None, "signature does not match the sender"
+        return VerificationOutcome.REJECTED, None, "signature is invalid or not the sender's"
 
     return VerificationOutcome.VERIFIED, VerifiedMessage(message=message, original=deepcopy(message)), ""
--- a/src/aleph/vm/agent/allocation/verify.py
+++ b/src/aleph/vm/agent/allocation/verify.py
@@ -124,11 +124,13 @@ def verify_entry(entry: dict) -> tuple[VerificationOutcome, VerifiedMessage | No
     ``reason`` is a safe, caller-facing string for REJECTED.
     """
     raw = entry.get("message")
-    if not raw:
+    if raw is None:
         return VerificationOutcome.UNVERIFIABLE, None, ""
     if not isinstance(raw, dict):
         # Everything else malformed answers REJECTED; a non-dict must not be
-        # the one shape that escapes with an AttributeError instead.
+        # the one shape that escapes with an AttributeError instead. Only a
+        # missing message means "nothing to verify": an empty string or list
+        # is a malformed one, so it is not routed to the fetch fallback.
         logger.warning("Refusing embedded message for %s: not an object", entry.get("item_hash"))
         return VerificationOutcome.REJECTED, None, "message is not an object"
     if raw.get("item_type") != ItemType.inline.value:
@@ -142,8 +144,12 @@ def verify_entry(entry: dict) -> tuple[VerificationOutcome, VerifiedMessage | No
         logger.warning("Refusing embedded message for %s: %s", entry.get("item_hash"), error)
         return VerificationOutcome.REJECTED, None, "message failed validation"
 
-    if str(message.item_hash) != str(entry.get("item_hash")):
-        logger.warning("Embedded message claims %s, plan lists %s", message.item_hash, entry.get("item_hash"))
+    requested = entry.get("item_hash")
+    if requested is None:
+        logger.warning("Plan entry carries a message (%s) but requests no item_hash", message.item_hash)
+        return VerificationOutcome.REJECTED, None, "plan entry has no item_hash"
+    if str(message.item_hash) != str(requested):
+        logger.warning("Embedded message claims %s, plan lists %s", message.item_hash, requested)
         return VerificationOutcome.REJECTED, None, "message does not match the requested item_hash"
 
     if not _signature_matches(message):
--- a/tests/supervisor/fixtures/signed_vprogram_message.json
+++ b/tests/supervisor/fixtures/signed_vprogram_message.json
@@ -1,0 +1,58 @@
+{
+  "chain": "ETH",
+  "channel": null,
+  "content": {
+    "address": "0x34924EF945b931d1E929375556f69CE2E7FE5dcc",
+    "allow_amend": false,
+    "environment": {
+      "internet": true
+    },
+    "metadata": {
+      "name": "demo-vprogram-exec-1.1"
+    },
+    "payment": {
+      "type": "credit"
+    },
+    "requirements": {
+      "node": {
+        "node_hash": "50fd1ed26416a945e561bf8e598412156a493e8d7bb3d6496829805a943ea419"
+      }
+    },
+    "resources": {
+      "memory": 2048,
+      "seconds": 30,
+      "vcpus": 1
+    },
+    "runtime": {
+      "comment": "",
+      "ref": "3f06acd38b47fc8045f202bc7fca49a62809fc694343fb415ebb6cb4c74ca83d"
+    },
+    "time": 1788288282.3883018,
+    "verification": {
+      "backend": "sev_snp",
+      "measurements": [
+        {
+          "platform": "sev_snp",
+          "registers": {
+            "launch": "614077fe4486516dd094eb71750cd0246cd2b199b911d5d7867b3923c4dac803d487ec654268baea7dff6790f515dca4"
+          },
+          "vcpu_type": "EPYC-v4"
+        }
+      ],
+      "policy": 196608
+    },
+    "volumes": [],
+    "workload": {
+      "hash_tree": "2ce324b3eca2728ff3df4ff4826db41f378dbfc87195bd4030d19bf09d64f8c9",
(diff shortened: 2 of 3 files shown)


foxpatch-aleph (changes requested):
The module's security model has a critical gap for InstanceMessage. The PR's stated goal is that the scheduler 'gains none over what they contain,' but InstanceMessage does not validate that `content == json(item_content)` — only ProgramMessage and VerifiableProgramMessage have the `check_content` validator in aleph_message. Since the signature covers `item_hash` (which binds to `item_content`), but the agent reads `verified.message.content`, a malicious scheduler can substitute a tampered `content` dict while keeping the legitimate signed `item_content`. I confirmed this is exploitable: a crafted InstanceMessage with correct item_content/signature but inflated resources in the `content` field passes verification. The fix is to either re-derive `content` from `item_content` (not trust the raw `content` field) or to explicitly validate `content == json.loads(item_content)` for InstanceMessage in this module. Beyond this, the code is well-structured with good test coverage and sound design decisions (UNVERIFIABLE fallback for non-EVM chains, dropping amended messages).

**`src/aleph/vm/agent/allocation/verify.py`** (line 100): CRITICAL: After `_parse` succeeds, there is no check that `message.content` matches `json.loads(message.item_content)` for InstanceMessage. Unlike ProgramMessage and VerifiableProgramMessage, InstanceMessage has no `check_content` validator in aleph_message, so `parse_message` accepts a message where `content` differs from `item_content`. Since the signature only covers `item_hash` (which binds to `item_content`), the scheduler can tamper with the `content` field freely. I confirmed this exploit: a message with correct `item_content`/`signature` but a `content` dict with `vcpus: 64` (instead of the signed `vcpus: 2`) passes verification. The agent would then use the tampered `content`. Fix options: (1) re-parse `content` from `item_content` rather than trusting the raw dict's `content` field, e.g. `message = _parse({**raw, 'content': json.loads(raw['item_content'])})`, or (2) add an explicit assertion `assert message.content.model_dump(exclude_none=True) == json.loads(message.item_content)` after parsing, rejecting on mismatch.

**`src/aleph/vm/agent/allocation/verify.py`** (line 43): EVM_CHAINS is missing many EVM-compatible chains that aleph_message.Chain supports: ARBITRUM (ARB), OPTIMISM (OP), LINEA, LISK, METIS, MODE, BLAST, BOB, FRAXTAL, INK, UNICHAIN, WORLDCHAIN, ZORA, etc. Messages from these chains will be UNVERIFIABLE and fall back to the fetch path. This is safe (no security issue) but means a significant fraction of legitimate EVM messages lose the fast path. Consider using a broader set or deriving from Chain attributes if available.

**`src/aleph/vm/agent/allocation/verify.py`** (line 113): `VerifiedMessage(message=message, original=message)` always sets `original` to the same object as `message`. This is presumably intentional for this PR (the `update_message` deepcopy that would differentiate them comes in a later wiring PR), but the dataclass structure is slightly misleading until then. The test `assert verified.original is verified.message` confirms this is expected — just noting it for future readers.

**`tests/supervisor/test_allocation_verify.py`** (line 72): The tamper test only covers the case where `item_content` is modified (which breaks the hash). There is no test for the actual exploitable path: tampering with the `content` field while leaving `item_content` intact. Adding a test like `test_content_field_tampered_is_rejected` would catch the critical issue described in the verify.py comment. Once the fix is in, this test should assert REJECTED.

**`tests/supervisor/test_allocation_verify.py`** (line 127): The `test_a_storage_item_type_is_unverifiable` test pops `item_content` but leaves `content` in the dict. This is fine since the early `item_type` check returns before parsing, but it's slightly confusing — a reader might think the test exercises a content mismatch. Consider also popping `content` for clarity, or adding a comment.

foxpatch-aleph (changes requested):
The module's security model has a critical gap: the docstring claims aleph_message validates `content == json(item_content)`, but it does not. `item_hash` is checked against `item_content` (the JSON string), and the signature covers `item_hash`, but the agent uses `message.content` (the parsed Pydantic model), which is populated independently from the input dict's `content` key with no cross-check against `item_content`. A malicious scheduler can keep `item_content` consistent with the user's signature while setting `content` to different values (e.g., more vCPUs, a different rootfs). I confirmed this exploit empirically: a user signs for 2 vCPUs, the scheduler embeds `content` with 64 vCPUs, and `verify_entry` returns VERIFIED with the 64-vCPU content. This must be fixed before merging — either re-derive `content` from `item_content` after parsing, or add an explicit check that the parsed content matches the signed `item_content`.

**`src/aleph/vm/agent/allocation/verify.py`** (line 5): The docstring claims `aleph_message`'s validators enforce `content == json(item_content)` for inline messages. This is incorrect — `BaseMessage` has a `check_item_hash` validator (item_hash == sha256(item_content)) and a `check_item_content` validator (item_content is valid JSON), but NO validator that `content` (the parsed Pydantic model) matches `item_content` (the JSON string). I confirmed this by reading the `BaseMessage` source and empirically: a message where `content` has vcpus=64 but `item_content` has vcpus=2 parses successfully. Since the agent uses `message.content`, not `item_content`, a scheduler can embed a message with different `content` than what the user signed, and `verify_entry` will return VERIFIED. This is the exact authority this module exists to deny.

**`src/aleph/vm/agent/allocation/verify.py`** (line 78): After `parse_message(raw)`, the `content` field is populated from the raw dict's `content` key, independently of `item_content`. Consider re-deriving content from item_content here, e.g. by removing the `content` key from `raw` before parsing so Pydantic must parse it from `item_content`, or by adding an explicit check: `if json.loads(message.item_content) != message.content.model_dump(mode='json'): raise ValueError('content does not match item_content')`. Without this, the two-gate security model has a hole: the signature binds `item_hash`, `item_hash` binds `item_content`, but neither binds `content`.

**`src/aleph/vm/agent/allocation/verify.py`** (line 113): `VerifiedMessage(message=message, original=message)` always sets `original` to the same object as `message`. Since amends were dropped, the `original` field is currently dead complexity. This is fine as preparation for future use, but consider noting it in a comment or deferring the field until it's needed.

**`tests/supervisor/test_allocation_verify.py`** (line 72): This test modifies both `item_content` and `content`, so it tests the case where `item_hash` no longer matches `item_content` (caught by the hash validator). The critical attack vector — modifying only `content` while leaving `item_content` intact — is not tested. Consider adding a test where `content` differs from `item_content` but both `item_content`/`item_hash` and the signature are consistent; this should be REJECTED but currently returns VERIFIED.

**`src/aleph/vm/agent/allocation/verify.py`** (line 43): Only 4 of ~20+ EVM-compatible chains are listed. Chains like ARBITRUM, OPTIMISM, LINEA, BASE, etc. all use secp256k1 EVM signatures and could be verified with the same code path. They fall through to UNVERIFIABLE instead. The PR notes acknowledge this is intentional, but the set could be expanded cheaply since the verification logic is identical for all EVM chains.

foxpatch-aleph (changes requested):
The design is sound but there is a critical security vulnerability: InstanceMessage does not enforce content == item_content in aleph_message's validators (only ProgramMessage and VerifiableProgramMessage do). The scheduler can tamper with the 'content' field of an INSTANCE message while keeping 'item_content' and 'item_hash' intact, producing a VERIFIED outcome with different VM resources/rootfs/volumes than the user signed for. This breaks the core security guarantee of the PR. The fix is to add an explicit check in verify_entry that content matches item_content for all message types (not just programs), or to re-derive content from item_content rather than trusting the content field. The existing test suite misses this because the tamper test modifies both item_content and content together.

**`src/aleph/vm/agent/allocation/verify.py`** (line 5): The docstring claims aleph_message enforces 'content == json(item_content)' for inline messages, but this is only true for ProgramMessage and VerifiableProgramMessage. InstanceMessage has no check_content validator. I verified empirically that an InstanceMessage with item_content containing vcpus=2 and content containing vcpus=64 passes parse_message and verify_entry returns VERIFIED. The scheduler can exploit this to tamper with VM configuration without invalidating the signature.

**`src/aleph/vm/agent/allocation/verify.py`** (line 100): After _parse(raw) succeeds, there is no check that message.content matches json.loads(message.item_content). For InstanceMessage, aleph_message does not enforce this. Consider adding an explicit check here, e.g. asserting that message.content.model_dump(exclude_none=True) == json.loads(message.item_content), or re-deriving content from item_content instead of trusting the content field from the scheduler.

**`src/aleph/vm/agent/allocation/verify.py`** (line 113): VerifiedMessage(message=message, original=message) sets both fields to the same object. If 'original' is intended to represent the pre-amend message for future use, it currently provides no independent value. Consider documenting why both fields exist or removing 'original' until it's needed.

**`tests/supervisor/test_allocation_verify.py`** (line 72): test_content_tampered_after_signing_is_rejected tampers both item_content AND content together. This does not catch the critical case where only 'content' is tampered while item_content stays intact (which passes for InstanceMessage since it lacks the check_content validator). A test that tampers only the 'content' field for an INSTANCE message would catch the vulnerability and should be added.

**`tests/supervisor/test_allocation_verify.py`** (line 18): The sign_message helper sets both 'content' (dict) and 'item_content' (json string) but does not verify they match. For InstanceMessage, parse_message accepts them even if they differ. Consider adding a test variant where content is tampered independently of item_content to guard against regression.

foxpatch-aleph (approved):
Well-designed security module with a clear and correct trust model. The two-gate approach (hash binding + signature verification) and the three-outcome design (VERIFIED/REJECTED/UNVERIFIABLE) are sound. The decision to drop amended_message is well-reasoned. The code is purely additive with no production caller yet, which is expected for a stacked PR. The import reordering in views/__init__.py and test_allocation_teardown.py is cosmetic. A few minor gaps in test coverage (only INSTANCE messages tested, no None-signature or non-executable-type tests) and a slight docstring inaccuracy about InstanceMessage validation are worth addressing but are not blocking.

**`src/aleph/vm/agent/allocation/verify.py`** (line 6): The docstring claims `content == json(item_content)` is enforced for inline messages generally, but `aleph_message` only enforces this for `ProgramMessage` and `VerifiableProgramMessage` (via their `check_content` validators). `InstanceMessage` has no such validator — it only enforces `item_hash == sha256(item_content)`. For INSTANCE messages, a scheduler could in principle provide a `content` field that differs from `item_content` while keeping the hash valid. This is not exploitable because the signature covers `item_hash` (not `content`), and `parse_message` will parse `content` as an `InstanceContent` pydantic model from the `content` field, but the docstring should clarify that the content==item_content binding applies to PROGRAM and V-PROGRAM only.

**`src/aleph/vm/agent/allocation/verify.py`** (line 113): Both `message` and `original` are the same object. This is correct given the design (no amends accepted, so the signed message IS the original), but it diverges from `load_updated_message` in `messages.py` where `original` is a pre-deepcopy snapshot and `message` gets its `use_latest` refs resolved. A future integrator wiring this into `start_persistent_vm` will need to understand that `verified.original` here means "the signed message before ref resolution" — and will need to deepcopy before calling `update_message` to preserve that semantics. A one-line comment here would save someone time later.

**`tests/supervisor/test_allocation_verify.py`** (line 61): All tests use INSTANCE messages. PROGRAM and V-PROGRAM paths through `_parse` and `verify_entry` are not exercised. Consider adding at least one test using the existing `tests/supervisor/fixtures/vprogram_message.json` fixture (re-signed) to cover the V-PROGRAM branch.

**`tests/supervisor/test_allocation_verify.py`** (line 139): Missing test cases worth adding: (1) a message with `signature: None` — the field is `Optional[str]` in `aleph_message`, and while the code correctly REJECTs it (the `except Exception` in `_signature_matches` catches the resulting error), this path should be tested; (2) a non-executable message type (e.g., AGGREGATE or STORE) — should be REJECTED by `_parse`, but this is untested.

foxpatch-aleph (commented):
The security reasoning is sound and I verified both gates independently: aleph_message (1.2.1-dev and pinned 1.3.1) does enforce item_hash == sha256(item_content) and content == item_content for inline messages, and the explicit .value usage in verification_buffer is load-bearing (f"{Chain.ETH}" renders "Chain.ETH" in this environment, and a signature over that buffer is correctly rejected). I probed adversarial inputs — stripped, empty, and garbage signatures; hash mismatches; missing entry hashes; POST-type messages; lowercase, OP, and non-EVM chains; storage items — and every case fails closed in the intended direction. Tests (8/8), mypy, and ruff format all pass, and the views/__init__.py churn matches what CI's style script produces. The remaining comments are design-contract questions rather than bugs: the aliased message/original fields are a trap for the next PR's caller (update_message mutates in place), a non-dict embedded message escapes the REJECTED contract with an AttributeError, and only INSTANCE messages are tested despite the V-PROGRAM type string never having been confirmed against live traffic.

**`src/aleph/vm/agent/allocation/verify.py`** (line 113): VerifiedMessage(message=message, original=message) returns the same object in both fields, and the test cements it. This dataclass mirrors load_updated_message's (message, original) contract, but there the original survives because it is deep-copied before update_message(), which mutates in place (update_with_latest_ref assigns obj.ref — agent/messages.py:54). If the next PR's caller runs `await update_message(verified.message)`, `original` is silently corrupted and the distinction the field exists to provide is gone. Either deepcopy for `original` here (cheap, once per allocation) or document that the caller must copy before updating — worth settling before the consumer lands.

**`src/aleph/vm/agent/allocation/verify.py`** (line 94): These raw.get() calls run before the try block, so a non-dict message (e.g. {"message": "..."}) raises AttributeError instead of returning an outcome tuple. Everything else in this module's contract says malformed -> REJECTED; an isinstance(raw, dict) check (or moving the accessor calls inside the try) would keep that uniform so a malformed plan entry cannot escape the REJECTED classification with an exception.

**`src/aleph/vm/agent/allocation/verify.py`** (line 43): The name EVM_CHAINS implies exhaustiveness, but the Chain enum ships many more EVM chains (OP, ARB, POL, LINEA, AURORA, ZORA, ...); messages from those land in UNVERIFIABLE and take the fetch path. That is the safe direction, but a name like VERIFIABLE_CHAINS, or a comment saying "chains whose EVM recovery we currently implement; extend as needed", would prevent future confusion.

**`tests/supervisor/test_allocation_verify.py`** (line 61): Only INSTANCE messages are exercised, but the module also accepts PROGRAM and V-PROGRAM. The buffer format was confirmed against live INSTANCE messages only; the dashed "V-PROGRAM" type string in the signed buffer was never validated against real traffic. A test signing a v-program message would at least pin this module's intent, and a live-message recovery for one v-program would close the gap for good.

foxpatch-aleph (approved):
This PR is a well-designed trust boundary for scheduler-embedded messages, and every load-bearing claim checks out under independent verification: the signature buffer format recovers the senders of 10 live INSTANCE messages from api3.aleph.im; both the installed aleph-message and the pinned 1.3.1 wheel enforce item_hash == sha256(item_content) for inline messages; and InstanceMessage genuinely lacks the check_content validator, making the re-derivation of content from item_content in _parse() the load-bearing fix — confirmed empirically, as a tampered content field is discarded in favor of the signed bytes. The decision to drop amended_message is correct given update_message only refreshes use_latest refs in place, and the REJECTED-not-fallback semantics for tampering attempts is the right call. The test suite (11 tests, not 8 as the stale PR description says) passes, adversarial probing of verify_entry produced no crashes or unsafe outcomes, and the only findings are minor: unrelated isort churn in views/__init__.py and a small falsy-shape inconsistency in verify_entry's first gate.

**`src/aleph/vm/agent/views/__init__.py`** (line 25): These two hunks are unrelated pre-commit/isort churn (import re-sorting plus collapsing the `supervisor_interface.types` import). Harmless, but they add noise to an otherwise tightly scoped diff — consider keeping the PR to the new module and its tests, or noting the lint fix in the PR description.

**`src/aleph/vm/agent/allocation/verify.py`** (line 127): Minor consistency nit: `if not raw:` routes falsy non-dict values (`""`, `[]`, `0`) to UNVERIFIABLE, while truthy non-dicts fall to REJECTED one line below, so a scheduler sending `[]` vs `["x"]` gets different outcomes. Both are safe (fetch vs. reject), but if the intent is that any non-object payload is malformed, checking `raw is None` for the 'no message' case and letting `isinstance` handle the rest would be more uniform. Non-blocking.

foxpatch-aleph (approved):
This module's security design is correct and I verified its central claims independently rather than trusting the description. The signature buffer format (chain\nsender\ntype\nitem_hash) recovers the signers of 12 live network messages across INSTANCE, PROGRAM and V-PROGRAM; the installed aleph_message confirms InstanceMessage lacks the check_content validator, so re-deriving content from the hash-bound item_content is indeed the necessary defense against scheduler-substituted content; and check_item_hash binds item_hash to sha256(item_content) for inline messages. I probed untested edge cases (None/empty signature, signed non-executable types, entry without item_hash, extra envelope fields, garbage item_content) and every one fails closed to REJECTED or UNVERIFIABLE as documented, with none of the unsigned envelope fields consumed downstream in src/. The REJECTED-without-fallback design creates no false availability gap since the fetch path parses through the same parse_message. The 11 tests pass, and the CI-enforced checks (ruff format, isort, mypy) are clean. Two minor non-blocking observations below.

**`tests/supervisor/test_allocation_verify.py`** (line 192): Non-blocking: of the three REJECTED reasons documented in the PR (bad signature, hash mismatch, non-executable type), the non-executable-type path has no test. I confirmed manually that a correctly-signed POST message answers REJECTED via the _parse narrow, but a test pinning that (e.g. sign_message(..., message_type="POST") with post-shaped content) would complete the documented contract.

**`src/aleph/vm/agent/allocation/verify.py`** (line 145): Nit: when the entry carries an embedded message but no item_hash key at all, this returns the reason "message does not match the requested item_hash", which reads oddly when there is no requested hash. Functionally fine (still REJECTED), just imprecise.

**`src/aleph/vm/agent/allocation/verify.py`** (line 114): Nit: `ruff check` (UP038) prefers `isinstance(message, InstanceMessage | ProgramMessage | VerifiableProgramMessage)`. Not CI-enforced today (only `ruff format` runs in CI, and there are ~108 pre-existing ruff check findings across the agent package), so purely cosmetic.

foxpatch-aleph (approved):
A carefully-designed security module with an unusually thorough rationale, and the two-gate security model holds up under scrutiny: I verified against the installed aleph_message source that item_hash is bound to item_content for inline messages on all executable types, and that the InstanceMessage content/item_content divergence hole the re-derivation defends against is real. Probing malformed inputs (extra keys, missing/garbage signature, non-executable type, hash mismatches) showed every path fails closed to REJECTED or degrades to UNVERIFIABLE exactly as documented. The decision to drop amended_message is correct — update_message only refreshes use_latest refs, so an embedded amend was never verifiable, and the fixture note checks out. Findings are non-blocking: the signature format is only tested against a helper that mirrors it, and a couple of documented behaviors lack test coverage.

**`tests/supervisor/test_allocation_verify.py`** (line 18): sign_message reimplements the assumed signature format, so the suite would pass even if verification_buffer disagreed with the real network format — the format is effectively only anchored by the one-time live check described in the PR description. Consider committing a real captured message (a golden vector) or asserting against the SDK's get_verification_buffer so a future buffer-format regression can't slip through unnoticed.

**`src/aleph/vm/agent/allocation/verify.py`** (line 141): Worth being aware that BaseMessage uses extra="forbid", so any benign envelope divergence — e.g. a scheduler passing a CCN-style payload with an extra id key — lands here and is REJECTED with no fallback, i.e. the allocation fails rather than degrading to fetch. Fail-closed is the right default and parity with the fetch path (same parse) holds for genuinely malformed messages, but it does make the embedded-message schema an exact contract between scheduler and agent: a schema drift between the two components will surface as 'allocations rejected', not as a fallback. Fine as-is; just documenting the load-bearing nature of the REJECTED-vs-UNVERIFIABLE split.

**`tests/supervisor/test_allocation_verify.py`** (line 18): Two small coverage gaps for behaviors the docstring claims: no test exercises a non-executable message type (e.g. type=POST) hitting the 'is not executable' REJECTED path, and none covers an inline message missing item_content (the ValueError in _parse). Both behave correctly when probed manually; tests would lock that in.

odesenfans:
Thanks for the reviews. The blocking finding (the `content` / `item_content` binding hole on `InstanceMessage`) was fixed earlier by re-deriving `content` from the signed `item_content` in `_parse`; the three later reviews confirm that independently. This comment covers the non-blocking tail.

## Taken

**Golden vector.** Agreed, and the concern was real: `sign_message` in the suite mirrors `verification_buffer`, so the tests were self-consistent rather than anchored. I confirmed the blind spot by swapping `chain` and `sender` in the buffer *in both the helper and the module* -- all 14 tests still passed. There is now a V-PROGRAM captured from api3.aleph.im, signed by a wallet we do not own, and it is the only test that fails under that mutation.

Two notes on how this landed:

- The suggested `get_verification_buffer` anchor is not available: it is not in `aleph_message`, and `aleph-sdk-python` is not a dependency here. I did not want to add one to this module, so the real capture does that job instead.
- `tests/supervisor/fixtures/vprogram_message.json` could not serve as the vector. Its `item_content` was edited and its `item_hash` re-derived in #1143 without re-signing, so its signature recovers a stranger. Its existing readers never verify signatures, so nothing noticed. The new fixture is separate and deliberately untouched.

While there I validated the buffer format against live traffic for all three executable types: 6/6 INSTANCE, 6/6 PROGRAM, 5/5 V-PROGRAM senders recovered. That closes the open question about the dashed `V-PROGRAM` type string, which several reviews flagged as never having been checked against real messages. It is correct.

**Coverage gaps.** Added tests for the non-executable type rejection and for an inline message with no `item_content`.

**Falsy-shape uniformity and the imprecise reason string.** Both fixed in a separate commit, with tests. `raw is None` is now the only "nothing to verify" case; every other shape is malformed and rejected. An entry with a message but no `item_hash` gets its own reason rather than claiming a mismatch against a hash nobody requested.

## Not taken

**`ruff check` UP038 (`isinstance` union syntax).** Only `ruff format` runs in CI, and there are ~108 pre-existing `ruff check` findings across the agent package. Changing this one call site would make the file inconsistent with its neighbours for no enforced benefit. Worth doing as a package-wide pass, not here.

**Routing `extra="forbid"` parse failures to UNVERIFIABLE instead of REJECTED.** The two reviews disagree here, and the later one is right that the failure modes differ: a scheduler-built envelope can diverge where a CCN-fetched one would not, so drift surfaces as rejected allocations rather than a fetch fallback. I am still leaving the behavior as-is. In practice the risk is narrower than it looks: I checked, and a raw CCN message verifies unmodified, `confirmations` / `confirmed` / `size` included, because those are declared fields on `BaseMessage`. So only genuinely novel keys trip it. Given the choice, I would rather this module be strict about input it cannot parse than quietly fall back on it, since unparseable input is exactly what it exists to scrutinize. Happy to revisit if scheduler-side schema drift ever actually bites.

https://claude.ai/code/session_01CYrne44Q6jEc3KcGPERJgB

foxpatch-aleph (changes requested):
The security design is sound and well-verified: the two-gate analysis is correct against aleph_message 1.2.1 (BaseMessage.check_item_hash binds item_hash to item_content for all inline messages; InstanceMessage genuinely lacks a check_content validator, so re-deriving content from item_content is a necessary mitigation), dropping the amended_message field is the right call given that update_message only resolves use_latest refs, and the live-network fixture pins the signing-buffer format. Tests are thorough and pass. One blocking issue: a plan entry whose embedded message carries an unhashable chain value (e.g. {"item_type": "inline", "chain": ["ETH"]}) raises TypeError: unhashable type out of verify_entry instead of returning a verdict, violating the module's own invariant that malformed entries always get a verdict — the same class of bug the head commit just fixed for non-dict messages. Suggested fix: coerce with str(raw.get('chain')) not in VERIFIABLE_CHAINS, or guard with an isinstance check, and add a regression test.

**`src/aleph/vm/agent/allocation/verify.py`** (line 138): This raises instead of returning a verdict when the scheduler sends an unhashable `chain` value: `verify_entry({"item_hash": "b"*64, "message": {"item_type": "inline", "chain": ["ETH"]}})` -> `TypeError: unhashable type: 'list'`, because set membership hashes the operand. That is the same class of escape hatch the in-code comment at line 130 ("a non-dict must not be the one shape that escapes with an AttributeError") and the head commit set out to eliminate. Wrapping with `str(...)` (`str(raw.get("chain")) not in VERIFIABLE_CHAINS`) or an isinstance-against-str guard keeps every malformed shape on the UNVERIFIABLE path. Worth a regression test alongside `test_an_empty_message_is_rejected_not_treated_as_absent`.

**`src/aleph/vm/agent/allocation/verify.py`** (line 157): Minor legibility nit, same spirit as the head commit: when `message.signature` is None, empty, or non-hex, `_signature_matches` fails via the exception path and logs "Could not recover the signer", but the caller-facing reason says the signature "does not match the sender", which reads as a recovered-and-different key. A reason like "invalid or mismatched signature" (or branching on the two failure modes) would be less misleading in the verdict.

odesenfans:
Both taken.

**Unhashable `chain`.** Confirmed, and it is slightly wider than reported: a dict trips it too, not just a list, since the failure is `in` against a set hashing its operand. Fixed with the `str()` coercion.

I put it on the UNVERIFIABLE path rather than REJECTED to stay consistent with the gate it lives in: the early screens ask "is this something we can verify", not "is this well-formed", which is why a `chain` of `None` or `"SOL"` and an `item_type` of `["inline"]` already degrade to fetch. An unusable chain is by definition not one we recover signatures for, so it belongs with them, and the CCN copy gets validated on the fetch path anyway. The neighbouring `item_type` gate compares with `!=` and was never exposed to this.

Regression test is parametrised over list, dict, set, `None` and `0`. I checked it actually bites: reverting the `str()` fails exactly the list and dict cases.

**Signature reason string.** Agreed it was misleading. It now reads "signature is invalid or not the sender's", which covers both the recovered-and-differs case and the never-recovered-anything case. I kept `_signature_matches` returning a bool rather than branching the two apart: the distinction is already in the logs, where `_signature_matches` logs the recovery failure with a traceback, and the caller-facing reason goes to the scheduler, which cannot act differently on the two.

21 tests pass, `ruff format` / `isort` / mypy clean.

foxpatch-aleph (changes requested):
The security design is correct and well-evidenced: the two-gate analysis (sha256 item_hash binding plus signature over chain/sender/type/item_hash) holds against the installed aleph_message source, the InstanceMessage content/item_content divergence hole is properly closed by re-deriving content from the signed bytes, ambiguous shapes safely degrade to the CCN fetch path, and the live-network fixture independently pins the buffer format (all 16 tests pass). However, verify_entry violates its own malformed-input contract: a message whose chain field is a JSON list or dict reaches `raw.get("chain") not in VERIFIABLE_CHAINS`, and set membership hashes the operand, raising TypeError instead of returning a verdict — the exact escape class the module's line 130 comment and the branch's own verdict-uniformity commit promise to prevent. Since chain is fully scheduler-controlled, this lets one crafted plan entry abort processing of the entire allocation rather than degrading that single entry. Please gate the check (e.g. `raw.get("chain") in VERIFIABLE_CHAINS if isinstance(raw.get("chain"), str) else ...`, or move the chain check after parsing against the enum) and add the missing non-string-chain test.

**`src/aleph/vm/agent/allocation/verify.py`** (line 138): This raises `TypeError: unhashable type: 'list'` (or 'dict') when the scheduler embeds a message whose `chain` is a JSON list or dict, e.g. `{"message": {"item_type": "inline", "chain": []}}`. Set membership hashes the operand before comparing, so the exception escapes `verify_entry` instead of returning UNVERIFIABLE — the same malformed-shape escape the comment at line 130 and the verdict-uniformity commit exist to prevent. Since `chain` is fully scheduler-controlled, one crafted entry can abort the caller's plan processing instead of degrading this entry to the fetch path. Suggest checking `isinstance(..., str)` first, or comparing the parsed enum after `_parse` (where the same dicts would fail pydantic and return REJECTED, which also matches the verdict for the non-object message shape). Worth a regression test alongside `test_a_message_that_is_not_an_object_is_rejected`.

**`src/aleph/vm/agent/allocation/verify.py`** (line 114): Minor: `ruff check` (rules selected in pyproject) flags this line UP038 (use `X | Y`) and `verify_entry` PLR0911 (9 returns). Neither is CI-enforced today (the style gate only runs `ruff format --diff`, and the agent tree already carries ~109 such findings), so feel free to ignore — noting only because the new file otherwise lints clean.

odesenfans:
This review raced the push: it is recorded against `e7833d96`, but `e7833d96` **is** the commit that fixes the unhashable-chain bug, and the review describes it as still live while reporting 16 tests, which was the count at the previous commit (`126a5f92`). The tree it actually assessed looks like the one before the fix.

Verified against the pushed commit on a clean detached checkout of `e7833d96`:

```
chain=[]        -> unverifiable
chain=["ETH"]   -> unverifiable
chain={"a": 1}  -> unverifiable
21 passed
```

The gate at line 141 there reads `if str(raw.get("chain")) not in VERIFIABLE_CHAINS:`. The regression test is parametrised over list, dict, set, `None` and `0`, and reverting the `str()` fails exactly the list and dict cases, so it does bite.

The diagnosis in the review was right, and the extra detail about one crafted entry aborting the whole plan is a fair characterisation of why it mattered. It is just already fixed. A re-review of the current head should come back clean on that point.

On the lint note: confirmed accurate, `UP038` and `PLR0911` are both selected in `pyproject.toml` (CI only runs `ruff format --diff`, which is why they do not gate). Still leaving both. `UP038` alone would not make the file clean, since `PLR0911` would remain, and clearing that one means restructuring `verify_entry` away from the flat sequence of guard clauses that makes its order of checks readable. The agent tree carries 105 of these findings; a package-wide pass is the right place to deal with them, not this PR.

foxpatch-aleph (approved):
This is a well-designed trust-boundary module and I found no security or correctness issues. I independently verified the PR's central claims against the aleph_message library source: InstanceMessage genuinely lacks the check_content validator that Program/VerifiableProgram messages have, so re-deriving content from the hash-bound item_content is the correct fix, and the signature buffer format is anchored by a test that recovers the signer of a live captured network message. Adversarial probes (type-switch and chain-switch after signing, missing/garbage signatures, garbage sender, divergent content, 0x-prefixed and uppercase hashes) all answer correctly and fail safe: REJECTED never degrades to the fetch fallback, and UNVERIFIABLE always falls back to a network-anchored fetch. The amended_message drop is sound — update_message only resolves use_latest refs, so an embedded resolved amend would be an unverifiable bare hash whose acceptance would grant the scheduler rootfs selection, matching the legacy path's behavior. The reasoning is documented exceptionally well, both in the module docstring and the tests, which are adversarial rather than happy-path. The comments below are minor consistency/robustness observations that do not block.

**`src/aleph/vm/agent/allocation/verify.py`** (line 120): Minor consistency observation: the module goes to lengths to answer every malformed *message* shape (non-dict, empty string, unhashable chain) with a verdict, but `entry` itself is only guarded by the type hint — a non-dict entry would raise AttributeError from `entry.get(...)`. That's fine if the plan parser (in the next stacked PR) guarantees dict entries, but given the defensive posture here it may be worth either a one-line docstring note on that precondition or the same narrowing guard, so the "every entry gets a verdict" property holds unconditionally.

**`src/aleph/vm/agent/allocation/verify.py`** (line 136): A message dict that is present but *missing* item_type (e.g. `{}`) takes the UNVERIFIABLE fallback, while a non-dict message is REJECTED. This fails safe — the fallback fetch is anchored by item_hash — but it's a slight inconsistency with the module's stated posture that malformed answers REJECTED. If a fetch is preferable for a shape we can't classify, a brief comment saying so would help future readers; otherwise this is fine as-is.

**`src/aleph/vm/agent/allocation/verify.py`** (line 134): Very minor nit: `entry.get("item_hash")` (and the pydantic error text at line 147) are scheduler-controlled and logged verbatim, so a malicious scheduler could inject newlines to forge log lines. The caller-facing `reason` strings are correctly static, so this only affects operator logs. Not worth blocking on; just flagging for awareness.

PR #1159, opened 2026-08-25
feat(agent): batch, release-aware capacity simulation

odesenfans:
Stacked on #1158. Third of the async-allocations stack.

A plan asks a different question than a single create: "if I gave you all of these, and dropped those". `check_capacity` answers only "does this one fit right now", which is wrong twice over for a plan:

- Judging each candidate independently makes the CRN answer yes three times for three VMs that only fit twice.
- "Allocate C and delete B" must admit C against B's *release*, or the CRN refuses work it is about to have room for.

## Changes

- `CapacityManager.simulate(candidates, *, releasing)` returns one `AdmissionVerdict` per candidate. Cumulative (each acceptance is committed before the next is judged) and release-aware (hashes the plan will stop are subtracted from the committed sums). It reserves nothing, which is what makes it safe for the upcoming capacity-check endpoint to call speculatively across several CRNs.
- The arithmetic moves to `_check_against`, shared by both entry points: `check_capacity` judges against the registry as it stands, `simulate` against sums it accumulates itself. One implementation means an advisory answer can never be stronger or weaker than the enforced one.
- Drops a docstring line that went stale with #1153 (the create path no longer passes `disk_mib=0`).

## How to test

`pytest tests/supervisor/test_agent_capacity.py` — 49 pass, of which 19 are new. The other 30 are the pre-existing `check_capacity` tests running unchanged through the extracted arithmetic, which is the proof that the refactor preserved behaviour.

The new tests pin `HOST_MEMORY_RESERVED_MIB` and `PROGRAM_MEMORY_RESERVED_MIB` wherever their arithmetic depends on them, as the older tests in the file already do, so a change to the defaults breaks the setting rather than these tests. The figures are computed from 2048 / 8192, making the instance bucket physical - 10240 MiB. 48 GiB fits exactly two 16384 MiB instances; the release test uses 40 GiB, where B plus C does not fit but C alone does. If those defaults change, recompute rather than nudging the numbers: a cumulative test that goes green because *nothing* fits proves nothing.

### DIFF
--- a/src/aleph/vm/agent/capacity.py
+++ b/src/aleph/vm/agent/capacity.py
@@ -162,6 +162,7 @@ class CapacityManager:
         committed_instance_memory_mib: int,
         committed_program_memory_mib: int,
         committed_vcpus: int,
+        committed_disk_mib: int = 0,
     ) -> None:
         """The admission arithmetic, against caller-supplied commitments.
 
@@ -197,7 +198,9 @@ class CapacityManager:
             committed_memory_mib = committed_program_memory_mib
             memory_cap_mib = program_memory_cap_mib
 
-        available_disk_mib = self._available_disk_bytes() // (1024 * 1024)
+        # Free space is a live figure, not a committed sum, so a batch caller
+        # passes what it has already promised to the candidates before this one.
+        available_disk_mib = max(self._available_disk_bytes() // (1024 * 1024) - committed_disk_mib, 0)
 
         errors: list[str] = []
 
@@ -254,13 +257,31 @@ class CapacityManager:
         """Judge a whole plan at once.
 
         Cumulative: each accepted candidate is committed before the next is
-        judged, so three VMs that only fit twice get two yeses and one no.
+        judged, so three VMs that only fit twice get two yeses and one no. This
+        covers memory, vCPUs and disk. Disk needs the accumulator because free
+        space is read live and no record of it exists until the volumes are
+        actually written.
 
         Release-aware: hashes the plan is about to stop are subtracted from the
         committed sums, so "allocate C, delete B" admits C against B's memory.
+        Their disk is not credited back: nothing has been deleted yet, so the
+        space is genuinely still occupied, and guessing otherwise would make
+        the advisory answer stronger than the enforced one.
 
         Side-effect free: nothing here reserves or holds anything, which is
         what makes it safe for the speculative capacity-check endpoint.
+
+        GPUs are out of scope. Card availability comes from the supervisor's
+        async host info and is claimed through stateful holds, neither of which
+        fits a synchronous side-effect-free call, so a GPU plan needs a
+        separate answer.
+
+        The third element of a candidate is the caller's memory-bucket choice,
+        NOT ResourceRequirements.is_instance. The two disagree for V-PROGRAMs:
+        requirements_from_message reports is_instance=False (the content is not
+        an InstanceContent) while a V-PROGRAM is committed to the instance
+        bucket, which is what _committed_resources and _admit both do. Pass
+        isinstance(content, (InstanceContent, VerifiableProgramContent)).
         """
         committed_instance, committed_program, committed_vcpus = self._committed_resources(None)
         for vm_hash in releasing:
@@ -273,8 +294,14 @@ class CapacityManager:
             else:
                 committed_program -= resources.memory
             committed_vcpus -= resources.vcpus
+        # A registry inconsistency must not make admission more permissive than
+        # an empty node.
+        committed_instance = max(committed_instance, 0)
+        committed_program = max(committed_program, 0)
+        committed_vcpus = max(committed_vcpus, 0)
 
         verdicts: list[AdmissionVerdict] = []
+        committed_disk = 0
         for vm_hash, requirements, is_instance in candidates:
             try:
                 self._check_against(
@@ -286,6 +313,7 @@ class CapacityManager:
                     committed_instance_memory_mib=committed_instance,
                     committed_program_memory_mib=committed_program,
                     committed_vcpus=committed_vcpus,
+                    committed_disk_mib=committed_disk,
                 )
             except InsufficientResourcesError as error:
                 logger.info("Plan candidate %s refused: %s", vm_hash, error)
@@ -298,6 +326,7 @@ class CapacityManager:
             else:
                 committed_program += requirements.memory_mib
             committed_vcpus += requirements.vcpus
+            committed_disk += requirements.disk_mib
             verdicts.append(AdmissionVerdict(vm_hash, True))
         return verdicts
 
--- a/src/aleph/vm/agent/capacity.py
+++ b/src/aleph/vm/agent/capacity.py
@@ -49,6 +49,20 @@ class ResourceRequirements:
     gpu_device_ids: list[str] = field(default_factory=list)
 
 
+@dataclass(frozen=True)
+class AdmissionVerdict:
+    """One candidate's admission answer.
+
+    ``code`` and ``detail`` are safe to hand back to the scheduler; the full
+    error text stays in the logs.
+    """
+
+    vm_hash: ItemHash
+    accepted: bool
+    code: str = ""
+    detail: str = ""
+
+
 def requirements_from_message(content: ExecutableContent) -> ResourceRequirements:
     """Extract the resources a message requests into a message-free DTO."""
     is_instance = isinstance(content, InstanceContent)
@@ -115,21 +129,50 @@ class CapacityManager:
         Two-bucket memory accounting: instances share
         physical - HOST_MEMORY_RESERVED_MIB - PROGRAM_MEMORY_RESERVED_MIB,
         programs share PROGRAM_MEMORY_RESERVED_MIB. vCPUs are capped at
-        physical cores times VCPU_OVERCOMMIT_FACTOR. Disk is only checked for
-        disk_mib > 0 (the reserve path; the create path passes 0).
+        physical cores times VCPU_OVERCOMMIT_FACTOR. Disk is checked whenever
+        disk_mib > 0, which every create path now passes (#1153).
 
         ``exclude_vm_hash`` skips that VM's own registry record from the
         committed sums: the create paths record the VM before admission (the
         early owner record, or a leftover record on a recreate), and its own
         record must not count against its own request.
         """
-        required_memory_mib = memory_mib
-        required_vcpus = vcpus
-        required_disk_mib = disk_mib
-
         committed_instance_memory_mib, committed_program_memory_mib, committed_vcpus = self._committed_resources(
             exclude_vm_hash
         )
+        self._check_against(
+            memory_mib=memory_mib,
+            vcpus=vcpus,
+            disk_mib=disk_mib,
+            max_volume_mib=max_volume_mib,
+            is_instance=is_instance,
+            committed_instance_memory_mib=committed_instance_memory_mib,
+            committed_program_memory_mib=committed_program_memory_mib,
+            committed_vcpus=committed_vcpus,
+        )
+
+    def _check_against(
+        self,
+        *,
+        memory_mib: int,
+        vcpus: int,
+        disk_mib: int,
+        max_volume_mib: int,
+        is_instance: bool,
+        committed_instance_memory_mib: int,
+        committed_program_memory_mib: int,
+        committed_vcpus: int,
+    ) -> None:
+        """The admission arithmetic, against caller-supplied commitments.
+
+        Split out so simulate() can judge a candidate against sums it is
+        accumulating itself, while check_capacity keeps judging against the
+        registry as it stands. One implementation, so an advisory answer can
+        never be stronger or weaker than the enforced one.
+        """
+        required_memory_mib = memory_mib
+        required_vcpus = vcpus
+        required_disk_mib = disk_mib
 
         physical_memory_mib = psutil.virtual_memory().total // (1024 * 1024)
         physical_cores = psutil.cpu_count() or 1
@@ -201,6 +244,62 @@ class CapacityManager:
                     "disk_mib": available_disk_mib,
                 },
             )
+
+    def simulate(
+        self,
+        candidates: list[tuple[ItemHash, ResourceRequirements, bool]],
+        *,
+        releasing: frozenset[ItemHash] = frozenset(),
+    ) -> list[AdmissionVerdict]:
+        """Judge a whole plan at once.
+
+        Cumulative: each accepted candidate is committed before the next is
+        judged, so three VMs that only fit twice get two yeses and one no.
+
+        Release-aware: hashes the plan is about to stop are subtracted from the
+        committed sums, so "allocate C, delete B" admits C against B's memory.
+
+        Side-effect free: nothing here reserves or holds anything, which is
+        what makes it safe for the speculative capacity-check endpoint.
+        """
+        committed_instance, committed_program, committed_vcpus = self._committed_resources(None)
+        for vm_hash in releasing:
+            record = self.registry.get(vm_hash)
+            if record is None or not record.message.resources:
+                continue
+            resources = record.message.resources
+            if isinstance(record.message, InstanceContent) or record.is_vprogram:
+                committed_instance -= resources.memory
+            else:
+                committed_program -= resources.memory
+            committed_vcpus -= resources.vcpus
+
+        verdicts: list[AdmissionVerdict] = []
+        for vm_hash, requirements, is_instance in candidates:
+            try:
+                self._check_against(
+                    memory_mib=requirements.memory_mib,
+                    vcpus=requirements.vcpus,
+                    disk_mib=requirements.disk_mib,
+                    max_volume_mib=requirements.max_volume_mib,
+                    is_instance=is_instance,
+                    committed_instance_memory_mib=committed_instance,
+                    committed_program_memory_mib=committed_program,
+                    committed_vcpus=committed_vcpus,
+                )
+            except InsufficientResourcesError as error:
+                logger.info("Plan candidate %s refused: %s", vm_hash, error)
+                verdicts.append(
+                    AdmissionVerdict(vm_hash, False, "insufficient_capacity", "not enough capacity on this CRN")
+                )
+                continue
+            if is_instance:
+                committed_instance += requirements.memory_mib
+            else:
+                committed_program += requirements.memory_mib
+            committed_vcpus += requirements.vcpus
+            verdicts.append(AdmissionVerdict(vm_hash, True))
+        return verdicts
 
     def _committed_resources(self, exclude_vm_hash: ItemHash | None) -> tuple[int, int, int]:
         """(committed_instance_memory_mib, committed_program_memory_mib,
--- a/src/aleph/vm/agent/capacity.py
+++ b/src/aleph/vm/agent/capacity.py
@@ -46,6 +46,8 @@ class ResourceRequirements:
     memory_mib: int
     disk_mib: int
     max_volume_mib: int = 0
+    # The memory bucket, from is_instance_bucket: True for a V-PROGRAM even
+    # though it is not an InstanceContent.
     is_instance: bool = False
     gpu_device_ids: list[str] = field(default_factory=list)
 
@@ -77,7 +79,7 @@ def is_instance_bucket(content: ExecutableContent) -> bool:
 
 def requirements_from_message(content: ExecutableContent) -> ResourceRequirements:
     """Extract the resources a message requests into a message-free DTO."""
-    is_instance = isinstance(content, InstanceContent)
+    is_instance = is_instance_bucket(content)
     volume_sizes_mib: list[int] = []
     if isinstance(content, InstanceContent) and content.rootfs:
         volume_sizes_mib.append(content.rootfs.size_mib)
@@ -264,7 +266,7 @@ class CapacityManager:
 
     def simulate(
         self,
-        candidates: list[tuple[ItemHash, ResourceRequirements, bool]],
+        candidates: list[tuple[ItemHash, ResourceRequirements]],
         *,
         releasing: frozenset[ItemHash] = frozenset(),
         available_gpus: list[GpuDevice] | None = None,
@@ -312,14 +314,8 @@ class CapacityManager:
         allocate C onto B's card" is answered no even though doing it in that
         order would work.
 
-        The third element of a candidate is the caller's memory-bucket choice,
-        NOT ResourceRequirements.is_instance. The two disagree for V-PROGRAMs:
-        requirements_from_message reports is_instance=False (the content is not
-        an InstanceContent) while a V-PROGRAM is committed to the instance
-        bucket, which is what _committed_resources and _admit both do. Pass
-        is_instance_bucket(content) rather than restating the rule.
         """
-        candidate_hashes = {vm_hash for vm_hash, _, _ in candidates}
+        candidate_hashes = {vm_hash for vm_hash, _ in candidates}
         committed_instance, committed_program, committed_vcpus = self._committed_resources(candidate_hashes)
         for vm_hash in releasing:
             if vm_hash in candidate_hashes:
@@ -342,7 +338,7 @@ class CapacityManager:
 
         verdicts: list[AdmissionVerdict] = []
         committed_disk = 0
-        for vm_hash, requirements, is_instance in candidates:
+        for vm_hash, requirements in candidates:
             refusal: tuple[str, str] | None = None
             try:
                 self._check_against(
@@ -350,7 +346,7 @@ class CapacityManager:
                     vcpus=requirements.vcpus,
                     disk_mib=requirements.disk_mib,
                     max_volume_mib=requirements.max_volume_mib,
-                    is_instance=is_instance,
+                    is_instance=requirements.is_instance,
                     committed_instance_memory_mib=committed_instance,
                     committed_program_memory_mib=committed_program,
                     committed_vcpus=committed_vcpus,
@@ -380,7 +376,7 @@ class CapacityManager:
                     committed_program += kept[1]
                     committed_vcpus += kept[2]
                 continue
-            if is_instance:
+            if requirements.is_instance:
                 committed_instance += requirements.memory_mib
             else:
                 committed_program += requirements.memory_mib
--- a/src/aleph/vm/agent/capacity.py
+++ b/src/aleph/vm/agent/capacity.py
@@ -273,9 +273,11 @@ class CapacityManager:
 
         Cumulative: each accepted candidate is committed before the next is
         judged, so three VMs that only fit twice get two yeses and one no. This
-        covers memory, vCPUs and disk. Disk needs the accumulator because free
-        space is read live and no record of it exists until the volumes are
-        actually written.
+        covers memory, vCPUs and aggregate disk. Disk needs the accumulator
+        because free space is read live and no record of it exists until the
+        volumes are actually written. Aggregate only: the per-volume check asks
+        whether the roomiest pool could hold the largest volume, and which pool
+        a volume lands on is a placement decision nothing models here.
 
         A candidate's own registry record never counts against it. A hash can
         already be recorded here and still be a candidate: a recreate, or an
@@ -304,6 +306,11 @@ class CapacityManager:
         memory alone. An advisory yes must never be read as one that covered
         the GPU.
 
+        Releases are not credited back for cards either, and for the same
+        reason as disk: the inventory is the cards not currently attached, so a
+        card still held by a VM the plan stops is absent from it. "Stop B,
+        allocate C onto B's card" is answered no even though doing it in that
+        order would work.
 
         The third element of a candidate is the caller's memory-bucket choice,
         NOT ResourceRequirements.is_instance. The two disagree for V-PROGRAMs:
--- a/src/aleph/vm/agent/capacity.py
+++ b/src/aleph/vm/agent/capacity.py
@@ -266,6 +266,7 @@ class CapacityManager:
         candidates: list[tuple[ItemHash, ResourceRequirements, bool]],
         *,
         releasing: frozenset[ItemHash] = frozenset(),
+        available_gpus: list[GpuDevice] | None = None,
     ) -> list[AdmissionVerdict]:
         """Judge a whole plan at once.
 
@@ -284,10 +285,17 @@ class CapacityManager:
         Side-effect free: nothing here reserves or holds anything, which is
         what makes it safe for the speculative capacity-check endpoint.
 
-        GPUs are out of scope. Card availability comes from the supervisor's
-        async host info and is claimed through stateful holds, neither of which
-        fits a synchronous side-effect-free call, so a GPU plan needs a
-        separate answer.
+        GPUs are judged from ``available_gpus``, the host's unattached cards,
+        which the caller reads from the supervisor because that read is async
+        and this call is not. Matching is cumulative like the rest: two
+        candidates wanting the only card of a kind get one yes. A card under a
+        live hold counts as taken, since the hold is some user's pending
+        create.
+
+        Without ``available_gpus`` there is no inventory to judge against, so a
+        candidate that asks for a card is refused rather than admitted on
+        memory alone. An advisory yes must never be read as one that covered
+        the GPU.
 
         The third element of a candidate is the caller's memory-bucket choice,
         NOT ResourceRequirements.is_instance. The two disagree for V-PROGRAMs:
@@ -313,6 +321,8 @@ class CapacityManager:
         committed_program = max(committed_program, 0)
         committed_vcpus = max(committed_vcpus, 0)
 
+        gpu_pool = None if available_gpus is None else self._unheld_gpus(available_gpus)
+
         verdicts: list[AdmissionVerdict] = []
         committed_disk = 0
         for vm_hash, requirements, is_instance in candidates:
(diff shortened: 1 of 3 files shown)


foxpatch-aleph (commented):
Clean refactor with well-tested new functionality. Extracting _check_against preserves the 30 existing tests, and the 4 new simulate tests cover cumulative admission, release-awareness, disk checking, and side-effect freedom. The bucketing logic in the release loop correctly mirrors _committed_resources. One substantive observation: disk is not tracked cumulatively in simulate — each candidate sees the live filesystem available disk without subtracting previously-accepted candidates' disk — so a plan with two disk-heavy VMs could get two yeses where sequential check_capacity calls would reject the second. This contradicts the stated principle that an advisory answer can never be stronger than the enforced one, but it's a limitation of the live-filesystem disk model (disk is not a committed sum), not a logic bug. Worth a docstring note so future readers don't assume full cumulative parity.

**`src/aleph/vm/agent/capacity.py`** (line 262): The docstring says 'an advisory answer can never be stronger or weaker than the enforced one,' but this is not fully true for disk: _check_against reads available disk from the live filesystem each time, and accepted candidates' disk_mib is not subtracted from any running total. Two candidates each needing 80% of available disk both pass. Memory and vCPUs are cumulative; disk is not. Consider adding a note to the simulate docstring clarifying that cumulative tracking applies to memory and vCPUs only, and that disk is a point-in-time check per candidate.

**`src/aleph/vm/agent/capacity.py`** (line 266): The release loop subtracts memory and vCPUs but not disk. This is consistent with the fact that disk is a live filesystem figure (not a committed sum), but it means 'allocate C, delete B' does not credit B's disk usage. Acceptable for an advisory endpoint, but worth noting in the docstring alongside the cumulative-disk caveat above.

foxpatch-aleph (approved):
Clean refactor that extracts the admission arithmetic into `_check_against` shared by both `check_capacity` and the new `simulate` method. The `simulate` method correctly implements cumulative (each acceptance committed before the next judgment) and release-aware (releasing VMs' resources subtracted from committed sums) batch admission. The bucketing logic in the release path matches `_committed_resources` exactly (InstanceContent + is_vprogram → instance bucket). The four new tests are well-designed with clearly explained math, covering cumulative admission, release-aware admission, disk rejection, and side-effect-freeness. The `AdmissionVerdict` DTO correctly avoids leaking internal error details to the scheduler. Minor non-blocking observations: disk space for releasing VMs is not subtracted (conservative but safe for an advisory call), committed values are not clamped to non-negative after release subtractions (would only matter with a registry inconsistency), and there's no test for releasing a non-existent hash or a combined cumulative+release scenario.

**`src/aleph/vm/agent/capacity.py`** (line 275): Minor: committed values are not clamped to non-negative after release subtractions. If a `releasing` hash has resources exceeding the committed sum (a registry inconsistency), `committed_instance`/`committed_program`/`committed_vcpus` could go negative, making admission too permissive. In practice this shouldn't happen with a consistent registry, but a `max(value, 0)` clamp would be defensive. Not blocking.

**`src/aleph/vm/agent/capacity.py`** (line 265): Worth noting in the docstring: `simulate` does not subtract disk space for releasing VMs. Disk is checked against physical free space (`_available_disk_bytes`), not a committed accumulator, so a plan that says "delete B (freeing disk) then create C (needing disk)" may be rejected if current free disk is insufficient even though the release would make room. This is the conservative/safe choice for an advisory call, but documenting it would help future readers understand the asymmetry between memory/vCPU (release-aware) and disk (not).

**`tests/supervisor/test_agent_capacity.py`** (line 542): Test coverage gap: no test for (a) releasing a hash not in the registry (the `record is None` guard at line 268 handles it but is untested), and (b) a combined cumulative+release scenario — e.g., release B, then admit two candidates where the freed space only fits one. These are minor gaps given the existing tests already prove the core mechanics.

foxpatch-aleph (commented):
The refactor is clean and correct: _check_against cleanly shares the admission arithmetic between check_capacity and simulate, the bucketing logic in the release loop matches _committed_resources, and cumulative accounting works as described. The main issue is that test_simulate_reserves_nothing is vacuous — it asserts holds == {} but simulate never touches holds, so the assertion passes regardless of whether simulate has side effects. There's also a coverage gap: no simulate test exercises the is_instance=False (program bucket) path. GPU requirements are silently ignored by simulate, which is consistent with check_capacity but undocumented.

**`tests/supervisor/test_agent_capacity.py`** (line 552): This test is vacuous: `manager.holds` starts as `{}` and `simulate()` never touches the holds dict, so `assert manager.holds == {}` passes regardless of whether simulate has side effects. To actually prove simulate leaves no trace, assert against the registry — e.g. that `manager.registry._records` is unchanged, or that `check_capacity` returns the same answer before and after a `simulate` call.

**`tests/supervisor/test_agent_capacity.py`** (line 485): The `_requirements` helper hardcodes `is_instance=True`, so all four simulate tests only exercise the instance-bucket path. The `is_instance=False` branch (program bucket) in simulate (line 298-299: `committed_program += ...`) is untested. Consider adding a test with a program candidate to cover that branch.

**`src/aleph/vm/agent/capacity.py`** (line 254): The docstring describes simulate as cumulative, release-aware, and side-effect free, but doesn't mention that GPU requirements (ResourceRequirements.gpu_device_ids) are silently ignored. This is consistent with check_capacity (which also doesn't check GPUs), but a caller reading the docstring might expect simulate to cover all resource dimensions. A one-liner noting that GPU admission is handled separately would help.

foxpatch-aleph (commented):
The refactor is clean and well-tested: extracting _check_against preserves the existing 30 tests, and the new simulate() method correctly implements cumulative memory/vCPU admission with release-awareness. The AdmissionVerdict DTO is well-designed. However, disk accounting is not cumulative in simulate() — each candidate is checked against live filesystem state rather than an accumulated disk figure, which contradicts the docstring's claim that 'each accepted candidate is committed before the next is judged.' This means a plan with two disk-heavy VMs that individually fit but jointly exceed free disk would be over-admitted. This may be acceptable for an advisory endpoint (disk state is hard to simulate without side effects), but the docstring should note the limitation. There is also no test covering the cumulative disk scenario.

**`src/aleph/vm/agent/capacity.py`** (line 256): The docstring says 'Cumulative: each accepted candidate is committed before the next is judged,' but this is only true for memory and vCPUs. Disk is checked against live filesystem state via _available_disk_bytes() on each call to _check_against, so accepted candidates' disk requirements are NOT subtracted before judging the next one. Two VMs that each fit in the available disk individually but not together would both be admitted. Consider either tracking an accumulated disk figure (like committed_instance/committed_vcpus) or qualifying the docstring to note that disk cumulativeness is not enforced.

**`tests/supervisor/test_agent_capacity.py`** (line 530): test_simulate_judges_disk only tests a single candidate exceeding disk limits. There is no test for the cumulative disk scenario — e.g., two candidates each with disk_mib within the limit, but together exceeding it. If disk cumulativeness is intentional, this test should document the limitation; if it's a gap, a cumulative disk test should be added alongside test_simulate_is_cumulative (which covers memory).

foxpatch-aleph (changes requested):
The refactor itself is clean and behavior-preserving (verified: all 34 tests pass, and the test arithmetic checks out against the current reservation defaults), and the release-loop bucketing is provably consistent with _committed_resources. The blocking issue: simulate's cumulativity only holds for memory and vCPUs — the disk check inside _check_against reads the live free-space figure fresh on every candidate, so a plan whose aggregate disk exceeds free space is over-admitted, which is exactly the independently-judged-candidate failure this PR was written to eliminate for memory. Secondary: the candidate tuple's is_instance element is a V-PROGRAM trap (it contradicts requirements_from_message's own is_instance for that content type), and GPU requirements are silently ignored. All three are small, actionable fixes or docstring qualifications, and the consuming endpoint hasn't landed yet so now is the cheapest time to address them.

**`src/aleph/vm/agent/capacity.py`** (line 256): The "Cumulative" claim holds for memory and vCPUs but not disk: _check_against recomputes available_disk_mib from live free space on every call (line 200), and nothing in the loop reduces it. A plan of three candidates each wanting 40 GiB against 50 GiB free therefore gets three yeses — the same independently-judged-candidate failure the PR description criticizes. Note the release loop can't compensate either: disk isn't a committed sum, so "delete B, whose disk is in use" doesn't add B's space back (that direction is at least conservative). Since the verdicts feed a speculative capacity-check endpoint, this is a wrong-answer bug rather than a safety one, but it contradicts this docstring. Either thread a running disk commitment through the loop (an optional available_disk_mib parameter on _check_against), or qualify the docstring and add a test pinning the current per-candidate disk behavior so the limitation is explicit before the endpoint lands.

**`src/aleph/vm/agent/capacity.py`** (line 250): The third tuple element duplicates ResourceRequirements.is_instance, and for V-PROGRAMs the two disagree: requirements_from_message sets is_instance=False for VerifiableProgramContent (it's isinstance(content, InstanceContent)), while the create path passes is_instance=True for V-PROGRAMs — that's exactly why _committed_resources buckets V-PROGRAMs into the instance bucket. A future endpoint author building candidates from requirements_from_message could naturally use requirements.is_instance for this element and silently mis-bucket V-PROGRAMs into the program bucket. Consider either deriving the bucket unambiguously (e.g. a single bucketed flag populated once at construction, so caller and message can't disagree) or documenting that the tuple element is the caller's bucketing decision which overrides requirements.is_instance.

**`src/aleph/vm/agent/capacity.py`** (line 262): ResourceRequirements.gpu_device_ids is ignored entirely, so simulate answers "fits" for plans requiring GPUs without consulting card availability (understandable — availability comes from the supervisor's async HostInfo and holds are stateful, which would fight the side-effect-free guarantee). Worth one docstring line saying GPU admission is out of scope, so the endpoint author knows it must be answered separately.

**`tests/supervisor/test_agent_capacity.py`** (line 513): The release test only covers memory subtraction; the release loop also subtracts vcpus and has an untested program-bucket branch (the else at capacity.py line 274). A plan that fits only after a release of a program-bucket VM, and one where the vCPUs are the binding constraint, would close both gaps.

foxpatch-aleph (commented):
Solid, carefully-reasoned change. The extraction of the admission arithmetic into _check_against is behavior-preserving for check_capacity (committed_disk_mib defaults to 0 and _available_disk_bytes is already clamped, so the old path is bit-identical), and simulate's cumulative/release-aware logic is correct: release bucketing matches _committed_resources exactly, negative sums are defensively clamped, nothing is mutated, and the exact-fit test figures all check out against the conf.py defaults (2048/8192/4.0). No correctness or security issues found. The suggestions below are test-robustness and API-shape improvements, none blocking.

**`src/aleph/vm/agent/capacity.py`** (line 305): simulate() silently accepts candidates whose requirements carry gpu_device_ids. The docstring scopes GPUs out, but the verdict gives a caller no way to tell 'GPU requirement not checked' apart from 'fits' — a scheduler reading accepted=True from the upcoming capacity-check endpoint could take it as a green light for a GPU VM. Consider refusing such candidates with a distinct code (e.g. gpu_unsupported) so an advisory yes can never be mistaken for one that covered the GPU.

**`src/aleph/vm/agent/capacity.py`** (line 368): Worth a note somewhere: this reads live pool free bytes and ignores committed_disk_mib, so a plan can pass simulate() yet fail at create — e.g. two pools with 400 GiB free each and two candidates each needing one 400 GiB volume: the aggregate check passes for both, the roomiest-pool check still passes for the second (nothing is written during simulate), yet at create the second finds no pool with room. Modeling pool placement is presumably out of scope for an advisory call, but it is a small hole in the 'advisory never stronger than enforced' invariant stated in _check_against's docstring.

**`src/aleph/vm/agent/capacity.py`** (line 132): Nit: as written this reads as if no create path passes disk_mib=0 anymore, but the post-download re-checks still do (run.py:447, run.py:506, run.py:590, migration/runner.py:333) — it's the _admit() calls that now pass the volume total. Something like 'the create paths pass the message's volume total at admission (#1153); their post-download re-checks pass 0' would be unambiguous.

**`tests/supervisor/test_agent_capacity.py`** (line 549): This test cannot detect a committing implementation: against the ~48 GiB instance bucket, a second call that saw the first candidate's 2048 MiB committed would still accept, so both calls pass either way. Assert the state directly — len(registry) == 0, manager.registry.get(_HASH_A) is None, holds == {} — or size the candidate so it only fits once, so the second call would actually flip.

**`tests/supervisor/test_agent_capacity.py`** (line 499): These figures depend on HOST_MEMORY_RESERVED_MIB/PROGRAM_MEMORY_RESERVED_MIB being 2048/8192. Older tests in this file pin both with mocker.patch.object whenever their arithmetic depends on them (see test_committed_memory_from_registry_is_bucketed); two lines here would make the new tests immune to a defaults change instead of relying on the PR description's note to recompute.

foxpatch-aleph (commented):
Correct and well-documented batch admission: the _check_against extraction provably preserves check_capacity (30 pre-existing tests unchanged), the cumulative disk accumulator fixes a real gap (free space has no committed sum), and the release logic is sound — frozenset prevents double-subtraction, the max(..., 0) clamp bounds registry inconsistency, and the release-loop bucketing is exactly equivalent to _committed_resources' isinstance check. Test arithmetic verified (48 GiB -> 38912 MiB, 40 GiB -> 30720 MiB) and the cumulative/release/disk tests genuinely discriminate against the bugs they target. All findings are non-blocking: the 'reserves nothing' test is non-discriminative (a leaking implementation still passes on its 55296 MiB bucket), the tuple-with-bool API forces callers to re-derive the V-PROGRAM bucketing that ResourceRequirements cannot express, GPU candidates are silently admitted despite gpu_device_ids being carried on the requirements, and the PR description's test count is stale (35/5, not 34/4).

**`tests/supervisor/test_agent_capacity.py`** (line 542): This test cannot detect the bug it is named for. On the 64 GiB host the instance bucket is 64*1024 - 2048 - 8192 = 55296 MiB, so even if simulate() leaked the first candidate's 2048 MiB commitment, the second identical call would still see 2048 + 2048 <= 55296 and answer True. Size the candidate above half the bucket (e.g. memory_mib=30000, so a leak flips the second verdict to False), or assert directly: `assert len(registry) == 0 and manager.holds == {}` after the call.

**`src/aleph/vm/agent/capacity.py`** (line 253): The third tuple element is a footgun the docstring has to spend six lines defusing: a caller passing requirements.is_instance (the obvious choice) silently mis-buckets V-PROGRAMs. Consider accepting ExecutableContent here and deriving both the requirements and the bucket internally (or a small NamedTuple with a `memory_bucket` field) so the V-PROGRAM rule lives in one place instead of being the caller's burden.

**`src/aleph/vm/agent/capacity.py`** (line 292): This bucketing predicate now exists in three forms: _committed_resources uses `isinstance(record.message, (InstanceContent, VerifiableProgramContent))`, here it is `isinstance(..., InstanceContent) or record.is_vprogram`, and the caller supplies a raw bool for candidates. A shared helper (e.g. `memory_bucket(content) -> bool`) would keep the enforced and advisory answers from drifting apart — which is the stated goal of the _check_against extraction.

**`src/aleph/vm/agent/capacity.py`** (line 305): Candidates carrying requirements.gpu_device_ids are admitted on memory/vCPU/disk alone — the GPU request is silently ignored, so the verdict reads `accepted: true` for a plan the CRN may have no cards for. Since the verdict is all a remote scheduler sees, consider refusing GPU candidates with a distinct code (e.g. `gpu_unchecked`) so the docstring's "out of scope" caveat is machine-readable rather than relying on the endpoint author having read it.

foxpatch-aleph (approved):
Solid, well-documented addition. The _check_against extraction is behavior-preserving (all 30 pre-existing tests pass unchanged through it), and simulate's semantics — cumulative admission, release-aware memory/vCPU accounting, deliberately conservative disk treatment, side-effect-free — are carefully reasoned and documented, including the V-PROGRAM is_instance footgun. The release-loop bucketing is exactly equivalent to _committed_resources, and the post-subtraction clamp is correctly placed. The four comments below are documentation nits, none blocking.

**`src/aleph/vm/agent/capacity.py`** (line 133): "which every create path now passes" is slightly broad: the migration runner's create path (migration/runner.py:333) still passes disk_mib=0. That is arguably fine since the disk is already staged by the download before admission, but consider narrowing the wording or noting the exception.

**`src/aleph/vm/agent/capacity.py`** (line 260): "This covers memory, vCPUs and disk" slightly overstates: the max-volume check (_check_max_volume, the roomiest-pool constraint) is not cumulative. Two candidates whose largest volume each fit the current roomiest pool can both be accepted even though creating the first shrinks the pool for the second. Pool placement isn't modeled here, so a one-word qualification ("aggregate disk") would make the docstring exact.

**`src/aleph/vm/agent/capacity.py`** (line 305): Unlike check_capacity, which documents its exclude_vm_hash rationale, simulate doesn't say what happens when a candidate's hash already has a registry record: the old record counts against the candidate unless it is also listed in releasing. Worth one docstring line before the upcoming endpoint PR builds on this contract.

**`tests/supervisor/test_agent_capacity.py`** (line 491): The PR description says "34 pass, of which 4 are new", but the branch head has 35 passing / 5 new after the final commit added test_simulate_is_cumulative_on_disk_too. Consider updating the description so reviewers running the suite don't get confused by the count.

odesenfans:
Rebased onto `dev-2.1`, and worked through the review tail. The disk-cumulativity finding that four reviews raised was already fixed by the second commit; the rest is below.

## GPU candidates are now actually judged

Rather than a docstring caveat or a `gpu_unchecked` code, `simulate` now checks availability for real. The obstacle I had claimed (availability is an async supervisor read, holds are stateful) turned out to be an argument about *where* the read happens, not whether it can happen: the caller passes the host's unattached cards in, and the matching is a pure function here.

It behaves like the rest of the simulation:

- **Cumulative** - cards are consumed from the pool, so two candidates wanting the only card of a kind get one yes.
- **All or nothing per candidate** - one that cannot get every card it asked for takes none, so it does not strand a card the next candidate could use.
- **Taken last** - only after memory, vCPU and disk have cleared, so a candidate refused on memory does not walk off with the cards.
- **A held card counts as taken**, since a hold is some user's pending create. Expired holds read as free but are *not* reaped: reaping is a write, and this is the call that must not touch the ledger.

With no inventory passed, a GPU candidate is refused with `gpu_unavailable` rather than admitted on memory alone. **This means #1172 must supply the inventory before GPU plans become placeable** - noted here so it does not get lost.

## Test hygiene

`test_simulate_reserves_nothing` could not fail. Its 2048 MiB candidate against a 55296 MiB bucket meant even a leaking implementation answered yes twice (`2048 + 2048` still fits). Now sized to 30000, where a leak flips the second verdict, plus direct assertions on the registry and the hold ledger. Confirmed the size is discriminative: two such candidates in one call answer `[True, False]`.

Also pinned `HOST_MEMORY_RESERVED_MIB` / `PROGRAM_MEMORY_RESERVED_MIB` in the tests whose arithmetic depends on them, matching what the older tests in this file already do, and added the missing release cases (a released hash with no record, and a release where vCPUs rather than memory bind).

Every new test was mutation-checked. Not consuming from the GPU pool fails only the cumulativity test; taking cards before the capacity check fails only the refused-on-memory test.

## Bucketing

Took the shared-helper suggestion. `is_instance_bucket(content)` now holds the V-PROGRAM rule, and the committed sums, the release loop and the docstring all defer to it. Worth noting the two predicates were already provably identical (`is_vprogram` is exactly `isinstance(message, VerifiableProgramContent)`), so this is a pure de-duplication with no behavior change. I left the candidate tuple shape alone.

## Corrected

The `_check_against` docstring claimed `disk_mib > 0` is what "every create path now passes". Five call sites still pass 0 (the post-download re-checks and the migration runner). Reworded.

## Not taken

The pool-placement hole. Genuine limitation - per-pool free space ignores `committed_disk` - but the worked example does not demonstrate it: two 400 GiB pools with two 400 GiB candidates place fine, one per pool. Modeling placement inside a side-effect-free advisory call is disproportionate, and the aggregate accumulator already stops the over-admission that bites in practice.

45 tests pass; `ruff format`, `isort`, `pyproject-fmt` and mypy clean.

foxpatch-aleph (approved):
Solid, well-tested addition. The arithmetic extraction into _check_against is clean, the cumulative/release-aware semantics are correct (accepted candidates commit memory/vCPUs/disk; GPU cards are consumed last and all-or-nothing; the caller's available_gpus list is copied, not mutated), and the 15 new tests are carefully reasoned with numbers derived from the config rather than magic. All 45 tests pass locally. The findings below are non-blocking contract and documentation observations for the upcoming capacity-check endpoint.

**`src/aleph/vm/agent/capacity.py`** (line 307): simulate() has no equivalent of check_capacity's exclude_vm_hash: a candidate hash that is already recorded on this CRN (e.g. a plan that keeps it here, or a recreate) is double-counted — its registry record plus its own candidate requirements. The caller can offset it by also listing the hash in releasing, but that is undocumented. Since the endpoint consuming this isn't built yet, consider either excluding each candidate's own registry record automatically, or adding a docstring line stating that candidates already recorded on this node must also appear in releasing.

**`src/aleph/vm/agent/capacity.py`** (line 288): Worth a docstring note: releases are not credited back for GPUs, only for memory/vCPUs. A card attached to a VM the plan stops does not appear in available_gpus (they are the *unattached* cards), so 'stop B, allocate C wanting B's GPU' is answered no even though create-after-delete would say yes. Same safe direction as the disk asymmetry, but unlike disk it isn't mentioned.

**`src/aleph/vm/agent/run.py`** (line 617): Minor: is_instance_bucket() was introduced so the rule is 'not restated wherever a bucket is picked', but run.py still states it literally here and at line 625 (is_instance=True for the V-PROGRAM path). These call sites could pass is_instance_bucket(content) instead.

**`src/aleph/vm/agent/capacity.py`** (line 353): Slight inconsistency with AdmissionVerdict's contract: for insufficient_capacity the detail is deliberately generic ('not enough capacity on this CRN') with the full text kept in logs, but for GPU refusals the raw gpu_refusal string is passed through. It contains no host figures so it is harmless, but if the intent is a uniform machine-readable detail, the two paths diverge.

**`tests/supervisor/test_agent_capacity.py`** (line 515): Test coverage nit: the documented release asymmetry (a released VM's disk is NOT credited back) and program-bucket releases (a released non-instance record freeing the program bucket) are both untested. Given the memory figures are hand-computed from config defaults, a regression there would be silent.

**`tests/supervisor/test_agent_capacity.py`** (line 491): Heads-up: the PR description says 34 tests pass with 4 new, but the diff adds 15 new simulate tests and 45 pass in total. The description looks stale; no code issue.

odesenfans:
All taken except one, and the description is updated (49 pass, 19 new).

## Candidate own-record exclusion

Confirmed and fixed properly rather than documented. A hash already recorded here and listed as a candidate refused itself: 16384 from its record plus 16384 as its own request against a 30720 MiB bucket.

`simulate` now excludes every candidate's record from the committed sums, the same discount `check_capacity` has always applied through `exclude_vm_hash`.

Worth noting why documenting it would not have been enough. The plan verdict in #1172 does compensate, via its `recreating` set, but that set is built only from hashes the supervisor still reports. An owner record left by a create that failed part way is invisible to it, so the caller could not fully enforce the rule it was being asked to remember.

Existing callers keep working: a hash that is both a candidate and in `releasing` is skipped by the release loop, since subtracting it again would credit memory nobody freed. That guard needed a specific test to be meaningful. My first attempt could not fail, because the `max(..., 0)` clamp absorbed the double subtraction; the committed test adds a second record so the sum stays positive and the bug is visible.

## Docstrings

Both qualifications taken. "Cumulative on disk" is now "aggregate disk", with the per-volume roomiest-pool check called out as the part that is not cumulative, and the GPU release asymmetry is written down next to the disk one it mirrors. Both behaviors are now pinned by tests, along with a program-bucket release.

## run.py

Taken, and slightly further than suggested: rather than passing `is_instance_bucket(content)` at each site, `_admit` derives it and the parameter is gone. That removes three call-site literals plus the docstring paragraph explaining the V-PROGRAM exception, and the post-download re-checks derive it too. I verified the substitution is value-preserving at every site (`ProgramContent` -> False, `InstanceContent` -> True, `VerifiableProgramContent` -> True), which matches the literals they replaced.

## Not taken

The GPU refusal detail passing through where `insufficient_capacity` is generic. That asymmetry is deliberate: the capacity detail was made generic to avoid leaking host memory figures, whereas `No available GPU matching device_id ...` contains no host state and tells the scheduler something it can act on, namely that this node is the wrong one for that card.

49 tests pass. `ruff format`, `isort`, `pyproject-fmt`, `yamlfix` and mypy clean; mypy on `run.py` is unchanged from its baseline, and the whole supervisor suite shows the same 38 pre-existing environment failures before and after.

foxpatch-aleph (changes requested):
Strong PR overall: the _check_against extraction is clean and behavior-preserving (all 49 tests pass, 30 pre-existing ones unchanged), the is_instance_bucket() centralization is a genuine improvement (verified behavior-preserving at every run.py call site), and the test suite for simulate() is unusually thorough and well-reasoned. However, there is one genuine bug in the new batch accounting: candidate hashes are excluded from the committed sums upfront, but the exclusion is never undone when a candidate is refused. For a recreate candidate that gets refused (e.g. on GPU), its still-running VM's memory stays invisible for the rest of the batch, so a later candidate can receive an advisory yes where the enforced check_capacity at create time would say no — contradicting the advisory-vs-enforced parity invariant this PR states twice. I reproduced this: on a 40 GiB host (30720 MiB instance bucket), A recorded at 16384 MiB and listed as a candidate, B new at 24576 MiB, no GPU inventory — simulate() refuses A on the GPU but accepts B, while the enforced create check for B refuses it (16384 + 24576 > 30720). The minimal fix is to re-add the refused candidate's registry record to the committed sums before continuing (the releasing loop already performs exactly that lookup), or, if the author deems this out of scope for the advisory endpoint, to document it as a known limitation in the simulate() docstring.

**`src/aleph/vm/agent/capacity.py`** (line 322): Excluding every candidate hash from the committed sums up front is only correct if every candidate is eventually accepted. When a candidate that already has a registry record — the recreate case the docstring explicitly supports — is refused (e.g. `gpu_unavailable`), its still-running VM's memory was deducted from the batch's committed sums and never restored, so later candidates are judged as if that VM did not exist. Reproduced on a 40 GiB host (instance bucket 30720 MiB): A recorded at 16384 MiB as a recreate candidate, B new at 24576 MiB, `simulate(candidates)` with no GPU inventory refuses A but accepts B; the enforced `check_capacity` for B's create then refuses it (16384 + 24576 > 30720). That is an advisory yes where the enforced answer is no, which the `_check_against` docstring ('an advisory answer can never be stronger or weaker than the enforced one') and the simulate docstring both claim cannot happen. Minimal fix: in both refusal branches, re-add the candidate's registry record's resources to the committed sums (the same `registry.get` + `is_instance_bucket` lookup the `releasing` loop already does) before `continue`. Alternatively, if this is judged acceptable for a speculative check, document the limitation here in the docstring.

**`src/aleph/vm/agent/capacity.py`** (line 373): Minor: `gpu_refusal` embeds the user-supplied `device_id` (an unvalidated free-form `str` in GpuProperties) into `detail`, while the `AdmissionVerdict` docstring says `code` and `detail` are safe to hand back to the scheduler and 'the full error text stays in the logs'. For GPU refusals the detail IS the full error text, so the insufficient_capacity verdict and this one treat 'detail' inconsistently. Either route GPU refusals through a generic, fixed detail string like the insufficient_capacity branch does, or relax the AdmissionVerdict docstring if echoing the requested device_id back to the scheduler is intended.

odesenfans:
Both taken. The first is a real bug and it was mine, introduced by the exclusion fix in the previous round.

## Refused candidates no longer stay discounted

Reproduced exactly as described before changing anything:

```
bucket 30720 MiB; A recorded at 16384 and refused on GPU; B wants 24576
  simulate                      -> [A refused (gpu_unavailable), B ACCEPTED]
  enforced check_capacity for B -> REFUSED
```

Discounting a candidate's own record is a bet that its request will replace it. When the candidate is refused the bet is off, and the rest of the batch was judged against a node emptier than it really is. The record now goes back on **both** refusal paths, not just the GPU one - the capacity branch had the identical hole, which the reproduction confirms.

One case worth naming, since it is the reason this is not a plain "always restore": a refused candidate that the plan is *also* stopping stays discounted, because the release has already accounted for it. All four combinations now behave:

```
A refused on GPU,       then B(24576) -> refused    (matches enforced)
A refused on capacity,  then B(24576) -> refused    (matches enforced)
A accepted,             then B(24576) -> refused    (unchanged)
A refused but released, then B(24576) -> accepted   (release frees it)
```

Reading what a record commits is now one helper, since the release loop wants the same figures with the opposite sign. Both mutations bite: removing the restore fails the two restoration tests, and dropping the `releasing` guard fails the release one.

## device_id no longer echoed

You changed my mind on this one. I had argued the specific reason was more useful than a generic string, but the point about it being unvalidated free-form text from the message is the part I had not weighed, and the scheduler already knows which device_ids it asked for, so echoing them back buys nothing. GPU refusals now carry a fixed detail and the specific reason stays in the logs, which is the same split the `insufficient_capacity` branch already made. `AdmissionVerdict`'s docstring is now true as written rather than needing to be relaxed.

53 tests pass; `ruff format`, `isort` and mypy clean.

foxpatch-aleph (approved):
Solid, carefully documented work. The simulate() logic is correct: I traced every test's arithmetic, probed the refusal-restoration edge cases with live code, and verified the _check_against extraction preserves check_capacity behavior (30 pre-existing tests pass unchanged). The run.py conversions are behavior-preserving (the persistent-program branch handles only ProgramContent, so is_instance_bucket is a no-op there), and the remaining call sites pick the bucket correctly. Security posture is good: the free-form device_id text is kept out of verdicts (explicitly tested) and confined to logs. The 23 new tests are all meaningful and several are engineered so the plausible implementation bug flips them. Only non-blocking nits remain, chiefly that the bare bool third tuple element is a footgun the type system can't catch.

**`src/aleph/vm/agent/capacity.py`** (line 267): Non-blocking design nit: the bare `bool` third element of the candidate tuple is a footgun the type system can't catch — a caller passing `requirements.is_instance` (which is `False` for V-PROGRAMs) compiles fine and quietly mis-buckets. The docstring warns loudly, but for the follow-up endpoint PR, consider carrying the bucket on `ResourceRequirements` itself (set by `requirements_from_message` via `is_instance_bucket`) so the mistake is impossible to express, or introducing a small candidate dataclass.

**`src/aleph/vm/agent/capacity.py`** (line 412): Worth remembering when the speculative endpoint lands: holds are user-agnostic here, so a card held by the plan's own user (e.g. taken via the reserve endpoint moments earlier) reads as taken and simulate answers `gpu_unavailable` where the real create would succeed. Conservative is the right default for an advisory answer, but the endpoint PR may want a story for this.

**`src/aleph/vm/agent/views/__init__.py`** (line 1016): Pre-existing and currently correct (this view validates messages as `InstanceContent`, so `requirements.is_instance` is always True), but it restates the bucket rule that `is_instance_bucket` was introduced to centralize. Consider migrating in a follow-up for uniformity.

**`src/aleph/vm/agent/capacity.py`** (line 345): Minor test-coverage nit: memory, disk and GPUs all have a direct cumulative test, but vCPUs are only exercised through the release path. The accumulation line shares the accepted-candidate path with memory so the risk is low; a two-candidates-vs-cap test would close it if you want symmetry.

odesenfans:
Approved, but the tuple footgun had come up in three separate reviews now, so I took the structural fix rather than leaving the docstring to guard it.

## The bucket now travels on the requirements

`requirements_from_message` sets `is_instance` from `is_instance_bucket`, so the field means "which memory bucket" rather than "is this an InstanceContent". The candidate tuple then drops the third element entirely, since it only ever existed to correct that field. Reaching for the obvious value is now the right answer, and the six-line docstring warning is gone.

I checked every consumer before changing it, since this alters what the field means:

- `views/__init__.py:1016` (the reservation dry-run) validates its body as `InstanceContent`, where both rules agree on `True`. No behaviour change - and it also resolves your third comment, since that call site now inherits the centralized rule instead of restating it.
- The create paths in `run.py` already chose the bucket by hand, so they are unaffected.

What does change: a V-PROGRAM reaching either path now reports the bucket it is actually admitted against. That was latent rather than live, but it was wrong.

The API change is a breaking one for #1172, which builds 3-tuples today. I am rebasing that PR next and it gets simpler there: the hand-rolled `isinstance(content, (InstanceContent, VerifiableProgramContent))` line and its explanatory comment both disappear.

Test evidence that the new shape is load-bearing: the program-bucket release test failed the moment the element was removed, because the bucket had to come from the requirements instead, and reverting `requirements_from_message` to the old `isinstance` rule fails the new V-PROGRAM test.

## vCPU coverage

Added the direct cumulative test. Dropping the accumulation line fails it and nothing else.

## Noted for the endpoint PR, not changed here

Holds being user-agnostic, so a card the plan's own user reserved moments earlier reads as taken and `simulate` answers `gpu_unavailable` where create would succeed. Conservative is right for an advisory answer, and fixing it means threading an owner through, which belongs with the endpoint that knows who is asking. Carried forward.

55 tests pass; `ruff format`, `isort`, `pyproject-fmt` and mypy clean, with mypy unchanged from its 5-error baseline. The full supervisor suite shows the same 38 pre-existing environment failures before and after.

foxpatch-aleph (approved):
Well-constructed PR. The cumulative, release-aware simulate() logic is correct in all the tricky interactions I traced: candidate records are excluded up front (replace semantics matching check_capacity's exclude_vm_hash), the release loop skips candidate hashes to avoid double-crediting memory, refused candidates get their records restored unless the plan also stops them, and the clamp-to-zero prevents registry inconsistencies from loosening admission. The _check_against extraction provably preserves check_capacity behavior (pre-existing tests pass unchanged; committed_disk_mib defaults to 0 so the disk expression is identical). The run.py centralization into is_instance_bucket is behavior-preserving — I verified VerifiableProgramContent is not a ProgramContent subclass and _is_spec_eligible matches InstanceContent only, so every call site's bucket is unchanged. Security is handled thoughtfully: verdict code/detail are fixed strings, the attacker-controlled device_id never reaches them (explicitly tested), and it stays in logs where !r escapes it. All 55 tests pass, plus the create-path suites. Only minor test-quality nits below; the aiohttp_client fixture errors in test_vprogram.py are a missing pytest-aiohttp plugin in this environment, not a PR issue.

**`tests/supervisor/test_agent_capacity.py`** (line 796): This test doesn't actually discriminate. With disk_bytes=10 GiB (10240 MiB) free and B's record carrying a 10000 MiB rootfs, an implementation that wrongly credited disk back on release would expose 20240 MiB — but C requests 50_000 MiB, so it's refused either way. A request in (10240, 20240] would make the test pin the behavior its docstring claims to cover. Same category of nit the PR description itself warns about ("a cumulative test that goes green because nothing fits proves nothing").

**`src/aleph/vm/agent/capacity.py`** (line 311): The documented behavior that releases are not credited back for GPU cards ("Stop B, allocate C onto B's card is answered no even though doing it in that order would work") has no test, unlike its disk twin. Since the docstring makes a specific behavioral claim, a test passing a released VM's card-less inventory while the released VM held the only card would pin it cheaply.

**`tests/supervisor/test_agent_capacity.py`** (line 925): Minor: the PR description says "49 pass, of which 19 are new", but the file has 55 tests with 25 new ones — the description appears stale. Not a code issue, but worth updating if the description is used as a changelog.

PR #1160, opened 2026-08-25
feat(agent): tear down loop and device-mapper devices when a VM is retired (2.1, PR C)

odesenfans:
## Summary

PR C of the 2.1 disk reclamation work (spec Section 5, S9: loop and device-mapper devices were never detached). Stacked on #1156.

- `storage.remove_devmapper(namespace, volume_name)`: the inverse of `create_devmapper`. `dmsetup remove --retry {ns}_{vol}`, `losetup -d` of every loop backed by the volume file (found by `losetup -j`, so a loop over another file is never touched), `rmdir /mnt/{ns}_{vol}`, and `dmsetup remove --retry {ns}_base` only when no other `{ns}_*` snapshot remains. The parent image's `/dev/mapper/{ref}` and its read-only loop are shared across VMs and are never removed here (the cache pass in PR D owns them).
- `retire.teardown_vm_devices`: `retire_vm` tears the devices down after `DeleteVm` returns and before the storage pass, for GONE / ERASE / FAILED_CREATE; RECREATE leaves them for the next create to reuse. Errors are logged, never raised, and never skip the storage release.
- In-place reinstall tears down after `stop_vm`, before the volume purge, so the `.btrfs` file is no longer pinned by a live target (the "still held by device mapper" branch in `purge.py` is now a real error path the reconciler retries).
- Name validation: the namespace keeps the strict item-hash check; the volume name is refused only for what `dmsetup` itself refuses (empty, `/`, `.`, `..`, over 127 bytes), so any name the create path accepted is torn down.
- `MOUNT_ROOT` hoisted into `storage.py` and shared with the reconciler.

## Known follow-ups
- `create_devmapper` builds one `{ns}_base` per namespace from the first parent-backed volume; a VM with two parent-backed volumes of different parents is already wrong on the create side (pre-existing). A volume literally named `base` collides with it. Teardown is correct either way.
- A stray loop over an already-unlinked file cannot be found via `losetup -j`.

## Test plan
- `tests/supervisor`: failing-ID set identical to base (compared by test id against the base commit in a throwaway worktree).
- New `tests/supervisor/test_devmapper_teardown.py` (exact command sequences and order, parent device never touched, `{ns}_base` sibling rule, awkward names, stale loop); `test_retire.py` (per reason, ordering, teardown failure does not abort the retire); `views/test_operator.py` (reinstall order teardown then purge).

### DIFF
--- a/src/aleph/vm/agent/views/operator.py
+++ b/src/aleph/vm/agent/views/operator.py
@@ -31,7 +31,7 @@ from aleph.vm.agent.vm.downloader import recreate_vm_volumes
 from aleph.vm.agent.vm.purge import purge_vm_staging, purge_vm_volumes
 from aleph.vm.agent.vm.reclaimable import retained_marker
 from aleph.vm.agent.vm.reconciler import creating
-from aleph.vm.agent.vm.retire import RetireReason, retire_vm
+from aleph.vm.agent.vm.retire import RetireReason, retire_vm, teardown_vm_devices
 from aleph.vm.agent.vm_registry import AgentVmRecord
 from aleph.vm.backup.archive import InsufficientDiskSpaceError
 from aleph.vm.backup.staging import download_volume_by_ref, get_backup_directory
@@ -767,6 +767,11 @@ async def operate_reinstall(request: web.Request, authenticated_sender: str) -> 
                 # the agent's job because the agent created them; the
                 # supervisor is handed resolved paths and never allocates.
                 await supervisor.stop_vm(vm_id)
+                # A parent-backed volume is a loop device under a dm
+                # snapshot: the purge below refuses to unlink its file while
+                # those are live, and create_devmapper would return the old
+                # device instead of rebuilding it.
+                await teardown_vm_devices(str(vm_hash), record)
                 deleted = await asyncio.to_thread(purge_vm_volumes, vm_hash, include_data_volumes=include_data_volumes)
                 try:
                     # Rebuild each volume on the pool it was deleted from: the
--- a/src/aleph/vm/agent/vm/reconciler.py
+++ b/src/aleph/vm/agent/vm/reconciler.py
@@ -55,6 +55,7 @@ from aleph.vm.agent.vm.reclaimable import (
 )
 from aleph.vm.agent.vm_registry import AgentVmRegistry
 from aleph.vm.conf import settings
+from aleph.vm.storage import MOUNT_ROOT  # /mnt/{namespace}_{volume name}
 from aleph.vm.storage_budget import parse_budget
 from aleph.vm.storage_pools import StoragePool, get_pools, iter_namespace_dirs
 from aleph.vm.supervisor_interface.abc import Supervisor
@@ -62,9 +63,6 @@ from aleph.vm.utils import create_task_log_exceptions
 
 logger = logging.getLogger(__name__)
 
-# Where create_devmapper mounts a volume while resizing it (storage.py mounts
-# /mnt/{namespace}_{volume name}).
-MOUNT_ROOT = Path("/mnt")
 STAGING_KINDS = ("vprogram", "snp-instance")
 
 # Namespace to the number of creates in flight for it. A count rather than
--- a/src/aleph/vm/agent/vm/retire.py
+++ b/src/aleph/vm/agent/vm/retire.py
@@ -30,10 +30,15 @@ from aleph_message.models import ItemHash
 
 from aleph.vm.agent.metrics import delete_records_for_vm
 from aleph.vm.agent.vm.backup import purge_vm_backups
-from aleph.vm.agent.vm.purge import purge_vm_side_dirs, purge_vm_storage
+from aleph.vm.agent.vm.purge import (
+    _checked_namespace,
+    purge_vm_side_dirs,
+    purge_vm_storage,
+)
 from aleph.vm.agent.vm.reclaimable import depends_on_from_content, mark_reclaimable
 from aleph.vm.agent.vm_registry import AgentVmRecord, AgentVmRegistry
 from aleph.vm.conf import settings
+from aleph.vm.storage import remove_devmapper
 from aleph.vm.supervisor_interface.abc import Supervisor
 from aleph.vm.supervisor_interface.errors import VmNotFoundError
 from aleph.vm.supervisor_interface.types import VmId
@@ -63,6 +68,31 @@ def set_after_gone_hook(hook: AfterGoneHook | None) -> None:
     """
     global _after_gone  # noqa: PLW0603
     _after_gone = hook
+
+
+async def teardown_vm_devices(namespace: str, record: AgentVmRecord | None) -> None:
+    """Remove the device-mapper snapshots and loop devices of a VM's
+    parent-backed volumes.
+
+    The agent creates them (``storage.create_devmapper``) and nothing else
+    removes them, so this is the only inverse. Best effort: a failure is
+    logged and the volume file stays behind, where the reconciler's dm guard
+    leaves it for the next pass rather than unlinking a file a loop device
+    still pins.
+    """
+    if record is None:
+        return
+    namespace = _checked_namespace(namespace)
+    for volume in getattr(record.message, "volumes", None) or []:
+        # An instance rootfs is a qcow2 overlay, not a dm snapshot, and it is
+        # not in `volumes` anyway; only a volume with a parent went through
+        # create_devmapper.
+        if getattr(volume, "parent", None) is None or not getattr(volume, "name", None):
+            continue
+        try:
+            await remove_devmapper(namespace, volume.name)
+        except Exception:
+            logger.exception("Device teardown of %s/%s failed", namespace, volume.name)
 
 
 async def retire_vm(
@@ -98,6 +128,10 @@ async def retire_vm(
     record = registry.get(item_hash)
     registry.forget(item_hash)
     await delete_records_for_vm(str(vm_hash))
+    # Before the storage pass: a volume file held by a live dm target cannot
+    # be unlinked usefully, and the marker written for a kept volume would
+    # describe a size the loop device still pins.
+    await teardown_vm_devices(str(vm_hash), record)
     try:
         await asyncio.to_thread(_release_storage, str(vm_hash), reason, record)
     except Exception:
--- a/src/aleph/vm/agent/vm/retire.py
+++ b/src/aleph/vm/agent/vm/retire.py
@@ -76,13 +76,24 @@ async def teardown_vm_devices(namespace: str, record: AgentVmRecord | None) -> N
 
     The agent creates them (``storage.create_devmapper``) and nothing else
     removes them, so this is the only inverse. Best effort: a failure is
-    logged and the volume file stays behind, where the reconciler's dm guard
-    leaves it for the next pass rather than unlinking a file a loop device
-    still pins.
+    logged and never aborts the retire, which still drops the records and
+    releases the rest of the storage. The volume file then stays behind,
+    pinned by its loop device, and the purge's dm guard refuses to unlink it
+    (that would free nothing and would not reset the volume either, since
+    ``create_devmapper`` reuses a live device). Nothing retries the teardown
+    once the record is gone: such a device waits for an operator.
     """
     if record is None:
         return
-    namespace = _checked_namespace(namespace)
+    try:
+        namespace = _checked_namespace(namespace)
+    except ValueError:
+        # Unreachable from the callers, which pass an ItemHash, but the
+        # best-effort contract holds for the check too: a teardown that
+        # cannot run must not abort the retire between the forget and the
+        # storage release.
+        logger.exception("Device teardown of %r skipped", namespace)
+        return
     for volume in getattr(record.message, "volumes", None) or []:
         # An instance rootfs is a qcow2 overlay, not a dm snapshot, and it is
         # not in `volumes` anyway; only a volume with a parent went through
--- a/src/aleph/vm/storage.py
+++ b/src/aleph/vm/storage.py
@@ -32,11 +32,17 @@ from aleph_message.models.execution.volume import (
 )
 
 from aleph.vm.conf import settings
-from aleph.vm.storage_pools import volume_path_for
+from aleph.vm.storage_pools import find_existing_volume, volume_path_for
 from aleph.vm.utils import fix_message_validation, run_in_subprocess
 
 logger = logging.getLogger(__name__)
 DEVICE_MAPPER_DIRECTORY = "/dev/mapper"
+# Where create_devmapper mounts a volume to resize its filesystem.
+MOUNT_ROOT = Path("/mnt")
+# What may be composed into a dmsetup device name. create_devmapper builds
+# names out of a VM hash and a volume name straight from the message, so a
+# name outside this set is one dmsetup would have refused to create.
+DEVICE_NAME_PATTERN = re.compile(r"^[\w\-]+$")
 
 # Runtimes and base images can be several hundred MiB and are sometimes served by slow
 # gateways (e.g. IPFS). We must not cap the *total* transfer time, or large-but-steady
@@ -460,13 +466,78 @@ async def create_devmapper(
     )
     await create_mapped_device(mapped_volume_name, snapshot_table_command)
 
-    mount_path = Path(f"/mnt/{mapped_volume_name}")
+    mount_path = MOUNT_ROOT / mapped_volume_name
     mount_path.mkdir(parents=True, exist_ok=True)
     await resize_and_tune_file_system(path_mapped_volume_name, mount_path)
     await chown_to_jailman(path_image_device_name)
     await chown_to_jailman(path_mapped_volume_name_base)
     await chown_to_jailman(path_mapped_volume_name)
     return path_mapped_volume_name
+
+
+async def detach_loop_devices(backing_file: Path) -> list[str]:
+    """Detach every loop device backed by ``backing_file``.
+
+    ``losetup -j`` lines look like ``/dev/loop3: [2049]:12 (/path/to/file)``.
+    A file can be attached more than once (a create that was interrupted
+    between the loop setup and the dmsetup create leaves one behind), so
+    every match is detached, not just the first.
+    """
+    stdout = await run_in_subprocess(["losetup", "-j", str(backing_file)])
+    detached: list[str] = []
+    for line in stdout.decode().splitlines():
+        device = line.split(":", 1)[0].strip()
+        if not device.startswith("/dev/loop"):
+            continue
+        await run_in_subprocess(["losetup", "-d", device])
+        logger.info("Detached loop device %s of %s", device, backing_file)
+        detached.append(device)
+    return detached
+
+
+async def remove_devmapper(namespace: str, volume_name: str) -> None:
+    """The inverse of ``create_devmapper`` for one volume.
+
+    Removes the snapshot device, detaches the loop devices of its backing
+    file, drops the resize mount point, and removes the per-VM base device
+    once no snapshot of this VM is left. The parent image's device and its
+    read-only loop are shared across VMs and are removed by the cache pass
+    when the image is evicted, never here.
+
+    Until this runs, the volume file is pinned by its loop device: unlinking
+    it would free no space and would not reset the volume either, since
+    ``create_devmapper`` returns early while the dm device exists (which is
+    what ``purge._held_by_device_mapper`` guards against).
+    """
+    mapped_name = f"{namespace}_{volume_name}"
+    if not DEVICE_NAME_PATTERN.match(namespace) or not DEVICE_NAME_PATTERN.match(volume_name):
+        logger.error("Refusing to remove an implausible device-mapper name: %r", mapped_name)
+        return
+
+    mapper = Path(DEVICE_MAPPER_DIRECTORY)
+    if (mapper / mapped_name).is_block_device():
+        await run_in_subprocess(["dmsetup", "remove", mapped_name])
+        logger.info("Removed device-mapper target %s", mapped_name)
+
+    backing_file = find_existing_volume(namespace, f"{volume_name}.btrfs")
+    if backing_file is not None:
+        await detach_loop_devices(backing_file)
+
+    mount_path = MOUNT_ROOT / mapped_name
+    if mount_path.is_dir() and not mount_path.is_mount():
+        try:
+            mount_path.rmdir()
+        except OSError:
+            logger.warning("Could not remove %s", mount_path, exc_info=True)
+
+    # The base device is per VM but shared by all of its parent-backed
+    # volumes (create_devmapper builds it once, from the first one), so it
+    # only goes once the last snapshot of this VM is gone.
+    base_name = f"{namespace}_base"
+    siblings = [path for path in mapper.glob(f"{namespace}_*") if path.name != base_name and path.is_block_device()]
+    if not siblings and (mapper / base_name).is_block_device():
+        await run_in_subprocess(["dmsetup", "remove", base_name])
+        logger.info("Removed device-mapper base %s", base_name)
 
 
 async def get_existing_file(ref: str) -> Path:
--- a/src/aleph/vm/storage.py
+++ b/src/aleph/vm/storage.py
@@ -39,10 +39,13 @@ logger = logging.getLogger(__name__)
 DEVICE_MAPPER_DIRECTORY = "/dev/mapper"
 # Where create_devmapper mounts a volume to resize its filesystem.
 MOUNT_ROOT = Path("/mnt")
-# What may be composed into a dmsetup device name. create_devmapper builds
-# names out of a VM hash and a volume name straight from the message, so a
-# name outside this set is one dmsetup would have refused to create.
-DEVICE_NAME_PATTERN = re.compile(r"^[\w\-]+$")
+# A namespace is an item hash, validated here too (agent.vm.purge does the
+# same for its delete paths) because the cost of being wrong is removing
+# another VM's device. ASCII only: a device name is bytes, and \w would match
+# letters that are not.
+NAMESPACE_PATTERN = re.compile(r"^[0-9a-zA-Z]{16,128}$", re.ASCII)
+# DM_NAME_LEN is 128 bytes including the terminating NUL.
+DEVICE_NAME_MAX_BYTES = 127
 
 # Runtimes and base images can be several hundred MiB and are sometimes served by slow
 # gateways (e.g. IPFS). We must not cap the *total* transfer time, or large-but-steady
@@ -495,6 +498,27 @@ async def detach_loop_devices(backing_file: Path) -> list[str]:
     return detached
 
 
+def device_name_for(namespace: str, volume_name: str) -> str | None:
+    """The dm name ``create_devmapper`` builds, or None if it could not have.
+
+    ``PersistentVolume.name`` is an arbitrary string and nothing on the create
+    path narrows it, so the only names that cannot exist are the ones
+    ``dmsetup create`` itself refuses: empty, containing a ``/``, ``.``, ``..``
+    and anything over the device-mapper name limit. A volume called ``data.1``
+    or ``my data`` has a live device and must be removable; refusing it here
+    would strand its loop device and, through ``purge._held_by_device_mapper``,
+    its disk.
+    """
+    if not NAMESPACE_PATTERN.match(namespace):
+        return None
+    if volume_name in ("", ".", "..") or "/" in volume_name:
+        return None
+    name = f"{namespace}_{volume_name}"
+    if len(name.encode()) > DEVICE_NAME_MAX_BYTES:
+        return None
+    return name
+
+
 async def remove_devmapper(namespace: str, volume_name: str) -> None:
     """The inverse of ``create_devmapper`` for one volume.
 
@@ -509,14 +533,16 @@ async def remove_devmapper(namespace: str, volume_name: str) -> None:
     ``create_devmapper`` returns early while the dm device exists (which is
     what ``purge._held_by_device_mapper`` guards against).
     """
-    mapped_name = f"{namespace}_{volume_name}"
-    if not DEVICE_NAME_PATTERN.match(namespace) or not DEVICE_NAME_PATTERN.match(volume_name):
-        logger.error("Refusing to remove an implausible device-mapper name: %r", mapped_name)
+    mapped_name = device_name_for(namespace, volume_name)
+    if mapped_name is None:
+        logger.error("Refusing to remove an implausible device-mapper name: %r/%r", namespace, volume_name)
         return
 
     mapper = Path(DEVICE_MAPPER_DIRECTORY)
     if (mapper / mapped_name).is_block_device():
-        await run_in_subprocess(["dmsetup", "remove", mapped_name])
+        # --retry: udev and blkid settle on the device for a moment after the
+        # VM exits, and a first "device or resource busy" must not strand it.
+        await run_in_subprocess(["dmsetup", "remove", "--retry", mapped_name])
         logger.info("Removed device-mapper target %s", mapped_name)
 
     backing_file = find_existing_volume(namespace, f"{volume_name}.btrfs")
@@ -536,7 +562,7 @@ async def remove_devmapper(namespace: str, volume_name: str) -> None:
     base_name = f"{namespace}_base"
     siblings = [path for path in mapper.glob(f"{namespace}_*") if path.name != base_name and path.is_block_device()]
     if not siblings and (mapper / base_name).is_block_device():
-        await run_in_subprocess(["dmsetup", "remove", base_name])
+        await run_in_subprocess(["dmsetup", "remove", "--retry", base_name])
         logger.info("Removed device-mapper base %s", base_name)
 
 
--- a/src/aleph/vm/storage.py
+++ b/src/aleph/vm/storage.py
@@ -533,12 +533,32 @@ async def remove_devmapper(namespace: str, volume_name: str) -> None:
     ``create_devmapper`` returns early while the dm device exists (which is
     what ``purge._held_by_device_mapper`` guards against).
     """
+    if not NAMESPACE_PATTERN.match(namespace):
+        logger.error("Refusing device teardown for an implausible VM hash: %r", namespace)
+        return
     mapped_name = device_name_for(namespace, volume_name)
     if mapped_name is None:
-        logger.error("Refusing to remove an implausible device-mapper name: %r/%r", namespace, volume_name)
+        logger.error(
+            "Refusing device teardown of VM %s for a volume name dmsetup could not have created: %r",
+            namespace,
+            volume_name,
+        )
         return
 
     mapper = Path(DEVICE_MAPPER_DIRECTORY)
+    mount_path = MOUNT_ROOT / mapped_name
+    if mount_path.is_mount():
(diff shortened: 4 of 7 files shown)


foxpatch-aleph (approved):
Well-implemented device-mapper teardown that correctly reverses create_devmapper: snapshot removal before loop detach, base device removal only when no siblings remain, parent image devices never touched. Name validation matches dmsetup's own rules (not stricter), ordering is correct (quiesce → teardown → purge), errors are best-effort and never block the retire, and RECREATE correctly skips teardown. No security issues — all subprocess calls use exec (not shell), namespace validation mirrors the existing ItemHash pattern, and is_mount() guards rmdir. Test coverage is thorough across all the important properties. The only minor nit is that _checked_namespace in teardown_vm_devices is outside the per-volume try/except, but it can't raise in practice since the namespace is always a validated ItemHash.

**`src/aleph/vm/agent/vm/retire.py`** (line 85): The `_checked_namespace` call is outside the per-volume try/except below. If it raised (bad namespace), it would abort the retire after `registry.forget()` and `delete_records_for_vm()` have already run, skipping the storage pass. In practice this can't happen since the namespace comes from a validated `ItemHash`, but the docstring says "Best effort" while this one path isn't caught. Consider wrapping it or moving the validation inside the loop, or accept that it's a structural invariant guaranteed by the caller.

**`tests/supervisor/views/test_operator.py`** (line 1311): The test verifies `order == ["teardown", "purge"]` but doesn't track `stop_vm` to confirm it happens before teardown. The code path makes this structural (stop_vm is on the line above teardown), but tracking it would make the ordering assertion complete.

foxpatch-aleph (changes requested):
Solid design overall: remove_devmapper mirrors create_devmapper in the right order, the {ns}_base sibling rule is safe (an item hash can never contain '_', so the glob can't cross VMs), there is no injection surface (argv lists only, namespace strictly validated), and failure containment is correct. However, the reinstall fix only covers the in-place branch: the non-in-place branch (confidential/ephemeral/V-program VMs) still retires RECREATE (devices kept) and then purges, so a parent-backed volume there keeps a live dm target, the purge refuses its file, and the recreated VM runs on unreset volumes — the exact bug this PR fixes, one call site short. The unnamed-volume skip is a narrower version of the same leak after an agent restart.

**`src/aleph/vm/agent/views/operator.py`** (line 803): This branch still has the bug the in-place branch above was fixed for: `retire_vm(RECREATE)` deliberately leaves the devices in place, then `purge_vm_volumes` runs directly. For a confidential/ephemeral/V-program VM with a parent-backed volume, `_held_by_device_mapper` is true, the `.btrfs` file is kept, and `create_devmapper` early-returns the live device — so the VM is recreated on its old, unreset volumes while the reinstall reports 200. Consider `await teardown_vm_devices(str(vm_hash), record)` before the `purge_vm_volumes` call here, exactly as in the in-place branch.

**`src/aleph/vm/agent/vm/retire.py`** (line 90): A parent-backed volume with no `name` gets one at create time — `host_volumes_from_message` mutates it in place to `unnamed_volume_{i}` — so the live in-memory record sees the name and teardown works. But a record rehydrated from the DB after an agent restart has `name=None`, this guard skips it, and the `{ns}_unnamed_volume_{i}` device is never torn down, which pins its `.btrfs` file through the purge guard forever. Deriving the same `unnamed_volume_{i}` fallback (enumerate with the loop index) would close this window.

**`src/aleph/vm/storage.py`** (line 514): Nit: the docstring's claim is slightly off for "." and "..": since the name is always prefixed with `{namespace}_`, the mapped names `{ns}_.`/`{ns}_..` are ones `dmsetup create` would accept, so refusing them is stricter than create was. Harmless in practice, but because `remove_devmapper` returns before `detach_loop_devices` on refusal, a create that failed after the `losetup` step would leave a stranded loop for those (and for >127-byte) names. Also note `$` in `NAMESPACE_PATTERN` matches a trailing newline, inherited from `purge._ITEM_HASH_PATTERN` — consistent with existing code, so just an observation.

**`src/aleph/vm/storage.py`** (line 545): `dmsetup remove --retry` retries indefinitely while the device is open. `delete_vm`/`stop_vm` should have released the holders, and udev/blkid are transient, but a wedged process holding the target would hang `retire_vm` — and the ERASE/forget handlers that await it — forever. Was a bounded retry (or a timeout around the call) considered?

foxpatch-aleph (approved):
Correct and well-tested inverse of create_devmapper: device names are validated against exactly the constraints dmsetup itself imposes, teardown ordering (dm target, loop, mount point, per-VM base) is sound and guarded by the sibling rule, and the shared parent device is never touched. Teardown is wired into retire_vm and the in-place reinstall at the right points, with failures logged rather than raised so the storage pass always runs. Minor nits: a `_checked_namespace` raise in teardown could skip `_release_storage` (no regression since the purge paths raise the same error), teardown keyed on the updated message can strand devices dropped by an amend, and the namespace regex is duplicated across purge.py and storage.py.

**`src/aleph/vm/agent/vm/retire.py`** (line 85): `_checked_namespace` raises before the loop, so an implausible namespace aborts here and skips `_release_storage` — contrary to "never skip the storage release". No regression in practice (the purge paths raise the same ValueError for the same input), but worth a note in the docstring or a local catch.

**`src/aleph/vm/agent/vm/retire.py`** (line 86): Teardown keys on `record.message` (the updated message). If an amend dropped a parent-backed volume, its `{ns}_<name>` device and loop are stranded: no `volumes` entry names it anymore, and the stale `.btrfs` file stays pinned. Worth listing in "Known follow-ups" alongside the stray-loop case.

**`src/aleph/vm/storage.py`** (line 46): NAMESPACE_PATTERN duplicates `_ITEM_HASH_PATTERN` from purge.py with the same regex. Consider hoisting one shared validator into storage.py (or a common module) so the two guards cannot drift.

**`src/aleph/vm/agent/views/operator.py`** (line 774): In rootfs_only mode this tears down data-volume devices the purge will not touch; harmless (backing files survive and create_devmapper rebuilds a snapshot over them) but extra dmsetup churn on every rootfs-only reinstall. Could scope teardown to the volumes being purged.

foxpatch-aleph (approved):
Correct inverse of create_devmapper with verified naming consistency, correct ordering in both retire and in-place reinstall, a sound sibling rule for the shared {ns}_base device, ASCII-only validation that matches dmsetup's own refusal rules, and thorough per-reason tests. Minor nits only: the no-op consequence of a failed teardown on the reinstall path, slightly broad Path monkeypatching in tests, and a fixture naming collision.

**`src/aleph/vm/storage.py`** (line 536): When `device_name_for` returns None, the volume file stays pinned and `create_devmapper` reuses the old device — on the reinstall path that means "reinstall" silently does not reset the volume. It is consistent with the best-effort design, but the error message could note that consequence so operators know why a purge refusal persists.

**`tests/supervisor/test_devmapper_teardown.py`** (line 34): `Path.is_block_device` is monkeypatched for all Path instances; scoping it to `/dev/mapper` paths (like `glob` is) would make the fixture tighter and avoid a real device coinciding with a test name.

**`tests/supervisor/reclaim_fixtures.py`** (line 22): `pool1 = tmp_path / "mnt" / "nvme1"` collides with the monkeypatched `MOUNT_ROOT = tmp_path / "mnt"` in the teardown test; harmless now, but a future pool1-based test that creates a mount point would collide.

foxpatch-aleph (approved):
This is careful, well-documented code: remove_devmapper is a faithful inverse of create_devmapper (raw volume names, per-VM base rule, shared parent device never touched), the name validation matches exactly what the create path could have produced, and the ordering in both retire_vm and the reinstall path (quiesce -> teardown -> purge) is right. Tests are thorough and all 35 new/changed tests pass; the operator-view failures I saw are a sandbox artifact (verified via an unmodified test). Remaining comments are observations and questions, not blockers: the potentially unbounded `dmsetup remove --retry` inside the retire/reinstall paths, the fact that record-less namespaces (orphans) and volumes dropped by a message amendment never get device teardown, and two small DRY nits.

**`src/aleph/vm/storage.py`** (line 545): `dmsetup remove --retry` retries until the removal succeeds, with no cap the caller controls. For the transient post-exit udev/blkid window that's exactly right, but a device that is genuinely held forever (e.g. a leaked mount from `resize_and_tune_file_system` whose umount failed, or a wedged holder) would stall `retire_vm` indefinitely — and in the reinstall path this runs inside an HTTP request handler, so the operator's reinstall would hang. Consider bounding this (e.g. `asyncio.wait_for` around the call, logging and moving on, since teardown is already best-effort). Also worth noting `--retry` requires a fairly recent device-mapper userspace (added in 1.02.172, 2022); on older nodes every remove fails with `unknown option`, which silently reduces teardown to a log line.

**`src/aleph/vm/agent/vm/retire.py`** (line 84): Two classes of leftover devices never reach teardown: (1) a namespace with no registry record — orphan namespaces purged by the reconciler (reconciler.py:343, :528) call `purge_vm_storage` with no teardown, so a live dm target makes `_held_by_device_mapper` refuse the unlink on every pass with nothing ever removing the device; (2) a volume dropped from a message after its device was created, and the unnamed-volume case (`host_volumes_from_message` assigns `unnamed_volume_{i}` in place on the message object, which survives an agent restart only until the record is re-parsed from JSON). Since `remove_devmapper`'s device names are all `{ns}_*` and namespaces are validated item hashes, a fallback sweep of `/dev/mapper/{ns}_*` when the record is None (or after the per-volume loop) would close all of these. Is this PR D's scope? If so, a pointer here would help.

**`src/aleph/vm/storage.py`** (line 44): `NAMESPACE_PATTERN` duplicates `purge._ITEM_HASH_PATTERN` — two sources of truth for "a namespace is an item hash" that could drift apart. Consider reusing the purge one (as retire.py does) or defining the single pattern in one place and importing it in both.

**`src/aleph/vm/agent/vm/retire.py`** (line 34): Minor: importing the private `_checked_namespace` cross-module matches the reconciler's existing `_ITEM_HASH_PATTERN` import, but here the validation is also redundant — `remove_devmapper` re-validates the namespace via `device_name_for` before composing any device name. Harmless defense in depth, but it means a namespace that is not an item hash raises ValueError in `teardown_vm_devices` *outside* the per-volume try/except, which would skip `_release_storage` and `purge_vm_backups` after the record is already forgotten. All callers pass an ItemHash so this is theoretical, but folding `_checked_namespace` inside the per-volume try (or dropping it) would make the function trivially exception-free.

foxpatch-aleph (approved):
Solid, well-scoped PR. remove_devmapper is a faithful inverse of create_devmapper: correct ordering, a name check that accepts exactly what dmsetup accepts (no wider, no narrower), no shell or path-traversal surface (argv-only subprocess, no '/', '..' possible in composed names), and the {ns}_base sibling rule is sound including partial-teardown recovery. retire_vm ordering (after quiesce, before storage release, skipped for RECREATE) matches the design and failures never skip the storage pass. All new and touched tests pass, and the only suite failures are pre-existing environment issues in untouched files. The comments below are non-blocking.

**`src/aleph/vm/agent/views/operator.py`** (line 808): Non-blocking observation: in this delete+recreate branch (ephemeral/confidential/vprogram VMs) teardown is never run, `retire_vm(RECREATE)` deliberately keeps the devices, and `purge_vm_volumes` then refuses to unlink a dm-held file — so a parent-backed data volume of such a VM is silently never erased by a reinstall, and the recreated VM reuses it via `create_devmapper`'s early return. This is pre-existing behavior rather than a regression, but now that `teardown_vm_devices` exists it may deserve a follow-up.

**`src/aleph/vm/agent/views/operator.py`** (line 774): Worth considering: if `teardown_vm_devices` fails here (logged, swallowed), the purge below skips the dm-held file, `recreate_vm_volumes` re-attaches the old device, the VM restarts with its old data, and the caller gets a 200. The failure is fail-safe but silent from the operator's perspective — the 'erase' they asked for didn't happen. Returning a 5xx (or noting it in the response) when the teardown of a volume failed would make the no-op visible.

**`src/aleph/vm/agent/vm/retire.py`** (line 90): Minor: `host_volumes_from_message` defaults an empty volume name to `unnamed_volume_{i}` (in place, before `create_devmapper` runs), so a parent-backed volume whose stored name is empty actually owns a device named `{ns}_unnamed_volume_{i}` that this skip will strand. In practice the in-place mutation means the stored record already carries the defaulted name, so this should be unreachable — a one-line comment saying the create path always leaves a non-empty name would make that explicit.

**`src/aleph/vm/agent/vm/purge.py`** (line 129): This comment ("nothing tears those down on stop yet") is now superseded by this PR: devices are torn down at retire/reinstall time by `teardown_vm_devices`. Consider updating it to point at that function so the guard reads as a safety net rather than as the only mechanism.

**`src/aleph/vm/agent/vm/purge.py`** (line 104): Nit: `_held_by_device_mapper` composes `{namespace}_{volume.stem}` directly, duplicating the composition that `storage.device_name_for` now owns and documents. Having the two agree is load-bearing (teardown clears exactly what the guard blocks), so reusing one helper would prevent them drifting apart.

foxpatch-aleph (approved):
This PR correctly fills a real gap (loop and dm devices were never torn down) with a careful, well-documented inverse of create_devmapper. Ordering is right (snapshot before loop, teardown after DeleteVm and before the storage pass, base device only after the last sibling snapshot), name validation is justified against the actual create path (I verified the sanitization in get_volume_path is dead code for parent-backed volumes, so awkward-but-legal device names must be accepted), and there is no injection surface: every subprocess call is an argv list and the glob pattern is built from a hash-validated namespace. Teardown is best-effort with per-volume error isolation that never skips the storage release, and the pre-existing _held_by_device_mapper guard in purge.py remains as the backstop. Tests are thorough and pass locally. Remaining points are nits: the non-in-place reinstall branch (operator.py:807) retires as RECREATE, so an ephemeral VM with a parent-backed volume has its reinstall silently not reset that volume (safe but surprising); retire.py imports the private _checked_namespace from purge.py; and NAMESPACE_PATTERN duplicates purge.py's _ITEM_HASH_PATTERN.

**`src/aleph/vm/agent/views/operator.py`** (line 807): This branch retires as RECREATE, which deliberately skips device teardown, and then purges volumes. For an ephemeral or confidential VM with a parent-backed volume, purge_vm_volumes will refuse to unlink the dm-held file and the subsequent create_vm_execution... will get create_devmapper's early return of the old device — so the reinstall silently does not reset that volume. Safe (guarded and logged), but worth a comment here or a follow-up ticket since it contradicts 'reinstall = reset to initial state' for that edge case.

**`src/aleph/vm/agent/vm/retire.py`** (line 34): Importing the private _checked_namespace from purge.py reaches into another module's underscore namespace. Since both modules now validate the same item-hash shape, consider promoting it (e.g. a public checked_namespace in purge.py, or moving it next to the shared NAMESPACE_PATTERN) so the two checks cannot drift.

**`src/aleph/vm/storage.py`** (line 46): Minor: NAMESPACE_PATTERN is byte-for-byte the same regex as purge.py's _ITEM_HASH_PATTERN. The comment notes the parallel, but a single shared constant would prevent the two from drifting apart over time.

foxpatch-aleph (changes requested):
The overall design is sound: correct teardown ordering, safe subprocess usage, a namespace-validated glob that cannot cross VMs, and thorough, passing tests (all 36 new/changed unit tests plus the reconciler/storage suites pass here). However, the name validation violates the PR's own stated invariant: `device_name_for` applies dmsetup's *device-name* rules (refuse "", ".", "..") to the *volume name*, but `create_devmapper` builds the device from the composite `{ns}_{volume_name}` — and `{ns}_.` is a name dmsetup happily creates. The message model accepts volumes named ".", "", and even name=None with a parent set (verified against aleph_message), so those volumes get a working device on create but are refused on teardown (falsy names are skipped outright in `teardown_vm_devices`), stranding the dm target, its loop device, and — through `purge._held_by_device_mapper` — the volume file, permanently, which is exactly the leak this PR closes. Validating the composite name instead (reject any '/' in it and enforce the 127-byte limit; the composite can never be ""/"."/".." given the validated hash prefix) would make the code match its docstring.

**`src/aleph/vm/storage.py`** (line 514): This applies dmsetup's *device*-name rules to the *volume* name. What dmsetup refuses is the composite `{ns}_{volume_name}` — which is never literally ".", ".." or empty, since the namespace prefix is a validated hash. A volume actually named "." or "" (or with the model-default `name=None`, which `PersistentVolume(parent=...))` accepts) creates device `{ns}_.` / `{ns}_` / `{ns}_None` and file `.btrfs` / `None.btrfs` on the create path, but is refused here — stranding the device, its loop, and via `purge._held_by_device_mapper` the volume file, forever. That contradicts the docstring's contract ('any name the create path accepted is torn down'). Suggest validating the composite instead: reject `/` in the composite and the >127-byte length (as you already do), and drop the `("", ".", "..")` special case, since the composite can never be those.

**`src/aleph/vm/agent/vm/retire.py`** (line 90): Same asymmetry from the create side: `volume.name` is optional (`PersistentVolume` defaults it to None), and `create_devmapper` will happily build `{ns}_None` (and `None.btrfs`) for a parent-backed volume with no name — so this falsy-name skip strands that device too. If `device_name_for` validates the composite name as above, this guard can go (or restrict the skip to `name is None` while letting `remove_devmapper` decide on "").

**`src/aleph/vm/agent/views/operator.py`** (line 807): Non-blocking observation: on this branch (confidential/ephemeral VMs), `retire_vm(RECREATE)` deliberately leaves the devices in place and the `purge_vm_volumes` on the next line then skips the dm-held `.btrfs` file, so a parent-backed persistent volume of a confidential VM is never actually reset by a reinstall (the recreate's `create_devmapper` early-returns the old device with its old contents). This matches the base branch's behavior, so it's no regression — but given this PR wires teardown into the in-place branch, tearing down here (after the retire, before the purge) would close the same hole. Worth confirming it was a deliberate scope decision.

**`src/aleph/vm/storage.py`** (line 46): Nit: this duplicates `purge._ITEM_HASH_PATTERN` (identical regex, minus the ASCII flag) — consider importing one from the other so the two delete paths can never drift. Also, the comment about `re.ASCII` is slightly off: an explicit `[0-9a-zA-Z]` class matches the same bytes with or without the flag (re.ASCII only affects \w/\d/\b), so the flag is harmless but not load-bearing here.

foxpatch-aleph (commented):
A careful, well-tested inverse of create_devmapper: the teardown order (quiesce -> snapshot device -> loop detach -> mount point -> shared base) is sound, the name validation prevents touching other VMs' devices or the shared parent image, and all subprocess calls are exec'd without a shell. I verified the external behaviors the code depends on (dmsetup remove --retry exists, losetup -j exits 0 on no match despite check=True). The gaps I found are edge cases or pre-existing: the inconsistent error posture around _checked_namespace/teardown_vm_devices in retire_vm, a stricter-than-necessary refusal of volume names "", ".", "..", the un-handled non-in-place reinstall branch (ephemeral VMs with parent-backed volumes still don't reset), and unnamed volumes that are renamed at create time but skipped at teardown. None block the merge.

**`src/aleph/vm/agent/vm/retire.py`** (line 85): This `_checked_namespace` call is the only statement in `teardown_vm_devices` that can raise, and it sits both outside the per-volume try and (in `retire_vm`) outside the try that guards `_release_storage`. If it ever did raise, the retire would abort after `registry.forget`/`delete_records_for_vm` but before the storage release — the exact outcome the surrounding except blocks exist to prevent, and at odds with the "errors are logged, never raised, and never skip the storage release" contract. It is practically unreachable for a validated ItemHash, but for consistency consider folding it into the best-effort posture (or calling teardown_vm_devices inside the same try as `_release_storage`).

**`src/aleph/vm/storage.py`** (line 514): Small nit on the docstring's premise: the kernel rejects only `/` (and NUL) in dm names, so the *composed* names for volume_name `""`, `"."`, `".."` — `{ns}_`, `{ns}_.`, `{ns}_..` — are ones `dmsetup create` would accept, meaning these are names the create path could have produced and teardown now refuses (stranding the loop device and pinning the disk, the bug this PR fixes). `""` is unreachable in practice because `host_volumes_from_message` renames unnamed volumes, and `"."`/`".."` are pathological, so this is fine to keep — but the comment claims symmetry with what dmsetup refuses and it is slightly stricter than that. A one-line note that this is a deliberate conservative cut would help future readers.

**`src/aleph/vm/agent/views/operator.py`** (line 808): Worth a follow-up: this branch (ephemeral/confidential VMs, `_supports_in_place_reinstall` false) purges volumes with no `teardown_vm_devices` call, so for an ephemeral QEMU VM with a parent-backed volume the purge guard refuses the unlink, the volume file survives, and the subsequent create early-returns the old dm device — i.e. the reinstall silently does not reset that volume. Pre-existing behavior and out of this PR's stated scope, but it is the same bug class fixed for the in-place branch above; `RECREATE`'s no-teardown rule only really holds for the delete+recreate of the *execution*, not for the partial purge that follows it here.

**`src/aleph/vm/agent/vm/retire.py`** (line 90): The `not getattr(volume, "name", None)` skip can strand a device: `host_volumes_from_message` assigns `unnamed_volume_{i}` to nameless persistent volumes at create time, so `{ns}_unnamed_volume_0` can be live while the record's message still has an empty name (if the stored content object is not the one mutated at create). Rare, but since the whole point of this function is "the only inverse", a volume name fallback mirroring `host_volumes_from_message` would close it.

**`src/aleph/vm/agent/vm/retire.py`** (line 80): Minor doc nit: "leaves it for the next pass" suggests a retry, but nothing retries the device teardown — the next reconcile pass will hit the same dm guard and keep refusing forever (operator territory). The sentence is technically about the file, not the devices, but a reader could infer a self-healing loop that doesn't exist.

**`tests/supervisor/test_devmapper_teardown.py`** (line 63): The test named "in order" only asserts the order of the two `dmsetup remove` calls plus membership of the losetup calls; the interleaving (that the loops are detached *after* the snapshot device is removed — the property that keeps `losetup -d` from failing with "in use") is untested. Asserting on `commands` directly (e.g. full prefix ordering) would pin the actual guarantee.

foxpatch-aleph (approved):
Solid, well-tested PR. Device teardown ordering is correct (snapshot dm target removed before detaching loop devices; base device removal correctly guarded by sibling-scan; shared parent device never touched). device_name_for properly mirrors create-time constraints (rejects '/', '.', '..', empty, >127 bytes; namespaces validated as item hashes matching [0-9a-zA-Z]{16,128}). No injection risk: run_in_subprocess uses argv lists. All 14 new devmapper tests and 22 retire tests pass; only test file that fails is test_operator_retire.py, which fails identically for all 12 tests including pre-existing ones due to environmental constraints (requires root + /opt/firecracker/vmlinux.bin), not this PR. Four minor, non-blocking comments below.

**`src/aleph/vm/agent/vm/purge.py`** (line 158): The comment "nothing tears those down on stop yet" is now outdated since this PR adds teardown_vm_devices. The _held_by_device_mapper check should stay as a safety net, but the comment could be updated to reflect that this is now a defensive guard rather than the expected case.

**`src/aleph/vm/storage.py`** (line 850): Minor: _teardown_vm_devices silently returns when record is None (VM unknown to the registry). Consider documenting this behavior on the function, since callers may not expect a no-op.

**`src/aleph/vm/agent/views/operator.py`** (line 412): The non-in-place branch of operate_reinstall doesn't call teardown_vm_devices. This appears correct since ephemeral/confidential VMs have no persistent volumes, but worth confirming that's the intent.

**`tests/supervisor/test_devmapper_teardown.py`** (line 132): The '# noqa: F811' comment is unnecessary here (function call, not a redefinition); ruff RUF100 flags it.

foxpatch-aleph (approved):
This PR is correct and carefully engineered. remove_devmapper mirrors create_devmapper faithfully: subprocess calls use argv lists (no shell/option injection — the composed device name always starts with the strict-ASCII namespace, and '/' is rejected so path joins can't escape), the {ns}_base sibling rule is safe (parent image devices are bare item hashes, namespaces are fixed-length, and the 'base'-named-volume collision resolves correctly in either volume order), and dmsetup remove --retry is confirmed available on the bookworm supervisor image. losetup -j exits 0 with no match, so the common no-loop-attached case won't raise. Ordering (quiesce -> teardown -> storage release) is guaranteed and explicitly tested; teardown failures are logged without aborting the retire. All 14 new devmapper tests pass, as do test_retire.py and test_reconciler.py (test_operator_retire.py couldn't run in this sandbox due to a missing pytest-aiohttp plugin — an environment issue predating this PR). Remaining findings are non-blocking: nothing ever retries a failed teardown (the record is forgotten before teardown runs, so the reconciler's dm guard only prevents the worse bug of unlinking a pinned inode, it never heals), the ephemeral/confidential reinstall branch still purges without teardown, and a stray loop over an already-unlinked backing file is unreachable via find_existing_volume (acknowledged in the PR description).

**`src/aleph/vm/agent/vm/retire.py`** (line 79): Non-blocking: since `retire_vm` runs this after `registry.forget`, a failure here can never be retried — the reconciler's dm guard only prevents unlinking a pinned inode; it never re-attempts the `dmsetup remove` itself. The "stays behind ... for the next pass" wording reads as if the next pass will resolve it, but the devices and files are stranded until an operator intervenes. Consider rewording to say the failure is permanent (until manual action) so the severity is clear.

**`src/aleph/vm/storage.py`** (line 548): When the backing file was already unlinked (e.g. a purge raced ahead, or an earlier retire's purge partially completed), `find_existing_volume` returns None and the loop device pinning the inode is never detached — the blocks stay held with no way to reach them via `losetup -j`. This is acknowledged in the PR description as a known follow-up; a short code comment here would spare future readers the archaeology.

**`src/aleph/vm/agent/views/operator.py`** (line 807): The else-branch (ephemeral/confidential reinstall) still purges volumes without tearing devices down, so a parent-backed volume on those VMs won't reset on reinstall (purge refuses the held file, and create_devmapper returns the old device). Pre-existing behavior and not a regression from this PR, but worth a follow-up note now that the in-place branch is fixed.

**`src/aleph/vm/agent/vm/retire.py`** (line 34): Nit: importing the private `_checked_namespace` couples retire to purge's internals. Storage can't host it (purge imports storage, so hoisting there would be a cycle), but promoting it to a public `checked_namespace` in purge would read better and drop the underscore.

**`src/aleph/vm/storage.py`** (line 563): Minor: the `/dev/mapper` glob runs even when `(mapper / base_name)` is not a block device; checking `base_name`'s existence first would skip the directory scan in the common no-base case. Purely cosmetic.

foxpatch-aleph (approved):
Solid, carefully-scoped PR. remove_devmapper correctly inverts create_devmapper (snapshot → loops → mount point → base only when no siblings remain), never touches the shared parent device, and is injection-safe (create_subprocess_exec, strict namespace validation, no path-reachable volume names). Ordering relative to quiesce and the storage pass is right and explicitly tested in both retire_vm and the reinstall path, RECREATE correctly skips teardown, and failure handling is best-effort without skipping storage release, as tested. New tests are meaningful and pass locally (36 in the retire/teardown files, 42 in the reconciler); the operator-view test only fails locally for lack of root, which is an environment limitation. Findings are all minor: the early return on an implausible device name happens before the loop detach, so the narrow case of a create interrupted after losetup but before dmsetup create with an over-long (or slash-containing) name strands its loop device and {ns}_base with no recovery path; the purge.py comment saying nothing tears these down is now partially stale for the retire/reinstall paths; NAMESPACE_PATTERN duplicates purge._ITEM_HASH_PATTERN; and teardown_vm_devices sits outside the try/except guarding _release_storage, though the namespace check is practically unreachable for a validated ItemHash.

**`src/aleph/vm/storage.py`** (line 539): This early return skips the loop detach too. If create_devmapper attached the loop (storage.py:464) and then `dmsetup create` failed because the composite name exceeds 127 bytes or contains a `/`, the loop device and `{ns}_base` are stranded with no recovery: device_name_for returns None forever, and since no dm device exists, purge._held_by_device_mapper won't stop the volume file being unlinked while the loop still pins its inode. Loop detach only needs the backing file, not the device name — consider attempting `find_existing_volume` + `detach_loop_devices` even when the name is implausible. Edge case (needs a >~62-char volume name), not blocking.

**`src/aleph/vm/agent/vm/purge.py`** (line 137): This comment ("nothing tears those down on stop yet") is now partially stale: retire_vm and the reinstall path do tear them down, so this branch is an error path there, not the norm. Still accurate for a plain stop_vm, but a rewording would help — the PR description itself frames this as "a real error path the reconciler retries" now.

**`src/aleph/vm/agent/vm/retire.py`** (line 134): teardown_vm_devices sits outside the try/except that guards _release_storage, yet a raise here (e.g. _checked_namespace) would abort the retire before the storage release and the backup purge — inconsistent with the "never abort the caller's sweep" rationale documented at line 138-143. Practically unreachable since vm_hash was already validated as an ItemHash (>= 20 chars), so this is a robustness nit, not a bug.

**`src/aleph/vm/storage.py`** (line 46): Minor: this is the same regex as purge._ITEM_HASH_PATTERN, now defined twice. Also, the comment's "\w would match letters that are not" rationale doesn't match the code, which uses an explicit [0-9a-zA-Z] class rather than \w — the re.ASCII flag is a no-op here. Consider sharing one pattern (e.g. purge importing it) and trimming the comment.

foxpatch-aleph (approved):
A careful, correct inverse of create_devmapper: ordering (dm target before its loop devices, teardown after DeleteVm quiesce but before the storage pass) is right, the name-validation asymmetry is justified — I confirmed the create path builds device names from the raw volume.name (get_volume_path's sanitization is dead code on the parent-backed branch), so any name dmsetup accepted must be accepted at teardown — and there is no injection surface (argv lists, strict namespace pattern, no '/' in volume names, no '-'-prefixed names). I verified 'dmsetup remove --retry' and 'losetup -j' semantics on this system, ran the new tests (36 pass) plus adjacent purge/reconciler/storage modules (103 pass; the remaining suite failures are pre-existing environment issues unrelated to the diff), and checked the {ns}_base sibling rule and the shared-parent exclusion. Remaining comments are non-blocking: the docstring's 'leaves it for the next pass' framing overstates the recovery story for a failed teardown (after the record is forgotten, nothing ever re-attempts remove_devmapper, so a persistent busy device strands the target, loop, and pinned file until manual action), _checked_namespace raises outside the per-volume try in teardown_vm_devices (unreachable from current callers), a still-mounted mount point will keep dmsetup remove failing where a best-effort umount first would help, and the validation-failure log does not distinguish namespace from volume name.

**`src/aleph/vm/agent/vm/retire.py`** (line 79): The phrase "the reconciler's dm guard leaves it for the next pass" suggests a retry that never happens: after registry.forget, no code path calls remove_devmapper again for this namespace, and the reconciler's guard only refuses the unlink — it cannot remove the device. A teardown failure therefore strands the dm target, its loop device, and the pinned volume file permanently, not just "for the next pass". Either say so plainly ("stays behind for the operator") or have the reconciler's orphan pass attempt remove_devmapper for devices that outlive their VM.

**`src/aleph/vm/agent/vm/retire.py`** (line 85): _checked_namespace raises, yet it sits outside the per-volume try even though this helper is documented as best-effort. It is unreachable from the current callers (vm_hash is an ItemHash), but a raise here would abort the retire between registry.forget and _release_storage — consider logging-and-returning like remove_devmapper does, for symmetry.

**`src/aleph/vm/storage.py`** (line 553): If the mount point is genuinely still mounted (e.g. an earlier resize_and_tune_file_system died between mount and umount), dmsetup remove --retry will keep failing with EBUSY and the whole teardown becomes unretryable. A best-effort `umount` before the removal would make this path self-healing.

**`src/aleph/vm/storage.py`** (line 538): The log message does not distinguish which input failed validation (namespace vs volume name), which makes the two rejection causes harder to diagnose from logs. Minor: include which one in the message.

foxpatch-aleph (changes requested):
The teardown design is sound and mirrors create_devmapper correctly (ordering, the {ns}_base sibling rule, never touching the shared parent device), name validation is appropriately scoped to what dmsetup itself refuses, there is no injection surface (argv-only, no '/' in volume names), and test coverage is thorough. One real bug blocks: Path.is_mount() (storage.py:553) only exists on Python 3.12+, but requires-python is >=3.10 and packaging targets Ubuntu 22.04 (3.10) and Debian 12 (3.11); on those every remove_devmapper raises AttributeError, which teardown_vm_devices swallows, leaving the mount point and the {ns}_base device leaked on every retire. Switch to os.path.ismount() as reconciler.py:492 already does.

**`src/aleph/vm/storage.py`** (line 553): Path.is_mount() was only added in Python 3.12, but pyproject declares requires-python ">=3.10" and the packaging targets include ubuntu-22.04 (Python 3.10) and debian-12 (3.11). On those interpreters this line raises AttributeError on every call, which teardown_vm_devices swallows and logs — so the /mnt/{ns}_{vol} directory and, worse, the {ns}_base device (removed only after this step) are never cleaned up, reintroducing exactly the leak this PR fixes. Use os.path.ismount(mount_path) instead, as the reconciler already does (reconciler.py:492).

**`src/aleph/vm/agent/vm/retire.py`** (line 85): This _checked_namespace call is outside the per-volume try/except, so an implausible hash raises ValueError mid-retire — after registry.forget() and delete_records_for_vm() but before _release_storage and purge_vm_backups, leaving the VM half-retired. That is inconsistent with the best-effort design of the surrounding code (_release_storage is deliberately wrapped). Since remove_devmapper validates the namespace itself via device_name_for (logging and refusing), consider dropping this call or moving it inside the try.

**`tests/supervisor/test_devmapper_teardown.py`** (line 101): This test (and test_remove_devmapper_detaches_a_stale_loop_without_a_dm_device below) does not patch MOUNT_ROOT or /dev/mapper, so remove_devmapper probes the real /mnt and globs the real /dev/mapper. It passes with the current fixture hash, but is slightly environment-dependent — consider passing a tmp_path-based MOUNT_ROOT like test_remove_devmapper_tears_down_in_order does.

foxpatch-aleph (approved):
Solid, carefully defensive implementation. The teardown order exactly inverts the documented create_devmapper chain, every failure path degrades conservatively (teardown never aborts the retire, the purge guard keeps dm-pinned files on disk, and nothing retries teardown once the record is gone), and input validation makes both path traversal and glob-pattern injection impossible. Tests are thorough (exact command sequences and ordering, sibling rule, stale loops, awkward names, leftover mounts) and all directly relevant suites pass in this sandbox; the broader suite's failures are pre-existing environment issues (root-only /var paths) affecting untouched test files equally. The four findings are non-blocking nits and pre-existing gaps.

**`src/aleph/vm/storage.py`** (line 514): Nit: refusing volume names `.` and `..` is slightly over-broad versus the docstring's claim that only names `dmsetup create` rejects are refused. The composed device name `{ns}_.` is not `.`, so dmsetup would happily create it — meaning a device can exist that teardown refuses to remove (leaving the volume file pinned by its loop device). Ultra edge case and the failure direction is conservative, but the docstring and the check disagree.

**`src/aleph/vm/agent/views/operator.py`** (line 808): The in-place branch now tears devices down before purging, but this else-branch (ephemeral/confidential rebuilds) still calls `purge_vm_volumes` with the devices live: for a parent-backed volume the purge guard refuses to unlink, and `create_devmapper` then returns the stale device, so a reinstall of such a VM silently keeps the old volume contents despite the erase-everything semantics. Behavior is identical to base, so not a regression, but it is the same class of gap this PR fixes on the other branch — worth a follow-up.

**`docs/architecture/vm-lifecycle.md`** (line 426): This sentence says the purge guard applies 'until dm teardown on stop exists' — that teardown is this PR, so the sentence is now stale (and teardown happens on retire/reinstall, not on stop). `docs/architecture/storage.md` likewise documents only `create_devmapper` with no mention of `remove_devmapper`.

**`src/aleph/vm/agent/vm/retire.py`** (line 34): Minor: importing the private `_checked_namespace` from purge, plus storage.py defining its own identical `NAMESPACE_PATTERN`, leaves the same invariant spelled twice. Consider exporting one shared validator from `purge.py` (or storage) so the two cannot drift.

PR #1161, opened 2026-08-25
feat(agent): bounded download caches, in-stream size cap and LRU eviction (2.1, PR D)

odesenfans:
## Summary

PR D of the 2.1 disk reclamation work (spec Section 4, S4 unbounded caches and S5 no download size cap; the parent-device half of Section 5). Stacked on #1160.

- **Size enforced in the stream.** `download_file_in_chunks(url, tmp, *, max_bytes)` rejects an oversize `Content-Length` before opening the file and aborts, unlinking the `.part`, the moment the byte count passes the limit. Same limits as before (`MAX_PROGRAM_ARCHIVE_SIZE`, `MAX_DATA_ARCHIVE_SIZE`, new `MAX_RUNTIME_ARCHIVE_SIZE`), now applied to every `download_file` caller: the four cache getters, immutable volume refs and restore staging. `FileTooLargeError` is a 400 on the create path. A cache hit touches the entry's mtime (the LRU signal) and falls through to a download if the file vanished under it.
- **Bounded caches.** `CACHE_BUDGET` (default 20% of the filesystem) per cache root. The reconciler's cache pass computes the referenced set (every item hash a live VM's message names: runtime, code, data, rootfs parent, volume parents and refs, V-PROGRAM workload and hash tree, confidential firmware and runtime, the VM's own message; plus every `.reclaimable` marker's `depends_on`) through one shared enumeration in `reclaimable.py`, then evicts unreferenced entries LRU; if still over budget with only retained-VM-referenced entries left it reclaims those retained directories oldest-first and then their entries; entries a live VM references are never evicted (log and stop). Evicting a parent image removes its shared `/dev/mapper/<ref>` and read-only loop only when nothing stacks on it (sysfs holders), and a sweep recovers loops left over a deleted cache file.
- **Admission.** `storage.set_cache_admission` runs before a download writes: it may evict unreferenced LRU entries only, never retained VMs, never anything while the live set is not trusted (registry and `list_vms` disagree), and refuses with `InsufficientResourcesError` (503) otherwise.

## Known limitations
- If the supervisor never answers `list_vms`, admission never evicts and downloads are refused once a cache reaches its budget (fail-closed).
- Per-root percentage budgets stack when several caches share one filesystem (spec-mandated shape).
- Admission's directory walks run on the event loop (bounded: a few stats per entry).

## Test plan
- `tests/supervisor`: failing-ID set identical to base across both tasks.
- New: `test_storage_download_cap.py`, `test_cache_eviction.py`; extended `test_reconciler.py`, `test_reclaimable.py`, `test_storage.py`, run-routing error mapping.

### DIFF
--- a/docs/architecture/storage.md
+++ b/docs/architecture/storage.md
@@ -348,8 +348,10 @@ message. One pass does, in order:
    under `keep`. Liveness is asked again immediately before anything is
    marked, purged or evicted, so a create that commits while the pass walks
    is not mistaken for an orphan.
-2. Delete `*.part` files older than the guard, in the four download caches
-   and in the pools (the downloader streams volumes in place).
+2. Delete `*.part` and `*.tmp` files older than the guard, in the four
+   download caches and in the pools (the downloader streams volumes in
+   place). Every write-then-rename in the agent uses one of the two suffixes,
+   and an interrupted one leaks the same way.
 3. Sweep the side directories keyed by VM hash: the confidential session
    directory, the SNP and V-PROGRAM staging directories, and
    `/mnt/{namespace}_{volume}` mount points. A directory that is still a
@@ -417,9 +419,21 @@ default). The pass evicts least recently used first, mtime being a real
 signal because the downloader touches an entry on every cache hit
 (`_touch_cache_hit`). What it may not touch is the point:
 
-- Entries a **live VM** names, meaning every `runtime`, `code`, `data`,
-  rootfs parent, volume parent and immutable volume `ref` of a message in the
-  agent registry. If live references alone exceed the budget the pass logs an
+- Entries a **live VM** names. `reclaimable.iter_content_refs` is the single
+  enumeration of those, so the two questions asked of a message (what a
+  retained volume depends on, what a live VM still needs) cannot drift apart:
+  `runtime`, `code`, `data`, the rootfs parent, volume parents and immutable
+  volume `ref`s, a V-PROGRAM's `workload.ref` and `workload.hash_tree`, a TEE
+  instance's `trusted_execution.firmware` and `trusted_execution.runtime`, and
+  the VM's own message entry (`<vm_hash>.json`). The confidential ones are not
+  a detail: a V-PROGRAM's workload image and hash tree are attached as the
+  VM's disks straight out of `DATA_CACHE`, so evicting one unlinks a running
+  VM's disk. The runtime bundle tarball is the one ref no message carries
+  (only the manifest names it); it is protected whenever the manifest itself
+  is cached, and is otherwise a re-downloadable artifact, since nothing runs
+  off the tarball (it is verified and extracted into the VM's staging
+  directory at create time).
+  If live references alone exceed the budget the pass logs an
   error and stops: that is a capacity problem admission should have refused,
   and eviction cannot fix it. The referenced set is read from the registry's
   messages, so a pass whose live set holds a hash the registry has no record
@@ -429,6 +443,13 @@ signal because the downloader touches an entry on every cache hit
   retained directories go first (oldest marker first, through
   `purge_vm_storage`), and a parent is evicted only once every retained VM
   that named it is gone and its purge actually removed the disks.
+
+The pass also sweeps what an earlier teardown left behind: a loop device
+still backed by a cache entry that is already unlinked (the kernel shows it as
+` (deleted)` in `/sys/block/loop*/loop/backing_file`). That is the recovery
+path for a teardown that crashed or failed, because once the entry is gone
+nothing else can name the loop that pins its blocks; the same holders check
+applies, so a leaked loop whose device is in use after all is left alone.
 
 Evicting a parent image also removes what `create_devmapper` builds once per
 image rather than once per VM: the `/dev/mapper/<ref>` target and the
@@ -450,6 +471,18 @@ rather than writing bytes that trigger an eviction storm. Directories that
 are not cache roots are none of its business: the downloader streams per-VM
 volumes in place, and those are admitted by the capacity checks and the pool
 budget.
+
+It is deliberately weaker than the pass, because it runs on the event loop
+inside a create. It unlinks unreferenced cache entries and nothing else: it
+never reclaims a retained VM directory (that is an `rmtree`, which belongs in
+the reconciler's worker thread), so what it cannot free it refuses and leaves
+to the next pass. And it evicts nothing at all unless the live set can be
+trusted: a pass publishes the live set only when both halves agree (the
+registry and a supervisor that answered `list_vms`), and admission additionally
+requires that every hash in it has a registry record, since a live VM without a
+message means the referenced set is incomplete. Before the first pass there is
+no live set, so admission evicts nothing and simply admits or refuses on the
+budget as it stands.
 
 ### Settings
 
@@ -549,7 +582,9 @@ pass against an empty registry would call every running VM an orphan.
 - `src/aleph/vm/agent/vm/retire.py`: `retire_vm` and `RetireReason`, the one
   way a VM's storage is released.
 - `src/aleph/vm/agent/vm/reclaimable.py`: the `.reclaimable` marker
-  (`mark_reclaimable`, `adopt`, `iter_reclaimable`, `reclaimable_bytes`).
+  (`mark_reclaimable`, `adopt`, `iter_reclaimable`, `reclaimable_bytes`) and
+  the one enumeration of the cache entries a message names
+  (`iter_content_refs`, `refs_from_content`, `depends_on_from_content`).
 - `src/aleph/vm/agent/vm/cache.py`: the cache budget, LRU eviction
   (`evict_caches`), the download admission hook (`admit_download`) and the
   parent image's device teardown (`remove_parent_device`).
--- a/docs/architecture/storage.md
+++ b/docs/architecture/storage.md
@@ -342,6 +342,14 @@ registry alone is not enough, because it is refilled at boot from the agent
 DB and that rehydration skips any record with an empty or unparseable
 message. One pass does, in order:
 
+0. Tear down the device-mapper snapshots and loop devices of every namespace
+   `/dev/mapper` still names and no live VM owns
+   (`teardown_namespace_devices`). Before the walk, not inside it: a volume
+   file a dm target still holds cannot be unlinked usefully, so a directory
+   whose devices are still up is refused by step 1 and would be refused
+   again on every later pass. This runs on the event loop (dmsetup has to be
+   awaited) and only when the supervisor answered `list_vms`, since removing
+   the device of a VM that is merely unlisted takes that VM's disk with it.
 1. Walk `{pool}/*/`. Each directory is live (the registry or the supervisor
    knows its hash), reclaimable (a marker is present) or an orphan
    (neither). Orphans are purged under `reap` and marked `reason: orphan`
@@ -363,7 +371,11 @@ message. One pass does, in order:
    evict oldest-first (by `reclaimable_since`) until under
    `VOLUME_RETENTION_BUDGET`. Under `reap` everything marked is given back,
    which is how a node switched from `keep` to `reap` drains.
-5. Bring the four download caches under `CACHE_BUDGET` (see below).
+5. Bring the four download caches under `CACHE_BUDGET` (see below). Skipped
+   entirely, with an ERROR, when the supervisor could not be listed: what a
+   cache holds for a live VM is read from that VM's message, so a live set
+   that is the registry alone cannot answer the question, and evicting on it
+   would unlink a running VM's runtime.
 6. Sweep expired backups.
 
 Nothing a create is still building is touched. Two independent guards say
@@ -415,9 +427,19 @@ cache's `MAX_*_ARCHIVE_SIZE` before it opens the file, and aborts the moment
 the bytes written pass it, so a lying `Content-Length` cannot fill a disk.
 
 Per root: each cache gets `CACHE_BUDGET` (20% of the filesystem it sits on by
-default). The pass evicts least recently used first, mtime being a real
+default), the message cache included. The spec suggested a smaller budget for
+it; it does not have one, and does not need one: its entries are a few
+kilobytes of JSON each, so it never approaches a budget sized for runtime
+images, and one setting is one thing for an operator to reason about. A root's
+usage is its finished entries plus what it owes to downloads that have not
+finished (`in_flight_bytes`): the allocated blocks of `.part` and `.tmp`
+files, and the size any admitted download was promised.
+
+The pass evicts least recently used first, mtime being a real
 signal because the downloader touches an entry on every cache hit
-(`_touch_cache_hit`). What it may not touch is the point:
+(`_touch_cache_hit`, called by `download_file` for the caches it serves and
+by `get_runtime_path` / `get_rootfs_base_path`, which short-circuit a hit
+before reaching it). What it may not touch is the point:
 
 - Entries a **live VM** names. `reclaimable.iter_content_refs` is the single
   enumeration of those, so the two questions asked of a message (what a
@@ -464,10 +486,17 @@ left alone: its devices belong to that create now.
 
 Admission is the same question one download ahead: `storage`'s cache
 admission hook (registered by the agent on `set_cache_admission`) runs inside
-`download_file_in_chunks` once `Content-Length` is known and before the file
-is opened. It evicts what it can, and raises `InsufficientResourcesError`
+`download_file_in_chunks` once the size is known and before the file is
+opened. It evicts what it can, and raises `InsufficientResourcesError`
 (mapped to 503 on the create path) when the download still would not fit,
-rather than writing bytes that trigger an eviction storm. Directories that
+rather than writing bytes that trigger an eviction storm. A response with no
+`Content-Length` is admitted for the worst case it is allowed to send, its
+`MAX_*_ARCHIVE_SIZE` cap, and refused when that does not fit: the alternative
+is a download charged for nothing at all. What is admitted is then charged to
+the download's `.part` path (`reserve_download`) until `download_file`
+releases it, so a second create arriving while the first is still writing
+sees the room the first was promised rather than only the bytes it has
+managed to write. Directories that
 are not cache roots are none of its business: the downloader streams per-VM
 volumes in place, and those are admitted by the capacity checks and the pool
 budget.
--- a/docs/architecture/storage.md
+++ b/docs/architecture/storage.md
@@ -361,7 +361,8 @@ message. One pass does, in order:
    evict oldest-first (by `reclaimable_since`) until under
    `VOLUME_RETENTION_BUDGET`. Under `reap` everything marked is given back,
    which is how a node switched from `keep` to `reap` drains.
-5. Sweep expired backups.
+5. Bring the four download caches under `CACHE_BUDGET` (see below).
+6. Sweep expired backups.
 
 Nothing a create is still building is touched. Two independent guards say
 so: a hash registered in the in-process "creating" set (every create path in
@@ -399,6 +400,56 @@ the whole transfer and records the imported VM in the registry (and the agent
 DB) once `CreateVm` returns. A namespace under the guard is skipped whole,
 `.part` files included: the directory mtime does not advance while a file
 inside it grows, so age is no evidence that a transfer has stopped.
+
+### The download caches
+
+The runtime, code, data and message caches are flat directories keyed by item
+hash and shared by every VM naming the same hash, and nothing used to remove
+anything from them. Two things now bound them
+(`src/aleph/vm/agent/vm/cache.py`).
+
+In the stream: `download_file_in_chunks` refuses a `Content-Length` above the
+cache's `MAX_*_ARCHIVE_SIZE` before it opens the file, and aborts the moment
+the bytes written pass it, so a lying `Content-Length` cannot fill a disk.
+
+Per root: each cache gets `CACHE_BUDGET` (20% of the filesystem it sits on by
+default). The pass evicts least recently used first, mtime being a real
+signal because the downloader touches an entry on every cache hit
+(`_touch_cache_hit`). What it may not touch is the point:
+
+- Entries a **live VM** names, meaning every `runtime`, `code`, `data`,
+  rootfs parent, volume parent and immutable volume `ref` of a message in the
+  agent registry. If live references alone exceed the budget the pass logs an
+  error and stops: that is a capacity problem admission should have refused,
+  and eviction cannot fix it. The referenced set is read from the registry's
+  messages, so a pass whose live set holds a hash the registry has no record
+  for skips the cache pass entirely rather than guess.
+- Entries a `.reclaimable` marker names in `depends_on`, the parent images
+  retained overlays are built on. Over budget with only those left, the
+  retained directories go first (oldest marker first, through
+  `purge_vm_storage`), and a parent is evicted only once every retained VM
+  that named it is gone and its purge actually removed the disks.
+
+Evicting a parent image also removes what `create_devmapper` builds once per
+image rather than once per VM: the `/dev/mapper/<ref>` target and the
+read-only loop device under it. A parent whose device still has holders (a
+VM's `<ns>_base` device is stacked on it) is not evicted at all, and an
+unanswerable question counts as held. The teardown runs on the event loop
+after the pass (`reconcile_now`), because the entry is already unlinked by
+then; `losetup -j` can no longer find the loop that pins its blocks, so the
+loop devices are found through sysfs, which keeps the backing path with a
+` (deleted)` suffix. An image a create downloaded again in the meantime is
+left alone: its devices belong to that create now.
+
+Admission is the same question one download ahead: `storage`'s cache
+admission hook (registered by the agent on `set_cache_admission`) runs inside
+`download_file_in_chunks` once `Content-Length` is known and before the file
+is opened. It evicts what it can, and raises `InsufficientResourcesError`
+(mapped to 503 on the create path) when the download still would not fit,
+rather than writing bytes that trigger an eviction storm. Directories that
+are not cache roots are none of its business: the downloader streams per-VM
+volumes in place, and those are admitted by the capacity checks and the pool
+budget.
 
 ### Settings
 
@@ -452,6 +503,10 @@ pass against an empty registry would call every running VM an orphan.
   and a placement that does not fit evicts them (oldest first) before it is
   refused (`src/aleph/vm/agent/capacity.py`,
   `src/aleph/vm/agent/resources.py`, `src/aleph/vm/storage_pools.py`).
+- A cache entry a live VM's message names is never evicted, and a parent
+  image whose device-mapper device still has holders is never evicted; both
+  questions fail closed when they cannot be answered
+  (`src/aleph/vm/agent/vm/cache.py`).
 - Every agent-side deletion goes through `retire_vm` with an explicit
   reason; nothing else under `src/aleph/vm/agent/` calls
   `supervisor.delete_vm` (`src/aleph/vm/agent/vm/retire.py`).
@@ -495,6 +550,9 @@ pass against an empty registry would call every running VM an orphan.
   way a VM's storage is released.
 - `src/aleph/vm/agent/vm/reclaimable.py`: the `.reclaimable` marker
   (`mark_reclaimable`, `adopt`, `iter_reclaimable`, `reclaimable_bytes`).
+- `src/aleph/vm/agent/vm/cache.py`: the cache budget, LRU eviction
+  (`evict_caches`), the download admission hook (`admit_download`) and the
+  parent image's device teardown (`remove_parent_device`).
 - `src/aleph/vm/agent/vm/reconciler.py`: `reconcile_storage`, the create
   guard (`creating`), the evictor (`make_room`) and the startup/periodic
   hooks the agent registers in `src/aleph/vm/agent/supervisor.py`.
--- a/docs/architecture/storage.md
+++ b/docs/architecture/storage.md
@@ -490,9 +490,16 @@ admission hook (registered by the agent on `set_cache_admission`) runs inside
 opened. It evicts what it can, and raises `InsufficientResourcesError`
 (mapped to 503 on the create path) when the download still would not fit,
 rather than writing bytes that trigger an eviction storm. A response with no
-`Content-Length` is admitted for the worst case it is allowed to send, its
-`MAX_*_ARCHIVE_SIZE` cap, and refused when that does not fit: the alternative
-is a download charged for nothing at all. What is admitted is then charged to
+`Content-Length` is a different question, and the hook is told the two
+figures separately (`content_length`, `max_bytes`) so it can tell them apart.
+A cap is not a measurement: `MAX_RUNTIME_ARCHIVE_SIZE` is 100 GiB, the
+ceiling for a runtime image and equally the ceiling for a few kilobytes of
+manifest, so charging it as a size would evict a whole cache root for a
+download that never needed the room. An unknown-length download therefore
+never evicts, is refused only when the root is already over its budget, and
+is charged `min(cap, budget)`: bounded, so a stream of them still runs the
+root over budget and the next one is refused, and never more than the budget
+itself. What is admitted is then charged to
 the download's `.part` path (`reserve_download`) until `download_file`
 releases it, so a second create arriving while the first is still writing
 sees the room the first was promised rather than only the bytes it has
--- a/docs/architecture/vm-lifecycle.md
+++ b/docs/architecture/vm-lifecycle.md
@@ -136,7 +136,9 @@ Admission happens in two layers that never overlap in what they check:
   own early registry record (recorded before the spec build, to make
   owner-auth answerable during a slow confidential download) is excluded
   from the committed memory/vCPU sums via `exclude_vm_hash`, or a create
-  would count its own request against itself. GPU admission is a separate
+  would count its own request against itself. Cache downloads are admitted
+  separately, against `CACHE_BUDGET`, inside the downloader itself (see
+  [storage.md](storage.md)). GPU admission is a separate
   reservation ledger (`CapacityManager.holds`, keyed by concrete `pci_host`)
   with a short-lived hold/resolve two-step: `reserve_gpus` (used by the same
   dry-run endpoint) holds a card for a user for `RESERVATION_TTL_SECONDS`,
@@ -423,8 +425,13 @@ path); `purge_vm_storage` removes those directories, the confidential
 session directory and the SNP/V-PROGRAM staging directory. A `.btrfs`
 volume whose device-mapper target is still present is refused with an
 error rather than unlinked (the loop device would pin the inode and
-`create_devmapper` would skip the rebuild), until dm teardown on stop
-exists.
+`create_devmapper` would skip the rebuild). That refusal is a backstop
+rather than a dead end: `retire_vm` tears the VM's dm snapshots and loop
+devices down before the storage pass (`teardown_vm_devices`, from the
+message's volumes, or `teardown_namespace_devices` off `/dev/mapper` when
+no record is left), and the reconciler does the same for every namespace
+no live VM owns before its walk, so a refused directory is reclaimed on
+the pass after its devices go.
 
 What `DeleteVm` does: stop the VM (tearing down its network state exactly
 as `StopVm` would, and releasing every handle the daemon holds on the VM's
--- a/src/aleph/vm/agent/run.py
+++ b/src/aleph/vm/agent/run.py
@@ -706,7 +706,8 @@ async def create_vm_execution_or_raise_http_error(
             text="This CRN cannot host the requested workload at this time.",
         ) from error
     except FileTooLargeError as error:
-        raise HTTPInternalServerError(reason=error.args[0]) from error
+        # An oversized resource is the message's fault, not the node's.
+        raise HTTPBadRequest(reason=error.args[0]) from error
     except VmStartupError as error:
         # Created but never reached RUNNING (terminal status or start timeout):
         # a distinct, expected outcome, not the generic "unhandled error".
@@ -756,7 +757,8 @@ def _raise_http_for_program_error(error: Exception, vm_hash: ItemHash) -> None:
             text="This CRN cannot host the requested workload at this time.",
         ) from error
     if isinstance(error, FileTooLargeError):
-        raise HTTPInternalServerError(reason=str(error) or "File too large") from error
+        # An oversized resource is the message's fault, not the node's.
+        raise HTTPBadRequest(reason=str(error) or "File too large") from error
     if isinstance(error, VmStartupError):
         # Created but never reached RUNNING (terminal status or start timeout):
         # a distinct, expected outcome, not the generic "unhandled error".
--- a/src/aleph/vm/agent/supervisor.py
+++ b/src/aleph/vm/agent/supervisor.py
@@ -522,7 +522,7 @@ def run():
     # And the third hook of the same kind: a download about to start asks the
     # cache budget for room, so a create that would blow it is refused before
     # the bytes land rather than after (spec section 4).
-    storage.set_cache_admission(lambda root, size: admit_download(app["vm_registry"], root, size))
+    storage.set_cache_admission(lambda tmp_path, size: admit_download(app["vm_registry"], tmp_path, size))
     app.on_startup.append(start_node_hash_discovery)
     app.on_cleanup.append(stop_node_hash_discovery)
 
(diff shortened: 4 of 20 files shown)


foxpatch-aleph (approved):
Well-architected PR with thorough documentation, comprehensive test coverage, and a consistent fail-closed design philosophy. The cache eviction logic, download size caps, admission control, and device teardown are all correctly implemented. Security concerns (path traversal, option injection, device-mapper safety) are properly handled. The accepted race conditions are narrow and well-documented. Minor observations below are non-blocking nits.

**`src/aleph/vm/backup/staging.py`** (line 37): Using `MAX_RUNTIME_ARCHIVE_SIZE` as the cap for a volume download is semantically incorrect — volumes are not runtimes. This works in practice because the default is 100 GiB (generous enough for volumes), but a dedicated `MAX_VOLUME_ARCHIVE_SIZE` or at least a comment explaining the reuse would be clearer.

**`src/aleph/vm/agent/vm/cache.py`** (line 84): The `_live_snapshot` module-level global is shared between the reconciler thread (writer via `record_live_snapshot`) and the admission hook on the event loop (reader via `live_snapshot`). This is safe under the GIL since the value is either `None` or an immutable `frozenset`, but a brief comment noting the threading model would help future readers.

**`src/aleph/vm/agent/vm/cache.py`** (line 326): In `remove_parent_device`, there is a TOCTOU race between `cache_path.exists()` (line 326) and `dmsetup remove` (line 334): a create could re-download the image and set up devices in that window. The docstring acknowledges this, and the window is narrow, but it could break a VM being built. Consider re-checking `cache_path.exists()` immediately before the `dmsetup` call, or using a lock shared with the create path.

**`src/aleph/vm/agent/vm/cache.py`** (line 648): `admit_download` calls `cache_roots()` which calls `Path.is_dir()` on every cache root. This is repeated on every download admission. Since cache roots don't change at runtime, caching the result (or invalidating on settings change) would avoid redundant stats on the event loop path. Minor performance nit.

**`src/aleph/vm/agent/vm/cache.py`** (line 660): After `evict_caches` runs in `admit_download`, `cache_entries(root)` is called again to recompute usage. This is a second directory walk on the event loop. The eviction result from `evict_caches` could be used to compute the post-eviction usage without a re-walk, though the current approach is simpler and correct.

foxpatch-aleph (approved):
Well-designed and thoroughly tested implementation of bounded download caches with LRU eviction, in-stream size caps, and download admission. The fail-closed semantics are consistently applied across the cache pass, admission hook, and device teardown. Input validation is thorough (_safe_ref, _plausible, _checked_namespace). Race conditions (create guard, TOCTOU on cache hits, concurrent download reservations) are handled correctly. FileTooLargeError correctly maps to 400 (client fault) and is not retried. Test coverage is comprehensive with 40+ tests across two new test files plus extensions to existing ones. The few minor observations (double directory walk in admit_download, cross-module mutable state, manifest JSON parsing) are non-blocking and consistent with the existing codebase patterns.

**`src/aleph/vm/agent/vm/cache.py`** (line 724): admit_download calls cache_entries(root) a second time after evict_caches already walked the directory. This doubles the I/O on the event loop for large caches. Consider having evict_caches return the post-eviction entry list or usage, or pass the already-computed entries forward. Non-blocking optimization.

**`src/aleph/vm/agent/vm/cache.py`** (line 216): _manifest_bundle_ref parses JSON from a cache entry by name. A cache entry whose name matches a manifest ref but whose content is arbitrary JSON could inject an extra protected ref (bundle.ref). The 64KB cap and _safe_ref check mitigate this, and the worst case is over-retention not data loss, but worth noting that the manifest content is trusted as-is.

**`src/aleph/vm/storage.py`** (line 100): _reserved_downloads is a plain dict accessed from both the event loop (reserve_download/release_download in download_file) and worker threads (reserved_downloads in in_flight_bytes, called via evict_caches in asyncio.to_thread). This is safe under the GIL (dict copy is atomic), consistent with the registry snapshot pattern, but fragile if the runtime ever changes.

foxpatch-aleph (approved):
This PR is a thorough, well-documented implementation of bounded download caches with LRU eviction, in-stream size caps, and a download admission hook. The design is consistently fail-closed: live VM references are never evicted, the cache pass is skipped when the live set is untrusted, and admission refuses rather than writes-then-evicts. The code correctly handles TOCTOU races (cache hit eviction mid-touch), double-counting of in-flight downloads, and command injection prevention via _safe_ref. The test suite is comprehensive, covering LRU eviction, live reference protection, retained-VM reclaim ordering, admission concurrency, unknown-length downloads, device teardown, and all the fail-closed paths. The FileTooLargeError→400 mapping is a correct fix. The only notable design trade-off is synchronous filesystem I/O on the event loop in the admission path, which the PR explicitly acknowledges as a bounded cost.

**`src/aleph/vm/agent/vm/cache.py`** (line 738): After `evict_caches` runs on line 736, `cache_entries(root)` is called again on line 738 to recompute usage. This is a second full directory walk on the event loop in the admission path. In principle the evicted entries could be subtracted from the pre-eviction usage instead, avoiding the re-walk. Not a bug — just a minor efficiency note for a path that already acknowledges its I/O cost.

**`src/aleph/vm/agent/vm/cache.py`** (line 243): `live_refs` iterates `list(registry.items())` which is GIL-safe per the existing `live_hashes` documentation. When called from the reconciler's `evict_caches` (via `asyncio.to_thread`), the registry could be mutated on the event loop concurrently. The `list()` call makes this safe, but a brief comment noting the thread-safety rationale (matching the one on `live_hashes`) would be welcome for consistency.

**`src/aleph/vm/agent/vm/cache.py`** (line 776): When `max_bytes` is `None`, the unknown-length download is charged the full `budget`. In practice this never happens (all callers pass `max_bytes`), but charging the entire budget for a single unknown-length download means a second concurrent one is always refused. If `max_bytes=None` ever becomes a real path, consider charging a smaller fraction.

foxpatch-aleph (changes requested):
Well-engineered and thoroughly tested, with consistent fail-closed discipline and good security hygiene around subprocess names. However, there is one genuine bug: iter_content_refs misses VerifiedVolume.hash_tree, so a V-PROGRAM's volume hash trees — downloaded into DATA_CACHE and attached as the running VM's disks (vprogram_launch.py:260-261) — are not in the never-evict set, contradicting the module's core invariant. The existing V-PROGRAM test uses empty volumes, so the gap is untested. Minor nits: the unknown-length admission charge of min(cap, budget) serializes all downloads on a root when cap >= budget; dry-run can over-report reclaimed parents; and the 'only live references' ERROR message is inaccurate when marker-pinned entries remain.

**`src/aleph/vm/agent/vm/reclaimable.py`** (line 153): This enumeration misses `VerifiedVolume.hash_tree`. A V-PROGRAM's verified volumes are served by `get_existing_file` twice each — for `volume.ref` and `volume.hash_tree` (vprogram_launch.py:260-261) — and both files are attached directly as the running VM's disks out of DATA_CACHE. Only `volume.ref` is yielded here, so `volume.hash_tree` is not in the referenced set and a cache pass can evict a live V-PROGRAM's hash-tree disk, exactly the failure mode this module documents it prevents. The workload's `hash_tree` is enumerated ("workload_hash_tree"), so this looks like an oversight; the test (`test_refs_from_content_covers_a_vprogram_workload`) sets `content.volumes = []`, which is why it goes unnoticed. Please yield `volume.hash_tree` alongside `volume.ref` and extend the test with a non-empty volumes list.

**`src/aleph/vm/agent/vm/cache.py`** (line 776): Non-blocking observation: when `max_bytes >= budget` (the common runtime-cap case: 100 GiB cap vs a 20% budget), this charges the entire budget, so the first unknown-length download refuses every subsequent download to that root — known-length ones included, since eviction cannot remove the reservation — until it finishes. The docstring's "a handful of concurrent unknown-length downloads" rationale only holds when cap < budget. Fail-closed and safe, but worth a sentence in the docstring (or a smaller charge, e.g. a fraction of the budget) if concurrent chunked downloads are expected.

**`src/aleph/vm/agent/vm/cache.py`** (line 619): In a dry run this branch skips `purge_vm_storage` entirely, so a directory whose purge would refuse (a dm target still holds it) still counts as reclaimed and its parent is reported evicted in the next line via `_evict_freed_parents`. The dry-run report can therefore promise more than a real pass would free. Consider checking the same refusal condition here, or noting the over-approximation.

**`src/aleph/vm/agent/vm/cache.py`** (line 532): This ERROR ("only live references remain; admission should have refused this load") also fires when what remains is marker-pinned entries whose directories were protected by `is_live`/`is_creating` or by a refused purge — not live references. The message points the operator at admission for a condition admission does not control. Consider wording it as "only pinned references remain".

foxpatch-aleph (approved):
## Review: PR #1161 — bounded download caches, in-stream size cap and LRU eviction

I reviewed the full diff against the repository state, read the new module and every touched file in context, and ran the PR-related test files (`test_storage_download_cap.py`, `test_cache_eviction.py`, `test_devmapper_teardown.py`, `test_reclaimable.py`, `test_reconciler.py`, `test_storage.py`, `test_supervisor_run_routing.py`, `test_retire.py`, `test_pr2_neutral_modules.py`) — **242 tests pass**. The 40 failures/errors in the full `tests/supervisor` suite are environment permission issues (`PermissionError: /var/cache/aleph`) and unrelated dns/auth/vprogram-launch tests, not PR logic.

**Correctness checks that passed:**
- The in-stream cap (`storage.py:141-164`) refuses a lying `Content-Length` before the file is opened and aborts mid-stream past the limit; `FileTooLargeError` is not in the retry exception list, so it propagates immediately (verified by `test_download_file_unlinks_the_part_and_does_not_retry`).
- Admission accounting: `in_flight_bytes` counts `max(reservation, .part bytes)` so an in-flight download is never double-counted, and `admit_download` charges `content_length` to the `.part` until `download_file` releases it in the `finally` (storage.py:252-254) — the concurrent-creates scenario is properly tested.
- Fail-closed rules are consistent between the pass (`reconciler._enforce_cache_budget`) and admission (`_may_evict_for_admission`: no snapshot → no eviction; unknown hashes → no eviction), and the pass skips entirely rather than guessing.
- The LRU signal is real: `get_runtime_path`/`get_rootfs_base_path` touch hits that never reach `download_file`, and vanished entries fall through to re-download.
- dm/loop teardown layering is sound: `_safe_ref` blocks dmsetup argument injection (no `/`, no `-` prefix, ≤127 bytes) and `run_in_subprocess` uses `create_subprocess_exec`; `orphan_device_namespaces` only reads plausible item-hash devices; holder checks fail closed.
- All six `download_file` callers now pass a cap, matching the PR claim.

```json
{
  "verdict": "APPROVE",
  "summary": "Bounded download caches, in-stream size caps and LRU eviction are implemented carefully and consistently with the stacked PRs. The fail-closed rules (never evict live-referenced entries, skip the cache pass when the live set cannot be trusted) are applied symmetrically to the pass and to admission, the in-flight reservation ledger prevents double admission, and the dm/loop teardown layering avoids unlinking files pinned by loop devices. All PR-related tests pass (242 tests); the few full-suite failures in this environment are unrelated permission issues. Three minor observations below: a narrow race in the leaked-loop sweep missing one guard the main teardown has, a duplicated admission test, and a small asymmetry in the inaccessible-cache-root branch.",
  "comments": [
    {
      "path": "src/aleph/vm/agent/vm/cache.py",
      "line": 820,
      "body": "The sweep's `dmsetup remove` here checks holders but not whether the cache entry was re-downloaded: `remove_parent_device` guards with `cache_path.exists()` (\"the image is back in the cache\"), but this path has no such guard. If a create re-downloads the image and rebuilds its `/dev/mapper/<ref>` device during the window where no holder is yet stacked, `parent_device_is_free` returns True and the sweep removes that device, breaking the create. Narrow window, and consistent with the documented residual races, but the same `cache_path.exists()` check would close it."
    },
    {
      "path": "tests/supervisor/test_cache_eviction.py",
      "line": 680,
      "body": `test_a_measured_download_still_evicts_to_make_room` duplicates `test_admit_download_evicts_to_make_room` (line 309) with an identical body (budget 8192, code root, `admit_download(registry, pools["code"] / "new.part", 8000)`). Drop one or fold the Content-Length path into a shared helper."
    },
    {
      "path": "src/aleph/vm/agent/vm/cache.py",
      "line": 728,
      "body": "In the OSError budget branch a known-length download reserves its size, but an unknown-length one (`content_length is None`) is admitted without any reservation at all. Minor: the budget is uncomputable there and the written `.part` bytes still show up through `in_flight_bytes`, but an unknown-length stream on an inaccessible root is invisible to admission until bytes land."
    }
  ]
}
```

foxpatch-aleph (approved):
Well-built bounded-cache work: the in-stream size cap, per-root CACHE_BUDGET eviction, and admission hook all verify against the registry's message enumeration, fail closed on unanswerable questions, and are backed by thorough tests (242 passing across the touched files; remaining suite failures are environmental and in files untouched by this PR). Minor dry-run accuracy and teardown-lock observations below.

**`src/aleph/vm/agent/vm/reconciler.py`** (line 727): `_release_cache_devices` runs outside the pass lock, so device teardowns of overlapping passes can overlap. Each operation is guarded (cache-entry exists check, holders check) and `sweep_leaked_cache_loops` is the recovery path, so this is safe, but worth keeping in mind if a future teardown grows stateful.

**`src/aleph/vm/agent/vm/cache.py`** (line 621): In dry-run, the `iter_namespace_dirs` leftover check is skipped along with the purge, so a dry run cannot detect the "purge left directories behind" case and may report parent images as freed that a real pass would keep. Consistent with dry-run reporting semantics but slightly over-promising.

**`src/aleph/vm/agent/vm/cache.py`** (line 889): Module-level `_live_snapshot` is deliberate and documented, and the tests reset it via monkeypatch — good. The residual create-before-record race is documented as accepted with the mtime touch as mitigation; worth keeping the docstring accurate if admission logic changes.

foxpatch-aleph (approved):
Solid, well-documented implementation of bounded download caches with LRU eviction, in-stream size caps, and admission control. The fail-closed rules are applied consistently across the pass, the orphan-device teardown and the admission hook; the single shared ref enumeration in reclaimable.py prevents the live-set/depends-on drift the PR warns about; reservation accounting (max of reserved vs written bytes, not their sum) is correct; and dmsetup/losetup targets are validated against option injection and path traversal. All 243 tests in the touched suites pass; the remaining suite failures are pre-existing environment issues in untouched files. Four minor observations: remove_parent_device could re-check holders before dmsetup remove (as sweep_leaked_cache_loops already does), the live snapshot going stale after a retire degrades admission until the next pass (fail-closed but undocumented), the .part/.tmp suffix knowledge is duplicated between cache.py and reconciler.py, and one ERROR message can misattribute what remains over budget. None are blocking.

foxpatch-aleph (approved):
Well-engineered PR that bounds the four download caches with a per-root CACHE_BUDGET, LRU eviction keyed on mtime touches, in-stream size caps and download admission, plus the shared-parent-device teardown. The fail-closed discipline is consistent across the cache pass, the admission hook and the orphan-device teardown (untrusted live set or unanswered supervisors mean no eviction, no cache pass, no device removal), admission is atomic under the event loop so the reservation accounting closes the concurrent-download race, and iter_content_refs as a single enumeration of message refs makes the depends_on and never-evict questions structurally unable to drift. Tests are thorough and genuinely exercise the tricky interleavings (evicted-mid-touch, unknown-length charging, purge refusals, device holders); all touched test files pass locally and I found no correctness or security bugs. The three comments below are non-blocking robustness/performance notes.

**`src/aleph/vm/agent/vm/reconciler.py`** (line 760): The teardown key is derived purely from the device name: any `/dev/mapper` entry whose pre-underscore prefix matches the loose `^[0-9a-zA-Z]{16,128}$` pattern and that no live VM or in-flight create owns gets `dmsetup remove`d. That covers every agent-built VM device, but it would also match a foreign crypt/LVM-style device named like `<16+ alphanumeric>_x`. Open/mounted devices refuse removal and this only runs when the supervisor answered, which limits the blast radius — but since VM namespaces are full 64-char item hashes in practice, tightening the prefix check (or cross-checking that the namespace exists in pool directories) would make the pass harder to surprise.

**`src/aleph/vm/agent/vm/cache.py`** (line 504): `live_refs(registry)` is computed before we know whether any root is over budget, and it is the expensive part: per live VM it traverses the message and, for every runtime/tee_runtime ref, re-reads and JSON-parses the cached manifest (`_manifest_bundle_ref`). Every admitted download pays that even when the caches are nowhere near budget. Computing it lazily — after the `state.over_budget` check, next to the per-root `_markers()` — would keep the common admission path at plain `iterdir`/`lstat`.

**`src/aleph/vm/agent/vm/cache.py`** (line 377): Small residual race worth recording: between `cache_path.exists()` returning False and `_loop_devices_backing(cache_path`, a create could re-download the image and `create_loopback_device` a loop on the new file, which this teardown would then detach. The window is milliseconds and the create must finish a multi-GB download inside it, so it is negligible — but it is a different race from the one the module header documents, and a comment noting it (or a holders-style recheck before `losetup -d`) would keep the two conscious.

foxpatch-aleph (approved):
A carefully engineered PR: the bounded caches fail closed on every dangerous question (untrusted live sets, held parent devices, unknown registry records), the single content-ref enumeration keeps the depends_on and never-evict questions from drifting apart, and the unknown-length admission logic correctly distinguishes a cap from a measurement. Tests are behavioral and cover the fail-closed branches, races, and dry-run semantics; all 247 tests in the affected suites pass. I found no correctness or security bugs. The only issues worth noting are one narrow TOCTOU in remove_parent_device (a loop device matched by path string can hit a concurrently re-downloaded image), a missing re-check in sweep_leaked_cache_loops, and the acknowledged event-loop cost of admission-time eviction. None block the merge.

**`src/aleph/vm/agent/vm/cache.py`** (line 389): TOCTOU: `_loop_devices_backing` matches by path string, so both the leaked loop (backing "P (deleted)") and a concurrently re-downloaded image's fresh loop (backing "P") match `backing.removesuffix(" (deleted)") == str(path)`. The `cache_path.exists()` guard at line 377 closes the common case, but a create that completes its download in the window between that check and this sysfs enumeration gets its new loop device detached. Restricting the match to backings that end with `DELETED_SUFFIX` — as `_deleted_cache_backings` already does — would close the window and matches the function's stated intent (sysfs is read precisely because the leaked loop's file is unlinked).

**`src/aleph/vm/agent/vm/cache.py`** (line 820): Symmetry nit: unlike `remove_parent_device`, the sweep has no "the image is back in the cache" re-check before `dmsetup remove --retry <ref>`. A create that re-downloaded the image and passed `create_devmapper`'s `is_block_device()` early-return but hasn't yet stacked `<ns>_base` can have the shared device removed under it (the holders check in `parent_device_is_free` only protects once the base exists). The window is tiny; mirroring the exists() check would make the two teardown paths consistent.

**`src/aleph/vm/agent/vm/cache.py`** (line 734): Confirmed as documented in the PR's known limitations: `evict_caches` from here walks every cache root (one lstat per entry), re-reads all reclaimable markers, and reads one manifest JSON per registry record — on the event loop, per download. Correct, but on a node with a large runtime cache this is O(entries) per download; if it ever shows up in profiles, caching the referenced set between passes is the lever.

foxpatch-aleph (changes requested):
Well-engineered, thoroughly tested and documented PR; budget accounting, fail-closed liveness checks and device teardown all check out, and all 259 tests in the touched suites pass. One genuine gap: iter_content_refs misses VerifiedVolume.hash_tree, so a live V-PROGRAM's extra verified volumes have their hash-tree entries (fetched from DATA_CACHE and attached as disks in vprogram_launch.py:260-263) treated as unreferenced and evictable, contradicting the module's own no-live-ref invariant and pinning the blocks of an unlinked file until the VM stops. Fix is two lines in _iter_volume_refs; a real-model test would prevent this class of silent MagicMock-only regression.

**`src/aleph/vm/agent/vm/reclaimable.py`** (line 201): Verified volumes are missing their hash tree. A V-PROGRAM's extra verified volumes are fetched as a pair — get_existing_file(str(volume.ref)) and get_existing_file(str(volume.hash_tree)) in vprogram_launch.py:260-261 — and both are attached as disks (DiskSpec at lines 262-263). _iter_volume_refs yields only the volume's ref (kind "volume"), so volume.hash_tree never enters the referenced set and the cache pass can evict a running VM's attached hash-tree disk. The workload pair is covered in _iter_measured_refs but the volume pairs are not; yield a "volume_hash_tree" kind here (and consider folding the workload/volume hash trees into one rule, since both are attached verbatim from DATA_CACHE).

**`tests/supervisor/test_reclaimable.py`** (line 324): Non-blocking: because refs_from_content is only tested against MagicMock, a wrong attribute name on a real model would be silent in tests and mean evicting a live VM's disk in production (this is how the VerifiedVolume.hash_tree gap above survives). A test that parses a real VerifiableProgramMessage / InstanceMessage (with volumes carrying ref and hash_tree) and asserts the enumerated ref set would pin the model field names the mocks cannot check.

**`src/aleph/vm/agent/vm/cache.py`** (line 736): Non-blocking: on the admission path evict_caches runs live_refs() (one small file read per manifest-kind ref) and _markers() (a walk of every namespace directory on every pool, re-done per over-budget root) synchronously on the event loop, for every cache download. reclaimable_bytes got a TTL cache for exactly this walk; the admission path re-does it uncached. The PR documents the walk cost as accepted, so just flagging it as the place to look if download latency ever matters on VM-dense nodes.

foxpatch-aleph (approved):
Well-engineered PR with no blocking issues. The in-stream size cap is airtight (Content-Length checked up front, mid-stream abort, no retry of FileTooLargeError, .part always unlinked), admission accounting cannot double-count, and both the cache pass and admission hook fail closed when the live set is untrusted. Path-injection defenses are thorough (anchored NAMESPACE_PATTERN, _safe_ref rejecting traversal/oversized names, sysfs holders read fail-closed). iter_content_refs as a single enumeration prevents the live-refs and depends-on sets from drifting apart. 200 tests pass (47 in test_cache_eviction.py plus 153 across test_storage_download_cap, test_reconciler, test_reclaimable, test_devmapper_teardown, test_retire, and test_storage); the residual races (create-vs-eviction, admission-vs-pass, orphan-device TOCTOU) are documented and mitigated via LRU touch, re-download checks, and fail-closed guards. Docstrings explain the reasoning behind every fail-closed decision. Two non-blocking observations below.

**`src/aleph/vm/agent/vm/reconciler.py`** (line 802): Non-blocking: is_creating() is evaluated at listing time, so a create that enters creating() between this glob and _teardown_orphan_devices' removal could have its devices torn down. The window is milliseconds and the affected create would fail loudly at dmsetup rather than silently, but a comment (or a re-check of is_creating before each removal) would make the invariant explicit.

**`src/aleph/vm/agent/vm/cache.py`** (line 751): Non-blocking: _admit_unknown_length charges min(max_bytes, budget) but never evicts, so consecutive unknown-length downloads can each consume up to the cap while the root is under budget, only being refused once the root crosses it. This is documented in the docstring and is a reasonable trade-off given Content-Length-less responses are rare in this pipeline; noting it for operational awareness.

foxpatch-aleph (approved):
Solid, well-guarded implementation. Admission runs before bytes land, is charged to the .part path so concurrent downloads see each other's promised room, and fails closed in exactly the same ways the reconciler pass does (no snapshot, live VMs without registry records). The retry loop correctly does not re-attempt FileTooLargeError or InsufficientResourcesError, so a refusal cannot become an admission storm. .part/.tmp exclusion prevents eviction from unlinking in-flight downloads, in_flight_bytes avoids double-counting a growing .part, and the leaked-loop sweep only detaches inodes the kernel itself reports as deleted, with a holders fail-closed. Security is clean: namespaces are validated before reaching dmsetup and all subprocess calls use list form. All 265 new and changed tests pass (the only environment failure, the web3 pytest_ethereum plugin import, is pre-existing and unrelated). Two non-blocking observations below; otherwise only comment-level notes.

**`src/aleph/vm/agent/vm/cache.py`** (line 736): Minor observation: evict_caches is called with needed={root: content_length} but iterates every cache root, so a download into RUNTIME_CACHE can also evict from an over-budget CODE/DATA/MESSAGE cache. That seems like the right behavior (the budget is per-root and enforced globally), but the admit_download docstring speaks of 'the cache' in the singular - worth a sentence saying the eviction sweep covers all roots, if it is indeed intended.

**`src/aleph/vm/agent/vm/cache.py`** (line 738): The budget recheck here is an O(entries) stat walk of the cache root on the event loop, per download, with no memoization. The worst part (reclaim_retained=False, i.e. no rmtree) is already handled, and the reconciler pass is the real enforcement, so this is only a latency consideration on caches with many entries - but if cache roots are expected to hold thousands of entries, a short-TTL size cache (similar to reclaimable_bytes) for _root_usage could be worth it.

foxpatch-aleph (approved):
Well-engineered and thoroughly tested. The in-stream size caps are airtight (Content-Length rejected before file creation, mid-stream abort unlinks the .part, admission runs before the first byte is written), the budgeted caches with LRU eviction never unplug a live VM's references, and the admission hook's synchronous position on the event loop makes its read-evict-reserve sequence atomic. The fail-closed design is consistent: an unlistable supervisor skips the cache pass, parent_device_is_free treats an unanswerable holders query as held, and orphan markers claim directories exclusively so a concurrent GONE marker's owner and parent pins survive. iter_content_refs as the single enumeration behind both depends_on and the live-ref set is the right factoring and prevents the two from drifting. No injection surface: _safe_ref guards every ref reaching dmsetup/losetup and namespaces go through _checked_namespace. All 265 tests across the touched files pass; the full-suite failures I could reproduce (test_dns_configuration, test_vprogram_launch, test_supervisor_translate) are pre-existing environment issues (sandbox lacks writable /var/lib/aleph and /var/cache/aleph) in files this PR does not touch. The three observations below are non-blocking.

**`src/aleph/vm/agent/vm/cache.py`** (line 420): Non-blocking: create_task_log_exceptions schedules remove_parent_devices as a fire-and-forget task, but asyncio only keeps a weak reference to tasks, so the interpreter may garbage-collect it mid-flight (the task and any of its exceptions would simply vanish). If that has ever been observed as devices left behind, keeping the returned task in a module-level set with a done-callback discard (the asyncio docs' own pattern) would close the gap; as is, the leaked-cache-loop sweep on the next reconcile pass is a plausible backstop, so this is an observation rather than a request.

**`src/aleph/vm/agent/vm/reconciler.py`** (line 829): Non-blocking: _release_cache_devices (and therefore sweep_leaked_cache_loops) runs unconditionally, including when the cache pass was skipped because live_known was False. That is safe — the sweep is holder-checked and fail-closed, so a loop device backing a live VM's file is never detached — but running it when the pass was deliberately skipped for distrust of the live set is surprising. A `if live_known` guard on the sweep, or a comment noting it is deliberately unconditional, would make the intent explicit.

**`src/aleph/vm/storage.py`** (line 703): Worth confirming this is intended: get_existing_file now enforces MAX_RUNTIME_ARCHIVE_SIZE (100 GiB) where the path was previously unbounded, so an immutable volume (or workload image, or hash tree) larger than 100 GiB now fails admission with InsufficientResourcesError where it would previously have downloaded. That reads as a deliberate policy tightening and 100 GiB comfortably covers every consumer (vprogram launches, snp staging, translate), but it is a behavior change for over-large immutable content that the PR description does not call out.

foxpatch-aleph (changes requested):
The overall design here is strong: fail-closed handling of the live set (skipping the cache pass when the supervisor can't be listed or a live VM lacks a registry record), download admission that distinguishes a Content-Length measurement from a max_bytes ceiling (unknown-length downloads charge min(cap, budget) and never evict), atomic .reclaimable markers with exclusive orphan writes, two-phase eviction that reclaims retained dirs to free parent images, sysfs-based leaked-loop recovery, and the in-stream size caps with FileTooLargeError correctly mapped to 400 and not retried. Test coverage is unusually thorough. However, there is one correctness bug in the invariant the whole cache pass rests on: iter_content_refs, the 'single enumeration of what a VM needs out of the download caches', misses the per-volume hash_tree refs of a confidential V-PROGRAM. vprogram_launch.py fetches both volume.ref and volume.hash_tree via get_existing_file into DATA_CACHE and attaches both as disks (DiskSpec(path=volume_hash_tree, ...)), but _iter_volume_refs only yields ('volume', ref). A live V-PROGRAM's volume hash tree therefore looks unreferenced to live_refs(), and evict_caches can evict it -- unlinking a running VM's disk, the exact failure the iter_content_refs docstring says this enumeration exists to prevent. The tests protect the workload hash tree but not the per-volume ones. A one-line fix (also yield each volume's hash_tree) plus a test closes it.

**`src/aleph/vm/agent/vm/reclaimable.py`** (line 209): This loop yields each V-PROGRAM volume's own ref but not its hash_tree. A VerifiedVolume has fields ref/hash_tree/roothash/comment, and vprogram_launch.py fetches BOTH into DATA_CACHE (volume_data = await get_existing_file(str(volume.ref)); volume_hash_tree = await get_existing_file(str(volume.hash_tree))) and attaches both as disks. Since iter_content_refs is the single source live_refs() uses to pin cache entries for live VMs, a live V-PROGRAM's per-volume hash tree appears unreferenced and evict_caches can evict it -- unlinking a running VM's disk, exactly what the docstring at line 243 ('Getting the second one short is how a running VM's disk gets unlinked') says this enumeration prevents. _iter_measured_refs already handles the workload image and its hash tree; the per-volume hash trees need the same. Suggested fix: after yielding the volume ref, also yield the volume's hash_tree when present, e.g. hash_tree = getattr(volume, "hash_tree", None) / if hash_tree: yield "volume_hash_tree", str(hash_tree). Please also add a test asserting a V-PROGRAM's volume hash trees stay pinned, mirroring the existing workload-hash-tree test.

**`src/aleph/vm/agent/vm/cache.py`** (line 362): Non-blocking: the fire-and-forget release_parent_devices leaves a narrow race. _may_evict checks parent_device_is_free at eviction time, then the teardown runs later via create_task_log_exceptions; a concurrent create that attaches its loop between the eviction check and this task's dmsetup remove --retry is in the window where /dev/mapper/<ref> has no holders yet, so this teardown can pull the shared device (and, via losetup -d, the loop) out from under it and fail the create. The cache_path.exists() re-check covers the re-download case, but re-checking holders here (e.g. skipping when not parent_device_is_free) or at least documenting the window would close it. The kernel refuses to remove a held dm device, so the residual exposure is limited to the pre-dmsetup-create instant and costs a failed (retryable) create rather than corruption -- hence non-blocking.

**`src/aleph/vm/agent/vm/reclaimable.py`** (line 252): Minor: iter_content_refs documents itself as the one enumeration of 'what a VM needs out of the download caches' so that depends_on and refs_from_content cannot drift apart, but nothing enforces that new disk sources are added here. The missed V-PROGRAM volume hash_tree (see the comment on _iter_volume_refs) is precisely this failure mode. Consider a short docstring note or a test that walks the disk specs a create attaches for each content type and asserts each backing cache ref appears in refs_from_content(output) -- that would turn the invariant from convention into something checked.

foxpatch-aleph (changes requested):
Well-engineered change with strong fail-closed discipline and excellent test coverage, but the central "single enumeration of what a live VM names" is incomplete: VerifiedVolume.hash_tree is fetched from DATA_CACHE and attached as a disk (vprogram_launch.py:260-261) yet _iter_volume_refs only yields volume.ref, so a live V-PROGRAM's volume hash-tree entries are evictable, violating the PR's own never-evict-what-a-live-VM-names invariant. The fix is a small getattr-based yield in _iter_volume_refs; with that (and ideally the admission fast-path) this is an approve.

**`src/aleph/vm/agent/vm/reclaimable.py`** (line 209): Missing ref kind: a V-PROGRAM's VerifiedVolume has a hash_tree alongside ref, and vprogram_launch.py:260-261 fetches both from DATA_CACHE via get_existing_file and attaches both as disks — exactly like the workload pair, whose hash_tree IS protected (kind "workload_hash_tree"). As written, a live V-PROGRAM's volume hash-tree entries are unreferenced and evictable by the cache pass, which is the class of omission this enumeration exists to prevent. Suggest yielding ("volume_hash_tree", str(volume.hash_tree)) alongside the volume ref via getattr, so regular PersistentVolume/RootfsVolume (no hash_tree) are unaffected.

**`src/aleph/vm/agent/vm/cache.py`** (line 734): This evicts on every measured download whenever the live set is trusted, even when the root already fits the download. evict_caches unconditionally computes live_refs(registry) — a JSON read of every cached runtime/TEE manifest, one per live record — plus cache_entries (a stat per entry) on all four roots, on the event loop, on every download attempt (up to 10 per file). Consider computing usage first and skipping the eviction entirely when usage <= budget, so the common case costs one directory walk without the manifest reads.

**`src/aleph/vm/agent/vm/cache.py`** (line 276): referenced_hashes is only used by tests (test_cache_eviction.py). Either use it in evict_caches' call path or trim it; as is, it duplicates the live_refs | marker-refs composition that evict_caches performs inline.

**`src/aleph/vm/agent/vm/cache.py`** (line 463): Naming nit: live_only holds item-hash refs (live_refs(registry)), not namespaces, which misreads next to the is_live callable in the same dataclass that takes a namespace. Something like live_refs would be unambiguous.

foxpatch-aleph (changes requested):
Well-engineered and carefully fail-closed, with excellent tests and docs; the download caps, reservation ledger, admission hook and device teardown all check out, and all new/modified tests pass. However, there is one genuine gap in the live-ref enumeration: a V-PROGRAM's verified-volume hash_tree (vprogram_launch.py:260-263 attaches both volume.ref and volume.hash_tree as the live VM's disks straight out of DATA_CACHE) is not yielded by _iter_volume_refs, so the cache pass can evict a running confidential VM's disk file — the same class of entry the PR explicitly protects for workload hash trees. Fix the enumeration and add a test with a non-empty content.volumes; the rest is minor.

**`src/aleph/vm/agent/vm/reclaimable.py`** (line 210): _iter_volume_refs yields each volume's ref but not its hash_tree. For a V-PROGRAM, VerifiedVolume (aleph_message.models.execution.vprogram) has fields ref/hash_tree/roothash, and vprogram_launch.py:260-263 attaches BOTH volume.ref and volume.hash_tree as the live VM's disks straight out of DATA_CACHE — the exact case this enumeration exists to protect (the workload pair gets the same treatment in _iter_measured_refs two functions up). I verified that for a live V-PROGRAM with one verified volume, refs_from_content() omits the hash-tree ref, so evict_caches can unlink a running confidential VM's disk file: the VM survives on its open fd, but the eviction frees nothing (blocks stay pinned) and violates the "never evicted" invariant. Suggested fix: also yield getattr(volume, "hash_tree", None) here (e.g. as kind "volume_hash_tree").

**`tests/supervisor/test_cache_eviction.py`** (line 107): _vprogram_record always sets content.volumes = [], so the verified-volume path of iter_content_refs is untested — which is why the missing volume.hash_tree yield above goes unnoticed. Please add a test with a V-PROGRAM record whose volumes contain a VerifiedVolume-like object with both ref and hash_tree, asserting neither entry is evicted (mirroring test_a_live_vprograms_workload_disks_are_never_evicted).

**`src/aleph/vm/agent/vm/cache.py`** (line 525): Minor: _markers() calls iter_reclaimable(), which walks every namespace directory across all pools and is not cached (unlike reclaimable_bytes, which caches by pool fingerprint). From evict_caches() on the admission path this runs on the event loop once per over-budget root, so on a node with many namespace dirs admission pauses the loop for a full marker walk; the "few stats per entry" note in the PR description may undersell this. Consider reusing the fingerprint cache or documenting the expected scale.

**`src/aleph/vm/storage.py`** (line 129): Nit: reserve_download() entries are only released in download_file's finally (keyed on owns_tmp). Every current caller of download_file_in_chunks goes through download_file, so this is fine today, but a future direct call against a cache .part path would leak a reservation for the process lifetime and permanently consume the root's budget. Worth one sentence in reserve_download's docstring stating that download_file is the intended owner of the reservation lifecycle.

foxpatch-aleph (approved):
This is a well-engineered PR: the in-stream size caps are correctly enforced both before the file is opened and per-chunk; the LRU admission/eviction logic is carefully fail-closed everywhere it can't answer a question (unlistable supervisor, missing registry record, held dm device, unreadable sysfs); and the shared ref enumeration in reclaimable.py genuinely unifies the depends_on and never-evict questions. Reservations are leak-free across the download retry loop, and in-flight accounting correctly avoids double-counting a growing .part. All 275 tests across the touched files pass. The findings below are performance/robustness observations and nits, none blocking.

**`src/aleph/vm/agent/vm/cache.py`** (line 734): admit_download runs the full evict_caches walk (all four roots, one iterdir+stat per entry, live_refs over the whole registry, manifest reads, shutil.disk_usage) on the event loop for every admitted download, with no fast path when the target root is already under budget — evict_caches only checks over_budget after it has already listed every root. On a node with large caches and frequent downloads this is a per-download O(total cache entries) syscall walk inside the create path. Consider checking _root_usage for the target root first and skipping evict_caches entirely when it fits, and/or short-circuiting the per-root entry listing before the budget check. The PR documents the walk as bounded, but a fast path would keep the common case off the loop entirely.

**`src/aleph/vm/agent/vm/cache.py`** (line 817): sweep_leaked_cache_loops aborts on the first dmsetup/losetup failure (the exception propagates out of the for loop), so one failing device skips the rest of the leaked loops until the next pass. remove_parent_devices, by contrast, isolates per-ref failures so one bad device never strands the others. Wrapping the per-item body in try/except would make the sweep equally robust.

**`src/aleph/vm/storage.py`** (line 683): remove_base_device duplicates the base-removal tail of remove_devmapper (lines 660-664): the same glob-siblings-then-dmsetup-remove block exists in both. remove_devmapper could call remove_base_device (or share a helper) so the two can't drift; the docstrings already describe them as the same operation.

**`src/aleph/vm/agent/vm/cache.py`** (line 746): The required/available figures in InsufficientResourcesError are floor-divided to MiB, so a sub-MiB download or remaining headroom reports required=0/available=0, which is confusing in a 503 payload ("0 MiB required, 0 MiB available" for a refused 512-byte download). Consider ceiling for required or reporting bytes when below 1 MiB.

**`tests/supervisor/test_reconciler.py`** (line 954): The `# noqa: F811` on test_make_room_never_evicts_a_directory_a_create_is_using appears stale: there is only one definition of this test in the file (it was presumably moved here), so the suppression redefinition-of-unused-name directive is doing nothing and can be dropped.

foxpatch-aleph (commented):

foxpatch-aleph (changes requested):
Overall this is a careful, well-documented implementation: the fail-closed rules are consistent (cache pass skipped when the live set is untrusted, admission refusing to evict without a snapshot or with an incomplete registry, parent devices treated as held when sysfs can't answer), the single ref enumeration in reclaimable.py is the right structural fix, the in-flight reservation accounting closes a real concurrent-admission hole, and the subprocess argument validation is sound. All of the PR's new tests pass when run. The one blocking problem is that a new test never actually executes: test_a_retire_without_a_record_asks_device_mapper_instead is missing @pytest.mark.asyncio, so pytest silently skips it (the project's pytest config has no asyncio_mode=auto), leaving the new record-less fallback in teardown_vm_devices without any coverage — I verified the test passes once the decorator is added. The remaining comments are design observations and nits, not blockers.

**`tests/supervisor/test_retire.py`** (line 331): This test is missing its `@pytest.mark.asyncio` decorator, so it never runs: pytest reports it as skipped with a PytestUnhandledCoroutineWarning, and the project's pytest config does not enable asyncio auto-mode. I verified the test passes when actually executed (`-o asyncio_mode=auto`), so please add the decorator — as it stands, the new record-less fallback in `teardown_vm_devices` (retire.py:137-140) has zero coverage, and a future regression there would silently delete disks at retire time.

**`src/aleph/vm/agent/vm/cache.py`** (line 673): Design observation: after any retire the registry drops the record while `_live_snapshot` keeps the hash until the next pass, so this check finds an 'unknown' namespace and admission stops evicting at all — an over-budget download then gets a 503 instead of making room. Under `VOLUME_RETENTION=reap` no pass is triggered after a GONE (the after-gone hook in retire.py is keep-only), so the window can last up to VOLUME_RECONCILE_INTERVAL. Fail-closed is the right instinct; consider triggering `reconcile_now` on GONE under reap as well, or otherwise narrowing the window, so a routine retire doesn't disable admission eviction for an hour.

**`src/aleph/vm/agent/vm/cache.py`** (line 736): Admission asks for room on one root but gets all four: `evict_caches` runs `cache_entries`/`in_flight_bytes` for every cache root on every admitted download, and once any root is over budget `_markers()` additionally walks every namespace directory on every pool — all synchronously on the event loop. The PR notes the directory walks, but the pool-wide marker walk plus the manifest reads in `_record_refs` are heavier than 'a few stats per entry' on a loaded node. Consider a `roots=[root]` filter for the admission call path, or at least call this out in the known limitations.

**`src/aleph/vm/agent/vm/cache.py`** (line 746): Nit: `available` is the room left inside CACHE_BUDGET, not the filesystem's free space, and `required` is the full download size even when most of it would fit — a client using these figures to decide whether a retry could succeed will be misled by the 503's numbers. Maybe worth naming the fields `cache_budget_mib`-style or documenting the semantics.

**`src/aleph/vm/agent/vm/cache.py`** (line 336): Nit: if RUNTIME_CACHE were ever configured to alias another cache directory (they share a filesystem by default, only differing in path), every evicted entry of that root would be treated as a parent image and routed through `remove_parent_device`. It degrades safely today (no device means no-op), just worth being aware of.

**`src/aleph/vm/storage.py`** (line 325): Nit: `get_message` writes `<hash>.json` directly and never goes through `download_file`, so message-cache hits are never mtime-touched and that cache's LRU order is download order. Harmless today (entries are a few KB) and the docs say so, but a one-line `_touch_cache_hit` on the hit path would make the message cache's LRU a use signal too, consistent with the other three.

PR #1162, opened 2026-08-25
feat(cli): aleph-vm storage status, list, reclaim and reconcile (2.1, PR E)

odesenfans:
## Summary

PR E of the 2.1 disk reclamation work (spec Section 6, operator CLI). Stacked on #1161. Last PR of the stack.

`aleph-vm storage ...`, runnable without the agent daemon (filesystem plus the agent DB):

- `status`: per pool live / reclaimable / budget / free bytes, then per cache root used / budget.
- `list [--reclaimable]`: hash, pool, size, reason, age.
- `reclaim <hash>`: purges one directory carrying a `.reclaimable` marker; refuses an unmarked or live hash; exits non-zero if the purge was refused (device-mapper hold).
- `reconcile [--dry-run] [--trust-registry]`: one reconciler pass (same passes and the same cache-device teardown as the daemon's), printing what it removed or would remove.

Fail-closed rule, same as the daemon: `reconcile` and `reclaim` build the live set from the registry union the supervisor's `list_vms()` when the supervisor is reachable (3 s connect timeout); when it is not, `reconcile` runs dry with a warning on stderr and exit code 2 unless `--trust-registry` is passed, and `reclaim` refuses. `status` and `list` never dial the supervisor.

## Known limitation
- The daemon's in-flight-create guard (`creating()`) is process-local; from the CLI only the age of the directory protects a create in flight, for its directory and its devices alike (the orphan-device scan now applies `VOLUME_CREATE_GUARD` too). A real `reconcile` while the daemon is creating a VM that has outlived the guard and is not yet in the registry DB or `list_vms` can remove that create. Stated in `reconcile --help` and the docs; the daemon's own pass is preferred when it is up.
- A CLI pass shares no lock with the daemon's passes; running both at once is benign (idempotent purges, exclusive marker writes).
- Exit code `3` marks a reconcile that downgraded to a dry run; `2` stays argparse's usage error.

## Test plan
- `tests/supervisor`: failing-ID set identical to base.
- New `tests/supervisor/test_storage_cli.py` (24 tests: real temp pools and markers, fake supervisor for all three reachability states, dry-run inertness, exit codes, device teardown seams).

### DIFF
--- a/docs/architecture/storage.md
+++ b/docs/architecture/storage.md
@@ -560,12 +560,19 @@ process*:
   applies), reclaim reports that and exits non-zero rather than claiming
   success.
 - `storage reconcile [--dry-run] [--trust-registry]`: run one
-  `reconcile_storage()` pass, printing what it removed (or, under
-  `--dry-run`, what it would remove), then tears down the devices of any
-  evicted runtime cache parent the same way `reconcile_now` /
-  `reconcile_at_startup` do (`remove_parent_device` per evicted ref, then
-  `sweep_leaked_cache_loops`), so an evicted entry never leaves
-  `/dev/mapper/<ref>` and its loop device pinning the deleted inode.
+  `reconcile_storage()` pass with the daemon's two device steps around it,
+  or the CLI would keep refusing what the daemon reclaims. Before the pass,
+  `_teardown_orphan_devices` removes the dm devices of every namespace no
+  live VM owns, so the purge that follows is not refused on a volume file a
+  target still holds; that step needs the supervisor's answer and is skipped
+  under `--dry-run` and when the supervisor is unreachable, `--trust-registry`
+  included (it buys a purge on the registry's word, not a teardown). After
+  the pass, and printing what it removed (or, under `--dry-run`, what it
+  would remove), it tears down the devices of any evicted runtime cache
+  parent the same way `reconcile_now` / `reconcile_at_startup` do
+  (`remove_parent_device` per evicted ref, then `sweep_leaked_cache_loops`),
+  so an evicted entry never leaves `/dev/mapper/<ref>` and its loop device
+  pinning the deleted inode.
 
 `reconcile` and `reclaim` can purge, so they apply the same fail-closed rule
 `reconciler._startup_refusal` applies to the daemon's own startup pass: a
--- a/docs/architecture/storage.md
+++ b/docs/architecture/storage.md
@@ -551,17 +551,20 @@ process*:
   orphan no pass has reached yet is not a live VM). Read-only, registry
   live set only.
 - `storage reclaim <hash> [--trust-registry]`: purge one reclaimable
-  directory now. Checks the `.reclaimable` marker first, purely locally, so
-  a typo or an unrelated hash fails instantly rather than waiting on a
-  supervisor dial that could only ever confirm what the marker check already
-  knows; a directory with no marker (it may belong to a live VM) is refused
-  the same way. Then refuses a hash the agent registry or the supervisor
-  considers live (see below), and refuses outright, without
-  `--trust-registry`, when the supervisor cannot be asked. Finally, if the
-  purge itself leaves the directory behind (a device-mapper target still
-  holds one of its volumes, the same guard `purge_vm_storage` always
+  directory now. Checks the name and the `.reclaimable` marker first, purely
+  locally, so a typo or an unrelated hash fails instantly rather than
+  waiting on a supervisor dial that could only ever confirm what those
+  checks already know; a name that is not a VM hash (a hand-made marker
+  under `backup_old`) and a directory with no marker (it may belong to a
+  live VM) are refused the same way. Then refuses a hash the agent registry
+  or the supervisor considers live (see below), and refuses outright,
+  without `--trust-registry`, when the supervisor cannot be asked. Finally,
+  if the purge itself leaves the directory behind (a device-mapper target
+  still holds one of its volumes, the same guard `purge_vm_storage` always
   applies), reclaim reports that and exits non-zero rather than claiming
-  success.
+  success. `reclaim` never tears devices down itself: `storage reconcile`
+  is the CLI path that does, for every VM nothing owns, so the refusal
+  points the operator there.
 - `storage reconcile [--dry-run] [--trust-registry]`: run one
   `reconcile_storage()` pass with the daemon's two device steps around it,
   or the CLI would keep refusing what the daemon reclaims. Before the pass,
--- a/docs/architecture/storage.md
+++ b/docs/architecture/storage.md
@@ -539,7 +539,8 @@ goes first", which is the only promise a node can honestly sell.
 
 Storage is agent-side and readable from the filesystem plus the agent DB, so
 `aleph-vm storage ...` (`src/aleph/vm/agent/storage_cli.py`, dispatched by
-`agent/cli.py` before its own argument parser) needs no running *agent
+`agent/cli.py` before its own argument parser, which points at
+`storage --help` in its own `--help` epilog) needs no running *agent
 process*:
 
 - `storage status`: per pool, live / reclaimable / cache / free bytes
@@ -547,14 +548,24 @@ process*:
 - `storage list [--reclaimable]`: hash, pool, size, reason, age. Read-only,
   registry live set only.
 - `storage reclaim <hash> [--trust-registry]`: purge one reclaimable
-  directory now. Refuses a directory with no `.reclaimable` marker (it may
-  belong to a live VM); an implausible or unrelated hash simply matches
-  nothing on disk, so it is refused the same way. Also refuses a hash the
-  agent registry or the supervisor considers live (see below), and refuses
-  outright, without `--trust-registry`, when the supervisor cannot be asked.
+  directory now. Checks the `.reclaimable` marker first, purely locally, so
+  a typo or an unrelated hash fails instantly rather than waiting on a
+  supervisor dial that could only ever confirm what the marker check already
+  knows; a directory with no marker (it may belong to a live VM) is refused
+  the same way. Then refuses a hash the agent registry or the supervisor
+  considers live (see below), and refuses outright, without
+  `--trust-registry`, when the supervisor cannot be asked. Finally, if the
+  purge itself leaves the directory behind (a device-mapper target still
+  holds one of its volumes, the same guard `purge_vm_storage` always
+  applies), reclaim reports that and exits non-zero rather than claiming
+  success.
 - `storage reconcile [--dry-run] [--trust-registry]`: run one
   `reconcile_storage()` pass, printing what it removed (or, under
-  `--dry-run`, what it would remove).
+  `--dry-run`, what it would remove), then tears down the devices of any
+  evicted runtime cache parent the same way `reconcile_now` /
+  `reconcile_at_startup` do (`remove_parent_device` per evicted ref, then
+  `sweep_leaked_cache_loops`), so an evicted entry never leaves
+  `/dev/mapper/<ref>` and its loop device pinning the deleted inode.
 
 `reconcile` and `reclaim` can purge, so they apply the same fail-closed rule
 `reconciler._startup_refusal` applies to the daemon's own startup pass: a
@@ -566,11 +577,24 @@ supervisor over the same gRPC socket the agent uses
 (`settings.SUPERVISOR_GRPC_SOCKET`, a few seconds' timeout) and union its
 `list_vms()` answer into the live set (`reconciler.supervisor_hashes`, the
 same id-to-hash mapping the daemon pass uses). When the supervisor cannot be
-reached: `reconcile` runs as a dry run and prints a warning explaining why,
-unless `--trust-registry` says to purge on the registry alone (`--dry-run`
-always wins, with or without the flag); `reclaim` refuses outright unless
-`--trust-registry` is given. `status` and `list` are read-only and never
-dial the supervisor.
+reached and an explicit `--dry-run` was not requested, `reconcile` still
+runs as a dry run (a preview, nothing removed) but treats this as an error:
+it prints a warning to stderr (`"registry-only"`, not `"trusted"`: the
+report on stdout stays parseable, uninterrupted by the warning) and exits
+`2`, unless `--trust-registry` says to purge on the registry alone. An
+explicit `--dry-run` always stays exit `0`, whether or not the supervisor
+answered: nothing was silently downgraded, the caller asked for a preview
+and got one. `reclaim` refuses outright unless `--trust-registry` is given.
+`status` and `list` are read-only and never dial the supervisor.
+
+Neither command can see a create the daemon is in the middle of:
+`reconciler.is_creating()` is in-process state of the running agent, so a
+CLI invocation (a separate process) never sees it. A create that has
+outlived `VOLUME_CREATE_GUARD` and has not yet landed in the registry or in
+`list_vms` looks exactly like an orphan to a real `reconcile` run from the
+CLI, which can then remove it out from under the daemon. `reconcile --help`
+says so; the daemon's own pass (startup, periodic, at-GONE) is preferred
+whenever the daemon is up, and this command is mainly for when it is not.
 
 ### First start on an existing node
 
--- a/docs/architecture/storage.md
+++ b/docs/architecture/storage.md
@@ -539,24 +539,38 @@ goes first", which is the only promise a node can honestly sell.
 
 Storage is agent-side and readable from the filesystem plus the agent DB, so
 `aleph-vm storage ...` (`src/aleph/vm/agent/storage_cli.py`, dispatched by
-`agent/cli.py` before its own argument parser) needs no running daemon:
+`agent/cli.py` before its own argument parser) needs no running *agent
+process*:
 
 - `storage status`: per pool, live / reclaimable / cache / free bytes
-  against the budgets.
-- `storage list [--reclaimable]`: hash, pool, size, reason, age.
-- `storage reclaim <hash>`: purge one reclaimable directory now. Refuses a
-  directory with no `.reclaimable` marker (it may belong to a live VM) and,
-  since it only ever matches a directory that already exists on disk, an
-  implausible or unrelated hash simply matches nothing.
-- `storage reconcile [--dry-run]`: run one `reconcile_storage()` pass,
-  printing what it removed (or, under `--dry-run`, what it would remove).
+  against the budgets. Read-only, registry live set only.
+- `storage list [--reclaimable]`: hash, pool, size, reason, age. Read-only,
+  registry live set only.
+- `storage reclaim <hash> [--trust-registry]`: purge one reclaimable
+  directory now. Refuses a directory with no `.reclaimable` marker (it may
+  belong to a live VM); an implausible or unrelated hash simply matches
+  nothing on disk, so it is refused the same way. Also refuses a hash the
+  agent registry or the supervisor considers live (see below), and refuses
+  outright, without `--trust-registry`, when the supervisor cannot be asked.
+- `storage reconcile [--dry-run] [--trust-registry]`: run one
+  `reconcile_storage()` pass, printing what it removed (or, under
+  `--dry-run`, what it would remove).
 
-The live set these commands act against is the agent registry alone
-(rehydrated from the agent DB on startup, `rehydrate_registry`): there is no
-daemon to ask `list_vms`, so this cannot cross-check against a supervisor's
-answer the way the startup pass does (`reconciler._startup_refusal`). An
-operator invoking this by hand already knows what is running; the CLI does
-not add a `--trust-registry` gate on top of the registry it already trusts.
+`reconcile` and `reclaim` can purge, so they apply the same fail-closed rule
+`reconciler._startup_refusal` applies to the daemon's own startup pass: a
+live set built from the registry alone purges the disks of VMs the
+supervisor still runs, since a fresh or partly-lost agent DB rehydrates into
+an incomplete registry (this was PR A's Critical 1). A CLI process holding
+only the registry is exactly that state, so both commands dial the
+supervisor over the same gRPC socket the agent uses
+(`settings.SUPERVISOR_GRPC_SOCKET`, a few seconds' timeout) and union its
+`list_vms()` answer into the live set (`reconciler.supervisor_hashes`, the
+same id-to-hash mapping the daemon pass uses). When the supervisor cannot be
+reached: `reconcile` runs as a dry run and prints a warning explaining why,
+unless `--trust-registry` says to purge on the registry alone (`--dry-run`
+always wins, with or without the flag); `reclaim` refuses outright unless
+`--trust-registry` is given. `status` and `list` are read-only and never
+dial the supervisor.
 
 ### First start on an existing node
 
--- a/docs/architecture/storage.md
+++ b/docs/architecture/storage.md
@@ -545,8 +545,11 @@ process*:
 
 - `storage status`: per pool, live / reclaimable / cache / free bytes
   against the budgets. Read-only, registry live set only.
-- `storage list [--reclaimable]`: hash, pool, size, reason, age. Read-only,
-  registry live set only.
+- `storage list [--reclaimable]`: hash, pool, size, reason, age. The reason
+  is the marker's for a reclaimable directory; an unmarked directory reads
+  `live` when the registry knows its hash and `unmarked` otherwise (an
+  orphan no pass has reached yet is not a live VM). Read-only, registry
+  live set only.
 - `storage reclaim <hash> [--trust-registry]`: purge one reclaimable
   directory now. Checks the `.reclaimable` marker first, purely locally, so
   a typo or an unrelated hash fails instantly rather than waiting on a
@@ -588,7 +591,9 @@ reached and an explicit `--dry-run` was not requested, `reconcile` still
 runs as a dry run (a preview, nothing removed) but treats this as an error:
 it prints a warning to stderr (`"registry-only"`, not `"trusted"`: the
 report on stdout stays parseable, uninterrupted by the warning) and exits
-`2`, unless `--trust-registry` says to purge on the registry alone. An
+`3` (not `2`, which argparse uses for a usage error, so a wrapper script
+can tell the two apart), unless `--trust-registry` says to purge on the
+registry alone. An
 explicit `--dry-run` always stays exit `0`, whether or not the supervisor
 answered: nothing was silently downgraded, the caller asked for a preview
 and got one. `reclaim` refuses outright unless `--trust-registry` is given.
--- a/docs/architecture/storage.md
+++ b/docs/architecture/storage.md
@@ -535,6 +535,29 @@ and an attacker does not need to stop paying to create them (create, forget,
 repeat). It means "kept as long as the budget and live demand allow, oldest
 goes first", which is the only promise a node can honestly sell.
 
+### CLI
+
+Storage is agent-side and readable from the filesystem plus the agent DB, so
+`aleph-vm storage ...` (`src/aleph/vm/agent/storage_cli.py`, dispatched by
+`agent/cli.py` before its own argument parser) needs no running daemon:
+
+- `storage status`: per pool, live / reclaimable / cache / free bytes
+  against the budgets.
+- `storage list [--reclaimable]`: hash, pool, size, reason, age.
+- `storage reclaim <hash>`: purge one reclaimable directory now. Refuses a
+  directory with no `.reclaimable` marker (it may belong to a live VM) and,
+  since it only ever matches a directory that already exists on disk, an
+  implausible or unrelated hash simply matches nothing.
+- `storage reconcile [--dry-run]`: run one `reconcile_storage()` pass,
+  printing what it removed (or, under `--dry-run`, what it would remove).
+
+The live set these commands act against is the agent registry alone
+(rehydrated from the agent DB on startup, `rehydrate_registry`): there is no
+daemon to ask `list_vms`, so this cannot cross-check against a supervisor's
+answer the way the startup pass does (`reconciler._startup_refusal`). An
+operator invoking this by hand already knows what is running; the CLI does
+not add a `--trust-registry` gate on top of the registry it already trusts.
+
 ### First start on an existing node
 
 The first pass on an upgraded node finds every directory leaked by the bugs
--- a/docs/architecture/storage.md
+++ b/docs/architecture/storage.md
@@ -596,12 +596,22 @@ and got one. `reclaim` refuses outright unless `--trust-registry` is given.
 
 Neither command can see a create the daemon is in the middle of:
 `reconciler.is_creating()` is in-process state of the running agent, so a
-CLI invocation (a separate process) never sees it. A create that has
-outlived `VOLUME_CREATE_GUARD` and has not yet landed in the registry or in
-`list_vms` looks exactly like an orphan to a real `reconcile` run from the
-CLI, which can then remove it out from under the daemon. `reconcile --help`
-says so; the daemon's own pass (startup, periodic, at-GONE) is preferred
-whenever the daemon is up, and this command is mainly for when it is not.
+CLI invocation (a separate process) never sees it. What protects such a
+create from the CLI is the age of its directory alone: the directory purge
+and the orphan-device teardown (`orphan_device_namespaces`) both skip a
+namespace whose directory is younger than `VOLUME_CREATE_GUARD`. The
+registry record reaches the DB only once the VM runs and `list_vms`
+answers only once `create_vm` returned, so a create that has outlived the
+guard before either happened looks exactly like an orphan to a real
+`reconcile` run from the CLI, which can then remove its directory and its
+devices out from under the daemon. A CLI pass also holds no lock against
+the daemon's own passes (`_pass_lock` is in-process too) and can run
+concurrently with one; that is benign, since purges are idempotent and
+tolerate a directory that vanished mid-pass and markers are written
+exclusively, but it is one more reason to prefer the daemon's pass.
+`reconcile --help` says so; the daemon's own pass (startup, periodic,
+at-GONE) is preferred whenever the daemon is up, and this command is mainly
+for when it is not.
 
 ### First start on an existing node
 
--- a/src/aleph/vm/agent/cli.py
+++ b/src/aleph/vm/agent/cli.py
@@ -22,7 +22,11 @@ logger = logging.getLogger(__name__)
 
 
 def parse_args(args):
-    parser = argparse.ArgumentParser(prog="aleph-vm", description="Aleph.im compute node agent")
+    parser = argparse.ArgumentParser(
+        prog="aleph-vm",
+        description="Aleph.im compute node agent",
+        epilog="Run 'aleph-vm storage --help' for VM storage reclamation (status, list, reclaim, reconcile).",
+    )
     parser.add_argument(
         "--system-logs",
         action="store_true",
--- a/src/aleph/vm/agent/cli.py
+++ b/src/aleph/vm/agent/cli.py
@@ -151,6 +151,11 @@ async def run_async_db_migrations():
 
 
 def main():
+    if len(sys.argv) > 1 and sys.argv[1] == "storage":
+        from aleph.vm.agent import storage_cli
+
+        sys.exit(storage_cli.main(sys.argv[2:]))
+
     args = parse_args(sys.argv[1:])
 
     log_format = (
--- a/src/aleph/vm/agent/storage_cli.py
+++ b/src/aleph/vm/agent/storage_cli.py
@@ -10,9 +10,19 @@ exactly that state, so ``reconcile`` and ``reclaim`` ask the supervisor
 daemon over the same gRPC socket the agent dials (``GrpcSupervisor``,
 ``settings.SUPERVISOR_GRPC_SOCKET``), with a short timeout, and union its
 answer into the live set. When the daemon cannot be reached, both commands
-fail closed: ``reconcile`` runs dry and warns, ``reclaim`` refuses, unless
-``--trust-registry`` says to proceed on the registry alone. ``status`` and
-``list`` are read-only and never touch the supervisor.
+fail closed: ``reconcile`` runs dry, warns on stderr and exits non-zero,
+``reclaim`` refuses, unless ``--trust-registry`` says to proceed on the
+registry alone. ``status`` and ``list`` are read-only and never touch the
+supervisor.
+
+What this process still cannot see, even with the supervisor reachable: a
+create the daemon is in the middle of. ``reconciler.is_creating()`` is an
+in-process set the running agent populates for the duration of a create; a
+CLI invocation is a separate process and never sees it, so a create that has
+outlived ``VOLUME_CREATE_GUARD`` and has not yet landed in the registry or
+in ``list_vms`` looks exactly like an orphan to a real ``reconcile`` here.
+The daemon's own pass (startup, periodic, at-GONE) is always preferred when
+the daemon is up; this command exists mainly for when it is not.
 """
 
 from __future__ import annotations
@@ -36,6 +46,8 @@ from aleph.vm.agent.vm.reclaimable import (
     reclaimable_bytes,
 )
 from aleph.vm.agent.vm.reconciler import (
+    _release_cache_devices,
+    _still_on_disk,
     live_hashes,
     reconcile_storage,
     supervisor_hashes,
@@ -51,11 +63,20 @@ from aleph.vm.supervisor_interface.abc import Supervisor
 # sit through a 30 second RPC deadline just to learn the daemon is down.
 SUPERVISOR_CONNECT_TIMEOUT_SECS = 3.0
 
+# Exit code for a reconcile that silently downgraded to a dry run because the
+# supervisor could not be asked and --trust-registry was not given. Distinct
+# from an explicit --dry-run, which is a success (exit 0): nothing was
+# downgraded, the caller asked for a preview and got one.
+DEGRADED_EXIT_CODE = 2
+
 logger = logging.getLogger(__name__)
 
 
 def parse_args(argv: list[str]) -> argparse.Namespace:
-    parser = argparse.ArgumentParser(prog="aleph-vm storage", description="VM storage reclamation")
+    parser = argparse.ArgumentParser(
+        prog="aleph-vm storage",
+        description="VM storage reclamation: status, list, reclaim, reconcile.",
+    )
     sub = parser.add_subparsers(dest="command", required=True)
     sub.add_parser("status", help="per-pool and per-cache usage against the budgets")
     list_parser = sub.add_parser("list", help="VM directories on every pool")
@@ -67,7 +88,19 @@ def parse_args(argv: list[str]) -> argparse.Namespace:
         action="store_true",
         help="purge using the registry alone when the supervisor cannot be asked whether the hash is running",
     )
-    reconcile_parser = sub.add_parser("reconcile", help="run one reconciler pass")
+    reconcile_parser = sub.add_parser(
+        "reconcile",
+        help="run one reconciler pass",
+        description=(
+            "Run one reconciler pass against the agent registry (unioned with the supervisor's "
+            "list_vms() when it can be reached). This process cannot see a create the daemon is "
+            "currently in the middle of: is_creating() is in-process state of the running agent, "
+            "so a create that has outlived VOLUME_CREATE_GUARD and is not yet in the registry or "
+            "list_vms looks like an orphan here and a real pass can remove it. Prefer the daemon's "
+            "own pass (it runs at startup, periodically, and after every GONE) when the daemon is "
+            "up; use this command mainly when it is not."
+        ),
+    )
     reconcile_parser.add_argument("--dry-run", action="store_true")
     reconcile_parser.add_argument(
         "--trust-registry",
@@ -86,7 +119,8 @@ def _human(size: int) -> str:
         if value < _UNIT_STEP or unit == "TiB":
             return f"{value:.1f} {unit}" if unit != "B" else f"{int(value)} B"
         value /= _UNIT_STEP
-    return f"{value:.1f} TiB"
+    msg = "unreachable: the loop above always returns on its last unit (TiB)"
+    raise AssertionError(msg)
 
 
 def _age(since: datetime) -> str:
@@ -185,6 +219,15 @@ def _list(out: TextIO, *, reclaimable_only: bool) -> int:
 
 
 def _reclaim(registry: AgentVmRegistry, vm_hash: str, out: TextIO, *, trust_registry: bool) -> int:
+    # Cheapest, purely local check first: a typo or an unrelated hash fails
+    # instantly instead of waiting out a supervisor dial that can only ever
+    # confirm what this check already knows.
+    marked = [directory for directory, _marker in iter_reclaimable() if directory.name == vm_hash]
+    if not marked:
+        out.write(
+            f"{vm_hash} is not reclaimable (no .reclaimable marker); refusing to purge a directory a VM may own\n"
+        )
+        return 1
     if vm_hash in live_hashes(registry):
         out.write(f"{vm_hash} is a live VM in the agent registry; refusing to purge it\n")
         return 1
@@ -199,24 +242,24 @@ def _reclaim(registry: AgentVmRegistry, vm_hash: str, out: TextIO, *, trust_regi
             "Pass --trust-registry to purge using the registry alone\n"
         )
         return 1
-    marked = [directory for directory, _marker in iter_reclaimable() if directory.name == vm_hash]
-    if not marked:
+    deleted = purge_vm_storage(vm_hash)
+    if _still_on_disk(vm_hash):
         out.write(
-            f"{vm_hash} is not reclaimable (no .reclaimable marker); refusing to purge a directory a VM may own\n"
+            f"Purge of {vm_hash} left directories behind: a device-mapper target still holds its volumes; "
+            "retry once it is torn down\n"
         )
         return 1
-    deleted = purge_vm_storage(vm_hash)
     out.write(f"Purged {vm_hash}: {deleted} volume file(s)\n")
     return 0
 
 
 def _reconcile(registry: AgentVmRegistry, out: TextIO, *, dry_run: bool, trust_registry: bool) -> int:
     live, reachable = _cli_live_set(registry)
-    effective_dry_run = dry_run
-    if not reachable and not trust_registry:
-        effective_dry_run = True
-        out.write(
-            "Warning: supervisor unreachable; showing what a trusted pass would purge; "
+    downgraded = not reachable and not trust_registry
+    effective_dry_run = dry_run or downgraded
+    if downgraded:
+        sys.stderr.write(
+            "Warning: supervisor unreachable; showing what a registry-only pass would purge; "
             "pass --trust-registry to purge using the registry alone\n"
         )
     report = reconcile_storage(registry, dry_run=effective_dry_run, live=live)
@@ -226,7 +269,12 @@ def _reconcile(registry: AgentVmRegistry, out: TextIO, *, dry_run: bool, trust_r
         out.write(f"  {'would purge' if effective_dry_run else 'purged'} {name}\n")
     for name in report.marked_orphans:
         out.write(f"  {'would mark' if effective_dry_run else 'marked'} {name}\n")
-    return 0
+    # Mirrors reconcile_now/reconcile_at_startup: tear down the devices of
+    # evicted parent images, then sweep whatever an earlier teardown left
+    # behind. _release_cache_devices no-ops under a dry run on its own, so
+    # calling it unconditionally here matches the daemon's own passes.
+    asyncio.run(_release_cache_devices(report, dry_run=effective_dry_run))
+    return DEGRADED_EXIT_CODE if downgraded and not dry_run else 0
 
 
 def run(args: argparse.Namespace, registry: AgentVmRegistry, out: TextIO) -> int:
--- a/src/aleph/vm/agent/storage_cli.py
+++ b/src/aleph/vm/agent/storage_cli.py
@@ -1,22 +1,25 @@
 """``aleph-vm storage``: show and drive reclamation from the node shell.
 
 Storage is agent-side and, by design, readable from the filesystem plus the
-agent DB, so these commands need no running daemon. They run the same code
-the reconciler runs.
-
-The live set these commands act against is the agent registry alone
-(rehydrated from the agent DB, ``rehydrate_registry``): there is no daemon to
-ask ``list_vms``, so unlike the startup pass (``reconciler._startup_refusal``)
-this cannot cross-check against a supervisor's answer. An operator invoking
-this by hand already knows whether the daemon is up; running it against a
-stopped daemon (the normal case, since the daemon holds no lock these
-commands need) is exactly the documented use.
+agent DB, so these commands need no *agent process* running. They run the
+same code the reconciler runs, including the same safety rule: a live set
+built from the registry alone purges the disks of VMs the supervisor still
+runs (that was PR A's Critical 1, guarded on the daemon side by
+``reconciler._startup_refusal``). A CLI process holding only the registry is
+exactly that state, so ``reconcile`` and ``reclaim`` ask the supervisor
+daemon over the same gRPC socket the agent dials (``GrpcSupervisor``,
+``settings.SUPERVISOR_GRPC_SOCKET``), with a short timeout, and union its
+answer into the live set. When the daemon cannot be reached, both commands
+fail closed: ``reconcile`` runs dry and warns, ``reclaim`` refuses, unless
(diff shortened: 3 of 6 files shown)


foxpatch-aleph (approved):
Well-structured CLI that correctly reuses the daemon's reconciler code paths with a sound fail-closed design. The marker-first check in reclaim, the supervisor union for reconcile/reclaim, and the degraded exit code semantics are all correct. Test coverage is thorough across all three supervisor reachability states, dry-run inertness, exit codes, and device teardown seams. The minor observations (no is_live re-check in reconcile, O(pools*n) in status, no numeric assertions in status test) are all non-blocking and either deliberate design choices or negligible for a CLI tool.

**`src/aleph/vm/agent/storage_cli.py`** (line 265): reconcile_storage is called without is_live, so it defaults to _snapshot_is_live(live) — the snapshot alone, without a live registry re-check. The daemon's path uses registry_is_live(registry, live) which re-checks the registry at removal time. This is correct for a CLI process (its registry is frozen at load time, so a re-check would add nothing), but a one-line comment explaining why is_live is intentionally not passed would help future readers understand the asymmetry.

**`src/aleph/vm/agent/storage_cli.py`** (line 183): live_bytes iterates iter_namespace_dirs() (all pools) inside the per-pool loop and filters by directory.parent == pool.path. This is O(pools x all_dirs) rather than O(all_dirs). Negligible for a CLI tool, but a single pass with grouping by parent would be cleaner.

**`tests/supervisor/test_storage_cli.py`** (line 60): test_status_lists_pools_and_caches checks for presence of headers and paths but doesn't assert on the live/reclaimable/budget/free values. Adding assertions on the numeric values (e.g., that LIVE bytes are non-zero, reclaimable bytes match the marker size) would strengthen this test.

foxpatch-aleph (approved):
Well-structured CLI that correctly reuses the daemon's reconciler code paths with a consistent fail-closed design. The four subcommands (status, list, reclaim, reconcile) handle all three supervisor reachability states correctly, with proper exit codes and stderr/stdout separation. Test coverage is thorough with 17 tests covering edge cases like DM-held directories, dry-run inertness, and cache device teardown. The only notable observation is that main() doesn't configure logging, so diagnostics from underlying functions (e.g., DM-hold warnings from purge_vm_storage) are silently lost when running via the CLI — the user-facing output covers the important cases but debug-level information is missing. All other findings are minor (private function imports, O(n*p) status iteration, a small test gap for --trust-registry with reachable supervisor).

**`src/aleph/vm/agent/storage_cli.py`** (line 299): main() doesn't configure logging (no `logging.basicConfig` or handler setup). The daemon's own main() in cli.py does this at lines 196-201. As a result, log messages from purge_vm_storage, reconcile_storage, and _release_cache_devices (e.g., 'Not deleting %s: its device-mapper target is still present') are silently lost. Consider adding minimal logging configuration, at least for WARNING level, so the underlying functions' diagnostics are visible to the operator.

**`src/aleph/vm/agent/storage_cli.py`** (line 48): Importing private (_-prefixed) functions `_release_cache_devices` and `_still_on_disk` from reconciler.py creates a coupling to internal implementation details. This is a deliberate tradeoff to reuse the exact same code paths as the daemon, which is defensible, but it means changes to reconciler internals could break the CLI silently. Non-blocking.

**`src/aleph/vm/agent/storage_cli.py`** (line 183): The status command calls iter_namespace_dirs() (all pools) inside a per-pool loop, then filters by directory.parent == pool.path. This is O(n*p) instead of O(n). For a CLI status command with a small number of pools this is negligible, but iter_namespace_dirs() could be called once and the results grouped by parent. Non-blocking.

**`tests/supervisor/test_storage_cli.py`** (line 139): Minor test gap: there's no test for `reclaim --trust-registry <hash>` when the supervisor IS reachable and lists the hash as running. The code correctly refuses it (line 235-238 ignores --trust-registry when the supervisor answered), but this behavior isn't explicitly tested. Adding a test would guard against a future change that accidentally lets --trust-registry override a positive supervisor answer.

foxpatch-aleph (approved):
Well-structured PR implementing the operator CLI for storage reclamation (status, list, reclaim, reconcile). The fail-closed rule correctly mirrors the daemon's own safety logic: reconcile dials the supervisor and degrades to a dry run with exit code 2 when unreachable (unless --trust-registry), reclaim refuses outright, and status/list never dial. The device teardown sequence (orphan devices before the walk, cache parent devices after) matches reconcile_now/reconcile_at_startup. The known limitation (is_creating() is in-process, so a CLI reconcile can remove a create the daemon is in the middle of) is clearly documented in the module docstring, the --help text, and docs/architecture/storage.md. Test coverage is solid (17 tests covering all three supervisor reachability states, dry-run inertness, exit codes, device teardown seams, and CLI dispatch). No correctness bugs or security issues found.

**`src/aleph/vm/agent/storage_cli.py`** (line 274): Non-blocking observation: the CLI calls `reconcile_storage` without `is_live=registry_is_live(registry, live)`, so it falls back to `_snapshot_is_live(live)` (the frozen snapshot). This is correct for the CLI context — the CLI's registry is loaded once and never mutated, unlike the daemon's live in-memory dict — but a brief comment noting why `registry_is_live` is intentionally omitted would help future readers who compare the two call sites.

**`src/aleph/vm/agent/storage_cli.py`** (line 184): Minor: `_status` calls `iter_namespace_dirs()` (all dirs) once per pool and filters by `directory.parent == pool.path`. For a node with many pools this is O(pools x dirs). Acceptable for a status command, but a single pass grouping by parent would be cleaner.

**`tests/supervisor/test_storage_cli.py`** (line 78): The `test_status_lists_pools_and_caches` test only covers `VOLUME_RETENTION="keep"`. Consider adding a test for `VOLUME_RETENTION="reap"` to verify the `budget=0` branch on line 194 of storage_cli.py.

foxpatch-aleph (commented):
This PR adds a well-designed operator CLI that faithfully reuses the daemon's reconciler and its fail-closed safety model: the live set is the registry unioned with supervisor.list_vms(), device teardown only runs when the supervisor answered, and the cache pass is skipped when the live set is untrusted. Security is sound (hash-pattern validation via _checked_namespace, no path traversal or injection), the 22 new tests and the pre-existing reconciler/reclaimable/retire/devmapper suites all pass, and the docs match the implementation. The findings below are observations rather than blocking issues.

**`src/aleph/vm/agent/storage_cli.py`** (line 168): The docstring claims this is 'the same fail-closed reasoning reconciler._startup_refusal applies', but only the reachability condition of _startup_refusal is implemented: the second condition (a supervisor that runs N > 0 VMs while the agent registry is empty, i.e. a lost/unreadable agent DB) is not replicated here, so a CLI reconcile in that state proceeds with real purges. Running VMs are still protected by the union with list_vms(), so this is a design choice rather than a bug — but the docstring overstates the equivalence, and the author may want to either implement the empty-registry refusal or soften the wording.

**`src/aleph/vm/agent/storage_cli.py`** (line 272): Worth noting in the docs: _teardown_orphan_devices has no age/mtime guard (unlike the namespace purge, which respects VOLUME_CREATE_GUARD), and is_creating() is always False in a separate process. So a CLI reconcile run while the daemon is creating a VM can tear down that VM's dm devices even while the create is still young — a sharper edge than the documented 'a create that has outlived VOLUME_CREATE_GUARD' framing suggests. The 'prefer the daemon's own pass when it is up' guidance mitigates it, but the reconcile --help text and docs frame the limitation more narrowly than the actual exposure.

**`src/aleph/vm/agent/storage_cli.py`** (line 216): Suggestion: an unmarked directory that is not in the registry (e.g. an orphan younger than VOLUME_CREATE_GUARD, which no pass has marked yet) is reported as reason "live". For an operator that label is misleading — something like "unmarked" or "unknown" would be more honest than "live".

**`src/aleph/vm/agent/storage_cli.py`** (line 259): Minor: downgraded is True even for an explicit --dry-run with an unreachable supervisor, so the operator gets the 'pass --trust-registry' warning on stderr for a run they never expected to purge anything from. Consider excluding dry_run from the downgraded flag so the warning only fires when something was actually downgraded.

**`src/aleph/vm/agent/storage_cli.py`** (line 71): Nit: DEGRADED_EXIT_CODE = 2 collides with argparse's usage-error exit code 2, so a caller cannot distinguish 'reconcile was degraded to a dry run' from 'bad arguments'. Picking 3 (or documenting the collision) would make automation slightly less ambiguous.

**`src/aleph/vm/agent/storage_cli.py`** (line 308): main() loads the DB-backed registry for every command, so `storage list` (which never uses it) fails when the agent DB is unreachable even though it is purely a filesystem walk. Loading the registry lazily, only for status/reclaim/reconcile, would make `list` usable in exactly the degraded states where an operator reaches for this CLI.

**`tests/supervisor/test_storage_cli.py`** (line 43): The fixture's docstring says it never lets a test read the host's real /dev/mapper, but it only patches reconciler_module.DEVICE_MAPPER_DIRECTORY; purge._held_by_device_mapper reads its own module-level copy of the constant and still inspects the real /dev/mapper. It is benign today (the reconcile tests use .qcow2 volumes, and the dm-held test monkeypatches Path.is_block_device), but worth either patching both modules or narrowing the docstring.

**`src/aleph/vm/agent/storage_cli.py`** (line 311): setup_pools() can raise StoragePoolConfigError and an unreachable agent DB will surface as a SQLAlchemy traceback; both land as raw tracebacks on the operator. A try/except that prints a one-line error and exits non-zero would match the polish of the rest of the command surface.

foxpatch-aleph (approved):
PR E adds `aleph-vm storage` (status, list, reclaim, reconcile) as the operator CLI for the 2.1 disk reclamation work. The fail-closed rule is implemented correctly and matches the daemon: reconcile/reclaim union the registry with the supervisor's list_vms() over the same gRPC socket, run dry with exit 2 (or refuse) when the daemon is unreachable unless --trust-registry is given, and mirror the daemon's device teardown steps around the pass. Hash validation and purge scoping are sound, the known is_creating() process-local limitation is documented, docs match the code, and the 22 new tests pass while the surrounding modules show no new failures.

**`src/aleph/vm/agent/storage_cli.py`** (line 226): Unlike the daemon's `_reclaimable_on` (reconciler.py:529), this marker scan does not filter implausible directory names. `iter_reclaimable()` yields a directory with a hand-made `.reclaimable` marker on an implausible name (e.g. `lost+found`), so `reclaim lost+found` reaches `purge_vm_storage` and `_checked_namespace` raises ValueError — a traceback instead of the clean refusal, still fail-closed but inconsistent with the daemon. Filtering with `_plausible(directory.name)` would make it a clean refusal.

**`src/aleph/vm/agent/storage_cli.py`** (line 301): `_load_registry()` rehydrates without creating tables or running migrations, so on a node where `aleph-vm` never ran (no agent DB yet), the `get_execution_records()` query fails and the operator gets a traceback instead of a usable command. The daemon's own startup only avoids this because it creates tables first; worth a guard or a documented precondition.

**`src/aleph/vm/agent/storage_cli.py`** (line 247): The "left directories behind" message names only a device-mapper hold, but `_still_on_disk` is also True after an OSError from rmtree (permissions, lost+found on a read-only mount). Mentioning the dm target as the common case is fine, but the wording could hint the operator to check the log (purge_vm_storage logs the actual cause) rather than assume dm.

foxpatch-aleph (approved):
The CLI faithfully mirrors the daemon's reconciler: verified against the actual implementations, all imports exist, the fail-closed rule and device-teardown parity with reconcile_now are correct, the purge path validates hashes and refuses live or dm-held directories, and the 22 new tests pass. The broader suite failures are environment noise in modules untouched by this PR.

**`src/aleph/vm/agent/storage_cli.py`** (line 182): The pool header row truncates `BUDGT` while the cache row at line 199 uses `BUDGET` — inconsistent truncation between the two tables.

**`src/aleph/vm/agent/storage_cli.py`** (line 277): `report.purged_orphans + report.evicted` prints budget-evicted entries under "purged" — wording is loose since eviction is a budget eviction, not a purge; consider "removed" or a separate line for evicted entries.

**`src/aleph/vm/agent/storage_cli.py`** (line 246): `purge_vm_storage` raises ValueError for an implausible hash, which here would crash the CLI with a traceback instead of a clean refusal — practically unreachable since markers only exist on agent-created dirs, but worth noting.

foxpatch-aleph (approved):
Well-built PR: the CLI faithfully reuses the daemon's reconciler and applies the same fail-closed live-set rule, with the degraded-dry-run and device-teardown semantics mirrored correctly. All 22 new tests pass, as do the existing reconciler/reclaimable suites, and the documentation matches the code precisely. The findings below are minor robustness and consistency nits, none of which are bugs or security issues: a missing plausibility guard on the reclaim path, no clean handling for a never-initialized agent DB, an exit-code overload, and a few doc/test nits.

**`src/aleph/vm/agent/storage_cli.py`** (line 246): _reclaim is the one purge path without the _plausible() guard the reconciler applies to every action: a directory carrying a hand-placed .reclaimable marker but named like a non-hash passes the marker check, then purge_vm_storage -> _checked_namespace raises ValueError and the CLI dies with a traceback instead of the clean refusal the other paths print. Validating vm_hash up front (or catching ValueError here) would make the error path consistent.

**`src/aleph/vm/agent/storage_cli.py`** (line 301): Unlike the agent's own main(), which creates the tables and runs alembic migrations before touching the registry, the CLI only sets up the engine. On a node where the agent has never started, rehydrate_registry -> get_execution_records raises OperationalError (no such table) and the operator gets a traceback. Since this tool is explicitly documented to run without the daemon, a clean error message for the uninitialized-DB case would be worth it.

**`src/aleph/vm/agent/storage_cli.py`** (line 71): DEGRADED_EXIT_CODE = 2 collides with argparse's usage-error exit code, so a wrapper script cannot distinguish a degraded dry run from a bad invocation by the exit status alone (though the stderr warning does differ). A distinct code such as 3 would be unambiguous.

**`src/aleph/vm/agent/storage_cli.py`** (line 48): Importing the reconciler's underscore-prefixed helpers is pragmatic, but the CLI is now a second stable consumer of _release_cache_devices, _still_on_disk and _teardown_orphan_devices. Promoting them to public names in the reconciler would make this cross-module coupling visible and stable.

**`src/aleph/vm/agent/storage_cli.py`** (line 88): The known limitation (a create in flight is invisible to this process) is documented for reconcile, but reclaim shares an analogous, smaller window: a create's adopt() can clear the marker after the CLI's iter_reclaimable() listing but before the purge. A sentence in the reclaim --help (and the docs section) would make the exposure symmetrical.

**`tests/supervisor/test_storage_cli.py`** (line 46): The _no_device_mapper fixture patches only reconciler_module.DEVICE_MAPPER_DIRECTORY; purge.py keeps its own module-level copy pointing at the real /dev/mapper, so _held_by_device_mapper still stats real device paths during tests. Harmless with random hashes in practice, but the fixture's stated guarantee is only half-enforced — patching purge.DEVICE_MAPPER_DIRECTORY too would make the hermeticity airtight.

foxpatch-aleph (approved):
Solid, carefully-documented operator CLI. The fail-closed live-set rule (registry ∪ supervisor list_vms, degraded dry-run with exit 2, teardown gated on a real supervisor answer) faithfully mirrors the daemon's _startup_refusal semantics, and the in-flight-create limitation is honestly documented everywhere it matters. All 22 new tests plus the related reconciler/reclaimable/retire suites pass, and I found no correctness or security bugs — only non-blocking robustness and wording nits.

**`src/aleph/vm/agent/storage_cli.py`** (line 246): An implausibly-named directory with a hand-made `.reclaimable` marker (e.g. `foo`) passes the marker check — `iter_reclaimable()` has no plausibility filter, unlike `reconciler._reclaimable_on` and `cache._markers()`, which exist precisely because a hand-made marker under a directory nobody named after a VM must not abort a pass — so `purge_vm_storage` raises an uncaught `ValueError` and the operator gets a traceback instead of a clean refusal. Nothing is deleted, so this is only UX/consistency, but filtering the marker scan through `_plausible` (or catching the ValueError) would match the rest of the codebase.

**`src/aleph/vm/agent/storage_cli.py`** (line 263): Minor wording: this warning also fires for an explicit `--dry-run` (which is arguably right — that preview is registry-only too), but the suggested remedy "pass --trust-registry to purge using the registry alone" is misleading in that combination, where --trust-registry changes nothing about the output. Consider gating the suggestion on `not dry_run` or rewording for the explicit-preview case.

**`src/aleph/vm/agent/storage_cli.py`** (line 212): `list` (and `reclaimable_bytes()` in `_status`) don't filter implausibly-named directories, so e.g. `lost+found` renders as a row with REASON `live`, and marked-but-implausible dirs count as RECLAIMABLE in status even though the reconciler will never reclaim them (`_reclaimable_on` filters them out). Cosmetic, but worth a `_plausible` filter if the output is meant to be script-parsed.

**`tests/supervisor/test_storage_cli.py`** (line 121): Nice test for the reclaim no-dial guarantee. A cheap mirror asserting that `status` and `list` never dial the supervisor would lock in the headline read-only guarantee from the docs and PR description, which is currently untested.

**`src/aleph/vm/agent/storage_cli.py`** (line 226): `marked` builds a list used only for truthiness; `any(directory.name == vm_hash for directory, _ in iter_reclaimable())` expresses the intent directly. Same nit for the unreachable `return 2` at the end of `run()` (argparse's `required=True` guarantees `command`) — could reuse `DEGRADED_EXIT_CODE` or assert.

foxpatch-aleph (approved):
Solid, carefully-reasoned CLI with faithful parity to the daemon's reconciler semantics and a consistently-applied fail-closed rule. All imports verified against the reconciler/reclaimable/purge modules; exit-code contract (degraded=2, explicit --dry-run=0, --trust-registry override) matches the docs exactly; the degraded preview is faithful to what --trust-registry would do (cache pass skipped in both via live_known=reachable); no path-traversal risk since vm_hash is only compared, never joined, and purge_vm_storage re-validates the hash pattern. All 22 tests pass (PR description says 17 — stale after the head commit added the teardown-parity tests), the reconciler/reclaimable suites (69 tests) still pass, and mypy is clean. The ruff ARG001 nits on unused `pools` fixture args match pre-existing patterns in test_reconciler.py, so they're consistent with the codebase.

**`src/aleph/vm/agent/storage_cli.py`** (line 212): `_list` iterates `iter_namespace_dirs()` unfiltered, so non-VM directories on a pool (e.g. `lost+found`, which `reconciler._plausible()` explicitly anticipates and skips) show up with reason "live". Harmless since `reclaim` requires a marker, but filtering with the reconciler's plausibility check (or printing reason "-") would avoid implying a VM owns them.

**`src/aleph/vm/agent/storage_cli.py`** (line 228): Minor inconsistency: `_reclaim` writes all refusals to stdout (the `out` handle), while `_reconcile` deliberately routes its warning to stderr so the stdout report stays parseable. For scripting, refusal messages would sit better on stderr too — worth aligning before operators write parsers around the current stream.

**`docs/architecture/storage.md`** (line 566): The known-limitation paragraph (a create the daemon is mid-flight is invisible cross-process, since `is_creating()` is in-process state) is spelled out for `reconcile`, but `reclaim` has the same exposure: a `list_vms()` answer can go stale between the check and `purge_vm_storage()`, and a create that adopted the marker right after the CLI read it is equally invisible. One sentence extending the limitation to `reclaim` (or noting it in `reclaim --help`, as done for `reconcile`) would make the contract airtight.

foxpatch-aleph (approved):
Well-executed operator CLI that reuses the reconciler's exact semantics, including the fail-closed live-set rule, device-teardown ordering, and the degraded-run exit-code matrix, all faithfully mirroring reconcile_now/reconcile_at_startup. The 22 new tests pass (as do the 80 reconciler/reclaimable tests), the in-flight-create limitation is honestly documented in --help and the architecture doc, and I found no correctness or security issues. Four minor nits below, none blocking.

**`src/aleph/vm/agent/storage_cli.py`** (line 71): DEGRADED_EXIT_CODE = 2 collides with argparse's own usage-error exit code (and the unreachable `return 2` fallback in run()): a wrapper script cannot distinguish a degraded reconcile from malformed arguments. Both mean "nothing purged", so this is cosmetic, but 3 (or 1) would avoid the overload.

**`src/aleph/vm/agent/storage_cli.py`** (line 201): status's cache USED column sums cache_entries only, while the budget pass counts used + in_flight_bytes(root) (see cache._root_usage). A node with active downloads can show a cache under budget in `status` while the reconcile pass evicts from it. Maybe include in_flight_bytes here for parity.

**`src/aleph/vm/agent/storage_cli.py`** (line 216): The reconciler filters implausible names (_plausible) everywhere it walks namespace dirs, but `list` doesn't, so on a real pool `lost+found` is reported with reason "live". Consider skipping non-hash names or labeling them differently so "live" keeps meaning "owned by a live VM".

**`src/aleph/vm/agent/storage_cli.py`** (line 226): Unlike _reclaimable_on, this marker pre-check doesn't filter implausible directory names, so a hand-placed marker under e.g. `lost+found` passes it and then dies with a ValueError traceback from purge_vm_storage's _checked_namespace instead of a clean refusal. Fails safe, but a _plausible filter here would keep the failure mode tidy.

foxpatch-aleph (approved):
A well-built operator CLI that faithfully mirrors the daemon's reconciler, including its fail-closed live-set rule (registry union supervisor, dry-run downgrade with exit 2, refusal without --trust-registry) and both device-teardown steps. The security posture is sound: every hash reaching a delete path is validated against the item-hash pattern, the marker check is local and cheap, and no path can traverse. Tests are thorough and genuinely exercise the three supervisor-reachability states. The minor findings are all non-blocking: the CLI's orphan-device teardown is not guarded by VOLUME_CREATE_GUARD (so the documented limitation is slightly understated), a hand-made marker under an implausibly named directory makes reclaim die with a ValueError traceback instead of a clean refusal, and _list prints non-VM directories like lost+found as 'live'.

**`src/aleph/vm/agent/storage_cli.py`** (line 273): _teardown_orphan_devices runs with no create guard here: is_creating() is always False in this process, and unlike the directory purge in _is_orphan there is no mtime/VOLUME_CREATE_GUARD check on devices. So a CLI reconcile concurrent with a daemon create tears down that create's dm targets even within the guard window, while the docstring frames the hazard as a create that has 'outlived VOLUME_CREATE_GUARD'. Impact is bounded (backing files survive, the purge then refuses the young directory), but consider a sentence in the module docstring / reconcile --help noting that the device teardown, unlike the purge, has no guard at all.

**`src/aleph/vm/agent/storage_cli.py`** (line 246): purge_vm_storage raises ValueError for an implausible namespace, and iter_reclaimable() (unlike the reconciler's _reclaimable_on, whose comment explicitly guards hand-made markers) does not filter implausible names. A hand-made .reclaimable marker under an implausibly named directory therefore passes the marked check and crashes here with a traceback instead of producing a clean refusal. Filtering the marked list with reconciler._plausible would match the daemon's behaviour.

**`src/aleph/vm/agent/storage_cli.py`** (line 218): _list prints every directory under a pool, so non-VM entries (lost+found on an ext4 pool, an operator's stray dir) show up with reason "live". The reconciler filters these with _plausible(); worth doing the same here, or at least being aware the output can contain names that are not VM hashes.

**`src/aleph/vm/agent/storage_cli.py`** (line 229): Refusal messages in _reclaim go to out (stdout), while _reconcile's degradation warning correctly goes to stderr. For operators scripting around exit codes, refusals on stderr would be more conventional; the current split is presumably for testability, but it is slightly inconsistent within the same CLI.

foxpatch-aleph (approved):
Solid, carefully documented operator CLI that faithfully mirrors the daemon's reconciler, including the fail-closed live-set rule, the device-teardown passes and the degraded exit-code semantics; the test suite is thorough and passes. Only cosmetic nits: _list does not filter non-hash directories the way the reconciler does (_plausible), a malformed VOLUME_RETENTION_BUDGET crashes _status with a traceback, _reclaim's refusals go to stdout while _reconcile's warning deliberately goes to stderr, and the documented is_creating() limitation applies equally to reclaim's marker-check-to-purge window.

**`src/aleph/vm/agent/storage_cli.py`** (line 212): _list iterates iter_namespace_dirs() without the _plausible() filter the reconciler applies (_reconcile_namespaces), so pool strays like lost+found are printed as rows with REASON=live. Consider filtering directory.name through the same item-hash pattern for consistency.

**`src/aleph/vm/agent/storage_cli.py`** (line 194): parse_budget raises ValueError on a malformed VOLUME_RETENTION_BUDGET, so a bad setting turns `storage status` into a traceback rather than a clean error. A try/except printing the message and returning non-zero would be friendlier.

**`src/aleph/vm/agent/storage_cli.py`** (line 228): _reclaim writes its refusals to stdout, while _reconcile deliberately routes its warning to stderr to keep stdout parseable. For scripting, refusal messages on stderr (matching the reconcile warning) would be more conventional.

**`src/aleph/vm/agent/storage_cli.py`** (line 246): The documented known limitation (is_creating() is process-local) applies to reclaim too: the daemon can adopt a retained marker for a re-create between the CLI's marker check and purge_vm_storage, so the CLI can delete a directory the daemon just adopted. The window is narrower than reconcile's, but worth a sentence next to the is_creating paragraph in docs/architecture/storage.md.

**`src/aleph/vm/agent/storage_cli.py`** (line 49): Importing the private helpers _release_cache_devices, _still_on_disk and _teardown_orphan_devices across the module boundary works but couples the CLI to reconciler internals; thin public aliases in reconciler.py would make that contract explicit. Non-blocking since the intent (mirror the daemon exactly) is documented.

foxpatch-aleph (approved):
Solid, well-documented operator CLI with real daemon parity: the live set (registry unioned with the supervisor's list_vms), the teardown-before/release-after device ordering, the live_known propagation into the cache pass, and the purge-refusal re-check all mirror reconcile_now exactly, and the fail-closed degradation (dry run + stderr warning + exit 2, unless --trust-registry) is a sensible design for a separate process. The 22 new tests are meaningful (real temp pools, all three supervisor reachability states, exit codes, both device-teardown seams) and pass, as do the adjacent reconciler suites. Only minor observations below; none block.

**`src/aleph/vm/agent/storage_cli.py`** (line 71): Nit: DEGRADED_EXIT_CODE = 2 collides with argparse's own usage-error exit code (e.g. 'aleph-vm storage --bogus' also exits 2, as does a missing subcommand). A script cannot distinguish "degraded reconcile" from "bad invocation". Not blocking, but a distinct code (3?) or a note in the docs would help automation.

**`src/aleph/vm/agent/storage_cli.py`** (line 194): parse_budget can raise ValueError on a malformed VOLUME_RETENTION_BUDGET, and the daemon would catch that in settings.check() — which the CLI deliberately never calls, since check() asserts /dev/kvm, firecracker, etc. Similarly, _load_registry will traceback (OperationalError) on a node whose agent DB was never created, since main() skips create_tables/migrations. Consider a top-level try/except in main() that prints a one-line error and exits non-zero; raw tracebacks are a rough first experience for an operator tool. Same unguarded ValueError applies to cache_budget_bytes in _status (its parse_budget is wrapped only for OSError).

**`src/aleph/vm/agent/storage_cli.py`** (line 216): Every unmarked directory is labeled reason "live", but an unmarked orphan (a crashed create older than VOLUME_CREATE_GUARD that a pass hasn't marked yet, i.e. anything reconciled under reap that never gets a marker) also shows as "live". Cosmetic, but "live" could mislead an operator into believing a VM owns it; something like "unmarked" would be more honest. The workflow stays coherent because reclaim then correctly refuses it for lack of a marker.

**`src/aleph/vm/agent/storage_cli.py`** (line 257): Observation: _reconcile implements only the "supervisor could not be asked" half of reconciler._startup_refusal; it does not reproduce the startup pass's second refusal ("registry empty while the supervisor runs N VMs" -> lost DB, refuse to purge anything). With the supervisor reachable the union genuinely protects the listed VMs, and the registry-only path requires the explicit --trust-registry flag, so this is defensible — but the module docstring's claim of "the same fail-closed rule _startup_refusal applies" slightly overstates the parity. Worth one sentence acknowledging that only the reachability half is enforced.

**`src/aleph/vm/agent/storage_cli.py`** (line 48): Minor style: importing the private helpers _release_cache_devices, _still_on_disk and _teardown_orphan_devices across modules is a smell, though here they are deliberate parity seams with the daemon's passes. If the reconciler is comfortable with these being part of its CLI surface, consider promoting them (or re-exporting under public names) in a follow-up so the contract isn't underscore-incidental.

**`src/aleph/vm/agent/storage_cli.py`** (line 134): _open_supervisor duplicates agent.supervisor.build_supervisor(settings). Duplicating it is understandable — importing agent.supervisor drags in the whole aiohttp webapp — but a brief comment saying why the ~2-line factory isn't reused would prevent a future "deduplicate this" refactor from accidentally importing the webapp module into the CLI.

**`src/aleph/vm/agent/storage_cli.py`** (line 249): Tiny nit: a directory left behind after the purge is always reported as "a device-mapper target still holds its volumes", but purge_vm_storage also refuses to rmtree on a plain permission OSError. purge_vm_storage logs the actual cause, so the message is mostly right in practice; wording it as "still held (e.g. by a device-mapper target)" would cover the rare other case.

**`src/aleph/vm/agent/cli.py`** (line 158): The hand-rolled sys.argv[1] check before argparse works (the agent parser has no positionals, so there's no collision) but reads a bit magic. A one-line comment explaining that "storage" is dispatched before the agent's parser because the storage commands are a separate argparse tree that must run without the agent daemon would help, matching the note already present in pyproject.toml.

foxpatch-aleph (approved):
This PR adds a well-engineered operator CLI that reuses the reconciler's primitives and mirrors the daemon's reconcile_now pass faithfully (live-set union with the supervisor, teardown-before-walk gated on reachability, live_known propagation, cache-device release after the walk). The fail-closed semantics are coherent and each branch (unreachable -> dry-run + stderr warning + exit 2; reclaim refuses without --trust-registry; status/list never dial) has a test, the marker-first ordering in reclaim is cheap-local-first and sensible, and every purge path goes through the existing _checked_namespace hash validation, so there is no traversal or injection surface. I verified all imports/signatures against the underlying modules, executed the pure helpers (_human/_age), and confirmed the tests follow the conventions of test_reconciler.py/test_agent_backups.py. The findings below are minor robustness/consistency nits rather than bugs; none blocks merging.

**`src/aleph/vm/agent/storage_cli.py`** (line 310): The agent's main() calls settings.check() right after settings.setup() (agent/cli.py:228-232), but the storage CLI stops at setup(). check() is where VOLUME_RETENTION_BUDGET/CACHE_BUDGET are validated (conf.py:602-606); without it, a malformed budget surfaces as a raw ValueError traceback from parse_budget inside status/cache_budget_bytes instead of the agent's clean error. Relatedly, the agent runs alembic migrations before using the DB while _load_registry() assumes the tables exist, so on a node whose agent has never started, all four commands will crash on rehydrate. Consider calling settings.check() here for parity.

**`src/aleph/vm/agent/storage_cli.py`** (line 257): This mirrors reconcile_now exactly, which is worth noting: reconcile_now is the periodic/at-GONE pass, and only reconcile_at_startup applies _startup_refusal's second canary (supervisor answers but the registry rehydrated empty -> agent DB was lost -> refuse entirely). The CLI will purge through that state, relying on list_vms being complete. The union does protect every VM the supervisor names, so this is a defense-in-depth difference rather than a bug, but since the docs and the module docstring frame reconcile/reclaim as applying "the same fail-closed rule reconciler._startup_refusal applies", consider reusing _startup_refusal's registry-empty check here, or stating explicitly that the CLI matches the periodic pass rather than the startup one.

**`src/aleph/vm/agent/storage_cli.py`** (line 222): _reconcile was deliberately designed to keep stdout parseable and route its degraded-mode warning to stderr, but _reclaim writes all of its refusal messages to out (stdout). A script running "aleph-vm storage reclaim X 2>/dev/null" will still see the errors on stdout. If the stdout/stderr split was intentional for reconcile, reclaim should probably get the same treatment; if keeping all reclaim output together was the goal, a word in the docstring would help.

**`src/aleph/vm/agent/storage_cli.py`** (line 71): Minor: exit code 2 for a downgraded reconcile collides with argparse's own usage-error exit code, so a caller cannot distinguish "aleph-vm storage reconcle" (typo) from "supervisor unreachable, ran dry". A distinct code (e.g. 3) would make scripted use unambiguous.

**`src/aleph/vm/agent/storage_cli.py`** (line 210): _list doesn't apply the _plausible() filter the reconciler uses (reconciler.py:156), so a pool mountpoint's lost+found directory -- called out by name at reconciler.py:387 -- will be listed with reason "live" as if it were a VM. Either filter with _plausible here or print a different reason for non-hash directories.

**`tests/supervisor/test_storage_cli.py`** (line 104): Two small coverage gaps: no test exercises reclaim on a VM spanning two pools (iter_reclaimable matches by name across pools and purge_vm_storage purges every pool), and no test exercises status under VOLUME_RETENTION=reap (the BUDGET column prints "0 B"). Both paths go through machinery already covered by the reconciler tests, so this is a nice-to-have.

foxpatch-aleph (approved):
Well-executed operator CLI that faithfully reuses the daemon's reconciler machinery while applying the same fail-closed live-set rule, with parity on both device-teardown steps and the retention-budget math verified against the reconciler. The code is carefully commented, the documented in-flight-create limitation is honest and accurate, and all 22 tests pass. Only minor nits remain: the lost+found listing in `storage list`, an unhelpful hint in the dm-held refusal message, and the exit-code collision with argparse.

**`src/aleph/vm/agent/storage_cli.py`** (line 212): _list (and the live-bytes sum in _status) iterate every directory in the pools without the _plausible() item-hash filter the reconciler applies in _reconcile_namespaces, so on a real node lost+found and any stray directory show up with reason "live". Cosmetic only — reclaim refuses anything unmarked — but slightly misleading in an operator-facing listing; consider filtering with reconciler._plausible(directory.name).

**`src/aleph/vm/agent/storage_cli.py`** (line 249): The dm-held refusal tells the operator to "retry once it is torn down" but not how. Since storage reconcile is the only command here that tears namespace devices down, pointing at it in this message would save the operator a documentation trip.

**`src/aleph/vm/agent/storage_cli.py`** (line 71): DEGRADED_EXIT_CODE = 2 collides with argparse's own exit code for usage errors, so a script cannot distinguish a downgraded reconcile from a malformed invocation. Fine if nothing needs to distinguish them today, but worth a comment or a distinct code if it ever matters.

**`tests/supervisor/test_storage_cli.py`** (line 107): Nice seam for the marker-before-dial ordering. The module docstring also promises status and list never dial the supervisor — a similar test asserting _open_supervisor is untouched for those two commands would lock that guarantee in the way test_reclaim_checks_the_marker_before_dialing_the_supervisor does for reclaim.

foxpatch-aleph (commented):
Solid, careful work: the CLI reuses the reconciler's own code and mirrors reconcile_now's device-teardown passes exactly, the fail-closed rule is implemented sensibly, and the test suite is thorough (22 tests, all passing, plus the base reconciler/reclaimable/storage suites). All signatures used match the existing modules, and the lazy GrpcSupervisor construction plus the getattr-based close are handled correctly. Findings are non-blocking: (1) the CLI implements only the unreachable-supervisor half of _startup_refusal, not the empty-registry-while-supervisor-runs-VMs half, though this matches the daemon's periodic pass rather than its startup pass; (2) the documented in-flight-create race applies to reclaim too, not just reconcile, but only reconcile documents it; (3) reclaim refusal messages go to stdout while reconcile's warning goes to stderr, and DEGRADED_EXIT_CODE = 2 collides with argparse usage errors; (4) main() skips settings.check() and loads the DB unconditionally, so status/list traceback on a missing agent DB.

**`src/aleph/vm/agent/storage_cli.py`** (line 165): _cli_live_set implements only the first condition of reconciler._startup_refusal (supervisor unreachable) but not the second (empty registry while the supervisor runs VMs, i.e. a lost agent DB). In that state the CLI proceeds and purges every stopped VM's retained directories, where the daemon's startup pass refuses. This is fine parity with the daemon's periodic pass, but the module docstring and docs/architecture/storage.md say the CLI applies the same fail-closed rule _startup_refusal applies to the startup pass, which slightly overpromises. Either mirror the second refusal here or note the difference in the docs.

**`src/aleph/vm/agent/storage_cli.py`** (line 222): The in-flight-create race documented for reconcile (help text and module docstring) applies equally to reclaim: the marker check passes, then the daemon's creating() context clears the marker via adopt(), and the hash is not yet in the registry or list_vms, so purge_vm_storage deletes the in-flight create's disks. Only reconcile documents this; consider one sentence on reclaim --help / in the docs too.

**`src/aleph/vm/agent/storage_cli.py`** (line 228): reclaim writes its refusal messages to out (stdout), while _reconcile writes its degraded-run warning to stderr. Refusals are diagnostics, not part of the report, so stderr would be the more conventional stream. Also note DEGRADED_EXIT_CODE = 2 collides with argparse's usage-error exit code, so a script cannot distinguish a downgraded reconcile from a malformed command line — an explicit comment documenting that collision (or a distinct code like 3) would help.

**`src/aleph/vm/agent/storage_cli.py`** (line 308): main() never calls settings.check() (the daemon path does), so a malformed VOLUME_RETENTION_BUDGET surfaces as a raw parse_budget traceback instead of the settings error, and storage_pools.setup_pools() errors likewise. Separately, the registry is loaded unconditionally, so status and list — advertised as needing no running agent — still die with a traceback when the agent DB is missing or unmigrated, even though they only need an empty registry. Catching OperationalError (or degrading to an empty registry with a warning) would make the read-only commands useful on a broken node, which is where an operator reaches for them.

foxpatch-aleph (approved):
Solid, well-documented addition. The reconcile path mirrors reconcile_now exactly (teardown gating, live_known plumbing, cache-device release), the fail-closed supervisor rules and exit codes match the documented matrix and are pinned by tests, and reclaim is traversal-safe via the marker check plus purge_vm_storage's hash validation. All 22 new tests pass along with the related reclaim/reconciler/purge/cache suites, and ruff format is clean. Only nit-level findings remain: a misleading "live" label in `list`, an O(pools×dirs) re-walk in `status`, an exit-code collision with argparse's usage-error code 2, five test functions missing the ARG001 noqa that a sixth already carries, and a now-stale "no subcommands" comment in pyproject.toml.

**`src/aleph/vm/agent/storage_cli.py`** (line 216): Minor: `_list` labels every unmarked directory "live", but it never consults the registry or the supervisor, so an unmarked orphan (e.g. while the daemon is down and no pass has marked it yet) prints as `live`. Since `status`/`list` are deliberately registry-only, an "unmarked" or "-" label would be less likely to make an operator read an unowned directory as owned.

**`src/aleph/vm/agent/storage_cli.py`** (line 184): Perf nit: `iter_namespace_dirs()` is re-enumerated once per pool inside this comprehension. On a node with many namespace dirs this makes `status` O(pools × dirs) with repeated stat work. Hoisting the directory list out of the per-pool loop would make it a single walk.

**`src/aleph/vm/agent/storage_cli.py`** (line 71): Nit: DEGRADED_EXIT_CODE = 2 collides with argparse's own usage-error exit code 2, so a calling script cannot distinguish "bad arguments" from "reconcile was downgraded to a dry run". A code outside {0, 1, 2} would keep the two unambiguous.

**`tests/supervisor/test_storage_cli.py`** (line 281): The unused `pools` fixture argument here and in the tests at lines 293, 344, 357 and 368 trips `ruff check` ARG001 (ARG is selected in pyproject, and line 121 already carries `# noqa: F811, ARG001` for the same pattern). CI only runs `ruff format --diff` so nothing breaks there, but anyone running `ruff check` locally gets five failures; add the same noqa (or keep the `pools` argument used) for consistency.

**`pyproject.toml`** (line 83): Follow-up nit: the hatch script comment "The aleph-vm console script ... is a flat argparse CLI, no subcommands" is now stale, since this PR adds the `storage` subcommand dispatched in agent/cli.py:main.

foxpatch-aleph (approved):
A careful, well-tested implementation. The fail-closed live-set rule (registry union supervisor list_vms) is correctly transplanted to a separate CLI process, the two device-teardown steps faithfully mirror reconcile_now/reconcile_at_startup including their dry-run and unreachable-supervisor skip conditions, exit-code semantics (explicit --dry-run = 0, silent downgrade = 2, reclaim refusal = 1) are coherent and covered by tests, and the docs match the code exactly, including an honest known-limitation section. All 22 new tests pass and existing reconciler/reclaimable suites are unaffected. Remaining findings are minor: the CLI's orphan-device teardown step lacks both the in-process creating() guard and the VOLUME_CREATE_GUARD mtime check the namespace purge has, so the documented in-flight-create window is wider for that one step (bounded impact: dm targets are detached but volume files survive); `list` doesn't filter implausible directory names like lost+found; reclaim's refusal messages go to stdout while reconcile's warnings go to stderr; the DB engine in _load_registry is never disposed; and no test asserts the documented 'status/list never dial the supervisor' guarantee. None of these block merging.

**`src/aleph/vm/agent/storage_cli.py`** (line 273): The CLI's `_teardown_orphan_devices` call has neither the daemon's in-process `is_creating()` guard nor the `VOLUME_CREATE_GUARD` mtime check that `_is_orphan` gives the namespace purge — so while the docs frame the in-flight-create limitation as beginning once a create has outlived VOLUME_CREATE_GUARD, the teardown step is exposed to a mid-flight create for its whole duration (dm devices exist before the VM lands in list_vms, and `orphan_device_namespaces` checks only liveness, not age). Impact is bounded — teardown detaches dm targets but leaves the volume files — but the window is wider than documented. Consider filtering teardown candidates by namespace-directory mtime under VOLUME_CREATE_GUARD, or extending the docs/help text to say the device teardown is unguarded too.

**`src/aleph/vm/agent/storage_cli.py`** (line 212): `_list` prints every subdirectory of every pool, including implausible names like `lost+found`, which then show up with reason "live". The reconciler filters these with `_plausible()` (reconciler._reconcile_namespaces); reusing the same filter here would keep `list` output to actual VM directories.

**`src/aleph/vm/agent/storage_cli.py`** (line 229): Small stream inconsistency: `_reconcile` deliberately keeps the degraded warning on stderr so the stdout report stays parseable, but `_reclaim`'s refusal/diagnostic messages (no marker, live hash, supervisor unreachable) all go to `out` (stdout) even though they accompany a non-zero exit. Errors arguably belong on stderr.

**`src/aleph/vm/agent/storage_cli.py`** (line 301): `_load_registry` creates an engine via `metrics.setup_engine()` (which also installs the global `AsyncSessionMaker`) but never disposes it. Harmless for a process that exits immediately after, but disposing the engine after rehydration would avoid GC noise on exit.

**`tests/supervisor/test_storage_cli.py`** (line 64): There's no test asserting the documented guarantee that `status` and `list` never dial the supervisor. The `dialed`-spy pattern already used in `test_reclaim_checks_the_marker_before_dialing_the_supervisor` would make this a cheap two-line addition for both commands.

**`src/aleph/vm/agent/storage_cli.py`** (line 127): Minor: `_age` renders a marker with a future `reclaimable_since` (clock skew between writer and this host) as a negative age (e.g. "-0d -1h"). Clamping at zero would read better.

foxpatch-aleph (approved):
The CLI is a faithful, fail-closed port of the daemon's reclamation logic: the union live set (registry ∪ list_vms), the marker-first reclaim refusal, the device teardown steps before/after the pass, and the dry-run/exit-code semantics all match reconcile_now's guarantees, and the known in-flight-create limitation is honestly documented in the help text and docs. All 22 tests pass, mypy is clean, and there are no security issues (the marker check requires a real on-disk directory and purge_vm_storage re-validates the hash pattern, so there is no traversal or injection surface). Non-blocking observations: (1) the docs claim 'the same fail-closed rule _startup_refusal applies', but only the supervisor-unreachable clause is mirrored -- the daemon's second refusal (empty registry while the supervisor runs VMs, i.e. a lost DB, refuse to purge anything at all) has no CLI counterpart, so a CLI reconcile will purge unrelated orphans in that state; the union still protects supervisor-listed VMs, so this is a parity/documentation imprecision rather than a safety hole; (2) 'list' labels unmarked directories as reason 'live', which is misleading for unmarked orphans; (3) the reconcile output prints purged/evicted/marked names but not cache_evicted entries; (4) no test asserts that status/list never dial the supervisor although the docs promise it; (5) DEGRADED_EXIT_CODE=2 collides with argparse's usage-error exit code.

**`src/aleph/vm/agent/storage_cli.py`** (line 165): The docstring (and docs/architecture/storage.md:577) says this applies 'the same fail-closed reasoning reconciler._startup_refusal applies', but only the first clause of _startup_refusal is mirrored. The daemon also refuses to purge anything when the supervisor answers with VMs while the registry is empty (a lost agent DB); here that state still purges unrelated orphans. The union set does protect every supervisor-listed VM, so this is safe, but the parity claim slightly overstates it -- worth a sentence in the docs or docstring, or an explicit refusal for parity.

**`src/aleph/vm/agent/storage_cli.py`** (line 216): An unmarked directory that no live VM owns (an orphan a pass has not marked yet) is printed with reason "live". Since surfacing those is one of list's purposes, a neutral label such as "unmarked" or "-" for the no-marker case would be more accurate than "live".

**`src/aleph/vm/agent/storage_cli.py`** (line 277): The reconcile output names purged orphans and evictions but not report.cache_evicted entries, which only appear in the summary counts. Printing them (they are Paths) would let an operator match the report against the cache directories that vanished, the same way the purged names allow.

**`tests/supervisor/test_storage_cli.py`** (line 64): There is a test that reclaim refuses before dialing the supervisor, but no test that status and list never dial it, even though the docs promise 'status and list never dial the supervisor'. A small test monkeypatching _open_supervisor to raise (or record) and running both commands would pin that property.

**`src/aleph/vm/agent/storage_cli.py`** (line 71): Minor: DEGRADED_EXIT_CODE = 2 collides with argparse's usage-error exit code, so a caller distinguishing 'degraded preview' from 'bad arguments' must parse stderr. A distinct code (3) or documenting the collision would help scripting.

foxpatch-aleph (approved):
Solid, faithful port of the daemon's reclamation logic into an operator CLI. The fail-closed design (registry union supervisor live set, degraded reconcile exiting 2, reclaim refusing without --trust-registry) matches the documented semantics exactly, and the device-teardown parity with reconcile_now/reconcile_at_startup is correct in every reachability/flag combination I traced. All imports, function signatures, and test seams resolve against the base branch, and the new module compiles. Only minor, non-blocking observations remain.

**`src/aleph/vm/agent/storage_cli.py`** (line 228): Minor stream-discipline nit: reclaim's refusal reasons are written to stdout (via out), while the degraded reconcile warning deliberately goes to stderr. A shell script running `aleph-vm storage reclaim <hash> > /dev/null` will miss the refusal reason. Consider sending these refusals to stderr, since they are error outcomes and the reconcile command already established that convention.

**`src/aleph/vm/agent/storage_cli.py`** (line 71): DEGRADED_EXIT_CODE = 2 collides with argparse's own exit code 2 for usage errors. `storage reconcile --unknown-flag` and a degraded reconcile are distinguishable only by text. Harmless in practice, but a different value (e.g. 3) would make the degraded case unambiguous for scripts.

**`src/aleph/vm/agent/storage_cli.py`** (line 257): Informational: this pass cannot be serialized against the daemon's own passes — `_pass_lock` in reconciler.py is process-local to the agent's event loop, so a CLI reconcile while the daemon is up can interleave with a periodic/GONE pass. The removals tolerate the races (vanished directories are skipped), so this is the same family as the documented is_creating() limitation, but the docstring frames that limitation only around in-flight creates. One sentence in the module docstring or docs/architecture/storage.md mentioning pass-overlap would make the "daemon's own pass preferred" advice fully motivated.

**`tests/supervisor/test_storage_cli.py`** (line 310): These two device tests patch remove_parent_device but not sweep_leaked_cache_loops, so the real sweep runs (unlike the daemon's equivalent at test_reconciler.py:866, which stubs it). It's harmless on a test host — the sweep reads /sys and only detaches loops backed by deleted files inside the configured cache roots — but stubbing it like the reconciler tests do would keep these hermetic and consistent.

**`tests/supervisor/test_storage_cli.py`** (line 201): The report printing in _reconcile has a dedicated branch for report.marked_orphans ("would mark"/"marked") that no test exercises — everything here runs under VOLUME_RETENTION=reap. A small test under keep asserting the marked lines would cover that output path.

foxpatch-aleph (approved):
Solid, carefully reasoned operator CLI. The fail-closed design faithfully mirrors the daemon's own reconciler passes (supervisor union into the live set, live_known propagation, teardown-before/release-after parity with reconcile_now), the is_creating() cross-process blind spot is honestly documented in the module docstring, --help, and docs, and the 3s supervisor dial cap is well-motivated given the 30s gRPC deadline. All 22 new tests pass and the related reconciler/reclaimable/retire suites stay green. I found no correctness or security issues; the findings below are minor polish items.

**`src/aleph/vm/agent/storage_cli.py`** (line 184): Minor: iter_namespace_dirs() is re-walked once per pool inside the live_bytes sum, so a node with N directories costs O(pools x dirs) stat walks per status run. Iterating once and grouping by directory.parent would avoid the repeated pool scans.

**`src/aleph/vm/agent/storage_cli.py`** (line 216): Minor: _list labels every directory iter_namespace_dirs() yields, including non-hash entries like lost+found, as reason "live". Filtering with the reconciler's _plausible() (or an "other" label) would avoid implying a non-VM directory belongs to a live VM.

**`src/aleph/vm/agent/storage_cli.py`** (line 249): The dm-held refusal says "retry once it is torn down" but nothing tells the operator how to tear it down; storage reconcile is the only CLI path that runs the orphan-device teardown. Consider pointing at it in the message.

**`src/aleph/vm/agent/storage_cli.py`** (line 48): Style: importing the underscore-private _release_cache_devices, _still_on_disk and _teardown_orphan_devices couples this CLI to reconciler internals; if the CLI is a permanent consumer, promoting these to public names would keep the contract explicit (mypy/refactor safety).

**`src/aleph/vm/agent/storage_cli.py`** (line 312): main() calls settings.setup() but never creates the DB tables or runs migrations the agent path does (agent/cli.py:239-242), so on a node where the agent has never started, rehydrate_registry() will surface as a raw SQLAlchemy traceback rather than a friendly error. Consider catching it or documenting the prerequisite.

**`src/aleph/vm/agent/storage_cli.py`** (line 286): Nit: DEGRADED_EXIT_CODE (2) collides with argparse's usage-error exit code, and _reclaim writes its refusal messages to stdout while _reconcile's warning goes to stderr. Both defensible (the reconcile report must stay parseable on stdout), just worth being aware that scripts can't distinguish a usage error from a degraded reconcile by code alone.

foxpatch-aleph (approved):
A well-engineered operator CLI that reuses the reconciler's own machinery (reconcile_storage, supervisor_hashes, the two device-teardown steps) rather than reimplementing it, with fail-closed semantics that faithfully mirror the daemon's: registry-union-supervisor live sets, a degraded dry run with exit 2 for reconcile, refusal for reclaim, and --trust-registry as an explicit opt-in that deliberately buys a purge but not a device teardown. All 22 new tests plus the existing reclamation suites pass, and the docs are accurate. No correctness or security issues found; the comments below are doc-accuracy and labeling nits.

**`src/aleph/vm/agent/storage_cli.py`** (line 272): Doc nit, not a code bug: the known-limitation text (module docstring line 22, reconcile --help, and docs/architecture/storage.md) says only a create that has "outlived VOLUME_CREATE_GUARD" is at risk. That is true of the directory purge (_is_orphan's mtime guard), but not of this teardown step: orphan_device_namespaces() has no age guard, only the process-local is_creating(), so a CLI reconcile run while the daemon is mid-create tears down that create's dm devices at any age, even well inside the guard. The 'prefer the daemon's pass' guidance still covers it, but the limitation statement would be more accurate if it said the guard protects the directories, not the devices.

**`src/aleph/vm/agent/storage_cli.py`** (line 216): Minor: an unmarked directory is labeled REASON=live, but an unmarked orphan (older than the guard, not yet marked by a pass) also prints as live. Since this command is exactly what an operator uses for manual triage, "live" for something merely unmarked is misleading; consider "-" or "unmarked" unless the directory's hash is actually in the registry live set (status/list never dial the supervisor, so the honest label for unknown would be "unmarked").

**`src/aleph/vm/agent/storage_cli.py`** (line 71): Nit: exit code 2 is also argparse's usage-error code, so a wrapper script cannot distinguish "degraded reconcile" from "bad arguments" without parsing stderr. Harmless for a human operator; worth noting in --help if scripts are expected to key off it.

**`src/aleph/vm/agent/storage_cli.py`** (line 274): Observation: a CLI reconcile shares no lock with the daemon's _pass_lock, so it can run concurrently with a daemon pass (periodic, at-GONE). The failure modes are benign — purge_vm_storage is idempotent and tolerant of a vanished directory, and markers are written exclusively — but the docs only state the creating() gap, not the concurrent-pass race. A sentence in the storage.md CLI section would round it out.

foxpatch-aleph (approved):
Well-engineered PR: the fail-closed live-set rule is faithfully ported from the daemon's reconciler to a standalone CLI, the new VOLUME_CREATE_GUARD check in orphan_device_namespaces correctly extends the directory-purge guard to device teardown for out-of-process passes, and the 22 new tests (plus the reconciler test) all pass. Fail-closed behavior, degraded-mode exit code 3, dry-run inertness, and device-teardown parity with reconcile_now were each verified against the implementation and are well documented. Only minor nits follow.

**`src/aleph/vm/agent/storage_cli.py`** (line 265): Minor robustness gap: `iter_reclaimable()` (unlike the reconciler's `_reclaimable_on`) does not filter implausibly-named directories, but `purge_vm_storage` -> `_checked_namespace` raises ValueError for a name not matching `^[0-9a-zA-Z]{16,128}$`. A hand-made `.reclaimable` marker under an implausibly-named directory (e.g. `backup_old`) makes `reclaim backup_old` pass the marker check and then crash with a traceback instead of exiting 1. Reconcile filters such names via `_plausible`; reclaim should too (or catch the ValueError) so a clean refusal is printed.

**`tests/supervisor/test_reconciler.py`** (line 930): This test mixes two clocks: `_age(young, 10_000)` sets the mtime relative to the fixed NOW (2026-08-24), but `reconcile_now`/`orphan_device_namespaces` use `datetime.now()`. It only passes while the real clock is more than VOLUME_CREATE_GUARD past NOW; every other test in this file threads `now=NOW` for determinism. Aging relative to `time.time()` (or passing a `now` seam) would make it robust to when the suite runs.

**`src/aleph/vm/agent/storage_cli.py`** (line 267): Small UX nit: the refusal says "retry once it is torn down", but the CLI offers no way to tear down a single namespace's devices — the operator has to run a full `storage reconcile` to do it. Worth a word in the message or the docs (the reclaim path never calls `_teardown_orphan_devices`).

**`src/aleph/vm/agent/storage_cli.py`** (line 291): Nit on test coverage: the skip of `_teardown_orphan_devices` in the *degraded* case (supervisor unreachable, no --trust-registry, effective_dry_run=True) is not asserted anywhere — `test_reconcile_leaves_orphan_devices_when_the_supervisor_is_unreachable` only covers the --trust-registry variant. A one-line assertion there or in the degraded test would pin this branch too.

**`src/aleph/vm/agent/storage_cli.py`** (line 55): The CLI reaches into the reconciler's private API (`_release_cache_devices`, `_teardown_orphan_devices`, `_still_on_disk`). Pragmatic and it guarantees the CLI runs byte-for-byte the daemon's teardown logic, but it couples the two modules silently; a rename of any of these will break the CLI with no compiler help. Fine to keep as-is given the reasoning, just flagging the maintenance coupling.

foxpatch-aleph (approved):
Solid, well-documented operator CLI that mirrors reconcile_now faithfully (teardown gating, live_known plumbing, cache-device release), with a correct and well-tested age-guard extension to orphan_device_namespaces. The fail-closed rule is consistently applied: status/list never dial the supervisor, reconcile downgrades to a dry run with exit 3, reclaim refuses without --trust-registry, and every purge path is behind _plausible/_checked_namespace. Tests cover all three supervisor reachability states, exit codes, and both device-teardown seams. Only minor findings remain: reclaim's marker-then-dial window is a few seconds wide rather than the daemon's few instructions, status/list have small efficiency and cosmetic issues, and main() lacks the DB/settings bootstrap the agent path performs. None are blocking.

**`src/aleph/vm/agent/storage_cli.py`** (line 254): The marker is read here, but the purge in _reclaim happens after the supervisor dial (up to SUPERVISOR_CONNECT_TIMEOUT_SECS later). A daemon create entering creating() in that window adopt()s the directory (clearing the marker) and the CLI still purges what the create is building on. The daemon documents its own residual window as "a few instructions" (reconciler.py:639-646); here it is seconds. Re-reading the marker after the dial (or before purge_vm_storage) would narrow it back to the daemon's window, and applying the VOLUME_CREATE_GUARD age check to the directory would mirror the protection reconcile's pass gives.

**`src/aleph/vm/agent/storage_cli.py`** (line 195): Minor: iter_namespace_dirs() is re-walked (with a directory_size_bytes stat pass) once per pool. On a node with many VMs and several pools this is O(pools^2) stats where one walk grouped by directory.parent would do.

**`src/aleph/vm/agent/storage_cli.py`** (line 229): Cosmetic: unlike the daemon's walk (_reconcile_namespaces skips implausible names with a debug log), list will show non-VM directories such as lost+found on a pool mountpoint with reason "unmarked". reclaim refuses them via _plausible, so it is only a display oddity.

**`src/aleph/vm/agent/storage_cli.py`** (line 327): This return 2 is unreachable: the subparsers group is required=True, so argparse exits before run() sees an unknown command.

**`src/aleph/vm/agent/storage_cli.py`** (line 341): main() calls settings.setup() but neither settings.check() nor the DB bootstrap the agent path performs (metrics.create_tables + alembic, agent/cli.py:239-242). On a node with no agent DB yet, rehydrate_registry dies with a raw OperationalError traceback, and a malformed VOLUME_RETENTION_BUDGET raises mid-status. A fail-closed crash is safe, but a clean message would suit a command that is explicitly meant to run without the daemon.

PR #1163, opened 2026-08-26
fix(nix): content-only initrd so the launch measurement tracks file content, not store paths

odesenfans:
## What

- `nix/initrd.nix` no longer uses `pkgs.makeInitrd`. The cpio now holds regular files at their final paths (busybox, init scripts, attest-agent, veritysetup/nft/cryptsetup, the dm/nft modules), asserts at build time that no `nix/store` entry exists and that every executable is static, and uses `pkgsStatic.busybox`.
- `nix/boot-smoke.sh`: boots `nix#image` under plain QEMU (no SEV) and asserts init reaches the verity mount, the firewall and the `/sbin/init` handoff.
- Golden file re-seeded (the initrd content changed).

## Why

`makeInitrd` copies each object's nix closure into the archive and symlinks `/bin/aleph-attest-agent` to `nix/store/<hash>-cargo-package-0.1.0/...`. The initrd bytes therefore embedded the agent's derivation hash, which covers the whole cargo workspace (`cleanCargoSource ../rust`): any Rust change anywhere moved all four launch measurements with a byte-identical agent binary. That is what made the golden check on #1140 fail after every rebase (diagnosis in the PR discussion). The closure also dragged glibc, headers and man pages into the guest; the previous busybox was dynamically linked and only ran because glibc rode along.

## Testing

- `gzip -dc initrd | cpio -t` lists exactly the 19 expected entries, no `nix/` prefix.
- `nix build ./nix#initrd --rebuild` is byte-identical.
- `nix/boot-smoke.sh` passes on KVM; with a wrong roothash it fails on the verity check as expected.
- `nix/check-golden-measurements.sh` passes against the re-seeded file.

Stacked on #1140. Follow-ups: own cargo workspace for the agent (next PR), source rev alongside the image.


foxpatch-aleph (approved):
This PR replaces pkgs.makeInitrd with a custom content-only cpio builder that eliminates nix store paths from the initrd, making the SEV-SNP launch measurement a function of file content rather than derivation hashes. The implementation is correct: the cpio recipe is reproducible (fixed mtime, sorted entries, root ownership, gzip -9n), the build-time static-link check covers all executables, and the contract check prevents store paths from re-entering the archive. The new boot-smoke.sh test validates the full boot chain under plain QEMU, and its marker strings correctly match init.sh output. The pkgsStatic.busybox switch is necessary and correct since the old dynamically-linked busybox would have no glibc to run against. The golden measurements are re-seeded and the existing CI workflow covers verification. No correctness, security, or quality issues found.

**`nix/boot-smoke.sh`** (line 9): Consider wiring boot-smoke.sh into a CI workflow (e.g., golden-measurements.yml or a separate one). Currently it's a manual-only test, so regressions in the initrd boot path could land undetected between golden-measurement CI runs.

PR #1164, opened 2026-08-26
refactor(rust,nix): aleph-attest-agent gets its own workspace, lockfile and narrowed nix source

odesenfans:
## What

- `rust/crates/aleph-attest-agent` is its own cargo workspace with its own `Cargo.lock` (seeded from the root lock: identical versions, verified). The root workspace `exclude`s it (its root `Cargo.lock` is pruned of the agent-only tree: no package added, no version changed; some dependency references lose their version qualifier once the duplicate older major leaves the graph); it still depends on `aleph-tee` by path.
- `test-rust.yml` runs fmt/clippy/test on the agent workspace too; `golden-measurements.yml` triggers on `rust/Cargo.toml`, `rust/crates/aleph-tee/**`, `rust/crates/aleph-attest-agent/**` (no longer on `rust/Cargo.lock`).
- `nix/flake.nix` builds the agent from a `lib.fileset` of exactly those paths, vendoring from the agent's lock (crane `cargoLock`/`cargoToml` + `postUnpack` into the nested workspace).

## Why

The agent is the measured guest binary. Sharing the supervisor's lockfile meant a dependency bump made for the daemon changed the guest TCB and the launch measurement as a side effect. Now a bump of the agent's dependencies is a deliberate change in its own lockfile (re-seed the golden file in the same PR), and nothing else under `rust/` can reach the agent derivation: verified by appending to `supervisor-daemon/src/lib.rs` (agent .drv unchanged) and to `aleph-tee/src/lib.rs` (agent .drv changes).

## Golden file

Re-seeded; initrd diff limited to bin/aleph-attest-agent.

## Testing

- `cargo fmt/clippy/test --locked` in both workspaces.
- `nix build ./nix#attest-agent`, `nix/boot-smoke.sh`, `nix/check-golden-measurements.sh`.

Stacked on the content-only initrd PR. Dependency bumps for the agent now happen in `rust/crates/aleph-attest-agent/Cargo.lock`.


foxpatch-aleph (approved):
This PR correctly separates aleph-attest-agent into its own Cargo workspace with an independent Cargo.lock, ensuring that supervisor-side dependency bumps can no longer change the measured guest TCB as a side effect. The nix flake's `lib.fileset` correctly narrows the agent derivation's source to exactly its measured inputs (root Cargo.toml, aleph-tee, and the agent crate), and the `postUnpack`/`cargoToml`/`cargoLock` crane configuration properly handles the nested workspace layout. The CI workflows correctly add fmt/clippy/test gates for the agent workspace and update the golden-measurements trigger paths from the now-irrelevant root `Cargo.lock` to `rust/Cargo.toml` and `rust/crates/aleph-tee/**`. The root and agent lockfiles are each self-consistent with matching dependency versions. The divergences.md documentation is accurately updated. The content-only initrd design (verified in initrd.nix) confirms the measurement isolation claim. No security or correctness issues found.

foxpatch-aleph (approved):
Correct and well-executed refactoring that isolates aleph-attest-agent into its own Cargo workspace with an independent Cargo.lock, narrowing the nix source to exactly the measured inputs. This achieves the stated goal: supervisor dependency bumps no longer change the guest launch measurement, and agent dependency changes are deliberate and reviewable. All verification checks passed: workspace inheritance works correctly (aleph-tee's .workspace=true resolves from the root Cargo.toml which is in the fileset), version pins are consistent between manifests, root lockfile pruning is correct (agent-only deps removed, version qualifiers dropped where duplicates eliminated), agent lockfile contains all transitive deps with matching versions, nix fileset narrowing covers exactly the measured inputs, CI workflows trigger on the right paths, and golden measurements are re-seeded. No bugs, security issues, or significant code quality issues found.

The `[workspace]` declaration without a `[workspace.dependencies]` section is correct but unusual. The comment block (lines 9-18) does a good job explaining why: aleph-tee inherits from the root workspace, and the root Cargo.toml is shipped in the nix fileset. Consider adding a brief note at the nix fileset site in flake.nix noting that the root Cargo.toml must stay in agentSrc for this inheritance to work, as a guard against future edits that might accidentally drop it.

The agentSrc fileset correctly unions exactly the root Cargo.toml, aleph-tee sources, and agent sources. This is the key mechanism that ensures non-measured rust/ changes cannot affect the agent derivation. The postUnpack cd into the nested workspace is correct for crane's build expectations.

Lockfile pruning is correct: agent-only packages (actix-web ecosystem, brotli, zstd, cookie, h2 0.3.x, http 0.2.x, etc.) removed, and version qualifiers dropped where duplicate major versions were eliminated (http, h2, digest, const-oid). All remaining entries have versions consistent with the agent's Cargo.lock.

Trigger path changes are correct: removing rust/Cargo.lock and adding rust/Cargo.toml + rust/crates/aleph-tee/** + rust/crates/aleph-attest-agent/** ensures the golden measurement check runs on exactly the changes that could affect the agent's measured dependency tree.

PR #1165, opened 2026-08-26
fix(agent): 404, not 500, for a /vm/<hash> the network has no message for

odesenfans:
## Summary
- \`storage.get_message\` raises \`FileNotFoundError\` when the API has no message under the requested hash (never existed, or forgotten), but \`try_get_message\` only mapped the older \`ClientResponseError(404)\` to an HTTP 404. The \`FileNotFoundError\` fell through to the unhandled-exception middleware: a 500 plus a full traceback in the journal on every hit.
- Seen on the mainnet SNP test node: something polls \`GET /vm/5a3081a9…/ip/6\` for a forgotten hash every few minutes, each one logging a 15-line traceback.
- Now a \`404 {"error": "Hash not found: <hash>"}\`, same as the connector 404 path.

## Test plan
- [x] new \`test_run_code_unknown_hash_is_404\` (full webapp, \`_fetch_aleph_message\` mocked to \`None\`)
- [x] \`tests/supervisor/views/test_run_code.py\` + \`test_view_errors.py\` pass locally; ruff format / isort clean
- [ ] CI

https://claude.ai/code/session_01LwF8MmMGqDPX9Y16VwPvUj

### DIFF
--- a/src/aleph/vm/agent/messages.py
+++ b/src/aleph/vm/agent/messages.py
@@ -18,6 +18,12 @@ async def try_get_message(ref: str) -> ExecutableMessage:
     """Get the message or raise an aiohttp HTTP error"""
     try:
         return await get_message(ref)
+    except FileNotFoundError as error:
+        # The API answered but has no message under this hash: never
+        # existed, or forgotten since. A caller's bad reference, not a
+        # node failure, so a 404 rather than the 500 an unhandled error
+        # would produce.
+        raise HTTPNotFound(reason="Hash not found", text=f"Hash not found: {ref}") from error
     except ClientConnectorError as error:
         raise HTTPServiceUnavailable(reason="Aleph Connector unavailable") from error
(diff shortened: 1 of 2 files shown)


foxpatch-aleph (approved):
The fix correctly addresses a real bug: `get_message` raises `FileNotFoundError` when the Aleph API has no message for a given hash, but `try_get_message` only mapped `ClientResponseError(404)` and `ClientConnectorError` to HTTP errors. The uncaught `FileNotFoundError` fell through to the 500 handler, producing a traceback on every request for a non-existent or forgotten hash. The new handler maps it to `HTTPNotFound` with the same response format as the existing 404 path, and the test correctly mocks `_fetch_aleph_message` to return `None`, exercising the exact code path through the full webapp middleware stack. The change is minimal, well-commented, and consistent with the existing error-handling patterns in the function.

PR #1166, opened 2026-08-26
feat(nix): source revision alongside the image outputs and the golden check

odesenfans:
## What

- `nix#image`, `nix#composeImage`, `nix#instanceImage` write `$out/source-rev` (`self.rev`, or `self.dirtyRev` on a dirty tree). Not part of anything measured, so the launch measurements are unaffected (golden check passes unchanged).
- `nix/check-golden-measurements.sh` reports the commit that last seeded the golden file next to the current HEAD on divergence.

## Why

A measurement should always be traceable to the source it was built from. The vprogram bundle already records `source.rev`; this covers raw nix outputs and the golden file.

Stacked on the attest-agent workspace PR.


foxpatch-aleph (approved):
The PR correctly adds a source-rev audit trail file to image output directories without affecting any measured artifacts, and improves the golden check error message to identify the seeding commit. The Nix fallback chain for sourceRev is idiomatic, the heredoc change is necessary and safe (no $ characters in the body), and the git commands have proper fallbacks. No security or correctness issues found.

**`nix/check-golden-measurements.sh`** (line 68): The heredoc is now unquoted (`<<EOF` instead of `<<'EOF'`) to interpolate `${seeded_at}` and `${head_rev}`. The body text currently contains no `$` characters, so this is safe. Just be aware that any future edit adding a literal `$` to this heredoc body would be silently expanded by the shell — consider a brief inline comment noting the heredoc is intentionally unquoted for variable interpolation.

**`nix/flake.nix`** (line 381): No test verifies that `source-rev` is actually written to the image outputs. This is a low-risk gap since the `runCommand` derivations are straightforward, but a simple assertion (e.g., in the golden check or a smoke test) that `source-rev` exists and is non-empty would make the audit trail self-verifying.

foxpatch-aleph (approved):
Clean, well-documented change that adds source revision traceability to image outputs and improves the golden measurement divergence diagnostic. The `sourceRev` binding correctly uses the Nix `or` fallback chain for clean/dirty/non-git trees and is deliberately kept out of anything measured, so launch measurements are unaffected. The shell script correctly switches to an expanding heredoc and has a robust fallback chain for git failures. The `fetch-depth: 0` change is necessary and well-justified. No correctness, security, or quality concerns.

**`nix/check-golden-measurements.sh`** (line 66): The double-fallback `|| echo unknown` followed by `:-unknown` is belt-and-suspenders: `git log` returns non-zero on failure (triggering `|| echo unknown`), but could also return zero with empty output if the file was added in the initial commit with no history. The `:-unknown` catches that. Good defensive coding.

**`nix/flake.nix`** (line 164): Minor note: `self.dirtyRev` requires nix >= 2.19. On older nix, the attribute is absent on a dirty tree and the `or` chain falls through to `"unknown"`, which is the correct degradation. The comment at line 159 mentions this. No issue, just confirming the fallback is sound.

foxpatch-aleph (approved):
Well-scoped PR that adds source revision traceability to the three image output directories without affecting any measured inputs, preserving the golden check. The heredoc fix (quoted to unquoted) for variable expansion is correct, the fetch-depth: 0 change is necessary for the git log command to resolve, and the fallbacks on git commands are belt-and-suspenders. Documentation updates in divergences.md and confidential.md are accurate. No correctness, security, or testing concerns.

**`nix/flake.nix`** (line 287): Non-blocking: the `workload` convenience output (lines 287-292) also bundles artifacts into a directory but does not get a `source-rev` file, unlike `image`/`composeImage`/`instanceImage`. If the workload volume's source provenance is worth tracking, consider adding it here too; if not, the current scope is fine.

foxpatch-aleph (approved):
This PR adds source revision traceability to the three image output directories (image, composeImage, instanceImage) via a `$out/source-rev` file, and enhances the golden measurement divergence message to report the last commit that seeded the golden file alongside the current HEAD. The `sourceRev` expression (`self.rev or self.dirtyRev or "unknown"`) correctly uses Nix's attribute-missing fallback semantics and is deliberately kept out of measured artifacts (initrd, cmdline), so launch measurements are unaffected. The `fetch-depth: 0` CI change is justified by the need to resolve `git log` history for the golden file's last seeding commit. The heredoc change from `<<'EOF'` to `<<EOF` is correct for the variable interpolation, and the values being interpolated are safe (git SHAs or "unknown"). Documentation updates to divergences.md and confidential.md are accurate and well-written. No correctness, security, or testing issues found.

**`nix/flake.nix`** (line 287): Minor: the `workload` output (the workload volume + verity sidecars staging directory) does not get a `source-rev` file, unlike the three image outputs. If traceability is wanted for all artifact directories, this could be added here too. Not blocking — `workload` is a staging dir for a single volume, not a full image.

**`nix/check-golden-measurements.sh`** (line 67): Nit: `seeded_at` already falls back to `unknown` via `|| echo unknown` on line 66, so the `${seeded_at:-unknown}` on line 67 is redundant for the git-failure case. It does cover the edge case where `git log` succeeds but returns an empty string, though that is unlikely given the golden file existence check at line 58. Harmless defensive coding.

PR #1167, opened 2026-08-27
aleph-vm 2.0

odesenfans:

## Changes

## How to test

Explain how to test your PR.

## Print screen / video

## Notes


foxpatch-aleph (changes requested):
This PR introduces a major architecture overhaul with SEV-SNP confidential VM support, a Rust TEE library, attestation agent/CLI, Nix-based image building, and a two-process agent/supervisor split. The cryptographic core (ARK pinning, domain separation, report data binding) is well-designed with strong test coverage. However, several security and correctness issues need to be addressed before merging: systemd services run as root with zero hardening, the supervisor-launcher uses an unqualified python3 PATH lookup while running as root, the proxy's reqwest client follows redirects (SSRF amplification), init.sh uses a bare wait that leaves a dead workload behind a live attested endpoint, and secret key names leak into HTTP error responses.

**`packaging/supervisor-launcher`** (line 16): SECURITY: The supervisor launcher defaults to bare `python3` which resolves via $PATH, while the controller launcher correctly uses `/usr/bin/python3`. Since the supervisor service runs as root (User=0), an attacker who can place a `python3` binary earlier in the PATH could get it executed as root. Change to `ALEPH_VM_SUPERVISOR_PYTHON:=/usr/bin/python3` to match the controller launcher.

**`packaging/aleph-vm/etc/systemd/system/aleph-vm-agent.service`** (line 7): SECURITY: Both the agent and supervisor services run as root (User=0, Group=0) with zero systemd hardening directives. The agent is a network-facing HTTP API server. At minimum, add NoNewPrivileges=true, ProtectKernelTunables=true, ProtectKernelModules=true, ProtectControlGroups=true, RestrictSUIDSGID=true, LockPersonality=true, and RestrictRealtime=true. The controller@.service already sets NoNewPrivileges and PrivateTmp, proving the team is aware of these. Consider running the agent as a non-root user since it only needs socket access to the supervisor.

**`rust/crates/aleph-attest-agent/src/main.rs`** (line 103): SECURITY: `reqwest::Client::new()` uses the default redirect policy which follows up to 10 redirects. If the upstream application returns a redirect to an internal address (e.g., http://169.254.169.254/), the proxy will follow it and return the response to the caller, creating an SSRF amplification vector. Use `reqwest::Client::builder().redirect(reqwest::redirect::Policy::none()).build()` to disable redirect following.

**`nix/init.sh`** (line 191): CORRECTNESS: The bare `wait` at the end of init.sh waits for all child processes. If the workload's /sbin/init dies but the attest-agent keeps running, the VM stays up with a live attested TLS endpoint proxying to a dead upstream. The compose flavor (init-compose.sh) correctly waits on the specific guest PID and powers off when it exits. init.sh should do the same: track the workload PID and poweroff when it exits, rather than waiting for all children including the attest-agent.

**`nix/init.sh`** (line 116): CORRECTNESS: When both ro and rw mounts fail, the script prints a warning but continues execution. It proceeds to prepare_chroot, setup_firewall, and start the attest-agent on a non-existent /mnt/root. The compose and instance flavors both poweroff on mount failure. This should also `exec /bin/busybox poweroff -f` when the writable fallback mount fails.

**`rust/crates/aleph-attest-agent/src/secrets.rs`** (line 210): SECURITY: Secret key names are leaked into HTTP error responses. Line 210 returns `format!("failed to write secret: {key}")` as a 500 error body, and line 330 returns the user-supplied key name in a 400 response. Secret key names can themselves be sensitive. Replace the 500 message with a generic `"failed to write secret"` (keep the key name in server-side logs only).

**`rust/crates/aleph-attest-agent/src/proxy.rs`** (line 80): SECURITY: The attestation error at line 80 returns `format!("attestation failed: {e:#}")` to the HTTP client, which could leak TEE backend internals, file paths, or system details. Similarly, line 161 returns `format!("upstream unreachable: {e:#}")` which could reveal internal network topology. Return generic messages to the client and log the detailed error server-side.

**`rust/crates/aleph-tee/src/sev_snp/verify.rs`** (line 195): SECURITY: The verify_cert_chain function does not check X.509 Basic Constraints (CA:TRUE/FALSE) or Key Usage extensions. Without these checks, if an attacker obtained a cert signed by ASK without CA:TRUE, they could sign additional sub-certificates. Add checks that ARK and ASK have CA:TRUE and keyCertSign key usage, and VCEK has CA:FALSE or no CA flag.

**`rust/crates/aleph-tee/src/sev_snp/verify.rs`** (line 91): SECURITY: The verify_sev_snp_report function does not check the report's `policy` field, which encodes whether debug mode is allowed, SMT, minimum VMPL, etc. A report generated with DEBUG enabled (allowing a debugger to inspect guest memory) would pass verification silently. Add a policy check or at minimum expose the policy field in the VerificationResult so callers can enforce it.

**`rust/crates/aleph-attest-cli/src/client.rs`** (line 357): SECURITY: No validation that the URL scheme is HTTPS before sending requests. For inject_secret in legacy mode (no owner key), an `http://` URL would transmit secrets in plaintext. For attested_request, an `http://` URL bypasses the entire attestation gate. Add an explicit check at the start of each public function: `if !url.starts_with("https://") { bail!("URL must use https://"); }`

**`packaging/aleph-vm/DEBIAN/postinst`** (line 8): SECURITY: `rm -fr /tmp/aleph` during upgrade is a TOCTOU risk. A local user could create a symlink at /tmp/aleph pointing to an arbitrary directory. Since postinst runs as root, this could delete arbitrary directories. Check that the path is not a symlink before removing, or use `rm -fr --` with a guard.

**`packaging/aleph-vm/etc/aleph-vm/supervisor.env`** (line 1): SECURITY: The supervisor.env file is world-readable (644). Operators add secrets at runtime (SENTRY_DSN, AUTHORIZED_ALLOCATION_SIGNERS, etc.). The file should be mode 640 with group ownership matching the service user to prevent unprivileged users from reading runtime secrets.

**`nix/init-common.sh`** (line 144): CORRECTNESS: The bind mounts for /proc, /sys, and /dev in prepare_chroot have no error checking. If any fails (target directory missing, mount syscall failure), the chrooted workload runs without proc/sys/dev, which could cause crashes or security issues. Add `|| fatal` checks for each bind mount.

**`nix/compose-rootfs.nix`** (line 191): SECURITY: The Podman policy.json uses `insecureAcceptAnything`, disabling all signature verification for container images. While only local `podman load` is used today, if a compose file ever references a remote image, it would be pulled and run without any signing check. Consider using a more restrictive policy that only allows local archives.

**`rust/crates/aleph-attest-agent/src/proxy.rs`** (line 64): LOW: The attestation endpoint accepts any hex-decodable nonce of any length with no maximum. An attacker could send an extremely long nonce causing excessive memory allocation. Add a maximum nonce length check (e.g., 128 bytes).

**`.github/workflows/codeql-analysis.yml`** (line 38): COVERAGE: CodeQL analysis only scans Python. The repository now contains a significant Rust codebase with security-critical attestation and crypto code. Add 'rust' to the language matrix to get automated security scanning of the Rust code.

**`packaging/aleph-vm/DEBIAN/postinst`** (line 4): LOW: The jailman user is created with `useradd jailman` (no flags), giving it a login shell and home directory. For a service account that should never log in, use `useradd --system --no-create-home --shell /usr/sbin/nologin jailman`.

**`packaging/aleph-vm/DEBIAN/postinst`** (line 57): CORRECTNESS: The socket wait loop (lines 57-60) has no error path if the socket never appears within 30 seconds. If the supervisor crashes at startup, the loop exits silently and the agent is restarted anyway, failing to connect. Add a check after the loop that the socket exists, with an error message if it does not.

foxpatch-aleph (changes requested):
This is a large, well-architected PR adding SEV-SNP confidential VM support with strong crypto fundamentals (ARK pinning, domain separation, VMPL gating, symlink-protected secret injection). However, several security issues warrant changes before merge: the VCEK curve is not validated as P-384, the init.sh dm-verity fallback silently downgrades to a writable mount, all systemd services run as root without hardening, supervisor.env is world-readable and shared with untrusted guest controllers, and postinst does rm -rf on /tmp paths without symlink checks. Proxy error responses leak internal details to external clients.

**`rust/crates/aleph-tee/src/sev_snp/verify.rs`** (line 426): The VCEK public key is extracted as an EC key but never validated to be on the P-384 (secp384r1) curve. AMD's VCEK is always P-384. A compromised or cache-poisoned KDS could supply a VCEK with a key on a weaker curve. Add a check: `ensure!(ec_key.group().curve_name() == Some(openssl::nid::Nid::SECP384R1), "VCEK key is not P-384");`

**`nix/init.sh`** (line 114): The fallback to a writable mount only warns and does not poweroff. This silently downgrades the security posture — a compromised host could modify rootfs contents, defeating the entire measured-boot guarantee. init-compose.sh:131-137 correctly makes this a fatal poweroff. Consider making this fatal as well, or at minimum logging at a higher severity level.

**`packaging/aleph-vm/etc/systemd/system/aleph-vm-controller@.service`** (line 6): The controller service runs QEMU with untrusted guest workloads but has no User= directive (defaults to root). A VM escape lands directly as root on the host. Consider running as a dedicated unprivileged user (e.g., the jailman user created in postinst but never used) with NoNewPrivileges=true, and add systemd hardening directives (ProtectSystem=, CapabilityBoundingSet=, SystemCallFilter=, etc.).

**`packaging/aleph-vm/DEBIAN/postinst`** (line 9): `rm -fr /tmp/aleph` will follow a symlink placed at that path by any user (since /tmp is world-writable), potentially deleting arbitrary directories. Use a symlink check (e.g., `if [ -L /tmp/aleph ]; then rm -f /tmp/aleph; else rm -rf /tmp/aleph; fi`) or `find /tmp/aleph -delete`.

**`rust/crates/aleph-attest-agent/src/proxy.rs`** (line 161): The error response `format!("upstream unreachable: {e:#}")` leaks the internal upstream URL (e.g., http://127.0.0.1:8080/path) to the external TLS client. Return a generic message to the client (e.g., "upstream unreachable") and log the details server-side only. The same applies to line 80 (`format!("attestation failed: {e:#}")`) which leaks the full attestation error chain.

**`rust/crates/aleph-tee/src/sev_snp/verify.rs`** (line 277): `Asn1Time::days_from_now(0)` rounds to midnight in the local timezone, not the current instant. This creates a window of up to ~24 hours where expired certificates are accepted. Consider using `Asn1Time::from_unix(now.timestamp())` for an exact time comparison.

**`rust/crates/aleph-tee/src/sev_snp/verify.rs`** (line 101): The report policy bits are not checked. Specifically, the debug bit (policy bit 3) should be rejected — a debug-enabled SEV-SNP guest has its register state inspectable by the hypervisor. The documentation tells callers to pin the measurement and bind a nonce, but should also mention checking the policy debug bit, or the verifier should reject it by default.

**`nix/udhcpc.script`** (line 59): The `for ns in $dns` loop is intentionally unquoted for word-splitting, but $dns is a DHCP-server-controlled variable. A malicious DHCP server can inject arbitrary lines into /etc/resolv.conf. Consider validating each $ns is a well-formed IP address before writing it to resolv.conf.

**`nix/init.sh`** (line 191): The bare `wait` waits for ALL children including the attest-agent. If the workload exits but the attest-agent keeps running, the VM stays alive serving an attested TLS endpoint with no workload behind it. init-compose.sh correctly waits on `guest_pid` and powers off. Consider applying the same pattern here.

**`packaging/aleph-vm/etc/aleph-vm/supervisor.env`** (line 1): supervisor.env is installed with 0644 permissions (world-readable) and is loaded by all three services including the per-VM controller running untrusted guest workloads. Operators will populate this with secrets. Set permissions to 0640 root:root, and consider splitting secrets into a separate env file not loaded by the controller service.

**`nix/compose-rootfs.nix`** (line 191): The container image policy is set to `insecureAcceptAnything`, disabling all signature verification. If the compose workload pulls images from a registry, they would be pulled and run unsigned. Document this limitation or add a signature verification policy for known registries.

**`packaging/aleph-vm/DEBIAN/prerm`** (line 4): prerm stops the agent and supervisor but not the per-VM controller@ instances. During a package upgrade, old controllers keep running with old code while the new supervisor starts. Consider stopping controller instances explicitly, or document why this is safe.

PR #1168, opened 2026-08-27
Post-2.0 review follow-ups: drop aleph-attest-cli, harden the attest-agent, fail-closed init

odesenfans:
Follow-ups from the review of #1167 (the 2.0 dev -> main merge). None of these were regressions against main; they are the items worth doing from the bot's list, plus one removal the review prompted.

## What changes

**Drop `aleph-attest-cli`.** It was the aleph-cvm donor's verifying client and nothing in this repo consumed it (no CI job, test, packaging or testnets reference; only a flake package output). The aleph-rs SDK's `attest` module is the real verifier and already covers what this crate did plus the guest-policy pin, TCB floor and platform-posture checks it never grew, so the reviewer's "policy field not checked" finding is addressed on the side that matters. Workspace member, flake package (+ the now-unused `workspaceSrc`), comments and architecture docs updated; `aleph-tee` is untouched.

**attest-agent hardening** (`rust/crates/aleph-attest-agent`):
- reqwest client uses `redirect::Policy::none()`: a reverse proxy relays 3xx, it must not chase them from inside the guest.
- 500/502 bodies from the attestation endpoint and the proxy are generic; the anyhow chain (backend, device path, upstream address) stays in the guest log. The secret-store 500 no longer echoes the key name.
- Attestation nonce capped at `MAX_NONCE_LEN = 128` bytes before hex decoding (the SDK sends 32), with unit tests.

**Init scripts fail closed** (`nix/init.sh`, `nix/init-instance.sh`): both ended in a bare `wait`, so a dead `/sbin/init` left the VM up with a live attested TLS endpoint proxying to nothing. They now wait on the guest pid and power off when it exits, as `init-compose.sh` already did. `init.sh` also powers off when both rootfs mounts fail or the rootfs has no `/sbin/init`, instead of warning and starting the agent over an empty chroot. This makes the "chrooted init must stay in the foreground" contract load-bearing; it is now documented in `confidential.md` and the LUKS rootfs example header.

**Golden measurements re-seeded**: the agent binary and the init scripts are measured initrd inputs. Regenerated with `check-golden-measurements.sh --update` and verified with the check mode against the committed tree. Anything pinning the previous measurements must migrate.

## Not done, deliberately
- Basic-constraints/key-usage checks on the AMD chain: the chain is a fixed pinned-ARK -> ASK -> VCEK triple, the check adds nothing.
- systemd hardening / non-root agent, `supervisor.env` mode, postinst `rm -fr /tmp/aleph`, `useradd jailman`: all identical on main before 2.0; worth a backlog entry, not this PR.
- Podman `insecureAcceptAnything`: images only come from the verity-verified workload disk via `podman load`.

## Verification
- `cargo test` (attest-agent workspace): 30 passed. `cargo fmt --check` + `cargo clippy --all-targets -D warnings` clean for the agent workspace and `aleph-tee`.
- `dash -n` on the three init scripts.
- `nix flake show` evaluates; `nix/check-golden-measurements.sh` matches.

https://claude.ai/code/session_01LHA73vjDFEsgw6GDMs9jvA


foxpatch-aleph (approved):
Well-constructed PR with correct attest-agent hardening (redirect policy, error sanitization, nonce cap), proper fail-closed init supervision in all three init scripts, and clean removal of aleph-attest-cli across workspace, flake, and docs. No correctness or security issues found. Two minor cosmetic nits: unused workspace dependency declarations (zeroize, rustls, actix-web) and a line-break in a build script comment.

PR #1169, opened 2026-08-27
nix: whitelist guest kernel config (golden CI 2h20m -> minutes)

odesenfans:
Stacked on #1168 (od/drop-attest-cli). Golden-measurement CI spent 2h13m of its 2h22m compiling the guest kernel (run 33083034607: `linux-6.18.44` started 14:41, `dm-modules` 16:54, everything after the kernel took 2.5 min). Cause: `structuredExtraConfig` over nixpkgs' distro config compiles ~7000 modules for a guest whose initrds ship seven, and any override loses the binary-cache hit.

## What changes

- `nix/kernel.nix` builds the same nixpkgs-pinned 6.18 LTS source through `linuxKernel.manualConfig` from a generated `.config`: `make allnoconfig KCONFIG_ALLCONFIG=nix/kernel-config.fragment`, then `olddefconfig`. Every fragment line is verified against the generated `.config` at build time, so a dependency Kconfig silently drops fails the build, not the boot.
- `nix/kernel-config.fragment` is a whitelist (`EXPERT=y`): virtio, EFI stub, SEV-SNP guest (TDX guest kept), ext4, the exact modular set the initrds and compose rootfs install (`dm-mod`, `dm-bufio`, `dm-verity`, `dm-crypt`, `nfnetlink`, `nf_tables`, `fuse`), cgroups/namespaces for the compose flavor, serial console, all CPU mitigations (allnoconfig drops them), and the hardening set kept and extended: 32-bit entry paths, kexec, io_uring, tracing, `/dev/mem`, LDT, iopl compiled out; kstack offset randomization, `RANDOM_KMALLOC_CACHES`, UMIP on.
- Options that were `=m` in the distro config but never shipped in any initrd (veth, bridge, tun, conntrack, overlayfs, AF_ALG) were absent at runtime already and stay out.
- Golden measurements re-seeded (all four move: the kernel bytes changed).
- Separate first commit: `nix/boot-smoke.sh` marker was stale since 7a636488 (init.sh prints `from /mnt/root`), every run reported FAILED after a good boot.

## Numbers

| | before | after |
|---|---|---|
| `.config` lines | 9955 | 848 |
| bzImage | 14.1 MB | 4.8 MB |
| modules output | 7267 files, 161 MB | 7 files, 428 KB |
| kernel build, 32 cores | (not measured) | 259 s |
| golden CI job | 2h22m | this PR's run will tell |

## Verification

- `nix build ./nix#image ./nix#composeImage ./nix#instanceImage ./nix#workload`: all flavors build, initrd static/no-store-path assertions and the `dm-crypt` modules.dep check pass.
- `nix/boot-smoke.sh` (QEMU/KVM, no SNP): dm-verity mount, nft firewall, attest-agent starts and fails only on the missing `/dev/sev-guest`.
- Diffed `=y` sets old vs new by hand; nothing the init scripts, cryptsetup, nft or podman rely on was dropped.
- Not verified here: a real SEV-SNP launch (no SNP host locally). Same `AMD_MEM_ENCRYPT`/`SEV_GUEST`/`EFI_STUB`/`TSM_REPORTS` set as before; the testnet TEE host is the place to confirm.

## Follow-ups (not in this PR)

- `.github/workflows/golden-measurements.yml` comments and `timeout-minutes: 300` describe a multi-hour job; adjust once this run gives the new number.
- The supervisor's graceful stop sends QMP `system_powerdown`; no guest flavor has an ACPI button handler (before or after this change), so it always escalates to `quit`. `ACPI_BUTTON=y` plus a tiny handler in init would make stops clean.


foxpatch-aleph (approved):
This PR replaces the nixpkgs distro kernel config (which compiled ~7000 modules for a guest shipping seven) with a whitelist configuration over `allnoconfig`, reducing bzImage from 14.1 MB to 4.8 MB and modules from 161 MB to 428 KB. The approach is sound: the fragment is verified line-by-line against the generated `.config` at build time, so any silently dropped Kconfig option fails the build rather than the boot. The kernel-config.fragment is well-organized with clear section headers and comments, covers all features the init scripts depend on (virtio, ext4, dm-verity/dm-crypt modules, nftables, cgroups/namespaces for compose, serial console, EFI stub, SEV-SNP guest), and extends the hardening set beyond the old config (adding RANDOMIZE_MEMORY, SLAB_FREELIST_*, RANDOM_KMALLOC_CACHES, UMIP, INIT_ON_ALLOC_DEFAULT_ON, INIT_STACK_ALL_ZERO, KERNEL_IBT, USER_SHADOW_STACK, LEGACY_VSYSCALL_NONE, RANDOMIZE_KSTACK_OFFSET, etc.). All CPU mitigations are explicitly listed since allnoconfig drops them — critical for a confidential guest with an untrusted hypervisor. The boot-smoke.sh marker fix is correct: the old marker was stale since init.sh changed to print `from $guest_root` (e.g. `/mnt/root`) instead of `from rootfs`. Golden measurements are re-seeded for all four flavors. Documentation in confidential.md and divergences.md is updated accurately. The `manualConfig` usage with `config = { CONFIG_MODULES = "y"; }` correctly avoids import-from-derivation. A few auto-selected dependencies (NFNETLINK, TSM_REPORTS, NET_CRC32C) are not in the fragment and thus not verified by checkPhase, but their absence would manifest as build-time failures (module .ko not found) or is covered by `select`/`depends on` resolution in olddefconfig. The compose rootfs's CNI bridge config references kernel features (CONFIG_BRIDGE, CONFIG_VETH) that were already absent at runtime in the old config — not a regression.

**`nix/kernel.nix`** (line 92): Consider adding a comment noting that `config = { CONFIG_MODULES = "y"; }` is a minimal hint for manualConfig and does NOT determine the actual kernel config (which comes from `configfile`). A future reader might mistake this for the real config and wonder why only CONFIG_MODULES is set.

**`nix/kernel-config.fragment`** (line 272): `CONFIG_NF_TABLES=m` depends on `CONFIG_NFNETLINK` (or selects it). NFNETLINK is not in the fragment, so it's not verified by checkPhase. If olddefconfig resolves it to `=y` (built-in) rather than `=m`, initrd.nix's extraction of `nfnetlink.ko` at line 84 would fail at build time — a safe failure mode, but adding `CONFIG_NFNETLINK=m` to the fragment would make the intent explicit and get the checkPhase coverage.

**`nix/kernel-config.fragment`** (line 280): The compose rootfs's CNI config (compose-rootfs.nix line 198-227) references a "bridge" plugin that creates Linux bridge interfaces, which requires `CONFIG_BRIDGE`. The PR notes this was already absent at runtime (was `=m`, never shipped). Not a regression, but worth noting that the compose flavor's CNI bridge networking may not work with this kernel config — it likely relies on slirp4netns instead. A comment in the fragment or compose-rootfs.nix documenting this would help future maintainers.

**`nix/boot-smoke.sh`** (line 44): The prefix match `"init: starting /sbin/init from "` is correct and works for all three init scripts (init.sh prints `from $guest_root`, init-compose.sh and init-instance.sh print `from rootfs`). Good fix.

foxpatch-aleph (commented):
Well-structured PR that replaces nixpkgs' distro kernel config with a minimal allnoconfig-based whitelist, cutting golden CI from 2h20m to ~20 minutes and shrinking the confidential guest's attack surface. The build-time verification of every fragment line against the generated .config is a strong safety mechanism. The boot-smoke.sh marker fix is correct (init.sh prints the mount point dynamically). The checkPhase logic is sound for both =y/=m entries (exact-line grep) and "is not set" entries (negation check). One observation: NFNETLINK is not in the fragment but is shipped as a module by initrd.nix and loaded by init.sh/init-compose.sh; it is implicitly pulled in by NF_TABLES' Kconfig dependency today, but adding it explicitly would make the seven-module set self-documenting and resilient to future Kconfig changes. The CNI bridge config in compose-rootfs.nix references kernel modules (BRIDGE, VETH, TUN) that are not in the fragment, but this matches the pre-change behavior (those modules were =m in the distro config but never shipped in any initrd/rootfs module set), so it is not a regression.

**`nix/kernel-config.fragment`** (line 272): NFNETLINK is not listed in the fragment, but initrd.nix:84 extracts nfnetlink.ko.xz from the kernel module output and both init.sh:58 and init-compose.sh:71 insmod it. NF_TABLES depends on NFNETLINK in Kconfig, so olddefconfig enables it as =m today. However, the checkPhase cannot verify this implicit dependency. Adding `CONFIG_NFNETLINK=m` to the fragment would make the seven-module set explicitly complete, self-documenting (the fragment comment on line 15 already lists nfnetlink as one of the seven), and resilient to future Kconfig dependency changes that could make NFNETLINK built-in (=y), which would cause initrd.nix to fail at build time trying to extract a nonexistent nfnetlink.ko.xz.

**`nix/initrd.nix`** (line 77): The comment says "NET_CRC32C=y here" but the new fragment (kernel-config.fragment) does not contain CONFIG_NET_CRC32C. The comment appears to reference the old distro config, not the new whitelist. Since kernels >= 6.15 removed libcrc32c as a separate module (the comment itself says this), the statement is slightly misleading in context of the new config. Not a bug, just a stale reference.

foxpatch-aleph (approved):
A well-executed whitelist kernel config that replaces nixpkgs' distro config overlay, reducing the golden CI from 2h20m to ~20min while shrinking the confidential guest's attack surface. The build-time checkPhase verification of every fragment line against the generated .config is a strong safety mechanism that catches silently-dropped Kconfig options at build time rather than boot time. The boot-smoke.sh marker fix is correct (init.sh prints the mount point path, not the literal "rootfs"). The kernel config is comprehensive: all SEV-SNP guest requirements, all modules the initrds ship, the full CPU mitigation set (which allnoconfig would otherwise drop), and an extended hardening set. The manualConfig usage avoids IFD by passing a minimal config attrset. The compose CNI bridge networking gap (CONFIG_BRIDGE absent, needed by the CNI bridge plugin in compose-rootfs.nix) is pre-existing and not a regression — the previous distro config had it as =m but no initrd ever shipped bridge.ko, so it was already non-functional at runtime. Golden measurements are properly re-seeded for all four flavors. The CI timeout reduction (300 -> 60 min) is well-justified with a reference run.

**`nix/kernel-config.fragment`** (line 272): CONFIG_NFNETLINK is not in the fragment but is auto-selected by olddefconfig as a dependency of CONFIG_NF_TABLES=m. This is correct behavior (the checkPhase only verifies explicit fragment lines, and auto-selected dependencies are handled by the kernel build system), but it's worth noting in a comment that NFNETLINK is expected to be auto-enabled, so a future reader doesn't think it was accidentally omitted.

**`nix/compose-rootfs.nix`** (line 204): The CNI bridge plugin config ("type": "bridge") requires CONFIG_BRIDGE in the kernel, which is not in the whitelist fragment. This is pre-existing (the distro config had CONFIG_BRIDGE=m but no initrd shipped bridge.ko, so it was already non-functional at runtime) and the compose flavor ships slirp4netns as an alternative. Consider adding a comment noting that CNI bridge mode requires a kernel module not present in the whitelist, so podman falls back to slirp4netns/rootless mode.

PR #1170, opened 2026-08-27
refactor(supervisor)!: remove the Python supervisor daemon

odesenfans:
## Why

The Rust daemon and controller have been the supervisor side that CI (`test-using-pytest.yml` rust leg, `test-rust.yml`) and the SEV-SNP testnet runs exercise, yet the Python daemon still shipped in the deb and was still the **launcher default**: `ALEPH_VM_SUPERVISOR_IMPL` fell back to `python` unless an operator flipped it, and the DigitalOcean testnet CRNs were running it (aleph-testnets#37 fixes that side). This PR makes the Rust implementation the only one.

## What goes

- `src/aleph/vm/supervisor/` (daemon, gRPC server, `LocalSupervisor`, Python controllers), `pool.py`, `models.py`, `hypervisors/`, `network/`, `systemd.py`, `aleph.vm.testing` (single-process harness) and the agent CLI flags that wired it (`--benchmark`, `--run-test-instance`, `--run-fake-instance`). The closure was computed from the import graph: nothing under `aleph.vm.agent` / `supervisor_interface` reaches any of it once the harness edge is cut.
- `packaging/supervisor-launcher`, `packaging/controller-launcher` and the `ALEPH_VM_SUPERVISOR_IMPL` flag: the units exec `/opt/aleph-vm/bin/aleph-vm-supervisor` and `aleph-vm-controller` directly. A stale `ALEPH_VM_SUPERVISOR_IMPL=` line in an existing `supervisor.env` is inert.
- Python-only deps: `qmp`, `nftables` bindings, `dbus-python`, `pyroute2`, `python-cpuid` (pyproject, Makefile pip line) and `python3-nftables`/`python3-dbus`/`python3-netifaces` from the deb Depends.
- 62 `tests/supervisor` modules that drove the Python daemon, `tests/network`, `tests/test_controller_launcher.py`, `tests/conformance/test_cloudinit_parity.py`, and the fixture generators `scripts/generate_rust_fixtures.py` / `generate_controller_argv_fixtures.py` (the committed Rust fixtures are frozen goldens now; comments updated).

## What stays and how

- `tests/conformance/oracle.py`: a frozen copy of the Python daemon's `PortMapping` schema and `spec_from_controller_configuration`, so the conformance suite keeps asserting the same DDL and the same GetVmSpec reconstruction. `host_ipv4` is now checked against `ip -4 addr` instead of `netifaces`.
[...]

PR #1171, opened 2026-08-27
feat(packaging)!: Rust supervisor daemon and controller by default

odesenfans:
## What

`ALEPH_VM_SUPERVISOR_IMPL` defaults to `rust` in `packaging/supervisor-launcher` and `packaging/controller-launcher`. An unknown value keeps its "log and run the default" semantics, so it lands on Rust as well; the Python daemon and controller only run when an operator sets `ALEPH_VM_SUPERVISOR_IMPL=python` explicitly, which stays the one-line rollback. The deb still ships both implementations; this is the 2.0 cutover without the code removal (that is #1170, held back).

Also: the integration conftest default follows the launcher (`rust`), the launcher dispatch tests pin the new table (`unset -> rust`, `bogus -> rust`), and the packaged `supervisor.env`, unit comments, Makefile note and `process-model.md` say so.

## Effect on CI

The droplet e2e jobs install the deb with a `supervisor.env` that does not set the flag, so from this PR on they exercise the Rust daemon and controller. The pytest integration matrix sets the flag explicitly and keeps both legs.

## Verification

- `tests/test_controller_launcher.py`: 5 passed with the new dispatch table (the tests exec the real script).
- `ruff format` clean; shellcheck runs in CI (`code-quality-shell`).
- On the testnets, aleph-testnets#37 runs the whole suite with `ALEPH_VM_SUPERVISOR_IMPL=rust` on every CRN: run 2 was 28/31 with the three failures being harness drift (CLI 0.17.0 name argument, upgrade test leaking into pr-tests), run 3 in progress.

### DIFF
--- a/docs/architecture/process-model.md
+++ b/docs/architecture/process-model.md
@@ -87,13 +87,14 @@ never uses it.
 Both a Python supervisor daemon (`src/aleph/vm/supervisor/daemon.py`) and a
 Rust one (`rust/crates/supervisor-daemon`) ship in the same `.deb`. Which one
 runs is an environment variable, `ALEPH_VM_SUPERVISOR_IMPL` (default
-`python`), read by a tiny dispatch script the systemd unit execs:
+`rust` since 2.0), read by a tiny dispatch script the systemd unit execs:
 `packaging/supervisor-launcher` for the daemon and
 `packaging/controller-launcher` for each per-VM controller. Both scripts read
 the same `/etc/aleph-vm/supervisor.env`, so one env change switches the
-daemon and every controller together; an operator typo falls back to
-`python` rather than failing to start. Cutover and rollback are therefore a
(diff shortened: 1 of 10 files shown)


foxpatch-aleph (approved):
Clean, well-executed cutover PR that flips the default supervisor implementation from python to rust across all relevant files. The dispatch logic, fallback behavior, tests, documentation, and CI matrix are all consistent. The change is low-risk: both implementations still ship, rollback is one env line, and CI sets the flag explicitly so neither leg is affected by the default. One minor stale doc reference found outside the diff.

**`docs/architecture/testing.md`** (line 58): This line still says "rather than silently defaulting to Python" but the default is now `rust`. The PR updated lines 197-200 in this same file but missed this earlier mention. Consider updating to "rather than silently picking one" to match the conftest's updated error message.

PR #1172, opened 2026-08-27
feat(agent): build a plan from a push and answer it immediately

odesenfans:
Task 4 of the async scheduler allocations stack (design: `docs/plans/2026-08-20-async-allocations-design.md`). Stacked on #1159.

`build_plan` verifies each entry and keeps the rejected ones out of the plan entirely, so nothing downstream can act on a message that failed verification. `compute_verdict` diffs the plan against what the supervisor reports and runs one cumulative admission over the additions, counting the removals as freed: a plan that drops B and adds C admits C against B's memory.

Both are pure and await-free. The handler must not yield between reading the supervisor's view and swapping the desired state, or the verdict it just returned could already be stale.

A VM pinned to another CRN via `requirements.node.node_hash` is refused with `node_mismatch` rather than admitted: the scheduler chooses which messages run here, and one naming a different node is not among them.

`AllocationState` is defined here but deliberately not mirrored from `VmStatus`. The supervisor owns what a VM is doing; these are the phases that precede it knowing the VM at all, and they stop existing once `create_vm` returns.

No endpoint calls this yet, so nothing changes at runtime.

## Tests

34 new in `tests/supervisor/test_allocation_verdict.py`, written before the code, three of them driving a real `CapacityManager` because a double cannot show what `simulate` does with a plan. Full `tests/supervisor` FAILED set identical to `dev-2.1` (44 environment-only failures, none in a file this PR touches); mypy at the `dev-2.1` error set exactly.

### DIFF
--- a/src/aleph/vm/agent/allocation/plan.py
+++ b/src/aleph/vm/agent/allocation/plan.py
@@ -7,26 +7,10 @@ down VMs that were migrated elsewhere during the downtime.
 
 from dataclasses import dataclass, field
 from datetime import datetime
-from enum import Enum
 
 from aleph_message.models import ItemHash
 
 from aleph.vm.agent.allocation.verify import VerifiedMessage
-
-
-class AllocationState(str, Enum):
-    """Agent-side phases, which all precede the supervisor knowing the VM.
-
-    Deliberately NOT a mirror of VmStatus: once create_vm returns, VmStatus is
-    the answer and this state stops existing. Two disjoint fields cannot drift,
-    whereas a merged enum would need maintaining in lockstep forever.
-    """
-
-    PLANNED = "planned"
-    RESOLVING = "resolving"
-    DOWNLOADING = "downloading"
-    SUBMITTING = "submitting"
-    FAILED = "failed"
 
 
 @dataclass
@@ -45,23 +29,6 @@ class AllocationPlan:
 
 
 @dataclass
-class FailureRecord:
-    """Why a planned VM is not running, and when we will try again.
-
-    Kept in the reconciler and NOT in AgentVmRegistry on purpose: the registry
-    is what CapacityManager sums committed resources over, and a VM that failed
-    to start must not count as committed.
-    """
-
-    code: str
-    message: str
-    attempts: int
-    first_failed_at: datetime
-    last_failed_at: datetime
-    next_retry_at: datetime
-
-
-@dataclass
 class PlanVerdict:
     """The immediate answer returned to the scheduler."""
 
@@ -69,5 +36,5 @@ class PlanVerdict:
     pending: list[ItemHash] = field(default_factory=list)
     unchanged: list[ItemHash] = field(default_factory=list)
     removing: list[ItemHash] = field(default_factory=list)
-    rejected: dict[ItemHash, dict] = field(default_factory=dict)
+    rejected: dict[str, dict] = field(default_factory=dict)
     retained: dict[ItemHash, str] = field(default_factory=dict)
--- a/src/aleph/vm/agent/allocation/plan.py
+++ b/src/aleph/vm/agent/allocation/plan.py
@@ -1,0 +1,73 @@
+"""What the scheduler asked for, and how far the agent got with it.
+
+The plan is held in memory only. After a restart the agent has no plan, and a
+reconciler with no plan deletes nothing: acting on a stale plan risks tearing
+down VMs that were migrated elsewhere during the downtime.
+"""
+
+from dataclasses import dataclass, field
+from datetime import datetime
+from enum import Enum
+
+from aleph_message.models import ItemHash
+
+from aleph.vm.agent.allocation.verify import VerifiedMessage
+
+
+class AllocationState(str, Enum):
+    """Agent-side phases, which all precede the supervisor knowing the VM.
+
+    Deliberately NOT a mirror of VmStatus: once create_vm returns, VmStatus is
+    the answer and this state stops existing. Two disjoint fields cannot drift,
+    whereas a merged enum would need maintaining in lockstep forever.
+    """
+
+    PLANNED = "planned"
+    RESOLVING = "resolving"
+    DOWNLOADING = "downloading"
+    SUBMITTING = "submitting"
+    FAILED = "failed"
+
+
+@dataclass
+class PlannedVm:
+    """One entry of the plan. State lives in the reconciler, not here: one owner."""
+
+    vm_hash: ItemHash
+    verified: VerifiedMessage | None = None
+
+
+@dataclass(frozen=True)
+class AllocationPlan:
+    plan_id: str
+    received_at: datetime
+    entries: dict[ItemHash, PlannedVm]
+
+
+@dataclass
+class FailureRecord:
+    """Why a planned VM is not running, and when we will try again.
+
+    Kept in the reconciler and NOT in AgentVmRegistry on purpose: the registry
+    is what CapacityManager sums committed resources over, and a VM that failed
+    to start must not count as committed.
+    """
+
+    code: str
+    message: str
+    attempts: int
+    first_failed_at: datetime
+    last_failed_at: datetime
+    next_retry_at: datetime
+
+
+@dataclass
+class PlanVerdict:
+    """The immediate answer returned to the scheduler."""
+
+    accepted: list[ItemHash] = field(default_factory=list)
+    pending: list[ItemHash] = field(default_factory=list)
+    unchanged: list[ItemHash] = field(default_factory=list)
+    removing: list[ItemHash] = field(default_factory=list)
+    rejected: dict[ItemHash, dict] = field(default_factory=dict)
+    retained: dict[ItemHash, str] = field(default_factory=dict)
--- a/src/aleph/vm/agent/allocation/verdict.py
+++ b/src/aleph/vm/agent/allocation/verdict.py
@@ -1,0 +1,133 @@
+"""Turning a request body into a plan, and a plan into an immediate answer.
+
+Everything here is pure and await-free by design: the handler must not yield to
+the event loop between reading the supervisor's view and swapping the desired
+state, or a concurrent push could invalidate the verdict it just returned.
+"""
+
+import logging
+from datetime import datetime
+from hashlib import sha256
+
+from aleph_message.models import (
+    ExecutableContent,
+    InstanceContent,
+    ItemHash,
+    VerifiableProgramContent,
+)
+
+from aleph.vm.agent.allocation.plan import AllocationPlan, PlannedVm, PlanVerdict
+from aleph.vm.agent.allocation.teardown import is_removable_by_allocation
+from aleph.vm.agent.allocation.verify import VerificationOutcome, verify_entry
+from aleph.vm.agent.capacity import requirements_from_message
+from aleph.vm.supervisor_interface.types import ConfidentialMode, VmInfo, VmStatus
+
+logger = logging.getLogger(__name__)
+
+# A VM the supervisor is already working on does not need re-creating.
+LIVE_STATUSES = (VmStatus.RUNNING, VmStatus.BOOTING, VmStatus.DEFINED)
+
+
+def compute_plan_id(hashes: list[str]) -> str:
+    """A stable identity for a plan, for correlation in logs and responses.
+
+    Order-independent, so the scheduler re-pushing the same set in a different
+    order is visibly the same plan.
+    """
+    return "sha256:" + sha256("\n".join(sorted(hashes)).encode()).hexdigest()
+
+
+def build_plan(body: dict, *, now: datetime) -> tuple[AllocationPlan, dict[ItemHash, dict]]:
+    """Verify every entry and assemble the plan.
+
+    Rejected entries are returned separately: they are answered in the response
+    and never enter the plan, so nothing downstream can act on them.
+    """
+    entries: dict[ItemHash, PlannedVm] = {}
+    rejected: dict[ItemHash, dict] = {}
+    for entry in body.get("vms", []):
+        vm_hash = ItemHash(entry["item_hash"])
+        outcome, verified, reason = verify_entry(entry)
+        if outcome is VerificationOutcome.REJECTED:
+            rejected[vm_hash] = {"code": "invalid_message", "message": reason}
+            continue
+        entries[vm_hash] = PlannedVm(vm_hash=vm_hash, verified=verified)
+    plan_id = compute_plan_id([str(h) for h in entries] + [str(h) for h in rejected])
+    return AllocationPlan(plan_id=plan_id, received_at=now, entries=entries), rejected
+
+
+def _retention_reason(record, info: VmInfo) -> str:
+    """Why an allocation push is not allowed to stop this VM."""
+    if record.uses_payment_stream:
+        return "payment_stream"
+    if record.uses_payment_credit:
+        return "payment_credit"
+    if info.gpus:
+        return "gpu"
+    if info.confidential_mode is not ConfidentialMode.NONE:
+        return "confidential"
+    return "operator_policy"
+
+
+def _targets_another_node(content: ExecutableContent, node_hash: str | None) -> bool:
+    """Whether the message pins itself to a CRN that is not us."""
+    requirements = getattr(content, "requirements", None)
+    node = getattr(requirements, "node", None) if requirements else None
+    required = getattr(node, "node_hash", None) if node else None
+    if not required:
+        return False
+    return node_hash is None or str(required) != str(node_hash)
+
+
+def compute_verdict(
+    plan: AllocationPlan,
+    *,
+    infos: list[VmInfo],
+    registry,
+    capacity,
+    node_hash: str | None = None,
+) -> PlanVerdict:
+    """The immediate answer: what we take, what we drop, what we refuse."""
+    verdict = PlanVerdict()
+    by_hash = {ItemHash(info.vm_id): info for info in infos}
+
+    for vm_hash, info in by_hash.items():
+        record = registry.get(vm_hash)
+        if vm_hash in plan.entries:
+            if info.status in LIVE_STATUSES or info.awaiting_confidential_init:
+                verdict.unchanged.append(vm_hash)
+            continue
+        if record is None or info.status is not VmStatus.RUNNING:
+            continue
+        if is_removable_by_allocation(record, info):
+            verdict.removing.append(vm_hash)
+        else:
+            verdict.retained[vm_hash] = _retention_reason(record, info)
+
+    candidates = []
+    for vm_hash, planned in plan.entries.items():
+        if vm_hash in verdict.unchanged:
+            continue
+        if planned.verified is None:
+            verdict.pending.append(vm_hash)
+            continue
+        content = planned.verified.message.content
+        if _targets_another_node(content, node_hash):
+            logger.info("Refusing %s: allocated to another node", vm_hash)
+            verdict.rejected[vm_hash] = {
+                "code": "node_mismatch",
+                "message": "this instance is allocated to a different node",
+            }
+            continue
+        # A v-program belongs in the instance memory bucket but is not an
+        # InstanceContent, the same distinction _admit makes in run.py.
+        is_instance = isinstance(content, (InstanceContent, VerifiableProgramContent))
+        candidates.append((vm_hash, requirements_from_message(content), is_instance))
+
+    for admission in capacity.simulate(candidates, releasing=frozenset(verdict.removing)):
+        if admission.accepted:
+            verdict.accepted.append(admission.vm_hash)
+        else:
+            verdict.rejected[admission.vm_hash] = {"code": admission.code, "message": admission.detail}
+
+    return verdict
--- a/src/aleph/vm/agent/allocation/verdict.py
+++ b/src/aleph/vm/agent/allocation/verdict.py
@@ -8,6 +8,7 @@ state, or a concurrent push could invalidate the verdict it just returned.
 import logging
 from datetime import datetime
 from hashlib import sha256
+from typing import Protocol
 
 from aleph_message.models import (
     ExecutableContent,
@@ -19,10 +20,33 @@ from aleph_message.models import (
 from aleph.vm.agent.allocation.plan import AllocationPlan, PlannedVm, PlanVerdict
 from aleph.vm.agent.allocation.teardown import is_removable_by_allocation
 from aleph.vm.agent.allocation.verify import VerificationOutcome, verify_entry
-from aleph.vm.agent.capacity import requirements_from_message
+from aleph.vm.agent.capacity import (
+    AdmissionVerdict,
+    ResourceRequirements,
+    requirements_from_message,
+)
+from aleph.vm.agent.vm_registry import AgentVmRecord
 from aleph.vm.supervisor_interface.types import ConfidentialMode, VmInfo, VmStatus
 
 logger = logging.getLogger(__name__)
+
+
+class _Registry(Protocol):
+    """The slice of AgentVmRegistry this module needs."""
+
+    def get(self, vm_hash: ItemHash) -> AgentVmRecord | None: ...
+
+
+class _Capacity(Protocol):
+    """The slice of CapacityManager this module needs."""
+
+    def simulate(
+        self,
+        candidates: list[tuple[ItemHash, ResourceRequirements, bool]],
+        *,
+        releasing: frozenset[ItemHash] = ...,
+    ) -> list[AdmissionVerdict]: ...
+
 
 # A VM the supervisor is already working on does not need re-creating.
 LIVE_STATUSES = (VmStatus.RUNNING, VmStatus.BOOTING, VmStatus.DEFINED)
@@ -37,16 +61,25 @@ def compute_plan_id(hashes: list[str]) -> str:
     return "sha256:" + sha256("\n".join(sorted(hashes)).encode()).hexdigest()
 
 
-def build_plan(body: dict, *, now: datetime) -> tuple[AllocationPlan, dict[ItemHash, dict]]:
+def build_plan(body: dict, *, now: datetime) -> tuple[AllocationPlan, dict[str, dict]]:
     """Verify every entry and assemble the plan.
 
     Rejected entries are returned separately: they are answered in the response
     and never enter the plan, so nothing downstream can act on them.
     """
     entries: dict[ItemHash, PlannedVm] = {}
-    rejected: dict[ItemHash, dict] = {}
+    rejected: dict[str, dict] = {}
     for entry in body.get("vms", []):
-        vm_hash = ItemHash(entry["item_hash"])
+        # This is the validation boundary for a body the scheduler controls, so
+        # a bad entry is data to reject, never an exception: one unusable hash
+        # must not take down the whole push.
+        raw_hash = entry.get("item_hash") if isinstance(entry, dict) else None
+        try:
+            vm_hash = ItemHash(str(raw_hash))
+        except Exception:
+            logger.warning("Refusing plan entry with an unusable item_hash: %r", raw_hash)
+            rejected[str(raw_hash)] = {"code": "invalid_message", "message": "unusable item_hash"}
+            continue
         outcome, verified, reason = verify_entry(entry)
         if outcome is VerificationOutcome.REJECTED:
             rejected[vm_hash] = {"code": "invalid_message", "message": reason}
@@ -56,8 +89,10 @@ def build_plan(body: dict, *, now: datetime) -> tuple[AllocationPlan, dict[ItemH
     return AllocationPlan(plan_id=plan_id, received_at=now, entries=entries), rejected
 
 
-def _retention_reason(record, info: VmInfo) -> str:
+def _retention_reason(record: AgentVmRecord, info: VmInfo) -> str:
     """Why an allocation push is not allowed to stop this VM."""
+    if not record.persistent:
+        return "non_persistent"
     if record.uses_payment_stream:
         return "payment_stream"
     if record.uses_payment_credit:
@@ -69,33 +104,40 @@ def _retention_reason(record, info: VmInfo) -> str:
     return "operator_policy"
 
 
-def _targets_another_node(content: ExecutableContent, node_hash: str | None) -> bool:
-    """Whether the message pins itself to a CRN that is not us."""
+def _required_node_hash(content: ExecutableContent) -> str | None:
+    """The CRN this message pins itself to, if it pins one."""
     requirements = getattr(content, "requirements", None)
     node = getattr(requirements, "node", None) if requirements else None
     required = getattr(node, "node_hash", None) if node else None
-    if not required:
-        return False
-    return node_hash is None or str(required) != str(node_hash)
+    return str(required) if required else None
 
 
 def compute_verdict(
     plan: AllocationPlan,
     *,
     infos: list[VmInfo],
-    registry,
-    capacity,
+    registry: _Registry,
+    capacity: _Capacity,
     node_hash: str | None = None,
 ) -> PlanVerdict:
     """The immediate answer: what we take, what we drop, what we refuse."""
     verdict = PlanVerdict()
     by_hash = {ItemHash(info.vm_id): info for info in infos}
+    unchanged: set[ItemHash] = set()
+    # Planned VMs the supervisor holds in a dead state are about to be created
+    # again. Their registry records still count as committed, so admission has
+    # to discount them or a recreate is judged against its own resources; the
+    # enforced path does the same thing with check_capacity's exclude_vm_hash.
+    recreating: set[ItemHash] = set()
 
     for vm_hash, info in by_hash.items():
         record = registry.get(vm_hash)
         if vm_hash in plan.entries:
             if info.status in LIVE_STATUSES or info.awaiting_confidential_init:
+                unchanged.add(vm_hash)
                 verdict.unchanged.append(vm_hash)
+            elif record is not None:
+                recreating.add(vm_hash)
             continue
         if record is None or info.status is not VmStatus.RUNNING:
             continue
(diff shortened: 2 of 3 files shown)


foxpatch-aleph (commented):
Well-structured pure functions with clear design documentation and good test coverage. The main concern is that `_retention_reason` doesn't account for non-persistent VMs, which pass through `is_removable_by_allocation` as non-removable but would get the misleading reason `"operator_policy"` instead of something like `"non_persistent"`. No test covers this case. The remaining observations are minor and non-blocking since no endpoint calls this code yet.

**`src/aleph/vm/agent/allocation/verdict.py`** (line 59): `_retention_reason` doesn't check `record.persistent`. `is_removable_by_allocation` returns `False` for non-persistent VMs (teardown.py:31), so they reach this function, but the reason returned would be `"operator_policy"` for a non-persistent VM with no payments/GPUs/confidential — which is misleading. Consider adding an early `if not record.persistent: return "non_persistent"` check, and a test for this case.

**`src/aleph/vm/agent/allocation/verdict.py`** (line 49): `entry["item_hash"]` will raise `KeyError` on malformed input missing this field. Since `build_plan` receives a raw request body, consider `entry.get("item_hash")` with a rejection for missing keys, so the response is a 400 with a clear reason rather than a 500. Not urgent since no endpoint calls this yet.

**`src/aleph/vm/agent/allocation/verdict.py`** (line 109): Membership check against `verdict.unchanged` (a list) is O(n). For large plans, converting `unchanged` to a set or using a separate `set` for the membership test would be more efficient. Minor, since plan sizes are expected to be small.

**`src/aleph/vm/agent/allocation/plan.py`** (line 40): `AllocationPlan` is `frozen=True` but `entries` is a mutable `dict[ItemHash, PlannedVm]`. The freeze only prevents field reassignment, not dict mutation. This is likely intentional (the reconciler updates `PlannedVm.verified` post-fetch), but a brief comment noting that the entries dict is deliberately mutable would help future readers.

foxpatch-aleph (commented):
Well-structured, well-tested code with clear documentation and sound logic. The compute_verdict function correctly handles the full lifecycle of plan entries (unchanged, removing, retained, pending, accepted, rejected) and the admission simulation correctly accounts for freed resources. Two minor observations: (1) build_plan uses entry["item_hash"] which will raise KeyError on malformed input, inconsistent with verify_entry's defensive entry.get("item_hash") — a malformed entry crashes the handler instead of being cleanly rejected; (2) AllocationPlan is frozen=True but holds a mutable dict, which is a common Python footgun though unlikely to cause issues given the single-owner design. Test coverage is good but lacks a test for build_plan's rejection path and the awaiting_confidential_init branch.

**`src/aleph/vm/agent/allocation/verdict.py`** (line 49): entry["item_hash"] will raise KeyError if the key is missing, while verify_entry (called on the same entry at line 50) uses entry.get("item_hash") defensively. Consider using entry.get("item_hash") here as well, or validating the key before accessing it, so a malformed entry is cleanly rejected rather than crashing the handler.

**`src/aleph/vm/agent/allocation/plan.py`** (line 40): frozen=True on a dataclass with a mutable dict field (entries) gives a false sense of immutability — the dict can still be mutated. This is unlikely to cause issues given the single-owner design, but if true immutability is desired, consider MappingProxyType or a frozen dict wrapper.

**`tests/supervisor/test_allocation_verdict.py`** (line 179): Missing test coverage for: (1) build_plan with a rejected entry — verify it ends up in the rejected dict and not in entries; (2) the awaiting_confidential_init=True path in compute_verdict, where a VM awaiting confidential init should be classified as unchanged rather than removing.

foxpatch-aleph (commented):
Well-structured pure functions with correct core logic and good design rationale. The main gaps are in test coverage: the build_plan rejection path, several _retention_reason branches (credit, gpu, confidential, operator_policy), non-running VmStatus handling, and awaiting_confidential_init are all untested. No bugs or security issues found — the comments are suggestions for improving test completeness.

**`tests/supervisor/test_allocation_verdict.py`** (line 165): The `build_plan` tests only use entries without a `"message"` key (which returns UNVERIFIABLE). No test exercises the REJECTED path (e.g. a message whose item_hash doesn't match, or a bad signature), and the rejected dict returned by `build_plan` is never asserted on. Consider adding a test that passes an entry with a malformed embedded message and verifies it appears in the rejected dict with code `"invalid_message"`.

**`tests/supervisor/test_allocation_verdict.py`** (line 100): Only `payment_stream` retention is tested. The `payment_credit`, `gpu`, `confidential`, and `operator_policy` branches of `_retention_reason` (verdict.py:59-69) are uncovered. Consider adding tests for each retention reason — especially `operator_policy`, which is the implicit fallback for a plain persistent VM.

**`tests/supervisor/test_allocation_verdict.py`** (line 58): All tests use the default `VmStatus.RUNNING`. The `LIVE_STATUSES` check (verdict.py:97-98) for BOOTING/DEFINED and the non-running skip (verdict.py:100) are untested. A test for a STOPPED VM that's in the plan (should be re-admitted as a candidate) and a STOPPED VM not in the plan (should be skipped) would document the intended behavior.

**`tests/supervisor/test_allocation_verdict.py`** (line 27): `awaiting_confidential_init=True` is never exercised in any test. This branch (verdict.py:97) treats a confidential VM waiting for init as `unchanged` if it's in the plan — worth a dedicated test since it's a subtle lifecycle edge case.

**`src/aleph/vm/agent/allocation/verdict.py`** (line 49): `entry["item_hash"]` uses bare key access while `verify_entry` uses `.get()` defensively for everything else. A malformed entry without this key will raise `KeyError`. If the handler validates the body shape before calling `build_plan`, this is fine — but the inconsistency with `verify_entry`'s defensive style is worth a brief comment.

**`src/aleph/vm/agent/allocation/verdict.py`** (line 55): The plan_id includes rejected hashes, so two pushes with the same accepted set but different rejected entries get different plan IDs. The docstring says "a stable identity for a plan" — rejected entries are explicitly excluded from the plan (line 44), yet they influence its identity. This seems intentional for response correlation, but worth a brief comment clarifying why.

foxpatch-aleph (changes requested):
Well-structured and consistent with the design: build_plan/compute_verdict are pure, the instance/v-program memory bucketing matches _admit and _committed_resources exactly, and the 12 new tests pass and follow the existing conventions in test_allocation_teardown.py. The blocking issue is build_plan's item_hash handling: it is the PR's own validation boundary for untrusted scheduler input, yet a single entry with a missing or malformed item_hash raises KeyError/UnknownHashError and takes down the whole push, while every other malformed field is handled as data to reject (see verify_entry's entry.get patterns). Since the next task wires this into a handler, that crash becomes a 500 on a plausible payload. The remaining comments are design questions worth resolving before the endpoint lands (node_hash=None conflated with node_mismatch; recreate double-counting against simulate's promise to never be weaker than the enforced path) plus dead code and small nits.

**`src/aleph/vm/agent/allocation/verdict.py`** (line 49): This is the one place a malformed entry crashes the whole push. `entry["item_hash"]` raises KeyError when the key is absent, and `ItemHash(...)` raises UnknownHashError on a non-hash value (verified against the installed aleph_message). Meanwhile `verify_entry` treats every other malformed field as data to reject, and even logs it via `entry.get("item_hash")` (verify.py:102, 106), which clearly anticipates it can be missing or wrong. Once a handler calls this, one bad entry 500s the entire push instead of producing the per-entry rejection the docstring promises. Please fold the conversion into the rejection path (try/except around both the lookup and the construction, rejecting with code `invalid_message`).

**`src/aleph/vm/agent/allocation/verdict.py`** (line 79): When `node_hash is None` the VM is rejected as `node_mismatch` with the message `this instance is allocated to a different node`, which is inaccurate: the agent just does not know its own hash yet (NodeIdentity discovery pending). The legacy path deliberately answers 503 here (views/__init__.py:819-820) so the scheduler retries instead of treating it as a definitive rejection. Consider a distinct code such as `node_hash_unknown`, or at least document that callers must pass a discovered hash — the `None` default makes the misleading mode easy to hit silently.

**`src/aleph/vm/agent/allocation/verdict.py`** (line 127): Recreate double-counting: a plan entry that exists on the supervisor in a non-live state (STOPPED/FAILED are not in LIVE_STATUSES) becomes a candidate, but it is neither in `removing` nor excluded from the committed sums, so its own stale registry record counts against its re-admission. `check_capacity` has `exclude_vm_hash` for exactly this in the enforced path (run.py:391); as written the verdict can answer `insufficient_capacity` for a recreate the real create would admit. It fails safe, but the advisory answer is then weaker than the enforced one, which simulate's docstring promises never happens. Worth a comment at minimum, or passing the not-live plan hashes to simulate as also-released/excluded.

**`src/aleph/vm/agent/allocation/plan.py`** (line 17): `AllocationState` and `FailureRecord` (line 48) are referenced nowhere in this PR, tests included. If they belong to a later task in the stack, they would land better with the code that uses them — as shipped they are untested and free to drift. (The `AllocationState` docstring is great, for the record; the placement is the question.)

**`src/aleph/vm/agent/allocation/verdict.py`** (line 87): `registry` and `capacity` are untyped while every other parameter is annotated, so the duck-typed contract (`registry.get`, `capacity.simulate`) is only discoverable from the tests. A small Protocol, or the concrete `AgentVmRegistry`/`CapacityManager`, would document it.

**`src/aleph/vm/agent/allocation/verdict.py`** (line 109): Membership in the `unchanged` list makes this an O(n^2) scan over the plan. Fine at current sizes; a set built in the first loop would keep it linear and read the same.

**`tests/supervisor/test_allocation_verdict.py`** (line 169): The two `build_plan` tests only assert plan_id stability. The rejection path (invalid embedded message -> `rejected` populated, entry kept out of the plan) has no coverage here, nor do the `retained` reasons other than `payment_stream` (gpu, confidential) or the `node_hash=None` case — the gap matters most for `build_plan`, since that is the code path a malformed push exercises first.

foxpatch-aleph (changes requested):
The core verdict logic is well-designed and verified: admission accounting mirrors the enforced path (recreating discounts match exclude_vm_hash, memory-bucket choice matches _admit, release-aware simulate wiring is correct), the node-pin handling reproduces the legacy 503/400 distinction, and retention reasons stay parallel to is_removable_by_allocation. Tests pass, mypy is clean, and the purity/await-free constraint holds. However, two verified defects sit at the validation boundary this PR exists to establish: build_plan raises TypeError on a non-list `vms` field, contradicting its documented never-raise contract for scheduler-controlled data, and duplicate item_hash entries can produce a simultaneously rejected-and-accepted answer. Both are trivial to fix now, before the handler in a later task builds on this contract.

**`src/aleph/vm/agent/allocation/verdict.py`** (line 72): This raises TypeError when `vms` is not a list: I verified `build_plan({"vms": None})` -> TypeError: 'NoneType' object is not iterable, and `build_plan({"vms": 5})` similarly. That contradicts the boundary's own contract ("a bad entry is data to reject, never an exception") for a body the scheduler controls — a JSON body that serializes `vms` as null crashes where every other malformed shape is answered. Consider `vms = body.get("vms") if isinstance(body.get("vms"), list) else []` (a string `vms` also iterates character-by-character into rejected entries, which is noisy).

**`src/aleph/vm/agent/allocation/verdict.py`** (line 87): A duplicate item_hash can end up in both `entries` and `rejected`: I verified that pushing the same hash twice — once with a message that fails verify_entry and once with one that passes — puts it in the plan AND in the rejected dict, so the response says both "rejected: A" and "accepted/pending: A". Duplicate entries also overwrite last-write-wins (a second unverified entry silently downgrades a first verified one to pending). Since rejected entries are deliberately excluded from the plan so nothing downstream acts on them, a hash should probably resolve to exactly one answer — either reject duplicates outright or make the first entry for a hash authoritative.

**`src/aleph/vm/agent/allocation/verdict.py`** (line 142): A VM dropped from the plan while BOOTING/DEFINED/STOPPING gets no verdict category at all — not removing, not retained, nothing — even though the desired-state swap will presumably act on it. If that's intentional (transient states resolve to RUNNING and get answered on the next push), it deserves a comment; if not, the immediate answer under-reports what the reconciler will do. Worth confirming against the reconciler task since this is the immediate-answer contract the scheduler will code against.

**`src/aleph/vm/agent/allocation/verdict.py`** (line 104): "operator_policy" is unreachable: _retention_reason is only called when is_removable_by_allocation returned False, and its predicate order mirrors that function exactly, so the fallthrough only fires if the two lists drift apart. The duplication is the real risk — adding a new retention rule to one function but not the other silently produces wrong answers. Consider a comment saying the fallthrough guards against drift, or deriving the reason inside teardown.py next to the predicate it explains.

**`src/aleph/vm/agent/allocation/verdict.py`** (line 85): Minor test gap: the tests cover the unusable-item_hash path into `rejected`, but not this branch — an entry whose embedded message is REJECTED by verify_entry (bad signature, hash mismatch). A test with a patched verify_entry returning (REJECTED, None, reason) asserting the entry never enters the plan would lock in the exclusion contract.

**`tests/supervisor/test_allocation_verdict.py`** (line 1): PR description nits (not code): it says 12 new tests but the file has 15, and it describes an `AllocationState` that does not appear anywhere in the diff — both look like stale text from an earlier iteration.

foxpatch-aleph (commented):
Well-designed and well-documented: build_plan keeps rejected entries out of the plan so nothing downstream acts on unverified data, compute_verdict is pure and await-free as claimed, the node_hash_unknown/node_mismatch split mirrors the legacy 503/400 path, and the recreate discount parallels check_capacity's exclude_vm_hash. All 15 tests pass and mypy is clean on the new modules. The findings are semantic subtleties rather than blocking bugs: the recreating discount is applied even when the recreate is rejected or still pending (leaving the stale record committed in the enforced path, making the advisory answer potentially stronger than enforcement for sibling candidates), VmStatus.STOPPING is treated as dead despite possibly still consuming resources, and the build_plan rejection path for verify_entry REJECTED outcomes (the security-relevant case) plus the awaiting_confidential_init branch lack test coverage.

**`src/aleph/vm/agent/allocation/verdict.py`** (line 180): The `releasing` set includes every `recreating` hash unconditionally, even when that recreate ends up rejected (insufficient capacity) or stays `pending` (verified is None, so it is never a candidate). In those cases the stale record remains committed in the enforced path — `check_capacity` only discounts it via `exclude_vm_hash` for its own create — so sibling candidates can be told `accepted` against memory that later enforcement refuses. That is the opposite of the invariant `CapacityManager._check_against` documents ("an advisory answer can never be stronger than the enforced one"). If the reconciler promises to forget dead records before creating anything else this is fine, but that contract lives in a later task; consider only discounting a recreate when it is actually admitted, or noting the promise here.

**`src/aleph/vm/agent/allocation/verdict.py`** (line 52): VmStatus.STOPPING is not in LIVE_STATUSES, so a plan entry in that state is classified as `recreating` and its committed memory/vCPUs are credited back via `releasing`, even though a stopping VM may still be consuming them. Transient, but worth a comment on LIVE_STATUSES confirming it is deliberate (and that the plan re-creating a VM mid-teardown is the intended read).

**`src/aleph/vm/agent/allocation/verdict.py`** (line 72): The comment frames build_plan as the validation boundary, but only entry-level shape is defended: a non-dict body raises AttributeError on `body.get`, and a non-list `vms` value (e.g. `{"vms": 5}`) raises TypeError, both escaping the boundary as exceptions. Fine if the future handler validates the body shape with a schema, but if this function is the boundary, `body.get("vms", []) or []` with an isinstance check on the list would make it airtight.

**`src/aleph/vm/agent/allocation/verdict.py`** (line 104): The `operator_policy` fallback is unreachable today: every way `is_removable_by_allocation` can return False (non-persistent, payment stream/credit, gpus, confidential) maps to an earlier reason. It only fires if the two functions drift apart, and then it silently reports a misleading reason. Consider a comment tying _retention_reason to is_removable_by_allocation so they are updated together.

**`tests/supervisor/test_allocation_verdict.py`** (line 182): The security-relevant rejection path — verify_entry returning REJECTED (tampered content, item_hash mismatch, bad signature) flowing through build_plan into the `rejected` dict — is never exercised; only the unusable-hash rejection is. A test reusing sign_message from test_allocation_verify.py with a tampered body would pin the {"code": "invalid_message", ...} response. Also uncovered: the awaiting_confidential_init -> unchanged branch (verdict.py:136) and the non_persistent/gpu/confidential retention reasons.

**`tests/supervisor/test_allocation_verdict.py`** (line 1): Description nits: the PR says 12 new tests but the file has 15, and it refers to an `AllocationState` type that does not exist in this diff (presumably the PlannedVm/AllocationPlan structures).

foxpatch-aleph (approved):
Solid, well-documented module with high test quality: the accepted/rejected/pending/removing/retained classification is correct against the supervisor's view, the recreating discount faithfully mirrors the enforced path's exclude_vm_hash, and node-pinning semantics match the legacy endpoint including the retry-vs-settled distinction. The separation of build_plan (validation boundary) from compute_verdict (pure diff against the supervisor) is clean, and the test that wires in the real CapacityManager to catch protocol drift is excellent. I found three minor edge cases, all verified by execution: (1) compute_plan_id merges accepted and rejected hashes into one sorted list, so {A accepted, B rejected} and {B accepted, A rejected} collide on the same plan_id; (2) pushing the same hash twice with contradictory embedded messages puts it in both rejected and plan.entries, violating the module's stated invariant that rejected entries never enter the plan; (3) a non-iterable body["vms"] (e.g. {"vms": 5}) raises TypeError, escaping the never-raise contract the code comment promises. None block: all require a malformed or hostile scheduler push and none are security-relevant.

**`src/aleph/vm/agent/allocation/verdict.py`** (line 83): Merging accepted and rejected hashes into one sorted list makes plan_id non-unique: a push where A is verified and B is rejected and a push where B is verified and A is rejected both hash to the same sorted list, and I confirmed they produce identical plan_ids. A separator between the two lists (or hashing them separately) would keep the identity faithful. Also worth reconsidering whether rejected entries should contribute to the identity at all: build_plan's docstring says rejected entries "never enter the plan", yet they change its plan_id, so two pushes of the same requested set get different ids when their junk entries differ.

**`src/aleph/vm/agent/allocation/verdict.py`** (line 80): A hash pushed twice — once with a bad embedded message, once with no message — ends up in both `rejected` and `plan.entries` (verified by execution), which contradicts the invariant that rejected entries never enter the plan and so nothing downstream can act on them: the answer to the scheduler is self-contradictory (both rejected and pending) and the pending entry proceeds to the CCN fetch. Popping the hash from `entries` when rejecting would close it: `rejected[vm_hash] = {...}; entries.pop(vm_hash, None); continue`.

**`src/aleph/vm/agent/allocation/verdict.py`** (line 67): The comment above promises a bad entry is "data to reject, never an exception", but this line raises TypeError when `vms` is a non-iterable like `{"vms": 5}`. A body shape like that should arguably be the handler's job to validate, but since this function is the declared validation boundary (and no endpoint calls it yet), either a `isinstance(..., list)` guard here or a docstring note that the handler must validate the body shape first would make the contract airtight.

**`src/aleph/vm/agent/allocation/verdict.py`** (line 131): Minor test-coverage nit: the `info.awaiting_confidential_init` short-circuit is the only branch of compute_verdict without a test, and the STOPPING/STOPPED-not-in-plan-is-skipped path (line 137) is untested too. Both are one-line test additions to the existing table.

foxpatch-aleph (commented):
Well-crafted, well-documented, well-tested code: rejected entries provably never enter the plan, the node-hash gate mirrors the legacy path's semantics exactly, and the retention-reason mirror is faithful for every reachable branch. The 16 tests pass, mypy is clean, and there are no security concerns in these pure functions. The main point to address is at verdict.py:172 — folding `recreating` into `releasing` is a no-op for candidate recreates (simulate already excludes candidate records), so its only real effect is freeing the old record for pending or node-refused entries, which is more permissive than the comment justifies and than the enforced path would be. Remaining items are edge cases: a `null` `vms` key raises out of build_plan contrary to its stated boundary contract, a duplicate hash can land in both `rejected` and the plan, `operator_policy` is unreachable, and only the `payment_stream` retention reason is test-pinned. None are dangerous since the advisory answer is re-checked by the enforced path and no endpoint calls this yet.

**`src/aleph/vm/agent/allocation/verdict.py`** (line 172): The `| recreating` part of `releasing` is a no-op for the case the code comment describes: `CapacityManager.simulate` already excludes candidate hashes from the committed sums and skips them in the release loop ("Already discounted above"), so a verified recreate is never judged against its own record regardless of this set. The only behavioral effect here is for recreates that are NOT candidates — a pending entry (`verified is None`) or a node-refused one — whose old registry record is then credited back to OTHER candidates. That is more permissive than the enforced path and more permissive than the comment implies. If the broader release is intended ("the plan intends to replace that record"), the comment should say so; if only the self-discount was intended, `recreating` should not be passed here at all. Note also that `test_a_planned_vm_the_supervisor_holds_dead_is_not_judged_against_itself` (tests/supervisor/test_allocation_verdict.py:208) asserts this against a mocked capacity, so it validates the plumbing rather than the semantics — the pending-recreate case, where the real behavior differs, has no test.

**`src/aleph/vm/agent/allocation/verdict.py`** (line 67): Confirmed experimentally: `body.get("vms", [])` raises `TypeError` for `{"vms": null}` or a non-iterable value, which escapes the boundary contract stated a few lines below ("a bad entry is data to reject, never an exception") at the body level. Fine if the endpoint pre-validates the body with pydantic, but `body.get("vms") or []` would keep the contract airtight here too.

**`src/aleph/vm/agent/allocation/verdict.py`** (line 82): A hash pushed twice — once with a message that fails verification, once bare — ends up in both `rejected` (line 80) and `plan.entries` (here, as pending), giving the scheduler contradictory answers for a single hash. Consider skipping a hash already decided, e.g. `if vm_hash in rejected: continue` at the top of the loop, or at least documenting that last-entry-wins applies.

**`src/aleph/vm/agent/allocation/verdict.py`** (line 137): A running VM absent from the plan whose hash has no registry record is skipped with no verdict entry: the scheduler is told neither `removing` nor `retained`, and the reconciler will do nothing. This is the right fail-safe, but since the scheduler is the one driving removals, silence here means it has no way to learn that nothing will happen. Consider reporting it via `retained` with a reason like `unknown_record` — or a comment pinning the intent if silence is deliberate.

**`src/aleph/vm/agent/allocation/verdict.py`** (line 99): `"operator_policy"` is unreachable as written: every case where `is_removable_by_allocation` returns False (not persistent, stream, credit, gpus, confidential) is caught by an earlier branch. Fine as a defensive default, but if it is meant to be a real policy bucket the ordering with `is_removable_by_allocation` needs a comment so the two stay mirrored.

**`tests/supervisor/test_allocation_verdict.py`** (line 101): Only the `payment_stream` retention reason is pinned by a test. Since `_retention_reason` must stay a faithful mirror of `is_removable_by_allocation`, a small parametrized case for `non_persistent`, `gpu`, and `confidential` would prevent the mirror from drifting silently. (Minor: the PR description says 12 new tests; the file has 16.)

odesenfans:
Thanks, both rounds. Every finding reproduced by execution before acting on it; five commits, one per concern.

**`| recreating` (parity break) — fixed, and the most serious of the batch.** Confirmed exactly as described. With a 30720 MiB bucket, a stale record holding 16384 MiB for a VM that is planned but still *pending* its message bought a second 16384 MiB instance an accepted verdict:

```
ADVISORY: accepted = [A]
ENFORCED: A refused -> committed 16384 MiB, required 16384, cap 30720
```

Nothing is stopping the pending VM, so nothing should have been credited. And your reading of the redundancy is right: `_committed_resources(candidate_hashes)` plus the "Already discounted above" skip means the set never did anything for the candidates it was written for. So it is gone rather than narrowed. You were also right that `test_a_planned_vm_the_supervisor_holds_dead_is_not_judged_against_itself` only proved plumbing; it is replaced by two tests over a real `CapacityManager`, one per direction, each checked against the mutation it exists to catch.

**Duplicate hash in both halves — fixed.** It reproduced in *both* orders, so a rejection is now sticky and also drops a hash an earlier entry had planned.

**`plan_id` collision — fixed.** The two halves are digested apart and then together. I used digests rather than a separator because a rejected key is whatever junk the push carried in place of a hash, so it can contain the separator. On your side question: rejected hashes still contribute, since dropping them would collide two genuinely different pushes.

**Malformed `vms` — fixed, but not the suggested way.** `body.get("vms") or []` is fail-dangerous here: an empty plan is a real instruction, the one that stops every VM this node runs and the scheduler no longer lists, so a body we cannot read must never resolve to one. There is a worse case neither review caught, which makes the point:

```
{"vms": "abc"} -> no raise, entries: [], rejected: ['None']
```

A string is walked character by character straight into a stop-everything plan. So `build_plan` now requires the list and raises otherwise, for the handler to answer 400; `{"vms": []}` still means what it says. Docstring updated, since the never-raise contract was about entries and the body shape is a different question.

**`operator_policy` unreachable — correct**, kept as the defensive default with a comment, and now backed by a parametrized test over all five reasons so the mirror with `is_removable_by_allocation` cannot drift silently.

**Coverage gaps — added.** `awaiting_confidential_init`, plus both silent skips. Worth noting my first attempt at the confidential test was not discriminative: I used `DEFINED`, which is already in `LIVE_STATUSES`, so it passed with the flag removed. It now uses a non-live status, which is the only case where the flag does any work.

**Not doing:** reporting a running VM with no registry record as `retained: unknown_record`. It is a real information gap, but an edge case of an edge case; the intent is pinned with a comment instead.

`ruff format`, `isort` and `mypy src/aleph/vm/` clean (47 errors, the dev-2.1 baseline unchanged). Suite is 44 failures against a detached dev-2.1 baseline of 44, all in `test_vprogram_launch.py`, `test_dns_configuration.py` and `views/test_operator_retire.py`, none touched here.

foxpatch-aleph (approved):
This PR is correct and carefully engineered. build_plan is a proper validation boundary: unusable hashes are rejected rather than raised, both duplicate-order cases are airtight (the "vm_hash in rejected" guard plus entries.pop), and unreadable bodies raise ValueError instead of being read as an empty plan, which matters because an empty plan stops everything the node runs. compute_verdict faithfully mirrors the legacy node-hash path, keeps releasing and candidates disjoint by construction, and is await-free as the design requires; _retention_reason safely mirrors is_removable_by_allocation with a fail-safe operator_policy fallback, and a test pins the mirror. The 33 tests pass, mypy is clean, and the three tests driving a real CapacityManager are a good guard against doubles drifting from the real simulate() signature. Only non-blocking observations remain.

**`src/aleph/vm/agent/allocation/verdict.py`** (line 208): compute_verdict never passes available_gpus to simulate, and the _Capacity Protocol can't express it, so any candidate requesting a GPU is systematically refused (simulate answers gpu_unavailable without an inventory). That is the right fail-safe for the advisory capacity-check endpoint, but if this verdict is authoritative for placement the scheduler can never allocate a GPU VM through this push path. If that is intended for now, a sentence in compute_verdict's docstring would help; otherwise consider threading the supervisor's inventory through the handler when task 5 wires the endpoint.

**`src/aleph/vm/agent/allocation/verdict.py`** (line 64): Minor: the within-half "\n".join(sorted(hashes)) is still ambiguous for rejected keys, which are arbitrary junk and may contain newlines — a push refusing the single key "a\nb" and one refusing the two keys "a" and "b" share a plan_id. plan_id is correlation-only so this is cosmetic, but the docstring's separator-safety argument only covers the between-halves join; hashing each key separately into the digest would close it if you care.

**`src/aleph/vm/agent/allocation/verdict.py`** (line 156): Nit: registry.get is executed even when vm_hash is in plan.entries, where the result is unused. Moving the lookup below the membership check would skip it, though the current shape reads fine.

**`tests/supervisor/test_allocation_verdict.py`** (line 319): Nit: AgentVmRegistry is already imported at module level (line 15); the local re-import here is redundant.

odesenfans:
Thanks, all four correct. Three cleanups pushed; GPUs stay for the endpoint PR.

**Rejected keys joined on a newline — fixed, and you caught me applying my own argument by halves.** The last round digested the two halves apart precisely because a rejected key is arbitrary junk that may hold the separator, then went on joining the keys inside each half on `"\n"`:

```
rejected keys 1: ['a\nb']
rejected keys 2: ['a', 'b']
SAME PLAN_ID: True
```

Each key is now digested before the join, which is what the docstring already claimed. Cosmetic in effect, but the comment was overstating the code.

**Dead `registry.get` — fixed, and it was my own doing.** Removing `recreating` deleted the `elif record is not None` that read it, so the lookup survived its only caller. Moved under the branch that uses it.

**Redundant import — fixed.** The same commit added the module-level `AgentVmRegistry` and left the local one behind; the rest of that block is hoisted with it.

**GPUs — documented here, threaded in the endpoint PR.** Confirmed the blast radius first: `_take_gpus` returns early on an empty `device_ids`, so only candidates actually asking for a card are refused, with `"GPU availability was not checked for this plan"`. The read is `supervisor.get_host_info()`, which is async, and this module is await-free by design so the handler cannot yield between reading the supervisor's view and swapping the desired state. Threading it therefore belongs to the endpoint that makes this verdict authoritative, not here. `compute_verdict`'s docstring now records the limitation, and it is an explicit item on that task rather than an implied one.

One wrinkle for whoever picks it up: GPU holds are user-agnostic, so a card the plan's own user has already reserved reads as taken by someone else, and a push re-stating an allocation that user just made can refuse itself.

34 tests pass, `ruff format` and `isort` clean, mypy at the `dev-2.1` baseline (47), suite 44 failures against a baseline of 44 in files this PR does not touch.

foxpatch-aleph (changes requested):
Overall this is careful, well-documented code: build_plan's hardening (list-type validation, duplicate handling in both orders, separator-safe order-independent plan IDs) is excellent, the recreate/release accounting in compute_verdict correctly leans on simulate's candidate discounting instead of double-crediting, _retention_reason is a faithful mirror of is_removable_by_allocation and a test pins that, and node-pin semantics exactly match the legacy path. Tests pass (34/34) and mypy is clean. One genuine bug blocks: compute_verdict converts supervisor vm_ids with an unguarded ItemHash(info.vm_id) at verdict.py:161, while the codebase's established convention (reconciler.supervisor_hashes and tasks.check_payment, both with docstrings explaining why) drops such ids with a warning, and build_plan in this same PR explicitly holds itself to "one unusable hash must not take down the whole push". A single VmInfo whose id is not an item hash (an operator or foreign-tenant VM — see the ownership note in views/__init__.py:250) will raise out of compute_verdict and fail the entire plan push. Two minor notes follow on status asymmetry and an untested safety property.

**`src/aleph/vm/agent/allocation/verdict.py`** (line 161): This unguarded `ItemHash(info.vm_id)` raises `UnknownHashError`/`ValueError` for any VmInfo whose id is not an item hash, taking the whole `compute_verdict` call — and thus the entire plan push — down with it. `infos` is the supervisor's view, and this codebase already defends this exact conversion everywhere else it happens: `supervisor_hashes` (reconciler.py:176, "an id that is not a plausible hash is dropped rather than raised on"), `check_payment` (tasks.py:373, "one unparseable id must not take the whole payment sweep down"), and the ownership note in views/__init__.py:250. `build_plan` in this same PR catches around its own `ItemHash(str(raw_hash))` for exactly this reason. Please wrap this conversion the same way — try `ItemHash(str(info.vm_id))` / `except (UnknownHashError, ValueError): log a warning and skip the info`, mirroring `supervisor_hashes`.

**`src/aleph/vm/agent/allocation/verdict.py`** (line 47): Minor asymmetry worth a comment: a VM in DEFINED or BOOTING counts as "live" when the plan still lists it, but when the plan drops it only RUNNING ones are reported (line 182 requires RUNNING for removing/retained). A VM stuck in DEFINED (e.g. a create that failed mid-flight) that the scheduler drops is therefore neither removed nor retained — it lingers unseen. If that's deliberate (failed/defined VMs are cleaned by another path), a one-line note saying so would help; if not, the removal gate may want widening.

**`src/aleph/vm/agent/allocation/verdict.py`** (line 217): Test gap: the docstring states "no plan can place a GPU VM" (no `available_gpus` reaches `simulate`), and the PR description repeats it, but no test exercises it — a candidate carrying `gpu_device_ids` should land in `rejected` with `gpu_unavailable`. Since this is a safety property (an advisory yes must never read as one that covered the GPU), one test would keep it from silently regressing; the related claim that "stop B, allocate C onto B's card" is refused (a GPU VM dropped from the plan is `retained` as `gpu`, and simulate never credits releases for cards) is also untested end-to-end.

odesenfans:
Blocker fixed in `7e502def`. Reproduced first:

```
RAISED: UnknownHashError -> Could not determine hash type: 'operator-scratch-vm'
=> the whole plan push fails, including the entries that were fine
```

`compute_verdict` now drops such an id with a warning the way `supervisor_hashes` and `check_payment` drop theirs. The argument that settled it is the one about this PR rather than the convention: `build_plan` catches around its own `ItemHash(str(raw_hash))` on the principle that one unusable hash must not take down the whole push, and `compute_verdict` forty lines later took the opposite stance on the same push. Test included, and checked against the unguarded conversion so it cannot regress quietly.

One correction, in your favour and against mine: the convention is 2-of-3, not universal. The legacy `update_allocations` does the same unguarded `ItemHash(info.vm_id)` at `views/__init__.py:576`, so this is a pre-existing hole I mirrored rather than one this PR invented. It wants the same guard, but as its own change rather than folded in here.

Leaving the other two, deliberately:

**Status asymmetry** — not widening the gate. The legacy path gates removal on `info.status is VmStatus.RUNNING` too, and `compute_verdict` exists to mirror it; widening would make the new path stop VMs the old one leaves alone, which is a behaviour change wearing a refactor's clothes. Your observation that a DEFINED VM dropped from the plan lingers unseen is correct, but it is a property of the allocation contract and belongs to the reconciler if it is ever addressed.

**GPU test** — fair hit, and it stays on the list rather than getting done here. `simulate` not crediting released cards is already pinned in `test_agent_capacity.py` and the `retained: gpu` half by the parametrized retention test, so the gap is specifically the refusal itself; it lands with the endpoint PR that threads the inventory through, where the property becomes load-bearing rather than incidental.

35 tests pass, `ruff format` and `isort` clean, mypy at the `dev-2.1` baseline (47), suite 44 failures against a baseline of 44 in files this PR does not touch.

foxpatch-aleph (approved):
This PR is solid, carefully defensive, and well tested. build_plan treats the scheduler-controlled body as data at the right boundary (raising only on a body shape that could be misread as an empty plan), and its duplicate-hash handling is correct in both orderings. compute_verdict's partition is exhaustive, _retention_reason mirrors is_removable_by_allocation for every input that can reach it, node-pinning matches the legacy notify_allocation semantics, and the pure/await-free design with a documented handler contract avoids the stale-verdict race. The plan-id digest provably resists both the planned/rejected merge collision and separator injection through junk keys. All 35 new tests pass, as do the neighboring allocation/capacity suites. Remaining findings are nits: a redundant exception tuple (UnknownHashError already subclasses ValueError), rejection keys collapsing for repeated unusable hashes, the undocumented STOPPING-status recreate path, and small test-coverage gaps (BOOTING/DEFINED, GPU-candidate refusal through compute_verdict).

**`src/aleph/vm/agent/allocation/verdict.py`** (line 157): Nit: `except (UnknownHashError, ValueError)` is redundant — UnknownHashError subclasses ValueError (it's the whole reason a bare ValueError would do here). Harmless as documentation of intent, but `ValueError` alone would be simpler.

**`src/aleph/vm/agent/allocation/verdict.py`** (line 99): Cosmetic: several malformed entries with the same unusable hash (e.g. two `{}` entries) collapse into a single `rejected["None"]` key, so the scheduler learns of one rejection where the push contained several. Keys must be strings so there is no clean fix, but it might be worth a note that the count is lossy.

**`src/aleph/vm/agent/allocation/verdict.py`** (line 188): This comment covers a planned VM the supervisor holds dead (STOPPED/FAILED), but a planned VM in VmStatus.STOPPING takes the same recreate-candidate path while still winding down in the supervisor. If that is intended (the future handler/reconciler presumably tolerates a create racing the old instance's teardown), a sentence saying so would help; if not, STOPPING may belong in the unchanged set or at least in a test.

**`tests/supervisor/test_allocation_verdict.py`** (line 62): Minor test-coverage gaps: only RUNNING and awaiting_confidential_init are exercised among the live-ish states — BOOTING and DEFINED (LIVE_STATUSES) have no test — and no test drives a GPU-requesting message through compute_verdict to demonstrate the promised gpu_unavailable refusal (simulate's side is well covered in test_agent_capacity.py, so this is about the verdict layer only).

PR #1173, opened 2026-08-27
feat(agent): background reconciler for scheduler allocations

odesenfans:
Task 5 of the async scheduler allocations stack (design: `docs/plans/2026-08-20-async-allocations-design.md`). Stacked on #1172.

Level-triggered: each pass re-reads the desired state and diffs it against what the supervisor reports, so an identical re-push is an empty diff and a plan arriving mid-convergence wins at the next step boundary. This is the seam pull mode plugs into: swapping "the scheduler pushed a plan" for "the agent fetched a plan" changes `submit()`'s caller and nothing else.

Creates run with bounded concurrency (`ALLOCATION_DOWNLOAD_CONCURRENCY`, default 3), because the download phase is agent-side and outside the supervisor's creation lock; `create_vm` itself stays serialized by that lock. Failures are remembered with an exponential backoff instead of vanishing, and are cleared when the VM starts or leaves the plan, so a hash the scheduler gave up on is not retried forever.

With no plan the reconciler deletes nothing. That is the post-restart state: acting on a remembered plan risks tearing down VMs migrated elsewhere during the downtime.

`pubsub` is resolved per use rather than held: `app["pubsub"]` is created by a startup hook registered after `setup_webapp`, and only when `WATCH_FOR_MESSAGES` is on, so reading it at construction would `KeyError` and caching it at task start would silently drop update-watching for reconciler-started VMs.

The supervisor event watcher now nudges the reconciler when a VM goes down, so one that dies while still planned is retried immediately rather than at the backstop interval.

The reconciler is constructed and its task started, but nothing submits a plan yet: the v2 endpoints land in the next PR. Until then it converges against `None` and does nothing.

## Tests

12 new in `tests/supervisor/test_allocation_reconciler.py`, written before the code; each drives `_converge_once()` with an injected clock, so no backoff assertion sleeps. Full `tests/supervisor` FAILED set identical to `main` (10 environment-only failures); mypy at the `main` error set exactly.

### DIFF
--- a/src/aleph/vm/agent/allocation/plan.py
+++ b/src/aleph/vm/agent/allocation/plan.py
@@ -7,10 +7,43 @@ down VMs that were migrated elsewhere during the downtime.
 
 from dataclasses import dataclass, field
 from datetime import datetime
+from enum import Enum
 
 from aleph_message.models import ItemHash
 
 from aleph.vm.agent.allocation.verify import VerifiedMessage
+
+
+class AllocationState(str, Enum):
+    """Agent-side phases, which all precede the supervisor knowing the VM.
+
+    Deliberately NOT a mirror of VmStatus: once create_vm returns, VmStatus is
+    the answer and this state stops existing. Two disjoint fields cannot drift,
+    whereas a merged enum would need maintaining in lockstep forever.
+    """
+
+    PLANNED = "planned"
+    RESOLVING = "resolving"
+    DOWNLOADING = "downloading"
+    SUBMITTING = "submitting"
+    FAILED = "failed"
+
+
+@dataclass
+class FailureRecord:
+    """Why a planned VM is not running, and when we will try again.
+
+    Kept in the reconciler and NOT in AgentVmRegistry on purpose: the registry
+    is what CapacityManager sums committed resources over, and a VM that failed
+    to start must not count as committed.
+    """
+
+    code: str
+    message: str
+    attempts: int
+    first_failed_at: datetime
+    last_failed_at: datetime
+    next_retry_at: datetime
 
 
 @dataclass
--- a/src/aleph/vm/agent/allocation/plan.py
+++ b/src/aleph/vm/agent/allocation/plan.py
@@ -5,13 +5,39 @@ reconciler with no plan deletes nothing: acting on a stale plan risks tearing
 down VMs that were migrated elsewhere during the downtime.
 """
 
+import logging
 from dataclasses import dataclass, field
 from datetime import datetime
 from enum import Enum
 
+from aleph_message.exceptions import UnknownHashError
 from aleph_message.models import ItemHash
 
 from aleph.vm.agent.allocation.verify import VerifiedMessage
+from aleph.vm.supervisor_interface.types import VmInfo, VmStatus
+
+logger = logging.getLogger(__name__)
+
+# A VM in one of these states is the supervisor's business already: creating it
+# again would be a second VM under the same hash.
+LIVE_STATUSES = (VmStatus.RUNNING, VmStatus.BOOTING, VmStatus.DEFINED)
+
+
+def by_hash(infos: list[VmInfo]) -> dict[ItemHash, VmInfo]:
+    """The supervisor's VMs, keyed by item hash.
+
+    An id that is not one is dropped rather than raised on, the way
+    supervisor_hashes and check_payment drop theirs: it names a VM the plan
+    says nothing about, so letting it through would take down a whole push or
+    wedge a whole convergence pass over something neither is about.
+    """
+    known: dict[ItemHash, VmInfo] = {}
+    for info in infos:
+        try:
+            known[ItemHash(str(info.vm_id))] = info
+        except (UnknownHashError, ValueError):
+            logger.warning("The supervisor lists a VM whose id is not an item hash: %r", info.vm_id)
+    return known
 
 
 class AllocationState(str, Enum):
--- a/src/aleph/vm/agent/allocation/reconciler.py
+++ b/src/aleph/vm/agent/allocation/reconciler.py
@@ -31,6 +31,12 @@ logger = logging.getLogger(__name__)
 # again would be a second VM under the same hash.
 LIVE_STATUSES = (VmStatus.RUNNING, VmStatus.BOOTING, VmStatus.DEFINED)
 
+# States a dropped VM can be torn down from. DEFINED and BOOTING are excluded
+# deliberately: those are mid-creation, and deleting one races the creation
+# that is still in flight. A VM stuck there is caught on a later pass once it
+# reaches one of these.
+TEARDOWN_STATUSES = (VmStatus.RUNNING, VmStatus.STOPPED, VmStatus.FAILED)
+
 
 class AllocationReconciler:
     """Holds the desired state and converges to it in the background."""
@@ -102,7 +108,10 @@ class AllocationReconciler:
                 logger.exception("Allocation reconcile pass failed; retrying at the next wake-up")
             try:
                 await asyncio.wait_for(self._wakeup.wait(), timeout=settings.ALLOCATION_RECONCILE_INTERVAL)
-            except TimeoutError:
+            except asyncio.TimeoutError:
+                # Not the builtin: the two are only the same class from 3.11,
+                # and pyproject still supports 3.10, where catching the builtin
+                # would let the first quiet interval kill the reconciler.
                 pass
             self._wakeup.clear()
 
@@ -123,7 +132,7 @@ class AllocationReconciler:
     async def _teardown_dropped(self, plan: AllocationPlan, infos: list[VmInfo]) -> None:
         for info in infos:
             vm_hash = ItemHash(info.vm_id)
-            if vm_hash in plan.entries or info.status is not VmStatus.RUNNING:
+            if vm_hash in plan.entries or info.status not in TEARDOWN_STATUSES:
                 continue
             record = self.registry.get(vm_hash)
             if record is None or not is_removable_by_allocation(record, info):
--- a/src/aleph/vm/agent/allocation/reconciler.py
+++ b/src/aleph/vm/agent/allocation/reconciler.py
@@ -1,0 +1,194 @@
+"""Converging the CRN onto the scheduler's plan, in the background.
+
+Level-triggered, not event-sourced: each pass re-reads the desired state and
+diffs it against what the supervisor reports, so a re-pushed identical plan is
+an empty diff and a plan arriving mid-convergence simply wins at the next step
+boundary. This is the seam pull mode plugs into: replacing "the scheduler
+pushed a plan" with "the agent fetched a plan" is a change to submit()'s
+caller and nothing else.
+"""
+
+import asyncio
+import logging
+from collections.abc import Callable
+from datetime import datetime, timedelta, timezone
+
+from aleph_message.models import ItemHash
+
+from aleph.vm.agent.allocation.plan import (
+    AllocationPlan,
+    AllocationState,
+    FailureRecord,
+)
+from aleph.vm.agent.allocation.teardown import is_removable_by_allocation, teardown_vm
+from aleph.vm.agent.run import start_persistent_vm
+from aleph.vm.conf import settings
+from aleph.vm.supervisor_interface.types import VmInfo, VmStatus
+
+logger = logging.getLogger(__name__)
+
+# A VM in one of these states is the supervisor's business already: creating it
+# again would be a second VM under the same hash.
+LIVE_STATUSES = (VmStatus.RUNNING, VmStatus.BOOTING, VmStatus.DEFINED)
+
+
+class AllocationReconciler:
+    """Holds the desired state and converges to it in the background."""
+
+    def __init__(
+        self,
+        *,
+        supervisor,
+        registry,
+        capacity,
+        expiry,
+        update_watcher,
+        pubsub_getter: Callable[[], object | None],
+        now: Callable[[], datetime] | None = None,
+    ) -> None:
+        self.supervisor = supervisor
+        self.registry = registry
+        self.capacity = capacity
+        self.expiry = expiry
+        self.update_watcher = update_watcher
+        # Resolved per use, never stored: app["pubsub"] is created by a startup
+        # hook registered after setup_webapp, and only when WATCH_FOR_MESSAGES
+        # is on, so it does not exist when this object is built.
+        self._pubsub_getter = pubsub_getter
+        self._now = now or (lambda: datetime.now(tz=timezone.utc))
+        self._desired: AllocationPlan | None = None
+        self._wakeup = asyncio.Event()
+        self._failures: dict[ItemHash, FailureRecord] = {}
+        self._states: dict[ItemHash, AllocationState] = {}
+
+    # ── Public surface ──
+
+    def submit(self, plan: AllocationPlan) -> None:
+        """Record a new desired state and wake the loop.
+
+        Await-free on purpose: the handler calls this in the same event-loop
+        turn as it computed the verdict it is about to return, so nothing can
+        change underneath it in between.
+        """
+        self._desired = plan
+        for vm_hash in list(self._failures):
+            if vm_hash not in plan.entries:
+                self._forget(vm_hash)
+        for vm_hash in list(self._states):
+            if vm_hash not in plan.entries:
+                self._forget(vm_hash)
+        self._wakeup.set()
+
+    def notify_vm_down(self, vm_id: str) -> None:
+        """A VM went STOPPED or FAILED. If the plan still wants it, converge."""
+        if self._desired and ItemHash(vm_id) in self._desired.entries:
+            self._wakeup.set()
+
+    def planned_hashes(self) -> set[ItemHash]:
+        return set(self._desired.entries) if self._desired else set()
+
+    def state_for(self, vm_hash: ItemHash) -> tuple[AllocationState | None, FailureRecord | None]:
+        """What the agent is doing about this VM, for the executions list."""
+        return self._states.get(vm_hash), self._failures.get(vm_hash)
+
+    async def run(self) -> None:
+        """Converge, then wait for a wake-up or the backstop interval."""
+        while True:
+            try:
+                await self._converge_once()
+            except asyncio.CancelledError:
+                raise
+            except Exception:
+                logger.exception("Allocation reconcile pass failed; retrying at the next wake-up")
+            try:
+                await asyncio.wait_for(self._wakeup.wait(), timeout=settings.ALLOCATION_RECONCILE_INTERVAL)
+            except TimeoutError:
+                pass
+            self._wakeup.clear()
+
+    # ── One pass ──
+
+    async def _converge_once(self) -> None:
+        plan = self._desired
+        if plan is None:
+            # Post-restart: no plan means delete nothing. A stale plan is worse
+            # than none, so the agent waits to be told rather than acting on
+            # what it remembers.
+            return
+
+        infos = await self.supervisor.list_vms()
+        await self._teardown_dropped(plan, infos)
+        await self._start_missing(plan, infos)
+
+    async def _teardown_dropped(self, plan: AllocationPlan, infos: list[VmInfo]) -> None:
+        for info in infos:
+            vm_hash = ItemHash(info.vm_id)
+            if vm_hash in plan.entries or info.status is not VmStatus.RUNNING:
+                continue
+            record = self.registry.get(vm_hash)
+            if record is None or not is_removable_by_allocation(record, info):
+                continue
+            logger.info("Plan %s dropped %s; tearing it down", plan.plan_id, vm_hash)
+            await teardown_vm(vm_hash, supervisor=self.supervisor, registry=self.registry)
+
+    async def _start_missing(self, plan: AllocationPlan, infos: list[VmInfo]) -> None:
+        live = {
+            ItemHash(info.vm_id) for info in infos if info.status in LIVE_STATUSES or info.awaiting_confidential_init
+        }
+        now = self._now()
+        todo = [vm_hash for vm_hash in plan.entries if vm_hash not in live and self._retry_due(vm_hash, now)]
+        if not todo:
+            return
+
+        semaphore = asyncio.Semaphore(settings.ALLOCATION_DOWNLOAD_CONCURRENCY)
+
+        async def start(vm_hash: ItemHash) -> None:
+            async with semaphore:
+                await self._start_one(vm_hash)
+
+        await asyncio.gather(*(start(vm_hash) for vm_hash in todo), return_exceptions=True)
+
+    def _retry_due(self, vm_hash: ItemHash, now: datetime) -> bool:
+        failure = self._failures.get(vm_hash)
+        return failure is None or failure.next_retry_at <= now
+
+    async def _start_one(self, vm_hash: ItemHash) -> None:
+        self._states[vm_hash] = AllocationState.DOWNLOADING
+        try:
+            await start_persistent_vm(
+                vm_hash,
+                self._pubsub_getter(),
+                supervisor=self.supervisor,
+                registry=self.registry,
+                capacity=self.capacity,
+                expiry=self.expiry,
+                update_watcher=self.update_watcher,
+            )
+        except Exception as error:
+            self._record_failure(vm_hash, error)
+            return
+        self._forget(vm_hash)
+
+    def _forget(self, vm_hash: ItemHash) -> None:
+        self._failures.pop(vm_hash, None)
+        self._states.pop(vm_hash, None)
+
+    def _record_failure(self, vm_hash: ItemHash, error: Exception) -> None:
+        now = self._now()
+        previous = self._failures.get(vm_hash)
+        attempts = (previous.attempts if previous else 0) + 1
+        delay = min(
+            settings.ALLOCATION_RETRY_BASE_INTERVAL * (2 ** (attempts - 1)),
+            settings.ALLOCATION_RETRY_MAX_INTERVAL,
+        )
+        code = getattr(getattr(error, "code", None), "value", "") or type(error).__name__
+        logger.warning("Starting %s failed (attempt %d): %s", vm_hash, attempts, error)
+        self._failures[vm_hash] = FailureRecord(
+            code=code,
+            message=str(error)[:200],
+            attempts=attempts,
+            first_failed_at=previous.first_failed_at if previous else now,
+            last_failed_at=now,
+            next_retry_at=now + timedelta(seconds=delay),
+        )
+        self._states[vm_hash] = AllocationState.FAILED
--- a/src/aleph/vm/agent/allocation/reconciler.py
+++ b/src/aleph/vm/agent/allocation/reconciler.py
@@ -40,6 +40,13 @@ logger = logging.getLogger(__name__)
 # deliberately: those are mid-creation, and deleting one races the creation
 # that is still in flight. A VM stuck there is caught on a later pass once it
 # reaches one of these.
+#
+# Wider than the removing list compute_verdict answers with, which is RUNNING
+# only: the answer names what this push stops that was up, while a pass sweeps
+# what the plan dropped whatever state it is in. Capacity stays conservative
+# under the difference, because a commitment is held by the registry record
+# rather than by the status, so a stopped VM left out of releasing has its
+# memory counted against the push that is about to free it.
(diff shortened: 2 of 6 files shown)


foxpatch-aleph (approved):
Well-structured level-triggered reconciler with sound design decisions: no-plan-means-delete-nothing for post-restart safety, lazy pubsub resolution to avoid startup-order KeyErrors, bounded download concurrency with exponential backoff for failures, and proper CancelledError propagation in the run loop. The 12 tests cover the core convergence logic thoroughly with an injected clock for deterministic backoff assertions. The supervisor event watcher integration is minimal and correct. I found no bugs or security issues — only minor test coverage gaps for `notify_vm_down` and the STOPPED-VM teardown behavior, which are non-blocking observations.

**`src/aleph/vm/agent/allocation/reconciler.py`** (line 82): `notify_vm_down` has no direct unit test in `test_allocation_reconciler.py`. The supervisor events test only verifies a MagicMock receives the call, not that a real reconciler wakes on a planned VM and stays asleep on an unplanned one. A small test calling `notify_vm_down` and checking `_wakeup.is_set()` against both a planned and unplanned hash would close the gap.

**`src/aleph/vm/agent/allocation/reconciler.py`** (line 126): Teardown only targets `VmStatus.RUNNING` VMs, so a STOPPED or FAILED VM dropped from the plan does not get its agent-side state (registry record, DB rows, staging dirs) cleaned by the reconciler. This is presumably by design (the supervisor event watcher drops that state via `_drop_vm_state`), but a test documenting that a STOPPED VM dropped from the plan is NOT torn down would make the intent explicit.

**`src/aleph/vm/agent/allocation/reconciler.py`** (line 149): `asyncio.gather(..., return_exceptions=True)` swallows `CancelledError` raised by individual child coroutines. In practice this is harmless — children don't cancel independently of the parent task — but worth noting that if `start_persistent_vm` ever raises `CancelledError` on its own, it would be silently eaten rather than propagated.

foxpatch-aleph (approved):
Well-structured level-triggered reconciler with correct backoff, concurrency bounding, and lazy pubsub resolution. The design choices (no-plan-means-no-deletes, level-triggered convergence, per-use pubsub) are sound and well-documented. Tests are thorough for converge logic with injected clocks. The findings are all non-blocking: a missing assertion on notify_vm_down in the events test, no direct unit test for notify_vm_down, a minor stale-failure-record edge case under mid-converge submit, and redundant return_exceptions=True. No correctness bugs or security issues found.

**`tests/supervisor/test_supervisor_events.py`** (line 130): The test adds `allocation_reconciler` to the app dict but doesn't assert the new behavior. Consider adding `app["allocation_reconciler"].notify_vm_down.assert_called_once_with(VM_ID)` after line 130 to verify the reconciler is actually nudged when a VM goes STOPPED.

**`src/aleph/vm/agent/allocation/reconciler.py`** (line 149): `return_exceptions=True` is redundant here: `_start_one` already catches all exceptions on line 167, so `gather` will never receive an exception result. Harmless as a safety net, but could be removed for clarity.

**`src/aleph/vm/agent/supervisor.py`** (line 194): `_reconcile_after_event_gap` calls `_drop_vm_state` for VMs that went down during the stream gap, but does not call `notify_vm_down` on the reconciler. A planned VM that died during the gap will only be retried at the backstop interval rather than immediately. This is consistent with the stated design (backstop catches missed wake-ups), but adding `app["allocation_reconciler"].notify_vm_down(vm_id)` here would close the gap.

**`tests/supervisor/test_allocation_reconciler.py`** (line 244): No direct test for `notify_vm_down` — it's only exercised indirectly through `test_supervisor_events.py` and without an assertion on the call. Consider a unit test verifying that `notify_vm_down` sets the wakeup event only when the VM is in the current plan.

**`tests/supervisor/test_allocation_reconciler.py`** (line 111): No test for the `_teardown_dropped` path when a dropped VM is STOPPED or FAILED (not RUNNING). The code correctly skips non-RUNNING VMs (line 126), but this behavior is untested. A test with `_info(HASH_B, status=VmStatus.STOPPED)` and an empty plan would verify the skip.

foxpatch-aleph (commented):
Well-structured level-triggered reconciler with sound convergence logic, proper exponential backoff, and excellent documentation. The pubsub lazy-resolution pattern correctly handles startup ordering. The core logic is correct and well-tested for the main paths. A few public methods (notify_vm_down, planned_hashes) lack test coverage, and the run() loop exception handling is untested. The _teardown_dropped only targeting RUNNING VMs is likely intentional but could benefit from a comment clarifying that STOPPED VMs are handled elsewhere. No bugs or security issues found.

**`tests/supervisor/test_allocation_reconciler.py`** (line 244): Missing test for `notify_vm_down()`: it's a public method called by the supervisor event watcher (supervisor.py:225), but no test verifies that it wakes the loop only when the down VM is still in the plan. A test should also verify it does NOT wake when there is no plan or when the VM is not in the plan.

**`tests/supervisor/test_allocation_reconciler.py`** (line 244): Missing test for `planned_hashes()`: this public method is used to expose the current plan's VM hashes but has no test coverage.

**`tests/supervisor/test_allocation_reconciler.py`** (line 244): Missing test for the `run()` loop's exception handling (line 101): the `except Exception` path that logs and retries is not exercised. Consider a test that patches `_converge_once` to raise, then verifies the loop continues.

**`tests/supervisor/test_allocation_reconciler.py`** (line 244): Missing test for `awaiting_confidential_init` in `_start_missing`: the `live` set (line 136) includes VMs with `awaiting_confidential_init=True` regardless of status, preventing double-start. This logic should have a test confirming a VM awaiting confidential init is not re-started.

**`src/aleph/vm/agent/allocation/reconciler.py`** (line 126): `_teardown_dropped` only acts on VMs with status `RUNNING`. A persistent VM in `STOPPED` status that was dropped from the plan won't be torn down here. If this is intentional (e.g. handled by the expiry manager), a brief comment would clarify the design intent.

foxpatch-aleph (commented):
Well-designed level-triggered reconciler with a clean concurrency story: _converge_once() re-reads the desired state each pass, teardown precedes start (freeing resources first), is_removable_by_allocation guards the destructive path, and the backoff math and pubsub lazy-resolution are all correct. Tests are thorough, inject the clock so backoff assertions never sleep, and all 13 pass; mypy is clean on the new file. Four non-blocking observations: (1) `except TimeoutError` only catches asyncio.wait_for timeouts on Python 3.11+ — on 3.10, which pyproject claims to support, the first backstop expiry would escape run() and silently kill the reconciler; `except asyncio.TimeoutError` is the robust spelling (matches migration/runner.py's existing practice). (2) _teardown_dropped only tears down RUNNING VMs, so a VM dropped from the plan while FAILED or STOPPED is never cleaned up and its registry record keeps capacity committed — deliberate conservatism or a gap worth documenting. (3) The supervisor/registry/capacity/expiry/update_watcher constructor params are untyped, weakening mypy on calls made through them. (4) The reconciler task is created with bare create_task rather than the file's existing create_task_log_exceptions, so an unexpected exit would go unlogged.

**`src/aleph/vm/agent/allocation/reconciler.py`** (line 105): On Python 3.10 (pyproject declares requires-python >= 3.10), asyncio.wait_for raises asyncio.TimeoutError, which is a distinct class from the builtin TimeoutError until they were unified in 3.11. There, the first backstop-interval expiry with no wake-up would escape run() and silently kill the reconciler task. It matches the existing pattern in storage.py:130, but `except asyncio.TimeoutError:` (as in agent/migration/runner.py:128) is the version-robust spelling and behaves identically on 3.11+.

**`src/aleph/vm/agent/allocation/reconciler.py`** (line 126): Only RUNNING VMs are ever torn down. BOOTING ones get caught on a later pass, but a VM dropped from the plan while FAILED or STOPPED never transitions to RUNNING, so it is never cleaned up: its disks and its registry record linger, and CapacityManager keeps counting its resources as committed. If restricting teardown to a stable running state is deliberate (e.g. to avoid deletes mid-transition), a comment saying so would help; otherwise consider including FAILED/STOPPED here.

**`src/aleph/vm/agent/allocation/reconciler.py`** (line 41): These params (supervisor, registry, capacity, expiry, update_watcher) are untyped while the rest of the interface is carefully annotated, which forfeits mypy checking of the calls made through them. teardown.py already imports AgentVmRegistry with no cycle, and this module already imports run.py, so typing them as Supervisor / AgentVmRegistry / CapacityManager / ExpiryManager / UpdateWatcher should be free.

**`src/aleph/vm/agent/supervisor.py`** (line 313): run()'s own except Exception covers pass failures, so anything else escaping the loop (e.g. the TimeoutError variant on Python 3.10) dies with no log until the task is GC'd. The file already has create_task_log_exceptions (used at line 284); using it here would make a dead reconciler visible in the logs.

foxpatch-aleph (approved):
Solid, well-documented background reconciler: the level-triggered converge loop has no lost-wakeup races, the Python 3.10 TimeoutError nuance is handled correctly, teardown excludes are consistent with start_persistent_vm's state machine, and the lazy pubsub resolution is justified. All 16 new tests pass and the broader suite failures are pre-existing environment issues unrelated to this change. Findings are nits: an in-flight start failure can re-insert a FailureRecord for a hash already dropped from the plan (lingering until the next submit), LIVE_STATUSES is duplicated between reconciler.py and verdict.py, three AllocationState members are currently dead, and the run() loop's wake/backstop machinery itself has no test coverage. No correctness or security blockers.

**`src/aleph/vm/agent/allocation/reconciler.py`** (line 177): Minor race: if submit() drops a hash from the plan while its _start_one is still in flight (inside the gather), submit prunes the record but this in-flight _record_failure re-inserts it. The stale FailureRecord then lingers in _states/_failures (and in state_for output) until the next submit() prunes it. Harmless for convergence since todo is built from plan.entries, but a `if vm_hash in self._desired.entries` guard here (or re-checking in _record_failure) would keep the reported state honest.

**`src/aleph/vm/agent/allocation/reconciler.py`** (line 32): LIVE_STATUSES is defined identically in verdict.py:52. Consider importing it from there (or a shared constants module) so the two cannot drift — they encode the same "the supervisor already owns this VM" semantics.

**`src/aleph/vm/agent/allocation/plan.py`** (line 25): PLANNED, RESOLVING and SUBMITTING are never set anywhere — only DOWNLOADING and FAILED are used. If the v2 endpoints in the next PR set them, fine; otherwise these are speculative additions to remove or defer.

**`src/aleph/vm/agent/allocation/reconciler.py`** (line 100): The run() loop's wake-up/backstop/exception-swallowing machinery is the subtlest part of this file, but every test drives _converge_once() directly. A test that runs run() with a fake clock against a raise-then-succeed _converge_once would protect the loop against future edits (e.g., someone "simplifying" the asyncio.TimeoutError catch).

foxpatch-aleph (commented):
Well-crafted PR with sound concurrency design: wait-then-clear wakeup ordering avoids lost wakeups, serialized passes prevent double-create races, and the lazy pubsub_getter is correctly reasoned and documented. All 16 tests pass. My findings are non-blocking robustness nits: (1) teardown_vm exceptions abort the whole pass while the start path has per-VM isolation, so one persistently failing delete blocks start convergence for everything else; (2) ItemHash(info.vm_id) validates and raises on malformed ids, which would wedge every pass; (3) AllocationState.PLANNED/RESOLVING/SUBMITTING and planned_hashes() are dead code pending the next PR; (4) an in-flight create can resurrect a failure record for a VM dropped by a newer plan. No security issues found.

**`src/aleph/vm/agent/allocation/reconciler.py`** (line 141): The start path isolates failures per VM (_start_one catches), but this teardown call does not: one dropped VM whose delete_vm persistently raises a non-VmNotFoundError aborts the pass before _start_missing runs, blocking start convergence for all other VMs on every pass. Consider a per-VM try/except-log-continue to match the isolation of the start path.

**`src/aleph/vm/agent/allocation/reconciler.py`** (line 134): ItemHash.__new__ validates the hash format, so a malformed vm_id from list_vms() raises UnknownHashError here and aborts the entire pass — every interval, for as long as that VM exists. vm_ids are message hashes in practice, so this is defensive only, but skipping un-parseable ids (or converting with a guard) would keep one bad row from wedging convergence.

**`src/aleph/vm/agent/allocation/plan.py`** (line 26): PLANNED, RESOLVING and SUBMITTING are never set anywhere in this PR — only DOWNLOADING and FAILED are. The reconciler calls start_persistent_vm as a black box, so these may not get a setter even in the follow-up. Worth confirming the next PR consumes them, otherwise they read as design intent the code doesn't have.

**`src/aleph/vm/agent/allocation/reconciler.py`** (line 177): If submit() drops VM X while X's create is still in flight, this re-creates a FailureRecord/FAILED state for a VM that is no longer planned, and carries stale attempts/first_failed_at if X re-enters a plan later. Impact is small (cleared on the next submit, never retried while out of plan), but checking membership in self._desired here would tighten it.

**`tests/supervisor/test_allocation_reconciler.py`** (line 145): Minor gap: awaiting_confidential_init VMs being treated as live (so they are neither restarted nor torn down) is subtle, load-bearing behavior in _start_missing with no test. One test with _info(...) plus awaiting_confidential_init=True would lock it in.

foxpatch-aleph (commented):
A well-engineered level-triggered reconciler with unusually good rationale comments and meaningful behavioral tests; the core loop, backoff, teardown exclusions, and lifecycle wiring are all correct as far as I can verify. The concerns are robustness edges rather than bugs: an unguarded ItemHash(info.vm_id) that verdict.py:_by_hash explicitly defends against would wedge every converge pass, and a single failing teardown_vm aborts the rest of the pass (including all starts) with no per-VM isolation. One design question to carry into the next PR: nothing gates _start_missing on capacity, so it matters whether capacity-rejected entries ever reach submit().

**`src/aleph/vm/agent/allocation/reconciler.py`** (line 134): This ItemHash(info.vm_id) is unguarded, while verdict.py:_by_hash in the same stack deliberately catches (UnknownHashError, ValueError) for the exact same supervisor listing, with a comment saying letting it raise takes down a whole push. Here one non-hash VM id makes every pass raise, and since list_vms() precedes both _teardown_dropped and _start_missing, all convergence (teardowns AND starts) wedges until that VM disappears. Consider reusing the guarded conversion from verdict.py here (and at line 145 in _start_missing).

**`src/aleph/vm/agent/allocation/reconciler.py`** (line 141): A raised teardown_vm propagates out of _converge_once and skips _start_missing for that pass. With a persistently failing teardown (e.g. a supervisor delete error) this blocks every planned start on every backstop interval, and unlike starts, teardowns have no failure memory or backoff — they retry at full rate forever. A per-VM try/except (log and continue) would isolate them the way _start_one already isolates starts.

**`src/aleph/vm/agent/allocation/reconciler.py`** (line 143): Design question for the v2 endpoint PR: _start_missing starts everything in the plan, but compute_verdict/simulate rejects entries for capacity. If the endpoint submits the raw plan, capacity-rejected entries get retried forever (and will start the moment capacity frees, after the scheduler was told they were rejected). Also note the concurrency here widens the agent-side admission race: check_message runs before create_vm and the registry record only lands after it returns, so up to ALLOCATION_DOWNLOAD_CONCURRENCY creates can each pass admission against the same committed sum. The legacy path started sequentially; worth stating the intended story here.

**`src/aleph/vm/agent/allocation/reconciler.py`** (line 32): LIVE_STATUSES is identical to the one in verdict.py:48 and shares its intent (\"a VM the supervisor is already working on does not need re-creating\"). Two copies in the same package can drift; consider defining it once in plan.py and importing it in both.

**`src/aleph/vm/agent/allocation/reconciler.py`** (line 80): Nit: the two loops could be one pass over set(self._failures) | set(self._states), or _forget could take a set. Harmless as is.

**`tests/supervisor/test_allocation_reconciler.py`** (line 1): run() has no coverage at all — in particular the asyncio.TimeoutError catch that was the point of the follow-up commit (a builtin TimeoutError escape on py3.10) and the wakeup set/wait/clear protocol. A test that runs one loop iteration with a short ALLOCATION_RECONCILE_INTERVAL and a pre-set _wakeup, plus one that lets the timeout fire, would pin both down. Also small gap: only DEFINED is tested for the teardown exclusion; BOOTING is untested.

odesenfans:
Rebased onto `dev-2.1` (one conflict, an import both sides added) and worked through the rounds. Six commits.

**Both robustness findings reproduced before fixing.**

The unguarded `ItemHash(info.vm_id)`:

```
RAISED: UnknownHashError -> Could not determine hash type: 'operator-scratch-vm'
=> teardowns AND starts both skipped; planned VM started 0 times
```

Worse than the `verdict.py` one you caught, because `run()` catches and logs: the task stays alive and converges nothing, every interval, for as long as that VM is listed. Rather than copy the guard, `LIVE_STATUSES` and the guarded reading now live once in `plan.py` and both modules import them, so the duplication you flagged separately is gone with it. `notify_vm_down` had the same unguarded conversion; no round mentioned it, and it is fixed too.

The unisolated teardown:

```
RAISED: RuntimeError -> supervisor delete failed
=> C started 0 times: one stuck teardown blocks every start, every pass
```

Caught per VM now, matching `_start_one`. The asymmetry inside one file was the tell.

**Also done:** the in-flight failure no longer resurrects a record for a dropped VM (raised twice); the reconciler task uses `create_task_log_exceptions` (cancellation is unaffected, since the wrapper catches `Exception` and `CancelledError` is not one); the collaborators are typed, which cost nothing, as you said, and left mypy unchanged at the branch baseline.

**`run()` now has coverage** — the wake-up, the backstop expiry, and a failed pass. The wake-up test is bounded, so a wake-up cleared before the wait fails in a second rather than sitting out the 3600s backstop it was meant to pre-empt. Every new test was checked against the mutation it exists to catch.

One caveat worth stating plainly: **the backstop test cannot pin the 3.10 nuance.** `asyncio.TimeoutError` and the builtin are the same class from 3.11, and the suite runs on 3.14, so swapping the spelling back does not fail anything. The comment in the code carries that, not the test.

Also added: `BOOTING` for the mid-creation exclusion, which `TEARDOWN_STATUSES` named and nothing asserted, and the `awaiting_confidential_init` case.

**On capacity gating** — the design question, answered rather than coded. `_start_missing` deliberately starts what the plan holds; the endpoint is what decides that only accepted entries reach `submit()`, so capacity-rejected entries never enter the plan and cannot be retried forever behind the scheduler's back. Your second point is the sharper one and you rediscovered a requirement the design wrote down and the code does not yet meet: concurrent creates each pass admission against the same committed sum, because the registry record only lands after `create_vm` returns. The design calls for provisional commitments in the reconciler for exactly this. That work lands in the endpoint PR alongside the capacity gating, since the two are one mechanism; it is not a rebase-sized change.

**Not changed:** `AllocationState.PLANNED/RESOLVING/SUBMITTING` and `planned_hashes()` are still unset and uncalled. They are consumed by the executions-list work, so they are early rather than dead; if that lands without them they should go.

24 tests pass, `ruff format` and `isort` clean, mypy at the `dev-2.1` baseline (47). Suite is 45 failures against a `dev-2.1` baseline of 45, measured in a detached worktree; all in `test_dns_configuration.py`, `test_vprogram_launch.py` and `views/test_operator_retire.py`, none touched here.

foxpatch-aleph (approved):
Well-designed level-triggered reconciler: the loop's wake-up protocol is race-free, the in-flight-create vs. plan-change race is explicitly handled, and teardown/start isolation plus bounded-concurrency downloads are correctly implemented and well commented. The async/TimeoutError handling is right for the Python 3.10 floor, teardown-status exclusions are deliberate and documented, and the lazy pubsub_getter wiring is correctly reasoned. The 24 new tests pass (as do all 88 allocation tests), and mypy shows no new errors in the changed files. All findings are minor: by_hash(infos) re-parses the supervisor list twice per pass; pubsub_getter is typed object | None where start_persistent_vm expects PubSub | None (invisible to mypy, which resolves the module to Any); PLANNED/RESOLVING/SUBMITTING enum members, planned_hashes() and state_for() are unused until the next PR; a persistently failing teardown logs a full stack trace every pass with no backoff; and there is no test asserting that re-pushing a plan containing a failed VM preserves its backoff. None of these block merge.

**`src/aleph/vm/agent/allocation/reconciler.py`** (line 139): Minor: _teardown_dropped and _start_missing each call by_hash(infos), so every pass constructs the ItemHash keys for the full supervisor list twice. Computing it once in _converge_once and passing the dict down would halve that (and keep the two views consistent by construction).

**`src/aleph/vm/agent/allocation/reconciler.py`** (line 56): pubsub_getter is typed Callable[[], object | None] but start_persistent_vm expects PubSub | None. mypy does not catch the mismatch (the module resolves to Any in practice), so a getter returning the wrong type would slip through type-checking. Callable[[], PubSub | None] would be tighter and self-documenting.

**`src/aleph/vm/agent/allocation/plan.py`** (line 51): PLANNED, RESOLVING and SUBMITTING are never assigned in this PR, and planned_hashes()/state_for() have no callers yet. Fine given the stated staging (v2 endpoints land next), but as merged this ships a few dead branches — a short note that these are for the next PR would help readers of this commit in isolation.

**`src/aleph/vm/agent/allocation/reconciler.py`** (line 158): Worth noting operationally: a VM the supervisor persistently refuses to delete now logs a full stack trace every pass (backstop interval), since teardowns carry no backoff. The per-VM isolation is the right call versus aborting the pass, but a rate-limited variant of this log might be worth considering if it ever gets noisy in production.

**`tests/supervisor/test_allocation_reconciler.py`** (line 261): Nice coverage overall. One gap: there is a test that a VM leaving the plan loses its failure record, but no test for the complement — re-pushing a plan that still contains the failed VM must preserve the backoff (submit() prunes only VMs not in plan.entries). Worth locking down since it is the property that stops a scheduler polling loop from hammering a broken VM.

foxpatch-aleph (changes requested):
Well-constructed level-triggered reconciler with unusually careful reasoning about no-plan teardown safety, backoff, and failure isolation, and the tests genuinely exercise the loop and its exclusions. However, there is one genuine race worth fixing before merge: a plan submitted while a pass is parked in list_vms() or an earlier teardown_vm() is not honored by _teardown_dropped, which still tears down VMs per the stale snapshot — destroying (GONE + volume retention) a VM the newer plan wants, where the design's own "newer plan wins at the next step boundary" promise is too late for a destructive action. The same class of race is already guarded in _start_one, so the asymmetry looks like an oversight rather than a choice. Two smaller observations: TEARDOWN_STATUSES diverges from compute_verdict's RUNNING-only removing report, and by_hash(infos) is computed twice per pass.

**`src/aleph/vm/agent/allocation/reconciler.py`** (line 152): This teardown acts on the pass's snapshot of `self._desired`, but `submit()` can replace the plan while this coroutine is parked in `await` — inside `list_vms()`, or inside an earlier `teardown_vm()`. A newer plan that re-adds a VM this snapshot dropped will still have that VM torn down, and GONE applies the volume retention policy, so "a plan arriving mid-convergence wins at the next step boundary" is too late for a destructive action. Since `submit()` always assigns a fresh object, an identity guard before each teardown is sufficient and race-free, e.g. `if self._desired is not plan: return` — skipping is safe because `submit()` sets the wakeup, so a pass with the fresh plan runs immediately. The same check in `_start_missing` would avoid a wasted create/teardown cycle, though a stale start is recoverable and less urgent. For symmetry, note the mirror-image race in `_start_one` (failure recorded after the plan dropped the VM) is already guarded, which is what makes this one look like an oversight rather than a choice.

**`src/aleph/vm/agent/allocation/reconciler.py`** (line 42): TEARDOWN_STATUSES removes from RUNNING, STOPPED and FAILED, but compute_verdict (verdict.py:184) only reports `removing` for VmStatus.RUNNING, so a dropped VM that is stopped or failed is torn down here without ever appearing in the scheduler-facing answer. The capacity direction stays conservative (under-releasing in simulate), so admission is unaffected, but is the divergence intended — verdict answers "what live capacity frees up", reconciler sweeps the rest — or should the verdict list them too? Either way a word in one of the two docstrings would help.

**`src/aleph/vm/agent/allocation/reconciler.py`** (line 144): Minor: `by_hash(infos)` is computed twice per pass, here and again in `_start_missing`. Computing it once in `_converge_once` and passing it down halves the work and guarantees the teardown and start halves see the same view.

odesenfans:
Thanks, the teardown race is real and I reproduced it before fixing it: a `list_vms` that calls `submit()` with a plan re-adding the VM, then returns it as RUNNING, gets it torn down. `VOLUME_RETENTION` defaults to `reap`, so the volumes go with it.

Fixed in `14510847`, with one difference from the suggestion. `if self._desired is not plan: return` abandons the whole teardown half, including the VMs the newer plan also drops, so a scheduler pushing faster than a teardown completes could stop making progress on teardowns entirely. Re-reading `self._desired` per VM and checking membership is just as race-free (the read and the `await` are one turn) and skips only what the newest plan wants. It also covers the "parked in an earlier `teardown_vm`" window, not just the `list_vms` one.

I left `_start_missing` unguarded on purpose: the check would narrow a window it cannot close, since the plan can change during `start_persistent_vm` too, and a stale start is torn down by the next pass. That is the asymmetry, a destroy is the one step the next pass cannot undo.

Also taken:

- `435b6981` keys the supervisor list once per pass and hands the dict to both halves.
- `c0044a9c` types `pubsub_getter` as returning `PubSub | None`. No import cycle, this module already imports `run`.
- `0209a9a1` adds the missing complement: a re-pushed plan that still lists a failed VM keeps its backoff. Pruning unconditionally in `submit()` passed the whole suite before this.
- `332d34d0` writes down why `TEARDOWN_STATUSES` is wider than the verdict's `removing`. It is deliberate, and capacity stays conservative under it because commitment is held by the registry record, not the status.

Not taken: rate-limiting the teardown stack trace, and the `_forget` two-loop nit. The unused `PLANNED`/`RESOLVING`/`SUBMITTING` members and `planned_hashes()`/`state_for()` land with the v2 endpoints in the next PR, as does capacity gating on what reaches `submit()` and the provisional commitments that close the admission race you noted.

Note that review 5145896451 was against `5ac55b33`, before the fixes pushed at 19:14. The guarded `ItemHash`, per-VM teardown isolation, shared `LIVE_STATUSES` and `run()` coverage it asks for are in `675fda04`, `1e7438a2` and `adb026b2`.

Suite: 45 failed / 1051 passed against a `dev-2.1` baseline of 45 failed / 1024 passed, same three unrelated files. mypy at the 47-error baseline, ruff and isort clean.

foxpatch-aleph (approved):
A careful, well-documented level-triggered reconciler with sound lifecycle handling. The teardown/start asymmetry (re-reading the plan immediately before an irreversible GONE, while letting starts ride on a one-pass-stale snapshot the next pass undoes) is correct, the failure/backoff bookkeeping handles every race I could trace (plan swap mid-teardown, plan drop during an in-flight create, re-pushed identical plans not resetting backoff), and the asyncio.TimeoutError-vs-builtin distinction is genuinely required given pyproject still supports 3.10. The PR is inert in production until the v2 endpoints land, which keeps its risk minimal. I could not execute the test suite in this environment due to a pre-existing web3 plugin ImportError, but the 26 tests are precise and directly exercise the behaviors the code comments claim. Minor non-blocking notes below.

**`src/aleph/vm/agent/allocation/plan.py`** (line 43): Three of the five AllocationState values (PLANNED, RESOLVING, SUBMITTING) are never assigned anywhere in this PR, and `planned_hashes()` / `state_for()` in reconciler.py have no callers outside tests — all forward declarations for the v2 endpoints in the next PR. Fine for a stacked series, but a one-line note in the enum docstring saying which PR consumes them would help readers who land here first.

**`src/aleph/vm/agent/allocation/reconciler.py`** (line 232): There is no terminal failure state or max-attempts cutoff: a planned VM that can never start (e.g., a permanently broken message) is retried forever at the capped 15-minute interval. Defensible given the plan is the authority and the scheduler can drop the hash (which clears the record), and `state_for` exposes `attempts` for observability — just confirming this is deliberate rather than an oversight.

**`src/aleph/vm/agent/allocation/reconciler.py`** (line 92): The two prune loops could be one pass over `set(self._failures) | set(self._states)`. Safe as written since both iterate snapshots, purely stylistic.

**`src/aleph/vm/agent/allocation/reconciler.py`** (line 187): For the record: a plan flip-flopping within the duration of one `teardown_vm` await can still destroy a VM the newer plan wants back (GONE reaps volumes, so the next pass re-creates it from its message, fresh). The window is one await long and the design comment covers the snapshot reasoning, so this is within the stated envelope — no change needed, just flagging the consequence.

**`tests/supervisor/test_allocation_reconciler.py`** (line 1): Could not run this suite locally: collection fails project-wide on a web3 plugin ImportError (`cannot import name 'ContractName' from 'eth_typing'`), an environment-only breakage unrelated to this PR. The tests were reviewed statically and the assertions (injected clock for backoff, list_vms-reentry race, peak-concurrency probe, loop-resilience trio) match the behaviors the production comments describe.

PR #1174, opened 2026-08-28
ci: trigger on any dev* branch, not a single hardcoded name

odesenfans:
`dev` was merged into `main` and deleted, so every workflow trigger list named a branch that no longer exists. A PR against `dev-2.1` ran **no CI at all**.

Rather than swap in the new name and be back here at 2.2, the six workflows now match the glob `dev*`, which covers `dev`, `dev-2.1` and whatever comes next.

| Workflow | push | pull_request |
|---|---|---|
| `build-deb-package-and-integration-tests` | ✓ | ✓ |
| `codeql-analysis` | ✓ | ✓ |
| `test-build-examples` | ✓ | ✓ |
| `test-rust` | ✓ | ✓ (keeps `od/**`) |
| `test-using-pytest` | ✓ | ✓ (keeps `od/**`) |
| `golden-measurements` | ✓ | ✓ (keeps `od/**`) |

One caveat worth knowing: the glob does not cross a slash, so `dev*` matches `dev-2.1` but **not** `dev/foo` — that would need `dev**`. Every release branch so far uses the dash form.

The `od/**` entries are untouched: they exist so a stacked PR (increment N+1 based on increment N's branch) carries CI before the stack reaches the release branch, which this does not affect. `deploy-main-on-staging` is `main`-only by design and is left alone.

This PR bootstraps itself: for `pull_request`, GitHub evaluates the workflow file from the merge ref, so the change is live for its own checks. Once merged, PRs already targeting `dev-2.1` pick it up the same way without needing a rebase.

`yamlfix --check .` passes, and every trigger list was verified by parsing the YAML rather than by eye.

### DIFF
--- a/.github/workflows/build-deb-package-and-integration-tests.yml
+++ b/.github/workflows/build-deb-package-and-integration-tests.yml
@@ -4,10 +4,10 @@ name: "Build Packages and tests on Droplets"
 
 on:
   push:
-    branches: [main, dev]
+    branches: [main, 'dev*']
   pull_request:
     # The branches below must be a subset of the branches above
-    branches: [main, dev]
+    branches: [main, 'dev*']
 
 # This action build the package for Ubuntu and Debian
 # Then run e2e integration tests on Digital Ocean droplets (on demand VM)
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -15,10 +15,10 @@ name: "CodeQL"
 
 on:
   push:
-    branches: [main, dev]
+    branches: [main, 'dev*']
   pull_request:
     # The branches below must be a subset of the branches above
-    branches: [main, dev]
+    branches: [main, 'dev*']
   schedule:
     - cron: '15 16 * * 0'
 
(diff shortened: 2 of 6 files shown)


foxpatch-aleph (approved):
Straightforward and correct CI trigger fix. All six workflow files that referenced the deleted `dev` branch are updated to `dev-2.1` in both `push` and `pull_request` triggers. The three workflows not mentioned in the PR (`deploy-main-on-staging`, `release-on-tag`, `pr-rating`) correctly don't reference `dev` and are appropriately left alone. The `od/**` stacked-PR branch patterns are preserved in the three workflows that had them, and their accompanying comments are updated to say `dev-2.1` instead of `dev`. No stale references to the old `dev` branch remain anywhere in `.github/workflows/`. The change is minimal, well-scoped, and self-documenting.

foxpatch-aleph (approved):
Clean, correct CI trigger fix. All six workflows that referenced the deleted `dev` branch are updated to the glob `dev*`, which covers `dev`, `dev-2.1`, and future dash-form dev branches. The three workflows with stacked-PR support correctly retain `od/**`. The three workflows that should not trigger on dev branches (deploy-main-on-staging, pr-rating, release-on-tag) are correctly left untouched. YAML quoting of `dev*` is correct, comments are updated consistently, and no stale references to bare `dev` remain.

PR #1175, opened 2026-08-28
feat(tee)!: VerificationResult carries registers, not a scalar measurement

odesenfans:
Increment 5 of the TDX design (#1144, §4.4), aleph-vm half.

`VerificationResult.measurement: Vec<u8>` becomes `registers: SevSnpRegisters { launch: Vec<u8> }`, the same shape `LaunchMeasurement.registers` took in aleph-message 1.3.0. A concrete struct now; it becomes an enum discriminated on TEE type when a second platform lands, exactly as on the message side.

**Breaking** for the JSON form (`registers.launch` replaces `measurement`). No consumer outside `aleph-tee` in this workspace.

The aleph-rs half (SDK + CLI, → 0.18.0) is a separate PR; the two are independent since `aleph-tee` does not depend on `aleph-types`.

Verified: `cargo test -p aleph-tee` 75 pass / 1 ignored (pre-existing Tier 2), fmt clean, clippy `--all-targets --all-features -D warnings` clean.


foxpatch-aleph (approved):
Clean, well-documented breaking change that replaces the scalar `measurement` field on `VerificationResult` with a `SevSnpRegisters` struct wrapping a `launch` field, mirroring the aleph-message side. The `hex_serde` attribute is correctly carried over, the verify function properly wraps the extracted measurement, doc comments are updated, and a thorough round-trip test covers the new JSON shape and the absence of the old field. No external consumers are affected in this workspace. No correctness or security issues found.

foxpatch-aleph (approved):
Clean, minimal, well-documented reshape. VerificationResult.measurement becomes registers: SevSnpRegisters { launch }, correctly mirrors the aleph-message 1.3.0 LaunchMeasurement.registers shape, and the sole constructor (verify_sev_snp_report) plus the doc comment are updated consistently. The golden-measurement re-seed is expected (aleph-tee is a measured input of the attest-agent initrd) and was regenerated in the same PR per the check script's own process, with a commit message explaining the cause. The new test covers the JSON shape, absence of the old key, and round-trip. Only non-blocking observations follow.

**`nix/golden-measurements.json`** (line 2): Deployment note worth stating explicitly in the PR: this is more than the documented JSON break — all four launch measurements moved, so any host or registry entry pinning the old compose/workload/instance measurements must migrate before it can boot-verify against a build of this tree. check-golden-measurements.sh's own guidance ("anything pinning the old measurement must migrate") applies; make sure that migration is coordinated alongside the aleph-rs 0.18.0 rollout.

**`rust/crates/aleph-tee/src/types.rs`** (line 51): Non-blocking: report_data is a strictly validated [u8; 64] (hex_serde_array rejects anything but exactly 64 bytes), but launch stays an unvalidated Vec<u8>, so a 3-byte or 100-byte hex string deserializes silently. Since the field is being reshaped anyway, a fixed [u8; 48] with the same treatment would have been free; fine to defer, just noting the asymmetry.

**`rust/crates/aleph-tee/src/types.rs`** (line 48): Non-blocking design observation: SevSnpRegisters lives in the TEE-agnostic types.rs next to TeeType::Tdx / NvidiaCc, so when a second platform lands the VerificationResult.registers field type itself will need to change (another breaking rename). A neutral name like Registers (with the SEV-SNP variant documented) would absorb that future without a rename, but matching the message-side naming is a defensible tiebreaker and the comment at line 42-46 already flags the plan.

odesenfans:
Follow-up pushed: `VerificationResult` now speaks **both** register formats on the wire — it emits the deprecated scalar `measurement` (hex duplicate of `registers.launch`) alongside `registers`, and accepts either on input; if both arrive they must agree or deserialization fails closed. Peers on either side of the 2.0 format change keep interoperating. The scalar is to be dropped after one release cycle. Goldens re-seeded for the aleph-tee edit.

foxpatch-aleph (approved):
The core change is correct and well-executed: VerificationResult.measurement becomes registers: SevSnpRegisters, and the serde try_from/into bridge through VerificationResultWire implements the right dual-write transition — it always emits both the new registers object and the deprecated scalar measurement alias, accepts either on deserialize, and fails closed when both are present but disagree. Round-trip behavior is sound, the doc comments on verify_sev_snp_report were updated consistently, and I confirmed there are no other consumers of VerificationResult in this workspace (the aleph-attest-agent crate, though built from aleph-tee, does not reference it). The golden-measurement regeneration is mechanically required — the measured agent binary is built from aleph-tee, so any source change moves all four launch measurements — but it is undeclared in the PR description, which per check-golden-measurements.sh means anything pinning the old measurements must migrate. Remaining items are minor: untested error branches (neither-field-present, non-hex legacy scalar) and the absence of length validation on launch (pre-existing looseness in hex_serde, not a regression). No correctness or security issues found; the fail-closed disagreement check in particular is the right call.

**`nix/golden-measurements.json`** (line 2): All four launch measurements move here but the PR description doesn't mention it. The change is mechanically required (the measured aleph-attest-agent binary is built from aleph-tee, so any source change moves the measurement), but per check-golden-measurements.sh this means anything pinning the old measurements — deployed verifiers, published image pins — must migrate. Please call this out in the PR description and confirm these values came from `check-golden-measurements.sh --update` on this exact tree (rather than toolchain/dependency drift, which the script warns would be a reproducibility regression rather than an intended change).

**`rust/crates/aleph-tee/src/types.rs`** (line 128): The `(None, None)` bail branch and the non-hex legacy-scalar error path (line 116/126) are the only branches of `try_from` not exercised by `verification_result_speaks_both_register_formats`. Both are reachable from JSON (both fields are `#[serde(default)]` Options), so a two-line addition to the test would close the gap: `json` with both `registers` and `measurement` removed should fail, and a non-hex `measurement` should fail.

**`rust/crates/aleph-tee/src/types.rs`** (line 51): Minor, pre-existing looseness rather than a regression: the doc says 48-byte launch digest, but neither `launch` nor the legacy scalar path validates length, so a peer can supply an empty or mis-sized digest (e.g. an empty-string legacy `measurement` decodes to an empty vec). Callers pin exact digests so the impact is low, but a length check here would give a much clearer error than a downstream pin mismatch. Consider when TDX lands and this becomes an enum.

**`rust/crates/aleph-tee/src/types.rs`** (line 74): The doc comment here is accurate, but note it contradicts the PR description, which claims this is breaking for the JSON form. On the wire it is not breaking for JSON consumers — the scalar alias is still emitted and accepted — only the Rust field access is breaking. Worth reconciling so downstream readers don't over-prepare for a wire break.

foxpatch-aleph (approved):
Solid, well-scoped change. The dual-format wire shim is correct: serialization emits both `registers` and the deprecated scalar `measurement`, deserialization accepts either alone or both, and the both-present case fails closed on disagreement rather than silently picking one. The golden-measurement re-seed is legitimate rather than suspicious — `aleph-tee` is a path dependency of the initrd-baked `aleph-attest-agent`, so any crate code change shifts the launch digest, and all four re-seeded values are consistent 48-byte digests, matching what `check-golden-measurements.sh` mandates. Scope claim verified: `VerificationResult` has no consumers outside `aleph-tee`, and the `SevSnpRegisters` naming mirrors the message-side type already imported from `aleph_message` in the Python tests. Remaining items are nits.

**`rust/crates/aleph-tee/src/types.rs`** (line 67): This `#[serde(with = "hex_serde_array")]` is now inert: with `try_from`/`into` at the container level, field-level serde attributes on `VerificationResult` are ignored and the wire shape comes entirely from `VerificationResultWire`. Consider dropping it or adding a note, so a reader doesn't assume it affects the JSON output.

**`rust/crates/aleph-tee/src/types.rs`** (line 52): Minor observation: `report_data` is strictly validated to exactly 64 bytes via `hex_serde_array`, but `launch` accepts any hex length (an empty string or a 3-byte value deserializes fine). Not a regression — the old `measurement: Vec<u8>` had the same property, and `VerificationResult` is a local verification output rather than a trust boundary — but since this is a fresh struct in a security-relevant path, a `len() == 48` check would be cheap symmetry.

**`rust/crates/aleph-tee/src/types.rs`** (line 128): The new test covers both-present-agree, scalar-only, registers-only, and both-present-disagree, but not this `(None, None)` bail branch nor the non-hex-legacy-measurement error path. Two small extra assertions would close those mutation-survival gaps.

PR #1176, opened 2026-08-28
feat: V-PROGRAM verified data volumes (host + guest)

odesenfans:
Makes `content.volumes` (VerifiedVolume) work end to end; the schema, CCN and CLI halves already shipped, this is the host + guest half. Client counterpart for compose binds: aleph-im/aleph-rs#383. Builds on #1169 (whitelist guest kernel).

## What changes

- **Agent** (`vprogram_launch.py`): attaches each volume as a (data, hash tree) EXTRA disk pair after the workload pair, in message order; writes a `{rootfs}.verified_volumes` sidecar (comma-joined roothashes) only when volumes are declared; fails closed on bad roothashes, more than 8 volumes, and runtimes whose cmdline template has no `{verified_volumes}` slot.
- **Daemon** (`snp_config_slice`): reads the sidecar and appends ` verified_volumes=h1,h2` after the workload token; strict 1..=8 lowercase-sha256 validation; absent sidecar leaves the cmdline byte-identical to today (parity with the CLI's template-token dropping).
- **Guest init** (`init-common.sh` `mount_verified_volumes`, wired in `init.sh` and `init-compose.sh`): anchored token validation, positional verity-open of device pairs (`vde/vdf`, `vdg/vdh`, ...), read-only `nodev,nosuid` mounts at `/volumes/<i>` over a tmpfs on the image-shipped `/volumes` dir (workload chroot for exec, platform chroot for compose). Every mismatch powers the VM off.
- **Runtime template**: `CMDLINE_TEMPLATE_EXEC_V1` gains ` verified_volumes={verified_volumes}` (new exec/compose manifests only; published 1.0 runtimes keep refusing volumes at the CLI and now at the agent too).
- **boot-smoke** phase 2: QEMU boot of the fib workload plus one host-built verity volume, asserting the init markers. Golden measurements re-seeded (initrd and compose rootfs changed).

Design: docs/plans/2026-07-08-confidential-vm-protocol-design.md (VerifiedVolume, section 8.1, decision 11). Plan: docs/plans/2026-08-28-vprogram-verified-volumes-implementation.md.

## Verification
Python and shell gates locally, boot-smoke both phases on the pre-fix-wave tree; Rust tests and the golden check run in CI (the local environment cannot run `cargo test`). Draft until CI is green; publishing exec-1.1/compose-1.1 runtimes and the aggregate update follow (plan Part 3).


odesenfans:
Deferred minors from the task and whole-branch reviews (none blocking; triaged as can-defer by the final review):

- `test_vprogram_launch.py`: no positive boundary test for exactly 8 volumes; `test_too_many_volumes_fails_closed` reuses one staged volume x9 (does not prove count-check-before-fetch with distinct refs); a mid-loop roothash failure leaves `get_existing_file` side effects for earlier volumes (mirrors the workload path).
- `lifecycle.rs`: six malformed-sidecar cases share one `#[test]`; a `verified_volumes` sidecar without a `workload_roothash` sidecar is only rejected by the guest, not the daemon.
- `nix/init-common.sh`: the shipped static busybox's ERE `{64}` interval support is exercised by boot-smoke's accept path only; a build-time assertion (like `checkedBusybox`) or a reject-path smoke would make it airtight. The `8` cap is a third copy of `MAX_VERIFIED_VOLUMES` (guest and daemon cannot import the schema).
- `nix/boot-smoke.sh`: the top-level and phase-1 `$log` temp files leak (local tool).

foxpatch-aleph (approved):
Solid, well-documented implementation with consistent fail-closed checks at every layer (schema, agent re-check, daemon cmdline gate, guest-init anchored validation). The positional (data, hash tree) disk contract is correctly mirrored between the agent's disk-append order and the guest's device-letter tables, and the daemon's sidecar validation matches the agent's writer exactly. All 43 touched Python tests pass locally; the busybox grep ERE intervals and mkfs-on-file assumptions in the smoke test were verified empirically. The findings below are non-blocking polish items.

**`src/aleph/vm/agent/vprogram_launch.py`** (line 266): Unlike the .roothash sidecar (cross-checked against the manifest) and the .workload_roothash sidecar (unconditionally overwritten), a verified_volumes sidecar shipped INSIDE the runtime bundle is left in place when content.volumes is empty. The daemon would then splice a verified_volumes token into the measured cmdline that the CLI never included, so every volume-less launch on that runtime fails measurement verification (and the guest powers off at wait_for_dev /dev/vde anyway). Fail-closed, but noisy and asymmetric: consider unlinking {rootfs}.verified_volumes when volume_roothashes is empty so the staged state always mirrors the message.

**`nix/boot-smoke.sh`** (line 33): run_phase reassigns the global log on every call, so once phase 2 starts the phase-1 mktemp file is orphaned (the EXIT trap only removes the latest $log). For a local tool this is cosmetic, but a `rm -f "$log"` before the reassignment would keep it tidy.

**`nix/init-common.sh`** (line 216): Nit: the 8 cap here and the two letter tables above duplicate MAX_VERIFIED_VOLUMES and the device-letter derivation in three places that must stay in sync with the schema. The FATAL message makes divergence loud rather than silent, so this is fine as-is — just flagging the coupling.

**`nix/boot-smoke.sh`** (line 103): Test-coverage gap: phase 2 exercises only the exec flavor, so the compose path (init-compose.sh's mount_verified_volumes /mnt/root and the compose-rootfs.nix /volumes mount-point) has no boot-level coverage — only the re-seeded golden measurement. Also, the run_phase failure message does not say which phase failed, which will matter now that there are two.

odesenfans:
Addressed two of the review's findings:

- 8096ebc1: a `{rootfs}.verified_volumes` file shipped inside the runtime bundle is now unlinked on volume-less launches, so the staged state always mirrors the message (regression test included).
- 1cb45f49: boot-smoke removes the previous phase's log before reassigning it and names the failing phase in the failure output.

The constant/letter-table triplication is accepted coupling (the guest and daemon cannot import the schema; divergence fails loudly). Compose-flavor boot coverage is deferred: either a cheap phase 3 here or the planned aleph-testnets e2e, pending a call on where it belongs.

odesenfans:
Decision: compose-flavor boot coverage stays out of boot-smoke; the aleph-testnets e2e (real compose workload + volume bind on the TEE host) will cover it end to end, podman half included.

foxpatch-aleph (approved):
No genuine bugs or security issues found. The three-layer fail-closed validation chain (agent → daemon → guest init) is thorough and consistent: roothashes are pattern-validated in Python, re-validated in Rust before splicing into the measured cmdline, and anchored-grep-validated in the guest before any device is touched. Byte-parity is preserved for volume-less messages (no sidecar written, stray bundle-shipped sidecars unlinked, daemon skips the token entirely). Positional device mapping (vde/vdf through vds/vdt) is correct for all 8 slots, and the tmpfs-over-/volumes approach cleanly handles the read-only verity root. Tests are comprehensive at every layer: 6 new Python tests, 3 new Rust tests, and a full boot-smoke phase 2 exercising a real two-volume dm-verity guest boot. All targeted tests pass locally. Remaining observations are minor.

**`src/aleph/vm/agent/vprogram_launch.py`** (line 243): The check verifies the template contains a {verified_volumes} slot but not that it appears after workload_roothash, so a runtime whose template reorders the tokens would boot with a different (but still self-consistent) cmdline than the CLI-computed measurement. This matches the pre-existing design for the other two slots and is acceptable, but worth noting for future hardening.

**`nix/boot-smoke.sh`** (line 1): Phase 2 asserts the volume-mount markers but does not read a file through the mounted volume to prove the data is actually readable by the workload. The mount options (ro,nodev,nosuid) and dm-verity verification are exercised, which is the important part; a content-read assertion would be a small robustness improvement.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 4902): The mixed &str/&String array literal relies on array-element LUB coercion (&String → &str via deref coercion), which is valid Rust but worth confirming compiles in CI since rustc was unavailable locally.

**`nix/init-common.sh`** (line 207): The grep -Eq pattern uses {64} which busybox grep supports (verified in the boot-smoke run where the valid path passes through the same regex), and the strict anchored shape-check happens before any word-splitting, which prevents the 'a,,b' re-indexing issue called out in the comment. Well-handled.

foxpatch-aleph (approved):
Solid, carefully fail-closed implementation. The agent-side staging (positional EXTRA disk pairs, {rootfs}.verified_volumes sidecar, slot/max/pattern refusal), the daemon-side cmdline splicing with strict 1..=8 lowercase-sha256 validation, and the guest verity-open of each device pair are all mutually consistent, and the byte-parity constraint for volume-less launches (no sidecar -> no token, matching the CLI template-token drop) is explicitly guarded by tests on both sides. I ran the full Python test file (44 passed), simulated the guest validation logic under real busybox sh against 11 malformed/valid shapes (all behave correctly, including the a,,b re-indexing hazard the anchored grep closes), and verified the vde/vdf..vds/vdt letter mapping is gap-free. No injection, mismeasurement, or fail-open path found; the Rust side could not be compiled locally but the tests read correctly against the existing helpers.

**`nix/init-common.sh`** (line 216): Minor: this 8 duplicates MAX_VERIFIED_VOLUMES (aleph_message, the Python agent import, and the new Rust const in lifecycle.rs), and the two device-letter lists at lines 230-231 encode the same cap. POSIX sh cannot import the constant, but a cross-reference comment would help a future bump update all of them. The letters do fail closed anyway (cut -f9 yields empty and wait_for_dev rejects /dev/vd), so this is purely a maintainability nit.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 2024): Minor asymmetry: on the opaque arm (tee.kernel_cmdline set) a stray {rootfs}.verified_volumes sidecar is silently ignored, whereas a {rootfs}.roothash sidecar is rejected as an ambiguous dual cmdline source. Ignoring is safe here (the cmdline is fully agent-specified, so no mismeasurement), but it might deserve a word in the comment for the next reader.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 2172): Note: the volume entries are validated lowercase-hex-only while the platform/workload roothash sidecars accept any ascii_hexdigit. Stricter is fine (the schema pins lowercase and the guest grep agrees); just flagging the intentional inconsistency in case it surprises anyone.

**`nix/boot-smoke.sh`** (line 46): Nit: expanding "${extra_drives[@]}" on an empty array (phase 1 passes no extra drives) is a hard error under set -u on bash < 4.4. Fine on any current NixOS bash; only relevant if this script is ever run on an old bash.

**`src/aleph/vm/agent/vprogram_launch.py`** (line 233): The PR description references docs/plans/2026-08-28-vprogram-verified-volumes-implementation.md and docs/plans/2026-07-08-confidential-vm-protocol-design.md, but neither exists in this branch (docs/plans/ only holds the compose design doc). If the plan doc was meant to be committed alongside the code, it is missing from the PR; otherwise ignore.

PR #1177, opened 2026-08-31
test(daemon): stop racing the live filesystem in the pooled-disk test

odesenfans:
`pooled_disk_dedups_same_filesystem_and_skips_dead_pools` compares two separate `statvfs` sweeps of `/` and asserts byte equality; any write landing between them fails it. It flaked twice in one day on #1175's CI (4 KiB drift both times: [run 1](https://github.com/aleph-im/aleph-vm/actions/runs/33189434490/job/98910841140), [rerun](https://github.com/aleph-im/aleph-vm/actions/runs/33189434490/job/98913996445)), and the `twice >= once` assertion carries the same race in the other direction.

Fix: inject the per-path probe into the dedup-and-sum core, test the logic against a fixed table (distinct devices sum, repeated device counts once, dead pool contributes zero), keep one race-free live smoke. `available_disk_bytes_pooled` keeps its signature and behavior.

Verified: `cargo test -p supervisor-daemon --lib host::` 4 pass; fmt + clippy (`--all-targets -D warnings`) clean.

### DIFF
--- a/rust/crates/supervisor-daemon/src/host.rs
+++ b/rust/crates/supervisor-daemon/src/host.rs
@@ -96,25 +96,44 @@ pub fn available_disk_bytes(path: &Path) -> Result<u64, DaemonError> {
 /// pools_disk_usage. Unreachable pools contribute zero instead of failing
 /// GetHostInfo: a dead disk must not take Health down.
 pub fn available_disk_bytes_pooled(paths: &[PathBuf]) -> u64 {
-    let mut seen_devices = std::collections::HashSet::new();
-    let mut free = 0u64;
-    for path in paths {
+    available_disk_bytes_pooled_with(paths, |path| {
         let device = match std::fs::metadata(path) {
             Ok(metadata) => metadata.dev(),
             Err(error) => {
                 tracing::warn!(path = %path.display(), %error, "volume pool inaccessible, ignoring");
-                continue;
+                return None;
             }
+        };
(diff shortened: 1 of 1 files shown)


foxpatch-aleph (approved):
Clean, behavior-preserving deflake. The dependency-injection split is correct: both subtle cases (metadata failure skipping device registration, statvfs failure still claiming its device with a zero contribution) carry over exactly from the old code, and the race the PR targets — comparing two live statvfs sweeps — is fully eliminated in favor of a fixed probe table plus a race-free live smoke test. The public API and its service.rs caller are untouched. Only nits: the dedup test passes the same path twice where the comment implies a second distinct path on the same device, and the empty-input assertion from the old test was dropped.

**`rust/crates/supervisor-daemon/src/host.rs`** (line 163): Nit: the comment says "b's device via a second path", but the test passes the same `PathBuf` twice. A distinct path string mapping to device 2 (e.g. `"/pool-b2" => Some((2, 40))` in the probe) would match the comment and the real two-pools-one-filesystem scenario the dedup exists for. Still exercises device-based dedup as written, so non-blocking.

**`rust/crates/supervisor-daemon/src/host.rs`** (line 172): The old test asserted the empty-input case (`available_disk_bytes_pooled(&[]) == 0`); that coverage was dropped. Trivially true today, but a one-line addition here would keep it.

PR #1178, opened 2026-08-31
fix(agent): SNP vCPU probe no longer requires nested virtualization

odesenfans:
Field report F1 (2026-08-31, community SNP CRN rollout): hosts running kvm_amd nested=0 (a legitimate hardening posture) advertise no SNP guest CPU models at all, because every QEMU EPYC model reports svm/npt/nrip-save unavailable and the probe filter required zero unavailable features. The operator had to re-enable nested virt to get a tee block, undoing his mitigation.

SNP guests never use nested virtualization, and the launch path passes -cpu without enforce (QEMU masks missing features with a warning), so the filter was strictly stricter than the launch. The filter now ignores the SVM feature family (CPUID 0x8000000A EDX plus the svm bit) when judging models; a genuinely missing feature still disqualifies.

Part of the 2.0.1 field-report fix series (see docs/plans/2026-08-31-snp-field-report-fixes-implementation.md on the follow-up branch). Stacked PR: od/snp-qemu-capability-gate builds on this.

### DIFF
--- a/src/aleph/vm/agent/vcpu_probe.py
+++ b/src/aleph/vm/agent/vcpu_probe.py
@@ -84,8 +84,10 @@ async def query_cpu_definitions() -> list[dict]:
 # guest never uses them, and the launch path passes -cpu without enforce, so
 # QEMU merely masks them with a warning. Disabling nested virt is a
 # legitimate host hardening posture and must not empty the advertisement.
-# The set is the SVM feature word (CPUID 0x8000_000A EDX) plus the svm bit
-# itself (0x8000_0001 ECX).
+# The set is every name QEMU gives to the SVM feature word (FEAT_SVM over
+# CPUID 0x8000_000A EDX in target/i386/cpu.c; gmet is on QEMU master and not
+# yet in a release) plus the svm bit itself (0x8000_0001 ECX): exactly the
+# features Linux KVM exposes only when kvm_amd runs with nested=1.
 NESTED_VIRT_FEATURES = frozenset(
     {
         "svm",
@@ -102,7 +104,7 @@ NESTED_VIRT_FEATURES = frozenset(
         "avic",
         "v-vmsave-vmload",
         "vgif",
-        "x2avic",
+        "gmet",
         "svme-addr-chk",
         "vnmi",
     }
--- a/src/aleph/vm/agent/vcpu_probe.py
+++ b/src/aleph/vm/agent/vcpu_probe.py
@@ -79,13 +79,46 @@ async def query_cpu_definitions() -> list[dict]:
             process.kill()
 
 
+# Features QEMU reports as unavailable when the host runs kvm_amd nested=0.
+# They exist only to run nested hypervisors inside the guest: an SEV-SNP
+# guest never uses them, and the launch path passes -cpu without enforce, so
+# QEMU merely masks them with a warning. Disabling nested virt is a
+# legitimate host hardening posture and must not empty the advertisement.
+# The set is the SVM feature word (CPUID 0x8000_000A EDX) plus the svm bit
+# itself (0x8000_0001 ECX).
+NESTED_VIRT_FEATURES = frozenset(
+    {
+        "svm",
+        "npt",
+        "lbrv",
+        "svm-lock",
+        "nrip-save",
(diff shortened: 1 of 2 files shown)


foxpatch-aleph (changes requested):
The core change is sound and I verified its premises: KVM really does zero the whole SVM feature word plus the svm bit when nested=0 (arch/x86/kvm/svm/svm.c:5522), the launch path really passes -cpu without enforce (rust/crates/aleph-tee/src/sev_snp/qemu.rs:63-66, so QEMU masks with a warning rather than failing), and the pinned launch measurement is insensitive to the masked bits (sevsnpmeasure hashes the CPUID page as zeros and uses only the CPU signature), so advertising models whose nested-virt features get masked does not break attestation parity. The set-difference filter logic is correct, backward compatible with a missing unavailable-features key, and both new tests pass. The blocking issue is that the hardcoded list does not match the QEMU SVM feature word the comment claims it mirrors: it is missing gmet (present in QEMU master's FEAT_SVM, enabled by EPYC-Milan-v4, EPYC-Genoa-v3 and EPYC-Turin-v2, and set by KVM behind the same if (nested) gate) and includes x2avic, a name no QEMU version ever reports. With every QEMU released so far (through 10.2) the code behaves correctly, but once a QEMU with gmet ships, those three models silently disappear from the advertisement on nested=0 hosts — the exact field-report failure mode recurring, with no error or failing test to catch it. The fix is a one-word addition (plus optionally dropping the phantom entry), so it is cheapest to do now.

**`src/aleph/vm/agent/vcpu_probe.py`** (line 89): This set is missing `gmet`. QEMU master's FEAT_SVM word (target/i386/cpu.c) names bit 17 `gmet`, and it is explicitly enabled by EPYC-Milan-v4, EPYC-Genoa-v3 and EPYC-Turin-v2; Linux KVM sets `X86_FEATURE_GMET` behind the same `if (nested)` gate as every other entry here. Once a node runs a QEMU that ships gmet, those models will report it in unavailable-features on a nested=0 host and silently drop out of the advertisement again — precisely the field bug this PR fixes. No released QEMU (through 10.2) has it yet, but since the comment states the set *is* the SVM feature word, adding `gmet` now is a one-line fix.

**`src/aleph/vm/agent/vcpu_probe.py`** (line 105): `x2avic` is not a QEMU feature name — no QEMU release, and not master, names any SVM-word feature that way (the FEAT_SVM word leaves that bit unnamed; `x2avic` is Linux KVM terminology). Harmless, but it suggests the list wasn't derived from the QEMU source the comment cites, and if QEMU ever does name the bit there is no guarantee it uses this spelling. Consider removing it, or noting its provenance.

**`tests/supervisor/test_vprogram_probe.py`** (line 42): Optional: since NESTED_VIRT_FEATURES is a hardcoded list that must track QEMU's FEAT_SVM word, a test asserting that every name in the set is ignored (e.g. one unavailable-features list covering the whole set) would guard against silent drift like a missing `gmet` when a new QEMU feature lands.

foxpatch-aleph (approved):
Correct and well-scoped fix, verified against both QEMU and kernel sources. I confirmed that NESTED_VIRT_FEATURES is exactly QEMU's FEAT_SVM feature word (all 16 names on master, plus svm itself from 0x80000001 ECX) and that Linux kvm_amd exposes every one of those features only when nested=1, so the field report's symptom is fully explained and per-feature ignoring is the right shape of fix. The filter still rejects models with any genuinely missing feature, handles missing/None unavailable-features, and the selection/advertisement path stays consistent via select_snp_vcpu_type. Tests are good and all 17 pass. One non-blocking nit: the gmet parenthetical in the comment is now stale.

**`src/aleph/vm/agent/vcpu_probe.py`** (line 88): The ignore list itself is accurate (I verified all 16 FEAT_SVM names plus svm against QEMU master), but this parenthetical is now stale: gmet shipped in QEMU v11.0.0 (it is absent from v10.0.x/v10.1.x but present in v11.0.0's cpu.c). Since QEMU 11.0 is released, consider rewording to just note that gmet is only reported by QEMU 11.0+ and that listing a name an older QEMU never reports is harmless. The inclusion stays correct either way.

PR #1179, opened 2026-08-31
feat(agent): truthful SNP capability advertisement (sev-snp-guest gate, operator surfacing)

odesenfans:
Field report F2: a host can pass every capability signal aleph-vm checks and still be unable to launch an SNP guest. Of the packaged targets only Debian 13 ships a QEMU with the sev-snp-guest object (>= 9.1); Ubuntu 24.04 has 8.2.2 and Debian 12 has 7.2, and nothing verified the object existed. Diagnosing this took the reporter and two node operators a day of custom host scripts.

Stacked on #1178. Four commits:

- The QMP probe now also runs qom-list-types and refuses to advertise SNP when the sev-snp-guest object is absent, carrying a human-readable reason (QEMU version included). An empty probe result is retried every 60s instead of being pinned for the process lifetime, so a live QEMU upgrade heals the advertisement without an agent restart (exactly the mid-test upgrade sequence from the report).
- /about/capability gains tee_unavailable_reason (only when SNP silicon is present), so the next person asks the node itself instead of writing host diagnostics. The scheduler-read /about/usage/system payload is unchanged.
- The agent logs a startup notice: INFO with the advertised models, or a WARNING naming exactly why confidential launches are disabled.
- docs/architecture/confidential.md documents the QEMU >= 9.1 host requirement.
- The QMP subprocess stream limit is raised to 1 MiB: qom-list-types replies already measure 43 KB against asyncio's 64 KiB default readline limit and grow with every QEMU release.

### DIFF
--- a/docs/architecture/confidential.md
+++ b/docs/architecture/confidential.md
@@ -105,6 +105,20 @@ in `MachineProperties` (`src/aleph/vm/agent/resources.py`,
 facts are different axes, and keeping them siblings leaves room for a future
 `tdx` or GPU-CC platform key without a schema break.
 
+### Host requirements for SEV-SNP
+
+SEV-SNP guest launches need QEMU >= 9.1, the first release with the
+`sev-snp-guest` QOM object the probe checks for. Of the packaged targets,
+only Debian 13 ships one; Ubuntu 24.04 (QEMU 8.2) and Debian 12 (QEMU 7.2)
+can serve plain SEV/SEV-ES but not SNP, whatever `query-cpu-definitions`
+reports. `kvm_amd nested=0` is fine, a legitimate hardening posture: the
+filter ignores nested-virt-only unavailable features since 2.0.1, so it no
+longer empties the advertisement. The probe retries every 60s after an empty
+result, so a live QEMU upgrade heals the advertisement without an agent
+restart. Check either end: `/about/capability`'s `tee_unavailable_reason`,
+or the agent's startup log line (`log_snp_launch_capability`,
+`src/aleph/vm/agent/supervisor.py`).
+
 ### The attestation stack
 
 Three crates, cleanly separated by role.
--- a/docs/architecture/confidential.md
+++ b/docs/architecture/confidential.md
@@ -109,9 +109,9 @@ facts are different axes, and keeping them siblings leaves room for a future
 
 SEV-SNP guest launches need QEMU >= 9.1, the first release with the
 `sev-snp-guest` QOM object the probe checks for. Of the packaged targets,
-only Debian 13 ships one; Ubuntu 24.04 (QEMU 8.2) and Debian 12 (QEMU 7.2)
-can serve plain SEV/SEV-ES but not SNP, whatever `query-cpu-definitions`
-reports. `kvm_amd nested=0` is fine, a legitimate hardening posture: the
+Debian 13 (QEMU 9.x) and Ubuntu 26.04 (QEMU 10.2) ship one; Ubuntu 24.04
+(QEMU 8.2) and Debian 12 (QEMU 7.2) can serve plain SEV/SEV-ES but not SNP,
+whatever `query-cpu-definitions` reports. `kvm_amd nested=0` is fine, a legitimate hardening posture: the
 filter ignores nested-virt-only unavailable features since 2.0.1, so it no
 longer empties the advertisement. The probe retries every 60s after an empty
 result, so a live QEMU upgrade heals the advertisement without an agent
--- a/src/aleph/vm/agent/resources.py
+++ b/src/aleph/vm/agent/resources.py
@@ -9,7 +9,10 @@ from pydantic import BaseModel, Field
 
 from aleph.vm.agent.aggregate import get_compatible_gpus, update_aggregate_settings
 from aleph.vm.agent.machine import get_cpu_info, get_hardware_info, get_memory_info
-from aleph.vm.agent.vcpu_probe import get_supported_snp_vcpu_types
+from aleph.vm.agent.vcpu_probe import (
+    get_snp_launch_capability,
+    get_supported_snp_vcpu_types,
+)
 from aleph.vm.conf import settings
 from aleph.vm.resources import GpuDevice
 from aleph.vm.sevclient import SevClient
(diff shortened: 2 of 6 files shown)


foxpatch-aleph (approved):
Solid, well-tested change that closes a real operational gap: hosts can pass every capability check yet be unable to launch SNP guests because their QEMU lacks the sev-snp-guest QOM object. The gate is enforced at the launch path (select_snp_vcpu_type consumes the same probe), empty results are retried so a live QEMU upgrade heals the advertisement, and the failure mode of the new 1 MiB stream limit is a retried probe failure rather than a crash. The scheduler-visible /about/usage/system payload is untouched as claimed. All 22 tests in the PR's suite pass; the only failures I saw elsewhere are pre-existing sandbox filesystem issues unrelated to this change. Remaining comments are non-blocking polish: the reason string publishes the exact QEMU version on a public endpoint, and persistent-empty hosts will emit a WARNING every 60 seconds indefinitely.

**`src/aleph/vm/agent/vcpu_probe.py`** (line 167): The reason string embeds the exact QEMU version (e.g. "QEMU (8.2.2)") and is served on the public, CORS-open /about/capability endpoint. The marginal risk is low — a guest can read the QEMU version from SMBIOS anyway — but an operator may not expect their exact hypervisor build to be broadcast to anyone who can reach the node. Consider a version-independent message, or gating the version to the startup log line where the audience is unambiguously the operator.

**`src/aleph/vm/agent/vcpu_probe.py`** (line 258): With the new retry-on-empty behavior, a host with SNP silicon and an old QEMU re-probes every 60s and logs this WARNING on each attempt — indefinitely, which is a lot of log volume for a condition that persists for months. Consider warning only on the first empty result (e.g. a flag on _ProbeCache) and logging repeats at INFO, letting log_snp_launch_capability remain the loud, one-time notice.

**`src/aleph/vm/agent/resources.py`** (line 286): This inlines the TeeProperties construction that _get_tee_properties already encapsulates. If a second platform (tdx) is ever added to TeeProperties, this copy can silently drift from what /about/usage/system advertises. A tiny shared helper (or having get_machine_capability call _get_tee_properties) would keep the two views identical.

**`src/aleph/vm/agent/vcpu_probe.py`** (line 91): Minor: `if version_info:` only checks the dict is non-empty, so a hypothetical greeting with version.qemu == {} would render "None.None.None". Checking for the presence of the major/minor/micro keys instead would be exact. Harmless in practice since the greeting comes from QEMU itself.

foxpatch-aleph (approved):
Correct, well-documented, and well-tested. The QMP greeting version parsing is right (verified against the QEMU spec: version lives under the QMP key), the sev-snp-guest QOM gate is a sound build-capability check complementary to the kernel check, and the probe cache state machine handles the empty-result/exception/success transitions correctly under its lock. The 1 MiB stream-limit bump is justified. Only non-blocking observations remain: the new tee_unavailable_reason field publishes the QEMU version on the unauthenticated /about/capability endpoint (consistent with the CPU/memory data already there, but worth a conscious sign-off), the startup hook can delay agent startup by up to PROBE_TIMEOUT_SECONDS rather than just log, and a few branches (no-launchable-EPYC-model reason, INFO path of the startup notice) lack test coverage.

**`src/aleph/vm/agent/resources.py`** (line 145): Note that about_capability() serves /about/capability publicly (no token check), so tee_unavailable_reason — including the exact QEMU version — becomes world-readable on SNP-capable nodes. That is mild CVE-selection disclosure, though consistent with the CPU model/frequency, memory and GPU inventory this endpoint already publishes, and the field is only populated on SNP-silicon hosts. Fine to merge as-is; just confirming the disclosure is deliberate.

**`src/aleph/vm/agent/supervisor.py`** (line 493): on_startup hooks run sequentially, so on an SNP host where qemu-system-x86_64 hangs (or is absent but the path resolves), this hook can delay agent startup by up to PROBE_TIMEOUT_SECONDS (15s). It can never fail startup — good — but if startup latency matters for the fleet, consider a create_task wrapper so the probe logs once it resolves rather than gating the readiness handshake. Not blocking.

**`src/aleph/vm/agent/resources.py`** (line 285): Minor: this inlines what _get_tee_properties() already does. Justified since you need both the vcpu types and the reason from a single probe call (and the probe result is cached), but if _get_tee_properties ever grows additional TEE platforms this second construction site will need to track it.

**`tests/supervisor/test_vprogram_probe.py`** (line 320): Good coverage overall. Two small gaps worth considering: (1) no test for the second branch of _capability_from_facts — sev-snp-guest present but every EPYC model genuinely unlaunchable — so its message string is unverified; (2) the INFO branch of log_snp_launch_capability (capability available at startup) is untested, and test_capability_carries_unavailability_reason mocks get_snp_launch_capability wholesale, so the wiring of a real probe result through get_machine_capability to tee_unavailable_reason is only covered on the probe side.

foxpatch-aleph (approved):
Solid, well-tested change. Gating SNP advertisement on the sev-snp-guest QOM object is the right mechanism (the type is registered unconditionally by target/i386/sev.c in QEMU >= 9.1, so qom-list-types is a reliable presence check), the empty-vs-failed cache distinction is handled correctly, the 1 MiB stream limit is a reasonable response to the 43 KB qom-list-types replies, and test coverage of the new behavior (QOM gating, live-upgrade healing, reason surfacing, startup notice) is thorough. All 23 probe tests pass; the test_views.py failures I reproduced on main and are pre-existing environment issues. No security concerns. Remaining points are non-blocking nits.

**`src/aleph/vm/agent/supervisor.py`** (line 441): This on_startup hook runs the QMP probe synchronously before the site starts serving, so a hung qemu delays agent startup by up to PROBE_TIMEOUT_SECONDS (15s). On a healthy SNP host it is ~100ms so this is fine, but consider create_task_log_exceptions (already used elsewhere here) if you would rather startup never block on it.

**`src/aleph/vm/agent/vcpu_probe.py`** (line 260): On a host with SNP silicon and an old QEMU this WARNING fires on every 60s retry-window expiry, forever (~1440 lines/day). Intentional nagging is defensible, but two related gaps: (1) there is no counter-signal when a live QEMU upgrade heals the advertisement — an operator watching logs sees the startup DISABLED warning but never an INFO that the node is now advertising; (2) if the noise is a concern, logging only on the empty->non-empty transition would cut it to one line per state change.

**`src/aleph/vm/agent/vcpu_probe.py`** (line 92): Nit: if the greeting's version dict were present but missing keys this renders "None.None.None". The `if version_info:` guard only checks non-emptiness. Practically impossible with real QEMU, but `.get(k, "?")` would be more defensive.

**`docs/architecture/confidential.md`** (line 112): The doc and the operator-facing reason string name Ubuntu 26.04 as a packaged target, but packaging/ has no ubuntu-26.04.dockerfile yet (commit message says PR 1181) — until that merges, an operator following the string to packaging/ finds only Debian 13. Also the "can serve plain SEV/SEV-ES but not SNP" list omits Ubuntu 22.04 (QEMU 6.2), which is a packaged target.

**`src/aleph/vm/agent/vcpu_probe.py`** (line 58): The QMP greeting-to-qemu_version parsing here is the one piece of the new logic without test coverage (the whole _converse loop is untested, as its predecessor was). A small fixture feeding a canned greeting line through a fake stdin/stdout would pin the version parsing; not blocking.

PR #1180, opened 2026-08-31
fix: SNP launches fail loudly instead of wedging invisibly

odesenfans:
Field report F3: a V-PROGRAM on a community CRN sat in awaiting_confidential_init forever, with the owner seeing only "VM is starting" and no error surfaced to anyone. Root cause analysis showed two layers, both fixed here:

- The Python supervisor only implements the SEV session flow: for any spec with tee set it creates the VM but never starts the controller, waiting for an owner initialization that never comes for SNP. It now rejects SEV-SNP specs up front with a VmSetupError naming the remedy (ALEPH_VM_SUPERVISOR_IMPL=rust), before touching the pool. The released 2.0.0 debs default to the Python supervisor (the rust flip in #1171 merged 4.5 hours after the tag was cut), so every stock 2.0.0 community node hits exactly this; 2.0.1 ships the rust default, and this guard keeps an operator-pinned python impl honest.

- RUNNING only proves the controller unit is active: an SNP guest can wedge inside firmware while QEMU lives on. Confidential creates (V-PROGRAM and SNP instance) now gate on the guest's RA-TLS attestation endpoint accepting a TCP connection within a bounded window (90s, module constants), failing the create loudly (teardown + scheduler retry) instead of leaving a zombie only the owner can watch. A healthy guest pays one get_vm plus a millisecond TCP connect.

Ops notes from review, for the release checklist: the scheduler's allocation-client HTTP timeout should exceed create + 120s (RUNNING wait) + 95s (gate worst case); watch the demo CRN's first V-PROGRAM launches for RA-TLS bind times approaching the 90s window.

Part of the 2.0.1 field-report fix series (with #1178 and #1179).

### DIFF
--- a/src/aleph/vm/agent/run.py
+++ b/src/aleph/vm/agent/run.py
@@ -294,6 +294,54 @@ async def _wait_until_running(
         await asyncio.sleep(interval)
 
 
+_ATTEST_READY_TIMEOUT_SECONDS = 90.0
+_ATTEST_READY_POLL_INTERVAL_SECONDS = 2.0
+
+
+async def _wait_until_attest_endpoint_listens(
+    supervisor: Supervisor,
+    vm_id: VmId,
+    attest_port: int,
+    *,
+    timeout: float | None = None,
+    interval: float | None = None,
+) -> None:
+    """Gate a confidential create on the guest's RA-TLS attestation service
+    accepting a TCP connection.
+
+    RUNNING only proves the controller unit is active: an SNP launch can wedge
+    inside firmware while QEMU lives on, and the attestation endpoint is the
+    guest's sole consumer-facing contract. A guest that never listens is a
+    failed launch and must fail the create loudly (the caller tears down and
+    the scheduler retries), not survive as a zombie the owner can only watch.
+    """
+    if timeout is None:
+        timeout = _ATTEST_READY_TIMEOUT_SECONDS
+    if interval is None:
+        interval = _ATTEST_READY_POLL_INTERVAL_SECONDS
+    info = await supervisor.get_vm(vm_id)
+    address = info.ipv4.address
+    if not address:
+        msg = f"VM {vm_id} has no IPv4 assignment to reach its attestation endpoint"
+        raise VmStartupError(msg)
+    deadline = asyncio.get_running_loop().time() + timeout
+    while True:
+        try:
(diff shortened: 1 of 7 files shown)


foxpatch-aleph (approved):
Both halves of the fix are correct and well-tested. The LocalSupervisor SNP guard rejects before any pool interaction (VmSetupError passes through translating_errors unmodified and the env-var remedy matches the packaging launcher scripts), and the new attestation gate is correctly wired into both confidential create paths inside the existing teardown handlers, so a wedged guest now fails the create loudly. The gate's poll loop is sound: at least one probe always runs, the deadline is checked after each failure, and OSError subclasses (refused/reset/DNS) retry while success closes the probe connection. The Python-supervisor rejection tests and the four new gate unit tests all pass, as do the wiring tests in test_snp_instance_run.py and test_vprogram_port_map.py; the only failures in the touched suites are pre-existing environment issues unrelated to this change (tests needing write access to /var/lib/aleph/vm). Remaining comments are minor robustness/documentation nits, none blocking.

**`src/aleph/vm/agent/run.py`** (line 322): Minor: the IPv4 is resolved once, outside the retry loop, so a first-poll empty address fails immediately with no retry. Since the gate only runs after _wait_until_running (a tap must exist for RUNNING) this is low-risk, but a one-line comment noting the address is expected to be present post-RUNNING would help future readers. If an empty-address retry is ever desired, it would also make the error path more forgiving.

**`src/aleph/vm/agent/run.py`** (line 330): Note for the ops checklist in the PR description: with the 5s per-probe wait_for plus the 2s interval, worst-case wall time is ~timeout + 5s + interval (~97s), marginally past the stated 95s. Also, after a failed probe the loop sleeps the interval and attempts one more connection past the deadline before re-checking, so the budget is soft at the boundary. Only relevant to scheduler allocation-client timeout tuning.

**`src/aleph/vm/agent/run.py`** (line 345): Non-OSError exceptions from writer.wait_closed() would escape the gate and tear down an otherwise-successful create. The realistic case (ConnectionResetError from an RST right after accept) is an OSError subclass and is caught, so this is theoretical, but catching Exception (or awaiting wait_closed defensively) would make the success path airtight.

**`tests/supervisor/test_vprogram_launch.py`** (line 521): Patching the global asyncio.open_connection couples these tests to the implementation calling asyncio.open_connection by attribute at call time; a switch to e.g. loop.open_connection would silently still pass. Acceptable for a unit test, just worth knowing.

PR #1181, opened 2026-08-31
feat(packaging): build a deb for Ubuntu 26.04 (resolute)

odesenfans:
Ubuntu 26.04 LTS ships QEMU 10.2.1, making it the second packaged target (after Debian 13) whose stock QEMU can launch SEV-SNP guests: the sev-snp-guest QOM object gate added in #1179 passes on it out of the box. This gives SNP-capable community CRNs an Ubuntu option instead of requiring Debian 13.

- packaging/ubuntu-26.04.dockerfile: identical to the other targets except the base image
- Makefile: all-podman-ubuntu-2604, requirements-ubuntu-26.04, repository-resolute, wired into the repositories and all-podman aggregates
- Both CI matrices (PR builds and release-on-tag) build and attach aleph-vm.ubuntu-26.04.deb
- README install snippet, flagged as the recommended Ubuntu for SEV-SNP hosts
- Drive-by fix: repositories/noble/conf/distributions was never committed although the repository-noble target references it, so the aggregate repositories target could not run; added it

Note for the repo settings: if the new build should gate merges like the others, add "Build ubuntu-26.04 Package" to the required status checks.

https://claude.ai/code/session_019H6oFfJpeqtZ34gWQqxuqY

### DIFF
--- a/.github/workflows/build-deb-package-and-integration-tests.yml
+++ b/.github/workflows/build-deb-package-and-integration-tests.yml
@@ -19,7 +19,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: ["debian-12", "debian-13", "ubuntu-22.04", "ubuntu-24.04"]
+        os: ["debian-12", "debian-13", "ubuntu-22.04", "ubuntu-24.04", "ubuntu-26.04"]
         include:
           - os: "debian-12"
             make_target: "all-podman-debian-12"
@@ -33,6 +33,9 @@ jobs:
           - os: "ubuntu-24.04"
             make_target: "all-podman-ubuntu-2404"
             artifact_name: "aleph-vm.ubuntu-24.04.deb"
+          - os: "ubuntu-26.04"
+            make_target: "all-podman-ubuntu-2604"
+            artifact_name: "aleph-vm.ubuntu-26.04.deb"
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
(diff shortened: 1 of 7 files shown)


foxpatch-aleph (approved):
Clean, well-scoped PR that faithfully follows the existing packaging patterns. I verified the new dockerfile is byte-identical to ubuntu-24.04.dockerfile except the base image, that the rustup default toolchain (1.90) matches rust/rust-toolchain.toml, that both CI matrices are correctly wired with make_target/artifact_name matching the Makefile, and that the drive-by fix is genuine: repositories/noble/conf/distributions is indeed absent on main while repository-noble is referenced by the aggregate repositories target, so that target could not have run before. The new resolute reprepro conf matches the trixie style. Two non-blocking observations below.

**`README.md`** (line 97): Minor: this is the only snippet pinning `2.0.1` while the other examples use `1.2.0` and the section intro says "Replace `1.2.0` with the latest release version." Consider matching the surrounding examples for consistency. Also note this URL only resolves once a release is tagged after this merge, since `aleph-vm.ubuntu-26.04.deb` doesn't exist in any prior release.

**`.github/workflows/build-deb-package-and-integration-tests.yml`** (line 143): Ubuntu 26.04 is added to the build matrices but not to the `run_on_droplet` end-to-end matrix, so the new deb is content-verified (dpkg --contents) but never installed and started on a real 26.04 host. If DigitalOcean has an ubuntu-26-04-x64 image, adding an entry here would exercise the actual install path (systemd units, dependency resolution against 26.04's system Python) the way Debian 13 gets. If the image isn't available yet, this can land as-is.

Issue #160, opened 2022-03-28
Upgrade guest kernel from 4.20 to 5.10


hoh:
See [Firecracker's Kernel Support Policy](https://github.com/firecracker-microvm/firecracker/blob/v1.0.0/docs/kernel-policy.md).

We also don't have the kernel build tools as code in this repository.

Comparative benchmarks would be useful to see if there is a performance gain/penalty.

hoh:
This was done.

PR #1183, opened 2026-09-01
build: pin released aleph-message 1.3.1 instead of the git ref

odesenfans:
The `MERGE GATE` git pin (`4ad261ff`, the pre-squash V-PROGRAM schema branch) is now covered by the [1.3.1 release](https://pypi.org/project/aleph-message/1.3.1/) on PyPI, so aleph-vm can use the same released version as pyaleph (aleph-im/pyaleph#1261).

The only delta between the pinned tree and 1.3.1 is aleph-im/aleph-message#160 (serde-strict V-PROGRAM scalars), which tightens CCN-side validation and does not affect the messages aleph-vm itself produces or parses.

Verified locally with `aleph-message==1.3.1` installed: all 152 tests in `tests/supervisor/test_snp_*`, `tests/supervisor/test_vprogram*` and `tests/vprogram/` pass.

https://claude.ai/code/session_01QkaSkidXtLL63prnqo7NS5

### DIFF
--- a/packaging/Makefile
+++ b/packaging/Makefile
@@ -42,10 +42,9 @@ debian-package-code:
 	python3 -m venv build_venv
 	build_venv/bin/pip install   --progress-bar off --upgrade pip setuptools wheel
 	# Versions below MUST mirror pyproject.toml so the .deb ships what the code is tested against.
-	# MERGE GATE: aleph-message is git-pinned to the V-PROGRAM schema ref, swap for a release with pyproject.toml.
 	# protobuf is pinned to 5.29.6: 5.29.0 has a compilation issue; 6.x needs grpcio/tools 1.71+.
 	# legacy-cgi is packaging-only: it backfills the `cgi` module removed from the Python 3.13+ stdlib.
-	build_venv/bin/pip install --no-cache-dir  --progress-bar off --target ./aleph-vm/opt/aleph-vm/ 'aleph-message @ git+https://github.com/aleph-im/aleph-message@4ad261ff25f595a64589d49bf36f514c402e7c5e' 'eth-account~=0.10' 'sentry-sdk==2.8' 'qmp==1.1' 'aleph-superfluid~=0.2.1' 'sqlalchemy[asyncio]>=2' 'aiosqlite==0.19' 'alembic==1.13.1' 'aiohttp-cors~=0.7.0' 'pydantic-settings~=2.6.1' 'pyroute2==0.7.12' 'python-cpuid==0.1.1' 'solathon==1.0.7' 'protobuf==5.29.6' 'grpcio>=1.70,<1.71' 'redis>=4.2' 'legacy-cgi>=1.0'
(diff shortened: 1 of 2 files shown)


foxpatch-aleph (approved):
Verified the bump: aleph-message 1.3.1 exists on PyPI and the wheel diff against the pinned tree (4ad261ff) is exactly PR #160 (serde-strict V-PROGRAM scalars plus tests). The PR's 152-test selection matches the claim; results are identical between the pinned git-ref build and 1.3.1, with the few failures being pre-existing sandbox environment issues unrelated to the bump. The tightening is ecosystem-consistent with pyaleph#1261 and handled gracefully in aleph-vm's parse paths. One minor note: the parse-side effect of the tightened validation is slightly broader than the PR claim states.

**`pyproject.toml`** (line 38): The delta claim slightly oversimplifies the parse side: aleph-vm parses V-PROGRAM messages via `parse_message` (tasks.py:157, storage.py:232 -> VerifiableProgramContent), so poisoned raw dicts with coerced scalars now raise ValidationError instead of being coerced. Behavior is identical for serde-decodable messages, the WS path catches ValidationError gracefully, and the tightening matches pyaleph#1261, so the bump is sound.

foxpatch-aleph (changes requested):
The pyproject.toml change itself is correct and its claims verify: 1.3.1 is on PyPI, the pinned tree and 1.3.1 differ only by aleph-message#160 (confirmed by diffing the installed pin against the 1.3.1 wheel), and the SNP/V-PROGRAM test suites behave identically under the pin and 1.3.1 (the only failures in both runs are sandbox /var/lib/aleph permission artifacts, with zero pydantic validation regressions). However, the PR is incomplete: packaging/Makefile:48 still installs aleph-message from the git ref 4ad261ff and packaging/Makefile:45 still carries the MERGE GATE comment, even though line 44 requires those versions to mirror pyproject.toml and pyproject.toml:31 instructs updating that Makefile on dependency changes. As it stands the .deb would ship the git ref while CI tests 1.3.1 — the exact divergence this PR is meant to close.

**`packaging/Makefile`** (line 48): This PR swaps the pin in pyproject.toml but leaves the mirrored Debian install here on the git ref (`aleph-message @ git+...@4ad261ff...`), and line 45 still carries the MERGE GATE comment this PR is supposed to close. Line 44 says these versions MUST mirror pyproject.toml so the .deb ships what the code is tested against, and pyproject.toml's own dependency block instructs updating this Makefile on dependency changes. Please install `aleph-message==1.3.1` here and drop the gate comment at line 45 so the .deb (and the `pip freeze` from extract_requirements.sh) matches CI.

odesenfans:
Re the review: both points addressed.

- **packaging/Makefile** — good catch, fixed in ea56ba19: the Debian install now pins `aleph-message==1.3.1` and the MERGE GATE comment is gone, so the .deb ships exactly what CI tests.
- **parse-side note on pyproject.toml** — agreed and intended: aleph-vm's `parse_message` paths now reject coerced scalars with ValidationError instead of coercing, which is the same serde-parity tightening as pyaleph#1261; the WS path already handles ValidationError gracefully, so no code change.

foxpatch-aleph (approved):
Clean, well-scoped dependency pin swap. I verified the two files stay consistent with each other, that no MERGE GATE comment or reference to the git ref 4ad261ff remains anywhere in the tree, and — by downloading the 1.3.1 wheel from PyPI and diffing it against the locally installed git-pinned tree — that the only delta is the strict-scalar change in models/execution/vprogram.py plus its tests, exactly as the PR description claims. I ran the full unit suite against both versions with byte-identical results (the pre-existing failures/errors reproduce identically under the old pin and are environment artifacts, not regressions), and confirmed pyaleph's main also pins 1.3.1, so CCN/VM validation stays in lockstep. The pyproject.toml comment realignment passes pyproject-fmt --check.

**`pyproject.toml`** (line 38): Out of scope for this PR, but note that `runtimes/aleph-debian-12-python/create_disk_image.sh:40` and `docker/vm_supervisor-dev.dockerfile:22` still install `aleph-message~=1.0.x` in their own contexts (guest runtime image / dev container). Worth a follow-up bump if those consumers ever need V-PROGRAM-aware models.

PR #1184, opened 2026-09-01
fix(nix): SEV-SNP guests need X86_PAT; the minimal kernel hung before console

odesenfans:
The whitelist guest kernel (#1169) never actually booted as an SEV-SNP guest: it hangs before the first console line. The VM stays "running", so the 90s RA-TLS gate (#1180) tears it down and the scheduler retries forever. Found by the aleph-testnets release-stack run ([aleph-testnets#38](https://github.com/aleph-im/aleph-testnets/pull/38)) — the first time the minimal kernel met real SNP hardware; boot-smoke is deliberately non-SNP, and the fragment was authored without an SNP host to test on (#1169 noted "SNP launch not verified locally").

**Bisect** (TEE server, EPYC 8024P, QEMU 10.2, direct kernel boot with kernel-hashes=on, all other artifacts equal):
- 1.0-runtime kernel (distro config): boots · minimal kernel: silent hang — with either OVMF, either initrd
- fragment + {EARLY_PRINTK, MTRR, X86_PAT, EFI_DXE_MEM_ATTRIBUTES, X86_MCE_AMD, EFI_HANDOVER_PROTOCOL}: **boots**
- fragment + {MTRR, X86_PAT}: **boots**
- fragment + {MTRR}: silent hang

Minimal fix: `CONFIG_X86_PAT=y` plus `CONFIG_MTRR=y` (PAT's Kconfig dependency). The same kernel boots fine as a non-SNP guest, which is why boot-smoke never caught it.

Golden measurements re-seeded (kernel change moves all four). Validated: boot-smoke phases 1+2 locally, and a real SNP boot of the rebuilt exec image on the TEE server reaching userspace (verity root mounted, init ran to its fail-closed poweroff on the dummy workload roothash, as expected).

Follow-up (not this PR): rebuild + republish the V-PROGRAM runtimes (the 2026.08.31 "1.1" bundles on mainnet carry the broken kernel; they were never added to the vm-images aggregate, so nothing references them), and consider an SNP boot-smoke tier that runs on the TEE server.

https://claude.ai/code/session_01QkaSkidXtLL63prnqo7NS5


foxpatch-aleph (approved):
Minimal, correct fix. Verified against the pinned kernel's actual Kconfig that X86_PAT depends on MTRR and is a prompted symbol under EXPERT (so allnoconfig answers it 'no', explaining both the original silent hang and why both lines are load-bearing); the comment's claims and the bisect are internally consistent. Both lines are enforced at build time by kernel.nix's checkPhase, the golden-measurement re-seed is complete (all four values) and CI-gated, and the comment documents the non-obvious rationale in the fragment's house style. One non-blocking nit: the comment could note why CONFIG_X86_PAT=y needs an explicit line even though upstream declares def_bool y (EXPERT=y exposes the prompt that allnoconfig answers 'no').

**`nix/kernel-config.fragment`** (line 68): Non-blocking: worth one extra sentence noting that the explicit `CONFIG_X86_PAT=y` line is load-bearing despite the upstream `def_bool y` — with `CONFIG_EXPERT=y` (line 88) X86_PAT becomes a prompted symbol, so `allnoconfig` answers it 'no'. A future reader could otherwise assume the line is redundant, or silently rely on it after an EXPERT change. Alternatively, that subtlety could go in the comment block added above.

odesenfans:
Re the review nit: added in ef5661be — the comment block now notes that `EXPERT=y` exposes X86_PAT's prompt, so `allnoconfig` answers it 'no' and the explicit line is load-bearing despite upstream's `def_bool y`. Comment-only change: `check-golden-measurements.sh` verified the generated .config and all four golden measurements are unchanged.

foxpatch-aleph (approved):
Minimal, correct fix with exemplary documentation. The two-line kernel config addition is verifiable against upstream Kconfig: X86_PAT depends on MTRR, and with EXPERT=y allnoconfig answers X86_PAT's exposed prompt 'no', so both lines are genuinely required — the comment at kernel-config.fragment:61-68 accurately records why. The whitelist is self-checking (kernel.nix verifies every fragment line against the generated .config, so a silently dropped option fails the build), and all four golden measurements are re-seeded as required, with the golden-measurements CI workflow triggering on nix/** to verify them. The empirical bisect methodology, the honesty about boot-smoke's non-SNP blind spot, and the explicitly deferred follow-ups (runtime republish, SNP smoke tier) all make this a trustworthy change.

**`nix/kernel-config.fragment`** (line 69): Minor placement nit: these are platform memory-attribute features, not confidential-computing ones — they might read more naturally in the platform section. Keeping them here under the SNP-boot comment is defensible, so not worth churn. Separately, FYI: CONFIG_MTRR=y registers the /dev/mtrr misc device (writes are CAP_SYS_ADMIN-gated), which is a negligible, unavoidable surface given X86_PAT depends on MTRR.

Issue #391, opened 2023-07-20
devmapper command fails when setting up a specific instance type


odesenfans:
Scheduling a specific instance on a CRN results in a 500. The instance is not started properly.

Instance:

```
{                                                                                                                                                                      
    "chain": "ETH",                                                                                                                                                                                        
    "channel": "ALEPH-CLOUDSOLUTIONS",                                                                                                                                                                     
    "confirmations": [],                                                                                                                                                                                   
    "confirmed": false,                                                                                                                                                                                    
    "content": {                                                                                                                                                                                           
        "address": "0xf379B2F8c29EeDc1fC7d189240d0aD7c9164aF64",                                                                                                                                           
        "allow_amend": false,                                                                                                                                                                              
        "environment": {                                                                                                                                                                                   
            "aleph_api": true,                                                                                                                                                                             
            "internet": true,                                                                                                                                                                              
            "reproducible": false,                                                                                                                                                                         
            "shared_cache": false                                                                                                                                                                          
        },                                                                                                                                                                                                 
        "metadata": {                                                                                                                                                                                      
            "name": "MyAwAZ"                                                                                                                                                                               
        },                                                                                                                                                                                                 
        "resources": {                                                                                                                                                                                     
            "memory": 8000,                                                                                                                                                                                
            "seconds": 30,                                                                                                                                                                                 
            "vcpus": 4                                                                                                                                                                                     
        },                                                                                                                                                                                                 
        "rootfs": {                                                                                                                                                                                        
            "parent": {                                                                                                                                                                                    
                "ref": "77fef271aa6ff9825efa3186ca2e715d19e7108279b817201c69c34cedc74c27",                                                                                                                 
                "use_latest": true                                                                                                                                                                         
            },                                                                                                                                                                                             
            "persistence": "host",                                                                                                                                                                         
            "size_mib": 80000                                                                                                                                                                              
        },                                                                                                                                                                                                 
        "time": 1689843819.372,                                                                                                                                                                            
        "volumes": [                                                                                                                                                                                       
            {                                                                                                                                                                                              
                "mount": "/mount/opt",                                                                                                                                                                     
                "name": "MyVolume",                                                                                                                                                                        
                "persistence": "host",                                                                                                                                                                     
                "size_mib": 64000
            }
        ]
    },
[...]

PR #1185, opened 2026-09-02
feat(agent): aleph-message 1.4.0, reject TDX instances before any I/O

odesenfans:
aleph-vm side of increment 1 of the Intel TDX design (aleph-message #161 / aleph-rs #391 / pyaleph #1266). Closes the schema increment.

- **`aleph-message==1.4.0`** (was 1.3.1; the delta is exactly the TDX schema PR). The supervisor can now parse a `mode: tdx` instance message instead of failing validation.
- **Clean rejection instead of a crash.** A tdx instance passed `is_snp_instance` as false and fell into `build_create_vm_spec`, which resolved `trusted_execution.firmware`; measured modes forbid a firmware ref, so it crashed on `get_existing_file(None)` with an error hiding the real cause. The plain path now rejects any measured mode it cannot build, naming the mode: `"tdx instances are not supported by this CRN"`.
- **Guards hoisted before the download.** The docstring promised validation before any I/O, but the SNP misrouting guard sat after `download_all()`; a rejected message would first download its rootfs. Both guards (SNP misrouting, unsupported measured mode) now run in the pre-I/O validation block, and both gained tests asserting the downloader never runs.

No launch-path change: SNP instances still route to `build_snp_instance_spec`, SEV instances still build here, V-PROGRAM stays schema-locked to `sev_snp`. The TDX launch path arrives with the verifier increments.

Gates: `tests/supervisor/` translate + SNP + routing + vprogram suites against the 1.4.0 package (106 passed), ruff format, isort, mypy.

### DIFF
--- a/packaging/Makefile
+++ b/packaging/Makefile
@@ -39,7 +39,7 @@ debian-package-code:
 	# Versions below MUST mirror pyproject.toml so the .deb ships what the code is tested against.
 	# protobuf is pinned to 5.29.6: 5.29.0 has a compilation issue; 6.x needs grpcio/tools 1.71+.
 	# legacy-cgi is packaging-only: it backfills the `cgi` module removed from the Python 3.13+ stdlib.
-	build_venv/bin/pip install --no-cache-dir  --progress-bar off --target ./aleph-vm/opt/aleph-vm/ 'aleph-message==1.3.1' 'eth-account~=0.10' 'sentry-sdk==2.8' 'aleph-superfluid~=0.2.1' 'sqlalchemy[asyncio]>=2' 'aiosqlite==0.19' 'alembic==1.13.1' 'aiohttp-cors~=0.7.0' 'pydantic-settings~=2.6.1' 'solathon==1.0.7' 'protobuf==5.29.6' 'grpcio>=1.70,<1.71' 'redis>=4.2' 'legacy-cgi>=1.0'
+	build_venv/bin/pip install --no-cache-dir  --progress-bar off --target ./aleph-vm/opt/aleph-vm/ 'aleph-message==1.4' 'eth-account~=0.10' 'sentry-sdk==2.8' 'aleph-superfluid~=0.2.1' 'sqlalchemy[asyncio]>=2' 'aiosqlite==0.19' 'alembic==1.13.1' 'aiohttp-cors~=0.7.0' 'pydantic-settings~=2.6.1' 'solathon==1.0.7' 'protobuf==5.29.6' 'grpcio>=1.70,<1.71' 'redis>=4.2' 'legacy-cgi>=1.0'
 	build_venv/bin/python3 -m compileall ./aleph-vm/opt/aleph-vm/
 
 debian-package-resources: firecracker-bins vmlinux target/bin/sevctl
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,10 +35,10 @@ dependencies = [
   "aiohttp-cors~=0.7.0",
   "aiosqlite==0.19",
   "alembic==1.13.1",
-  "aleph-message==1.4.0",
+  "aleph-message==1.4",
   "aleph-superfluid~=0.2.1",
   "eth-account~=0.10",
-  "grpcio>=1.70,<1.71",                                                                                   # generated _pb2_grpc.py embeds GRPC_GENERATED_VERSION=1.70.0; older installs RuntimeError on import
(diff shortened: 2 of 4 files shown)


foxpatch-aleph (changes requested):
The core change is correct and well-tested: hoisting the SNP misrouting guard above download_all() fixes a real validation-before-I/O contract violation, and the new measured-mode guard cleanly rejects tdx instances with the mode named (fail-closed for any future measured mode too). I verified the logic against the actual aleph-message 1.4.0 package from PyPI (is_snp, is_measured, tdx mode literal, TdxRegisters fields, and the credit-only rule for measured instances that motivates the payment field in the tests), and ran the translate and SNP suites against 1.4.0 in an isolated venv (28 passed). The one genuine issue: packaging/Makefile:42 still pins aleph-message==1.3.1, violating the file's own documented invariant ("Versions below MUST mirror pyproject.toml so the .deb ships what the code is tested against") and breaking from the convention established by the prior pin bump (7928a9b), which updated both files together. With 1.3.1 shipped in the .deb, a mode=tdx message fails pydantic validation upstream and never reaches the new guard, so the clean "tdx instances are not supported by this CRN" rejection this PR adds would never fire on production nodes. One-line fix; everything else is approvable as-is.

**`packaging/Makefile`** (line 42): Blocking: this pin must be bumped to 'aleph-message==1.4.0' along with pyproject.toml. The comment on line 39 states these versions "MUST mirror pyproject.toml so the .deb ships what the code is tested against", and the prior pin bump (7928a9b) updated both files together. With 1.3.1 shipped, a mode=tdx message fails pydantic validation (Literal["sev", "sev_snp"]) before it ever reaches build_create_vm_spec, so the clean named-mode rejection added in this PR never fires on production nodes — the exact behavior this PR exists to provide.

odesenfans:
daa65e57 takes the blocking finding: the .deb pin in packaging/Makefile now mirrors pyproject.toml, so shipped nodes parse mode=tdx and the named rejection fires instead of the upstream generic validation error. Same commit fixes the CI style failure (pyproject-fmt normalizes ==1.4.0 to the PEP 440-equivalent ==1.4 and realigns the comment column; both files use ==1.4 so they stay grep-identical).

foxpatch-aleph (approved):
Well-scoped, correct change. I verified the core assumptions against the actual aleph-message 1.4.0 wheel: is_measured exists and covers sev_snp+tdx, measured modes forbid the firmware ref (so the old fall-through really did crash at get_existing_file(None)), and measured instances are credit-only (which is why the tests add Payment(type=credit)). Both guards are correctly hoisted before download_all(), the SNP guard keeps its more specific routing error, and legacy SEV instances fall through both guards untouched. The InvalidBackendError propagates through the existing per-VM teardown in run.py:521 with no dangling registry record or leaked staging, identical to the pre-existing non-QEMU rejection path. Pins are consistent between pyproject.toml and packaging/Makefile (==1.4 is an exact PEP 440 match to 1.4.0), and the untouched ~=1.0.x pins are consumer-side. V-PROGRAM staying locked to sev_snp in the 1.4 schema confirms no vprogram change was needed. The new tests use the strong monkeypatch-to-raise pattern, guaranteeing rejection happens before any I/O. I could not re-run the suite locally (sandbox env has an unrelated web3/pytest import breakage and aleph-message 1.2.1.dev), but static verification against the real 1.4.0 package confirms the logic.

**`tests/supervisor/test_supervisor_translate.py`** (line 289): Minor: `fail_download_all` is duplicated verbatim in both new tests. The file's existing rejection tests share the `download_called`-flag pattern via `should_not_be_called`; the raise-on-call pattern is actually stronger, so consider lifting this helper to a module-level fixture or at least noting the pattern is intentional. Purely cosmetic — no action required.

**`src/aleph/vm/agent/run.py`** (line 100): Not changed by this PR, but the `_is_spec_eligible` docstring now lags `build_create_vm_spec`: with the new measured-mode guards it only documents the Firecracker rejection, while the translate path also rejects sev_snp (routing mistake) and tdx (no launch path). Since translate.py's docstring says the two must be kept in sync, a follow-up one-line mention here would help future readers. Non-blocking.

**`src/aleph/vm/agent/translate.py`** (line 85): Nice touch matching on `is_measured` rather than a literal `mode == "tdx"`: any future measured mode added to aleph-message is automatically rejected here instead of falling through to the firmware crash. Worth noting the tests exercise this via the real 1.4 schema (which also enforces credit-only payment for measured modes, hence the `payment` fixture addition).

PR #1189, opened 2026-09-03
fix(agent): keep the SNP instance attestation port in the desired forward set

odesenfans:
## Problem

Found while running the first end-to-end SNP confidential instance against a live 2.0.1 CRN: after a successful create, `/v2/about/executions/list` showed only port 22 in `mapped_ports` while the guest's RA-TLS endpoint (8443) kept serving over direct IPv6. The CLI discovers the attestation endpoint through that mapping, so `aleph instance attest` failed with "attestation port not yet mapped" (workaround: `--url`).

Root cause: the create path maps the runtime manifest's attestation port, but both port-forward reconcilers compute an instance's desired set as aggregate + always-SSH only:

- `reconcile_port_forwards` (port-forwarding aggregate watcher), and
- `reconcile_adopted_port_forwards` (re-adoption healing after an agent restart) — its instance branch, while V-PROGRAMs already have an attest-aware branch.

The first pass after create diffs the live forwards against that reduced set and removes the attestation mapping.

## Fix

Route every instance forward computation through one desired-set helper, `resolve_instance_desired_forwards`, which appends the manifest's RA-TLS port for `mode=sev_snp` instances (deduped against an aggregate-declared forward for the same port). Mirrors the V-PROGRAM pattern:

- `resolve_instance_attestation_port` re-derives the port from the runtime manifest alone (one cached STORE lookup, no bundle staging), like `resolve_vprogram_attestation_port`; the selection loop is factored into `select_attestation_port` and shared with `build_snp_instance_spec`.
- A manifest fetch failure raises instead of returning the reduced set, for the same reason `strict=` exists on the aggregate fetch: reconcilers skip healing rather than converge onto a set that strips the mapping.
- `finish_instance_create` now applies the attestation forward with the rest of the desired set; the create path's separate `add_port_forward` is gone and only the listen gate (`_wait_until_attest_endpoint_listens`) remains. This also means a migrated SNP instance comes up with its attestation mapping, like a fresh one.

## Tests

New `tests/supervisor/test_snp_instance_port_map.py` (mirroring `test_vprogram_port_map.py`): watcher reconcile keeps and restores the mapping, adoption healing includes it, a manifest fetch failure leaves forwards as found, a plain instance never consults the manifest, and an aggregate-declared 8443 is not duplicated. The existing create-path teardown test now exercises the real `finish_instance_create` instead of mocking around the removed manual add.

https://claude.ai/code/session_01AcVreAzpudAwcpP6fX7XGy

### DIFF
--- a/src/aleph/vm/agent/run.py
+++ b/src/aleph/vm/agent/run.py
@@ -34,6 +34,7 @@ from aleph.vm.agent.snp_instance_launch import (
     build_snp_instance_spec,
     is_snp_instance,
     remove_snp_instance_staging,
+    resolve_instance_attestation_port,
 )
 from aleph.vm.agent.translate import build_create_vm_spec, build_program_create_vm_spec
 from aleph.vm.agent.update_watcher import UpdateWatcher
@@ -155,6 +156,38 @@ async def resolve_port_forwards(vm_id: VmId, content, *, strict: bool = False) -
     return forwards
 
 
+async def resolve_instance_desired_forwards(vm_id: VmId, content, *, strict: bool = False) -> list[PortForwardSpec]:
+    """The full desired forward set for an instance: the user's aggregate
+    settings plus always-SSH (``resolve_port_forwards``), and, for a SEV-SNP
+    confidential instance, the runtime manifest's RA-TLS attestation port.
+
+    Every instance forward computation must go through here: a reconciler
+    that diffs against a set missing the attestation port tears its host
+    mapping down (this is how SNP instances lost their 8443 mapping in the
+    field: the aggregate watcher and the re-adoption healing path converged
+    on the aggregate+SSH set). The CLI discovers the guest's attestation
+    endpoint through that mapping.
+
+    A manifest fetch failure raises (VmSetupError) instead of returning the
+    reduced set, for the same reason ``strict=`` exists on the aggregate
+    fetch: a convergence caller must skip healing rather than converge onto
+    a set that strips the attestation mapping.
+    """
+    forwards = await resolve_port_forwards(vm_id, content, strict=strict)
+    if is_snp_instance(content):
+        attest_port = int(await resolve_instance_attestation_port(content))
+        if not any(int(f.vm_port) == attest_port and f.protocol is Protocol.TCP for f in forwards):
+            forwards.append(
+                PortForwardSpec(
+                    vm_id=vm_id,
+                    host_port=HostPort(0),
+                    vm_port=GuestPort(attest_port),
+                    protocol=Protocol.TCP,
+                )
+            )
+    return forwards
+
+
 async def _reconcile_forwards(supervisor: Supervisor, vm_id: VmId, desired_specs: list[PortForwardSpec]) -> None:
     """Diff `desired_specs` against what the hypervisor currently reports for
     `vm_id` and issue add/remove calls so the two converge. The hypervisor
@@ -178,7 +211,7 @@ async def reconcile_port_forwards(supervisor: Supervisor, vm_id: VmId, content) 
     Agent policy half of the old fetch_port_redirect_config_and_setup: compute
     the desired set, then converge through ``_reconcile_forwards``.
     """
-    await _reconcile_forwards(supervisor, vm_id, await resolve_port_forwards(vm_id, content))
+    await _reconcile_forwards(supervisor, vm_id, await resolve_instance_desired_forwards(vm_id, content))
 
 
 def resolve_vprogram_port_forwards(vm_id: VmId, attest_port: int | None) -> list[PortForwardSpec]:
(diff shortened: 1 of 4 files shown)


foxpatch-aleph (approved):
Correct, well-scoped fix for a real field regression: routing all instance forward computations through resolve_instance_desired_forwards guarantees the aggregate watcher, adoption healing, and the create/migration tail all include the SNP attestation port, and factoring select_attestation_port out of build_snp_instance_spec keeps the create-path gate and the reconcilers consistent by construction. The fail-closed choice (raise on manifest-fetch failure rather than converge on a reduced set) is handled correctly at every call site: the watcher catches and logs per-VM, adoption healing leaves forwards as found, and the create path tears down loudly. I verified the manifest schema makes the [0] fallback safe (attestation is min_length=1, transport is Literal["tcp"]), the cached re-fetch is genuinely cheap (download_file short-circuits on an existing file), the dedup handles SSH/UDP/aggregate collisions, and the new + updated test suites all pass.

**`src/aleph/vm/agent/run.py`** (line 178): Minor: the `int()` wrap is redundant — `resolve_instance_attestation_port` already returns `int` (and `select_attestation_port` returns the pydantic-validated port). Harmless, but inconsistent with the surrounding style where `int()` is only applied to the NewType-wrapped `GuestPort` values.

**`src/aleph/vm/agent/run.py`** (line 214): Non-blocking observation: on this watcher path the aggregate fetch inside `resolve_port_forwards` is still non-strict, so a transient aggregate-fetch failure during a watcher pass still converges an SNP instance onto SSH+attest and strips the user's aggregate-declared forwards. That behavior predates this PR and is out of scope here, but since the PR's whole theme is "never converge on a reduced set", a future change might want `strict=True` here too (the caller at tasks.py:229 already catches exceptions).

**`src/aleph/vm/agent/snp_instance_launch.py`** (line 106): Worth noting for future readers: `select_attestation_port` deliberately differs from the V-PROGRAM's `_select_attestation_port` (vprogram_launch.py:134) — the instance version falls back to the first descriptor's port instead of returning None, and doesn't filter on `transport.type == "tcp"`. That's pre-existing create-path semantics, just factored out, and safe because the manifest schema pins `attestation` to `min_length=1` and every transport to `Literal["tcp"]`. A one-line comment contrasting the two selectors might prevent someone "harmonizing" them later.

**`tests/supervisor/test_snp_instance_port_map.py`** (line 64): Two small nits: (1) `list_port_forwards` ignores its `vm_id` argument and always reports VM_ID — fine for a single-VM fake, but a `assert vm_id is not None`-style acknowledgment (like the vprogram fake's `if vm_id is None or ...` filter) would make the contract explicit. (2) `test_manifest_fetch_failure_skips_healing` only covers the adoption path; no test asserts the watcher entry point (`reconcile_port_forwards`) also skips on `VmSetupError` rather than converging — its safety currently relies on the pre-existing try/except in `_handle_port_forwarding_aggregate`.

**`tests/supervisor/test_snp_instance_run.py`** (line 209): The `resolve_instance_attestation_port` mocks added here and at lines 259 and 291 sit on paths that never reach `finish_instance_create` (GPU-rejected raises before create_vm; failure-cleans-staging fails at create_vm; the legacy-SEV test returns at awaiting_confidential_init). Harmless and defensive, but only the mocks at lines 116, 162, and 321 are load-bearing.

foxpatch-aleph (approved):
Solid fix that correctly routes all three instance desired-forward computations (aggregate watcher, re-adoption healing, create/migration tail) through the new resolve_instance_desired_forwards, closing the gap where reconcilers converged on aggregate+SSH only and tore down the SNP attestation mapping. Correctness verified against the manifest model guarantees (attestation min_length=1, tcp-only transports) and every caller's exception handling (per-VM try/except in the watcher, catch-all in adoption healing, teardown branch on create). The fail-closed design is sound and matches the existing strict= rationale. All 13 new/updated tests pass and a full supervisor sweep on both this branch and main shows zero regressions. Two non-blocking observations follow.

**`src/aleph/vm/agent/views/__init__.py`** (line 1088): Heads-up on a new failure surface from this PR: reconcile_port_forwards now propagates VmSetupError on a manifest-fetch failure, and operate_update calls it without a try/except — so a transient CCN error while fetching an SNP instance's runtime manifest turns into a 500 here, where the aggregate-fetch failure was previously swallowed. This is arguably the right fail-closed behavior (the notify genuinely didn't take effect), but worth being aware it's no longer only supervisor errors that escape this endpoint.

**`tests/supervisor/test_snp_instance_port_map.py`** (line 178): Minor test-fidelity nit: the SimpleNamespace makes is_snp_instance return False via its isinstance(content, InstanceContent) short-circuit, so the test exercises the non-instance branch rather than the trusted_execution is None branch that a real plain instance would hit. The property holds either way, but a real InstanceContent fixture (e.g. _make_qemu_instance_message() from test_supervisor_translate) would make this test guard the actual plain-instance path.

PR #1190, opened 2026-09-03
feat(agent): bind SNP instance unlock authority to the message sender

odesenfans:
## Problem

An SNP confidential instance created with `--on-behalf-of` measures `content.address` (the owner) into the cmdline's unlock slot, so only the owner key can inject the LUKS passphrase. For a browser-wallet owner (the whole point of on-behalf-of), that key is not available to the CLI: delegated deployments could be created but never unlocked. Found running the first live E2E.

## Change

Fill the measured `{owner}` slot with the address that **signed the INSTANCE message** instead: the key that deployed the instance (and encrypted/holds the passphrase) is the key that unlocks it.

- Directly-signed messages: sender == content.address, nothing changes.
- Delegated messages: the CCN already enforced the owner's security-aggregate authorization of the sender, so unlock authority still chains from the owner. A delegate that can deploy arbitrary instances on the owner's account gains nothing new by also being able to unlock its own deployments.
- The guest is untouched (it compares the EIP-191 signer against whatever the measured slot holds), so no runtime bundle republish and no measurement-computation change: the client simply computes the expected measurement with the signer's address, which is what the aleph-rs client ships.
- `content.address` remains the billing/ownership identity and may now be non-EVM as long as the sender is EVM-keyed. The rejection message always said "EVM-keyed sender"; the code now checks the sender it named.

Known trade-offs, accepted for v1 (an in-guest delegation-certificate scheme can layer on later without conflicting): the owner cannot unlock a delegated instance themself, and revoking a delegation does not affect already-running VMs (the address is baked into the launch measurement).

## Tests

`test_snp_instance_launch.py`: delegated create binds the sender (not content.address), non-EVM sender rejected regardless of address, non-EVM owner with EVM sender accepted, checksum-cased sender lowercased into the slot. `test_snp_instance_run.py` updated for the new signature (sender plumbed from the loaded message).

https://claude.ai/code/session_01AcVreAzpudAwcpP6fX7XGy

### DIFF
--- a/src/aleph/vm/agent/run.py
+++ b/src/aleph/vm/agent/run.py
@@ -532,7 +532,7 @@ async def create_vm_execution(
                 if requested_gpu_ids(content):
                     msg = "GPU passthrough is not supported on SEV-SNP instances yet"
                     raise VmSetupError(msg)
-                spec, attest_port = await build_snp_instance_spec(vm_hash, content)
+                spec, attest_port = await build_snp_instance_spec(vm_hash, content, str(message.sender))
             else:
                 spec = await build_create_vm_spec(vm_hash, content)
             # Agent-side admission, after the download so a failed download
--- a/src/aleph/vm/agent/snp_instance_launch.py
+++ b/src/aleph/vm/agent/snp_instance_launch.py
@@ -54,7 +54,7 @@ if TYPE_CHECKING:
 logger = logging.getLogger(__name__)
 
 # A lowercased 0x-prefixed 20-byte hex address: v1 confidential instances are
-# EVM-keyed only (owner-auth is EIP-191; see build_snp_instance_spec).
+# EVM-keyed only (the unlock authority signs EIP-191; see build_snp_instance_spec).
 _EVM_ADDRESS_PATTERN = re.compile(r"0x[0-9a-f]{40}")
 
 
@@ -148,7 +148,8 @@ async def build_snp_instance_spec(vm_hash: ItemHash, content: InstanceContent, s
     an on-behalf-of deployment the delegate (whose authorization from the
     owner the CCN already enforced) is the one that holds the passphrase and
     unlocks. content.address stays the billing/ownership identity and may be
-    non-EVM; the sender must be EVM-keyed (owner-auth is EIP-191).
+    non-EVM; the sender must be EVM-keyed (the unlock authority signs
+    EIP-191).
 
     Rejections fire in this order, all before any staging I/O: an
     attestation_port override, a host without SNP support, a policy above 32
@@ -171,7 +172,10 @@ async def build_snp_instance_spec(vm_hash: ItemHash, content: InstanceContent, s
         raise VmSetupError(msg)
     unlock_address = str(sender).lower()
     if not _EVM_ADDRESS_PATTERN.fullmatch(unlock_address):
-        msg = "SNP confidential instances require an EVM-keyed sender in v1 (owner-auth is EIP-191)"
+        msg = (
+            "SNP confidential instances require an EVM-keyed sender in v1 "
+            "(the EIP-191 unlock authority must be an EVM key)"
+        )
         raise VmSetupError(msg)
 
     if content.authorized_keys or content.variables:
--- a/src/aleph/vm/agent/snp_instance_launch.py
+++ b/src/aleph/vm/agent/snp_instance_launch.py
@@ -136,10 +136,19 @@ async def resolve_instance_attestation_port(content: InstanceContent) -> int:
     return select_attestation_port(manifest)
 
 
-async def build_snp_instance_spec(vm_hash: ItemHash, content: InstanceContent) -> tuple[CreateVmSpec, int]:
+async def build_snp_instance_spec(vm_hash: ItemHash, content: InstanceContent, sender: str) -> tuple[CreateVmSpec, int]:
     """Validate, fetch/stage the runtime bundle, and build the SNP CreateVmSpec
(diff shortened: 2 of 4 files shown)


foxpatch-aleph (approved):
A small, correct security-model fix. Binding the measured unlock-authority slot to the message signer rather than content.address fixes a real gap (on-behalf-of deployments with browser-wallet owners could never be unlocked) without weakening the model: the sender value is regex-validated before being formatted into the measured cmdline, the only call site plumbs the sender from the network-loaded message (so restarts stay deterministic), and a delegate binding their own address gains nothing they didn't already have, since the deployer supplies the LUKS passphrase in the first place. Relaxing the EVM requirement from content.address to the sender is consistent with billing/ownership code that already treats the address as opaque. Tests are thorough and all pass (21 + 6). Nits: the 'owner-auth is EIP-191' parenthetical in the rejection message is now ambiguous since the authority is the sender, not the owner; and test_rejects_non_evm_sender's 'regardless of content.address' claim would be airtight with a non-EVM owner passed too.

**`src/aleph/vm/agent/snp_instance_launch.py`** (line 141): Nit: after this change the authenticated party is the sender (unlock authority), not the owner — the "owner-auth is EIP-191" parenthetical is now ambiguous. Consider "(unlock-auth is EIP-191)" or "(the EIP-191 unlock-auth requires an EVM signer)".

**`tests/supervisor/test_snp_instance_launch.py`** (line 361): The docstring says the rejection is "regardless of content.address", but the test only exercises an EVM content.address. Passing a non-EVM owner too (e.g. snp_instance_content(address=NON_EVM_ADDRESS)) would make the claim airtight — though test_non_evm_owner_with_evm_sender_is_accepted already covers the mirror case, so this is minor.

**`tests/supervisor/test_snp_instance_launch.py`** (line 389): Clever but slightly non-obvious: the checksummed literal lowercases exactly to DELEGATE_LOWER, which is why the assertion two lines below compares against it. A one-word comment ("lowercases to DELEGATE_LOWER") would save the reader the detour.

foxpatch-aleph (approved):
Correct and well-documented change: binding the SNP unlock authority to the signature-verified message sender (rather than content.address) is sound — the CCN enforces the owner's security-aggregate authorization of the sender, so unlock authority chains from the owner, and a delegate only unlocks instances it deployed and holds the passphrase for. The value fails closed (strict 0x+40-lowercase-hex pattern checked before any staging I/O), is lowercased before both validation and interpolation so it matches the guest's own lowercasing in aleph-attest-agent, and the frozen {owner} manifest slot is correctly kept while the local variable was renamed. Tests pass (28/28) and cover the delegated-create binding, both non-EVM rejection cases, the non-EVM owner acceptance, and checksum-case lowering. Two non-blocking nits: the run.py plumbing test uses a sender equal to content.address so it cannot distinguish a regression back to content.address, and the amend path (load_updated_message resolves amends, so a delegate-signed amend moves the unlock authority to the amender) extends the documented owner-cannot-unlock trade-off to instances originally created by the owner and may merit a docstring note.

**`tests/supervisor/test_snp_instance_run.py`** (line 44): _SENDER is identical to snp_instance_content()'s default address, so `build_snp.assert_awaited_once_with(VM_HASH, content, _SENDER)` cannot distinguish a regression to `str(content.address)` in run.py — the delegated-create semantics would only be caught at the builder level. Using a delegate address distinct from content.address here (like DELEGATE_LOWER in test_snp_instance_launch.py) would prove the plumbing independently.

**`src/aleph/vm/agent/run.py`** (line 497): Worth noting in the docstring trade-off list: because load_updated_message resolves amends, the unlock authority follows the signer of the *latest* launched message. A delegate-signed amend of an instance the owner originally created moves the unlock slot to the delegate, extending the "owner cannot unlock a delegated instance" trade-off to instances the owner first deployed themself (and, conversely, can brick a VM if the amender doesn't hold the LUKS passphrase). Consistent with the design — the measurement binds the signer of the exact content being launched — but it is a sharp edge the accepted-trade-offs list in the PR description doesn't cover.

foxpatch-aleph (approved):
Clean, well-reasoned change. Binding the measured unlock-authority slot to the authenticated message sender rather than content.address is the right fix for delegated (on-behalf-of) SNP deployments, and the security argument holds: the CCN enforces the sender's signature and delegation before the message is fetchable, direct deployments are unaffected (sender == content.address), and the guest needs no change since it compares the EIP-191 signer against whatever the slot holds. Rejection ordering matches the docstring and all rejections still precede staging I/O. Test coverage is thorough (sender binding, non-EVM sender with both EVM/non-EVM owners, non-EVM owner with EVM sender, checksum lowercasing) and both suites pass locally. Only trivia below.

**`tests/supervisor/test_snp_instance_launch.py`** (line 359): Nit: consider `ids=["evm-owner", "non-evm-owner"]` on the parametrize so the two cases are distinguishable in CI output; the bare `address` param name is also slightly ambiguous next to the fixed `NON_EVM_ADDRESS` sender arg. The docstring covers it, so non-blocking.

**`tests/supervisor/test_snp_instance_run.py`** (line 44): Nit: `_SENDER` duplicates the OWNER_LOWER literal from test_snp_instance_launch.py. If these files are intentionally standalone (no shared fixture module), feel free to ignore.

**`nix/init-instance.sh`** (line 4): Comment now reads slightly stale: for an on-behalf-of deployment the party that "verifies the launch measurement ... then injects the LUKS passphrase" is the sender, not the owner in the content.address sense. Since the `owner=` slot name is a frozen contract this is cosmetic, but a one-line note that the slot holds the message sender would spare future readers a double take.

PR #1191, opened 2026-09-03
fix(agent): convergence reconciles must not converge on the SSH-only fallback

odesenfans:
Follow-up to #1189, from its review's non-blocking observation. Stacked on `od/snp-attest-port-reconcile`; retarget to `main` after #1189 merges.

## Problem

`resolve_port_forwards` falls back to SSH-only when the port-forwarding aggregate cannot be fetched. That fallback is right when establishing a VM's *first* forwards (worst case it starts with SSH only), but every *convergence* caller that diffs against it deletes the user's own forwards on a transient CCN error:

- the port-forwarding aggregate watcher (`_handle_port_forwarding_aggregate`), and
- the user-triggered refresh endpoint (`operate_update`).

The re-adoption healing path already guarded against exactly this with `strict=True`; these two did not.

## Change

`reconcile_port_forwards` gains a `strict=` parameter (default lenient, threaded through `resolve_instance_desired_forwards`, so an SNP instance's manifest-fetch failure is treated the same way), chosen per caller:

- **watcher**: `strict=True` — a failure skips the pass, caught and logged per-VM as before.
- **refresh endpoint**: `strict=True` — a failure now returns 503 with the forwards left unchanged, instead of 200 after silently tearing them down.
- **legacy-SEV post-init first setup** (`operator.py`): keeps the lenient default, with a comment: it establishes the first forwards, nothing exists to tear down, and SSH-only at least makes the VM reachable (the create path's rationale).

## Tests

`test_port_forward_reconcile_strict.py`: strict propagates the failure with no forwards touched; the lenient default still converges onto the SSH fallback (pinning why convergence callers must not use it); strict changes nothing when the fetch succeeds. Full sweep of every suite touching the changed call sites passes (146 tests).

https://claude.ai/code/session_01QUTt5c9gxStxywZVtbVnRG

### DIFF
--- a/src/aleph/vm/agent/run.py
+++ b/src/aleph/vm/agent/run.py
@@ -205,13 +205,23 @@ async def _reconcile_forwards(supervisor: Supervisor, vm_id: VmId, desired_specs
             await supervisor.add_port_forward(spec)
 
 
-async def reconcile_port_forwards(supervisor: Supervisor, vm_id: VmId, content) -> None:
+async def reconcile_port_forwards(supervisor: Supervisor, vm_id: VmId, content, *, strict: bool = False) -> None:
     """Drive the hypervisor's forwards to match the aggregate settings.
 
     Agent policy half of the old fetch_port_redirect_config_and_setup: compute
     the desired set, then converge through ``_reconcile_forwards``.
+
+    ``strict=`` follows ``resolve_port_forwards``' contract and the caller's
+    situation: a caller reconciling a VM that already has forwards (the
+    aggregate watcher, the user-triggered refresh endpoint) must pass
+    ``strict=True``, because converging onto the SSH-only fallback after a
+    transient aggregate-fetch failure would tear the user's own forwards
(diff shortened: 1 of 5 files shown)


foxpatch-aleph (approved):
Correct and well-executed follow-up to #1189. The strict flag is threaded properly: all convergence callers (aggregate watcher, refresh endpoint, re-adoption healing) now skip on transient fetch failures instead of tearing down user forwards, while first-setup callers (create path, legacy-SEV post-init) correctly keep the lenient SSH-only fallback. Call-site comments document each decision, and the tests pin both behaviors including the dangerous lenient one. I verified the three new test scenarios pass against the real code via direct asyncio execution (the sandbox's pytest-asyncio is broken), and confirmed no lint issues on changed lines. Two non-blocking observations: the endpoint's broad except conflates fetch failures with mid-reconcile supervisor failures (where "forwards left unchanged" may be inaccurate), and the new 503 path has no endpoint-level test.

Issue #1187, opened 2026-09-02
SEV-SNP DHCP fails on dual-stack Ubuntu hosts due to mixed resolver families and AppArmor lease denial


sapien-nts:
# SEV-SNP DHCP fails on dual-stack Ubuntu hosts due to mixed resolver families and AppArmor lease denial

## Environment

- aleph-vm 2.0.1 (`7928a9b6f587eb36a5e06033a2f4730c65850c44`)
- Ubuntu 24.04
- SEV-SNP enabled
- Dual-stack host resolver configuration
- AppArmor enforcing the distro `/usr/sbin/dnsmasq` profile

## Problem

An SNP execution can be allocated and start booting, but then die and be restarted before it remains reachable. Two independent dnsmasq failures can cause this behavior.

### 1. IPv6 resolvers are passed in DHCPv4 option 6

The Rust supervisor builds one numeric DHCP option 6 argument from the full resolver list. On a dual-stack host this can produce an argument such as:

```text
--dhcp-option=6,192.0.2.53,2001:db8::53
```

dnsmasq rejects that mixed-family option with:

```text
dnsmasq: bad command line options: bad IP address
```

Numeric option 6 is the DHCPv4 DNS server option and must contain only IPv4 addresses. This also differs from the Python configuration path, which splits `DNS_NAMESERVERS` into IPv4 and IPv6 lists.

### 2. AppArmor denies creation of the per-VM lease file

After correcting the resolver argument, dnsmasq can still be denied when it creates its per-VM lease database:

```text
apparmor="DENIED" operation="mknod" profile="/usr/sbin/dnsmasq" name="/var/lib/aleph/vm/dhcp/<vm-id>.leases" requested_mask="c" denied_mask="c"
```

The distro profile allows its standard lease locations and includes `local/usr.sbin.dnsmasq`, but the aleph-vm package does not currently add its lease path there.

## Expected behavior

- Build DHCPv4 option 6 from IPv4 resolvers only.
- Omit option 6 when no IPv4 resolver is available.
- Keep SNP DHCPv6 address allocation and router advertisements active.
- Preserve an error for malformed resolver values instead of silently discarding them.
- Allow the confined dnsmasq process to create and update `/var/lib/aleph/vm/dhcp/*.leases` without replacing the distro profile or administrator-owned local rules.

## Validation of the equivalent host-side workarounds

After applying both changes on an affected host:

- the per-tap dnsmasq process remained running;
[...]

PR #1192, opened 2026-09-03
fix(dhcp): keep IPv6 resolvers out of DHCPv4 option 6

sapien-nts:
An SNP launch currently passes every detected resolver to numeric DHCP option 6. A dual-stack list therefore produces an argument such as `--dhcp-option=6,192.0.2.53,2001:db8::53`, which dnsmasq rejects as a bad IP address and the scheduler subsequently restarts the VM.

Related GitHub issue: https://github.com/aleph-im/aleph-vm/issues/1187

- [x] The new code is clear, easy to read and well commented.
- [x] New code uses the standard library's IP address parser.
- [x] Changed behavior is documented at the `DhcpConfig` construction and argument-building seams.
- [x] All new behavior is covered by relevant unit tests.
- [x] Existing operator documentation remains accurate; no configuration key changed.
- [x] No dependency changed, so no Debian build-script update is required.

## Changes

`DhcpConfig::for_snp` now parses each configured resolver as an `IpAddr`, retains IPv4 resolvers for DHCPv4 option 6, and ignores valid IPv6 resolvers for that option. Malformed resolver values still fail configuration instead of being silently dropped, matching the Python setup behavior.

For example:

```text
before: --dhcp-option=6,192.0.2.53,2001:db8::53,192.0.2.54
after:  --dhcp-option=6,192.0.2.53,192.0.2.54
```

If the resolver list contains only IPv6 addresses, option 6 is omitted. The existing DHCPv6 single-address range and router advertisement configuration are unchanged, so IPv6 remains active for the guest.

## How to test

From the Rust workspace:

```shell
cargo test --locked -p supervisor-daemon dhcp
cargo fmt --check
```

The added tests cover a mixed IPv4/IPv6 list, an IPv6-only list, and a malformed resolver. Existing tests cover IPv4-only and empty lists.

Local `cargo fmt --check` and diff checks pass. The focused Cargo test could not run on the preparation Mac because its installed compiler is Rust 1.86 while the repository pins Rust 1.90; the dependency check stopped before compilation. The repository's Rust CI uses the pinned toolchain and will run the full locked test suite.

## Print screen / video

Not applicable.

## Notes

This PR intentionally does not add DHCPv6 DNS option 23. IPv6 address allocation and RA behavior remain unchanged, while the guest continues to receive reachable IPv4 DNS servers through DHCPv4 when they are available.

### DIFF
--- a/docs/architecture/networking.md
+++ b/docs/architecture/networking.md
@@ -238,7 +238,7 @@ only ever lease precisely that address (there is no MAC to key a
 `--dhcp-hostsdir` reservation on, since the SNP NIC has no fixed MAC).
 `DhcpConfig::for_snp` derives the config from the VM's `TapAssignment`
 (guest IP, gateway as DHCP option 3, tap-prefix netmask, the daemon's
-resolved nameservers as option 6 when any were found); `--port=0` disables
+resolved IPv4 nameservers as option 6 when any were found); `--port=0` disables
 dnsmasq's own DNS server so per-tap instances never collide on port 53.
 Each server runs as a transient systemd unit named
 `aleph-vm-dhcp-{vm_hash}.service`, launched via `systemd-run --collect -p
--- a/rust/crates/supervisor-daemon/src/dhcp.rs
+++ b/rust/crates/supervisor-daemon/src/dhcp.rs
@@ -72,8 +72,8 @@ pub struct DhcpConfig {
     /// The tap IPv6 prefix length (from the /124-per-VM scheme), the
     /// DHCPv6 `--dhcp-range` prefix and the RA on-link prefix.
     pub ipv6_prefix_len: u8,
-    /// The daemon's resolved nameservers, DHCP option 6. Empty omits the
-    /// option (the guest gets no DNS, the same as a nameserver-less host).
+    /// The daemon's resolved IPv4 nameservers, emitted in DHCPv4 option 6.
+    /// An empty list omits the option.
     pub nameservers: Vec<String>,
     /// The `--dhcp-range` lease-time literal (e.g. "1h").
     pub lease: String,
@@ -93,9 +93,10 @@ impl DhcpConfig {
     /// the VM hash.
     ///
     /// Fails (fail-closed) if the tap's IPv4 network CIDR carries no parseable
-    /// prefix length: silently falling back to `/32` would hand the guest an
-    /// unroutable single-host netmask, so the caller must surface the error
-    /// rather than boot an unreachable VM.
+    /// prefix length or if a nameserver is not a valid IP address: silently
+    /// falling back to `/32` would hand the guest an unroutable single-host
+    /// netmask, while silently dropping malformed DNS configuration would boot
+    /// a guest with no working resolver.
     pub fn for_snp(
         vm_hash: &str,
         tap: &TapAssignment,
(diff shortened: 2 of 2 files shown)


foxpatch-aleph (approved):
Correct, minimal fix for a real bug: dnsmasq rejects IPv6 addresses in DHCPv4 option 6, and the daemon's resolvectl-derived nameserver list can legitimately be dual-stack. The filter mirrors conf.py's DNS_NAMESERVERS_IPV4 split accurately, including the fail-closed behavior on malformed values (ipaddress.ip_address raises, and the new InvalidNameserver error matches that). The new error path is strictly better than the old silent failure, where dnsmasq died after Type=exec reported success and the SNP guest was left without DHCP. Tests cover mixed, IPv6-only, and malformed inputs, and no existing consumers or tests assume pass-through behavior. Two small documentation nits, neither blocking.

**`rust/crates/supervisor-daemon/src/dhcp.rs`** (line 96): The for_snp doc now covers the fail-closed path but never states that valid IPv6 nameservers are silently dropped — arguably the function's most surprising behavior. An IPv6-only host yields an SNP guest with no DNS at all (option 6 omitted, no DHCPv6 option 23). Consider adding a sentence here, e.g. noting that only IPv4 resolvers are forwarded to DHCPv4 option 6, since a reader of this constructor has no reason to look at the dnsmasq_args comment to discover the filtering.

**`docs/architecture/networking.md`** (line 241): This description of DhcpConfig::for_snp says "resolved nameservers as option 6"; after this change it is the resolved IPv4 nameservers. A one-word tweak keeps the operator docs precise.

foxpatch-aleph (approved):
Correct, minimal fix for a real bug: dual-stack resolver lists produced a DHCPv4 option 6 argument containing IPv6 addresses, which dnsmasq rejected and turned into a scheduler restart loop. Filtering at the DhcpConfig::for_snp seam keeps the stored config accurate, matches the Python reference behavior in conf.py (silently drop wrong-family, fail on malformed), and the new InvalidNameserver error propagates cleanly through the existing From<DhcpError> for RpcError conversion. The fail-closed choice introduces no regression since any value the stdlib parser rejects would also have been rejected by dnsmasq, just later and less legibly. Tests cover all three new branches (mixed, IPv6-only, malformed), docs are updated at every seam, and no security or dependency concerns exist.

**`rust/crates/supervisor-daemon/src/dhcp.rs`** (line 189): Non-blocking: valid IPv6 resolvers are dropped entirely, so a host whose resolvers are IPv6-only hands the guest no DNS at all (the --enable-ra RA carries no RDNSS, and option 6 is omitted). The PR notes this is intentional; worth a tracking issue for a --dhcp-option=23 follow-up so DHCPv6 can convey v6 resolvers, mirroring conf.py's DNS_NAMESERVERS_IPV6.

**`rust/crates/supervisor-daemon/src/dhcp.rs`** (line 116): Note for awareness: this fail-closed path is reachable in practice, not just via a bad DNS_NAMESERVERS env var: resolv_conf_dns_servers (net.rs:137) ports the [\w.]+ regex, so a resolv.conf line like `nameserver localhost` yields a non-IP token. Previously that failed later as an opaque dnsmasq unit failure; surfacing it here as InvalidNameserver is a clear improvement, and it is good that a test covers the branch.

PR #1193, opened 2026-09-03
fix(packaging): allow dnsmasq to create SNP DHCP lease files

sapien-nts:
The SNP supervisor starts one dnsmasq process per tap and stores its lease database at `/var/lib/aleph/vm/dhcp/<vm-id>.leases`. On Ubuntu with the distro dnsmasq AppArmor profile enforced, creation of that file is denied, dnsmasq exits, and the SNP execution is restarted.

Related GitHub issue: https://github.com/aleph-im/aleph-vm/issues/1187

- [x] The new code is clear, easy to read and well commented.
- [x] New code uses standard shell and AppArmor mechanisms.
- [x] New helper functions have focused names and comments where behavior is non-obvious.
- [x] Package lifecycle behavior is covered by a fixture test.
- [x] The package-owned policy fragment documents the permission and its purpose.
- [x] No dependency changed, so no Debian build-script update is required.

## Changes

The Debian package now ships a narrow AppArmor fragment containing:

```text
/var/lib/aleph/vm/dhcp/*.leases rw,
```

`postinst` adds one marked absolute include for that package-owned fragment to the distro-supported `local/usr.sbin.dnsmasq` file and conditionally reloads the dnsmasq profile. The helper preserves administrator content and file metadata, is idempotent across upgrades, accepts both modern `include` and legacy `#include` local hooks, and warns without modifying the local file if a custom base profile does not provide the hook.

On package removal, `prerm` removes only aleph-vm's managed lines after its services stop and reloads the profile while the package fragment still exists. Upgrades retain the include for `postinst` to refresh.

The package CI now runs a shell fixture covering install, repeat install, removal, a missing base-profile hook, absent or disabled AppArmor, parser failure, operator-rule preservation, and file-mode preservation. It also asserts that both new files are present in every supported `.deb` variant.

## How to test

Run the focused package test on any development host:

```shell
packaging/tests/test_dnsmasq_apparmor.sh
```

On an Ubuntu package test host, build and install the `.deb`, then verify the real profile and an SNP launch:

```shell
sudo apparmor_parser -r -W -T /etc/apparmor.d/usr.sbin.dnsmasq
sudo journalctl -k --since '10 minutes ago' | grep -iE 'apparmor.*dnsmasq.*DENIED' || true
```

The focused Bash syntax, fixture, staged-diff, and file-mode checks pass locally. A full `.deb` build was not run on the preparation Mac because Podman and `dpkg-deb` are unavailable there; the existing package matrix will build and inspect Debian 12/13 and Ubuntu 22.04/24.04/26.04 packages.

## Print screen / video

Not applicable.

## Notes

The change does not replace the distro dnsmasq profile and does not claim ownership of the administrator's local profile. If package removal scripts are forcibly skipped, the remaining include is harmless after the package-owned fragment disappears because it uses `include if exists`.

### DIFF
--- a/.github/workflows/build-deb-package-and-integration-tests.yml
+++ b/.github/workflows/build-deb-package-and-integration-tests.yml
@@ -47,6 +47,9 @@ jobs:
       - name: Initialize git submodules
         run: git submodule init
 
+      - name: Test dnsmasq AppArmor package integration
+        run: packaging/tests/test_dnsmasq_apparmor.sh
+
       # Runner image 20260726.254 ships podman 5.8.4 but leaves a stale crun 1.14.1
       # at /usr/bin/crun, breaking every podman run/build with "unknown version
       # specified". Point podman at the current crun in /usr/local/bin.
@@ -69,6 +72,8 @@ jobs:
           dpkg --contents packaging/target/${{ matrix.artifact_name }} | grep /opt/aleph-vm/bin/supervisor-launcher
           dpkg --contents packaging/target/${{ matrix.artifact_name }} | grep /opt/aleph-vm/bin/aleph-vm-controller
           dpkg --contents packaging/target/${{ matrix.artifact_name }} | grep /opt/aleph-vm/bin/controller-launcher
+          dpkg --contents packaging/target/${{ matrix.artifact_name }} | grep /usr/lib/aleph-vm/configure-dnsmasq-apparmor
+          dpkg --contents packaging/target/${{ matrix.artifact_name }} | grep /usr/share/aleph-vm/apparmor/usr.sbin.dnsmasq
 
       - uses: actions/upload-artifact@v4
         with:
--- a/packaging/aleph-vm/DEBIAN/postinst
+++ b/packaging/aleph-vm/DEBIAN/postinst
@@ -5,6 +5,11 @@ if ! id -u jailman > /dev/null 2>&1; then
   useradd jailman
 fi
 
+# The distro dnsmasq profile supports site-local additions. Keep our rule in a
+# package-owned fragment and add only its include to the administrator-owned
+# local file.
+/usr/lib/aleph-vm/configure-dnsmasq-apparmor install
+
 rm -fr /srv/jailer  # Upgrade from < 0.1.11
 rm -fr /tmp/aleph   # Upgrade from < 0.1.11
 mkdir -p /var/lib/aleph/vm/jailer
--- a/packaging/aleph-vm/DEBIAN/prerm
+++ b/packaging/aleph-vm/DEBIAN/prerm
@@ -6,3 +6,10 @@ for unit in aleph-vm-agent.service aleph-vm-supervisor.service; do
   systemctl stop "$unit" || true
   systemctl reset-failed "$unit" 2>/dev/null || true
 done
+
+# On removal, drop only aleph-vm's managed include after stopping its services
+# but while the package-owned policy fragment is still present. Keep it across
+# upgrades; postinst refreshes the include and reloads the new policy.
+if [ "${1:-}" = "remove" ] && [ -x /usr/lib/aleph-vm/configure-dnsmasq-apparmor ]; then
+  /usr/lib/aleph-vm/configure-dnsmasq-apparmor remove || true
+fi
--- a/packaging/aleph-vm/usr/lib/aleph-vm/configure-dnsmasq-apparmor
+++ b/packaging/aleph-vm/usr/lib/aleph-vm/configure-dnsmasq-apparmor
@@ -67,8 +67,8 @@ rewrite_local_file() {
     content_tmp="$(mktemp "$(dirname "$target")/.aleph-vm-dnsmasq-content.XXXXXX")"
     # Remove only lines owned by this helper. Any administrator additions,
     # including additions placed between our markers, remain intact.
-    awk -v begin="$BEGIN_MARKER" -v include="$INCLUDE_LINE" -v end="$END_MARKER" \
-      '$0 != begin && $0 != include && $0 != end { print }' \
+    awk -v begin="$BEGIN_MARKER" -v managed_include="$INCLUDE_LINE" -v end="$END_MARKER" \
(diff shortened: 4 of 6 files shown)


foxpatch-aleph (changes requested):
The design is sound and well-researched: I verified against real apparmor_parser 4.0.1 that the generated include-if-exists line parses and expands the lease rule inside the dnsmasq profile, that every supported distro's profile (apparmor 3.0.4 through master) has the local/usr.sbin.dnsmasq hook, that rw matches the distro's own lease rule, and that the rewrite logic is idempotent and preserves admin content and metadata. However, the PR contains two genuine portability bugs I reproduced on a Linux host: (1) awk -v include=... is rejected by gawk because include is a reserved keyword, and under set -eu the helper exits 2, so postinst fails and package install/upgrade breaks on any host whose awk is gawk (gawk takes over the awk alternative from mawk); (2) the test's stat -f '%Lp' ... || stat -c '%a' mode check fails on GNU/Linux because stat -f '%Lp' prints the filesystem status to stdout before exiting non-zero, so the fallback appends 640 to garbage and the comparison fails — meaning the new CI step added in this PR fails on the Linux runner (with gawk it fails even earlier at finding 1). With both worked around (mawk + GNU stat), the entire suite passes, so these are the only latent failures; both are small, localized fixes (rename the awk variable; use a platform-aware mode check).

**`packaging/aleph-vm/usr/lib/aleph-vm/configure-dnsmasq-apparmor`** (line 70): gawk reserves `include` as a keyword (for `@include`), so `awk -v include="$INCLUDE_LINE"` aborts with `gawk: fatal: cannot use gawk builtin 'include' as variable name`. Because of `set -eu`, the helper then exits 2 and postinst (which runs with `set -e`) fails — so package install/upgrade breaks on any host whose `awk` resolves to gawk (installing gawk takes over the `awk` alternative from mawk). I reproduced this end-to-end: the local file is left untouched and the helper returns 2. `begin` and `end` are fine; rename only this variable, e.g. `awk -v managed_include="$INCLUDE_LINE" ... '$0 != begin && $0 != managed_include && $0 != end { print }'`.

**`packaging/tests/test_dnsmasq_apparmor.sh`** (line 66): This mode check fails on GNU/Linux, so the new CI step cannot pass on the ubuntu-latest runner. GNU `stat` treats `%Lp` as a second (nonexistent) file operand: `stat -f '%Lp' "$LOCAL"` prints the *filesystem status* of `$LOCAL` to stdout and exits 1, so the `||` fallback appends `640` to that block and the `= 640` comparison fails. I verified the failure (`MODE CHECK FAILS (output is not just '640')`) and confirmed that switching to `stat -c '%a'` on GNU makes the whole suite pass (with mawk). Note this only worked on the author's Mac (BSD stat). Branch on the platform (e.g. try GNU `-c` first with a fallback to BSD `-f`) instead of relying on GNU stat failing without output.

**`packaging/tests/test_dnsmasq_apparmor.sh`** (line 108): `AA_ENABLED_EXIT=1 run_helper install` relies on bash's non-posix behavior where assignments preceding a function call are temporary. It works under the bash shebang, but in posix mode (or under another shell) the variable would persist and the parser-failure assertion at line 112 would silently break because `reload_profile` would return early. Consider passing these knobs explicitly (e.g. `env AA_ENABLED_EXIT=1 "$HELPER" ...` inside the helper wrapper) to make the test order-independent.

foxpatch-aleph (approved):
This change is correct, careful, and well tested. I validated the end-to-end behavior against the real Ubuntu 22.04 and 24.04 dnsmasq AppArmor profiles and the real apparmor_parser: the rule path matches the production lease directory ({EXECUTION_ROOT}/dhcp/{vm_hash}.leases with default /var/lib/aleph/vm), the distro profiles really do contain the <local/usr.sbin.dnsmasq> hook, the generated include line plus fragment parse cleanly under abi/3.0 and abi/4.0, and no parent-directory r permission is needed (man 5 apparmor.d defines directory r as listing access; the distro profile itself grants /var/lib/misc/dnsmasq.leases rw, without a /var/lib/misc/ r rule). The apparmor_parser -r -W -T flags are the correct --replace/--write-cache/--skip-read-cache reload idiom, and the non-fatal parser-failure handling mirrors dh_apparmor. I also probed idempotency, operator-content preservation, file-mode preservation (0640), symlinked and dangling-symlink local files, empty/blank local files (the post-remove state, which parses fine), and RETURN-trap cleanup of temp files — all behave as documented, and the fixture passes on this Ubuntu 24.04 host. The gawk -v include= rejection simulated by the test shim is accurate. Only minor nits remain.

**`packaging/aleph-vm/usr/share/aleph-vm/apparmor/usr.sbin.dnsmasq`** (line 3): Nit: this rule is tied to the default EXECUTION_ROOT. An operator overriding ALEPH_VM_EXECUTION_ROOT gets lease files elsewhere and will still hit AppArmor denials (the helper will not help). A one-line comment here ("matches the default EXECUTION_ROOT; override via this file's site-local rules if changed") would save future debugging time.

**`packaging/aleph-vm/usr/lib/aleph-vm/configure-dnsmasq-apparmor`** (line 32): Nit: the regex requires the hook line to end right after the closing bracket. A profile carrying a trailing comment on the hook line (e.g. `#include if exists <local/usr.sbin.dnsmasq> # site additions`) would not match, so install would warn and skip. None of the current distro profiles do this, so it is fine as is — just noting the assumption.

**`packaging/tests/test_dnsmasq_apparmor.sh`** (line 1): Minor coverage suggestion: the resolve_local_file() symlink branch (rewrite through a symlinked local file, plus the dangling-symlink and non-regular-file guards) is the subtlest part of the helper and is not exercised by the fixture — I verified both paths manually and they work. Consider adding a case for a symlinked local file to lock that behavior in.

**`packaging/aleph-vm/usr/lib/aleph-vm/configure-dnsmasq-apparmor`** (line 106): Observed while testing: after `remove` on a host where the local file was entirely ours, an empty local/usr.sbin.dnsmasq file remains behind. That is harmless (an empty include parses fine, and it matches dh_apparmor's behavior), just calling it out in case you would rather clean it up when the rewritten content is empty.

PR #1194, opened 2026-09-04
feat(tee): TDX certificate-chain and signature verification

odesenfans:
Increment 3 of the Intel TDX design (#1144): the certificate-chain and signature half of quote verification in `aleph_tee::tdx`, on top of #1186's parser. Additive, plus the golden re-seed the new measured-agent code requires.

**`tdx/certs.rs`**
- The Intel SGX Provisioning Certification Root CA is pinned in-crate (PEM vendored, fingerprint and serial documented). The quote's embedded root must be byte-identical to the pin, the same fail-closed discipline as the pinned AMD ARK: a wrong pin rejects genuine quotes rather than accepting forged ones. An `#[ignore]`d network test compares the pin against the root Intel currently serves, mirroring the SNP KDS liveness test.
- `verify_pck_chain()`: leaf/intermediate/root structure enforced, signatures walked down from the pinned root, validity windows for all three checked against the injected clock, and both CRLs applied (root CA CRL covers intermediates, PCK CRL covers leaves). CRL signatures are verified against certificates from the quote's own verified chain, never against the collateral's issuer-chain fields, so the collateral is not a trust input here. A CRL without `nextUpdate` is rejected rather than treated as never-expiring.

**`tdx/collateral.rs`** — the `sgx_ql_qve_collateral`-shaped JSON set (PEM chains, hex-DER CRLs, TCB Info / QE Identity bodies with detached hex signatures). Only the CRLs are consumed in this increment; the rest feeds the TCB walk (increment 4).

**`tdx/verify.rs`** — `verify_tdx_quote_chain(quote, collateral, now)`:
1. Intel QE vendor id gate.
2-3. PCK chain + CRLs (above).
4. QE report signature under the PCK leaf key.
5. Attestation-key binding: `qe_report.report_data[..32] == SHA-256(att_key || qe_auth_data)` — skipping this is a total break, since any key could otherwise sign the body.
6. Quote signature over `signed_region` under the attestation key.

The module docs spell out what a success does NOT establish: TCB acceptability (increment 4) and workload identity/freshness (the caller's register and report_data checks). **The verifier takes `now` as a parameter and never reads the clock** — the vendored collateral's windows all lie in the past, so an internal clock read would make the verifier untestable against archived collateral; production callers pass the real time and freshness stays enforced.

**Fixtures**: the Phala collateral pair for the existing v4 quote, plus the outdated-TCB quote (a v5) and its collateral — same MIT source and pinned commit as the quote fixtures, README updated. The outdated pair verifies cleanly here by design: "OutOfDate" is a TCB status for increment 4's walk, not a chain defect.

**Tests** (12 new): both genuine quotes verify end-to-end; the same collateral rejected once `now` passes its windows (both directions); swapped and garbage CRLs; non-Intel vendor id; tampered QE report; unbound attestation key; tampered TD report body; and a same-subject self-signed impostor root rejected by the byte-level pin.

**Golden measurements**: re-seeded; aleph-tee compiles into the measured attest agent, so the new modules move the launch measurements (same procedure as #1186, values from `check-golden-measurements.sh --update` at this tip).


foxpatch-aleph (approved):
This increment is solid, security-conscious work: the byte-identical root pin, CRL signatures checked against keys from the quote's own verified chain, injected clock, fail-closed handling of missing nextUpdate, and the attestation-key binding are all correct, and I independently verified the pin fingerprint, the QE report report_data offset of 320, the openssl 0.10.81 API usage (Asn1TimeRef::compare ordering, CrlStatus payload shapes), the test timestamps, and the flake fileset fix against the actual fixtures and crate sources. The test suite is thoughtful — particularly the same-subject impostor root and the both-directions collateral-expiry tests — and the module docs are unusually honest about what verification does not establish. The only gaps are minor: the revocation branch of check_crl is never exercised by any test, and the QE report offset constant is duplicated between verify.rs and quote.rs. Nothing blocks; the QE identity/TCB walk must land before this is treated as full attestation verification, but that is already documented and no production caller exists yet.

**`rust/crates/aleph-tee/src/tdx/certs.rs`** (line 90): The only branch of check_crl never exercised by the tests is the one that matters most: CrlStatus::Revoked. rejects_swapped_crls and rejects_garbage_crl cover parse and wrong-signer failures, but nothing confirms a cert actually listed on a CRL is rejected. Since check_crl takes arbitrary issuer certs, a small unit test with a self-signed cert and a CRL you issue yourself would close that loop without needing Intel's keys.

**`rust/crates/aleph-tee/src/tdx/verify.rs`** (line 35): QE_REPORT_DATA_OFFSET now duplicates knowledge already hardcoded in quote.rs's assert_structure (qe_report[320..352]). If the parser ever exported this constant (the SGX report layout is arguably the parser's domain), the two could not drift apart — minor, since a genuine quote fails loudly if either is wrong.

**`rust/crates/aleph-tee/src/tdx/certs.rs`** (line 70): check_crl relies on the signature check alone to reject a CRL issued by the wrong CA (X509_CRL_verify does not compare names). That is sound here because the issuer keys always come from the pinned chain, but comparing the CRL's issuer name against the issuer cert's subject would produce a much clearer error for the swapped-CRL case.

PR #1195, opened 2026-09-04
feat(tee): TDX TCB appraisal, platform gates and acceptance policy

odesenfans:
Increment 4 of the Intel TDX design (#1144), stacked on #1194 (chain verification). Completes `aleph_tee::tdx` quote verification: the TCB appraisal, platform gates, and acceptance policy on top of the chain half. Additive Rust plus a golden re-seed.

**Base**: this PR targets `od/tdx-dcap-chain` (#1194); review/merge #1194 first. The diff here is increment 4 alone.

**`tdx/pck_extension.rs`** — parses the platform identity from the PCK leaf's Intel SGX extension (OID `1.2.840.113741.1.13.1`): FMSPC, the 16 SGX component SVNs (CPUSVN), and PCESVN. A minimal, fully bounds-checked DER walk keyed on the exact OIDs, not a general ASN.1 parser, since the structure is fixed and the input untrusted.

**`tdx/tcb.rs`** — the appraisal:
- `TcbStatus` (7 variants, severity-ordered) with Intel's convergence rule; `TdxTcbPolicy` (accepted statuses + denied advisories) defaulting to accept `UpToDate` and `SWHardeningNeeded` only. `Revoked` is never acceptable and cannot be added to the accept set — the deliberate asymmetry with the SNP override.
- `verify_tcb_info` / `verify_qe_identity`: Intel's signed documents verified against their issuer chains up to the *pinned* root (increment 3's `verify_signer_chain`), signatures over the exact raw JSON bytes so a re-serialization can't slip a changed field past the check.
- The SVN walk (highest satisfied level wins, over SGX components + PCESVN + the quote's TDX components), the TDX module identity check (MRSIGNERSEAM against the Intel-signed module identity, module SVN ladder), and the QE Identity appraisal (MRSIGNER/ISVPRODID/masked MISCSELECT+ATTRIBUTES, ISVSVN ladder). Statuses converge worst-wins; the algorithm follows the DCAP reference field-for-field.
- Platform gates: `TD_ATTRIBUTES.DEBUG == 0` (a debuggable TD lets the host read guest memory — the single most important gate).

**`tdx/verify.rs`** — `verify_tdx_quote(quote, collateral, now, policy) -> TdxVerification`, composing the chain half with `evaluate_tcb`. The result carries the pinned registers and `report_data`, and its doc spells out that a success is still not a decision to trust a guest: the caller pins registers against the message's `LaunchMeasurement` and binds freshness through `report_data`.

**Fixtures**: reuses the Phala collateral pair (v4) and the outdated pair from #1194. Worth recording: the outdated sample's PCK reports a component SVN below every TCB level, so its walk finds no match (a rejection), which matches the DCAP reference's own behaviour on that sample — it's a "collateral/platform too old" negative, not a clean OutOfDate. Precise status transitions (fall-through to a lower level, the convergence rule) are covered by unit tests on the walk logic with synthetic TCB Info, independent of the fixtures.

**Tests** (18 new, 40 in the tdx module): genuine v4 → `UpToDate` end to end; FMSPC mismatch (v4 quote against the other FMSPC's collateral); forged TCB Info body (signature no longer covers it); debuggable-TD gate; synthetic walk fall-through to a lower status; convergence rule; the full `verify_tdx_quote` accepting a genuine quote and refusing to return registers for a rejected one; plus the PCK-extension parse pinned to exact FMSPC/CPUSVN/PCESVN values.

**Golden measurements**: re-seeded (aleph-tee compiles into the measured attest agent).


foxpatch-aleph (changes requested):
The appraisal logic itself (walk, convergence, platform gates, policy, pinning) is correct and carefully documented, and the QE report offsets and OID/DER walk check out. However, the TCB Info and QE Identity documents' validity windows (issueDate/nextUpdate) are never checked against `now`, even though the collateral is supplied input and every other collateral window (CRLs, cert windows) is enforced. Because each collateral field is independently signed, an attacker can replay arbitrarily old, genuinely-signed Intel collateral with lowered SVN thresholds and have an unpatched platform appraise as UpToDate forever — defeating the increment's purpose and diverging from the DCAP reference, which threads an expiration check through every verification path and by default rejects expired collateral.

**`rust/crates/aleph-tee/src/tdx/tcb.rs`** (line 226): Blocking: the TCB Info and QE Identity bodies carry `issueDate`/`nextUpdate` (both fixtures do), but neither `verify_tcb_info` nor `verify_qe_identity` checks them against `now`. The collateral is supplied input here, just like the CRLs whose windows `certs.rs::check_crl` does enforce — and since each collateral field is independently signed, an attacker can pair a fresh PCK CRL with arbitrarily old TCB Info/QE Identity. Old TCB Info means old (lower) SVN thresholds, so a platform missing recent microcode appraises as `UpToDate` against the stale document when current collateral would say `OutOfDate`. The DCAP reference threads `expiration_check_date` through every verification path and reports a `collateral_expiration_status`, and Intel's appraisal sample policies only accept expired collateral via an explicit `collateral_grace_period`. Also note `TcbInfo`/`QeIdentity` are private and don't deserialize those fields, so a caller can't enforce this outside the module. Suggested fix: deserialize `issueDate`/`nextUpdate` and fail (or consult a grace-period knob on `TdxTcbPolicy`, matching the reference) when `now` is outside the window, plus a test with `now_v4() + ~40 days` pinning the rejection.

**`rust/crates/aleph-tee/src/tdx/pck_extension.rs`** (line 123): `integer_u16` accepts negative DER INTEGERs: a first content byte with the MSB set (e.g. `0x80` = -128) is two's-complement negative, but this parses it as a large positive, which would inflate `pcesvn` in the walk (`platform.pcesvn < level.tcb.pcesvn`). Low risk since the source is an Intel-signed PCK, but rejecting values with the sign bit set is a one-line hardening that also documents the assumption.

**`rust/crates/aleph-tee/src/tdx/tcb.rs`** (line 69): The doc says Revoked "cannot be added to `accepted_statuses`", but `accepted_statuses` is a plain public `BTreeSet` that accepts `TcbStatus::Revoked`. The invariant is genuinely enforced in `evaluate_tcb` (good), so either soften the doc to say it's enforced there, or add a test pinning it: insert `Revoked` into a policy and assert a Revoked outcome is still rejected. Right now no test exercises the Revoked-vs-policy path, which is one of the security claims of this PR.

**`rust/crates/aleph-tee/src/tdx/tcb.rs`** (line 462): The `denied_advisories` enforcement is never executed by any test: the v4 fixture's accepted level carries no advisories, so this branch is dead in the suite. A synthetic-walk test (in the style of `walk_falls_through_to_a_lower_status`, which can put `advisoryIDs` on the matched level) with a denied advisory would cover the last policy dimension.

**`rust/crates/aleph-tee/src/tdx/tcb.rs`** (line 521): This asserts only `is_err()`, so it would also pass if the rejection came from an unrelated cause (e.g. an expired CRL window at `now_outdated()`). Asserting the error mentions "below every level" would pin the intended cause, as the sibling tests do for their failures.

**`rust/crates/aleph-tee/src/tdx/tcb.rs`** (line 95): Nit: "The outcome of a full TDX quote verification" — this is the TCB appraisal outcome; a full verification also returns registers and report_data (`TdxVerification` in verify.rs). Consider "The outcome of a TDX TCB appraisal" to avoid confusion.

foxpatch-aleph (changes requested):
The overall structure is strong: signature verification over raw JSON bytes, byte-level root pinning, a careful bounds-checked DER walk, and a correct severity ordering and convergence rule (I verified both against the dcap-qvl reference the fixtures come from, and re-ran the SVN/QE/module walks by hand against the fixture data — the genuine v4 quote does appraise UpToDate and the gate offsets are right). However, the appraisal is missing a core freshness gate: TCB Info and QE Identity carry issueDate/nextUpdate, but neither struct deserializes them, so only the certificate windows are checked. Intel's QVL rejects expired collateral documents; without that, a host can replay year-old Intel-signed collateral and keep a platform Intel has since downgraded appraised as UpToDate (the signer certs stay valid far longer than the documents). The verify.rs module doc's claim that freshness stays enforced in production is currently only true for CRLs and certs. Second, the TDX module identity check verifies MRSIGNERSEAM but omits the masked seam_attributes comparison the reference performs (this collateral pins attributes 00...00 under mask FF...FF, so a TD report with module attributes set should be rejected). Third, the module appraisal is skipped entirely when tee_tcb_svn[1] == 0 or tdxModuleIdentities is empty, where the reference falls back to the base tdxModule entry and still checks mrsigner — a fail-open in the identity check. All three are fixable within the current fixtures (the injected test clocks already sit inside both document windows).

**`rust/crates/aleph-tee/src/tdx/tcb.rs`** (line 226): Neither TcbInfo nor QeIdentity deserializes issueDate/nextUpdate, so the signed documents' own validity windows are never checked — only the signer-certificate windows in verify_signer_chain. Intel's QVL treats expired TCB Info/QE Identity as an error, and the reference implementation this PR's fixtures come from folds these windows into a collateral time window. Without this, a host can replay stale Intel-signed collateral indefinitely: a platform Intel has since downgraded (e.g. via a new advisory) continues to appraise as UpToDate, since the signer certs stay valid for years. The same gap applies to verify_qe_identity. The fixture clocks (2025-06-20 and 2026-02-19) are inside both windows, so adding the check won't break the tests — and it would let the doc claims in verify.rs ("freshness stays enforced in production") actually hold.

**`rust/crates/aleph-tee/src/tdx/tcb.rs`** (line 323): The module identity appraisal checks MRSIGNERSEAM but not seam_attributes. The reference also compares the TD report's SEAMATTRIBUTES against the identity's attributes under attributesMask, and this collateral pins attributes 00...00 under mask FF...FF — i.e. the module must run with no attributes set (notably the module's own DEBUG bit). A TD report with seam attributes set currently passes appraisal here but is rejected by the reference. The fields are already in the fixture collateral; it's a cheap addition to the same function.

**`rust/crates/aleph-tee/src/tdx/tcb.rs`** (line 312): This early return skips the module identity check entirely when tee_tcb_svn[1] == 0 or tdxModuleIdentities is empty. The reference instead falls back to the base tdxModule entry (and errors when it is absent), still enforcing the mrsigner/attributes comparison. As written this is fail-open: an identity check that silently disappears is worse than one that errors. Recommend mirroring the fallback, or at minimum treating a v3 TDX TCB Info with no identities as a rejection.

**`rust/crates/aleph-tee/src/tdx/tcb.rs`** (line 262): The walk takes the first satisfied level in JSON order. The reference deliberately does not rely on document order — it canonicalizes the levels (SGX components, then pcesvn, then TDX components) to mirror the QVL's sorted container. Intel publishes levels highest-first so this works in practice, but if a signed document ever shipped levels unordered, the walk could return a lower level's status (e.g. accept UpToDate when a higher satisfied level says OutOfDate). Sorting like the reference removes the assumption.

**`rust/crates/aleph-tee/src/tdx/certs.rs`** (line 198): Minor: verify_signer_chain checks the windows of the intermediate and signer but not of the pinned root, while verify_pck_chain does check the root's window. Since it's the same pinned cert, the two paths will disagree after 2049 — probably worth being consistent (either check it in both or neither).

**`rust/crates/aleph-tee/src/tdx/mod.rs`** (line 2): The module doc still says the TCB walk and platform gates "arrive in later increments" — this PR adds both, so the paragraph is now stale.

odesenfans:
Thanks, both rounds were sharp. Nine findings taken across four commits, one declined with rationale.

**Blocking — collateral freshness** (the big one): TCB Info and QE Identity now have their `issueDate`/`nextUpdate` checked against `now`, closing the replay of stale-but-signed collateral with lowered SVN thresholds. The two structs deserialize the dates and a document outside its window is rejected. I used a small self-contained RFC3339 parser rather than adding a date crate, deliberately — the agent this compiles into is the *measured* one, so a new dependency there moves the golden measurements and widens the trust surface for one fixed timestamp format. Hard window check (fail-closed) for now; a grace-period knob on `TdxTcbPolicy` is an easy follow-up if you ever want it operationally. Test at `now_v4() + 40 days` pins the rejection, and the `verify.rs` "freshness stays enforced in production" claim now actually holds for every collateral field.

**Module identity** (both R2 gaps): the check now also compares SEAMATTRIBUTES against the identity under `attributesMask`, and falls back to the base `tdxModule` entry (still enforcing MRSIGNERSEAM/attributes) when no per-version identity applies, instead of skipping the gate — the fail-open you flagged. A v3 TDX doc with no `tdxModule` is now a rejection.

**Policy paths tested**: extracted `check_policy()` so the Revoked-vs-policy and denied-advisory branches are unit-testable directly. New tests: Revoked rejected even when inserted into `accepted_statuses`, and a denied advisory rejected (and allowed when not denied). Softened the `Revoked` doc to say the rejection is enforced in `evaluate_tcb`, not by set membership.

**Smaller ones**: negative DER integer rejected in `integer_u16`; the outdated-sample test now asserts its "below every level" cause rather than bare `is_err()`; `verify_signer_chain` checks the pinned root's window like `verify_pck_chain`; `TdxTcbOutcome` doc reworded to "TCB appraisal outcome"; the stale mod.rs "later increments" line updated.

**Declined — sorting the TCB levels** (R2, line 262): I'd leave this. First, the reference doesn't actually sort — dcap-qvl iterates `tcb_levels` in document order and returns the first match, same as here. Second, and the reason it's safe: the walk only ever returns a level the platform *satisfies*, and among satisfied levels a lower one carries a strictly worse (more severe) status — so a hypothetical misordered document could only make us **stricter** (reject a platform that should pass), never accept one that shouldn't. The document is Intel-signed, so an attacker can't reorder it regardless. Happy to add the sort as belt-and-suspenders if you'd prefer, but it defends against an Intel bug with no security downside if it happened.

Golden re-seeded (all changes are in the measured aleph-tee). Gates green: 44 tdx tests, fmt, clippy `--all-targets`.

foxpatch-aleph (approved):
Solid, consistently fail-closed implementation of the DCAP TCB appraisal. The SVN walk, Intel convergence rule, module-identity and QE-identity gates, and the Revoked-is-never-acceptable invariant all match the DCAP reference semantics; the DER walker is bounds-checked with correct Intel OIDs; and the serde field mappings (including the advisory_i_ds -> advisoryIDs rename) check out against the real vendored collateral. The compose-and-gate API in verify.rs never returns registers for a rejected quote, and the doc explicitly scopes what a verification does not establish. Findings are documentation and test-coverage nits only.

**`rust/crates/aleph-tee/src/tdx/tcb.rs`** (line 76): This doc comment references "the SNP override, which admits any concrete named TCB", but no such override exists anywhere in this crate (a grep for it finds only this line). Either point at where the SNP override lives (another crate/increment?) or drop the comparison, since right now it is opaque to the reader.

**`rust/crates/aleph-tee/src/tdx/tcb.rs`** (line 577): The advisory path is only covered by the unit test on check_policy. The outdated fixture does carry advisories on its module-identity levels (INTEL-SA-01036/01099), but that sample fails inside walk_platform_tcb before any advisories are collected, so the end-to-end flow (collect advisories from real collateral, then block via a denied_advisories entry) is never exercised. Consider an end-to-end test that pairs the v4 quote with a policy denying an advisory that a synthetic or real document actually reports.

**`rust/crates/aleph-tee/src/tdx/pck_extension.rs`** (line 70): The length bytes are accumulated from a min()-clamped slice and only validated against buf.len() after the loop (line 73). It is correct, but hoisting the "truncated DER length" check above the loop would make the intent clearer and drop the saturating_sub clamp. Also worth a note that this walker accepts non-DER-canonical encodings — harmless here since the input is a signature-verified certificate.

**`rust/crates/aleph-tee/src/tdx/tcb.rs`** (line 286): parse_rfc3339_z validates month/day ranges but not hour/minute/second (or day-vs-month length), so e.g. hour 99 or Feb 30 parse to a shifted timestamp. Harmless for Intel-signed input, but a one-line range check would make the parser's guarantees obvious without relying on that.

foxpatch-aleph (changes requested):
High-quality security-critical work overall: I independently verified the DER walk, OIDs, SVN-ladder arithmetic, QE report offsets, and TcbStatus severity ordering against the real Intel fixtures, and everything checks out with a consistently fail-closed posture (pinned root, raw-bytes signatures, document-window anti-replay, Revoked as a hard invariant). One genuine algorithmic divergence blocks this: TcbStatus::converge implements Intel's convergence rule asymmetrically — only (component=OutOfDate, platform=config-needed) maps to OutOfDateConfigurationNeeded, whereas the reference applies the rule symmetrically, so OutOfDate.converge(ConfigurationNeeded) wrongly yields OutOfDate. The fix is a one-line mirrored match arm plus a test case. Secondary (non-blocking): verify_signer_chain lacks an adversarial impostor-chain test even though the PCK chain has rejects_foreign_root, the tcbLevels highest-first ordering is trusted where QVL sorts defensively, and the hand-rolled date parser is lenient on time fields.

**`rust/crates/aleph-tee/src/tdx/tcb.rs`** (line 58): The convergence rule is missing its mirrored case. Intel QVL's operator| treats the OutOfDate-plus-ConfigurationNeeded combination symmetrically: OutOfDate | ConfigurationNeeded and ConfigurationNeeded | OutOfDate both yield OutOfDateConfigurationNeeded. Here only (component, self) = (OutOfDate, ConfigurationNeeded|ConfigurationAndSwHardeningNeeded) hits the special case; the reverse falls through to max(), so OutOfDate.converge(ConfigurationNeeded) returns OutOfDate (4) instead of OutOfDateConfigurationNeeded (5). That combination is reachable: the platform walk yields OutOfDate on unpatched platforms while the QE identity ladder can yield a ConfigurationNeeded level — and TcbStatus::parse accepts those strings. The fixtures never carry ConfigurationNeeded statuses, so the tests can't catch it. Under the default policy both statuses are rejected so behavior is identical, but a policy accepting OutOfDate diverges fail-open from the reference the PR description claims to follow field-for-field. Suggest: match (component, self) { (OutOfDate, ConfigurationNeeded | ConfigurationAndSwHardeningNeeded) | (ConfigurationNeeded | ConfigurationAndSwHardeningNeeded, OutOfDate) => OutOfDateConfigurationNeeded, _ => component.max(self) } — and add the mirrored case to converge_worst_wins_and_config_rule.

**`rust/crates/aleph-tee/src/tdx/certs.rs`** (line 198): verify_signer_chain has no adversarial test. This function is what makes the TCB Info and QE Identity signatures meaningful — verify_signed_document checks the signature under whatever signer this chain supplies, so if the intermediate-against-pinned-root check regressed, rejects_forged_tcb_info_signature would still pass (an attacker's own chain would verify its own body). The PCK chain has rejects_foreign_root; suggest mirroring it here: swap collateral.tcb_info_issuer_chain for a same-subject self-signed impostor pair and assert rejection in verify_tdx_quote.

**`rust/crates/aleph-tee/src/tdx/tcb.rs`** (line 368): Both this walk and the QE/module ladders rely on Intel publishing tcbLevels highest-first. The DCAP reference sorts the levels defensively after parsing; here a mis-ordered (still genuine, Intel-signed) TCB Info would silently select the wrong status. The doc comment does state the assumption, so this is a hardening suggestion rather than a bug — either sort the levels before walking or note explicitly that the order comes from the signed document.

**`rust/crates/aleph-tee/src/tdx/tcb.rs`** (line 286): Minor: hour/minute/second are never range-checked (e.g. "99:99:99" parses fine), and day=31 is accepted for any month, rolling into the next month via days_from_civil. Only Intel-signed documents reach this code so it's harmless, but since the month/day check is right below, a symmetric range check (or a one-line comment noting the leniency) would keep the function honest.

**`rust/crates/aleph-tee/src/tdx/tcb.rs`** (line 76): Two doc nits: (1) "the SNP override, which admits any concrete named TCB" isn't resolvable from this repository — a reader of this crate has no way to find it, so a pointer to the design doc would help. (2) The comment says the Revoked rejection is "enforced in evaluate_tcb" where it actually lives in check_policy (admittedly private and only called from evaluate_tcb, so the meaning survives).

**`rust/crates/aleph-tee/src/tdx/tcb.rs`** (line 309): Only the expired direction of check_document_window is tested (rejects_expired_tcb_info). The not-yet-valid direction (now < issueDate) is a live rejection path that a one-line test with now = issueDate - 1s would pin.

odesenfans:
Round 4 taken in two commits: e7c631ad (code and tests) and 0f27a840 (golden re-seed). One item declined with references, one accepted with a correction to my own round-2 reply.

**Declined: the mirrored convergence arm.** Neither reference applies the rule symmetrically. Intel's QVL (`Src/AttestationLibrary/src/Verifiers/Checks/EvaluateTcb.cpp`, `convergeTcbStatuses`, in the `intel/confidential-computing.tee.dcap.qvl` repo that the DCAP tree pulls in as a submodule) starts from the platform status and reacts to exactly two component statuses: any `OutOfDate` component maps UpToDate/SWHardeningNeeded to OutOfDate and the two configuration statuses to OutOfDateConfigurationNeeded, and any `Revoked` component yields Revoked. Every other component status leaves the platform status untouched, so `(platform = OutOfDate, component = ConfigurationNeeded)` is `OutOfDate` there. dcap-qvl's `TcbStatus::converge_with_component` (`src/tcb_info.rs`) is the same match this PR carries. The suggested arm would return OutOfDateConfigurationNeeded for that pair, which diverges from both references rather than aligning with them. Reachability is also narrower than stated: the PCS API schema restricts Enclave Identity and TDX module `tcbStatus` to UpToDate/OutOfDate/Revoked. The pair is now pinned in `converge_worst_wins_and_config_rule` with the reference named, so the behaviour is explicit rather than implicit.

**Accepted: sort the levels, with a correction.** My round-2 reply said dcap-qvl walks in document order. It does not: `verify.rs` calls `canonicalize_tcb_levels()` before the walk, and Intel's `TcbLevel::operator>` feeds a sorted container. My "misordering can only make us stricter" argument also relied on a lower level never carrying a better status than a higher level the platform also satisfies, which is Intel's publishing discipline, not a property of the walk. `canonical_levels()` now orders the levels by SGX components, then PCESVN, then TDX components, highest first, and the module and QE ladders order by ISVSVN. `walk_does_not_trust_document_order` lists two satisfied levels lowest-first and expects the higher level's status.

**Also done:** `rejects_tcb_info_under_an_impostor_chain` (a same-subject self-signed signer/root pair, rejected at the pinned-root check, the signer-chain counterpart of `rejects_foreign_root`); `rejects_not_yet_valid_tcb_info` (issueDate minus one second); the date parser now range-checks hour, minute and second, and the day against the month with leap years; the Revoked doc names `check_policy`; the SNP comparison points at the aleph-rs SDK's `--min-tcb` `TcbFloorOverride` instead of an unnamed override.

Gates: 47 tdx tests, fmt, clippy `--all-targets` clean on the crate; golden re-seeded.

foxpatch-aleph (approved):
This PR is a high-quality, security-careful implementation of the TDX TCB appraisal. I verified the OID encodings, SGX report field offsets, TCB status severity ordering (matches Intel's enum), and the date parser, and independently simulated the platform TCB walk against both vendored fixtures — the v4 quote appraises UpToDate, the outdated sample falls below every level exactly as the PR description discloses, and the FMSPC-mismatch path exercises correctly. Failure modes are consistently fail-closed: unknown TCB statuses, malformed component counts, negative DER integers, mismatched module identities, and out-of-window documents all reject. The signature checks over raw JSON bytes under the pinned root, the Revoked short-circuit in check_policy, and the debug-TD gate are all sound and well-tested. The observations below are non-blocking: the converge() divergence from Intel for a Revoked component is strictly more conservative (Revoked is the worse verdict and is rejected unconditionally), module-2.0/TD-report-1.5 quotes are rejected at the MRSIGNER gate rather than silently under-appraised, and the remaining items are documentation/test-precision nits.

**`rust/crates/aleph-tee/src/tdx/tcb.rs`** (line 58): The test comment on converge_worst_wins_and_config_rule says Intel's convergeTcbStatuses reacts to an OutOfDate *or Revoked* component, but converge() only handles component == OutOfDate. For (component=Revoked, platform=ConfigurationNeeded) Intel would return OutOfDateConfigurationNeeded while this returns Revoked. The result is strictly more severe and Revoked is rejected unconditionally in check_policy, so this is fail-closed — but the comment and code disagree on what's implemented, which will confuse a future reader reconciling with the reference. Either drop 'or Revoked' from the comment or add the Revoked arm for exact parity.

**`rust/crates/aleph-tee/src/tdx/tcb.rs`** (line 447): tdx_module_status reads only tee_tcb_svn[0]/[1], so TD report 1.5's tee_tcb_svn2 (parsed in quote.rs but never consumed) is unappraised. A TDX module 2.0 platform reports its version in tee_tcb_svn2[1] while tee_tcb_svn[1] is 0, so it falls to the base tdxModule identity whose mrsigner is the all-zeros module-1.0 value and gets rejected at the MRSIGNERSEAM gate — fail-closed, but worth a note in the doc comment that module 2.0 quotes are rejected, not appraised, until tee_tcb_svn2 is handled.

**`rust/crates/aleph-tee/src/tdx/verify.rs`** (line 319): Minor: the comment says 'Genuine chain, but a debuggable TD' — after tampering raw[48 + 120] the quote signature no longer verifies, so in the composed verify_tdx_quote the failure actually surfaces at the chain stage ('quote signature'), not the debug gate. The assertion is .is_err() so the test passes either way; consider rewording the comment so a reader doesn't assume this exercises the gate in the composed path (the dedicated rejects_debuggable_td test in tcb.rs covers that).

**`rust/crates/aleph-tee/src/tdx/tcb.rs`** (line 529): The QE identity gates (MRSIGNER, ISVPRODID, masked MISCSELECT/ATTRIBUTES) have no adversarial test — understandable since every collateral field is Intel-signed, but qe_identity_status is unit-testable with a synthetic QeIdentity and a mutated qe_report, mirroring the synthetic walk tests. Same applies to the MRSIGNERSEAM/SEAMATTRIBUTES gates in tdx_module_status. Nice-to-have, not blocking.

**`rust/crates/aleph-tee/src/tdx/tcb.rs`** (line 598): evaluate_tcb takes pck_leaf as a parameter, so a caller could pass a leaf that doesn't belong to the quote; the doc comment does call out the precondition and verify_tdx_quote composes it correctly. If this API ever grows external callers, consider deriving the leaf from quote.signature.pck_chain_pem inside evaluate_tcb (or taking quote only) to remove the footgun.

PR #1196, opened 2026-09-04
feat(tee, attest-agent): NVIDIA CC part A, GPU evidence route

odesenfans:
## NVIDIA CC on SEV-SNP, part A of 5: attest-agent GPU evidence route

First PR of the confidential-GPU stack. Design: `docs/superpowers/specs/2026-09-04-nvidia-cc-design.md`, plan: `docs/superpowers/plans/2026-09-04-nvidia-cc-aleph-vm.md` (both in this PR).

### What

- `aleph-tee`: `gpu_nonce(served_key, client_nonce)`, the key-bound, domain-separated SHA-256 SPDM nonce (`aleph-gpu-nonce-v1`), with a pinned vector shared with the aleph-rs mirror.
- `aleph-attest-agent`: a `GpuEvidenceSource` trait whose production implementation runs NVIDIA's `nvattest collect-evidence` as a child process (chrooted into the verity rootfs by init) with the derived nonce as the last argument, parses its JSON, and refuses evidence answering any other nonce. The agent is a static musl binary inside a content-only initrd, so it cannot load NVIDIA's glibc NVML itself.
- New route `GET /.well-known/attestation/gpu?nonce=<hex>` returning `{tee_type:"nvidia-cc", client_nonce, gpus:[{arch,nonce,evidence,certificate}], boot_claims}` over the attested channel, enabled by `--gpu-claims` and `--gpu-collector`, 404 otherwise. The existing `/.well-known/attestation` response is byte-identical.

### Error contract of the GPU route

400 on an over-long or non-hex nonce (same bound as the SNP route), 404 on a runtime without GPU attestation, 503 `{"error": "gpu attestation busy"}` with `Retry-After` when another collection has held the GPU for more than ten seconds, 500 with a bare message when the collector fails (detail in the guest log). There is no "not attested" state: init powers the guest off when boot-time verification fails.

### Review round (2026-09-08)

- The collection is bounded end to end: the reader threads report over channels and the caller gives up on them at the same deadline as the child, so a helper that outlives the collector can no longer pin a worker and the GPU mutex. The timeout is injectable, and the kill path, the held-open pipe and the 16 MiB output cap each have a test.
- The route waits at most ten seconds for its turn at the GPU and answers 503 otherwise.
- `--gpu-claims` conflicts with `--insecure-plain-http`: plain mode has no served key to bind the evidence to, so the GPU route fails closed there like the SNP route does. Part D's init must skip the GPU flags when the insecure cmdline token is set.
- The boot claims must be a non-empty JSON array, so a broken extraction fails at start-up instead of being served.
- Spec and `report_data.rs` docs: the unreachable 503 "gpu not attested" is gone, the module doc no longer says "both schemes".

### Golden measurements

The agent binary is baked into the measured initrd, so this PR moves every base launch measurement and re-seeds `nix/golden-measurements.json` itself (last re-seed after the rebase onto dev-2.1 with the TDX TCB appraisal, #1195, which moved the agent binary too). The golden-measurements CI job is expected green here; B and C do not touch measured inputs but must rebase over these values.

### Testing

`cargo test` in `rust/` and in the agent's own workspace (49 tests in the agent crate); the pipe-draining collector has a 200 KiB output test (a poll-then-read loop deadlocked on the 64 KiB pipe buffer during review and was fixed).

### Stack

A (this) -> B `od/nvidia-cc-b-cc-probe` -> C `od/nvidia-cc-c-snp-argv` -> D `od/nvidia-cc-d-nix-gpu-image` -> E `od/nvidia-cc-e-agent`.


foxpatch-aleph (approved):
Solid, fail-closed first slice of the NVIDIA CC stack. I independently verified the pinned gpu_nonce vectors (all three match a Python recomputation), confirmed the wait()-after-try_wait() pattern at gpu.rs:152 is correct on Unix (std caches the exit status; try_wait stores it and wait returns the cached value, so no ECHILD), and confirmed the pipe-drain fix is sound with the 200 KiB regression test. The route is 404 when the GPU feature is off, nonce handling is shared with the SNP route, error detail stays in the guest log, and the CLI flags are mutually required. The dev-only Cargo.lock additions (tempfile) do not enter the measured binary. Remaining observations are minor robustness notes below; none block.

**`rust/crates/aleph-attest-agent/src/gpu.rs`** (line 137): On the timeout path the two reader-thread JoinHandles are dropped without being joined: the threads detach and normally exit once the pipes close. That is the right call for the grandchild-holds-the-pipe case (joining could block forever), but it means a collector that spawns a surviving grandchild would leak two threads per timeout. Worth a comment noting the detach is deliberate, since joining would be the worse option here.

**`rust/crates/aleph-attest-agent/src/proxy.rs`** (line 150): The lock is held across the entire web::block collect (up to COLLECT_TIMEOUT, 60 s) with no bound on how many callers queue and no timeout on acquisition, so N concurrent callers can each wait up to N*60 s with no backpressure. Callers are attested clients, so this is low risk, but a try_lock-based 503 or a bounded wait would turn pile-ups into fast failures instead of long hangs.

**`rust/crates/aleph-attest-agent/src/gpu.rs`** (line 40): The timeout/kill branch has no test coverage (the other collect() paths all do). A collector script that sleeps forever plus a shorter timeout parameterized for tests (or #[cfg(test)] override) would cover the kill-and-bail path, including the detach behavior above.

foxpatch-aleph (commented):
Part A of the NVIDIA CC stack is solid: the gpu_nonce scheme is correctly domain-separated and all three pinned vectors independently verify, the collector's pipe-draining fix is correct (and regression-tested at 200 KiB), the nonce-refusal logic in parse_output is fail-closed, the decode_nonce refactor keeps the SNP route byte-identical, and the CLI gates are fail-closed. The findings below are non-blocking: a spec/impl inconsistency on the 503 error contract, an unbounded join() after the collector child exits, a stale sentence in the report_data module doc, and the timeout branch being the one piece of collect() with no test coverage.

**`docs/superpowers/specs/2026-09-04-nvidia-cc-design.md`** (line 248): The spec added in this PR documents a 503 with {"error": "gpu not attested"} for "boot-time verification did not complete", but the implementation never emits it: gpu_attestation_endpoint (proxy.rs:141) returns 404 when GpuState is absent and 500 on collection failure, and init powers off on boot-verification failure, so the 503 state cannot occur. Either drop the 503 from the spec or serve it defensively, so the aleph-rs verifier authors don't wire up handling for an unreachable code.

**`rust/crates/aleph-attest-agent/src/gpu.rs`** (line 145): COLLECT_TIMEOUT only bounds the direct child: after try_wait() reports exit, stdout_handle.join() has no deadline. If the collector ever spawns a helper that inherits stdout and outlives the direct child (or a grandchild survives the kill() on the timeout path), the join blocks forever, the web::block worker stays pinned, and GpuState.lock is never released, so every subsequent GPU attestation request queues indefinitely. nvattest doesn't fork today and the command line comes from the measured initrd, so not blocking, but worth either bounding the reads or noting the invariant (the collector must not leak pipe holders).

**`rust/crates/aleph-tee/src/report_data.rs`** (line 34): With the third shape added, the closing sentence "Both schemes write SHA-384(...) into the first 48 bytes of the 64-byte field" reads as stale and now ambiguous: the GPU-nonce scheme writes a 32-byte SHA-256 and is not a report_data at all. Consider "Both report_data schemes write ..." (and the "two namespaces" on line 28 could become "the schemes").

**`rust/crates/aleph-attest-agent/src/gpu.rs`** (line 136): The timeout branch (kill + bail) is the one piece of collect() logic with no test: the fake-collector tests all exercise collectors that exit on their own. Consider making the timeout injectable (e.g. a collect_with_timeout(duration) used by collect()) so a test can pass a short bound against a sleeping script and assert the kill path, rather than waiting 60 s.

foxpatch-aleph (changes requested):
The GPU evidence route is well designed: gpu_nonce is correctly domain-separated (all three pinned vectors verified independently), the collector runs via argv with no shell, parsing is fail-closed (result_code, non-empty evidences, per-GPU nonce match), and the pipe-draining fix has a real regression test. But the PR does not compile its own test configuration: AppState gained the gpu field, yet the no_tee_state() test constructor in proxy.rs was not updated, so cargo test fails with E0063 and none of the new tests can have been run against this tree. Fix that one line and update the stale PR description about the golden-measurement re-seed.

foxpatch-aleph (approved):
Solid, well-tested implementation that follows the plan closely. I independently verified all three pinned gpu_nonce SHA-256 vectors (they match), confirmed the pipe-draining fix is correct and regression-tested, confirmed decode_nonce preserves the existing SNP route's behavior byte-for-byte, and found no injection or fail-open issues. The findings below are all non-blocking robustness and documentation nits.

**`rust/crates/aleph-attest-agent/src/gpu.rs`** (line 145): The reader-thread joins are not covered by COLLECT_TIMEOUT. The poll loop only bounds the direct child's lifetime; after it exits, `stdout_handle.join()` blocks until read_to_end sees EOF, which requires every inherited write-end of the pipe to close. If the collector (or anything it spawns that inherits stdout) leaves a write-end open, the join blocks forever: a web::block thread is stuck and, worse, the GpuState mutex is held indefinitely, so every subsequent GPU evidence request hangs with no bound. Corner case given the trusted busybox-chroot collector (chroot execs rather than forks), but the whole point of the timeout is bounding a wedged collector — consider a bounded wait (e.g. mpsc + recv_timeout) or a deadline over the whole operation.

**`nix/golden-measurements.json`** (line 2): The PR description contradicts the diff here: it says "The re-seed lands in part D (nix); expect the golden-measurements CI job to be red on A, B and C and green from D on", yet this PR contains the re-seed, and the commit message confirms it is deliberate ("the rebase onto dev-2.1 brought in the aleph_insecure_unattested init changes; both reach the measured initrd"). The code is right — PR A changes the measured agent binary, so it must re-seed here, and B/C don't touch measured inputs — but the stale description sets the wrong expectation for reviewers. The golden CI should be GREEN on this PR; if it is red, the re-seeded values are wrong and need regenerating. Please update the description.

**`rust/crates/aleph-attest-agent/src/main.rs`** (line 61): The GPU route does not fail closed in --insecure-plain-http mode. The SNP endpoint answers 500 through NoTeeBackend, but combining --insecure-plain-http with --gpu-claims/--gpu-collector would serve GPU evidence over unauthenticated HTTP, with a nonce derived from an empty served key. It's a local-testing-only mode and the flags won't be combined in practice (init-gpu.sh comes in part D and runs the attested path), but the stack's stated invariant is "nothing fails open" — either refusing GPU evidence when no attested identity exists, or a clap conflict between the flags, would keep the posture consistent.

**`rust/crates/aleph-attest-agent/src/main.rs`** (line 235): Nit: the mutual `requires` of --gpu-claims/--gpu-collector is untested. The tests module already has the pattern for exactly this kind of clap constraint (plain_http_flag_is_off_by_default_and_parses); adding a case for one flag without the other would pin the "both or neither" rule.

**`rust/crates/aleph-attest-agent/src/gpu.rs`** (line 40): Nit: COLLECT_TIMEOUT is a const, so the timeout path is untested (a test would have to wait 60 s). Making it injectable (a field with this const as the default) would let a fake slow collector pin the kill-and-fail behavior cheaply.

**`rust/crates/aleph-attest-agent/src/main.rs`** (line 143): Nit: boot_claims is served with no shape validation — any JSON value (null, a string) is accepted. Init writes the file in the measured image so it is trusted, but per the plan the init extraction uses a sed pipeline; requiring at least an array here would catch a broken extraction in part D before it reaches clients, rather than serving whatever the sed matched.

foxpatch-aleph (commented):
Solid, well-tested part A of the NVIDIA CC stack. The domain-separated gpu_nonce derivation (SHA-256 over DOMAIN_GPU_NONCE || served_key || client_nonce) is correctly designed: the served-key binding prevents relay across TLS channels, the client nonce gives liveness, and the domain tag prevents collision with the SNP key-bound and fresh schemes; all three pinned vectors independently verified. The collector implementation is careful - separate drain threads for stdout/stderr prevent the >64KiB pipe-buffer deadlock (with a regression test), the 60s timeout kills wedged collectors, parse_output fails closed on non-zero result codes, empty device lists, and nonce mismatches, and the tokio mutex serializes SPDM exchanges as designed. The route correctly 404s on GPU-less runtimes and shares the nonce bounds/hex validation with the SNP route. Two non-blocking observations below; also note the PR description says the golden re-seed lands in part D, but this PR re-seeds all four goldens itself (commit f5ae681), which keeps CI green but makes the description stale and forces B and C to rebase over the new measurements.

**`rust/crates/aleph-attest-agent/src/gpu.rs`** (line 115): The drain threads use unbounded read_to_end, so a wedged or hostile collector that keeps writing would grow guest memory without limit until COLLECT_TIMEOUT fires. Given nvattest runs inside the measured image this is defense-in-depth, but a simple take() with a cap (a few MiB, well above any legitimate evidence size) would bound the blast radius alongside the existing timeout.

**`nix/golden-measurements.json`** (line 2): PR description says 'The re-seed lands in part D (nix); expect the golden-measurements CI job to be red on A, B and C', but this PR re-seeds all four goldens itself in commit f5ae681. Re-seeding here is the better choice (CI stays green), but the description text is stale and B/C must rebase over these values - worth updating the description so the stack coordination is unambiguous.

foxpatch-aleph (approved):
Solid, carefully-tested first PR of the NVIDIA CC stack. The gpu_nonce scheme verifies against independently recomputed vectors, the child-process collector is bounded end to end (injectable timeout, 16 MiB per-pipe cap, deadline-shared pipe drains), the route's error contract matches the spec exactly, and the CLI fails closed in plain-HTTP mode. I verified the try_wait-then-wait pattern is safe against current std (exit status is cached), and that the golden re-seed story is consistent with the crane cargoLock build. Remaining comments are minor robustness/consistency nits; the golden measurement values themselves need CI to confirm since nix is unavailable in this environment.

**`rust/crates/aleph-attest-agent/src/gpu.rs`** (line 188): Minor: the kill path reaps only the direct child. If the collector is a shell/chroot chain that spawns helpers, a SIGKILL to the top process leaves grandchildren running with the GPU and holding the pipes (which the drain threads then wait on). The comment acknowledges surviving grandchildren for the pipe case; consider noting the process-group implication, or spawn with setsid and kill the group, in a follow-up.

**`rust/crates/aleph-attest-agent/src/gpu.rs`** (line 196): Minor: stderr is received with the same shared deadline, so it effectively gets `deadline - (time spent waiting on stdout)`. A child that exits just under the timeout with large output can fail spuriously with `stayed open` even though both pipes closed. This is the documented design (give up at the same deadline), but a tiny grace on the second receive would remove the edge; feel free to leave as is.

**`rust/crates/aleph-attest-agent/src/main.rs`** (line 160): Nit: the `--gpu-claims` file is read uncapped, while the rest of this crate bounds every input (secrets cap keys and values at 64 KiB, inject-secret caps bodies at 2 MiB). The file comes from measured init on tmpfs so the risk is low, but a size cap here would match the crate's fail-closed style.

**`rust/crates/aleph-attest-agent/src/proxy.rs`** (line 161): Nit: `Retry-After: 10` is hardcoded while `lock_wait` is configurable; they only agree because main.rs always passes GPU_LOCK_WAIT. Consider deriving the header from `gpu.lock_wait` (or a comment tying the two) so a future lock_wait change doesn't silently desynchronize the advertised retry.

**`rust/crates/aleph-attest-agent/src/proxy.rs`** (line 169): Nit: `aleph_tee::types::TeeType::NvidiaCc` already exists and serializes to exactly this string (see its test in aleph-tee). Using the enum instead of a bare string would prevent drift between the route response and the wire type the client parses.

**`rust/crates/aleph-attest-agent/src/proxy.rs`** (line 174): Minor test-coverage gap: the route tests cover 200/400/404/503 but not the 500 arm (a `GpuEvidenceSource` whose collect returns Err). A one-line fake would close it, since the 500 body is part of the documented error contract.

foxpatch-aleph (commented):
Solid, carefully-hardened part A. The gpu_nonce scheme and its pinned vectors verify independently, the route error contract (400/404/503/500) matches the spec, the pipe-drain collector correctly avoids the classic pipe-buffer deadlock while bounding child run and drain at one deadline, and all state constructors are updated consistently. No Rust toolchain was available in the sandbox, so the analysis is static. Four non-blocking findings; the most substantive is that the GPU mutex guard is held by the handler future, so a client disconnect mid-collection releases it while the web::block closure still runs, letting a determined client stack concurrent collectors and break the documented one-SPDM-exchange-at-a-time invariant. None of the findings fail open or expose secrets.

**`rust/crates/aleph-attest-agent/src/proxy.rs`** (line 159): The lock guard lives in the handler future, so if actix cancels the handler (client disconnects mid-collection), the guard drops and releases the GPU lock while the `web::block` closure keeps running to completion on the blocking pool. A new request can then acquire the lock and start a second collector, breaking the documented "One SPDM exchange at a time" invariant; a client can deliberately stack concurrent collectors by connecting, acquiring the lock, then hanging up. Consider acquiring the lock inside the blocking closure (pass an owned lock handle through `web::block`) so the serialization tracks the collector's actual lifetime.

**`rust/crates/aleph-attest-agent/src/proxy.rs`** (line 161): The Retry-After value is hardcoded as "10" but the actual wait bound is the configurable `gpu.lock_wait` (tests already vary it to 50 ms). Derive the header from `gpu.lock_wait` so the two cannot drift apart if GPU_LOCK_WAIT changes.

**`rust/crates/aleph-attest-agent/src/gpu.rs`** (line 196): An over-cap stderr fails the collection via `?` even when the child exited 0 and stdout parsed fine, but stderr is purely diagnostic (it is only logged on failure). A collector that is merely chatty on stderr would fail otherwise-successful collections. Fail-closed is defensible, but ignoring a stderr cap error on the success path (with a log line) would be friendlier without weakening anything the client can observe.

**`rust/crates/aleph-attest-agent/src/gpu.rs`** (line 188): `child.kill()` only signals the direct child, but the documented production command line is a busybox chain that may spawn helpers (the survivor test proves grandchildren happen). After a timeout, such helpers keep running in the guest with an open SPDM session until they exit on their own. Spawning the collector with `Command::process_group` (Rust 1.64+) and killing the group would make the timeout effective against helpers rather than only preventing them from pinning the agent's mutex.

PR #1197, opened 2026-09-04
feat(daemon, agent): NVIDIA CC part B, CC-mode probe and capability advertisement

odesenfans:
## NVIDIA CC on SEV-SNP, part B of 5: CC-mode probe and capability advertisement

Stacked on part A (merged as #1196). Design: `docs/superpowers/specs/2026-09-04-nvidia-cc-design.md`.

### What

- Daemon `gpu_cc.rs`: reads each idle NVIDIA card's confidential-computing mode from the BAR0 register NVIDIA's gpu-admin-tools reads (`0x590` Blackwell, `0x1182CC` Hopper, bits `[1:0]`), through a bounds- and alignment-checked read-only mmap of the sysfs `resource0` file. Device-id ranges vendored from `gpu/devid_chips.py`.
- `GpuDevice.cc_mode` on the inventory (omitted when unprobed, so the inventory JSON is byte-identical for today's fleets), proto `GpuDevice.cc_mode = 5`, the Python model and `proto_convert`.
- The probe holds the world read guard for its whole loop. CreateVm registers a VM's cards under the write lock before it boots anything, so a card is either already attached (skipped) or cannot become attached until the probe is done: the register is never read under a guest. Results are cached per card; the 60 s TTL arrives in part E, and on this branch alone the probe runs per `GetHostInfo` (a passive one-page register read on idle cards, microseconds each).
- Agent: `properties.tee.nvidia_cc.devices` lists cards probed `on`, gated on SNP launch capability (a CC GPU on a host that cannot launch SNP is not a capability). A supervisor the agent cannot reach withholds the block with a warning and leaves the rest of `/about/capability` intact. Part E extends the block to `/about/usage/system`.

### Review round (2026-09-08)

- Probe gate: the read guard now spans the probe, not just the attached-set snapshot.
- `/about/capability` no longer fails as a whole when the supervisor is down; `nvidia_cc` is additive.
- The BAR0 read asks the kernel for the page size instead of assuming 4 KiB.
- The little-endian test covers the Hopper offset it used to only write.
- Not taken: moving the 60 s probe TTL down from part E. The per-request cost on this branch is negligible, and moving it would only churn E's rebase.

### Review round (2026-09-09)

- A probe that errors, or reads a register encoding with no mode, now evicts the card's cached mode instead of leaving a stale `on` advertised. Unit test added.
- Not taken: unifying the two world snapshots in `host_info` (the `attached` set for `available_gpus_json` and the probe's own guard). The window predates this PR, any inventory is stale once the response leaves the daemon, and CreateVm re-validates attachments under the write lock.
- The CC probe is now a seam on `DaemonState`: production reads the register, hermetic state never finds a mode, and the `GetHostInfo` integration test injects a probe that answers `on` for every card. No test opens sysfs whatever the host's inventory, and the wire test now checks the attached card was never asked.

### Golden measurements

No measured input changes here: the daemon and agent are host-side. Rebased over the merged part A and dev-2.1's current goldens.

### Testing

Rust unit tests on captured `resource0` fixtures for all three modes and the reserved encoding, an integration test over the gRPC `GetHostInfo`, Python tests for the model, conversion, advertisement, and the supervisor-outage path.


foxpatch-aleph (commented):
This is a careful, well-structured PR: the BAR0 probe is bounds- and alignment-checked, the unsafe mmap is sound, the serde skip keeps the fleet inventory byte-identical, the proto field is backward compatible, and the test coverage is genuinely good (hermetic integration test via the Ada device id outside the CC ranges, injected probe, wire round-trips). Python tests pass. My findings are non-blocking: (1) the probe gate still has a TOCTOU window between reading the attached set and the mmap read, so the stated invariant 'never read under a guest' is narrowed rather than enforced — holding the world read guard across the probe loop would close it, since CreateVm registers the attachment under the write lock before booting; (2) /about/capability on SNP hosts now 500s when the supervisor is down, where it previously never touched the supervisor, and each request triggers a full GPU re-probe on this branch (TTL lands in part E — worth confirming part B never deploys alone); (3) the 4096 page constant in read_bar0_u32 is x86-only-correct; (4) minor dead test data in the gpu_cc.rs little-endian test.

**`rust/crates/supervisor-daemon/src/service.rs`** (line 435): Re-reading the attached set inside the blocking task narrows the race but does not close it: there is still a window between this snapshot and the actual `mmap`/register read in the loop below, during which a concurrent `CreateVm` registers the GPU attachment in the world view (lifecycle.rs takes the world write lock and inserts the entry *before* booting). The probe can therefore still read a register of a card the guest is about to own, which contradicts the module invariant stated in gpu_cc.rs ("the register is never read under a guest") and the comment here ("must never trust a stale snapshot"). Since CreateVm registers the attachment under the world write lock before the boot, holding this `blocking_read()` guard across the probe loop would make the gate airtight at essentially no cost. The impact is benign in practice (the register read is passive and the card is excluded from `available` anyway), hence comment rather than request-changes.

**`src/aleph/vm/agent/resources.py`** (line 316): Before this PR, `get_machine_capability` never talked to the supervisor; now every `/about/capability` request on an SNP-capable host hard-fails with a 500 if the supervisor socket is down or the RPC errors, and on this branch each request also triggers a full BAR0 re-probe of every free NVIDIA card (the 60s TTL arrives in part E — worth confirming part B never ships alone). Since `nvidia_cc` is purely additive information, catching a supervisor failure here and advertising `nvidia_cc=None` (or the tee block without it) would keep the endpoint's availability independent of the new feature. Consider also giving the `supervisor` parameter a type annotation, since it is now part of the function's contract.

**`rust/crates/supervisor-daemon/src/gpu_cc.rs`** (line 127): The hardcoded `4096` page size is only correct on 4K-page systems; on a host with larger pages (e.g. aarch64 with 64K) the computed `base` would not be page-aligned and `mmap` would fail with EINVAL, degrading the probe to unknown. SEV-SNP is x86_64-only so this is theoretical, but `libc::sysconf(libc::_SC_PAGESIZE)` would make it airtight for free.

**`rust/crates/supervisor-daemon/src/gpu_cc.rs`** (line 221): Minor: the write at `bytes[0x1182cc % 0x2000..][..4]` is dead data — the file is only 0x2000 bytes so the Hopper offset is never exercised, and `read_bar0_u32(&path, 0x1182cc)` is never called in this test. Either read it back (with a file of at least 0x1182d0 bytes) or drop the write to avoid implying the Hopper path is covered here.

foxpatch-aleph (approved):
Solid, carefully documented implementation. The BAR0 probe is bounds- and alignment-checked before its unsafe mmap with a correct SAFETY justification, error paths are uniformly fail-safe (log and advertise nothing), the wire format stays byte-identical for fleets without CC cards, and the vm_info_message signature change is propagated consistently. The remaining observations are minor: a small TOCTOU window the code comment slightly overstates as closed, stale cache entries never being evicted, one line of dead test fixture data, and the acknowledged per-GetHostInfo probe cost that part E's TTL addresses.

**`rust/crates/supervisor-daemon/src/service.rs`** (line 435): The doc comment says the attached set is read "immediately before use" so the probe gate never trusts a stale snapshot, but a TOCTOU window remains: between `blocking_read()` returning (releasing the lock) and `probe()` executing its mmap, a concurrent CreateVm can register a new attachment and the card still gets probed under a guest. The window is microseconds and the consequence (one 4-byte volatile read on a guest-owned card) is benign, but if the "never read under a guest" invariant is meant to be airtight, holding the read guard across the probe (blocking CreateVm's write lock for the mmap duration) would close it. Either way, the comment should acknowledge the residual race rather than imply freshness eliminates it.

**`rust/crates/supervisor-daemon/src/service.rs`** (line 454): `Ok(None) => {}` never evicts a previously cached mode. If a card was probed `on` and a later probe reads the reserved 0b10 encoding (or errors), the stale `on` stays cached and advertised indefinitely. Since `probe_cc_mode` also returns `Ok(None)` for non-CC-capable device ids (which never have a cache entry), removing the entry on both `Ok(None)` and `Err` would be safe and would make the cache self-healing.

**`rust/crates/supervisor-daemon/src/gpu_cc.rs`** (line 221): This line writes a Hopper-mode value at `0x1182cc % 0x2000` but nothing ever reads it — the file is only 0x2000 bytes, so `read_bar0_u32(&path, 0x1182cc)` would take the past-the-end error path, which the test never asserts. Either grow the fixture to 0x1182d0 bytes and assert the Hopper offset reads back, or drop the line.

**`rust/crates/supervisor-daemon/src/service.rs`** (line 282): On this branch alone, every GetHostInfo synchronously mmaps every free NVIDIA card's BAR with no timeout. A wedged card whose BAR read hangs would block GetHostInfo — and therefore both /about/usage/system and /about/capability — indefinitely. The PR description notes the 60 s TTL cache arrives in part E's fix wave; just flagging that the parts B and E branches should not ship independently.

**`src/aleph/vm/agent/resources.py`** (line 287): Minor: the `list[dict]` type hint is untyped enough that `gpu["device_id"]` and `gpu.get("cc_mode")` rely entirely on the daemon's JSON contract. A `dict[str, Any]` or a small TypedDict would document the expected shape for future readers.

foxpatch-aleph (approved):
This PR is well-executed: the BAR0 probe's unsafe mmap is carefully bounds-checked, the world-lock discipline around the probe is sound (verified that CreateVm registers GPUs under the write lock before booting, so 'never read under a guest' holds), the wire evolution is genuinely additive (byte-identical inventory for unprobed fleets, tested), and the agent's supervisor-outage path correctly catches the closed SupervisorError vocabulary while update_aggregate_settings swallows its own fetch errors. Test coverage is thorough on both sides. Remaining nits are documentation and observation level.

**`rust/crates/supervisor-daemon/src/service.rs`** (line 417): The doc says a probe error "leaves the card unknown, which advertises nothing", but the cache entry is never removed on error: a card that previously probed `on` and whose probe now fails keeps advertising its stale cached mode, not nothing. Only a never-probed card is truly unknown. Either remove the entry on error, or reword the doc to "keeps its last known value".

**`rust/crates/supervisor-daemon/src/service.rs`** (line 282): Observation: the `attached` snapshot at the top of host_info is taken from a different world-guard acquisition than the one inside refresh_cc_modes, so a card attached between the two is skipped by the probe (good) but still reported as available in this response's available_gpus_json. This TOCTOU pre-dates the PR and is only slightly widened by the probe, and CreateVm re-validates attachments, so it is fine — just noting the two snapshots are not consistent with each other.

**`rust/crates/supervisor-daemon/tests/read_only.rs`** (line 427): The `unprobed` assertion for 0000:01:00.0 implicitly requires that `/sys/bus/pci/devices/0000:01:00.0/resource0` does not exist on the test machine: refresh_cc_modes will really probe that card (10de:2b85 is in the Blackwell range) on every GetHostInfo, and on a dev box with a vfio-bound GPU at exactly that address the probe would succeed and flake the assertion. Extremely marginal, but worth knowing the test is not fully hermetic against GPU-equipped hosts.

foxpatch-aleph (changes requested):
The design is solid and the Python half is fully consistent (models, wire conversion, SNP gating, supervisor-outage handling, tests all check out and pass). However, the final commit of this branch (forget a card's CC mode when its probe stops answering) breaks the new Rust integration test: get_host_info_reports_probed_cc_modes_and_omits_unprobed_ones seeds CcMode::On for the free card 0000:02:00.0, but the same GetHostInfo RPC probes that card, probe_cc_mode returns Ok(None) because 10de:26b1 is an Ada id outside the CC ranges, and the new cache.remove eviction deletes the seeded entry before the inventory is serialized, so both cc_mode assertions in the test fail. The test needs reworking (e.g. make the free card non-NVIDIA so the refresh loop skips it, or make the probe injectable through DaemonState) before merge.

**`rust/crates/supervisor-daemon/tests/read_only.rs`** (line 426): This test fails as written after the eviction change in 40fad4b. The seed at line 397 inserts CcMode::On for 0000:02:00.0, but the GetHostInfo RPC under test first runs refresh_cc_modes (service.rs:282), which probes that card: it is NVIDIA and unattached, and its device id 10de:26b1 is outside the CC ranges, so probe_cc_mode returns Ok(None) and the new cache.remove branch (service.rs:459) evicts the seeded entry before host_info serializes the inventory. probed["cc_mode"] is therefore Null, and this assertion plus the available[0]["cc_mode"] one at line 437 both fail. The commit message's assumption that non-CC device ids 'never hold an entry' is true in production but not for this seeded fixture. Either make the free card non-NVIDIA (the refresh loop skips non-NVIDIA vendors without touching the cache) or plumb an injectable probe through DaemonState so the integration test can pin the mode.

foxpatch-aleph (approved):
This PR is well-executed. The BAR0 probe is carefully bounded (alignment, length, overflow, page-size checks), the lock protocol that guarantees the register is never read under a guest is sound and I verified CreateVm indeed registers GPUs under the world write lock before booting; the probe-gate-cached-mode design and the eviction-on-error semantics are correct and well tested. The agent side correctly gates nvidia_cc on SNP capability and degrades gracefully on supervisor outage. Remaining observations are nits: the available-GPU filter uses an attached-set snapshot taken before the probe, widening a pre-existing race window; identical CC-on cards produce duplicate entries in the advertised devices list; and the sysconf page-size filter would be airtight with a page >= 4 floor for the SAFETY argument. Nothing blocking.

**`rust/crates/supervisor-daemon/src/service.rs`** (line 269): The `attached` snapshot is taken before the CC probe runs and reused afterwards to filter `available_gpus_json`, so the probe widens the pre-existing window in which a card attached by a concurrent CreateVm still shows up as available. You explicitly declined to unify the two snapshots (stale-once-response argument is fair, and CreateVm re-validates under the write lock), but moving this snapshot to *after* the `refresh_cc_modes` await would narrow the window at zero cost and no churn. Non-blocking.

**`src/aleph/vm/agent/resources.py`** (line 296): Two identical CC-on cards produce two identical entries in `nvidia_cc_properties.devices` (same `device_id`, no `pci_host`). If consumers treat the list as a set of distinct capabilities the duplicate is harmless but redundant; if they count entries to gauge capacity, that should be stated in the docstring. Worth confirming intent either way.

**`rust/crates/supervisor-daemon/src/gpu_cc.rs`** (line 130): Minor hardening: the `is_power_of_two()` filter alone permits a hypothetical page size of 2, for which the SAFETY comment's `within + 4 <= page` argument would not hold (a 4-byte read could run past a 2-byte mapping). Real Linux never returns < 4096 here, but a `.filter(|page| page >= 4)` (or is_power_of_two plus a min) would make the safety argument self-contained in the code rather than relying on Linux behavior.

**`rust/crates/supervisor-daemon/src/service.rs`** (line 442): The world read guard is held across the full probe loop, so a pathological BAR read (card in a bad power state returning all-0xFFFF would be fine; a wedged device could stall) would also stall every CreateVm waiting on the write lock, not just tokio workers (which `spawn_blocking` protects). Given each read is microseconds and part E adds a TTL that reduces probe frequency, this is acceptable, but it is an single-host availability coupling worth remembering.

PR #1198, opened 2026-09-04
feat(controller, daemon): NVIDIA CC part C, SNP passthrough argv and the daemon gate

odesenfans:
## NVIDIA CC on SEV-SNP, part C of 5: passthrough argv and the daemon gate

Stacked on part B. Design: `docs/superpowers/specs/2026-09-04-nvidia-cc-design.md`.

### What

- Controller: `snp_gpu_args` emits, per card, `pcie-root-port` plus `vfio-pci,host=...,rombar=0` and once `-fw_cfg name=opt/ovmf/X-PciMmio64Mb,string=<mb>`; no `x-vga`, and the measured `-cpu` model is untouched. A config without GPUs adds no bytes: every existing SNP conformance fixture is unchanged; a new `gpu.json` fixture pins the GPU form.
- Daemon `gpu_bar.rs`: the OVMF 64-bit MMIO window is sized from the card's sysfs BARs (64-bit prefetchable memory BARs, next power of two, doubled, 1 GiB floor), never a constant, and carried to the controller as `pci_mmio64_mb`.
- `snp_config_slice` admits GPUs only when every card is in the inventory and reads `on` from its BAR0 register at create time, through part B's probe seam, never from the inventory cache; anything else, including a card that cannot be read, stays `InvalidBackend`. The read is safe there: the cards were validated unattached and the gate runs under the world write lock with creation serialized. Part E's fix wave additionally restricts admission to the measured verity arm.
- `{rootfs}.cmdline_extra` sidecar: the runtime's fixed `swiotlb=262144` token reaches the measured cmdline through an allowlist (`swiotlb=<digits>` only), spliced between `workload_roothash=` and `verified_volumes=`, the template's position.

### Review round (2026-09-09)

- The gate probes the card instead of trusting the cache: a card switched off between two host-info polls was fail-open. Tests inject the probe alongside the window reader and cover the off, devtools, no-mode, unreadable and cached-only arms.
- The `swiotlb=` allowlist refuses a leading zero: the kernel parses the value with base auto-detection, so `0262144` would have been octal.
- The MMIO window reader takes its sysfs root as a parameter and is tested against a fixture devices tree (two cards summed before rounding, domain-less pci_host, missing resource file).
- The argv builder's no-window arm is pinned: root ports and vfio devices, no fw_cfg entry.

### Testing

Controller argv unit tests and parity fixtures; daemon gate tests for on/off/devtools/no-mode/unreadable/absent cards and both sidecar arms; the Python forward-compat config test unchanged.


foxpatch-aleph (approved):
Solid, well-documented increment that matches the design spec byte-for-byte. The controller's snp_gpu_args correctly emits root ports plus vfio-pci and the OVMF MMIO window while deliberately preserving the measured -cpu model and dropping x-vga; the daemon gate is fail-closed (inventory + cached CC mode), the BAR-driven window sizing math and Linux IORESOURCE flag constants are correct, and the cmdline_extra allowlist is airtight (strict swiotlb=<digits> prefix, trimmed, capped sidecar). Tests cover the gate's on/off/devtools/unknown/absent arms, the argv form, a new fixture, and both sidecar arms; docs and divergence entry 68 are updated accurately. Remaining findings are minor coverage gaps and operational notes, none of which block.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 2058): Operational note: the gate trusts the create-time cached CC mode. `refresh_cc_modes_with` never re-probes attached cards and keeps the last value on a probe error, and the check is not repeated at launch — so a card cached `on` that was later switched `off` would still pass. Fail-closed on the cold start (cache miss -> InvalidBackend) is right, but the stale-`on` direction is fail-open; part E's on-demand probe should also decide whether to re-verify at launch. Worth a short comment here so the freshness assumption is explicit.

**`rust/crates/supervisor-controller/src/qemu.rs`** (line 141): Missing test for the `pci_mmio64_mb == None` arm: GPUs present but no window means no `-fw_cfg` is emitted and OVMF falls back to its (too small) default 64-bit MMIO window, which surfaces as a BAR-allocation failure in the guest. That path is reachable via a hand-written or old-daemon config; a one-line `snp_gpu_args(&gpus, None)` assertion alongside the existing ones would pin it.

**`rust/crates/supervisor-daemon/src/gpu_bar.rs`** (line 50): `gpu_mmio64_mb` itself (the sysfs path construction + file read) is untested — the lifecycle test injects the reader and only `parse_resource_file` is unit-tested. Given `gpu_cc.rs` already demonstrates the temp-dir pattern for `resource0`, a small test driving `gpu_mmio64_mb` against a fake devices directory would close the gap.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 2200): Nit: the allowlist accepts leading zeros, but the kernel parses `swiotlb=` with base 0 (`simple_strtoul`-style), so `swiotlb=0262144` would be interpreted as octal, silently mismeasuring the bounce buffer. The trusted agent writes a fixed value so this can't happen today; rejecting a leading `0` would make the allowlist match what the kernel will actually do.

foxpatch-aleph (approved):
Part C is correct and well-tested: the SNP argv builder emits exactly the design-doc device form (root port, vfio-pci, fw_cfg window) without touching any measured input, the MMIO window is sized from real sysfs BARs with correct IORESOURCE flag decoding, and the daemon gate fails closed on every non-CC state. The sidecar refactor is behavior-preserving and the swiotlb allowlist is genuinely closed. Remaining comments are non-blocking observations for the author.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 2083): The gate trusts the `gpu_cc_modes` cache, which is only refreshed by `get_host_info`. On a freshly restarted daemon (or an agent that skips GetHostInfo), the first SNP+GPU create fails closed even for a card in CC mode. The PR description says part E adds on-demand probing, so this is presumably the intended stack ordering — but between parts C and E landing, an SNP+GPU create requires a preceding GetHostInfo call to succeed. Worth a note in the code (or a pointer to part E) so the coupling is not lost.

**`rust/crates/supervisor-daemon/src/gpu_bar.rs`** (line 36): Two nits in one line: (1) a line whose flags say 64-bit prefetchable MEM but `end < start` is silently skipped rather than rejected — malformed sysfs geometry would silently under-size the window (fail-closed in effect since the GPU won't start, but silently); (2) `end - start + 1` overflows (panic in debug) if `end == u64::MAX`. `checked_add` on the `+1`, or rejecting the malformed line, would make both airtight. Not reachable with real sysfs today.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 2140): The opaque arm passes the agent cmdline through verbatim, so it never receives the `cmdline_extra` splice — an SNP+GPU VM on the opaque (LUKS) arm boots only with whatever swiotlb sizing the agent rendered. The daemon gate admits GPUs on both arms without enforcing the token. If that's intentional (GPU runtime is the verity arm until part E), a one-line comment on the opaque arm return saying so would prevent a future reader from assuming the window sizing is the only GPU-specific handling needed.

**`rust/crates/supervisor-controller/src/qemu.rs`** (line 124): Minor coverage gap: `snp_gpu_args` with GPUs present and `pci_mmio64_mb: None` (devices emitted, no `-fw_cfg`) is untested — that's the branch an operator-crafted or half-adopted config could hit, and it's the one combination the unit tests skip.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 2072): The inventory-membership check duplicates `validate_spec_gpus` (lifecycle.rs:1602), which already ran on the create path with `InsufficientResources`. As defense-in-depth for direct callers it's fine, but note the two error codes differ, so a non-inventory GPU on an SNP spec reports `InvalidBackend` here while the same GPU on a plain spec reports `InsufficientResources` — harmless, just worth knowing if anyone greps logs.

foxpatch-aleph (approved):
This is a careful, well-executed increment. The create-time GPU gate correctly probes the card's BAR0 register through the injected seam rather than trusting the inventory cache, fails closed on every non-On answer (off, devtools, reserved encoding, unreadable, absent-from-inventory), and updates the cache either way; its safety argument (probe under creation_lock plus the world write lock, after validate_spec_gpus confirmed the cards unattached) checks out against the actual call path, and it cannot race refresh_cc_modes since that holds a world read guard across its own probes. The MMIO window sizing matches the kernel's real IORESOURCE_MEM/PREFETCH/MEM_64 flag values and the design doc's power-of-two-doubled rule, with the sum taken across cards before rounding; the argv builder emits root port + vfio-pci + fw_cfg exactly as the design specifies, deliberately avoids gpu_args' -cpu host and x-vga, and adds zero bytes for GPU-less configs (fixture pinned). The swiotlb allowlist is a genuine closed allowlist — digits only, no leading zero (octal under base-0 kernel parsing), length-capped — spliced at the template's position between workload_roothash and verified_volumes. Test coverage is thorough across both arms and all gate outcomes, and the two docs are consistent with the code. Only non-blocking nits remain.

**`rust/crates/supervisor-daemon/src/gpu_bar.rs`** (line 38): Minor robustness nit: `end - start + 1` can overflow u64 on a degenerate line (e.g. start=0x0, end=0xffffffffffffffff with matching flags), which panics in debug builds and wraps to 0 in release. The input is a kernel-provided sysfs file for a card in the inventory, so this is not attacker-reachable, but `end.saturating_sub(start).saturating_add(1)` (or a checked_add) would make the debug-build panic impossible. Relatedly, a line with `end < start` is silently skipped rather than an error, while a short line is a hard error — fine given sysfs invariants, but a brief comment saying the skip is deliberate would help.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 2243): Non-blocking, tracking note: in the opaque-cmdline arm (tee.kernel_cmdline set, line 2143) a staged {rootfs}.cmdline_extra sidecar is silently ignored while the GPU is still admitted and the MMIO window sized — so a GPU VM on that arm could launch without the swiotlb token and no error surfaces. The agent owns the opaque cmdline so this is defensible, and part E's restriction of GPU admission to the measured verity arm closes it; flagging so the interim gap is on the record and part E doesn't drop it.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 2085): Observation: the BAR0 probe (blocking mmap + volatile read, plus the resource-file reads for the window) now runs inside the world write lock, since build_written_config is invoked in the write-guard scope at line 2740. That placement is the load-bearing part of the safety argument (the probe cannot race refresh_cc_modes' read-guard probes or a concurrent create), and the reads are single registers, but a hung vfio device would now stall every world-lock RPC for the probe's duration rather than just this create. The code comment covers the reasoning; if probe latency under a wedged card ever becomes a concern, the escape hatch is a timeout in gpu_cc::read_bar0_u32 rather than moving the probe.

PR #1199, opened 2026-09-04
nix: NVIDIA CC part D, the gpuImage flavor

odesenfans:
## NVIDIA CC on SEV-SNP, part D of 5: the gpuImage flavor

Stacked on part C. Design: `docs/superpowers/specs/2026-09-04-nvidia-cc-design.md`. Depends on part A at runtime (init passes `--gpu-claims` and `--gpu-collector` to the agent).

### What

- `gpuKernel`: the whitelist kernel plus a GPU fragment (MMU notifier, dma-buf, PCIe hotplug/ASPM, resizable BAR, firmware loader; DRM and VT off). Base kernel store path unchanged.
- `nix/nvidia.nix`: NVIDIA 595.71.05 open modules (`nvidia.ko`, `nvidia-uvm.ko` only) built against `gpuKernel`, the raw unpatched driver userland at `/opt/nvidia/lib` (container-toolkit contract: workloads link against it with their own libc), and the Ampere-and-newer GSP firmware. Unfree allowance is a narrow predicate.
- `nix/nvat.nix`: NVIDIA's attestation SDK (`libnvat` 1.2.2 and `nvattest`) built offline from pinned sources, including the Rust `regorus` engine through corrosion with a committed lock; build-path and vendored-crate references stripped from the runtime closure.
- `nix/init-gpu.sh`: with an NVIDIA card on the bus, init points the kernel's firmware search path at the rootfs copy of the GSP blobs, loads the modules, runs `nvattest attest --verifier local` against NVIDIA's RIM and OCSP services, requires every claim's `measres` to be `success` (counted against the claim objects), sets the ready state with `nvidia-smi conf-compute -srs 1` and reads it back (negative pattern first), hands the claims to the agent, and bind-mounts the driver userland into the workload chroot. Every GPU failure powers off. Without a card the image boots on like the base image and serves no GPU route; a client whose manifest declares a `gpu` block requires GPU evidence and gets none, which is where that case fails closed. In `aleph_insecure_unattested=1` mode the agent starts through the shared plain-HTTP path and never receives the GPU flags, which it refuses next to `--insecure-plain-http`.
- `gpuImage`, `gpuMeasurement`, `gpu.json` for the bundle builder; `boot-smoke.sh --gpu` proves the fail-closed no-GPU path; the golden check gets `--base-only`/`--gpu-only` and the GPU measurement runs in its own gated CI job.

### Review round (2026-09-09)

- The GSP firmware was unreachable: the blobs live in the verity rootfs, but the kernel resolves firmware requests against the initramfs root whatever chroot asks, so the first open of the card would have failed and every GPU boot powered off. init now sets the firmware loader's runtime search path to the rootfs copy before the insmod and checks it stuck. The blobs stay out of the measured initrd.
- `measres` is lowercase in the pinned SDK (`success`, `fail`, `not-run`, `absent`, see `nv_attestation/claims.h`); the init grep was right, the spec and plan said `Success` and are corrected.
- The no-GPU path continues rather than powering off, as the plan and the smoke test intended; the description above and the former measurement caveat said otherwise and are corrected.
- The GPU CI gate's input list gains `nix/kernel-config.fragment`; the full nvattest result and its log get the same 0600 as the extracted claims.
- Rebased over parts B and C and over the `aleph_insecure_unattested=1` mode: `boot-smoke.sh --gpu` now runs as a phase of the reworked script, and init-gpu.sh starts the agent through `start_attest_agent` in that mode.

### Review round 2 (2026-09-09)

- One device node per card: init counts the NVIDIA display-class devices on the bus and creates `/dev/nvidia0..N-1`, minors in probe order. A V-PROGRAM can own several cards and a lone `/dev/nvidia0` hid the rest from the workload.
- The GPU golden gate lists every measured input directly (ovmf.nix, the root Cargo.toml, the attest-agent and aleph-tee crates as prefixes), matching the workflow's trigger paths.
- nvidia-smi does not need libstdc++: readelf on 595.71.05 shows nvidia-smi, libnvidia-ml and libcuda need only the five glibc sonames, so the loader path is sufficient. The userland derivation now asserts that, so a driver bump that starts needing the gcc runtime fails the build rather than powering the guest off.
- The plan's `"Failure"` grep reads `"fail"` like the SDK.
- Not taken: deduplicating the two GPU fragment lines already in the base fragment. The concatenated config is identical either way, and the edit would rebuild the GPU kernel for a cosmetic change; the fragment doubles as the list of what the driver needs.
- `gpuMeasurement` re-seeded: init-gpu.sh is in the measured initrd, so the device-node change moves it. The base four are untouched.

### Review round 3 (2026-09-09)

- A workload that cannot get the driver userland powers off: a failed `/opt/nvidia/lib` bind into the workload chroot, or a workload image without the mount point, used to warn and leave the guest serving GPU evidence for a workload that could not use the card. Both arms are fatal now; the base flavors never reach the bind.
- The GPU golden job boots the GPU image without a GPU after the measurement compare (`boot-smoke.sh --gpu`, same KVM recipe as boot-smoke.yml), the first CI coverage of init-gpu.sh's no-GPU branch and marker lines.
- The gate lists the two udhcpc scripts, which every initrd ships.
- The plan's part-A snippets quote the implemented attest-agent test again (`Success` is an opaque fixture there); aligning proxy.rs is a follow-up.
- All five goldens re-seeded: init-common.sh is in every measured initrd.

### Review round 4 (2026-09-09)

- The GPU job's gate also lists what the job runs: the check script, the smoke script and the workflow file.
- The attestation result and log are created 0600 under a subshell umask instead of chmod'ed after nvattest returns; `gpuMeasurement` re-seeded.
- Kept: the fatal `/opt/nvidia/lib` mount-point check fires on the no-GPU boot as well. The GPU image is only launched for a manifest with a `gpu` block, and that block requires the workload to ship the mount point, so a workload without it is a manifest-to-image mismatch whatever the bus holds. Failing at the first boot beats a warning that surfaces once a card turns up. The smoke test boots the platform rootfs with no workload and is unaffected.
- Noted, no change: the subset golden modes verify what they build and never diff the whole file; the weekly all-mode run catches a lingering entry.

### Golden measurements

All four base measurements move (`init-common.sh` gained the driver bind), plus the new `gpuMeasurement`; seeded once at the tip of the rebased branch. Every runtime bundle (platform, compose, exec, instance) must be republished with the release.

### Untested on hardware

The firmware search-path handoff (the loader parameter is documented, the open driver's request timing is read from its source), the exact `nvidia-smi conf-compute -grs` wording, whether `AMD_IOMMU` and the hotplug options are inert or needed, the hardware-model string for `accepted_models`, and the GPU golden job's duration (timeout is a guess at 120 minutes). See `docs/operators/nvidia-cc.md` (part E) and the Tier 2 checklist in the plan.


foxpatch-aleph (changes requested):
The plumbing is impressively careful (faithful init.sh cloning, correct golden-measurement subset logic, a collector command that matches the agent's whitespace-split contract, rootfs reproducibility levers matching rootfs.nix), but two findings make the GPU flavor non-functional on real hardware. First, the GSP firmware ships only in the verity rootfs (/mnt/root/lib/firmware/nvidia), while the kernel's firmware loader resolves request_firmware() against the initramfs root at insmod time — the open module requests GSP firmware during PCI probe, so the load fails with ENOENT and every GPU-present boot powers off at the nvattest stage; the design explicitly keeps the firmware out of the initrd but no bridge (bind-mount or firmware_class.path) exists. Second, the fail-closed measres gate greps for lowercase "success", while both the design spec and this PR's own plan specify the value as "Success" — if the plan/spec reflect the pinned CLI, measres_ok is always 0 and every verified GPU boot fatals with "measurement comparison failed". Both failures are fail-closed (the runtime powers off rather than serving unverified GPU claims), so there is no security hole, but the feature would never come up on a host. Also worth fixing: the PR description's claim that a GPU-less boot powers off contradicts the implemented (and planned) no-GPU path, and the CI gate's gpu_inputs list omits nix/kernel-config.fragment.

**`nix/init-gpu.sh`** (line 215): The GSP firmware is unreachable at insmod time. `request_firmware()` resolves `nvidia/...gsp*.bin` against the kernel's current root — the initramfs — not the chroot-mounted verity rootfs, and the open module requests GSP firmware during PCI probe, i.e. synchronously inside `insmod` when the device is already present. The blobs ship only in gpuRootfs (`/mnt/root/lib/firmware/nvidia`, gpu-rootfs.nix:39); initrd.nix ships only the two .ko files. So probe fails with ENOENT, the device never comes up, and every GPU-present boot powers off at (or before) the nvattest stage. The kernel fragment's comment ("Firmware loading for the GSP blob from the rootfs") and the spec's "not the initrd" decision show the intent, but nothing bridges the two. Either bind-mount `/mnt/root/lib/firmware` over `/lib/firmware` (or set `firmware_class.path`) before the insmod, or ship the blobs in the initrd after all.

**`nix/init-gpu.sh`** (line 274): Case mismatch on the measres gate: this greps for lowercase `"success"`, but both the design spec (§6.4: `measres == "Success"`) and this PR's own plan sketch grep for `"Success"`/`"Failure"` with a capital S. If the plan/spec reflect the pinned nvattest CLI's actual output, `measres_ok` will be 0 while `claims_count > 0`, and every *verified* GPU boot fatals with "measurement comparison failed". Please reconcile against the pinned nv-attestation-cli source (the stringification of `nva_measurement_result` in its JSON dump) before merge — this is the core fail-closed gate, and unlike the acknowledged `-grs` wording, the measres value is statically checkable against the vendored sources.

**`nix/init-gpu.sh`** (line 311): Note: the PR description says "a GPU runtime without a GPU on the bus powers off", but this path deliberately continues without GPU attestation (matching the plan, line 2045, and the boot-smoke --gpu expectation that this exact line prints). The code is right; the description's security claim should be corrected so readers don't rely on a fail-closed behavior that doesn't exist (fail-safety here is deferred to the client requiring GPU evidence).

**`.github/workflows/golden-measurements.yml`** (line 136): `nix/kernel-config.fragment` is a direct input to `gpuKernel` (kernel.nix concatenates it with the GPU fragment before the line-by-line config check), but it's missing from `gpu_inputs`. It's benign today — a change there also moves the base measurement, so the base job fails and a full `--update` re-seeds the GPU entry too — but since the list presents itself as the GPU chain's input set, adding it costs nothing and keeps the gate self-contained.

**`nix/gpu-rootfs.nix`** (line 43): Minor: /run/aleph/gpu-attest.json holds the full attestation result (including the detached EAT) at default perms while init-gpu.sh chmod 0600s the extracted claims file. Since the claims were deemed worth 0600, the fuller document arguably deserves the same treatment (only the initramfs is visible there, so low impact).

foxpatch-aleph (commented):
A careful, consistently fail-closed addition: the gpuKernel fragment, the offline nvat build (pinned FetchContent inputs, committed Cargo.lock, path remaps), the raw driver userland extraction, and the init-gpu.sh stage are all well-reasoned and heavily documented, and the golden-measurement/CI split is wired correctly (I verified the subset-extraction sed against the committed golden file and the measured cmdline matches boot-smoke byte-for-byte). The claims-counting validation and the negative-first ready-state readback are genuinely good. The findings below are observations rather than clear bugs: the nvidia-smi loader-path question and the CI gate omissions both stem from the parts of this PR that are explicitly untested on real hardware, and every one of those paths fails closed (poweroff) rather than open.

**`nix/init-gpu.sh`** (line 207): gpu_nvattest and gpu_smi only put /opt/nvidia/lib:/opt/nvidia/glibc on LD_LIBRARY_PATH. nvidia-smi is a C++ binary and its DT_NEEDED commonly includes libstdc++.so.6 and libgcc_s.so.1 in addition to the glibc sonames; those exist in the chroot only inside the copied /nix/store closure (via nvat), which is not on any loader path. If the 595 binary carries those sonames, gpu_smi fails at exec and every GPU boot power off via gpu_fatal. Since this is fail-closed it is not a security issue, but it is worth confirming on the first Blackwell host — or pre-emptively adding the gcc runtime dirs to the rootfs alongside /opt/nvidia/glibc the same way glibc is pinned.

**`.github/workflows/golden-measurements.yml`** (line 137): gpu_inputs omits nix/ovmf.nix, rust/Cargo.toml and rust/crates/aleph-attest-agent/** — all three are measured inputs of gpuMeasurement (sev-snp-measure hashes OVMF.fd, and the agent binary is baked into gpuInitrd). Today the gate stays correct only indirectly: those files also move the base measurements, so the base job fails or the golden file is re-seeded, and nix/golden-measurements.json is in the list. That backstop works, but it is subtle enough that a one-line comment (or just adding the three paths) would keep the direct list honest for future readers.

**`nix/init-gpu.sh`** (line 234): Only /dev/nvidia0 (minor 0) is created, while the claims-counting logic above explicitly handles multiple per-GPU claims. If more than one NVIDIA device ever sits on the PCI bus, GPUs 1..N verify fine but get no device node, so the workload cannot see them. If single-GPU-per-VM is the contract, a one-line comment saying so (next to the mknod block) would make the asymmetry deliberate; otherwise a loop over the detected PCI devices is needed.

**`docs/superpowers/plans/2026-09-04-nvidia-cc-aleph-vm.md`** (line 2027): This review round corrected the "Success" casings, but this grep for '"measres" *: *"Failure"' is dead the same way: the SDK serializes the enum lowercase (fail), so the pattern can never match. Harmless since the plan is a historical document and the implementation replaced this approach with the counting logic, but for consistency with the correction note it should read "fail".

**`nix/kernel-config-gpu.fragment`** (line 25): CONFIG_PCI_MSI=y and "# CONFIG_VT is not set" are already in the base fragment, so the concatenated GPU config repeats them. Harmless (KCONFIG_ALLCONFIG and the line-by-line check both accept duplicates), but the GPU fragment could stay minimal and rely on the base whitelist.

**`nix/gpu-rootfs.nix`** (line 75): The +64 MiB headroom over du -sm (base rootfs.nix uses +10) is fine today, but the driver userland plus nvat's closure is several hundred MB; a driver bump that pushes past the headroom fails the build at mkfs.ext4 loudly, which is the right failure mode — just worth knowing that the failure will surface here rather than as a small diff.

foxpatch-aleph (commented):
Well-executed part of the NVIDIA CC stack: the GPU init stage is consistently fail-closed, the kernel/measurement machinery cleanly isolates the flavor from base outputs, and the nvat build is fully pinned and offline. I verified the claims extraction and counting logic by simulation (correctly rejects fail/not-run/missing-measres and empty claims), the subset-mode sed against the committed golden file, and that the --gpu-collector string matches the agent's split_whitespace-plus-append-nonce execution model. All findings are non-blocking gaps/nits, the main one being that the new CI gate omits two files that move gpuMeasurement.

foxpatch-aleph (commented):
A high-quality, carefully-reasoned addition: fail-closed GPU verification in init-gpu.sh (verified to be a faithful copy of init.sh plus the one GPU stage), clean measurement isolation between the base and GPU chains, a genuinely offline nvat build, and a correct subset-comparison design in check-golden-measurements.sh. No correctness or security bugs found. Four minor, non-blocking observations: the GPU CI gate's input list omits the udhcpc scripts (which are baked into gpuInitrd and thus move gpuMeasurement), the plan doc's edited part-A test snippets now diverge from the actual proxy.rs code, the driver-userland bind-mount into the workload chroot warns rather than failing like every other GPU error, and the --gpu boot smoke is local-only so CI never exercises init-gpu.sh.

**`.github/workflows/golden-measurements.yml`** (line 139): The gpu_inputs list is missing nix/udhcpc.script and nix/udhcpc6.script. Both are baked into gpuInitrd (initrd.nix lines 102-103) and therefore move gpuMeasurement, but a PR touching only those files would hit the "no GPU input changed; skipping" branch and never run the GPU golden check. The weekly cron backstop catches it eventually, but the gate's comment ("Inputs that can move gpuMeasurement") currently overpromises. Adding the two paths makes the list complete against everything initrd.nix ships for this flavor.

**`docs/superpowers/plans/2026-09-04-nvidia-cc-aleph-vm.md`** (line 468): This edit (and the one at line 482) changes the part-A proxy.rs test snippet to "success", but rust/crates/aleph-attest-agent/src/proxy.rs:407/422 still uses "Success". The fixture is passthrough so it doesn't affect behavior, but the plan now misquotes the code it documents — either keep the plan matching the implemented test, or align proxy.rs to the real SDK spelling in a follow-up.

**`nix/init-common.sh`** (line 190): Every other failure in the GPU stage powers the VM off, but a failed /opt/nvidia/lib bind into the workload chroot only warns — the guest then serves GPU attestation evidence while handing the workload the /dev/nvidia* nodes (which arrive via the /dev bind) without the driver libraries it needs to use them. Consider making this fatal on the GPU flavor, or at least confirming the console WARNING is surfaced somewhere an operator will see it before a client is confused by an attested GPU that can't run CUDA.

**`nix/boot-smoke.sh`** (line 18): Since --gpu is local-only, CI never builds the GPU image and the no-GPU branch of init-gpu.sh (plus the marker strings this script greps for) can regress without any CI noticing. The GPU chain build cost makes this a defensible trade, but a typo in "init: no NVIDIA GPU present; running without GPU attestation" would only surface when someone happens to run the smoke locally — worth either a note in the part-E Tier 2 checklist or, if the golden-measurements-gpu runner already pays for most of the chain, considering a smoke phase there.

foxpatch-aleph (commented):
A careful, fail-closed implementation: init-gpu.sh is a clean copy of init.sh plus one self-contained GPU stage, the nvattest claims extraction and per-claim measres enforcement are correct (verified empirically with busybox), the gate list covers everything that can move gpuMeasurement, and the agent-side collector contract matches. All GPU failures power off, which is the right posture. The three observations below are non-blocking: a gap in the GPU job's own gate list (changes to boot-smoke.sh / check-golden-measurements.sh / the workflow never exercise the GPU job), the /opt/nvidia/lib fatal in prepare_chroot firing even when no GPU was attested, and a brief world-readable window on gpu-attest.json before its chmod 0600.

**`.github/workflows/golden-measurements.yml`** (line 139): The gpu_inputs gate list covers everything that can move gpuMeasurement, but the job also runs `./nix/check-golden-measurements.sh --gpu-only` and `./nix/boot-smoke.sh --gpu` (plus its own QEMU/KVM steps) — none of nix/boot-smoke.sh, nix/check-golden-measurements.sh, or this workflow file is in the list, so a PR changing only those skips the GPU job entirely and first exercises the change at the weekly cron. Consider adding them, since the PR explicitly describes this job as the only CI coverage of init-gpu.sh's no-GPU branch and marker lines.

**`nix/init-common.sh`** (line 193): This fatal is keyed on the image shipping the driver userland, not on gpu_claims being set: on the GPU image's no-GPU boot path (gpu_total == 0), a workload image without the /opt/nvidia/lib mount point is fatally rejected even though no GPU was attested — a client with no attested GPU at all, versus the "attested GPU that cannot run CUDA" case the comment motivates. That is a defensible strict reading of the image contract (and matches the fail-closed bias of the flavor), but it's broader than the stated rationale; worth confirming it's intended that non-GPU workload images can never run on the GPU flavor.

**`nix/init-gpu.sh`** (line 271): gpu-attest.json (and the .log) are created by the redirection with the default umask, so they sit world-readable until the chmod 0600 at line 278, after nvattest exits. No competing process exists in the guest in that window (the workload and agent start later), so this is not exploitable today — but creating the files 0600 before invoking nvattest would close the window for free.

**`nix/check-golden-measurements.sh`** (line 103): Note that the subset modes (--base-only / --gpu-only, which are the only modes CI runs) never diff the golden file itself, so an entry for an output that is no longer produced lingers unnoticed until a weekly-cron `--update` / all-mode run. That is consistent with the documented intent (subset runs verify what they build), just worth knowing when adding or removing outputs.

foxpatch-aleph (approved):
Solid, carefully fail-closed implementation. I verified the gpuMeasurement input graph is fully covered by the CI gate list, the claims extraction/counting logic is sound for nlohmann dump(4) output, the firmware search-path handoff is correct (kernel firmware requests resolve against the init namespace, so the /mnt/root copy is what the loader finds), the prepare_chroot double-preparation guard handles all GPU x workload boot combinations correctly, and init-gpu.sh's agent invocation matches the part-A CLI contract (gpu-claims requires gpu-collector; both conflict with --insecure-plain-http). The golden re-seeding is consistent (init-common.sh is in every measured initrd, so all four base measurements legitimately move). Only minor observations follow; none block.

**`.github/workflows/golden-measurements.yml`** (line 89): Neither CI job ever runs check-golden-measurements.sh in "all" mode anymore (base job uses --base-only, GPU job uses --gpu-only, including on the weekly schedule), so the structural file diff that catches a lingering or missing golden entry is dead in CI. The PR's round-3 note ("the weekly all-mode run catches a lingering entry") is therefore no longer accurate. In practice this is cosmetic -- --update always writes the full base+gpu set -- but consider running the base job without the flag on schedule (cost is the same two subset runs combined), or update the note.

**`nix/init-gpu.sh`** (line 330): The negative ready-state pattern can miss a key/value-style readback such as "Ready: false" or "Ready state: no": the negative alternation (not-ready|disabled|off) wouldn't match, and the positive grep would match the key and pass. Since the exact -grs wording is explicitly unverified, consider adding false|\bno\b to the negative alternation so that reading fails closed until tuned on hardware (matching the stated intent that every ambiguous reading is fatal).

**`nix/init-gpu.sh`** (line 283): Minor hardening: unlike the claims extraction right below (which carefully reasons about indent depth), this grep is not anchored, so it matches a "result_code": 0 at any nesting depth. If the claims ever embed a per-device result_code while the top-level is nonzero, the grep would pass on the nested zero. Anchoring to '^    "result_code"' would bind it to the top level for free, consistent with the surrounding analysis.

**`nix/nvidia.nix`** (line 43): Small comment nit: the list is a fixed enumeration (libcuda.so.${version}, ...), not wildcards, and a renamed library would fail the cp, not "the ls". The behavior is the desired one; only the wording is off.

PR #1200, opened 2026-09-04
feat(vprogram, agent): NVIDIA CC part E, manifest gpu block, agent resolution, docs

odesenfans:
## NVIDIA CC on SEV-SNP, part E of 5: manifest, agent, docs, and the final-review fixes

Last part of the stack, rebased onto dev-2.1 with parts A to D merged. Design: `docs/superpowers/specs/2026-09-04-nvidia-cc-design.md`.

### What

- Runtime manifest `gpu` block (`GpuRuntimeSpec`: vendor, arch, driver_version, accepted_models, library_path) and the `gpu` bundle flavor; the cmdline template validator now pins every placeholder token to its key (`roothash={platform_roothash}` and friends) and allows only the closed fixed-token set, closing a kernel-parameter smuggling channel that also affected the existing exec and luks templates.
- Agent: `resolve_confidential_gpus` matches only cards probed `on`; `build_vprogram_spec` fails closed on more than one GPU, a runtime without a `gpu` block, a vendor mismatch, or less than 2048 MiB; the `cmdline_extra` sidecar is written from the template; `run.py` resolves the card after staging.
- `tee.nvidia_cc` is advertised on `/about/usage/system` as well as `/about/capability` (one shared builder).
- Whole-branch fixes that survived the merged parts: GPUs refused on the opaque-cmdline SNP instance arm; the CC-mode sweep bounded to one per minute when no card changed hands (the gate probes the card itself since part C); `result_code` check anchored to the top level; MMIO window clamped at 4 TiB; Turing GSP blob dropped; an ETXTBSY test flake fixed. The collector cap and the on-demand gate probe landed in parts A and C.
- `docs/operators/nvidia-cc.md` runbook and the architecture notes in `docs/architecture/confidential.md`.

### Update 2026-09-07: the message asks for a GPU family and a count

aleph-message #162 was reshaped: `content.gpu` is `ConfidentialGpuRequirement { vendor, arch: hopper|blackwell, count: 1..=8, models: optional narrowing to device ids, mode: "cc" }` instead of a one-entry `gpus` list of exact device ids. Four commits follow it here:

- Daemon: each inventory card carries its `arch`, derived from the device-id table the BAR0 probe already uses (`GpuDevice.arch = 6` on the proto, `skip_serializing_if` on the JSON inventory so a fleet without known cards keeps its bytes). The agent keeps no table of its own.
- Agent: `NvidiaCcDevice.arch` is advertised on both `/about` endpoints; `resolve_confidential_gpus(arch, count, models, owner)` picks `count` distinct CC-mode cards of the family, honouring holds and consuming the owner's only once the whole requirement is met; the shortage error names `confidential_gpu` with the family and the candidates.
- Launch: a count above one is refused as `VmSetupError` (single-GPU passthrough is what NVIDIA validates on the RTX PRO 6000 Blackwell Server Edition), as is a manifest that drives another architecture.
- Docs: spec section 5.1, architecture notes and runbook updated; NVIDIA's mode matrix (SPT, PPCIe, MPT) and the MIG/vGPU-under-CC status recorded.

### Rebase and review round (2026-09-09)

- Rebased over the merged A to D and dev-2.1's retire path, allocation endpoints and `aleph_insecure_unattested=1` mode. Dropped what those parts already carry (collector cap, on-demand gate probe, probe eviction) and the no-GPU poweroff, which part D's review settled the other way: the spec and the plan now describe the continue path, with the client requiring GPU evidence whenever the manifest declares a `gpu` block.
- A cmdline template token may appear only once: a repeated `swiotlb=262144` passed the validator and failed at create against the daemon's one-token allowlist.
- `/about/usage/system` refreshes the settings aggregate once per poll; the gpu block and the tee block share the map.
- The confidential-GPU candidate filter reads as three gates.
- Not taken: a timeout around the gate's register read. A read from a mapped BAR completes or hits the root port's completion timeout within milliseconds; it is not a process that can hang, and part C's merged gate does the same read under the same lock.
- Compatibility: the cmdline-template hardening rejects any already-published third-party runtime manifest whose template falls outside the closed placeholder and fixed-token sets, at fetch time with a clear `VmSetupError`. Every in-repo template passes.
- Goldens re-seeded at the tip: the attest-agent test edit moves the measured binary (line-table metadata, see the follow-up), and gpuMeasurement also carries the `result_code` anchor and the Turing drop.

### Dependency

aleph-message 1.5.0 (released 2026-09-07 from #162) is pinned in `pyproject.toml`; the two `getattr` fallbacks on `content.gpu` and the skip guard on the GPU launch tests are gone.

### Follow-ups

- Tier 2 hardware run on the SNP host with the RTX PRO 6000 Blackwell Server Edition (checklist in the plan's ledger and the runbook); `accepted_models` is a publish blocker for the GPU runtime.
- Any edit to the attest-agent crate, tests included, moves the measured binary (source line-table metadata), so every such edit needs a golden re-seed; strip that metadata or narrow the crane source.
- Kernel fragment minimisation (hotplug, IOV, IOMMU) once the card is available.
- aleph-rs client verifier: `docs/plans/2026-09-04-nvidia-cc-verifier.md` on `od/nvidia-cc-verifier`; scheduler placement on `properties.tee.nvidia_cc.devices`.


foxpatch-aleph (approved):
This is a careful, fail-closed final installment: the cmdline-template validator now pins every placeholder token to its key and restricts literals to a closed set (I tried several bypass shapes — glued text, doubled braces, multi-placeholder tokens — all rejected), canonical JSON stays byte-stable for pre-existing manifests (gpu is the only Optional field), the agent's GPU gates fire before staging with proper cleanup, the daemon-side sidecar is capped and allowlisted, and the opaque-arm refusal, probe TTL, MMIO clamp, and collector cap are all correct and well tested. Docs match the code everywhere I checked, including the runbook's failure signatures and the boot-smoke expectations against init-gpu.sh's actual output. Two non-blocking notes below.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 2087): Note that this on-demand probe runs under the world write lock: build_written_config (the only production caller of snp_config_slice) is invoked from inside the `let mut world = state.world.blocking_write()` block in create_vm_inner, whereas the sweep path (refresh_cc_modes_with) deliberately probes lock-free on the blocking task. A single config-register read is normally instant, but a read against a wedged card has no timeout and would stall every world operation (Health, GetHostInfo, other creates) — the same failure mode COLLECT_TIMEOUT guards against in the attest agent. Probably fine for a one-shot cold-cache probe; just confirming this trade-off is intentional.

**`src/aleph/vm/agent/vprogram_launch.py`** (line 320): Small asymmetry: findall() collects every match, and _check_cmdline_tokens validates tokens independently (no uniqueness), so a template legally carrying `swiotlb=262144` twice produces a sidecar of `swiotlb=262144 swiotlb=262144` that the daemon's allowlist then rejects at create time — the manifest validator accepts an input the daemon refuses. Fail-closed either way, so not a bug, but rejecting duplicate fixed tokens in the validator (or deduping here) would make the manifest the single source of truth. Relatedly, the `\d{1,9}` in CMDLINE_EXTRA_TOKEN is dead generality today since the validator pins the only legal value to exactly `swiotlb=262144`; harmless, and it does decouple this regex from a future value bump.

**`src/aleph/vm/vprogram/manifest.py`** (line 106): For awareness: this hardening is a breaking schema change for any already-published third-party runtime manifest whose template doesn't fit the new closed sets — such manifests will now fail RuntimeManifest validation at fetch time (VmSetupError) rather than launching. All in-repo templates (V1, EXEC, GPU, LUKS) pass, and the tests cover each rejection shape, so this looks intended; just flagging the compatibility impact.

foxpatch-aleph (approved):
Consistent, well-tested part E of the NVIDIA CC stack: the manifest `gpu` block, the agent family/count resolution, and the shared tee-block builder all fail closed where they must, and the cmdline-template validator properly closes the placeholder-smuggling channel. Probe caching with a TTL, the on-demand cold-cache probe, MMIO clamping and the collector cap are correct and test-covered. Minor nits only: an exactly-at-cap collector output is treated as capped, a repeated swiotlb token is manifest-valid but rejected by the daemon's single-token allowlist, and one compound filter line could be split for readability.

**`rust/crates/aleph-attest-agent/src/gpu.rs`** (line 115): `capped = buf.len() >= MAX_COLLECTOR_OUTPUT` also rejects an output exactly at the cap. Since `take(N)` stops at exactly N bytes, an at-cap read is truncated anyway, so failing is correct — but worth a one-line comment on the `>=` choice so a future reader doesn't "fix" it to `>`.

**`src/aleph/vm/vprogram/manifest.py`** (line 44): `CMDLINE_FIXED_TOKENS_V1` allows `swiotlb=262144` to appear twice (each token individually passes), but the daemon's cmdline_extra allowlist accepts only one `swiotlb=<digits>` token, so such a template fails closed at create. Not reachable from real manifests today; consider a note or a per-token uniqueness check here.

**`src/aleph/vm/agent/capacity.py`** (line 339): The candidate filter line is long; splitting the `models` clause into a separate condition would make the three gating conditions (cc_mode, arch, models narrowing) easier to scan.

foxpatch-aleph (approved):
Well-executed final part of the NVIDIA CC stack. The fail-closed gates are sound: the daemon refuses GPUs on the opaque-cmdline SNP arm and probes cold-cache cards only after proving inventory membership; the probe TTL is keyed on the attached set so anonymous /about callers cannot drive register reads; the MMIO window clamps saturating; the collector cap bounds output without truncating into parse; the cmdline template validator pins every placeholder token to its key, closing a kernel-parameter smuggling channel that affected the existing exec and luks templates; and the agent's family/count resolution honours holds and consumes them only once the whole requirement is met. Docs accurately describe the daemon settings and the trust model, and tests cover every new gate. Only minor non-blocking observations.

**`src/aleph/vm/agent/resources.py`** (line 224): `await _network_gpu_models()` inside `_tee_properties` calls `update_aggregate_settings()` per call, so a `/about/usage/system` poll on a SNP-capable host now refreshes the settings aggregate twice — once here, once via `_gpus_from_host_info` for the usage's gpu block. Consider building the map once per poll and passing it through.

foxpatch-aleph (approved):
Solid, well-guarded part E: the manifest gpu block is a strict closed schema, agent resolution is atomic and lock-protected, the launch path fails closed before any staging I/O, and the cmdline-template hardening (closed fixed-token sets, pinned placeholder->key pairs, duplicate rejection) is verified sound against smuggle/glue/spoof vectors. The GPU gates (count>1, missing gpu block, vendor/arch mismatch, 2048 MiB floor) close the surface between message and runtime, and the daemon-side guarantees (world write lock, live CC-mode probe at create gate) cover the agent's post-resolution race window the same way the instance path already accepts. Two non-blocking nits below.

**`src/aleph/vm/agent/vprogram_launch.py`** (line 193): Non-blocking: the GPU gates validate count, manifest gpu-block presence, vendor, and arch, but never cross-check that a manifest with a gpu block actually carries swiotlb=262144 in its boot.cmdline_template (the GPU_V1 fixed token). A mispackaged GPU runtime whose template omits the token would launch without the IOMMU bounce buffer and degrade into a failed GPU init in-guest rather than a clean agent-side VmSetupError. Since the manifest is already fetched here, a one-line check like "if manifest.gpu is not None and 'swiotlb=262144' not in manifest.boot.cmdline_template: raise VmSetupError" (and possibly the reverse, refusing swiotlb in a gpu-less runtime) would catch mispackaging at launch time. Today it only degrades, never mis-boots, so this is a hardening suggestion, not a bug.

**`docs/operators/nvidia-cc.md`** (line 60): Non-blocking docs nit: the 'Confirm the advertisement' section would benefit from mentioning that the cc_mode annotation is cached with a ~60 second sweep TTL (CC_PROBE_TTL, a fresh sweep only runs when the attached set changes). An operator who just flipped CC mode may otherwise wonder why /about/usage/system still reports cc_mode "off" for up to a minute, and poking the endpoint alone won't force a refresh. Also worth a sentence that a create is never served from this cache: the create gate re-reads the register live, so a stale advertisement cannot cause a mis-launch.

foxpatch-aleph (approved):
This part of the NVIDIA CC stack is correct, fail-closed at every gate, and well-tested. The manifest gpu block is strictly validated (closed Literals, safe path pattern, exclude_none keeping non-GPU manifest hashes stable); the cmdline template hardening closes the glued-token smuggling hole while both manifest formats share one validator so they cannot drift. The agent launch path performs all GPU gates before any staging I/O, cleans up stale sidecars, and converts every parse/integrity failure into VmSetupError. resolve_confidential_gpus consumes hold-ledger entries only on success, leaving the ledger untouched on shortage, and the InsufficientResourcesError payload is scheduler-friendly. The Rust daemon re-probes CC mode under the world write lock at create time (never trusting the sweep cache), bounds public-poller register reads with a 60s TTL keyed on the attached set, and saturates MMIO64 window math safely before the 4 TiB clamp. Advertisement is unified behind a single _tee_properties builder, in-guest verification greps the top-level result_code and poweroffs on any failure, and test coverage exercises every new behavior including ledger invariants and ETXTBSY retry. No correctness or security defects found; only non-blocking operational notes below.

**`src/aleph/vm/agent/resources.py`** (line 353): Non-blocking version-skew note: nvidia_cc_properties now requires gpu.get("arch"), so an agent from this branch pointed at an older daemon (proto field 6 unset) will silently withhold the whole nvidia_cc block. That is the right fail-closed default and it is documented in the docstring, but it is worth remembering when rolling out: agent and supervisor-daemon must be upgraded together, otherwise a CC-capable node will simply stop advertising its GPUs rather than advertise incomplete ones. Consider a one-line note in the rollout/runbook docs stating this co-deployment requirement.

**`src/aleph/vm/agent/vprogram_launch.py`** (line 241): Non-blocking design note: manifest.gpu.accepted_models is not cross-checked on the host against the selected card's model string — enforcement happens in the guest via the device certificate during nvattest attest (fail-closed poweroff). That is the architecturally correct place (the host cannot be trusted to enforce it anyway), but it means a user who narrows the message models to a card the runtime does not accept will pay a full boot before the guest refuses. If this becomes an operational annoyance, the agent could pre-filter candidates against the network-derived model map the way it already narrows by device_id; purely an efficiency nicety, not a correctness gap.

**`src/aleph/vm/agent/capacity.py`** (line 919): Non-blocking: the InsufficientResourcesError available payload lists every matching candidate including cards currently held by other users, without marking which. The same data is already published by /about/usage/system, so this leaks nothing new, but annotating held candidates (e.g. {"device_id": ..., "arch": ..., "held": true}) would make the error message more actionable for debugging scheduler placement failures. Optional polish.

PR #1201, opened 2026-09-07
fix(daemon): reboot recreates the SNP per-tap DHCP server with an empty lease db

odesenfans:
## Problem

Field report from mainnet (SNP instance `cb302e2c…` on computevm1): `aleph instance reboot` leaves the guest looping on DHCPv6 `received DHCP NAK` for as long as the reporter watched (~9 min, 178 retries); stop + start recovers. The guest never reaches the attestation agent, so unlock is dead for the duration.

## Root cause

Two behaviors stack:

1. **Host:** `reboot_vm` (persistent branch) only restarts the controller unit. The per-tap dnsmasq survives with its lease database intact. `StopVm`, by contrast, tears the server down and removes the lease file, and `StartVm` recreates it empty — which is why stop+start works.
2. **Guest:** busybox `udhcpc6` builds a stable DUID-LL from the MAC but picks its IA_NA **IAID as `rand()` per run** (`d6_dhcpc.c`). dnsmasq keys DHCPv6 leases on DUID+IAID, so the rebooted guest is a *new binding*; the deliberately single-address `--dhcp-range` is still leased to the previous boot's IAID for the remaining lease hour, dnsmasq has nothing to allocate, and replies with an error status. On that reply busybox resets its state machine and packet counter, so the `-n -t 5` bound in `init-common.sh` never trips and init hangs before starting the attestation agent. IPv4 is unaffected (dnsmasq keys v4 leases on the stable MAC).

In-guest reboots cannot bypass this fix: SNP QEMU keeps `--no-reboot`, so every reboot funnels through the RebootVm RPC.

## Fix

In the persistent `reboot_vm` branch, for SNP VMs with networking, recreate the per-tap DHCP server **before** the unit restart, mirroring the stop/start pair: best-effort `dhcp.stop` (removes the lease file), fatal `dhcp.start` (a reboot without a leasable address is a brick). No renew can race the gap — the guest's DHCP clients exit after configuring the interface.

Fixes already-published runtimes: no initrd change, no re-measurement, no new `snp-1.0` publish needed.

## Tests

- `rebooting_an_snp_vm_recreates_the_dhcp_server_with_a_fresh_lease_db`: reboot wipes the lease DB via the stop path, stands up a fresh server with the same allocated IP, and orders dhcp stop → dhcp start → unit restart.
- `rebooting_a_plain_vm_touches_no_dhcp_server`: the SNP gate holds on the reboot path.

`cargo test -p supervisor-daemon --lib`: 349 passed. fmt + clippy clean.

Follow-up (separate, next measured runtime bump): bound `udhcpc6` with `timeout` in the guest init so a NAK storm degrades to IPv4-only instead of hanging init.

### DIFF
--- a/rust/crates/supervisor-daemon/src/lifecycle.rs
+++ b/rust/crates/supervisor-daemon/src/lifecycle.rs
@@ -1165,6 +1165,31 @@ pub fn reboot_vm(state: &DaemonState, vm_id: &str) -> Result<(VmEntry, bool), Rp
     }
     let old_status = status_snapshot(state, &entry);
     let unit = entry.unit_name();
+    // Recreate the per-tap DHCP server with an EMPTY lease database, exactly
+    // like the stop/start pair does. The rebooted measured guest re-solicits
+    // with a fresh random IAID (busybox udhcpc6), so a DHCPv6 lease surviving
+    // from the previous boot pins the single-address range to a binding the
+    // new boot can never match: dnsmasq answers NoAddrsAvail until the lease
+    // expires and the guest loops in init before the attestation agent, with
+    // unlock unreachable for up to the whole lease time. The stop (which
+    // removes the lease file) is best-effort like the StopVm one; the start
+    // is fatal like the StartVm one, since a reboot without a leasable
+    // address is a brick. No renew can race the gap: the guest's DHCP
+    // clients exit after configuring the interface.
+    if entry.config.snp().is_some() && networking_enabled(state, &entry) {
+        let lease_file = dhcp::lease_file_path(&dhcp_lease_dir(state), vm_id);
+        if let Err(dhcp_error) = state.dhcp.stop(vm_id, &lease_file) {
+            tracing::warn!(vm_id, %dhcp_error, "cannot stop the DHCP server, continuing");
+        }
+        let tap = tap_assignment(state, vm_id)?;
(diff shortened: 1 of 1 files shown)


foxpatch-aleph (approved):
The fix correctly mirrors the stop/start pair on the persistent reboot path: a best-effort dhcp.stop (which removes the lease file via the shared lease_file_path source of truth) followed by a fatal dhcp.start, both gated on SNP+networking exactly like the create, restart and teardown paths, and ordered before the unit restart so the rebooting guest never races a stale server. The no-renew-race claim holds because the guest's DHCP clients are bound with -q (exit after lease). Lock usage follows established start_vm practice; the ephemeral branch is deliberately untouched. Two new tests with ordering assertions cover the changed behavior; minor nits only.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 1184): Non-blocking: `tap_assignment(state, vm_id)?` and `for_snp` run after the best-effort stop, so if one fails the reboot error leaves the DHCP server down (lease file removed) while the VM keeps running. This is recoverable — `DhcpBackend::start` replaces a leftover unit and `start_vm` recreates on the next start — and `for_snp` here is essentially unreachable (the same tap data already succeeded at create), but a one-line note that a config failure after the stop is recoverable via retry could save a future reader the analysis.

PR #1202, opened 2026-09-09
feat(agent): the v2 allocation endpoints, answering now and converging later

odesenfans:
Task 6 of the async allocations plan, and the first PR of the stack with a production caller: until now nothing reached `agent/allocation/`.

## The endpoints

**`POST /v2/control/allocations`** builds the plan from the push, answers for each VM at once (202 with `accepted` / `pending` / `unchanged` / `removing` / `rejected` / `retained` and the `plan_id`), and hands the reconciler the plan less what the answer refused. The handler reads the supervisor's list and the GPU inventory first, both async; from there to `submit()` nothing yields, so the answer returned is the one the plan was recorded against.

**`POST /v2/control/capacity/check`** judges the same entries the way a create would, with no side effects: no plan recorded, nothing held, nothing running read as dropped. An entry without its message answers `message_required` rather than being fetched. The headroom rides along, computed from the same figures admission uses.

Both share `build_plan` as their one validation boundary (a body that is not a plan is 400 on either, never read as the empty plan), and both take a plan-sized body. The legacy route is untouched.

## The three things carried here from the #1172 and #1173 reviews

**GPU inventory.** `compute_verdict` takes `available_gpus` and passes it to `simulate`; the handler reads it through a now-public `CapacityManager.available_gpus()`. A GPU VM can be placed. Along the way, a real wrinkle: `simulate` treated every live hold as taken, but the reserve endpoint exists precisely so a user can hold a card *before* asking the scheduler, so the verdict would have refused the user's own VM. `simulate` now lets a candidate use a hold its own user placed, mirroring `resolve_gpus(consume_own_hold=True)`; `ResourceRequirements` carries the owner for it.

**Capacity gating.** `narrow_plan` drops the refused entries before `submit()`. Left in, they'd be retried forever at backoff rate and started the moment room appeared, on a node the scheduler was told had refused them. Pending entries stay (nothing judged them), and so does the plan id (it names the push, not the host's room at the time).

**Provisional commitments.** This turned out to be half done already. The instance and V-PROGRAM paths in `create_vm_execution` record *before* `check_message(exclude_vm_hash=...)`, so a concurrent create's admission already sees the record: the early record is the provisional commitment. Only the program path recorded after `create_vm` returned, which is exactly the window the reconciler's concurrent starts open. Programs now take the same shape (record, admit, build, create, wait, one failure path that retires), and a parametrized test pins all three paths to record-before-admit. No new ledger.

## Two findings the plan did not have

- **The app's body ceiling.** `web.Application()` was built with aiohttp's default 1 MiB `client_max_size`, so a 2 MiB plan body would be cut off inside `request.read()`, inside the verifier's broad `except`, and reported as a 401. The route-level 8 MiB cap alone fixes nothing. The app ceiling rises to the plan cap; every other signed route keeps 1 MiB through the verifier's own cap, which is now a per-route parameter that answers 413 (only once the signer is checked, so an anonymous client still learns nothing).
- **The plan's handler code was stale in three places** (`compute_verdict` takes `node_hash`; `simulate` takes 2-tuples; `capacity_check`'s `body.get("vms", [])` was the fail-dangerous shape refused in #1172). All adapted.

## Residuals, stated rather than hidden

- One `await` (the `had_volumes` snapshot) sits between record and admission, so two peers on a tight host can refuse *each other* for a pass. Conservative, self-heals through backoff, and cannot normally arise since gating only submits entries `simulate` admitted together.
- Disk is still judged live at admission, as before: two concurrent creates can both pass the disk check before either writes. The second fails at the write, loudly, not silently.
- The capacity check judges capacity only; node pinning is the plan route's business.

## Verification

Every new test was mutation-checked against the specific defect it exists to catch (23 mutations, each failing exactly its own test). Suite: 45 failed / 1107 passed against a freshly measured `dev-2.1` baseline of 45 / 1077, same three unrelated files (root-owned `/var/cache/aleph`, `/var/lib/aleph/vm`). mypy at the 47-error baseline, ruff and isort clean.

Task 7 (`allocation` + `state` in `/v2/about/executions/list`) follows.

### DIFF
--- a/src/aleph/vm/agent/allocation/plan.py
+++ b/src/aleph/vm/agent/allocation/plan.py
@@ -46,6 +46,11 @@ class AllocationState(str, Enum):
     Deliberately NOT a mirror of VmStatus: once create_vm returns, VmStatus is
     the answer and this state stops existing. Two disjoint fields cannot drift,
     whereas a merged enum would need maintaining in lockstep forever.
+
+    The reconciler sets DOWNLOADING and FAILED, the only two it can observe.
+    The rest are the executions list's to report from the plan: PLANNED for
+    an entry the loop has not reached, RESOLVING for one still waiting on its
+    message, SUBMITTING for one handed to the supervisor.
     """
 
     PLANNED = "planned"
--- a/src/aleph/vm/agent/allocation/reconciler.py
+++ b/src/aleph/vm/agent/allocation/reconciler.py
@@ -230,6 +230,12 @@ class AllocationReconciler:
         self._states.pop(vm_hash, None)
 
     def _record_failure(self, vm_hash: ItemHash, error: Exception) -> None:
+        # There is no terminal failure, on purpose. The plan is the authority
+        # on what should run here, so a VM it still lists is still owed an
+        # attempt, at the capped interval; giving up would leave a listed VM
+        # not running with nothing outside this node able to tell. The
+        # scheduler dropping the hash is what ends the retries, and submit()
+        # clears the record then.
         now = self._now()
         previous = self._failures.get(vm_hash)
         attempts = (previous.attempts if previous else 0) + 1
--- a/src/aleph/vm/agent/allocation/verdict.py
+++ b/src/aleph/vm/agent/allocation/verdict.py
@@ -27,6 +27,7 @@ from aleph.vm.agent.capacity import (
     requirements_from_message,
 )
 from aleph.vm.agent.vm_registry import AgentVmRecord
+from aleph.vm.resources import GpuDevice
 from aleph.vm.supervisor_interface.types import ConfidentialMode, VmInfo, VmStatus
 
 logger = logging.getLogger(__name__)
@@ -46,6 +47,7 @@ class _Capacity(Protocol):
         candidates: list[tuple[ItemHash, ResourceRequirements]],
         *,
         releasing: frozenset[ItemHash] = ...,
+        available_gpus: list[GpuDevice] | None = ...,
     ) -> list[AdmissionVerdict]: ...
 
 
@@ -149,15 +151,15 @@ def compute_verdict(
     registry: _Registry,
     capacity: _Capacity,
     node_hash: str | None = None,
+    available_gpus: list[GpuDevice] | None = None,
 ) -> PlanVerdict:
     """The immediate answer: what we take, what we drop, what we refuse.
 
-    No GPU inventory reaches simulate from here, so a candidate asking for a
-    card is refused rather than admitted on memory alone. Reading the host's
-    cards is an async call on the supervisor and this stays await-free, for
-    the reason at the top of the module, so that read belongs to the handler
-    that will make this answer authoritative and has yet to be written. Until
-    it is, no plan can place a GPU VM.
+    ``available_gpus`` is the host's unattached cards. Reading them is an
+    async call on the supervisor and this stays await-free, for the reason at
+    the top of the module, so the handler reads them first and hands them in.
+    Without them simulate refuses every candidate that asks for a card, which
+    is the right answer to give when the cards were not looked at.
     """
     verdict = PlanVerdict()
     known = by_hash(infos)
@@ -216,10 +218,26 @@ def compute_verdict(
             continue
         candidates.append((vm_hash, requirements_from_message(content)))
 
-    for admission in capacity.simulate(candidates, releasing=frozenset(verdict.removing)):
+    admissions = capacity.simulate(candidates, releasing=frozenset(verdict.removing), available_gpus=available_gpus)
+    for admission in admissions:
         if admission.accepted:
             verdict.accepted.append(admission.vm_hash)
         else:
             verdict.rejected[admission.vm_hash] = {"code": admission.code, "message": admission.detail}
 
     return verdict
+
+
+def narrow_plan(plan: AllocationPlan, verdict: PlanVerdict) -> AllocationPlan:
+    """The plan the reconciler is handed: the push, less what the answer refused.
+
+    A refused entry must never reach the loop. Left in, it would be retried
+    forever at backoff rate, and started the moment room appeared, on a node
+    the scheduler was told had refused it and has since placed it elsewhere
+    from. Pending entries stay: nothing judged them, and the fetch that will
+    is the loop's own. The identity stays too, since it names the push, and a
+    re-push of the same set is the same plan whatever the host had room for
+    the first time.
+    """
+    entries = {vm_hash: planned for vm_hash, planned in plan.entries.items() if vm_hash not in verdict.rejected}
+    return AllocationPlan(plan_id=plan.plan_id, received_at=plan.received_at, entries=entries)
--- a/src/aleph/vm/agent/capacity.py
+++ b/src/aleph/vm/agent/capacity.py
@@ -539,7 +539,11 @@ class CapacityManager:
 
         ``gpus`` is None rather than empty when no inventory was given:
         unknown is not zero, and the caller that has the inventory is the one
-        that read it from the supervisor.
+        that read it from the supervisor. It lists the cards free of any live
+        hold, the node-wide view: there is no owner to ask for here, so a card
+        one user holds is absent even though simulate would let that user's
+        own candidate take it. The per-candidate answer is the one to trust
+        for a given VM; this figure is what anyone else could count on.
         """
         caps = self._caps()
         committed_instance, committed_program, committed_vcpus = self._committed_resources(())
--- a/src/aleph/vm/agent/capacity.py
+++ b/src/aleph/vm/agent/capacity.py
@@ -58,6 +58,9 @@ class ResourceRequirements:
     # though it is not an InstanceContent.
     is_instance: bool = False
     gpu_device_ids: list[str] = field(default_factory=list)
+    # Whose VM this is, for the GPU ledger: a hold this address took is
+    # available to it, the way resolve_gpus consumes an owner's own hold.
+    owner: str | None = None
 
 
 # The filename suffixes a volume's file carries in ``{pool}/{vm_hash}/``. A
@@ -198,6 +201,7 @@ def requirements_from_message(
         max_volume_mib=max(volume_sizes_mib, default=0),
         is_instance=is_instance_bucket(content),
         gpu_device_ids=requested_gpu_ids(content),
+        owner=str(address) if (address := getattr(content, "address", None)) else None,
     )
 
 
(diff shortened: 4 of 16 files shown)


foxpatch-aleph (commented):
A high-quality PR: the v2 plan/capacity-check endpoints are correctly wired, the no-yield window between the supervisor reads and submit() holds, narrow_plan's exclusion of refused entries is right, and the record-before-admit reordering closes the concurrent-program-create hole conservatively (it can only under-admit). Tests are thorough and the pure-logic suite passes locally (221 tests across the verdict/capacity/auth/create-path files; the aiohttp-client tests could not run in this sandbox due to a missing pytest-aiohttp plugin, not a code issue). One real finding: RequestTooLarge is raised by the verifier but only caught by the decorator, so the three handlers in views/__init__.py that call authenticate_api_request directly (update_allocations, recreate_network, regenerate_proxy) now answer 500 for an oversized body where they previously answered 401 and where the PR's stated design is 413 — contradicting the PR description's claim about "every other signed route". That, plus a small robustness nit in _read_plan (JSONDecodeError-only catch turns a non-UTF-8 body into a 500 rather than a 400), is worth a small follow-up; neither blocks the core functionality.

**`src/aleph/vm/agent/views/__init__.py`** (line 546): This handler (and its siblings at lines 732 and 765) call `authenticate_api_request` directly, but `RequestTooLarge` is only caught by the `requires_allocation_auth` decorator. With the app ceiling raised to 8 MiB, an oversized body on these three routes now propagates out of the handler and error_middleware converts it into a 500 — previously these routes answered 401, and the PR description states oversized bodies "answer 413". Consider catching `RequestTooLarge` here (or mapping it to `web.HTTPRequestEntityTooLarge` in `error_middleware`) so the direct-call sites get the same 413 semantics as the decorated routes.

**`src/aleph/vm/agent/views/allocations_v2.py`** (line 41): Catching only `JSONDecodeError` misses `UnicodeDecodeError` (a `ValueError` but not a `JSONDecodeError`), which `request.json()` raises for a body that isn't valid UTF-8 — that escapes to a 500 via error_middleware instead of the intended 400. Catching `ValueError` covers both cases.

**`src/aleph/vm/agent/supervisor.py`** (line 281): Note that raising the app-wide ceiling to 8 MiB also widens the buffer limit for routes with no cap of their own (the `operate_*` machine endpoints), not just the signed routes the comment discusses. Acceptable for the control-plane API, but the comment's framing ("each signed route enforces its own cap below it") only accounts for the signed surface.

odesenfans:
Both findings are real, reproduced before fixing: a 2 MiB signed body to `/control/allocations` answered `500 {"error_type": "RequestTooLarge"}`, and `{"vms": "\xff\xfe"}` on the v2 route answered 500 with the decoder's message.

- `6ca0d79c` makes `RequestTooLarge` a subclass of `web.HTTPRequestEntityTooLarge`. It is then an `HTTPException`, which `error_middleware` already renders as a JSON 413, so every handler calling the verifier directly gets the right answer with no per-site catch and nothing new in the middleware; the decorator returns the same object. The end-to-end test on the legacy route is the mirror of the v2 one.
- `1cad85c7` catches `ValueError` in `_read_plan`, covering the `UnicodeDecodeError`, with a test sending bytes that are not text. The plan refusal now answers with a fixed text instead of `str(error)`, which is what the CodeQL alert keyed on.
- `9219a678` rewrites the ceiling comment to say it covers the JWK-authenticated operator routes too.

The description's "every other signed route keeps its 1 MiB" was true; what was wrong was the implicit claim that over it meant 413 everywhere. It does now.

Suite: 45 failed / 1109 passed against the 45 / 1077 baseline, same three unrelated files. Both new tests mutation-checked (plain `Exception` base and `JSONDecodeError`-only catch each fail exactly their own test). mypy at 47, ruff and isort clean.

foxpatch-aleph (approved):
Solid, carefully-reasoned PR. The v2 endpoints are wired correctly: narrow_plan gates both verdict-time and build-time rejections before submit (the rejected-dict merge happens first), the GPU own-hold fix mirrors resolve_gpus without corrupting the shared inventory list (pool is copied), record-before-admit is now consistent across all three create paths with the right retire-on-failure semantics, and the 413 plumbing works for both decorated routes and legacy routes that call the verifier directly (verified error_middleware catches HTTPException). The no-yield window between the supervisor reads and submit() holds since everything downstream is synchronous. Non-blocking: the app-wide client_max_size bump to 8 MiB also covers the unauthenticated /vm/ routes (build_asgi_scope buffers the full body), an 8x per-request buffering increase on the public interface that the code comment doesn't mention; and the v2 route's over-the-plan-cap 413 path is untested.

**`src/aleph/vm/agent/supervisor.py`** (line 283): Raising the app-wide client_max_size to 8 MiB also raises the per-request body ceiling for the unauthenticated, public-facing routes: build_asgi_scope (run.py:84) does `await request.read()` on /vm/{ref} and hostname-routed code-execution POSTs, so every one of those requests can now buffer 8 MiB where the previous default capped them at 1 MiB. The comment above this line documents the signed and operator routes but not this class of callers. Deliberate and bounded, but worth one explicit sentence here (or a follow-up to bound the ASGI routes separately) so the next reader knows the tradeoff was considered.

**`tests/supervisor/test_allocations_v2.py`** (line 201): The suite pins the v2 route accepting a body over the legacy cap (this test) and the legacy route answering 413, but never the mirror case: a body over MAX_SIGNED_PLAN_BODY_BYTES on /v2/control/allocations answering 413 rather than 202. The code path is the same RequestTooLarge raise the legacy test exercises, so this is only a small gap — but since this route is the one the ceiling exists for, one request asserting the 413 would pin the contract end to end.

**`src/aleph/vm/agent/views/allocations_v2.py`** (line 110): Minor note on API shape: keys merged from build_plan's rejected map are unvalidated raw strings (e.g. "None" for an entry with no item_hash), so the response's rejected map can carry junk keys that never were hashes. This matches what the plan route answers and is harmless, just be aware consumers cannot assume every key is a hash.

odesenfans:
`26f1ad8b` drops the app-wide `client_max_size` change. The cap is now the route's alone: `requires_allocation_auth(max_body_bytes=...)` re-bounds the request with `request.clone(client_max_size=cap)` before anything reads it (the clone shares the payload and applies its own limit on read, keeps `app` and `match_info`), and hands the clone to the verifier and the handler. The app default stays at 1 MiB, so the operator routes are back where they were and a plan-sized body gets through the two v2 routes and nowhere else.

Verified by execution that a clone under a 1 MiB app default reads a 2 MiB body, and by mutation that removing the clone fails exactly the over-the-legacy-cap test (the aiohttp 413 is then swallowed inside the verifier as a 401 again). Suite unchanged at 45 / 1109, mypy 47, ruff and isort clean.

odesenfans:
Thanks. The app-wide ceiling concern is the stronger version of the one raised earlier, and it is gone as of `26f1ad8b`: the cap is re-bound per route through `request.clone(client_max_size=...)`, the app default stays at 1 MiB, and the `/vm/` routes buffer exactly what they did before.

`3d7c27d7` adds the mirror test you asked for: a body over `MAX_SIGNED_PLAN_BODY_BYTES` on `/v2/control/allocations` answers 413 and nothing reaches the reconciler. Mutation-checked: dropping the cap check fails it together with the legacy one.

On the `rejected` keys: agreed, and by design. `build_plan` keeps the raw key precisely because it is whatever junk the push carried in place of a hash, so the scheduler can correlate the refusal with what it sent; a consumer reading `rejected` should treat keys as opaque. That matches what the plan route answers.

foxpatch-aleph (approved):
Solid, carefully reasoned PR. The two new endpoints share build_plan as a single validation boundary, the per-route body-cap mechanism (clone with a raised client_max_size rather than raising the app-wide ceiling) is correct against aiohttp 3.10's clone/read semantics, the 413 answer is correctly ordered after the signer check so it leaks nothing to anonymous clients, and the record-before-admit move for the program path is the right fix for the reconciler's concurrent starts. The no-yield window between the two async reads and submit() holds as documented, narrow_plan cannot collide rejected entries with plan entries, and the own-hold rule in _take_gpus mirrors _match_requests/consume_own_hold exactly. All new and changed tests pass locally (16 + 116 + 105 + 15 across the touched files). Remaining observations are minor and non-blocking.

**`src/aleph/vm/agent/capacity.py`** (line 551): Minor inconsistency in the same response: _take_gpus now lets a candidate use its own user's live hold, but headroom still lists cards via _unheld_gpus, which drops every live hold regardless of owner. A capacity-check answer can therefore admit a candidate for a card while capacity["gpus"] omits it. Conservative and advisory-only, so not blocking, but a scheduler correlating the two halves of the response may find it surprising. Consider documenting (or mirroring) the own-hold exception here.

**`src/aleph/vm/agent/views/allocation_auth.py`** (line 305): Edge case at the exact boundary: aiohttp's Request.read() raises HTTPRequestEntityTooLarge when body_size >= client_max_size, while the verifier allows content_length <= max_body_bytes. A body of exactly MAX_SIGNED_PLAN_BODY_BYTES bytes therefore passes the verifier's check and is then rejected by the clone's read — the effective cap is one byte below the constant. Harmless (still a 413 on an authenticated route), and pre-existing for the legacy cap, but worth a line in the RequestTooLarge docstring if anyone ever sizes a plan body to the byte.

**`tests/supervisor/test_allocations_v2.py`** (line 85): The 2s client timeout and the 1s wait_for on started are tight for a loaded CI runner: the POST returns 202 immediately, but the reconciler has to be scheduled and reach _start_one within 1s. It passes comfortably here, but consider slightly larger margins or awaiting started before the POST completes to keep this from flaking.

**`src/aleph/vm/agent/capacity.py`** (line 462): Nit: _check_against re-reads HOST_MEMORY_RESERVED_MIB / PROGRAM_MEMORY_RESERVED_MIB from settings for its error message after HostCaps.read() already folded the same two into instance_memory_mib / program_memory_mib. Both read the same settings so they cannot disagree today, but HostCaps now owns the caps — carrying these two fields there (or logging from caps) would keep one source of truth.

odesenfans:
All four taken, and the boundary one turned out worse than described: a body of exactly the cap answered **401**, not 413. aiohttp refuses at its limit (`>=`) inside `request.read()`, and that is aiohttp's own `HTTPRequestEntityTooLarge`, not our subclass, so the verifier's broad `except` swallowed it as a bad signature. Probed all three sizes before fixing: cap-1 → 202, cap+0 → 401, cap+1 → 413.

- `c76fb385` bounds the request one over the cap, so the verifier's `>` on Content-Length stays the authority and an exact-cap body passes both; aiohttp's 413 is re-raised alongside ours in case it ever fires. Test at exactly the cap, mutation-checked (cloning at the cap fails it and only it).
- `e4a6a2bf` documents what `headroom()["gpus"]` is a view of: the node-wide figure, free of any live hold, with no owner to ask for; the per-candidate answer is the one to trust for a given VM.
- `87f2d7a6` widens the test bounds to 10s. They bound failure, not waiting.
- `8839dc73` moves the two reservations into `HostCaps`; `_check_against` reads them from there.

Suite 45 / 1111 against the 45 / 1077 baseline, mypy 47, ruff and isort clean.

foxpatch-aleph (approved):
Well-designed and thoroughly executed. The two new endpoints share build_plan as a single validation boundary, the auth changes correctly thread a per-route body cap through aiohttp's request lifecycle (verified against the pinned aiohttp 3.10.11 internals: clone shares the payload, client_max_size is enforced at read time, and the +1 keeps a body of exactly the cap passing), and narrow_plan's dict-key membership is sound since ItemHash is a str subclass. The record-before-admit change for the program path reuses the existing retire path so a refused create leaves no phantom record, and the parametrized test pins all three paths. All new and affected tests pass locally (89 + 165); ruff format is clean with the pinned 0.4.6 and mypy reports nothing new on the changed files. The two comments below are observations about pre-existing parity and a residual window, neither of which blocks.

**`src/aleph/vm/agent/allocation/verdict.py`** (line 198): A message-less entry is answered 'pending' and never sees the node-pinning check below (it requires the message), and neither the reconciler's fetch path nor start_persistent_vm re-checks pinning — so a hash-only entry pinned to another node will be started here anyway. This matches the legacy route's hash-only gap (parity, not a regression), but the narrow_plan docstring's "the fetch that will [judge them] is the loop's own" overstates what the loop judges: it re-verifies the message and runs admission, but node pinning is only enforced for entries that arrived with their message embedded.

**`src/aleph/vm/agent/run.py`** (line 839): The on-demand program path (_ensure_program_vm) still records after create_vm returns, so the provisional-commitment window this PR closes in create_vm_execution remains here: a reconciler-started create's admission won't count an in-flight on-demand program create. The per-VM creation_lock serializes only on-demand creates for the same hash, so the two paths can still race each other's sums. Not a regression and the supervisor's physical-memory backstop exists, but it may be worth a follow-up (or a comment) since this PR's own argument applies to it.

PR #1203, opened 2026-09-09
feat(agent): report allocation state in the v2 executions list

odesenfans:
Task 7 of the async allocations plan, the last one: `/v2/about/executions/list` reports what the agent is doing about a VM before, and beside, what the supervisor says.

## Two disjoint fields per entry

- **`state`**: the supervisor's `VmStatus`, verbatim. `running` stays, now an alias of it for existing consumers.
- **`allocation`**: the agent's phases the supervisor cannot see, or `null` when it has nothing to add. The loop's own `downloading` and `failed` (with the failure's `code`/`message`, `attempts` and `next_retry_at`), or `planned` / `resolving` derived from the plan for an entry the loop has not reached, the difference being whether the create will have to fetch the message first.

A VM the plan lists that the supervisor does not yet know is listed with `state: null` and the same keys as a live entry (`networking: {}`, all-null `status` times, `vm_type` from the registry record if a recreate left one), so the scheduler can tell "working on it" from "never heard of it", which it used to have to guess, and a reader iterating the list does not special-case the ones not up yet. A VM the supervisor knows can still carry an allocation block: the failure that keeps a dead one dead, or a recreate in flight.

## Two small things on the reconciler

- `pending_hashes()`: the planned entries the push carried no message for. It is the one fact the list needs from the plan that `planned_hashes()` did not give.
- `AllocationState.SUBMITTING` is removed. Nothing could set it without a hook in the create path, and an enum member nothing sets is a state a reader will wait for in vain. `PLANNED` and `RESOLVING`, flagged as unused on #1173 and #1202, now have their consumer.

## Verification

8 endpoint tests through the real app over a fake supervisor, plus one unit test for `pending_hashes`. Six mutations (derived states swapped, derivation overriding the loop's own state, dedupe removed, allocation dropped for known VMs, `state` nulled, `pending_hashes` emptied), each caught by its intended test. Suite 45 failed / 1126 passed against a freshly measured `dev-2.1` baseline of 45 / 1117, same three unrelated files. mypy at the 47-error baseline, ruff and isort clean.

That closes the plan. Left from its "after" list: the design doc's sections 6 and 11 are stale (they predate `check_message` and the merged prerequisites) and live on the unmerged `od/async-allocations-design` branch.

### DIFF
--- a/src/aleph/vm/agent/allocation/plan.py
+++ b/src/aleph/vm/agent/allocation/plan.py
@@ -48,15 +48,14 @@ class AllocationState(str, Enum):
     whereas a merged enum would need maintaining in lockstep forever.
 
     The reconciler sets DOWNLOADING and FAILED, the only two it can observe.
-    The rest are the executions list's to report from the plan: PLANNED for
-    an entry the loop has not reached, RESOLVING for one still waiting on its
-    message, SUBMITTING for one handed to the supervisor.
+    The executions list derives the other two from the plan, for an entry the
+    loop has not reached: PLANNED when the push carried its message, RESOLVING
+    when the create will have to fetch it first.
     """
 
     PLANNED = "planned"
     RESOLVING = "resolving"
     DOWNLOADING = "downloading"
-    SUBMITTING = "submitting"
     FAILED = "failed"
 
 
--- a/src/aleph/vm/agent/allocation/reconciler.py
+++ b/src/aleph/vm/agent/allocation/reconciler.py
@@ -112,6 +112,17 @@ class AllocationReconciler:
     def planned_hashes(self) -> set[ItemHash]:
         return set(self._desired.entries) if self._desired else set()
 
+    def pending_hashes(self) -> set[ItemHash]:
+        """The planned VMs whose message the push did not carry.
+
+        They are started like any other; the fetch happens inside the create.
+        Exposed so the executions list can say "waiting on its message" for
+        one the loop has not reached yet, where a verified entry is "planned".
+        """
+        if self._desired is None:
+            return set()
+        return {vm_hash for vm_hash, planned in self._desired.entries.items() if planned.verified is None}
+
     def state_for(self, vm_hash: ItemHash) -> tuple[AllocationState | None, FailureRecord | None]:
         """What the agent is doing about this VM, for the executions list."""
         return self._states.get(vm_hash), self._failures.get(vm_hash)
--- a/src/aleph/vm/agent/views/__init__.py
+++ b/src/aleph/vm/agent/views/__init__.py
@@ -290,7 +290,6 @@ async def list_executions(request: web.Request) -> web.Response:
     )
 
 
-@cors_allow_all
 def _allocation_block(state: AllocationState | None, failure: FailureRecord | None) -> dict | None:
     """What the agent is doing about a VM, for the executions list; None once
     the agent has nothing to add to the supervisor's word."""
@@ -304,6 +303,7 @@ def _allocation_block(state: AllocationState | None, failure: FailureRecord | No
     }
 
 
+@cors_allow_all
 async def list_executions_v2(request: web.Request) -> web.Response:
     """List all executions. Returning their status and ip.
 
--- a/src/aleph/vm/agent/views/__init__.py
+++ b/src/aleph/vm/agent/views/__init__.py
@@ -22,6 +22,8 @@ from pydantic import ValidationError
 from aleph.vm import haproxy
 from aleph.vm.agent import payment, status
 from aleph.vm.agent.aggregate import update_aggregate_settings
+from aleph.vm.agent.allocation.plan import AllocationState, FailureRecord
+from aleph.vm.agent.allocation.reconciler import AllocationReconciler
 from aleph.vm.agent.allocation.teardown import is_removable_by_allocation, teardown_vm
 from aleph.vm.agent.capacity import CapacityManager, requested_gpu_ids
 from aleph.vm.agent.custom_logs import set_vm_for_logging
@@ -206,6 +208,9 @@ def _datetime_from_ns(ns: int) -> datetime | None:
     )
 
 
+_TIMES_KEYS = ("defined_at", "preparing_at", "prepared_at", "starting_at", "started_at", "stopping_at", "stopped_at")
+
+
 def _times_dict(info: VmInfo) -> dict[str, datetime | None]:
     """The VmExecutionTimes-shaped dict the v2 endpoint has always served."""
     return {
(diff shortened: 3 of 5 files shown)


foxpatch-aleph (commented):
Well-executed change: clean separation of the supervisor's `state` from the agent's `allocation`, a sensible derivation of PLANNED/RESOLVING for entries the loop hasn't reached, and thorough tests through the real app (all 35 pass; mypy clean on the touched files). The only defect found is a misplaced `@cors_allow_all` decorator, which is cosmetic here since CORS is actually applied via `cors_routes` + `cors.add(route)` in `setup_webapp`, plus one non-blocking security consideration about surfacing raw exception text on the public endpoint.

**`src/aleph/vm/agent/views/__init__.py`** (line 293): The `@cors_allow_all` decorator ended up on `_allocation_block` when the helper was inserted above the view: `list_executions_v2` (line 307) is now the only handler in `cors_routes` without it. It has no runtime effect either way — `custom_cors` only sets a `{name}_cors_config` attribute consumed by class-based `CorsViewMixin` views, which this app never uses; real CORS comes from `cors_routes` + `cors.add(route)` in `setup_webapp` — but the placement breaks the convention and will confuse readers into thinking the helper is a view. Please move the decorator back onto `list_executions_v2`.

**`src/aleph/vm/agent/views/__init__.py`** (line 302): Non-blocking consideration: `allocation.error.message` surfaces `str(error)[:200]` from arbitrary exceptions during `start_persistent_vm` on an unauthenticated, CORS-open endpoint, so internal details (filesystem paths, fetch URLs) could leak in a failure string. That's arguably consistent with what this endpoint already exposes (VM IPs, host IP, mapped ports), so fine to ship — just flagging to confirm it was considered.

odesenfans:
Both points reviewed against the code.

**`@cors_allow_all` on `_allocation_block`**: right, the decorator stayed on the old line when the helper was inserted above the view. Moved back in 8790a979. Confirmed the "no runtime effect" reading: `custom_cors` only sets a `<name>_cors_config` attribute nothing in `src/` reads, and both list routes answered identical `Access-Control-*` headers on GET and preflight before the fix. The commit adds a test pinning those headers on `/v2/about/executions/list`, which guards the route's place in `cors_routes` (a mutation moving it to `other_routes` fails the test).

**`allocation.error.message` on a public endpoint**: considered, keeping it. The reconciler records the create path's raw exception, so the text can be a typed `SupervisorError`, a `ClientResponseError` (fetch URL) or an `OSError` (node path). The precedent is broader than what this endpoint already lists: `error_middleware` returns `str(exc)` plus the exception type, uncapped, for any unhandled exception on every route, and the legacy allocation endpoint returns `repr(error)` per VM. This field is the tightest of the set (200 chars, paired with the `code` that is the actionable part, same text as the reconciler's warning log). Contents are public item hashes and API URLs, and standard `/var/lib/aleph/vm` paths; no credential-bearing path into it. If that precedent should be tightened, `error_middleware` is the place, as its own change.

foxpatch-aleph (approved):
This PR is correct and well-tested. The two-disjoint-fields design (supervisor's VmStatus verbatim in 'state', agent-only phases in 'allocation') is clean and avoids the enum-drift problem the docstrings describe; deriving PLANNED/RESOLVING in the view from the plan for unreached entries is consistent with pending_hashes() and with the verdict's own 'pending' terminology. I verified there is no interleaving race (all reconciler reads are synchronous after the supervisor awaits), that the app key dependency always exists, that the dedupe check is type-safe, and that VmType.from_message_content cannot raise for a registry record. All new tests plus the reconciler suite pass, and mypy is clean on the changed files; the only failures I could reproduce are environmental (permission on /var/cache/aleph in test_operator_retire.py), matching the author's stated baseline. Three non-blocking notes below, chiefly a pre-existing state-bookkeeping wrinkle in _start_one that this PR makes observable: when a plan drops a VM mid-create and that create then fails, the early return leaves a stale DOWNLOADING in _states, which the new list will publish on a supervisor-known VM until the next push prunes it.

**`src/aleph/vm/agent/allocation/reconciler.py`** (line 234): Non-blocking, but this early return leaves the stale DOWNLOADING that _start_one set at its top: submit() pruned _states when the plan dropped the VM, but the assignment at line 215 happened before the await, so _states still holds DOWNLOADING for a VM no plan lists. Before this PR nothing read _states, so it was harmless; now the executions list will report allocation.state == 'downloading' on a supervisor-known VM (e.g. one held FAILED) until the next submit() prunes it. A self._forget(vm_hash) before this return would keep the published state truthful. Edge case (plan push landing exactly during an in-flight create that then fails) and self-healing on the next push, hence non-blocking.

**`src/aleph/vm/agent/views/__init__.py`** (line 301): Worth a conscious sign-off: allocation.error.message publishes str(error)[:200] — raw exception text that can include internal URLs and filesystem paths — on an unauthenticated endpoint. This is consistent with what the surface already exposes (VmInfo.status_message, /about/executions/records), so I don't consider it a defect, but if the list is meant for scheduler consumption only, a code-only error or a sanitized message would be tighter.

**`tests/supervisor/test_executions_list_allocation.py`** (line 150): The view docstring promises 'a create in flight for one it holds dead' is reported, but only the FAILED-on-known-VM combo is covered here; a DOWNLOADING-on-known-VM case (the recreate-in-flight half of the claim) would pin that too. Very minor — the view code handles both identically via state_for.

odesenfans:
Thanks for the approval. On the three notes:

**Stale DOWNLOADING after a mid-create plan drop (reconciler.py:234)**: does not happen. `submit()` prunes `_states` as well as `_failures` for every hash the new plan lacks (lines 92-97), so the DOWNLOADING set at the top of `_start_one` is gone by the time the failed create reaches the early return, and nothing re-adds it. That exact interleaving is what `test_a_start_that_fails_after_the_plan_drops_it_is_not_remembered` pins: the fake create submits a plan without the VM, then raises, and `state_for` is asserted `(None, None)`. Checked by mutation both ways: removing the `_states` prune from `submit()` fails that test, and adding the suggested `self._forget` before the return changes nothing (27/27 still pass). Left as is.

**`allocation.error.message`**: signed off, see the earlier comment for the reasoning. On "if the list is meant for scheduler consumption only": it is not, the console reads it from the browser, which is why the route sits in `cors_routes`; a human-facing reader is the one that benefits from the text.

**DOWNLOADING beside a supervisor-known VM**: fair, only the failure half of the docstring's claim was pinned. Added in 22602d3d, with a mutation (attach the block to known VMs only when FAILED) that fails exactly that test.

foxpatch-aleph (approved):
Solid, well-documented change. The two-disjoint-fields design (supervisor state verbatim + agent allocation block) avoids enum drift, the dedupe between supervisor entries and planned hashes is correct, the removed SUBMITTING enum member has no lingering references, and submit()'s pruning prevents stale allocation blocks. All 10 new endpoint tests and 27 reconciler tests pass locally. Remaining observations are minor: a retry in flight keeps a stale next_retry_at in the past alongside state "downloading", the view's three reconciler reads are mutually consistent only incidentally (nothing awaits between them), and the unauthenticated endpoint now exposes raw failure text — consistent with what it already exposes, but worth a conscious accept.

**`src/aleph/vm/agent/views/__init__.py`** (line 357): The pending loop reads the reconciler three times (pending_hashes(), planned_hashes(), state_for() per hash) and their mutual consistency rests on nothing awaiting in between — true today, but a torn read would silently appear if someone later added an await inside this loop. The handler in allocations_v2.py documents this discipline explicitly ("nothing may yield"); a one-line comment here, or a single snapshot-style accessor on the reconciler, would make the invariant explicit rather than incidental.

**`src/aleph/vm/agent/allocation/reconciler.py`** (line 215): _start_one sets DOWNLOADING without clearing a previous FailureRecord, so during a retry the executions list reports state "downloading" together with the previous attempt's error and a next_retry_at already in the past. Arguably informative, but a consumer keying off next_retry_at may find it ambiguous; worth either a comment saying the failure is the previous attempt's, or clearing it when a retry starts.

**`src/aleph/vm/agent/views/__init__.py`** (line 301): Worth a conscious accept: this endpoint is unauthenticated and CORS-open, and failure.message is str(error)[:200] from the create path, which can embed internal paths or URLs. It is consistent with what this endpoint already exposes (host IPs, networking), and the shape is deliberately scheduler-facing, so not a blocker — just noting the exposure is now richer than it was.

**`src/aleph/vm/agent/views/__init__.py`** (line 373): Minor duplication: this re-derives VmType.from_message_content(record.message).name, which is the record branch of _vm_type_name — the fallback there can't be reused since there is no VmInfo for a VM the supervisor doesn't know. Fine as is; flagging only in case the derivation grows.

PR #1204, opened 2026-09-09
fix(examples): keep the confidential image's encrypted root readable by the OVMF's GRUB

odesenfans:
`examples/example_confidential_image/build_debian_image.sh` formats the LUKS root with a bare `mkfs.ext4`, so the filesystem features come from the build host's `mke2fs.conf`. On Ubuntu 26.04 (e2fsprogs 1.47.2) that now includes `orphan_file` and `metadata_csum_seed`, and the GRUB embedded in the confidential OVMF cannot read the result. The guest unlocks the volume and stops right after:

```
Entering grub config
Attempting to decrypt master key...
Slot 0 opened
Failed to find any grub configuration on the encrypted volume
```

`/boot/grub/grub.cfg` is present on the volume; GRUB just cannot read the filesystem. Ubuntu 24.04 (e2fsprogs 1.47.0) did not enable either feature, which is why images built there boot.

Pin the feature set with `mkfs.ext4 -O ^orphan_file,^metadata_csum_seed`. Found on the new aleph-testnets TEE server (aleph-testnets #39/#40 carry the same change for the vendored copy). Users building encrypted images on a recent Ubuntu hit the same thing, so this is user-facing, not just CI.

https://claude.ai/code/session_01RnureenxugkU48aFn5HL6h

### DIFF
--- a/examples/example_confidential_image/build_debian_image.sh
+++ b/examples/example_confidential_image/build_debian_image.sh
@@ -140,7 +140,14 @@ echo -n "${DISK_PASSWORD}" >"${KEY_FILE}"
 
 sudo cryptsetup --batch-mode --type luks1 --key-file "${KEY_FILE}" luksFormat "${OS_PARTITION_DEVICE_ID}"
 sudo cryptsetup open --key-file "${KEY_FILE}" "${OS_PARTITION_DEVICE_ID}" "${MAPPER_NAME}"
-sudo mkfs.ext4 "${MAPPED_DEVICE_ID}"
+# The GRUB embedded in the confidential OVMF is what opens this volume and
+# loads its grub.cfg, and its ext2 driver cannot read a filesystem created
+# with the orphan_file and metadata_csum_seed features: the guest unlocks
+# the volume, prints "Failed to find any grub configuration on the encrypted
+# volume" and halts. e2fsprogs 1.47.2 (Ubuntu 26.04) enables both by
+# default, so turn them off explicitly instead of depending on the build
+# host's mke2fs.conf.
+sudo mkfs.ext4 -O ^orphan_file,^metadata_csum_seed "${MAPPED_DEVICE_ID}"
 
 echo "Copying root file system to the new OS partition..."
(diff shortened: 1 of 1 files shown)


foxpatch-aleph (approved):
The fix is correct and well-validated: both orphan_file (incompat 0x10000) and metadata_csum_seed (incompat 0x0400) set incompatible-feature bits that the pre-2.12 GRUB embedded in the confidential OVMF (aleph/efi-secrets fork, edk2-stable202205) rejects, and the -O ^feature,^feature syntax is valid mke2fs usage that clears only those two features while leaving the rest of the host defaults (including metadata_csum, which GRUB reads fine) intact. I verified the syntax works on e2fsprogs 1.47.0 and that the failure mode described matches how GRUB handles unknown incompat bits. The comment accurately captures the why, and no security or injection concerns exist since the feature list is static and device variables stay quoted. One non-blocking caveat: mke2fs hard-fails on unrecognized feature names, and orphan_file was only introduced in e2fsprogs 1.47.0, so build hosts with e2fsprogs 1.46.x (Ubuntu 22.04, Debian 11, RHEL 9) now abort with "Invalid filesystem option set: ^orphan_file" where the old code worked. Consider documenting the e2fsprogs >= 1.47.0 requirement in the example README.

**`examples/example_confidential_image/build_debian_image.sh`** (line 150): Verified this is correct — the syntax works on e2fsprogs 1.47.0 and both cleared features match the incompat bits the OVMF's pre-2.12 GRUB rejects. One caveat worth noting: mke2fs hard-errors on unknown feature names, and `orphan_file` only exists as a recognized name since e2fsprogs 1.47.0 (Feb 2023). Build hosts with e2fsprogs 1.46.x (Ubuntu 22.04 LTS, Debian 11, RHEL 9) — where the old bare `mkfs.ext4` worked, since those versions never enable these features — will now abort with `Invalid filesystem option set: ^orphan_file`. Since the README already lists build requirements, consider adding e2fsprogs >= 1.47.0 there (or probing with `mkfs.ext4 -n` before committing to the flag). Non-blocking.

PR #1205, opened 2026-09-09
docs: drop AI plan documents from the tree

odesenfans:
## Summary
docs/superpowers/ (nvidia-cc plan and spec) and docs/plans/2026-08-19-compose-runtime-port-design.md were tracked in dev-2.1, but repository policy is that design and plan documents produced while developing a feature do not ship. This removes them and inlines what the surviving code and docs actually relied on from the compose design doc, so nothing is lost by removing it.

## Changes
- `docs/superpowers/plans/2026-09-04-nvidia-cc-aleph-vm.md`, `docs/superpowers/specs/2026-09-04-nvidia-cc-design.md`: removed, nothing referenced them.
- `docs/plans/2026-08-19-compose-runtime-port-design.md`: removed after inlining its contract at its referrers.
- `docs/architecture/divergences.md` entry 64: replaced the "see the design doc for the full contract" pointer with the actual contract in prose (workload volume layout, the closed compose subset, the single attested port, fail-closed semantics, and why the CRN needs no code changes for this flavor).
- `.gitignore`: added `/docs/superpowers/` and `/docs/plans/`, both used locally during development, so they cannot be committed again by accident.
- `nix/compose-rootfs.nix` and `nix/flake.nix` still name the removed `docs/plans/2026-08-19-compose-runtime-port-design.md` path in comments. This is deliberate, not a miss: CI on this branch showed that editing those two comments to drop the reference moved `composeMeasurement` in the golden-measurements job, while the other three measurements stayed put, so the compose rootfs derivation evidently hashes the nix source files themselves, comments included. Rewriting those comments is a measurement-affecting change, not a docs no-op, so this PR leaves them exactly as they were on `dev-2.1` rather than re-seeding the golden file for a comment edit. Fixing the underlying reproducibility wart (why comment text feeds the measured boot chain at all) is a follow-up.

## Test plan
Docs-only change, no tests. `git grep -n 'docs/plans\|docs/superpowers' -- docs/ nix/ src/ rust/` now returns only pre-existing code-comment references (in `src/` and `rust/`) plus the two deliberately-unchanged `nix/` comments explained above; none point at the removed `docs/superpowers/` path. The pre-existing `src/`/`rust/` references are separate PR Q4's scope.


foxpatch-aleph (commented):
Docs-only PR that correctly removes the three plan/design documents and inlines the compose contract where its three referrers needed it. I verified the inlined claims against the code: the workload-volume layout matches nix/compose-rootfs.nix (docker-compose.yml at root, images/*.tar, at-least-one-required), the single-reachable-port claim matches the nftables policy drop with only tcp dport 8443 accepted in init-compose.sh, fail-closed semantics match the fatal() helper and the fatal-on-compose-exit line, and workload.contract is genuinely unread by the launch path. No dangling references to the removed docs remain, and the .gitignore entries are correctly anchored. Two minor nits introduced by the edit: a now-dangling "(see below)" pointer in compose-rootfs.nix (the subset description it pointed at lived in the deleted doc; it now lives in divergences.md entry 64, which the comment should cite), and an awkwardly-closed eight-line parenthetical in flake.nix left by splicing into an existing sentence. Also note the removal deletes the only in-repo rationale for the insecureAcceptAnything podman policy and the RAM-sizing guidance, which is fine per the stated policy but should be a conscious call.

**`nix/compose-rootfs.nix`** (line 150): "(see below)" is now a dangling pointer: in the deleted design doc it referred to the Compose subset / Out-of-scope sections, but nothing below in this file describes the subset. The description now lives in docs/architecture/divergences.md entry 64 (inlined by this PR) — cite that, or drop the parenthetical. As written, this is exactly the class of stale doc reference the PR set out to remove.

**`nix/flake.nix`** (line 20): The parenthetical opened at line 10 ("(compose-rootfs, initially excluded too, ...") now closes here, seven lines later, so the entire inserted sentence about the exec/compose topology inversion sits inside one giant aside, and the leftover "). The" is a seam from the original sentence ("...as /init; see <doc>). The encrypted-rootfs ..."). Cleaner: close the paren after "as /init." and start the inversion explanation as a fresh sentence.

**`nix/compose-rootfs.nix`** (line 188): Non-blocking scope note: deleting the design doc also deletes the only in-repo rationale for the `{"default": [{"type": "insecureAcceptAnything"}]}` policy (the no-pull-path + verity-volume argument) and the guidance that resources.memory must exceed the summed image archive sizes. Nothing surviving referenced those sections so no pointer breaks, but a future reviewer of this policy.json now has no in-tree explanation — confirm that information loss is acceptable under the no-plan-docs policy.

foxpatch-aleph (approved):
A clean, well-scoped docs-only removal: the dangling references check out (nothing points at the deleted superpowers docs, no tracked file is hidden by the new .gitignore rules), and the divergences.md entry 64 inlining faithfully preserves the compose contract's five core points rather than just deleting the design doc. Two things deserve follow-up but none block: (1) the PR's diagnosis of the measurement coupling is likely half right — the compose-rootfs.nix comment genuinely feeds the measured boot chain because it sits inside the baked /sbin/init heredoc, but the plain top-of-file comment in nix/flake.nix is very likely NOT a measurement input, and the "reproducibility wart" framing may send the follow-up chasing a nonexistent bug; (2) the insecureAcceptAnything image-trust rationale and the restart-accepted-but-ignored detail are now undocumented in the repo. The conservative choice not to re-seed golden measurements for a comment edit is the right call regardless.

**`nix/flake.nix`** (line 13): The measurement-input claim is probably over-broad here. Unlike the compose-rootfs.nix:151 comment (which sits inside the `cat > rootfs/sbin/init <<'INIT'` heredoc and is genuinely baked into the measured rootfs), this is a plain top-of-file nix comment that no derivation reads — `grep -rn 'flake.nix' nix/*.nix nix/*.sh` shows nothing copies flake.nix content into a build. The CI experiment that moved composeMeasurement edited both files in one commit, so the flake.nix edit was never isolated and was most likely a red herring. Worth verifying in CI (edit only this line, confirm composeMeasurement stays put) before tracking a "why comment text feeds the measured boot chain" wart that, for flake.nix, probably doesn't exist. If it does stay put, this comment can be reworded to drop the stale doc reference after all.

**`nix/compose-rootfs.nix`** (line 186): The deleted design doc's `insecureAcceptAnything` soundness rationale (no pull path exists in the compose init; every archive rides the dm-verity workload volume, so integrity is covered by the launch measurement rather than image signing) now exists nowhere in the repo — the rootfs ships an accept-anything policy in a confidential-computing image with no in-repo explanation. A nix comment placed *outside* the INIT/POLICY heredocs would document this without moving any measurement (heredoc content is what's baked; surrounding nix comments are not). Same applies to the accepted-but-ignored `restart` key, see next comment.

**`docs/architecture/divergences.md`** (line 81): Fidelity nit in the inlined contract: the deleted doc said `restart` is *accepted but ignored* (with a warning), but the inlined accepted list (`image`, `command`, `entrypoint`, `environment`, `depends_on`, `tmpfs`) omits it, and `restart` does not appear in the rejected list either — so a reader of the new "contract of record" would conclude `restart` is rejected by `deny_unknown_fields` when it is actually accepted. One clause fixes it.

**`.gitignore`** (line 41): Fine as-is, but note the interaction with the ~15 pre-existing `docs/plans/...` references across proto/, scripts/, src/, rust/ and tests/ (e.g. proto/README.md:8 points at docs/plans/2026-05-28-aleph-vm-architecture-backport-design.md as a mirror doc). Those files don't exist in the tree today, and this rule cements that they never will — consistent with the stated policy, but worth handling in the follow-up cleanup (PR Q4) rather than leaving long-term dead pointers in shipped docs like proto/README.md.

**`nix/compose-rootfs.nix`** (line 151): For the record, this is the comment that genuinely is a measurement input: it sits inside the embedded init heredoc, so it is baked into the guest rootfs verbatim, and any edit here changes the roothash and composeMeasurement. Keeping it verbatim (dead doc reference and all) rather than re-seeding the golden file for a comment edit is the right trade-off. The PR description's generalization — "the compose rootfs derivation evidently hashes the nix source files themselves" — is the wrong mechanism though: only this heredoc-internal text is hashed, not the nix sources generally, which is why the flake.nix comment is a different case (see the flake.nix comment).

foxpatch-aleph (commented):
The mechanical content of the PR is correct and verified: the three documents are removed, no surviving doc links to them, the .gitignore entries are properly root-anchored, and the inlined compose contract in divergences.md entry 64 accurately matches the code (volume layout, fail-closed init, single 8443 firewall rule, insecureAcceptAnything policy, contract-string neutrality of the launch path). However, the PR's stated reason for leaving nix/flake.nix:13 unchanged is incorrect: a plain nix comment outside any string cannot affect any derivation, and the flake's only self-reference (sourceRev) feeds only the non-measured source-rev files — the composeMeasurement move CI observed was caused entirely by the compose-rootfs.nix:151 edit, which genuinely is measurement-affecting because it sits inside the measured init heredoc. flake.nix:13 could therefore be cleaned up in this PR at zero golden-measurement cost, and the follow-up the PR proposes is mischaracterized: the wart is not that comment text feeds the measured chain, but that this one reference lives inside a string baked into the image. Second, the in-guest comment at compose-rootfs.nix:148-151 still points readers to the removed doc's Out-of-scope section for why the compose init deliberately shadows /tmp/secrets, but that rationale was not inlined anywhere, so it is now lost.

**`nix/flake.nix`** (line 13): This comment is NOT measurement-affecting and could be cleaned up in this PR at zero cost. It is a plain nix comment outside any string, so it cannot affect evaluation; the flake's only self-reference is sourceRev (line 158), which feeds only the non-measured source-rev files. The composeMeasurement move you observed in CI was caused entirely by the compose-rootfs.nix:151 edit (which sits inside the `cat > rootfs/sbin/init <<'INIT'` heredoc and is baked into the measured rootfs), not by this one. Editing both comments in one push confounded the observation. As a side effect, the follow-up is mischaracterized: the wart is not that comment text feeds the measured boot chain, but that this one reference happens to live inside a string literal that is part of the image.

**`nix/compose-rootfs.nix`** (line 151): Correctly left as-is (editing it would move composeMeasurement), but note this in-image comment points to the removed doc's Out-of-scope section for why the compose init deliberately shadows the /tmp/secrets bind mount, and that rationale was not inlined into divergences.md entry 64 or anywhere else in the surviving tree. One sentence in entry 64 (compose v1 rejects volumes: outright, so injected-secret delivery to a container has no expression in the aleph.compose/1 schema; the tmpfs shadowing is the fail-safe) would close the gap so nothing is lost by the doc's removal.

PR #1206, opened 2026-09-09
fix(agent): validate a namespace with ItemHash's rules before any delete

odesenfans:
## Summary
The guard on every storage delete path (`purge._checked_namespace`, plus three copies of the same regex) accepted any 16 to 128 alphanumeric string as a VM namespace, while an item hash is only 64 lowercase hex characters or an IPFS CID. A directory an operator dropped on a volume pool and called `backupsfromjanuary` therefore looked like a VM: under `VOLUME_RETENTION=reap` the reconciler's namespace pass called `purge_vm_storage` on it, which unlinked every file and removed the directory, and only then did `purge_vm_staging` build an `ItemHash` from the same name and raise into the pass's swallowed `except`. The operator's data was gone first and the error was logged after.

There is now a single validator, `storage.vm_namespace`, which returns an `ItemHash` only when the name parses as one. The purge guard refuses on it before any filesystem access, and the passes that walk the pools ask the same question, so they skip what the guard would refuse instead of tripping over it mid-pass.

## Changes
- `src/aleph/vm/storage.py`: added `vm_namespace(name) -> ItemHash | None` and `is_vm_namespace(name) -> bool`; removed `NAMESPACE_PATTERN` and pointed its three users (`device_name_for`, `remove_devmapper`, `remove_base_device`) at the new check.
- `src/aleph/vm/agent/vm/purge.py`: `_checked_namespace` now validates with `vm_namespace` and returns the parsed `ItemHash`, so `purge_vm_staging` no longer re-parses the name after the volumes are already deleted. `_ITEM_HASH_PATTERN` is gone.
- `src/aleph/vm/agent/vm/reconciler.py`: `_plausible` delegates to `is_vm_namespace` (the name is kept, it is imported by `agent/storage_cli.py`, which a concurrent PR chain is editing).
- `src/aleph/vm/agent/vm/cache.py`: dropped its own `_plausible` copy in favour of `is_vm_namespace`.
- `docs/architecture/storage.md`: the reconcile section described the namespace check only as "validated by `purge._checked_namespace`"; it now states the rule.

## Test plan
- `tests/supervisor/test_vm_purge.py`: the implausible-namespace parametrisation gains `backupsfromjanuary`, `deadbeefdeadbeef` and an uppercase 64-char string; a new test pins that `purge_vm_storage` on such a directory raises with its contents still on disk; another pins that a 64-hex hash and a `Qm...` CID both still purge.
- `tests/supervisor/test_reconciler.py`: a directory an operator named is neither reaped (its file survives, `purged_orphans` empty) nor marked reclaimable under `keep`; a hex hash and a CIDv0 still reconcile as orphans.
- All five new tests fail on the base commit (the reap case fails on the deleted file, which is the data loss) and pass after the fix.
- Checks: `pytest tests/supervisor` (1186 passed, 1 skipped, 1 failed: `test_log.py::test_make_logs_queue`, identical on the base commit), `ruff format --diff`, `isort --check-only --profile black`, `mypy src/aleph/vm/ tests/`, all clean.

### DIFF
--- a/docs/architecture/storage.md
+++ b/docs/architecture/storage.md
@@ -385,7 +385,12 @@ an mtime younger than `VOLUME_CREATE_GUARD`. Both matter, since a create
 that outlives the guard is only protected by the first, and a create that
 somehow escaped the context manager is still protected by the second.
 Everything a pass touches is under a directory the agent created and is
-keyed by a hash validated by `purge._checked_namespace`.
+keyed by a hash that `storage.vm_namespace` accepted: the name has to parse
+as an `ItemHash` (64 lowercase hex characters, or an IPFS CID), so a
+directory an operator dropped on a volume pool is never a VM. The delete
+paths go through `purge._checked_namespace`, which refuses on the same rule
+before any filesystem access; the passes that walk the pools ask
+`is_vm_namespace` first and skip what the guard would refuse.
 
 The startup pass is stricter than the rest: it refuses to purge anything
 (it runs dry and logs why) when the supervisor did not answer `list_vms`,
--- a/src/aleph/vm/agent/vm/cache.py
+++ b/src/aleph/vm/agent/vm/cache.py
@@ -42,7 +42,7 @@ from collections.abc import Callable, Collection, Iterable
 from dataclasses import dataclass
 from pathlib import Path
 
-from aleph.vm.agent.vm.purge import _ITEM_HASH_PATTERN, purge_vm_storage
+from aleph.vm.agent.vm.purge import purge_vm_storage
 from aleph.vm.agent.vm.reclaimable import (
     MANIFEST_REF_KINDS,
     ReclaimableMarker,
@@ -57,6 +57,7 @@ from aleph.vm.resources import InsufficientResourcesError
 from aleph.vm.storage import (
     DEVICE_MAPPER_DIRECTORY,
     DEVICE_NAME_MAX_BYTES,
+    is_vm_namespace,
     reserve_download,
     reserved_downloads,
 )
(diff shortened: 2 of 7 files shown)


foxpatch-aleph (commented):

foxpatch-aleph (approved):
Genuine and serious bug, cleanly fixed: the old 16-128 alphanumeric guard let an operator-named directory on a volume pool pass for a VM, and purge_vm_storage deleted its contents before the staging step's ItemHash() raised. Delegating to ItemHash's own validation is the right call - I verified in the installed aleph_message that from_hash accepts only Qm.../bafy... CIDs and 64-char lowercase hex, raises only UnknownHashError (a ValueError), so the new vm_namespace's except clause is airtight and every existing except ValueError caller remains correct. The guard now runs before any filesystem access and returns the parsed ItemHash, eliminating the re-parse-after-delete window, and the reconciler/cache passes now skip what the guard would refuse instead of raising mid-pass. All in-repo users of the removed patterns are migrated, the create path (which builds device names by f-string) is unaffected, tests pin the data-loss scenario for both retention modes plus the still-purges cases, and the docs are updated accurately. The purge/reconciler/cache/retire/devmapper suites all pass here; the full-suite failures in this sandbox are environment artifacts (no write access to /var/cache/aleph, older aleph_message missing ConfidentialGpuRequirement), not caused by the diff. Two tiny nits below.

**`tests/supervisor/test_vm_purge.py`** (line 174): Nit: the "every shape" parametrizations (here and in test_reconciler.py's test_every_shape_of_item_hash_still_reconciles) cover a 64-hex storage hash and a Qm... CIDv0, but not the bafy... CIDv1 branch that ItemHash/ItemType.from_hash also accepts (59-char base32, checked in aleph_message's from_hash). One more parametrize entry would pin all three accepted shapes on both the purge and reconcile paths.

**`src/aleph/vm/agent/vm/reconciler.py`** (line 156): Nit: now that _plausible is a pass-through to is_vm_namespace, its docstring explains what it does but not why the wrapper exists at all - i.e. that agent/storage_cli.py imports it and a concurrent PR chain is editing that file. A one-line note (or having storage_cli import is_vm_namespace directly once that chain lands) would spare future readers the indirection.

PR #1207, opened 2026-09-09
fix(tee): appraise the TD report 1.5 second TCB vector

odesenfans:
## Summary
A TD report 1.5 quote body carries two TDX SVN vectors, `tee_tcb_svn` and `tee_tcb_svn2`: the pair exists because a TD-preserving update swaps the TDX module under a running TD, so one describes the module TCB the TD launched on and the other the one it runs on now. `evaluate_tcb` appraised the first vector only, so a TD whose other vector matches no published TCB level, or names a TDX module major version Intel does not list, or sits below every level of its module identity, was accepted on the strength of the vector that happened to be walked, and a lower level that only the second vector falls into was never seen.

The fix appraises every vector the body carries, in full: the platform level walk and the TDX module identity and SVN ladder alike. Either vector failing fails the appraisal, and the worse of the two statuses is the answer. That is stricter than Intel's own handling, which propagates a failure of the second appraisal but keeps the first vector's status and only flags that a relaunch is advised; there is no advisory outcome here, so a status this policy would refuse for one vector is refused for the TD.

## Changes
- `rust/crates/aleph-tee/src/tdx/tcb.rs`: `appraise_tdx_svn_vector` appraises one SVN vector (level walk plus TDX module block); `appraise_platform_tcb` runs it for `tee_tcb_svn` and, on a 1.5 body, for `tee_tcb_svn2`. Each vector's failures carry a context naming that vector, so an error off a 1.5 body says which of the two tripped. TD report 1.0 bodies keep their single appraisal, so no verdict changes for v4 quotes.
- `rust/crates/aleph-tee/src/tdx/tcb.rs`: `tdx_module_status` takes the report body and the SVN vector instead of the whole quote, because the module major version and module SVN both come out of the vector and each vector names its own module.
- `rust/crates/aleph-tee/src/tdx/tcb.rs`: the two vector results combine with a new symmetric `TcbStatus::worse` (max over the declared severity order) rather than `converge`. `converge` is the asymmetric platform-versus-component rule and would make the outcome depend on which vector was walked first (`ConfigurationNeeded` then `OutOfDate` gives `OutOfDateConfigurationNeeded`, reversed gives `OutOfDate`). `converge` still combines the module and QE statuses into a platform status, which is what it is for.
- `rust/crates/aleph-tee/src/tdx/tcb.rs`: the advisory union, now needed a third time, moves into a small `merge_advisories` helper shared with the module and QE merges.
- `nix/golden-measurements.json`: re-seeded with `nix/check-golden-measurements.sh --update`. The attest agent binary links aleph-tee and is hashed into the measured initrd, so any edit to this crate moves every pinned launch measurement even when no verdict changes. The file now carries five entries: the base rebase picked up `gpuMeasurement` from the confidential-GPU runtime flavor.

### Review follow-ups
- `rust/crates/aleph-tee/src/tdx/tcb.rs`: only the second vector labelled its failures, so an error reading "the platform TCB is below every level in Intel's TCB Info" left an operator unable to tell which vector tripped. The first vector's appraisal now carries the symmetric context naming `tee_tcb_svn`.

## Test plan
- `rust/crates/aleph-tee/src/tdx/tcb.rs` (unit tests, hand-built TCB Info as the surrounding tests do):
  - `report_15_second_svn_vector_is_appraised_too`: a body whose `tee_tcb_svn2` is below every level is rejected (naming `tee_tcb_svn2`) while `tee_tcb_svn` alone satisfies the top level; a second vector that does meet the level still appraises UpToDate.
  - `report_15_second_vector_module_version_must_be_listed`: a second vector naming module version `TDX_02`, absent from `tdxModuleIdentities`, is rejected; the listed version on both vectors appraises normally.
  - `report_15_second_vector_walks_the_module_svn_ladder`: a second vector whose module SVN is below every level of its identity is rejected, and against a ladder that does list it, the module status it lands on is taken up.
  - `report_15_takes_the_worse_of_the_two_svn_vectors`: each vector matches a different level; the worse status wins in either direction and both levels' advisories are reported.
  - `report_15_status_does_not_depend_on_the_vector_order`: the `ConfigurationNeeded` / `OutOfDate` pair that separates the symmetric rule from `converge`, asserted in both orders.
  - `report_10_body_walks_once`: a 1.0 body has no second vector and is unaffected.
  - `converge_worst_wins_and_config_rule`: extended with the `worse` cases.
  - `outdated_platform_is_below_every_level`: asserted on the bare top-level message, which is now the vector context rather than the walk's own bail, so it formats the whole error chain and also pins that the launch vector is named.
  - Each new test was watched fail against the code without its own change (second vector unappraised, module gate on the launch vector only, `converge` in place of `worse`).
- Checks after the rebase onto `dev-2.1` (all exit 0, read bare): `cargo fmt --check`, `cargo clippy --locked --all-targets -- -D warnings`, `cargo test --locked -p aleph-tee` (131 passed, 2 ignored), and fmt plus clippy with `--manifest-path crates/aleph-attest-agent/Cargo.toml`, since that workspace compiles aleph-tee under its own lock. `nix/check-golden-measurements.sh` matches after the re-seed.

## Known gap, not in this PR
`walk_platform_tcb` compares all 16 TDX components, where Intel skips components 0 and 1 when the module major SVN is above 0. Stricter than the reference and pre-existing, so it is left alone here.

An end-to-end `evaluate_tcb` test on a genuine signed v5 quote with a 1.5 body is still missing: no such fixture or matching Intel collateral is available, so the two-vector paths are pinned at the unit level against synthetic TCB Info instead.


foxpatch-aleph (approved):
I've completed my review. Here's my assessment:

## Analysis

**The fix is correct and matches Intel's reference behavior.** `appraise_platform_tcb` (tcb.rs:451) walks the TCB levels for `tee_tcb_svn`, and for a 1.5 body (`body.v15.is_some()`) walks again for `tee_tcb_svn2`, converging statuses and unioning advisories — exactly how dcap-qvl/Intel's verifier treat a TD report 1.5. Fail-closed in both directions: either walk finding no level fails the appraisal with the vector named in the context.

Points I verified by hand:

- **Walk logic**: the second walk reuses `walk_platform_tcb` unchanged, including `canonical_levels` highest-first sorting, so document order can't skew either vector's result.
- **1.0 bodies unaffected**: `v15` is `None` for a v4 quote's 1.0 body, so the second walk is skipped; `report_10_body_walks_once` pins this.
- **Module status**: `tdx_module_status` (tcb.rs:489) still keys off `tee_tcb_svn[0..1]`, which is correct — the module identity (MRSIGNERSEAM, module SVN) describes the module the TD launched on; only the platform TCB evaluation needs both vectors.
- **Tests**: I traced all three new tests against the code — the "worse status wins" test exercises both orderings and asserts the advisory union; the "second vector appraised too" test fails on the old code (returns `UpToDate`) as claimed. I could not execute the suite (no Rust toolchain in this environment), so I'm relying on the reported green checks.
- **Converge asymmetry**: `converge` is order-sensitive (`OutOfDate.converge(ConfigurationNeeded) == OutOfDate` but `ConfigurationNeeded.converge(OutOfDate) == OutOfDateConfigurationNeeded`), so swapping which vector lands in `self` vs. `component` can shift a status between those two. Both are rejected by the default policy, so this has no practical impact — just worth knowing.

Minor observations (non-blocking):

- When the *first* vector matches no level on a 1.5 body, the error doesn't name which vector failed, whereas the second walk does. Symmetric context would change the error surface asserted in `outdated_platform_is_below_every_level`, so the current asymmetry is defensible.
- `merge_advisories` is a clean de-duplication of three identical merge loops; taking `from` by value is fine given all callers hand over ownership.

```json
{
  "verdict": "APPROVE",
  "summary": "Correct, fail-closed fix: appraise_platform_tcb now walks Intel's TCB levels for both the tee_tcb_svn and tee_tcb_svn2 vectors of a TD report 1.5 body, converges the statuses the way the module and QE statuses already converge, unions the advisories, and fails the appraisal when either walk finds no level — matching Intel's own verifier. TD report 1.0 bodies keep the single walk, so v4 quotes are unaffected, and tdx_module_status correctly continues to key off the launch module identity. The three new tests cover the second-vector rejection, the worse-status convergence in both orderings, the advisory union, and the unchanged 1.0 path; I traced them against the code and they exercise the bug as claimed (I could not execute cargo in this environment). The merge_advisories helper is a clean de-duplication and the doc comments are excellent. Two non-blocking notes below."
}
,
  "comments": [
    {
      "path": "rust/crates/aleph-tee/src/tdx/tcb.rs",
      "line": 461,
      "body": "Non-blocking: converge is order-sensitive (OutOfDate.converge(ConfigurationNeeded) is OutOfDate, but the reverse is OutOfDateConfigurationNeeded), so which vector is \"self\" vs \"component\" can shift the reported status between those two. Both are rejected by the default policy so there is no practical impact, but a one-line comment noting the chosen order would help future readers."
    },
    {
      "path": "rust/crates/aleph-tee/src/tdx/tcb.rs",
      "line": 456,
      "body": "Non-blocking observability nit: only the second walk gets an error context naming the vector. When the first vector of a 1.5 body matches no level, the error does not say which vector failed. Symmetric context would change the error surface asserted in outdated_platform_is_below_every_level, so the current asymmetry is understandable — feel free to ignore."
    }
  ]
}
```

**`rust/crates/aleph-tee/src/tdx/tcb.rs`** (line 461): Non-blocking: converge is order-sensitive (OutOfDate.converge(ConfigurationNeeded) is OutOfDate, but the reverse is OutOfDateConfigurationNeeded), so which vector is "self" vs "component" can shift the reported status between those two. Both are rejected by the default policy so there is no practical impact, but a one-line comment noting the chosen order would help future readers.

**`rust/crates/aleph-tee/src/tdx/tcb.rs`** (line 456): Non-blocking observability nit: only the second walk gets an error context naming the vector. When the first vector of a 1.5 body matches no level, the error does not say which vector failed. Symmetric context would change the error surface asserted in outdated_platform_is_below_every_level, so the current asymmetry is understandable — feel free to ignore.

foxpatch-aleph (approved):
Correct, fail-closed fix. A TD report 1.5 body carries two TDX SVN vectors and previously only tee_tcb_svn was appraised; now both get the full treatment (platform level walk plus TDX module identity and SVN ladder), either vector's failure fails the appraisal, and the second vector's errors are clearly labeled. The new symmetric TcbStatus::worse (max over the severity order) is the right combining rule for two peer results of the same appraisal — converge is asymmetric and would make the outcome depend on vector order, which the dedicated order-independence test pins. Under the default policy the max-composition is exactly 'both vectors must individually pass', and the advisory union means denied-advisory policies now see the second vector's advisories too. v4/1.0-body behavior is unchanged. Tests are thorough, each targeting one escape the old code allowed; the synthetic-TCB-Info approach is sound given no v5 collateral fixture exists. Only comment-level nits remain.

**`rust/crates/aleph-tee/src/tdx/tcb.rs`** (line 486): Minor doc nit: "a status this policy would refuse for one vector is refused for the TD" is strictly true only when accepted_statuses is downward-closed in the severity order (the default {UpToDate, SwHardeningNeeded} is). A custom policy accepting OutOfDate but refusing ConfigurationNeeded would pass a mixed (ConfigurationNeeded, OutOfDate) TD as OutOfDate. Same information loss converge already has for platform-vs-component, so acceptable — but the comment could soften the claim to "for the default policy" or "for any refusal-conservative accept set".

**`rust/crates/aleph-tee/src/tdx/tcb.rs`** (line 494): Symmetry nit: the second vector's failures are labeled with the "appraising the TD report 1.5 second TCB vector (tee_tcb_svn2)" context, but the first vector's are unlabeled. Unambiguous with only two vectors, but a symmetric .context("appraising the TD report TCB vector (tee_tcb_svn)") would make diagnosing a launch-vector failure marginally easier for an operator reading the error string.

**`rust/crates/aleph-tee/src/tdx/tcb.rs`** (line 1226): Nice touch that this test asserts the error propagates through the outer context rather than silently converging — that is the behavior Intel's own reference has (failure of the second appraisal propagates), and it distinguishes it from the status-combination path tested separately.

foxpatch-aleph (approved):
This PR closes a real appraisal gap: a TD report 1.5 body's tee_tcb_svn2 vector was previously ignored entirely, so a TD running under a module TCB matching no published level, an unlisted module major version, or a lower module SVN could pass on the strength of the launch vector alone. The fix appraises both vectors in full (platform level walk plus the per-vector TDX module identity and SVN ladder), fails if either fails, and combines the two peer statuses with a new symmetric TcbStatus::worse (max over the severity order) rather than converge — and I hand-verified that converge really is order-dependent for the ConfigurationNeeded/OutOfDate pair, so worse is the correct peer rule. TD report 1.0/v4 bodies keep their single appraisal, verified by test and by the parser (v15 is only set for 648-byte bodies). The change is fail-closed throughout: the second vector's module gate re-checks MRSIGNERSEAM/SEAMATTRIBUTES against its own identity and rejects on mismatch, and the deliberately stricter-than-Intel worse-of-two outcome is well documented in both code and PR. Tests are thorough and their error-string assertions match the actual messages; the golden-measurements re-seed follows the documented procedure. I could not run cargo in this environment (no toolchain), so verification was static. Minor nits: the first vector's failure lacks a symmetric error context naming tee_tcb_svn, and there is no end-to-end evaluate_tcb test with a genuine signed v5/1.5 fixture (appears to be a collateral-data limitation, with unit-level synthetic coverage compensating).

**`rust/crates/aleph-tee/src/tdx/tcb.rs`** (line 494): Nit: the second vector's failure is wrapped in a `.context(...)` naming tee_tcb_svn2, but the first appraisal's failure propagates bare, so on a 1.5 body an error like "the platform TCB is below every level" doesn't say whether tee_tcb_svn or tee_tcb_svn2 tripped. A symmetric context (e.g. `.context("appraising the TD report TCB vector (tee_tcb_svn)")` or similar) on the first call would make production diagnostics unambiguous. Non-blocking.

**`rust/crates/aleph-tee/src/tdx/tcb.rs`** (line 499): Confirmed the `worse` over `converge` choice is correct: `ConfigurationNeeded.converge(OutOfDate)` yields OutOfDateConfigurationNeeded via the `(component, self)` match arm while `OutOfDate.converge(ConfigurationNeeded)` falls through to max and yields OutOfDate, so converge would make a 1.5 body's outcome depend on evaluation order. The symmetric max is the right peer rule, and the strictness delta versus Intel's keep-first-status-plus-advisory handling is intentional and documented. No change needed.

foxpatch-aleph (approved):
The fix is correct and strictly fail-closed: both TDX SVN vectors of a TD report 1.5 are now appraised in full (platform level walk plus module identity and SVN ladder), either vector failing fails the appraisal, and the worse of the two statuses is the answer. Using a symmetric `worse` (max) rather than the asymmetric `converge` across the two vectors is the right call — I verified that `converge` would make the outcome depend on vector order (`ConfigurationNeeded`/`OutOfDate` yields `OutOfDateConfigurationNeeded` one way and `OutOfDate` the other). 1.0 bodies keep a single appraisal so v4 quotes are unaffected apart from an added, operator-friendly error context. Tests are thorough and hand-verification of each against the walk and ladder logic checks out; the golden-measurement re-seed is procedurally required since aleph-tee ships in the measured attest-agent binary. No security or correctness issues found.

**`rust/crates/aleph-tee/src/tdx/tcb.rs`** (line 567): Minor observation (non-blocking): appraising the second vector gates `body.mrsignerseam`/`body.seam_attributes` — a single per-body value — against the identity each vector names. All Intel TDX module identities share one MRSIGNER so this is moot today, and it fails closed if it ever weren't, but a TD-preserving update to an identity with a distinct MRSIGNER would be rejected even though MRSIGNERSEAM reflects the launch module. Worth a sentence in the doc comment if that ever becomes a real scenario.

**`rust/crates/aleph-tee/src/tdx/tcb.rs`** (line 497): The `.context("appraising the TD report launch TCB vector (tee_tcb_svn)")` also applies to 1.0 bodies where there is only one vector; the wording "launch TCB vector" still reads fine there, and the existing `outdated_platform_is_below_every_level` test was correctly updated for the new message.

PR #1208, opened 2026-09-09
chore: prune unused Python deps, fix deb and CI package lists

odesenfans:
## Summary
pyproject.toml still declared jsonschema, py-cpuinfo, pyyaml and schedule as dependencies with zero real imports left anywhere in src/, tests/ or examples/, and the deb control file carried the matching apt packages plus a stray debootstrap dependency that only the runtimes/ build scripts (build-time, not the installed node) need. Control was also missing pciutils: the daemon's lspci probe (rust/crates/supervisor-daemon/src/lspci.rs) aborts startup without it when ALEPH_VM_ENABLE_GPU_SUPPORT is set. CI apt lines and Justfile still installed python3-nftables/python3-netifaces, leftover from before the Python supervisor daemon was replaced by the Rust one (nftables is now driven through the `nft` CLI in rust/crates/supervisor-daemon/src/nft.rs).

Stacked on #1205.

## Changes
- pyproject.toml: drop jsonschema, py-cpuinfo, pyyaml, schedule. sev-snp-measure is kept: it is imported directly by src/aleph/vm/vprogram/measurement.py.
- packaging/aleph-vm/DEBIAN/control: drop python3-jsonschema, python3-cpuinfo, python3-yaml, python3-schedule, debootstrap; add pciutils.
- .github/workflows/test-using-pytest.yml, test-rust.yml, test-build-examples.yml, Justfile: drop the now-unused python3-nftables, python3-netifaces, python3-cpuinfo, python3-jsonschema from the apt-get/install-system-deps lines.
- .github/workflows/test-using-pytest.yml: also drop types-PyYAML from the pip install line, the mypy stub for the pyyaml dependency this PR removes.
- docker/vm_supervisor-dev.dockerfile: drop python3-nftables, the same leftover apt package as the CI lines (the image already installs the nftables CLI package separately).
- src/aleph/vm/agent/README.md: fix the matching install snippet, replacing python3-nftables with plain nftables (the daemon still needs the `nft` CLI).
- packaging/Makefile does not list any of these packages, so no mirroring change was needed there.

## Test plan
- No unit tests apply to a packaging manifest and CI apt lines; the deb build and droplet CI are the verification.
- `pyproject-fmt --check pyproject.toml` (hatch linting env `mPr7Fnpb`): exit 0, no change.
- Verified every removal with `git grep` across src/, tests/, examples/, scripts/, nix/, runtimes/, .github/ before touching anything; confirmed sev-snp-measure has a live import in src/aleph/vm/vprogram/measurement.py and was kept, and confirmed python3-nftables and types-PyYAML have no remaining consumers before dropping them from the dev dockerfile and the pytest workflow's pip line.
- Validated the edited workflow YAML files parse (`yaml.safe_load`).


foxpatch-aleph (commented):
The dependency-removal claims all verify cleanly: jsonschema, py-cpuinfo, pyyaml and schedule have zero remaining imports in src/, tests/ or examples/ (the cpu_info/schedule hits are lshw-based locals and a local function, not the packages); sev-snp-measure is correctly kept for its live import in src/aleph/vm/vprogram/measurement.py; debootstrap is only used at runtime-image build time so dropping it from the deb Depends is safe; and the pciutils addition is a real bug fix since rust/crates/supervisor-daemon/src/service.rs:96 propagates a missing-lspci failure with ? when ENABLE_GPU_SUPPORT is set, aborting daemon startup. packaging/Makefile is unaffected and the deb Version field is stamped at build time. Two minor completeness gaps remain, both non-blocking.

**`docker/vm_supervisor-dev.dockerfile`** (line 8): Leftover: this dev dockerfile still installs python3-nftables, the same dead package this PR removes everywhere else. Worth dropping here for consistency (the daemon now drives nftables via the `nft` CLI, and plain `nftables` is already installed later on this line).

**`.github/workflows/test-using-pytest.yml`** (line 43): Now that pyyaml is dropped from pyproject.toml and nothing in the tree imports yaml, the `types-PyYAML` stub package pip-installed here is also stale and could be removed alongside this cleanup (harmless either way, since mypy runs with ignore_missing_imports = true).

foxpatch-aleph (approved):
Well-scoped dependency pruning, thoroughly verified. All four removed Python packages (jsonschema, py-cpuinfo, pyyaml, schedule) have zero remaining imports anywhere in the repo, and sev-snp-measure is correctly retained with a live import in src/aleph/vm/vprogram/measurement.py:18. The pciutils addition to the deb Depends is a genuine fix: the Rust daemon invokes lspci at startup when GPU support is enabled and the error propagates (rust/crates/supervisor-daemon/src/service.rs:96), so the package was a missing hard runtime dependency. Dropping debootstrap from the deb is safe since only the build-time runtimes/ scripts use it and the deb ships no script invoking it; it's correctly retained in the CI/Justfile lists that build runtime images. The nftables package is correctly kept everywhere given the daemon drives the nft CLI. Two minor leftovers: docker/vm_supervisor-dev.dockerfile:8 still installs python3-nftables, and .github/workflows/test-using-pytest.yml:43 still installs the types-PyYAML stub even though yaml imports are gone. Neither is blocking.

**`docker/vm_supervisor-dev.dockerfile`** (line 8): Leftover: this dev Dockerfile still installs python3-nftables, the only apt list the pruning missed. Nothing imports the nftables Python bindings anymore (the daemon drives the `nft` CLI), so it can be dropped here for consistency.

**`.github/workflows/test-using-pytest.yml`** (line 43): types-PyYAML can be dropped from this pip install too: since pyyaml was removed from dependencies and nothing imports yaml anywhere, the mypy stub package is no longer needed. Harmless either way.

foxpatch-aleph (approved):
Thorough, well-verified dependency prune. I independently confirmed every removal: jsonschema, py-cpuinfo, pyyaml and schedule have zero imports (and zero dangling attribute references) across all .py files; sev-snp-measure is correctly kept for its live import in src/aleph/vm/vprogram/measurement.py; python3-nftables/netifaces have no consumers since nftables moved to the nft CLI; debootstrap's only use is the build-time runtimes/ create_disk_image.sh and the CI jobs that still install it; and types-PyYAML is safe to drop since mypy-covered paths contain no yaml imports. The pciutils addition to the deb control is a genuine bug fix — lspci is invoked at runtime by both the Rust daemon and the Python code in src/aleph/vm/resources.py, and was previously only satisfied incidentally. I also verified the packaging/Makefile really does omit these packages and that the edited workflow YAMLs parse. No correctness, security, or quality issues found.

**`packaging/aleph-vm/DEBIAN/control`** (line 6): Nit: the PR description attributes the pciutils need only to the Rust daemon's lspci probe, but the Python GPU probe in src/aleph/vm/resources.py:103,147 also shells out to `lspci` on the installed node — so this fix covers both paths. No change needed, just a doc observation.

PR #1209, opened 2026-09-09
fix(agent): keep a refused VM out of the reconciler's teardown set

odesenfans:
## Summary
The convergence loop deletes every VM the plan does not list, so the rule it needs is that **a VM hash the push listed, in any form, is never one that push dropped**. Neither half of that rule held.

A planned VM the supervisor holds STOPPED or FAILED is not "unchanged", so `compute_verdict` sizes it as a capacity candidate; `simulate` charged it its full declared disk with none of the held-volume discount `check_message` applies, so a nearly full node answered `insufficient_capacity`. `narrow_plan` dropped the hash from the plan the reconciler converges on, and `_teardown_dropped` read that absence as a deletion: it retires the VM GONE, which drops the record and the DB rows and reaps the disks under the default `VOLUME_RETENTION=reap`. The scheduler had been told "rejected", which is not "deleted". `node_hash_unknown`, transient right after an agent restart, refuses every node-pinned VM the same way, so a restart on a busy node could wipe them.

The same held for an entry whose embedded message does not verify: `build_plan` rejects it and the hash never entered the plan at all, so one corrupt entry, from a scheduler bug or a bad CCN read, deleted a VM running here perfectly well.

Now every hash the push named is carried by the plan, refused or accepted, and a pass tears down only what the push never named. `simulate` applies the held-volume discount too, so a re-listed VM is not refused for disk it already occupies.

The answer had to say the same thing as the loop, and did not. A RUNNING VM whose embedded message `build_plan` refused is out of `plan.entries`, so `compute_verdict` fell through to the drop branch and listed it under `removing`, and the e2e test's own payload carried one hash under `rejected` and under `removing` at once. A scheduler that believes `removing` stops naming the hash, and the next push, naming it nowhere, is exactly the teardown this PR carries the refusals forward to prevent, so the protection was defeated one push later. The list also reached `simulate` as released capacity, crediting a still-running VM's memory to the other candidates and making the advisory answer more permissive than the enforced check. Any hash the plan `lists()` is now answered under `retained`, with `refused` as the reason, and never reaches `simulate` as released.

## Changes
- `allocation/plan.py`: `AllocationPlan` gains `refused` and a `lists()` predicate, so a plan can say whether the push named a VM at all, not just whether it wants it started.
- `allocation/verdict.py`: `build_plan` names the hashes it refused over a message it would not verify (an entry whose hash could not be parsed is left out: it names no VM here, so it has nothing to protect), `narrow_plan` carries those through and adds the ones the answer refused.
- `allocation/verdict.py`: `compute_verdict` answers a hash the plan lists under `retained` (`refused`) instead of `removing`, so the answer and the loop agree and the VM's memory is not credited to the rest of the plan.
- `allocation/reconciler.py`: `_teardown_dropped` skips any hash the push named, on the pass snapshot and on the mid-pass re-read alike.
- `allocation/reconciler.py` and `allocation/plan.py` comments: the refused set was described as holding transient refusals only. `node_mismatch` is permanent, so both now say the set holds every refusal and that waiting for the push to stop naming the VM is the safe reading of either kind. `simulate`'s docstring notes the synchronous storage-pool walk it does per candidate on the caller's await-free path.
- `agent/capacity.py`: the disk arithmetic `check_message` did inline moves into `discounted_disk`, and `simulate` goes through it per candidate. `ResourceRequirements` carries the declared volumes it was summed from so the lookup is possible; requirements built as bare scalars are judged as given. The batch disk accumulator charges the discounted figure, so space a candidate already holds is not taken from the rest of the plan.

## Test plan
- `tests/supervisor/test_allocations_v2.py`: end to end, a push naming a RUNNING VM with a corrupt embedded message answers `invalid_message` and the next reconcile pass does not tear the VM down (on the unfixed code the mock teardown is awaited twice). The same payload now asserts `removing == []` and `retained[hash] == "refused"`, pinning the answer and the loop to one story.
- `tests/supervisor/test_allocation_reconciler.py`: a VM the answer only refused is not torn down (STOPPED and FAILED), a refusal does not spare the VMs the push left out, and the mid-pass re-read covers a refusal like it covers a re-add.
- `tests/supervisor/test_allocation_verdict.py`: a RUNNING hash the plan refused is `retained` rather than `removing`, and `simulate` is called with an empty `releasing`; narrowing carries the refused hashes, for a stopped VM refused `insufficient_capacity`, for one refused `node_hash_unknown`, and for one `build_plan` refused over its message; an unparseable hash is refused without being protected.
- `tests/supervisor/test_agent_capacity.py`: `simulate` discounts disk the candidate already holds, still charges a candidate that holds nothing, and does not charge a held volume to the rest of the plan.
- Checks: `ruff format --diff` and `isort --check-only --profile black` on every changed file, and `mypy src/aleph/vm/ tests/` from the worktree root, all clean. The test suites are left to CI on this round.

### DIFF
--- a/src/aleph/vm/agent/allocation/plan.py
+++ b/src/aleph/vm/agent/allocation/plan.py
@@ -88,13 +88,18 @@ class PlannedVm:
 class AllocationPlan:
     """What the push asked for, once the answer has had its say.
 
-    ``refused`` are the hashes the push listed and the answer turned down.
-    They are out of ``entries`` because nothing is to start them, and they are
-    named here because the convergence loop reads a hash the plan does not
-    hold as one the scheduler took away, and deleting a VM means reaping its
-    disks. "Rejected" is not "deleted": the scheduler still believes the VM
-    exists, and every refusal the answer gives is temporary anyway (no room
-    right now, a node that has not read its own hash back yet).
+    ``refused`` are the hashes the push listed and this node turned down,
+    over a message it would not verify or for want of room. They are out of
+    ``entries`` because nothing is to start them, and they are named here
+    because the convergence loop reads a hash the plan does not hold as one
+    the scheduler took away, and deleting a VM means reaping its disks.
+    "Rejected" is not "deleted": the scheduler still believes the VM exists,
+    and the shapes that produce a refusal (a full host, a node that has not
+    read its own hash back since it restarted, a corrupt message from a buggy
+    scheduler or a bad CCN read) are all things that pass.
+
+    A hash the push sent in a form we could not parse is absent from both:
+    it names no VM on this node, so there is nothing for it to protect.
     """
 
     plan_id: str
--- a/src/aleph/vm/agent/allocation/plan.py
+++ b/src/aleph/vm/agent/allocation/plan.py
@@ -86,9 +86,25 @@ class PlannedVm:
 
 @dataclass(frozen=True)
 class AllocationPlan:
+    """What the push asked for, once the answer has had its say.
+
+    ``refused`` are the hashes the push listed and the answer turned down.
+    They are out of ``entries`` because nothing is to start them, and they are
+    named here because the convergence loop reads a hash the plan does not
+    hold as one the scheduler took away, and deleting a VM means reaping its
+    disks. "Rejected" is not "deleted": the scheduler still believes the VM
+    exists, and every refusal the answer gives is temporary anyway (no room
+    right now, a node that has not read its own hash back yet).
+    """
+
     plan_id: str
     received_at: datetime
     entries: dict[ItemHash, PlannedVm]
+    refused: frozenset[ItemHash] = frozenset()
+
+    def lists(self, vm_hash: ItemHash) -> bool:
+        """Whether the push this plan came from named this VM at all."""
+        return vm_hash in self.entries or vm_hash in self.refused
 
 
 @dataclass
--- a/src/aleph/vm/agent/allocation/plan.py
+++ b/src/aleph/vm/agent/allocation/plan.py
@@ -93,10 +93,13 @@ class AllocationPlan:
     ``entries`` because nothing is to start them, and they are named here
     because the convergence loop reads a hash the plan does not hold as one
     the scheduler took away, and deleting a VM means reaping its disks.
-    "Rejected" is not "deleted": the scheduler still believes the VM exists,
-    and the shapes that produce a refusal (a full host, a node that has not
-    read its own hash back since it restarted, a corrupt message from a buggy
-    scheduler or a bad CCN read) are all things that pass.
+    "Rejected" is not "deleted": the scheduler still believes the VM exists.
+    Most of the shapes that produce a refusal pass on their own (a full host,
+    a node that has not read its own hash back since it restarted, a corrupt
+    message from a buggy scheduler or a bad CCN read), and one does not: a VM
+    allocated to another node stays allocated to it. The set holds every
+    refusal all the same, transient or permanent, because waiting for the push
+    to stop naming the VM is the safe reading of both.
 
     A hash the push sent in a form we could not parse is absent from both:
     it names no VM on this node, so there is nothing for it to protect.
--- a/src/aleph/vm/agent/allocation/reconciler.py
+++ b/src/aleph/vm/agent/allocation/reconciler.py
@@ -168,9 +168,12 @@ class AllocationReconciler:
             # answer refused is named: the scheduler was told the VM was
             # rejected, not that it was deleted, so it still believes the VM
             # is here, while a teardown retires it GONE and reaps its disks.
-            # Every refusal is temporary too, so the loop would be destroying
-            # a VM over a full disk or over a node hash it has not read back
-            # since its last restart.
+            # The set holds every refusal, transient or not. Most of them pass
+            # on their own (a full disk, a node hash not read back since the
+            # last restart), and one does not: a VM allocated to another node
+            # stays allocated to it. Waiting for a push to stop naming the VM
+            # is the safe reading either way, since the scheduler that placed
+            # it elsewhere is the one that will stop naming it here.
             if plan.lists(vm_hash) or info.status not in TEARDOWN_STATUSES:
                 continue
             # The plan is re-read here rather than taken from the pass, which
--- a/src/aleph/vm/agent/allocation/reconciler.py
+++ b/src/aleph/vm/agent/allocation/reconciler.py
@@ -164,7 +164,14 @@ class AllocationReconciler:
 
     async def _teardown_dropped(self, plan: AllocationPlan, known: dict[ItemHash, VmInfo]) -> None:
         for vm_hash, info in known.items():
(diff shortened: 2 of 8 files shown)


foxpatch-aleph (changes requested):
Confirmed: `compute_verdict` never consults `plan.refused` when composing `removing`, so the guard exists only in the reconciler.

## Review

The core fix is sound and well-executed:

- **The invariant holds in the loop.** Both refusal routes (`build_plan`'s verify failures, `compute_verdict`'s admission rejections) land in `refused`, `narrow_plan` carries them forward, and `_teardown_dropped` consults `lists()` on both the pass snapshot and the mid-pass re-read. The duplicate-entry interplay in `build_plan` (`vm_hash in rejected` guard plus `entries.pop`) keeps `entries` and `refused` disjoint. Unparseable hashes correctly stay out of `refused`.
- **The capacity refactor is behavior-preserving.** `discounted_disk` is a faithful extraction of the arithmetic `check_message` did inline; charging the discounted figure into `committed_disk` is right because `_available_disk_bytes()` already excludes space the held files occupy. Requirements built as bare scalars keep the old path, so the pre-existing `simulate` tests are unaffected.

The one genuine gap is on the answer side: a RUNNING VM whose message `build_plan` refused is not in `plan.entries`, so `compute_verdict` still appends it to `verdict.removing` (verdict.py:198) and credits its memory as released to the other candidates via `simulate(releasing=...)` (verdict.py:230) — while the reconciler, per this very PR, will never tear it down. The scheduler is told the VM is going away when the whole point of the change is that it is not; if it trusts `removing` and drops the hash from its next push, the VM is destroyed one push later and the protection is defeated. On the parent commit the answer was at least self-consistent. The e2e test's own payload contains both `rejected` and `removing` for the same hash. The answer needs to skip hashes the plan `lists()` too (e.g., report them as `retained` with a reason).

```json
{
  "verdict": "REQUEST_CHANGES",
  "summary": "The loop-side fix is correct and well-tested: both refusal routes feed the new AllocationPlan.refused set, narrow_plan carries them, and _teardown_dropped honors plan.lists() on both the pass snapshot and the mid-pass re-read, so a refused VM is no longer read as a deletion. The capacity refactor is a faithful extraction of check_message's discount arithmetic, and charging the discounted figure into the batch accumulator is the right accounting since _available_disk_bytes() already excludes space held files occupy. However, the fix only covers the reconciler: compute_verdict still answers a RUNNING VM whose message build_plan refused (the PR's own headline scenario) as \"removing\" and credits its committed memory as released to the other candidates via simulate(releasing=...), even though the loop now keeps that VM running. The answer and the convergence now contradict each other, and a scheduler that trusts \"removing\" and drops the hash from its next push will have the VM destroyed one push later anyway, defeating the protection. The answer side needs to skip hashes the plan lists() as well.",
  "comments": [
    {
      "path": "src/aleph/vm/agent/allocation/verdict.py",
      "line": 198,
      "body": "A refused hash is not in plan.entries, so a RUNNING VM whose embedded message build_plan refused falls through to this branch and is answered \"removing\" — while the reconciler, per this PR, will never tear it down. The e2e test's own payload carries both rejected[vm_hash] and removing: [vm_hash] for the same VM. Two consequences: (1) the scheduler is told the VM is going away, and if it trusts that and drops the hash, the next push tears the VM down anyway, defeating the protection in the exact corrupt-message scenario this PR fixes; (2) verdict.py:230 passes this list as releasing to simulate, crediting the still-running VM's committed memory to the other candidates — an advisory yes weaker than the enforced check, against the module's own stated principle. compute_verdict should skip hashes the plan lists() when composing removing (report them as retained with a reason, or at minimum exclude them from the releasing credit).",
    },
    {
      "path": "src/aleph/vm/agent/capacity.py",
      "line": 742,
      "body": "Note: this puts a synchronous storage-pool directory walk on compute_verdict's await-free critical path, once per candidate, with the event loop held — a large plan now costs N namespace walks per push. check_message already does the same per create and _check_against already stats the pools per candidate, so this follows an established pattern, but it might deserve a line in the simulate docstring since the handler contract (\"nothing may yield\") makes the blocking window easy to overlook.",
    },
    {
      "path": "src/aleph/vm/agent/allocation/reconciler.py",
      "line": 171,
      "body": "\"Every refusal is temporary too\" is not true of every refusal the answer can give: node_mismatch is permanent, and narrow_plan carries it into refused like the transient ones. Keeping such a VM running until the push stops naming it is the safe reading and I think the right call, but the comment (and the matching \"are all things that pass\" phrasing in AllocationPlan's docstring) overstates the case — worth a word so a future reader does not assume the set is transient by construction.",
    },
    {
      "path": "tests/supervisor/test_allocations_v2.py",
      "line": 200,
      "body": "The \"not torn down\" assertion is the right one for the loop, but this test's response payload is also where the \"simultaneously rejected and removing\" inconsistency is visible — asserting removing == [] here (once the verdict is fixed) would pin the answer and the loop to the same semantics.",
    }
  ]
}
```

foxpatch-aleph (approved):
The fix is correct and well-executed: carrying refused hashes alongside the plan entries, with a lists() predicate used at both teardown checkpoints, closes the gap where a refused VM was retired GONE and its disks reaped, and the answer now agrees with the loop (retained:refused instead of removing). Extracting discounted_disk unifies the enforced and advisory disk paths, and the simulate discount fixes a real bug where a re-listed stopped VM was refused for space its own files occupy. Tests cover the chain end to end, including the mid-pass re-read race and the negative case. Two subtle observations, neither blocking: (1) under VOLUME_RETENTION=retain, a re-listed candidate's held-but-reclaimable bytes are no longer reserved in the batch accumulator, so other candidates in the same plan can be admitted against them in the advisory answer — a looseness the enforced path re-judges, but it's the one direction the docstring doesn't document; (2) a refused VM that is also non-removable by policy now reports 'refused' rather than its policy reason in retained.

**`src/aleph/vm/agent/capacity.py`** (line 728): Non-blocking: with VOLUME_RETENTION=retain, a re-listed candidate's held files are still counted as reclaimable-free by _available_disk_bytes(), and now that only the delta is charged, committed_disk no longer implicitly reserves them for the rest of the batch. Another candidate in the same plan can therefore be admitted against bytes the re-listed VM will adopt. The enforced path re-judges at create time (or placement evicts the retained directory, which is the designed mechanism), so this is advisory-only looseness — but it is the one spot where the advisory answer is now more permissive than the enforced one, and the docstring only claims the 'never stricter' direction. A sentence in the simulate docstring noting the retain case would make the trade-off explicit.

**`src/aleph/vm/agent/allocation/verdict.py`** (line 206): Minor: a hash the push refused that is also non-removable by policy (payment stream, GPU, confidential) now answers retained:"refused" where it previously answered its policy reason ("payment_stream", etc.). Both mean kept, so nothing operational changes, but the scheduler loses the why in that overlap. Worth a word in the comment if it was deliberate.

**`src/aleph/vm/agent/capacity.py`** (line 748): When requirements.volumes is set, the recomputed figures silently replace requirements.disk_mib / max_volume_mib. That is correct today because requirements_from_message is the only production constructor and derives both from the same volumes, but a future caller that assembles requirements with scalars independent of volumes would be quietly overridden. The docstring's 'built from a message' framing covers it; just flagging the coupling.

PR #1210, opened 2026-09-09
fix(cli): give aleph-vm storage its config, logs and database

odesenfans:
## Summary
A hand-run `aleph-vm storage` command was missing everything the systemd units set up for the agent daemon, so it ran against a configuration the node does not have. The units inject `/etc/aleph-vm/supervisor.env` with `EnvironmentFile=`, and pydantic reads only the process environment plus a cwd `.env`: on a node configured with `VOLUME_RETENTION=keep`, `aleph-vm storage reconcile` therefore ran with the `reap` default and evicted every retained directory (with non-default pools it died in `setup_pools`, with a non-default execution root it read an empty DB). `_load_registry` also skipped the table creation and alembic migrations the daemon's startup runs, so `status` / `list` on a fresh install crashed with a raw `OperationalError` and left an empty sqlite file behind. And nothing configured logging, so every "Deleted volume" / "Removed volume directory" / "Marked reclaimable" line was dropped.

This loads the node's environment file (`--env-file`, else `$ALEPH_VM_ENV_FILE`, else the packaged path) and rebuilds the settings singleton from it, shares the daemon's database setup, configures a stderr handler at INFO, and turns `storage` into a real argparse subparser so the agent's global flags placed before the verb parse.

## Changes
- `src/aleph/vm/agent/storage_cli.py`: `run_parsed()` sets the process up before any verb runs. It loads the env file (existing environment values still win; a file named explicitly, by `--env-file` or by `$ALEPH_VM_ENV_FILE`, that does not exist is an error rather than a silent fall back to the defaults), installs a named stderr log handler at INFO with `--loglevel` (validated against the known level names, so an unknown one is a clean usage error rather than a raw `ValueError` traceback), and runs the shared database setup before the registry is rehydrated. `status` and `list` are read-only, so they refuse a database that is not there and exit 1 rather than create one: an operator pointed at the wrong execution root sees the mistake instead of an empty listing. An invalid value in the env file (a `pydantic.ValidationError` from the settings reload) is caught and printed field by field to stderr with exit 2, instead of escaping as a traceback. The verb parser moved into `add_arguments()`, reused by the standalone parser and by the new `add_subparser()`; the verb's dest is `storage_command` now that `command` belongs to the agent parser.
- `src/aleph/vm/agent/cli.py`: the daemon's table creation plus migrations factored into `initialise_database()`, called from `main()` and from the storage CLI. `storage` registered as a subparser instead of being dispatched by sniffing `sys.argv[1]` before `parse_args`, and a `--loglevel LEVEL` global flag added alongside `-v` / `-vv`, validated against the same tuple of level names the storage subparser uses (the tuple lives here, since the storage module already imports this one), so `aleph-vm --loglevel verbos storage status` is a usage error too. The subparser's own `--loglevel` defaults to `SUPPRESS` so it does not clobber what the parent parsed.
- `pyproject.toml`: the comment claiming a flat CLI with no subcommands was wrong.
- `docs/architecture/storage.md`: the CLI section documents the env file, the logging, and the database rules.

## Test plan
- `tests/supervisor/test_storage_cli.py`: nine new or rewritten tests, plus five more from the Foxpatch review rounds: an unknown `--loglevel` value is a usage error and not a traceback on the storage parser and on the agent parser ahead of the subcommand, `reconcile` creates a missing database through the real table creation and migrations (the plumbing fixture now restores the whole settings namespace after each test, since an env file rebuilds every field), `$ALEPH_VM_ENV_FILE` naming a missing file refuses exactly as `--env-file` does, and an invalid value in the env file exits 2 with the offending field named on stderr. The env file reaches the settings when the CLI runs and a named file that is missing exits 1; `status` refuses a missing database, creates no file, and does not call the initialiser; the migrations run before the registry is read, and a real empty sqlite file is brought up to date end to end through the real alembic path; a `reconcile` purge reports its INFO lines on stderr; the agent parser accepts a global flag before the verb, and the storage subparser keeps its own `--loglevel`.
- `pytest tests/supervisor/test_storage_cli.py` 35 passed.
- `ruff format --diff`, `isort --check-only --profile black`, `mypy src/aleph/vm/ tests/`: all clean (exit 0). A whole-`tests/supervisor` collection currently fails on this checkout on two unrelated files (`test_snp_instance_run.py`, `test_vprogram_launch.py`: `ImportError: cannot import name 'ConfidentialGpuRequirement'`), from the confidential-GPU feature merged into `dev-2.1` after this chain was opened (#1200) needing a newer `aleph_message` than this venv has installed.

### DIFF
--- a/docs/architecture/storage.md
+++ b/docs/architecture/storage.md
@@ -538,10 +538,9 @@ goes first", which is the only promise a node can honestly sell.
 ### CLI
 
 Storage is agent-side and readable from the filesystem plus the agent DB, so
-`aleph-vm storage ...` (`src/aleph/vm/agent/storage_cli.py`, dispatched by
-`agent/cli.py` before its own argument parser, which points at
-`storage --help` in its own `--help` epilog) needs no running *agent
-process*:
+`aleph-vm storage ...` (`src/aleph/vm/agent/storage_cli.py`, registered as a
+subparser on `agent/cli.py`'s own parser, which points at `storage --help`
+in its `--help` epilog) needs no running *agent process*:
 
 - `storage status`: per pool, live / reclaimable / cache / free bytes
   against the budgets. Read-only, registry live set only.
@@ -579,6 +578,34 @@ process*:
   (`remove_parent_device` per evicted ref, then `sweep_leaked_cache_loops`),
   so an evicted entry never leaves `/dev/mapper/<ref>` and its loop device
   pinning the deleted inode.
+
+Running with no agent process also means setting the process up the way the
+systemd units set it up for the daemon, which the CLI does before it reads
+anything:
+
+- **Configuration.** `EnvironmentFile=/etc/aleph-vm/supervisor.env` in the
+  units is what gives the daemon its settings; nothing gives them to an
+  operator's shell, and pydantic reads only the process environment and a
+  `.env` in the working directory. The CLI therefore loads that file itself
+  (`--env-file`, else `$ALEPH_VM_ENV_FILE`, else the packaged path) and
+  rebuilds the settings singleton from it, logging which file it used or
+  that there was none. Values already in the environment win, so a one-off
+  override on the command line still works, and a `--env-file` that does
+  not exist is an error rather than a silent fall back to the defaults: a
+  pass run with `VOLUME_RETENTION` at its `reap` default on a node
+  configured to keep would evict every retained directory.
+- **Logging.** Most of what a pass has to report it reports through the
+  logger the reconciler, the purge and the marker already write to, so the
+  CLI configures a stderr handler at `INFO` (`--loglevel`, or the agent
+  CLI's own `-v` / `-vv` / `--loglevel` placed before the verb). Without it
+  those lines are dropped and warnings arrive as bare last-resort lines.
+- **Database.** `cli.initialise_database()` (the tables, then the alembic
+  migrations) is shared with the daemon's startup and runs before the
+  registry is rehydrated: an unmigrated or absent database otherwise fails
+  with a raw sqlite error. `status` and `list` are the exception; being
+  read-only, they refuse a database that does not exist and exit `1` rather
+  than create an empty one, so an operator pointed at the wrong execution
+  root sees the mistake instead of an empty listing.
 
 `reconcile` and `reclaim` can purge, so they apply the same fail-closed rule
 `reconciler._startup_refusal` applies to the daemon's own startup pass: a
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -80,8 +80,9 @@ allow-direct-references = true
 platforms = [ "linux" ]
 
 [tool.hatch.envs.default.scripts]
-# The aleph-vm console script (aleph.vm.agent.cli:main) is a flat argparse
-# CLI, no subcommands; the supervisor daemon is the Rust binary under rust/.
+# The aleph-vm console script (aleph.vm.agent.cli:main) runs the agent with
+# no subcommand; `aleph-vm storage <verb>` is its one subcommand. The
+# supervisor daemon is the Rust binary under rust/.
 agent = "aleph-vm {args:--help}"
 config = "aleph-vm --print-settings --do-not-run {args}"
 
--- a/src/aleph/vm/agent/cli.py
+++ b/src/aleph/vm/agent/cli.py
@@ -21,6 +21,12 @@ from .custom_logs import setup_handlers
 logger = logging.getLogger(__name__)
 
 
+# Validated against --loglevel, here and on the storage subcommand: an
+# unknown name must be a clean argparse usage error (exit 2), not a raw
+# ValueError traceback out of logging.Logger.setLevel further down.
+LOG_LEVEL_NAMES = ("DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL")
+
+
 def parse_args(args):
     parser = argparse.ArgumentParser(
         prog="aleph-vm",
@@ -127,11 +133,12 @@ def parse_args(args):
         "--loglevel",
         dest="loglevel",
         type=str.upper,
+        choices=LOG_LEVEL_NAMES,
         # -v/--verbose is registered first and already supplies the default,
         # so this one must not offer a second one: argparse would keep the
         # first anyway, and SUPPRESS says so out loud.
         default=argparse.SUPPRESS,
-        help="Log level by name (DEBUG, INFO, WARNING, ERROR)",
+        help="Log level by name (DEBUG, INFO, WARNING, ERROR, CRITICAL)",
     )
     subparsers = parser.add_subparsers(dest="command")
     # Imported here rather than at module scope: storage_cli imports this
--- a/src/aleph/vm/agent/cli.py
+++ b/src/aleph/vm/agent/cli.py
@@ -123,6 +123,22 @@ def parse_args(args):
         default=False,
         help="Authorize the developer's SSH keys to connect instead of those specified in the message",
     )
+    parser.add_argument(
+        "--loglevel",
+        dest="loglevel",
+        type=str.upper,
+        # -v/--verbose is registered first and already supplies the default,
+        # so this one must not offer a second one: argparse would keep the
+        # first anyway, and SUPPRESS says so out loud.
+        default=argparse.SUPPRESS,
+        help="Log level by name (DEBUG, INFO, WARNING, ERROR)",
+    )
+    subparsers = parser.add_subparsers(dest="command")
+    # Imported here rather than at module scope: storage_cli imports this
+    # module for the shared database setup below.
+    from aleph.vm.agent import storage_cli
+
+    storage_cli.add_subparser(subparsers)
     return parser.parse_args(args)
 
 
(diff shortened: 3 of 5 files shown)


foxpatch-aleph (approved):
Well-engineered change with thorough, high-quality tests; I verified the 32 storage CLI tests pass, formatting is clean, and the tricky argparse SUPPRESS handling is both correct and necessary (confirmed empirically that a subparser default clobbers the parent's parsed value on Python 3.12). The shared initialise_database(), fail-closed env-file loading, and read-only DB refusal are all sound, and settings singleton rebuilding works correctly under pydantic-settings v2. Remaining findings are non-blocking polish items: an invalid --loglevel value surfaces as a raw ValueError traceback instead of a usage error, a $ALEPH_VM_ENV_FILE pointing at a missing file is silently ignored even though the PR's own fail-closed rationale applies to it, and the -v default (settings.LOG_LEVEL) is captured before the env file is loaded so ALEPH_VM_LOG_LEVEL does not affect the CLI's default verbosity.

**`src/aleph/vm/agent/storage_cli.py`** (line 483): An invalid `--loglevel` value (e.g. `--loglevel verbos`) reaches `root.setLevel("VERBOS")` in `_setup_logging` and dies with a raw `ValueError: Unknown level: 'VERBOS'` traceback. `choices=` (or a try/except around setLevel that exits 2) would turn it into a clean usage error, consistent with the PR's own care around operator ergonomics. Verified: `type=str.upper` alone does not validate.

**`src/aleph/vm/agent/storage_cli.py`** (line 467): A missing file is only fatal for `--env-file`; when the path comes from `$ALEPH_VM_ENV_FILE` it falls through with an INFO log, so an operator with a typo'd env var silently runs on the built-in defaults — exactly the hazard the docstring names ("running on the defaults they were trying to override is worse than refusing"). Consider treating an explicit env-var path as an error too, or at least WARNING.

**`src/aleph/vm/agent/storage_cli.py`** (line 455): `settings.__dict__.update(Settings().__dict__)` works under pydantic v2 for reads, but note it does not update `__pydantic_fields_set__`, so any `model_dump(exclude_unset=True)` on the singleton after a reload would report a stale field set. Nothing in the storage CLI path does that today (`settings.display()` is unused here), so this is informational — a short comment noting the limitation would help future readers.

**`src/aleph/vm/agent/cli.py`** (line 64): The `-v` default is `settings.LOG_LEVEL` captured at parse time, i.e. before `_load_env_file()` runs, so `ALEPH_VM_LOG_LEVEL` in the node's env file does not affect the storage CLI's default verbosity (only an explicit `--loglevel` does). Small inconsistency with the PR's stated goal of running on the node's config; fine if intentional, but worth a word in the docs alongside the `--loglevel` mention.

**`tests/supervisor/test_storage_cli.py`** (line 439): The `_restore_root_logger` autouse fixture restores the root logger's handlers and level, but `_setup_logging` also pins `aiosqlite` and `sqlalchemy.engine` to WARNING and those levels leak for the rest of the session. Harmless today, but restoring them (or recording and resetting in the fixture) would keep the isolation airtight.

foxpatch-aleph (approved):
This PR is correct and well-executed. The argparse subparser conversion works exactly as claimed (verified against 7 invocation patterns, including the SUPPRESS-default subtlety that prevents the subparser from clobbering the parent's parsed loglevel), the env-file loading correctly precedes settings setup with existing environment values winning, and the settings-reload via __dict__ swap propagates because no reachable module captures settings at import time. The fail-closed choices are right: a missing --env-file is an error rather than a silent run on defaults, and status/list refusing to create a database is sound. All 32 tests pass, and ruff/isort/mypy are clean. Only minor nits remain.

**`src/aleph/vm/agent/storage_cli.py`** (line 467): Minor inconsistency: when $ALEPH_VM_ENV_FILE names a path that does not exist, the CLI silently proceeds on defaults with only an INFO line, while the same operator intent expressed via --env-file is a hard error. If the env var is considered equally explicit, the missing-file branch in _load_env_file (the `if explicit` check) could also consider a value that came from the environment; otherwise the asymmetry deserves a word in the docstring.

**`src/aleph/vm/agent/storage_cli.py`** (line 483): `--loglevel TRACE` (any unknown level name) reaches logging.Logger.setLevel and raises a raw ValueError traceback instead of a clean argparse usage error. A type= function or choices= on the argument would fail with exit 2 and a proper message. Same applies to the agent-level --loglevel in cli.py.

**`src/aleph/vm/agent/storage_cli.py`** (line 489): Nit: `Path(str(settings.EXECUTION_DATABASE))` is a redundant round-trip — the setting is already a Path by the time Settings.__init__ has filled in the derived default. `settings.EXECUTION_DATABASE` alone would do.

**`tests/supervisor/test_storage_cli.py`** (line 469): This test implicitly depends on ALEPH_VM_EXECUTION_DATABASE being absent from the process environment: load_dotenv(override=False) will not replace it, and the settings reload replaces the plumbing fixture's monkeypatched EXECUTION_DATABASE, so the DB check would then look at the environment's path instead. A `monkeypatch.delenv("ALEPH_VM_EXECUTION_DATABASE", raising=False)` would make it airtight.

**`src/aleph/vm/agent/storage_cli.py`** (line 455): An invalid value in the env file (e.g. ALEPH_VM_VOLUME_RETENTION=sometimes) surfaces as a raw pydantic ValidationError traceback from run_parsed. This matches what the daemon does at import time, so it is consistent rather than wrong, but a CLI catching ValidationError to print the offending field would be friendlier to the operator who just wrote the file.

foxpatch-aleph (changes requested):
The core plumbing is solid and well tested: env-file loading with correct fail-closed semantics for both --env-file and $ALEPH_VM_ENV_FILE, correct ordering (settings.setup() before EXECUTION_DATABASE is read, SUPPRESS defaults so the subparser doesn't clobber the parent's loglevel, read-only verbs refusing to create a DB, shared initialise_database with clean ValidationError reporting). Tests are thorough (35 pass), ruff format and isort are clean, and mypy reports no errors in the changed files. However, there is one confirmed gap: the new agent-level --loglevel in cli.py is unvalidated, so `aleph-vm --loglevel verbos storage status` still dies with a raw ValueError traceback — the exact failure mode this PR's head commit claims to fix and its own test asserts must be a clean argparse usage error. Adding choices to the agent parser's flag (mirroring _LOG_LEVEL_NAMES, e.g. via a lazy import to avoid the storage_cli cycle) closes it; the fix is one line.

**`src/aleph/vm/agent/cli.py`** (line 126): This new agent-level `--loglevel` has no `choices`, unlike the storage subparser's flag at storage_cli.py:135-146. I verified by execution that `aleph-vm --loglevel verbos storage status` parses fine and then crashes in `_setup_logging` with a raw `ValueError: Unknown level: 'VERBOS'` — the exact traceback the head commit ('fail closed on ... a bad loglevel') and `test_an_unknown_loglevel_is_a_usage_error_not_a_traceback` say should be a clean usage error. The daemon path has the same issue via `logging.basicConfig(level=...)` at cli.py:206. Re-using `choices` (the `_LOG_LEVEL_NAMES` tuple, imported lazily to avoid the storage_cli cycle) would make both flags behave identically.

**`src/aleph/vm/agent/storage_cli.py`** (line 145): Nit: the help text says '(DEBUG, INFO, WARNING, ERROR)' but `_LOG_LEVEL_NAMES` (and argparse's own error message) also accept CRITICAL. The agent parser's `--loglevel` help has the same omission. Cosmetic, but the two should agree with the choices.

**`src/aleph/vm/agent/storage_cli.py`** (line 518): Minor test gap: nothing covers a write verb (reconcile/reclaim) creating the database when it is absent — the only end-to-end migration test (`test_an_unmigrated_database_is_brought_up_to_date`) starts from an existing empty file, and the `plumbing` fixture stubs `initialise_database` for everything else. A test that `reconcile` on a missing DB logs the 'Creating the agent database' line and proceeds would pin this branch down.

**`tests/supervisor/test_storage_cli.py`** (line 469): Minor isolation nit: `_reload_settings()` rebuilds every field of the global settings singleton, and only the fields explicitly monkeypatched (VOLUME_RETENTION, EXECUTION_DATABASE) are restored afterwards. It happens to be harmless today because the env file only sets fields the tests also patch, but a snapshot/restore of `settings.__dict__` in the `plumbing` fixture (or a dedicated fixture) would make that guarantee explicit rather than incidental.

foxpatch-aleph (approved):
This PR correctly gives the hand-run storage CLI the configuration, logging, and database setup the systemd units provide to the daemon. I verified all 37 tests pass against this checkout and drove the real CLI end-to-end: global flags before the verb now parse, an unknown --loglevel is a clean usage error at both parser levels, a missing --env-file or $ALEPH_VM_ENV_FILE refuses with exit 1, an invalid value in the env file exits 2 naming the field, status refuses a missing database without creating it, and a write verb creates and migrates the DB through the real alembic path. I also probed the empty-DB-file edge case and confirmed initialise_database() runs whenever the file exists, so a 0-byte leftover from the old bug is migrated rather than crashing status/list. The fail-closed design is consistent, no other code depends on the old sys.argv dispatch, no security issues (no injection, no secret exposure in error output), and the docs match the verified behavior. Only non-blocking nits remain.

**`src/aleph/vm/agent/storage_cli.py`** (line 482): load_dotenv() raises PermissionError when the file exists but is not readable, which escapes run_parsed as a raw traceback since only ValidationError is caught there. Given this function's own reasoning — a named file that cannot be used must refuse rather than run on defaults — an unreadable file deserves the same clean refusal as a missing one: catch OSError around this call and return False (same exit-1 path). It already fails closed, so this is polish, not a hazard.

**`src/aleph/vm/agent/storage_cli.py`** (line 484): load_dotenv returns False for an empty (but readable) file, so this INFO line then claims a load that did not happen. Cosmetic; checking the return value would keep the message honest.

**`src/aleph/vm/agent/storage_cli.py`** (line 463): Works today (verified: pydantic stores plain fields in __dict__ and Settings has no PrivateAttr/computed fields), but note it silently skips anything pydantic keeps outside __dict__ — worth a comment at least, so a future PrivateAttr on Settings doesn't get dropped here unnoticed.

PR #1211, opened 2026-09-09
fix(agent): hold a reserve, not the whole budget, for an unmeasured download

odesenfans:
## Summary
A download whose response carries no `Content-Length` was charged `min(cap, budget)`
against its cache root, and since `MAX_RUNTIME_ARCHIVE_SIZE` is 100 GiB that is the
whole budget on any realistic root. One chunked-encoding response then held the
entire root for its duration: a concurrent measured download saw usage already at
the cap, evicted every unreferenced entry in the root trying to free room that a
guess was holding, and was refused anyway. The cache was wiped and the create
failed, for a body that may have been a few kilobytes of manifest.
An unmeasured download is now charged `UNKNOWN_LENGTH_RESERVE` (2 GiB by default,
never more than its own cap nor than the budget), and a reservation records whether
its figure is a measurement or a ceiling: a ceiling never sets an eviction target.
Stacked on #1206.

## Changes
- `src/aleph/vm/storage.py`: reservations are a `DownloadReservation` (size plus
  `measured`), so the cache can tell a `Content-Length` from a cap after admission,
  not just during it. `reserve_download` takes `measured` as a keyword.
- `src/aleph/vm/agent/vm/cache.py`: `in_flight_bytes` takes `count_ceilings`. A
  ceiling reservation counts only the bytes it has actually written when the
  question is how much to evict (`evict_caches`), and in full when the question is
  whether the next download fits (`admit_download`). `_admit_unknown_length` charges
  `_unknown_length_charge`: the smallest of the reserve, the download's cap and the
  budget.
- `src/aleph/vm/conf.py`: `UNKNOWN_LENGTH_RESERVE`, defaulting to `2G`, validated
  through `parse_budget` alongside `CACHE_BUDGET`.
- `docs/architecture/storage.md`: the admission paragraph and the settings table.

## Test plan
- `tests/supervisor/test_cache_eviction.py`: a measured download admitted beside an
  unknown-length one with nothing evicted (the reported scenario, failing before);
  a ceiling reservation never driving the pass to evict; the bytes a ceiling
  download writes still counting as they land; total usage inside the budget once
  both downloads complete.
- `tests/supervisor/test_storage_download_cap.py`: the reservation assertions follow
  the new type.
- Checks: `pytest tests/supervisor` (only the pre-existing environment failure
  `test_log.py::test_make_logs_queue`, missing `systemd.journal.Reader`), `ruff
  format --diff`, `isort --check-only`, `mypy` on the five changed Python files, all
  exit 0.

- `cache.py`: the OSError fallback in `_unknown_length_charge` returned the whole budget, the hold this PR removes. It now resolves the reserve against a zero total, which yields the configured size when the reserve is absolute and nothing when it is a percentage of the disk that just refused to answer.
- `tests/supervisor/test_cache_eviction.py`: the two tests whose expected figure was really `min(reserve, budget)` under the 2 GiB default now set `UNKNOWN_LENGTH_RESERVE` explicitly, plus two tests for both shapes of the fallback.

### DIFF
--- a/docs/architecture/storage.md
+++ b/docs/architecture/storage.md
@@ -438,7 +438,8 @@ kilobytes of JSON each, so it never approaches a budget sized for runtime
 images, and one setting is one thing for an operator to reason about. A root's
 usage is its finished entries plus what it owes to downloads that have not
 finished (`in_flight_bytes`): the allocated blocks of `.part` and `.tmp`
-files, and the size any admitted download was promised.
+files, and the size any admitted download was promised (the greater of the
+two for a download still being written, never their sum).
 
 The pass evicts least recently used first, mtime being a real
 signal because the downloader touches an entry on every cache hit
@@ -502,9 +503,17 @@ ceiling for a runtime image and equally the ceiling for a few kilobytes of
 manifest, so charging it as a size would evict a whole cache root for a
 download that never needed the room. An unknown-length download therefore
 never evicts, is refused only when the root is already over its budget, and
-is charged `min(cap, budget)`: bounded, so a stream of them still runs the
-root over budget and the next one is refused, and never more than the budget
-itself. What is admitted is then charged to
+is charged `UNKNOWN_LENGTH_RESERVE` (2 GiB by default), never more than its
+own cap nor than the whole budget. The reserve is a modest, deliberately
+arbitrary figure: charging the budget itself let one chunked response hold a
+whole root, so every other download on it was refused for as long as that one
+ran. It is still charged against the next admission, so a stream of
+unknown-length downloads ends in a refusal, and it is reconciled against the
+truth as the body lands, since the bytes on disk are counted the moment they
+are written. A ceiling never sets the eviction target either: a root's usage
+counts a guessed reservation only at the bytes it has actually written when
+the question is how much to evict, and at the full figure when the question
+is whether the next download fits. What is admitted is then charged to
 the download's `.part` path (`reserve_download`) until `download_file`
 releases it, so a second create arriving while the first is still writing
 sees the room the first was promised rather than only the bytes it has
@@ -534,6 +543,7 @@ budget as it stands.
 | `VOLUME_RECONCILE_INTERVAL` | `3600` | Seconds between periodic passes |
 | `VOLUME_CREATE_GUARD` | `600` | Seconds a young directory or `.part` file is assumed to be an in-flight create |
 | `CACHE_BUDGET` | `20%` | Cap on each download cache (runtime, code, data, message) |
+| `UNKNOWN_LENGTH_RESERVE` | `2G` | Room held for a download whose response carries no `Content-Length` |
 
 `keep` is not a promise of "forever": reclaimable disks are unpaid storage,
 and an attacker does not need to stop paying to create them (create, forget,
--- a/src/aleph/vm/agent/vm/cache.py
+++ b/src/aleph/vm/agent/vm/cache.py
@@ -799,10 +799,15 @@ def _unknown_length_charge(root: Path, budget: int, max_bytes: int | None) -> in
     """The room to hold for a body nobody measured: the smallest of the
     reserve, this download's own cap and the budget."""
     try:
-        reserve = parse_budget(settings.UNKNOWN_LENGTH_RESERVE, shutil.disk_usage(str(root)).total)
+        total = shutil.disk_usage(str(root)).total
     except OSError:
-        logger.warning("Cache directory %s is not accessible; holding its whole budget", root, exc_info=True)
-        return budget
+        # A disk nobody can measure must not resurrect the whole-budget hold:
+        # resolve the reserve against a zero total instead, which yields the
+        # configured size when it is written as an absolute one and nothing
+        # when it is a percentage of the disk that just failed to answer.
+        logger.warning("Cache directory %s is not accessible; holding only the reserve", root, exc_info=True)
+        total = 0
+    reserve = parse_budget(settings.UNKNOWN_LENGTH_RESERVE, total)
     figures = [reserve, budget] if max_bytes is None else [reserve, budget, max_bytes]
     return min(figures)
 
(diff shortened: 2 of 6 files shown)


foxpatch-aleph (approved):
Sound, well-tested fix: unmeasured downloads now hold a bounded UNKNOWN_LENGTH_RESERVE instead of the whole budget, and the measured/ceiling distinction is threaded cleanly through reservations, the eviction target, and admission. All 68 tests in the two touched files pass; isort is clean; the ruff/mypy complaints in my sandbox are on pre-existing lines (tool-version drift), not on this diff. The new tests reproduce the reported scenario (a measured download beside a chunked one), pin the eviction-target semantics, and verify reconciliation as bytes land. Only non-blocking nits: the OSError fallback in _unknown_length_charge returns the full budget (the behavior this PR removes) though it is unreachable in practice, and two pre-existing tests implicitly depend on the 2G default exceeding their tiny budgets.

**`src/aleph/vm/agent/vm/cache.py`** (line 805): The OSError fallback here returns the full `budget` — the exact behavior this PR removes. It is effectively unreachable, because `admit_download` already succeeded at `cache_budget_bytes(root)`, which performs the same `shutil.disk_usage` call and returns early on OSError. If it ever did fire it would resurrect the bug; reserving 0 (or `max_bytes`) would be a safer fallback. Non-blocking.

**`tests/supervisor/test_cache_eviction.py`** (line 647): This assertion (and the one in `test_two_unknown_length_downloads_are_both_charged_the_capped_figure`) implicitly depends on the default `UNKNOWN_LENGTH_RESERVE` (\"2G\") exceeding the tiny absolute budget — the expected `DownloadReservation(8192, ...)` is really `min(reserve, budget)`. If the default ever changes, these break confusingly. The new tests monkeypatch the setting explicitly; doing the same here would make them robust. Non-blocking.

**`src/aleph/vm/agent/vm/cache.py`** (line 180): This conditional is dense (measured/counted/or-written in one line), but the docstring above explains it well and it is the single place the rule lives. No change requested.

foxpatch-aleph (approved):
This PR is a well-reasoned fix for a real problem: an unmeasured download held min(cap, budget) — effectively the whole 100 GiB runtime cap — against its cache root, so one chunked response blocked every other download and wiped the cache on the way to being refused. The replacement design is coherent and internally consistent: reservations now carry a measured flag so admission counts a ceiling in full (leaving room for the body to land) while eviction counts only the bytes actually written (a guess never costs an entry its place), with reconciliation as the body lands. The charge is correctly bounded by min(reserve, budget, max_bytes), the OSError fallback resolves against a zero total (the reverse of the bug it fixes, and both fallback shapes are tested), the new setting is validated alongside CACHE_BUDGET, and the release/rename flow in download_file leaves no accounting gap. All 70 tests in the touched files pass, and the new tests cover the reported scenario, both fallback shapes, the eviction-target exclusion, and the as-bytes-land reconciliation. The mypy and ruff findings in my environment are all in pre-existing untouched lines (tool version skew). Two minor, non-blocking nits below.

**`src/aleph/vm/agent/vm/cache.py`** (line 180): Minor readability nit: the one-line conditional in the accumulator — `total += max(reservation.size_bytes, written) if reservation.measured or count_ceilings else written` — is the densest line in the PR. An explicit `if reservation.measured or count_ceilings: ... else: ...` (or folding it into the loop with an early skip of ceiling reservations when not `count_ceilings`) would make the three cases easier to scan. Not blocking; the docstring does explain it.

**`src/aleph/vm/agent/vm/cache.py`** (line 802): `_unknown_length_charge` calls `shutil.disk_usage` again even though `admit_download` just resolved `budget` from the same stat via `cache_budget_bytes(root)` one call earlier. Passing the total (or folding the reserve resolution into the caller) would remove the duplicate syscall and make the OSError branch here unreachable in practice rather than a second failure path. Purely a nit — the current shape keeps the signature simple and the branch is defensive, so not blocking.

foxpatch-aleph (approved):
The fix is correct and well-executed: charging UNKNOWN_LENGTH_RESERVE instead of min(cap, budget) removes the whole-budget hold, and the DownloadReservation/measured split lets eviction targets discount guessed figures while admission checks still count them in full. All call sites of reserve_download and reserved_downloads were updated, the new setting is validated at startup and documented, and both changed test suites pass (70 tests) when run locally. The remaining observations are non-blocking polish: a dense one-line ternary in in_flight_bytes, a redundant disk stat plus a nearly-unreachable fallback in _unknown_length_charge, over-budget log lines that now under-report usage held by unmeasured downloads, and a missing test pinning that a measured reservation still counts its promised size when count_ceilings=False.

**`src/aleph/vm/agent/vm/cache.py`** (line 180): This one-liner is correct but dense: `total += max(reservation.size_bytes, written) if reservation.measured or count_ceilings else written`. A statement-level `if reservation.measured or count_ceilings: total += max(...) else: total += written` would make the two policies (measured figures are honored in full, ceilings only when asked) read at a glance.

**`src/aleph/vm/agent/vm/cache.py`** (line 798): _unknown_length_charge re-resolves the disk with shutil.disk_usage even though admit_download already called cache_budget_bytes (which stats the same root) one call earlier — and its OSError fallback is therefore only reachable if statfs fails between two adjacent calls. Passing the already-resolved total (or resolving the reserve where the budget is resolved) would drop the redundant stat and shrink the extra fallback path. Not blocking; the fallback is still correct and directly tested.

**`src/aleph/vm/agent/vm/cache.py`** (line 544): Minor: the over-budget log lines report the ceiling-excluded usage, so while an unmeasured download holds room the reported "is N bytes over" understates what the root actually holds. Cosmetic, but worth knowing when reading these logs during an incident.

**`tests/supervisor/test_cache_eviction.py`** (line 771): Good coverage of the ceiling side, but nothing pins the complementary half of the condition at cache.py:180 — that a *measured* reservation still counts its promised size, not just written bytes, when count_ceilings=False. Since `reservation.measured or count_ceilings` is the exact distinction this PR adds, a test showing an in-flight measured download (promised 8000, written 0) still drives evict_caches to evict for a later download would lock in the behavior.

PR #1212, opened 2026-09-09
test(tee): pin the TDX acceptance policy against real Intel collateral

odesenfans:
## Summary
The TDX "outdated" fixture never reaches an OutOfDate verdict: its PCK reports SGX component 7 at SVN 3 while every level of the matching TCB Info demands 5, so the walk ends "below every level" and the appraisal fails before a status is decided. Nothing therefore exercised `check_policy`'s "not accepted by policy" arm against a real Intel document, and no test asserted a non-empty advisory list off one either, so a typo in the `advisoryIDs` serde rename (a defaulted field) would hand every level an empty list, silently disable `denied_advisories`, and leave the suite green. This PR adds the two tests that close both gaps and corrects three pieces of doc text that describe the crate as SEV-SNP only.
Stacked on #1207.

## Changes
- `rust/crates/aleph-tee/src/tdx/tcb.rs`: two tests. `outdated_collateral_carries_the_published_advisories` parses the outdated collateral's signature-verified TCB Info and pins each level's advisories, including the nineteen the lowest rung carries. `out_of_date_level_is_refused_by_the_acceptance_policy` takes that same document and the fixture's real PCK platform, moves two SGX components so the walk lands on a published OutOfDate level (raising component 7, which blocks every level, and dropping component 4, which the top level rules out), then asserts the default policy refuses the result, a policy admitting OutOfDate takes it, and denying one of the level's advisories refuses it again. The existing `outdated_platform_is_below_every_level` comment now names the actual cause instead of saying "whatever the exact cause".
- `rust/crates/aleph-tee/tests/fixtures/tdx/README.md`: the outdated fixture's row claimed a platform "at an OutOfDate TCB level". It now says what the fixture actually exercises.
- `rust/crates/aleph-tee/Cargo.toml`: the crate description said "SEV-SNP attestation library" although the whole TDX quote and DCAP collateral path lives here.
- `rust/crates/aleph-tee/src/report_data.rs`: the module header spoke only of an SEV-SNP report carrying the 64 free bytes; a TDX quote carries them too, signed under the Intel-certified attestation key.
- `nix/golden-measurements.json`: re-seeded with `nix/check-golden-measurements.sh --update`, and it needs saying because this file pins the trusted launch chain. The attest agent binary links aleph-tee and is hashed into the measured initrd, and it embeds the panic locations of the sources it links, so even a pure doc reflow in `report_data.rs` moves all the pinned launch measurements. No behaviour changes with them. The file now carries five entries: the base rebase picked up `gpuMeasurement` from the confidential-GPU runtime flavor.

### Review follow-ups
- `rust/crates/aleph-tee/src/report_data.rs`: the paragraph edited here left one 95-column line in a file that otherwise wraps around 80. rustfmt does not reflow comments, so nothing in CI catches it. Re-wrapped, and the golden file re-seeded on top of it for the reason above.

## Test plan
- `rust/crates/aleph-tee/src/tdx/tcb.rs`: the two new tests pin the OutOfDate policy arm and the advisory parsing. Both were checked for teeth by temporarily adding `#[serde(rename = "advisoryIDz")]` to the three `advisory_i_ds` fields: both fail on empty advisory lists (as does the existing `report_15_takes_the_worse_of_the_two_svn_vectors`), and every other test stays green.
- Checks after the rebase onto the #1207 branch (all exit 0, read bare): in `rust/`, `cargo fmt --check`, `cargo clippy --locked --all-targets -- -D warnings`, `cargo test --locked -p aleph-tee` (133 passed, 2 ignored). `nix/check-golden-measurements.sh` matches after the re-seed.
- Earlier on this branch, the same three with `--manifest-path crates/aleph-attest-agent/Cargo.toml`: fmt and clippy exit 0, `cargo test` 49 passed. One unrelated flake on the first run, `gpu::tests::collect_runs_the_program_with_the_nonce_appended` failing with "Text file busy" on its temporary collector script; it passes on rerun and touches nothing in this PR.


foxpatch-aleph (approved):
This PR is correct and closes a real coverage gap. I independently verified every data-level claim against the vendored fixtures: the outdated collateral's TCB Info publishes exactly three levels with the pinned advisory lists (0/5/19), the fixture's PCK platform carries component 7 at SVN 3 (all levels demand 5) and component 4 at 4 (top level demands 4, OutOfDate rungs demand 3), so the two-component mutation lands the walk on the published OutOfDate level as the test comment describes. Both tee_tcb_svn vectors (module SVN 7 and 13) resolve to the TDX_01 identity's UpToDate rung with no advisories, so the exact-equality assertion on the level's five advisories holds, and now_outdated() sits inside the collateral's validity window. The corrected comments (tcb.rs, README) and the Cargo.toml/report_data.rs description fixes are all accurate. Both new tests demonstrably have teeth against an advisoryIDs rename typo or a broken check_policy arm. No issues found; one trivial cosmetic nit.

**`rust/crates/aleph-tee/src/report_data.rs`** (line 9): Trivial nit: this re-wrapped comment line is now 95 characters, noticeably above the ~80-column wrap used elsewhere in the file (the longest pre-existing line is 84). rustfmt doesn't reflow comments so CI is unaffected — feel free to ignore.

foxpatch-aleph (approved):
The substance of this PR is excellent: the two new tests close a real gap (the check_policy not-accepted arm and the advisoryIDs serde rename were previously untested against real Intel collateral), and I independently verified every fixture-dependent expectation: the outdated PCK reports cpusvn component 7 at 3 where all levels demand 5, the cpusvn[7]=5 / cpusvn[4]=3 mutations land the walk exactly on the level-1 OutOfDate rung (component 4 excludes only the UpToDate level, pcesvn 13 matches level 1 before level 2), both TDX vectors ([7,1,3] and [13,1,3]) name TDX_01 at SVN >= 6 so the module appraises UpToDate with no extra advisories, and the exact five-advisory list and the 19-advisory lowest rung match the collateral byte for byte. The corrected test comment and README row are now accurate, and the Cargo.toml and report_data.rs doc fixes are correct. The nix/golden-measurements.json re-seed is legitimately required (the report_data.rs doc reflow shifts embedded panic file/line locations in the measured agent binary) and was regenerated with the repo's own script, but the PR description never mentions it — worth adding given this file pins the trusted launch chain. Only nits remain.

**`rust/crates/aleph-tee/src/tdx/tcb.rs`** (line 818): Independently verified against the fixtures: the outdated PCK reports cpusvn [3,3,2,2,4,1,0,3,...] with pcesvn 13; all three levels demand component 7 at SVN 5; setting cpusvn[4] to 3 excludes only the UpToDate level; and pcesvn 13 means the walk lands on the level-1 OutOfDate rung rather than the lowest one. Both TDX SVN vectors ([7,1,3] and [13,1,3]) name TDX_01 at SVN >= 6, which appraises UpToDate with no advisories, so the converged status is OutOfDate and the exact five-advisory assertion holds. Solid test, and the explanatory comments make the fixture surgery auditable.

**`nix/golden-measurements.json`** (line 1): The PR description's Changes section omits this re-seed entirely; it is only explained in the commit message. For a file that pins the trusted launch chain (and per check-golden-measurements.sh, anything pinning the old measurements must migrate), please call it out in the PR body too: the report_data.rs doc reflow shifts embedded panic file/line locations in the measured agent binary, so all four pinned measurements move even though no behavior does. The re-seed itself is legitimate and was done with the repo's own script in the same PR.

**`rust/crates/aleph-tee/src/report_data.rs`** (line 9): Cosmetic: this line is 95 characters while the rest of the file wraps at ~82 — the tail of the paragraph was not re-wrapped after the edit. rustfmt does not check comment wrapping, so nothing in CI will catch it.

foxpatch-aleph (approved):
A high-quality test PR that I verified against the real fixtures rather than taking on trust. I independently decoded the outdated quote's PCK leaf (cpusvn [3,3,2,2,4,1,0,3,...], pcesvn 13) and the collateral's TCB Info, and every pinned value checks out: the three-level canonical order, the exact 5-advisory list on the middle rung, the 19 on the lowest (with all three sampled IDs present), and the mutated-platform walk landing exactly on the pcesvn-13 OutOfDate rung with both TDX vectors appraising the TDX_01 module as UpToDate (so status and advisory list are exactly as asserted). The corrected comments and the README/Cargo/report_data doc rewordings are all factually accurate, and both new tests have real teeth against the advisoryIDs serde-rename failure mode they target. No Rust/nix toolchain was available here (the background cargo run failed on the missing binary), so the suite and the golden-measurement re-seed rest on CI, which enforces both.

**`nix/golden-measurements.json`** (line 2): The re-seed cannot be validated in this review environment (no nix toolchain); CI must run nix/check-golden-measurements.sh before merge. The doc-only and test-only source edits shifting the measured chain is consistent with the repo's stated expectation that any source change moves the launch measurement, so this looks like the documented companion change rather than anything suspicious — but since the golden file is the tamper-evidence anchor for the measured agent, it is worth confirming the CI job actually ran and passed on this head.

**`rust/crates/aleph-tee/src/tdx/tcb.rs`** (line 815): Minor: the lowest rung is pinned by count (19) plus three sampled IDs. Pinning the full 19-element list — as the middle rung is pinned — would make the test robust against a partial parsing bug that drops or reorders individual advisories while keeping the count plausible. The current form is a reasonable trade-off against a 19-line literal, though; not blocking.

**`rust/crates/aleph-tee/src/tdx/tcb.rs`** (line 846): The 5-advisory list is duplicated verbatim in out_of_date_level_is_refused_by_the_acceptance_policy below. If either the fixture or Intel's collateral ever changes, both literals must be edited in lockstep — a small shared const (e.g. OUTDATED_RUNG_ADVISORIES) would tie the two assertions to one source of truth. Cosmetic only.

foxpatch-aleph (approved):
Well-executed test-coverage PR. I independently replayed the appraisal walk against the vendored fixture bytes in Python: the outdated quote's PCK platform is cpusvn [3,3,2,2,4,1,0,3,...] with pcesvn 13, and after the test's mutation (cpusvn[7]=5, cpusvn[4]=3) both of the quote's real TDX SVN vectors land on the published OutOfDate level carrying exactly the five asserted advisories, with the TDX module gate passing (MRSIGNERSEAM all-zeros matches the TDX_01 identity) and the module ladder contributing UpToDate with no advisories — so the merged advisory list is exactly what the test pins. The first test's expectations (3 levels; empty/5/19 advisories; canonical order) also match the fixture exactly. I confirmed the claimed gaps are real: no prior test hit the "not accepted by policy" arm, and no test asserted non-empty advisories off a real Intel document. All three doc corrections are accurate, and the golden-measurement re-seed is mechanistically coherent (report_data.rs reflow shifts embedded panic locations in the measured agent), though I could not rebuild the Nix measurements locally and rely on the golden-measurements workflow to verify the values. Only minor non-blocking nits below.

**`rust/crates/aleph-tee/src/tdx/tcb.rs`** (line 846): Minor: the five-entry advisory list is duplicated verbatim between the two new tests (here and at line 801-810). If the fixture is ever re-vendored, both need editing in lockstep — a shared `const` would keep them in sync. Purely optional; duplicating independent pins is also defensible.

**`rust/crates/aleph-tee/src/tdx/tcb.rs`** (line 826): Observation: since this drives `appraise_platform_tcb` + `check_policy` directly, the "not accepted by policy" arm is still not exercised end-to-end through the public `evaluate_tcb` with a real document — if `evaluate_tcb` ever stopped calling `check_policy`, `genuine_v4_platform_is_up_to_date` (which only exercises the Ok path) would stay green. That is inherent to the fixture (a doctored cpusvn cannot pass chain verification), so unit level is the right place; just worth knowing the arm's end-to-end wiring rests on `evaluate_tcb` being a thin composition.

**`nix/golden-measurements.json`** (line 1): Could not rebuild these locally (no nix in the review environment). The re-seed is consistent with the report_data.rs line shift changing embedded panic locations in the measured agent, and CI (golden-measurements.yml) verifies it; flagging only that anything downstream pinning the old measurements must migrate, which the check script itself calls out.

PR #1213, opened 2026-09-09
fix(agent): drop a retired VM from the live set the retire forgets it

odesenfans:
## Summary
The download admission hook evicts only when it can trust the live set a storage pass published: a hash in that set with no registry record means a running VM whose message nobody can read, so admission fails closed and evicts nothing. `retire_vm` forgot the registry record but left the retired hash in the published set, and under the default `VOLUME_RETENTION=reap` no pass follows a retire. Every retired VM therefore read as a live VM without a record until the next periodic pass (up to `VOLUME_RECONCILE_INTERVAL`, an hour): in that window every download logged an ERROR and a download that did not fit the budget was refused with a 503 while the cache was full of evictable entries.

The fix keeps one owner of the set. `cache.py` gains `forget_live`, and `retire_vm` calls it where it calls `registry.forget`, so both halves move together for every reason that drops the record. An unset set stays unset: "no pass has run yet" is still not "nothing is live", and only a pass may turn one into the other.

Stacked on #1211.

## Changes
- `src/aleph/vm/agent/vm/cache.py`: new `forget_live(namespace)` next to `record_live_snapshot`, the only other writer of `_live_snapshot`. It removes one hash and returns without touching an unset snapshot.
- `src/aleph/vm/agent/vm/retire.py`: `retire_vm` calls `forget_live` immediately after `registry.forget`, so it runs for GONE, ERASE and FAILED_CREATE and not for RECREATE (which keeps the record and brings the VM straight back).
- `docs/architecture/storage.md`: the admission section now says a pass is not the only writer of the published live set, and why a retire has to drop its hash.

## Test plan
- `tests/supervisor/test_retire.py`: four new tests. A record-dropping retire (parametrized over the three reasons) removes the hash and leaves the rest of the set alone; RECREATE leaves the set untouched; a retire before the first pass leaves the set unset; and the end to end pin, a reaping GONE retire followed by a download that needs room, which now evicts the stale entry with no "no registry record" ERROR and no `InsufficientResourcesError`. The three parametrizations and the end to end test failed on the base commit (the last one with the 503).
- An autouse fixture in that file resets the two module-level pieces of state the file now touches (the live snapshot and the reserved downloads), matching the convention in `test_cache_eviction.py`.
- Checks: `pytest tests/supervisor/test_retire.py tests/supervisor/test_cache_eviction.py tests/supervisor/test_reconciler.py` (146 passed, 1 skipped) and the whole `tests/supervisor` suite (1196 passed, 1 pre-existing journald failure in `test_log.py`, identical to the base). `ruff format --diff`, `isort --check-only --profile black` and `mypy` all clean on the three Python files. No Rust touched.

### DIFF
--- a/docs/architecture/storage.md
+++ b/docs/architecture/storage.md
@@ -534,6 +534,14 @@ message means the referenced set is incomplete. Before the first pass there is
 no live set, so admission evicts nothing and simply admits or refuses on the
 budget as it stands.
 
+A pass is not the only thing that changes the published set: `retire_vm` drops
+the VM it retires from it, right where it forgets the registry record. The two
+halves have to move together, and under `reap` no pass follows a retire, so a
+hash left behind would read as a live VM with no record until the next periodic
+pass (an hour by default) and stop admission evicting for that whole window. A
+retire never turns an unset live set into an empty one: only a pass may say
+that nothing is live.
+
 ### Settings
 
 | Setting | Default | What it does |
--- a/src/aleph/vm/agent/vm/cache.py
+++ b/src/aleph/vm/agent/vm/cache.py
@@ -96,6 +96,28 @@ def record_live_snapshot(live: Collection[str]) -> None:
     _live_snapshot = frozenset(str(namespace) for namespace in live)
 
 
+def forget_live(namespace: str) -> None:
+    """Drop a VM the agent has just retired from the live set.
+
+    A retire quiesces the VM in the supervisor and forgets its registry
+    record, so the two halves the pass reconciles both stop naming it. Until
+    the next pass republishes the set, a hash left here reads to admission as
+    a live VM with no record, which is precisely the doubt that stops it
+    evicting anything: every download then logs an error and one that does
+    not fit is refused with the cache full of evictable entries. Retiring
+    under the retention that keeps volumes runs a pass straight after, but
+    reaping does not, and the periodic pass can be an hour away.
+
+    An unset snapshot stays unset: "no pass has run yet" is a different
+    answer from "nothing is live", and only the pass may turn one into the
+    other.
+    """
+    global _live_snapshot  # noqa: PLW0603
+    if _live_snapshot is None:
+        return
+    _live_snapshot = _live_snapshot - {str(namespace)}
+
+
 def live_snapshot() -> frozenset[str] | None:
     return _live_snapshot
 
--- a/src/aleph/vm/agent/vm/retire.py
+++ b/src/aleph/vm/agent/vm/retire.py
@@ -31,6 +31,7 @@ from aleph_message.models import ItemHash
 
 from aleph.vm.agent.metrics import delete_records_for_vm
 from aleph.vm.agent.vm.backup import purge_vm_backups
+from aleph.vm.agent.vm.cache import forget_live
 from aleph.vm.agent.vm.purge import (
     _checked_namespace,
     purge_vm_side_dirs,
@@ -192,6 +193,12 @@ async def retire_vm(
     item_hash = vm_hash if isinstance(vm_hash, ItemHash) else ItemHash(str(vm_hash))
     record = registry.get(item_hash)
     registry.forget(item_hash)
+    # The cache's live set is the other half of what admission checks a record
+    # against, and it is only refreshed by a storage pass: leaving a retired
+    # hash in it makes every download until the next pass look like it is
+    # racing a live VM the agent has no message for, so nothing is evicted and
+    # a download that needs room is refused.
+    forget_live(str(vm_hash))
     await delete_records_for_vm(str(vm_hash))
     # Before the storage pass: a volume file held by a live dm target cannot
     # be unlinked usefully, and the marker written for a kept volume would
--- a/tests/supervisor/test_retire.py
+++ b/tests/supervisor/test_retire.py
@@ -3,13 +3,18 @@
 from __future__ import annotations
 
 import json
+import os
+import time
 from unittest.mock import AsyncMock, MagicMock
 
 import pytest
 from aleph_message.models import InstanceContent, ItemHash
-from reclaim_fixtures import VM_HASH, pools, volume  # noqa: F401
+from reclaim_fixtures import OTHER_HASH, VM_HASH, pools, volume  # noqa: F401
 
+import aleph.vm.agent.vm.cache as cache_module
 import aleph.vm.agent.vm.retire as retire_module
+import aleph.vm.storage as storage_module
+from aleph.vm.agent.vm.cache import admit_download
 from aleph.vm.agent.vm.reclaimable import MARKER_NAME, read_marker
 from aleph.vm.agent.vm.retire import RetireReason, retire_vm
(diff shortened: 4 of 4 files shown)


foxpatch-aleph (approved):
This PR correctly fixes a real bug: after retire_vm forgot a registry record, the hash stayed in the published live snapshot, so admission read it as a live VM with no record and refused to evict for up to a full reconcile interval (with a 503 when a download needed room). The fix is placed exactly where the record is dropped, after the RECREATE early-return and with no awaits between registry.forget and forget_live, so the two halves stay atomic on the event loop; forget_live is idempotent and correctly preserves an unset (None) snapshot as distinct from empty, and retire_vm is the only production path that drops records. I verified all new and surrounding tests pass (31 in test_retire.py, 116 in test_cache_eviction/test_reconciler), and that the parametrized and end-to-end tests fail on the base behavior once the fix is reverted, so they genuinely pin the bug. The autouse state fixture mirrors the existing convention in test_cache_eviction.py, and the documentation update is accurate. Only minor, non-blocking observations remain.

**`src/aleph/vm/agent/vm/retire.py`** (line 201): Residual race, not blocking: a retire that completes while a pass is inside `await supervisor_hashes(supervisor)` (reconciler.py:242) can still leak its hash into the published set, because `_live_set` snapshots `registry_live` before that await and `record_live_snapshot(live)` (reconciler.py:248) publishes the union afterwards. The window is one `list_vms` RPC and it self-heals at the next pass, so this is strictly narrower than the pre-PR behavior — but worth knowing, or guarding, if the ERROR ever resurfaces intermittently.

**`tests/supervisor/test_retire.py`** (line 428): Minor: this asserts on a substring of the ERROR message ("no registry record") rather than the logger or message format, so a future rewording of `_may_evict_for_admission`'s text would silently turn this into a no-op. The `not old.exists()` assertion carries the regression check either way, so this is fine to leave — just noting the fragility.

foxpatch-aleph (approved):
The fix is correct, minimal, and well-targeted: forgetting the registry record and the live-set hash in the same place keeps the two halves admission checks against in sync, and I confirmed registry.forget is the only record drop site so nothing is missed. RECREATE correctly returns before forget_live, the unset-snapshot invariant is preserved, and both mutations run on the event loop so there are no torn reads. Tests are thorough (including the end-to-end 503 reproduction) and all pass locally along with the cache eviction and reconciler suites; docs are accurate. One non-blocking observation: a reconcile pass that completes its registry read before a retire but publishes after it can still republish the retired hash into the snapshot (the window spans the awaited list_vms in _live_set), reintroducing the stale-hash state until the next pass; the consequence is fail-closed and bounded, and closing it would require locking that seems disproportionate to the window.

**`src/aleph/vm/agent/vm/cache.py`** (line 99): Non-blocking: a narrow residual race remains. _live_set (reconciler.py:216) reads the registry and then awaits supervisor.list_vms() before record_live_snapshot, so a retire that completes inside that await can have its hash republished into the snapshot after forget_live already removed it — the same stale-hash-without-record state, until the next pass. The consequence is fail-closed (admission logs an ERROR and evicts nothing) so it is safe, and closing the window would need locking or re-reading the registry after the await; probably not worth it, but worth keeping in mind if the ERROR ever reappears in production logs.

foxpatch-aleph (approved):
Correct, minimal fix for a real bug: retiring a VM left its hash in the admission live set, so every download for up to a reconcile interval logged an ERROR and a download needing room was refused with a 503 while the cache was full of evictable entries. Hooking forget_live beside registry.forget keeps the two halves of the invariant together, and registry.forget has no other call site, so no record-dropping path is missed. RECREATE correctly returns before the forget, string normalization matches what the pass publishes (str(ItemHash) on both sides), and the unset-stays-unset semantics preserve the fail-closed default. Tests are real pins: I verified the end-to-end test fails with forget_live neutralized, the full retire suite passes, and mypy/ruff-format are clean. One residual, non-blocking race remains: a reconcile pass whose `await supervisor.list_vms()` straddles a concurrent retire can republish a stale set containing the retired hash — a one-await window that fails closed and self-heals at the next pass, so it is strictly better than the pre-fix behavior.

**`src/aleph/vm/agent/vm/retire.py`** (line 201): Residual, non-blocking race worth a line in the comment: a reconcile pass that has already read `live_hashes(registry)` and is awaiting `supervisor.list_vms()` can complete and call `record_live_snapshot` after this `forget_live`, republishing a set that still contains the just-retired hash — undoing this removal until the next pass. The window is a single await rather than every retire, and the failure mode is the pre-existing fail-closed one (admission logs and evicts nothing), so this is fine, but a note that a concurrent pass may still resurrect the hash would set expectations for whoever reads the ERROR next.

**`src/aleph/vm/agent/vm/retire.py`** (line 201): Tiny nit: `forget_live(str(vm_hash))` could use the already-normalized `str(item_hash)` from three lines up for symmetry with the `registry.forget(item_hash)` call above. Behaviorally identical for both input types (ItemHash is a str subclass), so feel free to ignore.

**`src/aleph/vm/agent/vm/cache.py`** (line 99): Good docstring — it captures the two easy-to-miss invariants (unset stays unset; the pass is the only thing that may publish an empty set). One scenario it could also name: a VM re-created with the same hash right after a GONE retire is safe to have absent from the snapshot, because eviction protects live entries via `live_refs(registry)`, not via this set — the snapshot is only the trust check. A future reader may otherwise worry that removing the hash here exposes a just-recreated VM's cache entries.

foxpatch-aleph (approved):
A correct, minimal fix for a real availability bug: retire_vm dropped the registry record but left the hash in the admission live set, so under VOLUME_RETENTION=reap every download for up to a full reconcile interval was refused with a 503 while evictable entries sat in the cache. forget_live is placed beside the only registry.forget call site in src/, keeps the fail-closed None semantics, and correctly skips RECREATE; the change can only relax the coarse admission gate, never the per-entry referenced set, so no live VM loses protection. New tests pass (30 passed, 1 skipped in test_retire.py, plus the eviction and reconciler suites), and the end-to-end failure mode matches code reading of the base. Docs are accurate. Only a minor fail-closed race remains, noted below.

**`src/aleph/vm/agent/vm/cache.py`** (line 99): Non-blocking: there is a narrow residual window in the other direction. `_live_set` reads `live_hashes(registry)` before `await supervisor_hashes(supervisor)` and only publishes at `record_live_snapshot(live)` afterwards (reconciler.py:235-248), so a retire that commits during that await is reintroduced by the pass and re-triggers the pre-fix symptom until the next periodic pass. It fails closed (no eviction, no unsafe unlink) and self-heals within one interval, so it is fine as-is — but a one-line note in the docstring (or storage.md) would spare the next reader the search.

**`tests/supervisor/test_retire.py`** (line 408): Nice end-to-end pin. One small suggestion: the test asserts the eviction happened and the error was absent, but never asserts the download was actually admitted; since admit_download raises InsufficientResourcesError on refusal the test would still fail via the raise, so this is purely a readability nit — consider `admit_download(...)` on its own line before the `with caplog` block if you ever add a second download to the test.

PR #1214, opened 2026-09-09
fix(agent): back off when a guest crash-loops after it reached RUNNING

odesenfans:
## Summary
The allocation reconciler's backoff only covered a start that raised. A start that worked called `_forget`, erasing the record, so when the guest later died the FAILED event woke the loop, `_start_missing` found no record and read the retry as due, and `start_persistent_vm`'s FAILED branch retired the VM as RECREATE and rebuilt it from scratch. A guest that panics seconds after boot was therefore rebuilt at boot speed, disks and all, for as long as the plan listed it (v1 rebuilt the same guest once per scheduler push). This counts a post-RUNNING death as an attempt on the same exponential ladder, so the first rebuild is still immediate and each death after it waits longer, capped at `ALLOCATION_RETRY_MAX_INTERVAL`.

Stacked on #1209.

## Changes
- `allocation/reconciler.py`: `_start_one` takes the `VmInfo` the supervisor holds for the VM when it has one, which is what makes the start a rebuild rather than a first create. On success it now records that rebuild as an attempt instead of forgetting, on the same ladder a failing create climbs. The backoff bookkeeping moved out of `_record_failure` into `_note_attempt`, shared by both callers, so exactly one attempt is charged per pass either way.
- `allocation/reconciler.py`: `_forget_settled` drops the record of a VM that has been up longer than `ALLOCATION_RETRY_MAX_INTERVAL`, the longest wait the backoff can impose. The record has to outlive the successful start that follows a rebuild, or nothing would gate the next one, but it must not be immortal: a VM that crashed once a month ago deserves its rebuild at once, not at the cap.
- `allocation/reconciler.py`: a planned VM that is down and not yet due is reported `failed`, so the v2 executions list shows the attempts and `next_retry_at` instead of a dead VM the agent looks to have no opinion about. No change was needed in `_allocation_block`, which already renders that pair.
- `allocation/plan.py`: `FailureRecord`'s docstring says what a record on a running VM now means (it has died recently, not that it is down).

## Test plan
- `tests/supervisor/test_allocation_reconciler.py`: six tests, four of which failed before the fix. A guest that boots and dies is rebuilt at once, counted, and its next rebuild is gated by the base interval then by double; the ladder doubles and is capped at the max; a VM that stays up accumulates no attempts; a rebuilt VM that stays up loses its record once it has been up for the max interval (and keeps it a second before); a VM waiting out its rebuild reports `failed` with the attempts and the retry time; a rebuild the plan dropped mid-flight is not recorded, the rule the failing create already followed.
- pytest on the four files that import the touched modules plus `test_allocation_teardown.py` and `test_agent_capacity.py`: `197 passed`. The wider related set (15 files): `307 passed`.
- `ruff format --diff`, `isort --check-only --profile black`, `mypy` on the three changed files: exit 0, `Success: no issues found in 3 source files`.

### DIFF
--- a/src/aleph/vm/agent/allocation/plan.py
+++ b/src/aleph/vm/agent/allocation/plan.py
@@ -61,7 +61,12 @@ class AllocationState(str, Enum):
 
 @dataclass
 class FailureRecord:
-    """Why a planned VM is not running, and when we will try again.
+    """What it is taking to keep a planned VM running, and when we try again.
+
+    Two things are counted as one attempt: a start that raised, and a rebuild
+    of a VM the supervisor held dead, which is what puts a crash-looping guest
+    on the same backoff. So a record can outlive a successful start, and one
+    on a running VM means it has died recently, not that it is down now.
 
     Kept in the reconciler and NOT in AgentVmRegistry on purpose: the registry
     is what CapacityManager sums committed resources over, and a VM that failed
--- a/src/aleph/vm/agent/allocation/reconciler.py
+++ b/src/aleph/vm/agent/allocation/reconciler.py
@@ -205,7 +205,18 @@ class AllocationReconciler:
             if info.status in LIVE_STATUSES or info.awaiting_confidential_init
         }
         now = self._now()
-        todo = [vm_hash for vm_hash in plan.entries if vm_hash not in live and self._retry_due(vm_hash, now)]
+        self._forget_settled(live, now)
+        todo: list[ItemHash] = []
+        for vm_hash in plan.entries:
+            if vm_hash in live:
+                continue
+            if not self._retry_due(vm_hash, now):
+                # Down and waiting out its backoff. Saying so is what lets the
+                # executions list report the wait and the time it ends, rather
+                # than a dead VM the agent appears to have no opinion about.
+                self._states[vm_hash] = AllocationState.FAILED
+                continue
+            todo.append(vm_hash)
         if not todo:
             return
 
@@ -213,7 +224,9 @@ class AllocationReconciler:
 
         async def start(vm_hash: ItemHash) -> None:
             async with semaphore:
-                await self._start_one(vm_hash)
+                # A VM the supervisor lists that is not live is one it holds
+                # dead: the loop started it before, so this start is a rebuild.
+                await self._start_one(vm_hash, known.get(vm_hash))
 
         await asyncio.gather(*(start(vm_hash) for vm_hash in todo), return_exceptions=True)
 
@@ -221,7 +234,24 @@ class AllocationReconciler:
         failure = self._failures.get(vm_hash)
         return failure is None or failure.next_retry_at <= now
 
-    async def _start_one(self, vm_hash: ItemHash) -> None:
+    def _forget_settled(self, live: set[ItemHash], now: datetime) -> None:
+        """Drop the record of a VM that has been up long enough to call healthy.
+
+        A rebuild after a death counts as an attempt, so the record has to
+        outlive the successful start that follows it, or nothing would gate
+        the next rebuild. It cannot be immortal either: a VM that crashed once
+        a month ago deserves its rebuild at once, not at the capped wait. The
+        longest wait the backoff can impose is the threshold, past which the
+        next death is a new problem rather than the tail of the old one.
+        """
+        settled = timedelta(seconds=settings.ALLOCATION_RETRY_MAX_INTERVAL)
+        for vm_hash, failure in list(self._failures.items()):
+            if vm_hash in live and now - failure.last_failed_at >= settled:
+                self._forget(vm_hash)
+
+    async def _start_one(self, vm_hash: ItemHash, down: VmInfo | None = None) -> None:
+        """Start a planned VM. `down` is what the supervisor holds for it when
+        it already has one, which makes this start a rebuild of a dead VM."""
         self._states[vm_hash] = AllocationState.DOWNLOADING
         try:
             await start_persistent_vm(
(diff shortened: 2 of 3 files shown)


foxpatch-aleph (approved):
This PR fixes a genuine and well-diagnosed gap: the backoff only covered starts that raised, so a guest that crash-looped after reaching RUNNING was rebuilt from scratch at boot speed forever. Charging a successful rebuild as an attempt on the same exponential ladder, bounding the record's lifetime via _forget_settled, and reporting FAILED while a rebuild waits are all correct and mutually consistent; the mid-flight plan-drop guard mirrors the existing failing-create rule exactly. Tests are thorough — six new tests, four of which fail without the fix — and the full related set (197 tests), mypy, and formatting all pass, matching the author's claims. The ruff check findings present on the directory are pre-existing on untouched code. Only minor, non-blocking semantic notes remain.

**`src/aleph/vm/agent/allocation/reconciler.py`** (line 290): Minor semantic note: `down` can also be STOPPED (which `start_persistent_vm` resumes in place, a cheap stop/start rather than a delete+recreate) or STOPPING, and both are charged on the same ladder as a FAILED crash rebuild. That is almost certainly intended — a guest that halt-loops needs backoff as much as one that panics — but the `_start_one` docstring's "rebuild of a dead VM" slightly overstates the STOPPED case. Consider mentioning that any non-live status counts.

**`src/aleph/vm/agent/allocation/reconciler.py`** (line 246): `last_failed_at` is the moment the rebuild's `_note_attempt` ran, i.e. after `start_persistent_vm`'s readiness barrier, so "up for MAX_INTERVAL" is measured from when the VM reached RUNNING. That is a sound approximation and avoids needing a separate "started_at" field; no change requested, just confirming the reasoning holds if `start_persistent_vm` ever returns before RUNNING.

foxpatch-aleph (approved):
Correct fix for a real gap: the old code's _forget on every successful start meant a guest that died after reaching RUNNING was rebuilt at boot speed forever. Charging a post-RUNNING death on the same exponential ladder as a failing create is sound — exactly one attempt is charged per pass either way, the rebuild path re-reads self._desired after the await so a plan dropped mid-flight is not resurrected, and _forget_settled's threshold equals the delay cap, so a record is only ever dropped once its next_retry_at is already past (no retry is ever skipped). The new FAILED marking in _start_missing only fires when a record exists, so the executions list always has attempts and next_retry_at to render. down can only carry STOPPING/STOPPED/FAILED, all of which start_persistent_vm handles. Tests are thorough (six new, four demonstrating the old bug), 229 pass across the related files, ruff and mypy are clean. Two non-blocking notes below.

**`tests/supervisor/test_allocation_reconciler.py`** (line 333): Non-blocking: the _start_one docstring makes a point of charging the STOPPED (cheap in-place resume) and STOPPING (waited out, then recreated) paths on the same ladder, but only the FAILED path is tested. A one-line `status=VmStatus.STOPPED` case through _boots_then would pin that decision, since a future refactor could plausibly argue a resume should be exempt.

**`src/aleph/vm/agent/allocation/reconciler.py`** (line 314): Observation, not a change request: while the rebuilt VM is up, the executions list shows nothing about the lingering record (state popped, and _allocation_block returns None on state None), so attempts are only visible while the VM is down. That matches the "supervisor knows the VM again" reading in the docstring; just flagging that an operator watching the list won't see the crash history of a currently-running VM.

foxpatch-aleph (approved):
Sound fix for a real crash-loop hammer: a successful rebuild is now charged on the same backoff ladder as a failing create, with exactly one attempt per pass per VM, a symmetric mid-flight plan-drop guard, and a well-reasoned record-expiry policy in _forget_settled. The new FAILED state for down-and-waiting VMs composes cleanly with the existing _allocation_block rendering, which I confirmed needs no change. Tests are targeted (four of six fail without the fix per the author), and I verified the full reconciler, executions-list, verdict, and allocations suites pass alongside ruff, isort, and mypy. The remaining observations are nits.

**`src/aleph/vm/agent/allocation/reconciler.py`** (line 310): Nit: this comment is slightly misleading. Popping the state makes `_allocation_block` return `None` (it early-returns on `state is None`), so the attempt count is NOT reported while the VM runs either — nothing of the record is visible until the VM dies again. Consider rewording to say the agent has nothing at all to report once the supervisor knows the VM again.

**`src/aleph/vm/agent/allocation/reconciler.py`** (line 252): Worth a note in the docstring: the attempt is charged from the pass's `known` snapshot, but `start_persistent_vm` re-fetches the VM itself. If the VM transitioned FAILED→RUNNING between the two (operator restart, concurrent on-demand start), the no-op "already running" branch is still charged an attempt, inflating the next backoff by one notch. Self-correcting on the next pass and inherent to the level-triggered design, so not blocking.

**`tests/supervisor/test_allocation_reconciler.py`** (line 333): The `_start_one` docstring claims a cleanly STOPPED (or STOPPING) VM resumed by the loop is charged like a FAILED rebuild, but the charge tests only exercise FAILED. Parametrizing `_boots_then` over `VmStatus.STOPPED` (and asserting the `vm_stopped` code) would pin that deliberate behavior.

foxpatch-aleph (approved):
Correct and well-targeted fix: charging a successful rebuild on the same exponential ladder as a failed create closes the crash-loop hole, and the record's lifetime is bounded sensibly (only dropped for live VMs, so a crash-loop keeps resetting the clock while a one-off crash ages out at the cap). The mid-flight plan-drop guard correctly mirrors the failing-create rule, `down` can never carry a live status, and the executions list already renders the attempts/retry pair. Six new tests cover the ladder, the cap, the not-due reporting, the settled-record boundary, and the dropped-plan case; all 37 tests in the reconciler suite and the wider related set pass locally, as do ruff/isort/mypy. Two non-blocking nits below.

**`src/aleph/vm/agent/allocation/reconciler.py`** (line 217): Unlike `_start_one`, which re-reads `self._desired` before every state mutation, this write acts on the pass-start snapshot of `plan.entries`. A push that drops this VM while the pass is parked in `list_vms()` or a teardown has already pruned the state in `submit()`, and this line can put a `FAILED` entry back for a VM nothing will retry — the exact resurrection `_start_one` guards against, and the class of race `_teardown_dropped` documents and re-reads for. The window is narrow and the next push clears it, so this is cosmetic, but a `self._desired` check (or iterating the fresh plan) would make it consistent.

**`tests/supervisor/test_allocation_reconciler.py`** (line 333): `_boots_then` is only exercised with FAILED and RUNNING. The `VmStatus.STOPPED` resume path is called out in `_start_one`'s docstring as charged like a panic ("a guest that halt loops needs the wait as much as one that panics"), but no test pins a rebuild of a STOPPED VM to the ladder. A one-line parametrize over the status would cover it.

PR #1215, opened 2026-09-09
fix(cli): keep aleph-vm storage from purging behind a running agent

odesenfans:
## Summary
A real `aleph-vm storage reconcile` ran the reconciler's walk from a process
that cannot see the one thing the agent holds: the creates it has in flight
(`reconciler.creating()` is in-process state). A migration import that
outlives `VOLUME_CREATE_GUARD` before its DB record exists therefore read as
an orphan, and the pass removed the directory the create was building on,
plus its devices, with no lock against the agent's own pass. This PR turns
that documented caveat into a rule keyed on the agent process, and fixes
three smaller safety gaps in the same file: the CLI live set mirrored only
half of the daemon's `_startup_refusal` (an empty registry plus a reachable
supervisor purged on the supervisor's list alone), every failed dial became a
silent `None` that ended in the advice to purge on the registry alone (on a
permission error, on the 3 s deadline, on a malformed reply, and even on an
explicit `--dry-run`), and `reclaim` checked the `.reclaimable` marker before
the supervisor dial and purged afterwards without looking again.

The two processes are independent and the distinction matters: the **agent**
is this Python service, which owns the reconciler and the creates; the
**supervisor** is the Rust daemon that runs the VMs and answers `list_vms`.
The node an operator needs this command for is the awkward one, VMs running
under a healthy supervisor while the agent is down.

Stacked on #1210.

## Changes
- `storage_cli.py`, three-way purge rule in `_reconcile` and `_reclaim`:
  - the agent is running, or cannot be ruled out as running: a real pass and
    `reclaim` are refused, reason on stderr, exit 3. `--dry-run` still runs
    and notes that its listing may name a directory the agent is creating.
  - the agent is down and the supervisor answers: the full pass, including
    `_teardown_orphan_devices` before the walk (nothing is creating, and the
    live set is registry union `list_vms`).
  - the supervisor does not answer: the existing `--trust-registry` gate,
    unchanged, and no device teardown on that path.
- `storage_cli.py`, agent probe: `_probe_agent` connects to the agent's own
  HTTP bind (`SUPERVISOR_HOST`/`SUPERVISOR_PORT`, that setting's legacy name)
  with a 2 s timeout. It needs no privileges and no unit name. Only a refused
  connection returns `STOPPED`; anything else is `UNKNOWN`, which counts as
  running (`AgentProbe.may_be_running`). Being wrong that way costs one
  refused command; being wrong the other way deletes the disks of a VM
  mid-create. A wildcard bind is probed on both loopbacks and called stopped
  only when both refuse: asyncio's `create_server` sets `IPV6_V6ONLY`, so an
  agent bound to `::` accepts on `::1` and refuses on `127.0.0.1`. An address
  family the kernel cannot reach counts with the refusals, or an IPv4-only
  node would answer "cannot tell" for ever.
- `storage_cli.py`, streams: `run()` and the verbs take `out` and `err`.
  What the command found or achieved goes to stdout, every warning, refusal
  and diagnostic to stderr, `reclaim` included, so a refusal leaves stdout
  empty rather than mixing a reason into a report that does not exist.
- `storage_cli.py`, ordering: `_reconcile` probes the agent before dialling
  the supervisor, so a pass the agent refuses does not first sit through the
  3 s deadline for an answer it discards.
- `storage_cli.py`, live set: `_cli_live_set` returns a `LiveSet` carrying
  the daemon's own `_startup_refusal` verdict, imported rather than restated,
  so the registry-empty case ("the agent DB was lost") refuses in the CLI
  exactly as it does in the daemon. `_reclaim_refusal` asks it too.
- `storage_cli.py`, diagnostics: `_ask_supervisor` returns a
  `SupervisorAnswer` (`ANSWERED` / `DOWN` / `UNKNOWN`, plus the exception
  class and message) instead of `None`. `DOWN` is only claimed when it can be
  backed up: *its own* socket is missing (`os.stat`, since `Path.exists()`
  turns a stat this user may not make into a plain `False`) or refuses a
  connection. A refused connection in the exception chain counts too, but a
  missing file there does not (the client opens more than its socket): that
  case falls through to the stat. Fixed in the Foxpatch review round: the
  chain was walked one cause at a time and returned as soon as it found a
  `PermissionError`, so a `ConnectionRefusedError` sitting deeper in a chain
  whose head is a `PermissionError` went unnoticed; the whole chain is now
  checked for a refusal before it is checked for a permission error, so a
  refusal anywhere in the chain is proof of `DOWN` regardless of what comes
  before it.
  The `--trust-registry` advice is printed only then; a permission error, a
  deadline or a malformed reply say so and stop, and an explicit `--dry-run`
  is told its preview is registry-only.
- `storage_cli.py`, reclaim race: the marker is read once more immediately
  before `purge_vm_storage`, since a re-create adopts its retained
  directories by clearing their markers and one that started during the dial
  is in no answer this process holds.
- `storage_cli.py`, exit codes: refusals carry theirs (`Refusal`). "Not this
  hash" stays 1; a node whose agent is live or whose agent database was lost
  is 3, since nothing about the hash is wrong and a wrapper has to tell the
  two apart.
- `docs/architecture/storage.md`: the CLI section states the three states,
  both fail-closed probes, and the caveat that `--trust-registry` on a
  verified-down supervisor can still purge under a live VM (a supervisor
  process that dies leaves its QEMU processes running).
- `storage_cli.py`, wording (Foxpatch): the reclaim refusal used to point at
  `storage list --reclaimable` to preview a purge, but that verb does not run
  a reconcile pass; it now points at `storage reconcile --dry-run`, which
  does. The module docstring also now notes that aiohttp runs the agent's
  `on_startup` hooks before its listener binds, so a starting agent can
  briefly read as stopped; the create guard covers most of that window.

## Test plan
- `tests/supervisor/test_storage_cli.py`: the three states (a live agent and
  an unprobeable one refuse `reconcile` and `reclaim` at exit 3 while
  `--dry-run` still lists and warns; an agent that is down with the
  supervisor answering runs the full pass, tears down the orphan namespace's
  devices and evicts cache parents; an unanswered supervisor keeps the
  `--trust-registry` gate and tears nothing down). Plus: the probe itself on
  a listening socket, a refused port and a probe that raises; the
  registry-empty refusal on both verbs; `PermissionError` and `TimeoutError`
  dials named on stderr with no `--trust-registry` advice while
  `ConnectionRefusedError` gets it; an explicit `--dry-run` on a down daemon
  told its preview is registry-only; a marker cleared during the dial making
  `reclaim` refuse; and `_socket_reach` on a listening socket, a missing one
  and an unset path.
- Review follow-ups pinned: an agent bound to `::1` alone is found through a
  `::` wildcard; a wildcard is stopped only when both loopbacks refuse; a
  `FileNotFoundError` from elsewhere in the dial does not advise
  `--trust-registry` while the socket is answering; a refused pass never
  dials the supervisor; and every reclaim refusal is asserted on stderr with
  stdout empty.
- Foxpatch review round: a `ConnectionRefusedError` deeper in the chain than
  a `PermissionError` still classifies the daemon as `DOWN`.
- `pytest tests/supervisor/test_storage_cli.py` 58 passed; with the
  reconciler and reclaimable files, 145 passed.
- `ruff format --diff`, `isort --check-only --profile black`, and `mypy
  src/aleph/vm/ tests/` all exit 0.

### DIFF
--- a/docs/architecture/storage.md
+++ b/docs/architecture/storage.md
@@ -625,20 +625,39 @@ Both "is it running" questions are answered fail-closed, and neither claims
 more than it can back up. The **agent** is called stopped only when nothing
 accepts a connection on its bind address; that is a loopback connect with a
 2 s timeout, it needs no privileges and no unit name, and anything listening
-there counts as the agent. A probe that fails any other way (a name that
-does not resolve, a socket this user may not open, the timeout) leaves the
-agent possibly running, and the refusal stands: being wrong that way costs
-an operator one refused command, being wrong the other way deletes the disks
-of a VM mid-create. The **supervisor** is called *down* only when its socket
-is missing (`os.stat`, not `Path.exists()`, which turns a stat this user may
-not make into a plain False) or refuses a connection. Every other failed
-dial (a socket this user may not open, the 3 s deadline, a reply that does
-not parse) leaves the daemon's state unknown. The difference is what the
+there counts as the agent. A wildcard bind (`0.0.0.0`, `::`, empty) is
+probed on *both* loopbacks and called stopped only when both refuse, because
+asyncio's `create_server` sets `IPV6_V6ONLY` on an `AF_INET6` socket: an
+agent bound to `::` accepts on `::1` and refuses on `127.0.0.1`, so a probe
+that asked only the IPv4 loopback would call a running agent stopped and
+purge behind it. An address family the kernel cannot reach at all
+(`EAFNOSUPPORT`, `ENETUNREACH` and friends) counts with the refusals, since
+nothing can be serving on a loopback that is not there. A probe that fails
+any other way (a name that does not resolve, a socket this user may not
+open, the timeout) leaves the agent possibly running, and the refusal
+stands: being wrong that way costs an operator one refused command, being
+wrong the other way deletes the disks of a VM mid-create. The **supervisor**
+is called *down* only when *its own socket* is missing (`os.stat`, not
+`Path.exists()`, which turns a stat this user may not make into a plain
+False) or refuses a connection. A refused connection in the dial's exception
+chain counts too, but a missing *file* there does not: the client opens more
+than its socket, and only a stat of the socket path can tell a stopped
+daemon from a running one that tripped over something else. Every other
+failed dial (a socket this user may not open, the 3 s deadline, a reply that
+does not parse) leaves the daemon's state unknown. The difference is what the
 operator is told: the exception class and message always, and the advice to
 pass `--trust-registry` only when the daemon is verified down, since on
 anything else it would invite exactly the purge the supervisor union exists
 to prevent. An explicit `--dry-run` is told its preview is registry-only
 instead, having asked for nothing that could be downgraded.
+
+Every verb writes what it found or achieved (the status table, the listing,
+the reconcile report, the "Purged ..." line) to stdout, and every warning,
+refusal and diagnostic to stderr, `reclaim` included: a wrapper parses one
+stream without filtering the other, and a refusal leaves stdout empty rather
+than mixing a reason into a report that does not exist. The verbs take both
+streams as arguments, so the tests read them apart without capturing the
+process's own.
 
 `--trust-registry` remains the sharpest tool here even so, and the narrow
 advice does not blunt it: a supervisor that is verifiably down has not
--- a/docs/architecture/storage.md
+++ b/docs/architecture/storage.md
@@ -562,27 +562,62 @@ in its `--help` epilog) needs no running *agent process*:
   under `backup_old`) and a directory with no marker (it may belong to a
   live VM) are refused the same way. Then refuses a hash the agent registry
   or the supervisor considers live (see below), and refuses outright,
-  without `--trust-registry`, when the supervisor cannot be asked. Finally,
-  if the purge itself leaves the directory behind (a device-mapper target
-  still holds one of its volumes, the same guard `purge_vm_storage` always
-  applies), reclaim reports that and exits non-zero rather than claiming
-  success. `reclaim` never tears devices down itself: `storage reconcile`
-  is the CLI path that does, for every VM nothing owns, so the refusal
-  points the operator there.
+  without `--trust-registry`, when the supervisor is down. The marker is
+  read once more immediately before the purge: a re-create adopts its
+  retained directories by clearing their markers, and one that started
+  while the supervisor was being asked is in no answer this process holds,
+  so the second read is what keeps the purge off the disks a create is at
+  that moment building on. Finally, if the purge itself leaves the
+  directory behind (a device-mapper target still holds one of its volumes,
+  the same guard `purge_vm_storage` always applies), reclaim reports that
+  and exits non-zero rather than claiming success. `reclaim` never tears
+  devices down, and neither does any other CLI verb: the refusal points the
+  operator at the agent, whose own pass reclaims the device.
 - `storage reconcile [--dry-run] [--trust-registry]`: run one
-  `reconcile_storage()` pass with the daemon's two device steps around it,
-  or the CLI would keep refusing what the daemon reclaims. Before the pass,
-  `_teardown_orphan_devices` removes the dm devices of every namespace no
-  live VM owns, so the purge that follows is not refused on a volume file a
-  target still holds; that step needs the supervisor's answer and is skipped
-  under `--dry-run` and when the supervisor is unreachable, `--trust-registry`
-  included (it buys a purge on the registry's word, not a teardown). After
-  the pass, and printing what it removed (or, under `--dry-run`, what it
-  would remove), it tears down the devices of any evicted runtime cache
+  `reconcile_storage()` pass, with what the supervisor answers deciding
+  which passes are allowed to run at all:
+  - **The supervisor answers.** A real pass is refused (exit 3, the reason
+    on stderr, nothing written). The agent is running, and the set of
+    creates it has in flight is in-process state this command cannot see, so
+    a create that outlives `VOLUME_CREATE_GUARD` before its DB record exists
+    (a long migration import routinely does) would read as an orphan here.
+    A CLI pass also holds no lock against the agent's own. The agent runs
+    that pass itself at startup, periodically and after every VM goes away,
+    and it sees the creates too, so there is nothing here for a CLI pass to
+    add. `--dry-run` is still allowed and is how a running node is
+    inspected: it takes the supervisor's list into the live set exactly as a
+    real pass would.
+  - **The supervisor does not answer.** Nothing is being created, so the age
+    of the directory is a sound guard again and this is the command for a
+    node whose agent will not start. The live set is the registry alone,
+    which is not a safe one, so the pass runs dry, warns on stderr and exits
+    3 unless `--trust-registry` says to proceed on the registry's word.
+  After the pass, and printing what it removed (or, under `--dry-run`, what
+  it would remove), it tears down the devices of any evicted runtime cache
   parent the same way `reconcile_now` / `reconcile_at_startup` do
   (`remove_parent_device` per evicted ref, then `sweep_leaked_cache_loops`),
   so an evicted entry never leaves `/dev/mapper/<ref>` and its loop device
-  pinning the deleted inode.
+  pinning the deleted inode. The orphan-namespace teardown the daemon's
+  passes run first is not repeated here: tearing a device down takes the
+  disk of a VM that was merely unlisted, and the one answer that made it
+  safe (the supervisor's) is now the answer that refuses the pass outright.
+
+Both purge paths call the daemon's own `_startup_refusal` rather than
+restating its conditions, so a CLI pass and a daemon pass cannot drift
+apart: an empty registry while the supervisor runs VMs means the agent DB
+was lost, and both refuse on that alone.
+
+Whether the daemon is *down* is a claim the CLI only makes when it can back
+it up: a socket that is missing (`os.stat`, not `Path.exists()`, which turns
+a stat this user may not make into a plain False) or one that refuses a
+connection. Every other failed dial (a socket this user may not open, the
+3 s deadline, a reply that does not parse) leaves the daemon's state
+unknown. The difference is what the operator is told: the exception class
+and message always, and the advice to pass `--trust-registry` only when the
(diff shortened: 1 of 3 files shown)


foxpatch-aleph (approved):
This PR closes a genuine safety gap: a CLI `reconcile` could purge the directory and devices of an in-flight create because `creating()` is in-process state of the agent, and the CLI held no lock against the agent's own pass. The fix — refusing a non-dry-run reconcile whenever the supervisor answers — is the right call, and is implemented cleanly: `_startup_refusal` is imported rather than restated so the CLI and daemon verdicts cannot drift, the supervisor-reach classification is conservative in exactly the right direction (only a missing socket or a refused connection claims DOWN; permission errors, deadlines and malformed replies stay UNKNOWN and withhold the --trust-registry advice), and the reclaim marker re-check after the dial closes the real race with `adopt()`. Removing `_teardown_orphan_devices` is behavior-preserving since its only reachable invocation is now the refused path. All 43 storage-CLI tests pass, as do the related reconciler/reclaimable suites. Remaining nits are non-blocking: the lost-agent-DB warning from `_startup_refusal` is not surfaced on the dry-run path, `_socket_reach` could verify the path is a socket via `stat.S_ISSOCK` before calling a refused connect 'down', the window between the second marker read and the unlink in `purge_vm_storage` is narrowed but not eliminated, and `DEGRADED_EXIT_CODE` now also means 'refused'.

**`src/aleph/vm/agent/storage_cli.py`** (line 556): When the supervisor answers, the registry is empty while it runs VMs, and --dry-run is passed, `live_set.refusal` (the lost-agent-DB verdict) is computed but never surfaced — the preview silently reports orphans against a live set the daemon itself would refuse to act on. No writes happen, but printing the refusal on the dry-run path too would keep the preview honest in exactly the scenario the refusal exists for.

**`src/aleph/vm/agent/storage_cli.py`** (line 302): Minor: if SUPERVISOR_GRPC_SOCKET points at a regular (non-socket) file, `os.stat` succeeds and `connect()` raises ConnectionRefusedError on Linux, so it is classified DOWN and the purge advice is unlocked. Checking `stat.S_ISSOCK(st.st_mode)` after the stat would make "verified down" airtight. Unlikely configuration, but this is the classification that licenses purging.

**`src/aleph/vm/agent/storage_cli.py`** (line 499): Worth noting in the comment: the window between this check and the unlink inside `purge_vm_storage` is still TOCTOU-open (a create's `adopt()` clearing the marker in that instant). The check narrows the race from a ~3s dial to microseconds, which is a real improvement — just not a full close.

**`src/aleph/vm/agent/storage_cli.py`** (line 109): `DEGRADED_EXIT_CODE` now also covers the refused-while-the-agent-answers case, so the name has drifted from its original "silently downgraded to dry run" meaning. The comment block documents it, but a rename (e.g. NOT_RECONCILED_EXIT_CODE) would have kept the name honest. Cosmetic.

foxpatch-aleph (approved):
Well-designed, fail-closed safety hardening of the storage CLI. The three-state purge rule, the agent probe (both loopbacks, UNKNOWN counts as running), the supervisor DOWN/UNKNOWN classification, the re-read of the reclaimable marker before purge, and the stdout/stderr split are all correct and thoroughly tested (54 CLI tests, 118 with the reconciler/reclaimable suites, all passing). I independently verified the IPV6_V6ONLY claim against asyncio's create_server, confirmed SUPERVISOR_HOST/PORT is solely the Python agent's bind (the Rust daemon uses only a Unix socket), and checked that the changed run() signature has no stale callers. The fail-closed directions all err on the safe side, and exit-code semantics preserve pre-existing behavior. The comments below are non-blocking observations.

**`src/aleph/vm/agent/storage_cli.py`** (line 315): Non-blocking: aiohttp fires the on_startup hooks (which include reconcile_at_startup and the allocation reconciler launch) before site.start() binds the listening socket, so during agent startup there is a narrow window where the agent process is alive and working while nothing accepts on its bind address, and the probe reports STOPPED. It is inherent to port probing, the VOLUME_CREATE_GUARD still covers young directories for most of it, and it is a strict improvement over the pre-PR state — but the module docstring's "called stopped only when nothing accepts a connection" carries this caveat. A one-line mention in the docstring would make the limit of the probe's knowledge complete.

**`src/aleph/vm/agent/storage_cli.py`** (line 447): Nit: the chain is walked front-to-back and PermissionError short-circuits, so a ConnectionRefusedError sitting *deeper* in a chain whose head is a PermissionError goes unnoticed (UNKNOWN wins where DOWN was provable). Exotic, and it errs on the safe side — mentioning only in case you want the ConnectionRefusedError check hoisted out of the loop, which would also simplify the reasoning to "any refusal anywhere in the chain is proof".

**`src/aleph/vm/agent/storage_cli.py`** (line 307): Minor: this errno set is applied uniformly, but ENETUNREACH/EHOSTUNREACH mean "family unusable" only for the loopback targets of a wildcard probe. For a non-wildcard SUPERVISOR_HOST they can mean "address unreachable", which is weaker evidence the agent is not listening. The "both processes read the same settings" argument keeps it safe (an agent that could not bind would not be running), so this is fine as-is; just noting the reasoning relies on the shared-configuration invariant rather than on the errno itself.

**`src/aleph/vm/agent/storage_cli.py`** (line 631): Wording nit: in the reclaim refusal, "to see what it will find" reads as if 'storage list' will find something, but the antecedent is the agent's own reconcile pass. Something like "run 'storage reconcile --dry-run' to preview what the agent's own pass will find" would be unambiguous.

**`src/aleph/vm/agent/storage_cli.py`** (line 449): FYI: on Python 3.10 asyncio.TimeoutError is not builtins.TimeoutError, so a genuine wait_for deadline falls through to _socket_reach instead of this branch. Outcomes converge (a wedged daemon's socket still accepts, so UNKNOWN; a missing socket means DOWN is right anyway), so no change needed — just be aware this check only short-circuits on 3.11+.

foxpatch-aleph (approved):
Well-engineered, consistently fail-closed safety hardening of the storage CLI. The three-state purge rule, the agent probe (with correct dual-loopback handling of wildcard binds and the EAFNOSUPPORT-family classification), the classified SupervisorAnswer that replaces silent None, and the reuse of the daemon's _startup_refusal are all sound, well-documented, and thoroughly tested (58 tests, all passing, including real-socket IPv6 and refused-port cases). ruff format and isort pass with the pinned versions; mypy is clean for the changed file; remaining suite failures are environmental and unrelated. Only non-blocking nits found.

**`src/aleph/vm/agent/storage_cli.py`** (line 605): Minor diagnostic asymmetry: when --trust-registry lets a reclaim proceed on an unanswered supervisor, nothing reaches stderr, whereas _reconcile prints "Purging on the registry alone, as --trust-registry asked" in the equivalent state (_unanswered_warning). A wrapper reading the audit trail sees the purge happen with no record that it ran against an unconfirmed live set. Consider emitting the same note here.

**`src/aleph/vm/agent/storage_cli.py`** (line 462): Nit: this TimeoutError check inspects only the head of the chain, while the ConnectionRefusedError and PermissionError checks above walk the whole chain. It is functionally fine (a timeout buried under e.g. a gRPC RpcError head falls through to _socket_reach, which is the conservative answer), but the asymmetry is slightly surprising given the docstring's chain-walking emphasis. Worth a one-line comment, or walking the chain here too for uniformity.

**`src/aleph/vm/agent/storage_cli.py`** (line 663): Style nit: this "no longer marked reclaimable" branch writes directly to err and returns 1, bypassing the Refusal dataclass it sits beside (which exists to pair a message with its exit code). Moving it into _reclaim_refusal as a Refusal (checked just before returning) would keep the "every refusal goes through one shape" story intact and make the exit code auditable in one place.

**`src/aleph/vm/agent/storage_cli.py`** (line 316): Nit: _FAMILY_UNAVAILABLE is applied to non-wildcard targets too, so a SUPERVISOR_HOST naming an unreachable non-loopback address reads as STOPPED rather than UNKNOWN. That is the right call for the single-node deployment this CLI targets (and the docs frame these errnos as loopback facts), but a comment saying the set also covers explicitly-configured hosts would help a future reader. Also, the PR description's claim that "ruff check" exits 0 is not accurate repo-wide (thousands of pre-existing findings, and it is not a CI gate) - only ruff format and isort are enforced, and both pass.

foxpatch-aleph (approved):
This PR is carefully engineered and correct. The three-state purge rule is applied consistently in both reconcile and reclaim, and every liveness probe is fail-closed: the agent probe only claims STOPPED on refused connections (or family-unavailable errnos) on every loopback a wildcard bind could answer on; the supervisor classifier walks the whole exception chain for a refusal before checking for permission errors (the last commit's fix), treats a missing file anywhere in the chain as no proof, and falls back to an os.stat/connect probe of the socket itself; and the reclaim marker is re-read after the supervisor dial to close the adopt-during-dial race. Importing _startup_refusal rather than restating it keeps CLI and daemon refusals from drifting. The stream split (stdout for findings, stderr for refusals) is carried through every verb, exit codes are documented and wrapper-distinguishable, and the docs are accurate down to details I verified against the code (aiohttp binds after on_startup; errno list). All 60 storage-CLI tests plus 94 reconciler/reclaimable tests pass, and mypy is clean; the ruff format diff I see is version drift (0.9.10 vs the pinned 0.4.6 implicit-concat join), not a PR defect. Remaining nits are non-blocking.

**`src/aleph/vm/agent/storage_cli.py`** (line 457): On Python 3.10 (the project floor), asyncio.wait_for raises asyncio.exceptions.TimeoutError, which only became an alias of the builtin TimeoutError in 3.11. There, a timeout falls through to _socket_reach instead, which is still conservative (missing socket → DOWN, present socket → UNKNOWN), so this is behavior-neutral, but the branch is effectively 3.11+; a comment or checking both would make the floor explicit.

**`src/aleph/vm/agent/storage_cli.py`** (line 603): This verified-down refusal exits 1 while its reconcile analogue (a downgraded pass) and the lost-DB refusal in the same verb exit 3, and by the PR's own rationale — nothing about the hash is wrong, the command may not run here — a wrapper can't distinguish it from a bad hash. Presumably kept at 1 for backwards compatibility with the pre-PR behavior; worth a word in the Refusal docstring if so.

**`src/aleph/vm/agent/storage_cli.py`** (line 660): Minor wording: the marker could have been cleared by anything (a concurrent reclaim, another operator), not only a create adopting its directory. The refusal is right in every case; only the asserted cause is narrower than the possibilities.

PR #1216, opened 2026-09-09
fix(agent): put the retained marker back when a create does not commit

odesenfans:
## Summary
`creating()` adopts a VM's retained directories on entry (clearing the `.reclaimable` marker) before the create is known to succeed, and a create that then failed left them unmarked. The failure path retires `RECREATE` when the volumes already existed, which touches no storage, so the next reconcile pass saw an unmarked directory and re-marked it as an orphan: no owner, no `depends_on`, and a fresh `reclaimable_since`. The owner could no longer erase their own retained data (the operator erase endpoint 404s while there is no marker, and refuses a marker with no owner once the pass writes one), the parent images the retained volumes are built on lost their pin against cache eviction, and the directory went to the back of the eviction queue on every attempt. The allocation reconciler retries a failed create every cycle, so one that keeps failing cycled this indefinitely.

`adopt()` now returns the markers it removed and `creating()` puts them back on any exit other than a normal return, so a failed re-create leaves the directory exactly as it found it: retained, with its owner, its parent-image pins and its place in the eviction queue.

Stacked on #1213.

## Changes
- `src/aleph/vm/agent/vm/reclaimable.py`: `adopt()` returns the markers it cleared, keyed by directory (it counted them before). New `restore_markers()` writes them back exclusively, so a marker written while the create ran (a retire of the same hash) is the newer record and stays, and a directory the failed create's own teardown purged is skipped rather than recreated. Only `size_bytes` is re-measured: a create that failed part way may have left more or less on disk than it found, while the rest of the marker is the record that has to survive unchanged.
- `src/aleph/vm/agent/vm/reconciler.py`: `creating()` keeps what `adopt()` returned and restores it on any exception (a cancellation of the task running the create included), before the create guard comes down, so no pass can see the directory unmarked and unguarded in between. A restore that itself fails is logged and swallowed: the create's own failure is what the caller has to see. The docstring now states the rule.
- `docs/architecture/storage.md`: the marker section says adoption is undone when a create does not commit, and what the two exceptions are.

## Test plan
- `tests/supervisor/test_reconciler.py`: a failed create puts the retained marker back with its owner, `depends_on`, reason and timestamp; the eviction order survives repeated failed attempts; a create that commits leaves the directory adopted; a purged directory is not re-marked; a marker written during the create survives; a failed inner create does not re-mark a directory the outer create still holds; a restore that raises does not mask the create failure.
- `tests/supervisor/test_reclaimable.py`: `adopt()`/`restore_markers()` round trip across both pools a VM spans, `size_bytes` is re-measured, a directory that is gone is skipped.
- `tests/supervisor/views/test_operator_retire.py`: end to end on the real marker machinery, an owner's erase of retained disks still returns 200 after a failed re-create (it was a 404 before the fix).
- Checks: `pytest tests/supervisor` (1207 passed, 1 skipped, 1 pre-existing journald failure in `test_log.py::test_make_logs_queue`, identical on the base commit), `ruff format --diff`, `isort --check-only --profile black`, `mypy` on the changed files and on `src/aleph/vm/` (all exit 0). No Rust touched.

- `tests/supervisor/test_reconciler.py`: a test that a create cancelled while inside the context also restores the markers it adopted, the half of the `BaseException` clause `Exception` does not cover. It fails when the clause is narrowed.
- Not changed: the partial-adopt corner (an `adopt` that raises after clearing one pool's marker). Those directories fall back to orphan recovery, which is what happened before this PR.

### DIFF
--- a/docs/architecture/storage.md
+++ b/docs/architecture/storage.md
@@ -332,6 +332,13 @@ purge is immediate.
 A create for the same hash adopts the directory (the marker is unlinked and
 the volumes are reused through the existing sticky placement in
 `volume_path_for`), which is what `reconciler.creating()` does on entry.
+Adoption is undone when the create does not commit: leaving the guard by
+raising puts the adopted markers back as they were, so a re-create that
+keeps failing (the allocation reconciler retries it every cycle) does not
+turn retained disks into an owner-less orphan directory with no parent-image
+pins and a timestamp that resets on every attempt. A directory the failed
+create's own teardown purged is not re-marked, and a marker written while
+the create ran (a retire of the same hash) is the newer record and stays.
 
 ### The reconciler
 
@@ -754,7 +761,8 @@ pass against an empty registry would call every running VM an orphan.
 - `src/aleph/vm/agent/vm/retire.py`: `retire_vm` and `RetireReason`, the one
   way a VM's storage is released.
 - `src/aleph/vm/agent/vm/reclaimable.py`: the `.reclaimable` marker
-  (`mark_reclaimable`, `adopt`, `iter_reclaimable`, `reclaimable_bytes`) and
+  (`mark_reclaimable`, `adopt`, `restore_markers`, `iter_reclaimable`,
+  `reclaimable_bytes`) and
   the one enumeration of the cache entries a message names
   (`iter_content_refs`, `refs_from_content`, `depends_on_from_content`).
 - `src/aleph/vm/agent/vm/cache.py`: the cache budget, LRU eviction
--- a/src/aleph/vm/agent/vm/reclaimable.py
+++ b/src/aleph/vm/agent/vm/reclaimable.py
@@ -16,8 +16,8 @@ import json
 import logging
 import os
 import time
-from collections.abc import Iterator
-from dataclasses import asdict, dataclass
+from collections.abc import Iterator, Mapping
+from dataclasses import asdict, dataclass, replace
 from datetime import datetime, timezone
 from pathlib import Path
 from typing import Literal
@@ -307,15 +307,54 @@ def mark_reclaimable(
     return written
 
 
-def adopt(namespace: str) -> int:
-    """A create for this hash takes its retained directories back."""
+def adopt(namespace: str) -> dict[Path, ReclaimableMarker]:
+    """A create for this hash takes its retained directories back.
+
+    Returns the markers it removed, keyed by directory, so a create that
+    does not commit can put them back (``restore_markers``). A directory
+    whose marker was unreadable or corrupt is adopted like any other and
+    simply has nothing to give back.
+    """
     namespace = _checked_namespace(namespace)
-    cleared = 0
+    adopted: dict[Path, ReclaimableMarker] = {}
     for directory in iter_namespace_dirs(namespace):
+        marker = read_marker(directory)
         if clear_marker(directory):
             logger.info("Adopted retained volumes in %s", directory)
-            cleared += 1
-    return cleared
+            if marker is not None:
+                adopted[directory] = marker
+    return adopted
+
+
+def restore_markers(adopted: Mapping[Path, ReclaimableMarker]) -> int:
+    """Put back the markers an adopt cleared, for a create that then failed.
+
+    Adoption happens before the create is known to succeed, and a failed
+    create must leave the directory as it found it. Left unmarked it is only
+    an orphan to the next pass, which re-marks it with no owner (so the
+    owner can no longer have their own retained data erased) and no
+    depends_on (so the cache may evict the parent image the retained volumes
+    are built on), and with a fresh timestamp that moves it to the back of
+    the eviction queue on every retry.
+
+    Restored exclusively: a marker written while the create ran is a newer
+    record of the same directory (a retire of this very hash) and stays. A
+    directory the failed create's own teardown purged is skipped rather than
+    recreated.
+    """
+    restored = 0
+    for directory, marker in adopted.items():
+        if not directory.is_dir():
+            continue
+        # size_bytes is a measurement of what the directory holds, and a
+        # create that failed part way may have left more or less than it
+        # found. The rest of the marker (since when, whose, what it is built
+        # on) is the record that has to survive unchanged.
+        current = replace(marker, size_bytes=directory_size_bytes(directory))
+        if write_marker(directory, current, exclusive=True):
+            logger.info("Restored the reclaimable marker in %s after a create that did not commit", directory)
+            restored += 1
+    return restored
 
 
 def retained_marker(namespace: str) -> ReclaimableMarker | None:
--- a/src/aleph/vm/agent/vm/reconciler.py
+++ b/src/aleph/vm/agent/vm/reconciler.py
@@ -67,6 +67,7 @@ from aleph.vm.agent.vm.reclaimable import (
     iter_reclaimable,
     mark_reclaimable,
     read_marker,
+    restore_markers,
 )
 from aleph.vm.agent.vm.retire import teardown_namespace_devices
 from aleph.vm.agent.vm_registry import AgentVmRegistry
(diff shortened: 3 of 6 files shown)


foxpatch-aleph (approved):
This PR correctly fixes a real bug: a failed create left adopted retained directories unmarked, and since the failure path retires RECREATE (which touches no storage), the next pass re-marked them as owner-less orphans with fresh timestamps — costing the owner their erase authorization, dropping parent-image eviction pins, and resetting eviction order on every retried create. The fix (adopt returns the cleared markers; creating() puts them back on any non-normal exit) is well-scoped and its edge cases are handled thoughtfully: exclusive restore preserves a newer marker written during the create, purged directories are skipped, size_bytes is re-measured while the rest of the record survives unchanged, and the restore runs before the guard comes down to close the unmarked-and-unguarded window. I verified all race interactions (nested creates, concurrent re-marking, teardown purges) against the existing reconciler invariants, ran the reconciler (72), reclaimable (32), and operator-retire (13) test files — all green, including the new end-to-end erase test — and confirmed ruff/isort/mypy are clean on the changed files. Two minor, non-blocking notes below concern a partial-adopt failure corner and the untested cancellation path.

**`src/aleph/vm/agent/vm/reconciler.py`** (line 146): Minor: if adopt() raises partway (e.g. clear_marker fails on the second pool after clearing the first), the markers cleared before the raise are not restored, since adopted is only assigned on success. Those directories fall back to the pre-fix orphan-recovery path, so this is strictly better than base — but if the partial-adopt corner ever matters, adopt could accumulate its result into a caller-supplied dict the way it now returns one. Not blocking.

**`src/aleph/vm/agent/vm/reconciler.py`** (line 148): The docstring and comment say cancellation of the task running the create is covered by catching BaseException, which is correct — but no test exercises the CancelledError path (every failure test raises RuntimeError). A small test that asyncio.CancelledError also restores the markers would pin the one half of the BaseException clause that Exception does not cover.

foxpatch-aleph (approved):
The fix is correct and carefully engineered: adopt() hands back the markers it cleared, and creating() puts them back on any non-commit exit, before the guard comes down, so a retried failing create can no longer strip an owner, the parent-image pins, or the eviction-queue position from a retained directory. All the edge cases I traced hold up: exclusive writes make the restore safe against a concurrent retire of the same hash, purged directories are skipped, the inner-create overlap is a no-op (tested), and the mirror overlap (outer fails after a sibling committed) only leaves a stale marker on a live VM's directory, which _is_orphan clears and _evict refuses to act on. Test coverage is thorough, including the BaseException/cancellation half and an end-to-end erase test; I ran the reconciler and reclaimable suites locally (105 passed) — the operator-view suite needs the containerized environment. Nits below are documentation and test style only.

**`src/aleph/vm/agent/vm/reconciler.py`** (line 148): One residual worth a line in the docstring, given how diligently this codebase documents its residual windows (cf. the window comment in _evict): when two sibling creates for one hash overlap and the one that adopted the markers fails after the other committed, this restore writes the marker onto a directory a live VM now owns. It self-heals — _is_orphan clears a marker on a live/creating namespace with a warning, _evict refuses to evict live ones, and operate_erase only consults retained_marker when there is no registry record — but until the next pass the marker does inflate reclaimable_bytes for admission accounting. A one-sentence note here (or in restore_markers' docstring) would spare the next reader some archaeology.

**`src/aleph/vm/agent/vm/reclaimable.py`** (line 315): Minor docstring nit: for a corrupt marker, read_marker deletes the file as a side effect, so clear_marker then returns False and the directory is neither logged as adopted nor recorded — its outcome matches the unreadable case, but "adopted like any other" is not quite what happens on that path. The filesystem ends up in the same state either way; just a wording tightening if you care.

**`tests/supervisor/views/test_operator_retire.py`** (line 292): Style nit: setup_webapp is handed a supervisor that is immediately discarded, then a second _fake_supervisor() is built on the next line and installed over it. Building one fake, overriding its get_vm, and passing it to setup_webapp would read a little cleaner. (I could not run this suite outside the container — /var/cache/aleph and /opt/firecracker paths — so CI is the arbiter here.)

foxpatch-aleph (approved):
This PR correctly fixes a real bug: a failed create left adopted directories unmarked, causing the next reconcile pass to re-mark them as owner-less orphans with no parent-image pins and a resetting eviction timestamp. The implementation is careful and correct — restore runs under the create guard before it drops (no unmarked-and-unguarded window), BaseException covers task cancellation, exclusive marker writes let a concurrent GONE retire's newer marker survive, purged directories are skipped, and nested creates restore only what each adopted. The only production caller of adopt() is creating(), so the return-type change is safe, and the stale-marker-under-a-live-VM corner (overlapping creates where the adopter fails and the other commits) is declined by the pre-existing is_live check in _evict. Tests are thorough (restore semantics, eviction-order preservation, cancellation, committed-create guard, purged-directory skip, concurrent-retire marker survival, restore-failure masking, nested creates, and an end-to-end operator erase test on the real marker machinery). All storage-related test modules pass in this environment, and mypy/ruff/isort are clean on the changed files; the test_operator_retire.py failures observed here are environment-only (missing /var paths and /opt/firecracker/vmlinux.bin) and affect pre-existing tests identically.

**`src/aleph/vm/agent/vm/reclaimable.py`** (line 321): Minor: the added read_marker() call has a side effect — it unlinks a corrupt marker — so for a corrupt marker clear_marker() now returns False and the "Adopted retained volumes in %s" log line is skipped. Corrupt-marker adoptions are now silent (before this change they were logged as ordinary adoptions). Log-only, and the docstring's claim that such a directory is "adopted like any other" is accurate, so this is a nit at most — a debug-level line or a note might be worth it.

foxpatch-aleph (approved):
The fix is correct and well-constructed: adoption is undone on any non-commit exit of creating(), the restore runs while the create guard is still up, and the exclusive write preserves any newer marker written during the create. The adopt() return-type change has exactly one production caller, all failure paths in run.py (RECREATE leaves storage alone, FAILED_CREATE purges it) are consistent with the restore being the only re-marker, and the erase endpoint logic confirms the end-to-end claim (404 without a marker, 403 without an owner, 200 with the restored owner). Tests are thorough — restore on exception and cancellation, over-restoring guards, purged directories, newer-marker survival, restore failure not masking the create failure, nested creates, and eviction-order preservation — and all relevant suites pass locally (143 tests), with ruff format and mypy clean on the changed files. The operator retire suite could not run in this sandbox (requires root for /var/cache/aleph), an environment limitation rather than a code issue. Two non-blocking observations below.

**`src/aleph/vm/agent/vm/reclaimable.py`** (line 321): Non-blocking observation: read_marker followed by clear_marker is a two-step TOCTOU — a marker replaced between the two (a concurrent GONE retire's os.replace) is cleared here and the older, previously-read record is what restore_markers puts back on failure. That is strictly no worse than the pre-PR behavior (the newer marker was lost either way) and the exclusive restore prevents clobbering anything newer still, but the window is worth a line in the docstring so future readers know it was considered.

**`src/aleph/vm/agent/vm/reclaimable.py`** (line 310): The partial-adopt corner (an adopt that raises after clearing one pool's marker loses those markers to orphan recovery) is called out in the PR description as deliberately unchanged, but a brief note here in adopt()'s docstring would spare future readers the archaeology.

foxpatch-aleph (approved):
The change is correct and well-engineered: adopt() now returns what it cleared, and creating() puts the markers back on any non-normal exit, while the create guard is still up, so a failed re-create cannot regress a retained directory into an owner-less orphan with a reset eviction timestamp. The exclusive write correctly defers to a marker written during the create, purged directories are skipped, and re-measuring size_bytes is the right call. I traced the contextmanager semantics (re-raising the same exception object propagates unchanged to the caller, and the restore runs synchronously with no await points, so a second cancellation cannot interrupt it), verified the orphan re-stamp and erase-404/403 failure modes against the base code, and confirmed the return-type change has a single production caller. Test coverage is thorough, including the cancellation half, the commit guard, the purged-directory skip, the concurrent-retire race, restore-failure masking, and an end-to-end test through the operator erase endpoint; docs match the implementation. One non-blocking note: an exception raised after the VM has actually started (e.g. persist_record in run.py failing after create_vm succeeded) will restore a marker onto a live VM's directory; the reconciler already clears markers on live-VM directories on the next pass (reconciler.py:381-386), so this self-heals, but it may be worth a line in the docstring. I could not execute the test suite because this environment's Python dependencies are broken, so test runs are left to CI as the PR itself anticipates.

**`src/aleph/vm/agent/vm/reconciler.py`** (line 148): Non-blocking: any exception leaving the with-block triggers the restore, including one raised after the create actually committed its work — e.g. persist_record at run.py:526 failing after create_vm succeeded. The restored marker then sits on a live VM's directory until the next pass clears it via the live-VM cleanup in _is_orphan (reconciler.py:381-386), and until then reclaimable_bytes() over-counts by this VM's size. This self-heals, so it is fine as is, but the docstring's rule ('any way out other than a normal return means the create did not commit') slightly overstates the case; a sentence acknowledging the transient marker on a live VM might help future readers.

**`src/aleph/vm/agent/vm/reclaimable.py`** (line 316): Minor docstring nit: for a directory whose marker is corrupt, read_marker() unlinks it as a side effect and returns None, so clear_marker() then returns False and the directory is neither logged as adopted nor returned — it is effectively adopted (unmarked) with nothing to give back, which the sentence says, but a reader could take 'adopted like any other' to mean it also appears in the returned dict. Consider phrasing it as 'is cleared like any other and simply has nothing to give back'.

PR #1217, opened 2026-09-09
fix(agent): leave a VM its owner stopped alone; a push acknowledges it, never restarts it

odesenfans:
## Summary
Stopping a persistent VM through the operator API leaves it STOPPED and still
defined, and the allocation loop read that as a VM it owed a start: the
supervisor's down event nudged the reconciler, STOPPED is not a live status,
and the next pass called `start_persistent_vm`, whose STOPPED branch resumes
the VM in place. The owner's stop came undone within seconds, and since the
crash backoff landed it also cost a rebuild attempt, so a VM the owner stopped
and started climbed the crash ladder and showed up as `failed` in the
executions list on the way.

The rule this PR settles on is that a VM this node already holds is
acknowledged by a push, in any state, and only its owner starts it again. A
push never restarts a VM in STOPPED or STOPPING, a stop event no longer wakes
the loop, and `compute_verdict` answers such a hash under `unchanged` instead
of sizing it through `simulate`: the scheduler's belief that the VM is
allocated here is exactly what has not changed. Sizing it would let a node
that is tight on room refuse a VM it is already holding, and a refusal is
what takes the VM out of the plan the loop converges on.

Accounting does not move. The stopped VM keeps its registry record, so its
memory, vCPUs and disk stay committed and the headroom advertised to the
scheduler stays reduced by them. Forgetting them on a stop is how a node
over-provisions: it would promise the same room to somebody else and have
nowhere to put the VM when its owner starts it again.

Teardown is unchanged, and it is how the scheduler unallocates a stopped VM:
a push that stops naming the hash tears it down like any other. FAILED is
unchanged too, nobody's decision, so the loop still rebuilds it under the
crash-loop backoff of #1214.

This diverges from v1 deliberately. In v1 an allocation push restarted a
stopped VM, so an owner's stop lasted until the next push and no longer. In
2.1 the push acknowledges the stop and the owner is the only one who ends it.

Stacked on #1214.

## Changes
- `allocation/plan.py`: new `STOPPED_STATUSES` (STOPPING, STOPPED), next to
  `LIVE_STATUSES`, because the verdict needs it as much as the loop does.
- `allocation/reconciler.py`: `notify_vm_down` takes the status (required:
  both callers have one) and does not wake the loop for a stop.
  `_start_missing` skips a planned hash the supervisor reports STOPPED or
  STOPPING, clearing its allocation phase and keeping any failure record as
  attempt history, so a VM that failed twice and then sat stopped climbs the
  ladder from where it was. `_start_one` documents that only a VM held FAILED
  now reaches it with a `down` info.
- `allocation/verdict.py`: a planned hash the supervisor holds stopped or
  stopping is answered `unchanged` and never reaches `simulate`.
- `supervisor.py`: the event watcher and the post-outage reconcile pass the
  status along with the nudge (None when the VM is gone from the list
  entirely, which still wakes the loop).
- `capacity.py`: `_candidate_disk`'s docstring said a re-listed stopped VM is
  sized as a candidate, which is now FAILED's case and not STOPPED's.
- `docs/architecture/vm-lifecycle.md`: the paragraph in "Stop, start, reboot"
  states the rule, the capacity guarantee and the v1 divergence.

STOPPING is treated as a stop as well: a pass that caught a stop in flight
would take `start_persistent_vm`'s wait-until-gone path and recreate the VM
from scratch, the same failure through a slower door.

## Test plan
- `tests/supervisor/test_allocation_reconciler.py`: an owner-stopped VM is
  left alone by a pass and by a re-pushed plan; a stop does not wake the loop
  while a failure does; a VM that died is still rebuilt on a nudge; a stopped
  VM reports no allocation state of its own; a VM caught mid-stop is left
  alone too; a VM this node already holds stopped is not started by a first
  plan either; a stopped VM whose last start failed still reports why; a
  stopped VM the plan stops naming is still torn down. The watcher test pins
  the status travelling with the nudge.
- `tests/supervisor/test_allocation_verdict.py`: a stopped or stopping entry
  is `unchanged` and `simulate` is called with no candidates; the headroom
  the node advertises stays reduced by a stopped VM's memory; a newcomer that
  would only fit if that VM were forgotten is refused
  `insufficient_capacity`. The recreate fixtures that used STOPPED as a
  stand-in for "a VM to build again" now use FAILED, the only status that
  still means that.
- `tests/supervisor/test_executions_list_allocation.py`: a planned VM the
  owner stopped renders `state: stopped` and no allocation block.
- Checks: `ruff format --diff` (exit 0), `isort --check-only --profile black`
  (exit 0), `mypy src/aleph/vm/ tests/` from the worktree root (Success, exit
  0). The test suites are left to CI on this round.

### DIFF
--- a/docs/architecture/vm-lifecycle.md
+++ b/docs/architecture/vm-lifecycle.md
@@ -253,6 +253,37 @@ have to restart. `StopVm` and `StartVm` on an ephemeral program are flatly
 "stopped but still defined" state for a program (no on-disk config, no
 systemd unit to restart), so the daemon refuses rather than faking one.
 
+The agent's allocation loop respects that stop, and respects it for good.
+The loop converges the node onto the scheduler's plan, and a planned VM that
+is down is normally started again on the next pass, but a VM that was stopped
+is a decision somebody made: the owner's through
+`/control/machine/{ref}/stop`, or the guest's own shutdown. The loop
+therefore treats STOPPED and STOPPING as terminal, full stop. A stop event
+does not wake it, its backstop pass leaves the VM alone, and a plan naming
+the VM does not restart it either: the owner stopped it, and the owner is who
+starts it again, through the same operator API. The executions list carries
+no allocation block for it, since the supervisor already reports STOPPED and
+the agent has nothing to add. It does keep reporting a start it was asked to
+make and could not: a VM left stopped because its create failed still carries
+the failure and the attempt count, which is the node's news rather than the
+owner's. FAILED is the opposite case, nobody's decision, so the loop rebuilds
+it on the event, damped by the retry backoff. Without the split an owner
+could not keep a planned VM stopped at all, since the plan is level-triggered
+and re-pushed for as long as the VM is allocated here.
+
+What the scheduler is told is that the VM is here. A plan entry for a stopped
+VM is answered `unchanged`, the same answer a running one gets, because the
+scheduler's belief that the VM is allocated to this node is exactly what has
+not changed. The node keeps committing its memory, vCPUs and disk for it too:
+the definition and the volumes are still allocated, and forgetting them would
+let the node promise the same room twice. Unallocating a stopped VM stays the
+scheduler's own move, made by dropping the hash from the plan, which the loop
+reads as a teardown like any other.
+
+This diverges from v1, where an allocation push restarted a stopped VM. The
+node answered every push by starting whatever it held down, so an owner's
+stop lasted until the next push and no longer.
+
 ### Idle expiry
 
 Idle teardown is agent policy, not something the supervisor knows about: the
--- a/src/aleph/vm/agent/allocation/plan.py
+++ b/src/aleph/vm/agent/allocation/plan.py
@@ -22,6 +22,14 @@ logger = logging.getLogger(__name__)
 # again would be a second VM under the same hash.
 LIVE_STATUSES = (VmStatus.RUNNING, VmStatus.BOOTING, VmStatus.DEFINED)
 
+# States a VM only reaches because it was told to: the owner's stop through the
+# operator API, or a guest shutting itself down. The definition and the disks
+# survive, so the VM is down rather than lost, and starting it again is the
+# owner's call and nobody else's. STOPPING is here too because a stop caught
+# mid-flight is still a stop: a caller that read it as work to do would wait
+# the VM out and then recreate it from scratch.
+STOPPED_STATUSES = (VmStatus.STOPPING, VmStatus.STOPPED)
+
 
 def by_hash(infos: list[VmInfo]) -> dict[ItemHash, VmInfo]:
     """The supervisor's VMs, keyed by item hash.
--- a/src/aleph/vm/agent/allocation/reconciler.py
+++ b/src/aleph/vm/agent/allocation/reconciler.py
@@ -6,6 +6,22 @@ an empty diff and a plan arriving mid-convergence simply wins at the next step
 boundary. This is the seam pull mode plugs into: replacing "the scheduler
 pushed a plan" with "the agent fetched a plan" is a change to submit()'s
 caller and nothing else.
+
+One asymmetry runs through the loop: a VM that is stopped and a VM that is
+dead are not the same thing. A stop is somebody's decision, the owner's
+through the operator API or the guest's own, and only its owner undoes it:
+the loop never starts a stopped VM, not on a down event, not on its own
+backstop interval, and not because a plan lists it. A VM that failed is
+nobody's decision, so the loop rebuilds it on the event, damped by the
+backoff. Without that split the owner could not keep a planned VM down at
+all, since the plan is level-triggered and re-pushed for as long as the VM is
+allocated here.
+
+What the scheduler is told about a stopped VM is that it is here: the plan
+naming it is answered "unchanged", and the node goes on committing its
+memory, vCPUs and disk for it, because the definition and the volumes are
+still allocated. Unallocating it stays the scheduler's own move, made by
+dropping the hash from the plan, which the loop reads as a teardown.
 """
 
 import asyncio
@@ -18,6 +34,7 @@ from aleph_message.models import ItemHash
 
 from aleph.vm.agent.allocation.plan import (
     LIVE_STATUSES,
+    STOPPED_STATUSES,
     AllocationPlan,
     AllocationState,
     FailureRecord,
@@ -97,9 +114,18 @@ class AllocationReconciler:
                 self._forget(vm_hash)
         self._wakeup.set()
 
-    def notify_vm_down(self, vm_id: str) -> None:
-        """A VM went STOPPED or FAILED. If the plan still wants it, converge."""
+    def notify_vm_down(self, vm_id: str, status: VmStatus | None) -> None:
+        """A VM went down. If the plan still wants it up, converge now.
+
+        `status` is what the supervisor reported, or None when the VM is gone
+        from its list entirely. A stop is not a death: the loop has nothing to
+        do about a VM somebody stopped, so a stop does not wake it, while a VM
+        that failed or vanished is rebuilt at once rather than at the backstop
+        interval.
+        """
         if self._desired is None:
+            return
+        if status in STOPPED_STATUSES:
             return
         try:
             vm_hash = ItemHash(str(vm_id))
@@ -210,6 +236,25 @@ class AllocationReconciler:
         for vm_hash in plan.entries:
             if vm_hash in live:
                 continue
+            info = known.get(vm_hash)
+            if info is not None and info.status in STOPPED_STATUSES:
+                # Somebody stopped this VM, so it stays stopped: the plan
+                # listing it says the scheduler still holds it here, not that
+                # the node should overrule the owner. What the agent still has
+                # to report is whether its last attempt failed: a stop is not
+                # the agent's news to tell, since the supervisor already says
+                # STOPPED, but a start the node was asked to make and could
+                # not is. Only the phase is cleared, and the failure record is
+                # deliberately left standing: it is the attempt history, and a
+                # VM that failed twice, sat stopped a while and then crashed
+                # should climb the ladder from where it was rather than from
+                # the bottom. _forget_settled drops it once the VM has been up
+                # long enough to call healthy.
+                if vm_hash in self._failures:
+                    self._states[vm_hash] = AllocationState.FAILED
+                else:
+                    self._states.pop(vm_hash, None)
+                continue
             if not self._retry_due(vm_hash, now):
                 # Down and waiting out its backoff. Saying so is what lets the
                 # executions list report the wait and the time it ends, rather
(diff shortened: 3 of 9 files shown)


foxpatch-aleph (approved):
The change is correct and well-tested. The split between 'stopped' (owner's/guest's decision, held until the next push) and 'failed' (nobody's decision, rebuilt on the event under the crash backoff) is implemented cleanly: the status partition over VmStatus is complete, the _start_owed accounting is sound under single-threaded asyncio (submit replaces the set wholesale; live VMs and completed starts discard their entry, which also correctly absorbs a push landing mid-start), and the one-start-per-push invariant holds even when the push-driven restart fails. All call sites of notify_vm_down are updated, FAILED VMs keep the #1214 backoff, the executions-list behavior is consistent with the allocation-block semantics, and the docs match the code. All 145 allocation-related tests pass; the unrelated failures in the supervisor suite are environmental (missing /var/lib/aleph).

**`src/aleph/vm/agent/allocation/reconciler.py`** (line 256): Minor asymmetry: this branch pops `_states` but not `_failures`, so a FailureRecord from a failed push-driven restart lingers in `_failures` while the VM waits for the next push. It is invisible in the executions list (since `_allocation_block` returns None when state is None) and cannot gate a future push-driven restart, but `state_for` will hand a raw consumer a `(None, failure)` pair whose `next_retry_at` will never be honored. Keeping the record does preserve attempt history across pushes, which is arguably right — but if so, a word saying why `_failures` deliberately outlives `_states` here would save the next reader the same double-check.

**`src/aleph/vm/agent/allocation/reconciler.py`** (line 66): Extending stop semantics to STOPPING is deliberate and documented ("a stop caught mid-flight is still a stop"), but no test exercises a VM listed as STOPPING at pass time — i.e., that a mid-stop VM is left alone when not owed, and that the push-driven start takes the wait-then-recreate path in `start_persistent_vm` (run.py:1101). Given the comment calls out exactly what the old behavior was, a test pinning it would be cheap.

**`tests/supervisor/test_allocation_reconciler.py`** (line 481): The one-start-per-push invariant is only tested on the success path. Worth adding: a push-driven restart of a stopped VM that fails (`fail=True`) followed by another pass asserting no second start — the discard at reconciler.py:263 is what guarantees it, and a regression there (e.g., moving the discard to after a successful start) would go unnoticed by the current tests.

**`docs/architecture/vm-lifecycle.md`** (line 249): The paragraph is accurate, but note the corollary it implies: since the plan is level-triggered and re-pushed periodically, an owner's stop only holds until the next push, and the node has no channel to tell the scheduler "I'm holding this one stopped" — the scheduler's execution list will show the VM flapping stopped/running at push cadence. That is the documented v1-compatible semantics, so no change needed; flagging it only so the trade-off is a conscious one.

foxpatch-aleph (approved):
The stopped/dead split is correctly implemented and the generation-stamped `_start_owed` protocol is sound: I traced the mid-pass, double-push, failed-start and dropped-VM sequences and found no way a push is swallowed or answered twice, nor a way the loop starts a stopped VM without one. All `notify_vm_down` callers are updated, the no-attempt-charged rule is consistently applied in both `_start_missing` and `_start_one`, tests are comprehensive and pass (129 across the allocation suites), and ruff format is clean. The three comments are non-blocking: one rationale comment that overstates what the code delivers, one reporting wrinkle for a stopped VM carrying an old crash record, and one docs wrapping nit.

**`src/aleph/vm/agent/allocation/reconciler.py`** (line 274): This rationale ("a VM that failed twice, sat stopped a while and then crashed should climb the ladder from where it was") is undercut by `_start_one`: a *successful* push-driven restart of a stopped VM takes the `down.status in STOPPED_STATUSES` branch and calls `_forget`, wiping the record. So the ladder survives only across *failed* restart attempts while stopped; the moment one succeeds, a later crash starts from the bottom (and `_forget_settled` can never age the record out while the VM sits stopped, since it only drops records of live VMs). If the current behavior is intended — which reads as reasonable, mirroring `test_a_successful_start_clears_a_previous_failure` — consider softening this comment; if the comment's guarantee is intended, the stopped-restart path needs a timed drop rather than `_forget`.

**`src/aleph/vm/agent/allocation/reconciler.py`** (line 281): This surfaces any failure record as `allocation.state: failed`, but the record is not always "a start the node was asked to make and could not": a VM that crashed, was rebuilt (record: `vm_failed` / "rebuilt after the supervisor reported it failed"), and was then stopped by its owner within ALLOCATION_RETRY_MAX_INTERVAL (900s) will render `state: stopped` alongside an allocation block claiming "failed" with a stale `next_retry_at`, for as long as it stays stopped. Cosmetic and arguably fine (it is genuine crash history), but worth noting since the docs frame the lingering block strictly as a failed start.

**`docs/architecture/vm-lifecycle.md`** (line 264): This line runs past the file's ~78-column wrap style ("...never marked answered by the one that could not have read it. FAILED is the opposite case, nobody's decision, so the loop"). Worth re-wrapping for consistency with the surrounding paragraphs.

foxpatch-aleph (approved):
This PR is a correct and well-executed split between "stopped" and "dead" in the allocation loop. The status space is partitioned cleanly (LIVE_STATUSES / STOPPED_STATUSES / FAILED) with no gaps or overlaps, both callers of notify_vm_down were updated to pass the status, and the None-status (VM vanished) path still wakes the loop. Ordering in _start_missing is right: the stopped check precedes _retry_due, so a stopped VM is never misreported as waiting out backoff, while its failure record is deliberately kept as attempt history. The verdict change to "unchanged" is the correct answer — sizing a held-stopped VM as a candidate could refuse it for capacity and drop it out of the plan the loop converges on — and the two tests against the real CapacityManager verify the capacity guarantee both ways. All 103 tests in the touched suites pass, mypy is clean on changed files, and the docs (including the deliberate v1 divergence) are thorough. Only minor nits remain.

**`src/aleph/vm/agent/allocation/verdict.py`** (line 191): Nit: with STOPPED/STOPPING now answered unchanged above, the fall-through this comment describes is FAILED-only. The comment still reads correctly ("dead" now unambiguously means FAILED), but naming the statuses that actually reach it would save a reader a moment of doubt.

**`src/aleph/vm/agent/allocation/reconciler.py`** (line 254): Minor presentational consequence worth being aware of: a stopped VM carrying a failure record reports state "failed" with a next_retry_at, even though nothing will retry it while it is stopped — the field renders as inert history until the owner restarts. The docstrings and tests suggest this is deliberate, so just a note.

**`src/aleph/vm/agent/allocation/reconciler.py`** (line 128): Suggestion: the status=None wake path (VM gone from the supervisor list entirely, reached via _reconcile_after_event_gap) has no direct test — the gap-reconcile function itself is untested. The live-watcher path is covered, so this is a nice-to-have, not a blocker.

foxpatch-aleph (approved):
A correct, well-documented and well-tested fix. The three halves of the change reinforce each other: the reconciler leaves STOPPED/STOPPING VMs alone (both on the down event and on the backstop pass), the verdict answers a planned stopped VM as `unchanged` without sizing it — which is load-bearing, since it keeps the hash in `plan.entries` (protecting it from the teardown sweep) and prevents a tight node from refusing a VM it already holds — and capacity stays committed through the registry record, so a stop cannot over-provision the node. The deliberate v1 divergence, the treatment of STOPPING, and the FAILED-vs-STOPPED split are all documented and pinned by tests. All relevant test suites pass locally, mypy is clean, and ruff format is clean on the changed files. Only minor nits follow, none blocking.

**`tests/supervisor/test_allocation_reconciler.py`** (line 649): Minor: this test's name and docstring went stale with the change it was updated for. It emits a STOPPED event, which now deliberately does NOT nudge the loop — that is the PR's whole point — yet the name says "nudges the reconciler when a VM goes down" and the docstring says "A VM that dies while still planned must be retried at once, not at the backstop interval." The assertion correctly pins that the status travels with the call; consider renaming (e.g. "the watcher passes the status with the down report") and rewording the docstring to match.

**`src/aleph/vm/agent/supervisor.py`** (line 212): Note: `_reconcile_after_event_gap`'s nudge-with-status has no direct test — the gap path was untested before this PR too, so this is a pre-existing gap, and the live-stream twin in watch_supervisor_events is covered. Not blocking, but the two call sites of notify_vm_down now both encode the stop-vs-death split, and only one is pinned.

**`src/aleph/vm/agent/allocation/reconciler.py`** (line 240): Worth knowing rather than fixing: a pass that read list_vms() just before an owner's stop can still undo the stop — start_persistent_vm re-reads the status and resumes a STOPPED VM in place, and no event will put it back. The window is inherent to level-triggered convergence without coordinating with the operator stop endpoint, and this change shrinks the exposure from "every pass" to "in flight only", so it is a large improvement rather than a regression. If it ever matters, the operator stop path could invalidate the reconciler's snapshot.

**`src/aleph/vm/agent/allocation/reconciler.py`** (line 253): Documented and deliberate, but worth calling out for consumers of the executions list: a stopped VM with a failure record keeps allocation.state == "failed" for as long as it stays stopped, since _forget_settled only drops records for live VMs. A reader of the listing sees state: "stopped" alongside allocation: {state: "failed"}. The distinction (the supervisor's word vs the agent's) is explained in vm-lifecycle.md, so this is just something to watch as consumers appear.

foxpatch-aleph (approved):
This PR correctly fixes the owner-stop race by teaching all three layers of the allocation loop to leave a stopped VM alone: compute_verdict answers a stopped hash 'unchanged' instead of sizing it (a capacity refusal would otherwise evict it from the plan the loop converges on), _start_missing skips it while deliberately preserving any failure record as attempt history, and notify_vm_down no longer wakes the loop for a stop (while still waking for FAILED or a vanished VM). I verified the capacity-accounting claim (an owner stop keeps the registry record, so memory/vCPUs/disk stay committed), that both callers of the changed notify_vm_down signature are updated, that the STOPPING case avoids start_persistent_vm's wait-and-recreate path, and that a stopped VM dropped from the plan is still torn down. All four touched test suites pass locally (45+47+11+19); the ruff and mypy findings on the full scope are pre-existing artifacts of local tool-version skew (project pins ruff==0.4.6 and mypy==1.8.0), with the only in-file mypy error at supervisor.py:281 being pre-existing code untouched by this PR. Tests are thorough and the docs accurately describe the rule, the capacity guarantee, and the v1 divergence. Only cosmetic nits remain.

**`tests/supervisor/test_allocation_verdict.py`** (line 639): Cosmetic: the docstring wasn't re-wrapped after the rename — this line runs to 89 chars while its neighbors sit at ~76. `unchanged, and a node with no room left refuses it. Narrowing has to keep it out of the` could be re-flowed to match the wrapping style of the rest of the file.

**`src/aleph/vm/agent/allocation/reconciler.py`** (line 253): Observation, not a defect: because the failure record is deliberately kept while the VM is stopped, an owner-stopped VM that recently crash-looped (record younger than ALLOCATION_RETRY_MAX_INTERVAL) will render `allocation.state: "failed"` with attempts and a stale next_retry_at for as long as it stays stopped. The reasoning is documented and the tradeoff is sound — clearing the record would reset the crash ladder — but it may be worth a sentence in vm-lifecycle.md noting that a stopped VM can carry a failed allocation block from a recent crash, since the current docs only describe the create-failed case.

PR #1218, opened 2026-09-09
fix(daemon): resume a suspended GPU before the CC probe and cache the answer

odesenfans:
## Summary
The NVIDIA CC-mode probe read BAR0 without taking a runtime-PM reference. vfio-pci runtime-suspends a card nobody has opened, a function in D3hot answers MMIO with all ones, and the low two bits of `0xffffffff` are the devtools encoding, so an idle card was advertised as `devtools` and then refused by the SEV-SNP create gate: a host with working CC hardware would look like a host with none. On top of that the refresh ran on every `GetHostInfo`, which the agent calls for every unauthenticated `/about/capability` request, so any internet client could make the host mmap device memory on every idle NVIDIA card it has, as often as it liked.

The probe now pins a suspended card awake for the read and restores the host's own `power/control` value afterwards, rejects an all-ones answer as unreadable instead of decoding it, and caches each card's answer (successful or not) for 60 seconds. The create gate is untouched: it still reads the card on every create, under the creation lock.

The third commit is the review fixes, listed under Changes.

The second commit closes the matching hole on the other side: a card a guest owns is never probed, so after a restart every adopted VM's cards reported an empty `cc_mode` for the rest of the VM's life. A card held by an SEV-SNP VM needs no read, because the create gate admitted it only after reading CC-on and the mode cannot change while the guest holds the card, so it is now entered as CC-on on the gate's word. Plain passthrough VMs went through no gate and their cards stay unknown.

## Changes
- `gpu_cc.rs`: `hold_runtime_power_on` writes `on` to `power/control` when `power/runtime_status` says the device is not active, waits up to 200 ms for it to come back, and puts the previous value back on drop. A device with no runtime-PM files, or one already active, is not written to at all. `probe_cc_mode` rejects `0xffffffff` as unreadable rather than reporting devtools. `read_bar0_u32` now returns the `io::Error` (annotated with the file it came from) so the caller can build a typed error around it.
- `error.rs`: `DaemonError::GpuRegisterRead { pci_host, source }` and `DaemonError::GpuUnreadable { pci_host }` replace the formatted string on the probe path.
- `service.rs`: `gpu_cc_modes` holds `ProbedCcMode { mode, probed_at }` instead of a bare mode. A card whose last answer is younger than `CC_MODE_TTL` (60 s) is not read again, and a failed probe is cached as "no mode" so an unreadable card is not retried on every request either. `cc_mode_of` is unchanged from a caller's point of view: an entry with no mode advertises nothing.
- `main.rs`: one probe pass on the blocking pool before the socket is bound, so a card that is free at boot has a mode from the first request.
- `service.rs` (third commit): a refresh pass takes a dedicated lock before the world read guard, so a burst of `GetHostInfo` calls costs one pass and the callers behind it read what it wrote. Overlapping passes were not just wasteful: both could read `power/control` before either wrote it, so the second restored `on` and left the card pinned awake with the host's setting lost (reproduced: the file came back `on\no`). The world read guard still covers the probes, since it is what stops a card being read after a concurrent create handed it to a guest, and its doc comment now states the real budget (up to 200 ms per stale suspended card, with tokio's write-preferring RwLock making a mid-pass `CreateVm` wait) instead of claiming microseconds. `main.rs` names the same bound for the pre-bind pass.
- `lifecycle.rs` (third commit): deleting a VM drops its cards' cache entries. The card is free hardware again and its mode can be switched, so the old answer must not be served for the rest of its freshness window.
- `gpu_cc.rs` (third commit): a device whose `power/control` cannot be read or written no longer fails the probe. The resume step is an improvement on the read, not a condition of it, so the register is read as it was before the step existed and an all-ones answer still fails closed.
- `service.rs` (second commit): the refresh enters CC-on for any NVIDIA card attached to a VM whose config resolves as SEV-SNP (`QemuVmConfig::snp()`), without reading the hardware. An existing entry is never overwritten, so a card probed before it was attached keeps its own answer and its own age; once the VM is deleted the card leaves the attached set, its entry is long stale, and the next refresh reads it as usual.

### Review follow-ups, round 3
- `src/lifecycle.rs`: a stop now evicts the VM's cards from the CC-mode
  cache, through the same `forget_cc_modes` helper `delete_tracked_vm`
  uses, on both the persistent-VM and the program stop paths. The seed
  guard alone only stopped a *new* seed; the entry the create gate wrote
  while the guest was live stayed and kept the card advertised CC-on after
  an in-session stop. Test: seed the card the way the gate does, stop the
  VM, `cc_mode_of` is `None`.
- `src/service.rs`: the seed requires `started_at_ns != 0` as well as
  `stopped_at_ns == 0`, so a VM adopted while the systemd bus was
  unreachable, or a Defined entry that never started, is not seeded CC-on.
  Test: a never-started SNP VM's card is not seeded.
- `src/lifecycle.rs`: the create gate's comment now states the write-lock
  cost: the probe resumes a suspended card under the world write lock, up
  to 200 ms per suspended card, during which every world reader stalls.
- `src/gpu_cc.rs`: `hold_runtime_power_on` writes `on` and polls only when
  `runtime_status` is `suspended`, `suspending` or `resuming`; `active`,
  `unsupported`, `disabled` or anything else reads immediately with no
  hold. Test: an `unsupported` card is neither written to nor waited for,
  and its register is still read.
- Known and left as is: `StartVm` after a stop re-seeds the card CC-on
  from the gate's pre-stop reading at the next refresh pass (up to the
  60 s TTL late) without re-probing. A card re-moded while the VM was
  down would be advertised stale until the next create; the guest's own
  boot-time GPU attestation is what catches it for the owner. Making
  StartVm re-probe under the VM lock would close that, and is a follow-up.

### Round 4: two cache windows, the long one a setting
- `src/gpu_cc.rs`, `src/config.rs`: a decoded mode is now served from the
  cache for `ALEPH_VM_GPU_CC_MODE_TTL` seconds (default 3600) and an answer
  with no mode (probe failed, all ones, reserved encoding) for a fixed
  60 s. The mode changes only when an operator runs NVIDIA's tool against
  an idle card, and admission never depends on the cache (the create gate,
  and the start gate in #1240, read the hardware), so the timer only bounds
  inventory staleness; a minute for every answer woke every idle card out
  of D3 once a minute for as long as the node lived, driven by host-info
  polling. The failed-read tier stays short because an operator's mode
  change ends with a card reset, and a sweep landing during it reads all
  ones: cached for an hour, that would hide the card for an hour.
- `src/service.rs`: the pass-level skip is gated on the shortest window,
  so a card cached as unreadable is re-read within the minute even when
  every other card is fresh for an hour; the per-card gate is where the
  long window pays.
- `docs/architecture/confidential.md`: states the two windows and the
  setting.
- Tests: a decoded mode is fresh past a minute and stale past the long
  window; an unreadable answer is stale past 60 s; a pass with one fresh
  decoded card and one stale unreadable card re-reads only the latter; the
  setting parses its default and an override.

## Test plan
- `gpu_cc.rs`: new tests over a fake sysfs tree under a temp dir. An all-ones register is rejected at probe level (the test also pins that the decoder alone would have said `devtools`); a `runtime_status = suspended` card gets `on` written to `power/control` and the previous value restored; an active card is never written to; a card with no runtime-PM files probes as before; a card with no CC architecture is neither written to nor read; the freshness window behaves.
- `service.rs`: two refreshes inside the TTL read each card once, a failed probe is remembered and not retried, and a zero window reads both cards again. The existing attached-set and stale-eviction tests were updated for the new cache value.
- `service.rs` (second commit): a fake world holding one adopted SEV-SNP VM and one adopted plain passthrough VM, each with a card, refreshed with a probe that panics if it is ever called. The confidential VM's card reports CC-on through both `cc_mode_of` and the VM's own `VmInfo.gpus[0].cc_mode`; the passthrough card stays unknown.
- `service.rs` (third commit): two refresh passes run concurrently against a fake sysfs card that never resumes. Asserts one register read and `power/control` back at `auto`. Without the pass lock this fails twice over: two reads, and `power/control` left as `on\no`.
- `lifecycle.rs` (third commit): a VM created with a card and a seeded cache entry, deleted; the card advertises nothing afterwards.
- `gpu_cc.rs` (third commit): a card that claims to be suspended but has no `power/control` still reads its register.
- Checks from `rust/`, re-run after all three commits: `cargo fmt --check` (0), `cargo clippy --locked --all-targets -- -D warnings` (0), `cargo test --locked -p supervisor-daemon` (0, 387 passed). No NVIDIA hardware here, so the register read itself is exercised against a regular file, as it already was.

## Rebased onto dev-2.1 (#1199 + #1200), plus a review fix

#1200 landed its own answer to the same rate-limiting problem: a sweep-level gate in `refresh_cc_modes_with` (`CcSweep` plus `CC_PROBE_TTL`) that skips the whole sweep when the attached set has not changed since the last one. Both gates are kept, and they now share one lifetime, `gpu_cc::CC_MODE_TTL`; the duplicate `CC_PROBE_TTL` constant is gone. The sweep gate is the cheap outer short circuit, the per-card `ProbedCcMode` freshness the inner one that still holds when the sweep runs because some other card changed hands. The two colliding tests were split into `refresh_cc_modes_repeats_a_sweep_only_when_a_card_changed_hands_or_the_ttl_passed` and `refresh_cc_modes_serves_a_fresh_answer_without_reading_the_card` over one shared fixture, the second forcing the sweep so it exercises the per-card gate alone. Every `GpuDevice` literal gained #1200's new `arch` field, and the adopted-SNP test now also asserts the architecture reaches `VmInfo`.

Review fix, as its own commit: the CC-on seed keyed on `entry.config.snp().is_some()`, which also seeded a stopped SNP VM's card. The seed is only sound under a live guest, where nothing can switch the mode the create gate read; a stopped VM's QEMU is gone and the card is idle hardware an operator can re-mode. The predicate now also requires no stop stamped (`entry.times.stopped_at_ns == 0`), the cheap indicator the status mapping itself uses. Such a card stays out of the sweep (the config still claims it) so it advertises nothing until the VM is deleted. New test: `a_stopped_snp_vms_card_is_not_seeded_cc_on`.

Checks re-run after the rebase: `cargo fmt --check` exit 0, `cargo clippy --locked --all-targets -- -D warnings` exit 0 (which type-checks every test). The test suites were not run locally this round (some need root to unwind DHCP servers); CI runs them.

### DIFF
--- a/docs/architecture/confidential.md
+++ b/docs/architecture/confidential.md
@@ -301,9 +301,18 @@ Blackwell, `0x1182cc` on Hopper, bits `[1:0]`), through the card's sysfs
 runs only against cards no VM's world-view entry currently attaches, reads
 the attached set fresh on the blocking task immediately before probing (a
 stale snapshot could otherwise race a concurrent `CreateVm`), and caches
-each card's mode until its attachment state changes; a card a guest owns is
-never read. A probe error or an unrecognized device id leaves the card's
-mode unknown, which advertises nothing.
+each card's answer; a card a guest owns is never read. A probe error or an
+unrecognized device id leaves the card's mode unknown, which advertises
+nothing. The cache is dropped for a card when its attachment state changes
+(a stop or a delete forgets it), and otherwise ages out on two windows: an
+hour for an answer that decoded to a mode, `ALEPH_VM_GPU_CC_MODE_TTL`
+seconds where an operator wants it shorter, and a fixed minute for an
+answer that carries none, so a card read while it was being reset (all
+ones, cached as unreadable) comes back within the minute rather than
+staying hidden for the hour. Reading the register wakes an idle card out
+of runtime suspend, so the windows are also what keeps the unauthenticated
+`/about` path from driving register reads at the request rate; the create
+and start gates never trust the cache and read the card themselves.
 
 **Gate.** `snp_config_slice` (`rust/crates/supervisor-daemon/src/lifecycle.rs`)
 admits a GPU onto an SEV-SNP spec only when the probed mode is exactly `On`
--- a/rust/crates/supervisor-daemon/src/config.rs
+++ b/rust/crates/supervisor-daemon/src/config.rs
@@ -186,6 +186,16 @@ pub struct Settings {
     /// unset bounds the reservation by total RAM minus the per-node headroom.
     /// Only read when `numa_hugepages` is on. Rust-only (no Python oracle).
     pub numa_hugepages_limit_mb: Option<u64>,
+
+    /// ALEPH_VM_GPU_CC_MODE_TTL, in SECONDS, default
+    /// [`crate::gpu_cc::DEFAULT_CC_MODE_TTL_SECS`] (3600). How long a GPU
+    /// confidential-computing mode that was read successfully is served
+    /// from the cache before the card is read again. The read wakes an idle
+    /// card out of runtime suspend, so this is the rate at which host-info
+    /// polling touches idle hardware; shorten it on a node whose cards are
+    /// re-moded often. An answer that carries no mode is held for a fixed
+    /// minute instead, whatever this says. Rust-only (no Python oracle).
+    pub gpu_cc_mode_ttl: u64,
 }
 
 /// conf.py DnsResolver: how DNS_NAMESERVERS is derived when unset.
@@ -374,6 +384,9 @@ impl Settings {
             .get_u32("NUMA_HUGEPAGES_HEADROOM_MB")?
             .unwrap_or(crate::hugepages::DEFAULT_HEADROOM_MB);
         let numa_hugepages_limit_mb = env.get_u64("NUMA_HUGEPAGES_LIMIT_MB")?;
+        let gpu_cc_mode_ttl = env
+            .get_u64("GPU_CC_MODE_TTL")?
+            .unwrap_or(crate::gpu_cc::DEFAULT_CC_MODE_TTL_SECS);
         let dns_resolution = match env.get("DNS_RESOLUTION") {
             None => DnsResolution::Detect,
             Some(value) if value == "detect" => DnsResolution::Detect,
@@ -428,6 +441,7 @@ impl Settings {
             numa_hugepages,
             numa_hugepages_headroom_mb,
             numa_hugepages_limit_mb,
+            gpu_cc_mode_ttl,
         })
     }
 
--- a/rust/crates/supervisor-daemon/src/config.rs
+++ b/rust/crates/supervisor-daemon/src/config.rs
@@ -657,6 +657,27 @@ mod tests {
     }
 
     #[test]
+    fn the_gpu_cc_mode_ttl_defaults_to_an_hour_and_is_tunable() {
+        // The long tier is the only one an operator sets: a node whose
+        // cards are re-moded often can shorten it, and the short tier for
+        // an answer with no mode is fixed in the probe module.
+        assert_eq!(
+            Settings::from_vars(vars(&[])).unwrap().gpu_cc_mode_ttl,
+            crate::gpu_cc::DEFAULT_CC_MODE_TTL_SECS
+        );
+        assert_eq!(
+            Settings::from_vars(vars(&[("ALEPH_VM_GPU_CC_MODE_TTL", "120")]))
+                .unwrap()
+                .gpu_cc_mode_ttl,
+            120
+        );
+        // A non-integer (or negative) value is a hard error, like the
+        // other int settings.
+        assert!(Settings::from_vars(vars(&[("ALEPH_VM_GPU_CC_MODE_TTL", "often")])).is_err());
+        assert!(Settings::from_vars(vars(&[("ALEPH_VM_GPU_CC_MODE_TTL", "-1")])).is_err());
+    }
+
+    #[test]
     fn defaults_mirror_conf_py() {
         let settings = Settings::from_vars(vars(&[])).unwrap();
         assert_eq!(settings.execution_root, PathBuf::from("/var/lib/aleph/vm"));
--- a/rust/crates/supervisor-daemon/src/error.rs
+++ b/rust/crates/supervisor-daemon/src/error.rs
@@ -41,6 +41,19 @@ pub enum DaemonError {
     #[error("GPU probe failed: {0}")]
     GpuProbe(String),
 
+    #[error("cannot read the confidential-computing register of the GPU at {pci_host}: {source}")]
+    GpuRegisterRead {
+        pci_host: String,
+        #[source]
+        source: std::io::Error,
+    },
+
+    #[error(
+        "the GPU at {pci_host} answers 0xffffffff: it is powered down or off the bus, so its \
+         confidential-computing mode cannot be read"
+    )]
+    GpuUnreadable { pci_host: String },
+
     #[error("Device vendor not compatible")]
     IncompatibleGpuVendor,
 
--- a/rust/crates/supervisor-daemon/src/gpu_cc.rs
+++ b/rust/crates/supervisor-daemon/src/gpu_cc.rs
@@ -691,19 +691,44 @@ mod tests {
         );
     }
 
+    /// An answer of `mode` that a probe gave `age` ago.
+    fn aged(mode: Option<CcMode>, age: Duration) -> ProbedCcMode {
+        ProbedCcMode {
+            mode,
+            probed_at: Instant::now()
+                .checked_sub(age)
+                .expect("the process started after the ages used here"),
+        }
+    }
+
     #[test]
-    fn a_probe_answer_is_fresh_only_inside_its_window() {
+    fn a_decoded_mode_outlives_an_answer_that_carries_none() {
+        // The two tiers, at the ages that separate them: a mode is still
+        // served a minute in and only expires at the long window, while an
+        // answer with no mode (a card read during its reset, say) is read
+        // again after the short one, so a blink hides a card for a minute
+        // and not for an hour.
+        let windows = CcCacheWindows::with_mode_ttl(Duration::from_secs(3600));
+        assert_eq!(windows.unreadable, UNREADABLE_CC_MODE_TTL);
+        assert_eq!(windows.shortest(), UNREADABLE_CC_MODE_TTL);
+
         let answer = ProbedCcMode::now(Some(CcMode::On));
-        assert!(
-            answer.is_fresh(CcCacheWindows::with_mode_ttl(Duration::from_secs(
-                DEFAULT_CC_MODE_TTL_SECS
-            )))
-        );
-        assert!(!answer.is_fresh(CcCacheWindows {
+        assert_eq!(answer.mode, Some(CcMode::On));
+        assert!(answer.is_fresh(windows));
+
+        assert!(aged(Some(CcMode::On), Duration::from_secs(61)).is_fresh(windows));
+        assert!(!aged(Some(CcMode::On), Duration::from_secs(3601)).is_fresh(windows));
+        assert!(aged(None, Duration::from_secs(30)).is_fresh(windows));
+        assert!(!aged(None, Duration::from_secs(61)).is_fresh(windows));
+
+        // A zero-length pair expires everything, which is how the refresh
+        // tests force a read.
+        let expired = CcCacheWindows {
             mode: Duration::ZERO,
             unreadable: Duration::ZERO,
-        }));
-        assert_eq!(answer.mode, Some(CcMode::On));
+        };
+        assert!(!ProbedCcMode::now(Some(CcMode::On)).is_fresh(expired));
+        assert!(!ProbedCcMode::now(None).is_fresh(expired));
     }
 
     #[test]
--- a/rust/crates/supervisor-daemon/src/gpu_cc.rs
+++ b/rust/crates/supervisor-daemon/src/gpu_cc.rs
@@ -270,9 +270,9 @@ impl Drop for RuntimePowerHold {
 /// Pin a runtime-suspended card awake for a probe. vfio-pci lets a device
 /// nobody has opened runtime-suspend, and MMIO reads to a function in
 /// D3hot come back as all ones, which the register decoder cannot tell
-/// from a real answer. Returns `None`, having written nothing, when the
-/// device exposes no runtime PM or is already active. Waits up to
-/// `timeout` for the kernel to report it active.
+/// from a real answer. Returns `None`, having written nothing, unless the
+/// kernel reports the device on its way into or out of runtime suspend.
+/// Waits up to `timeout` for the kernel to report it active.
 ///
 /// Never fails the probe: a device whose runtime-PM files cannot be read
 /// or written is read as it is, exactly as it was before there was a
@@ -289,7 +289,15 @@ fn hold_runtime_power_on(device_dir: &Path, timeout: Duration) -> Option<Runtime
             return None;
         }
     };
-    if status.trim() == "active" {
+    // Only a device on its way into or out of D3 is worth holding. An
+    // "active" one is already awake, and a device with no runtime PM says
+    // "unsupported" (some drivers report other words still): it is powered,
+    // it will never report "active", so writing "on" would change the
+    // host's setting for nothing and then burn the whole resume budget
+    // waiting for a transition that cannot come. "resuming" stays in: a
+    // device mid-resume is exactly what the budget is there to wait for.
+    let status = status.trim();
+    if !matches!(status, "suspended" | "suspending" | "resuming") {
         return None;
     }
     let control_path = device_dir.join("power/control");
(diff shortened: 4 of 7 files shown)


foxpatch-aleph (approved):
This PR is correct and well-tested. The runtime-PM hold is properly RAII-scoped (the control write precedes hold construction, so no restore-without-ownership window), the all-ones rejection closes a genuine fail-open bug (0xffffffff decoding as devtools and advertising bogus confidential capacity), and the 60s TTL closes a real resource-exhaustion vector on the unauthenticated capability path. I verified the startup probe runs after build_world_view so adopted VMs' cards are in the world view and skipped, that lock ordering (world guard -> gpu_cc_modes) is consistent in both refresh and the create gate, that the Copy closure used twice with map_err compiles, and that caching mode=None is observationally equivalent to the old remove() for every cc_mode_of caller. The create gate still reads hardware, so up to 60s of staleness in the advertised inventory cannot leak into SEV-SNP admission decisions. Remaining nits are minor: a runtime_status of "unsupported" burns the full 200ms poll before the read proceeds anyway, the resume wait adds bounded latency under the world read guard (and the creation write lock), one error path is untested, and the _resumed binding name hides that it is an RAII guard.

**`rust/crates/supervisor-daemon/src/gpu_cc.rs`** (line 292): Nit: a device whose `runtime_status` is `unsupported` (or otherwise never goes active) pays the full 200 ms poll here even though it will never report active — reading it immediately would be equivalent. Consider breaking out early on `suspended`-by-another-name values like `unsupported`. Non-blocking: the read proceeds either way and fails closed on all ones.

**`rust/crates/supervisor-daemon/src/gpu_cc.rs`** (line 279): Missing test: nothing exercises the `GpuRegisterRead` path where `power/runtime_status` exists but reading/writing `power/control` fails (e.g. a card with no control file, or an unwritable one). A small fixture-only test alongside `a_suspended_card_is_pinned_awake_and_released_afterwards` would pin the NotFound-to-error mapping and the `Err` (as opposed to `Ok(None)`) exit.

**`rust/crates/supervisor-daemon/src/gpu_cc.rs`** (line 333): `_resumed` reads like a resume result, but it is the RAII guard keeping the card awake for the read; `let _hold = ...` would make the lifetime scope clearer at a glance.

**`rust/crates/supervisor-daemon/src/service.rs`** (line 458): Worth being aware of: the new up-to-200 ms resume wait now runs while the world read guard is held here (and under the world write lock / creation lock in the create gate at lifecycle.rs:2097), so a host with several suspended cards that never resume can hold the guard ~200 ms per card per refresh. Bounded and fail-closed, so not blocking — just noting the latency floor it adds for CreateVm writers behind the guard.

foxpatch-aleph (approved):
A careful, well-documented fix for a real bug: a runtime-suspended vfio-pci card answers MMIO with all ones, whose low bits decode as "devtools", so idle cards were advertised as confidential capacity and then refused by the SNP create gate. Resuming the card for the read, rejecting 0xffffffff as unreadable, caching probe answers (including failures) behind a 60s TTL, and serializing refresh passes are all sound; I verified the sleep loop only runs on the blocking pool, that lock ordering (pass lock -> world read -> gpu_cc_modes) is consistent with create/delete (world write -> gpu_cc_modes), that the pre-bind pass runs after world adoption so adopted cards are skipped, and that the create gate still fails closed on unreadable cards. Tests are deterministic and target the right invariants, including a concurrency test that reproduces the power/control corruption the pass lock fixes. The remaining observations are minor edge cases (below), none of which block.

**`rust/crates/supervisor-daemon/src/service.rs`** (line 512): Minor edge case: the seeding predicate is `entry.config.snp().is_some()`, but the justification ("the mode cannot change while the guest holds the card") only holds while the guest is running. A tracked-but-stopped SNP VM still counts as attached, so a card an operator switched while the VM was down gets seeded CC-on. Impact is low — pre-PR behavior was equivalent (the create gate's entry persisted for attached cards too), the card is withheld from available inventory either way, and a fresh create re-reads the hardware — but if a cheap status check exists, it would tighten the claim in the comment.

**`rust/crates/supervisor-daemon/src/gpu_cc.rs`** (line 271): As the test plan states, the runtime-PM resume path has never run against real hardware. If writing "on" to power/control fails to bring some kernel/vfio-pci combination out of D3hot, every idle-card probe fails closed (GpuUnreadable -> cached no-mode -> no CC capability advertised, SNP creates refused). That is the designed failure direction, but it silently disables CC capability on affected hosts, so it is worth validating the write-"on"/poll-runtime_status sequence on a staged card before rolling this out.

**`rust/crates/supervisor-daemon/src/service.rs`** (line 499): Pre-existing gap, unchanged by this PR, but worth knowing now that the probe writes to sysfs: cards owned by failed-reattach (hidden) or untracked VMs are absent from world.entries, so a refresh will read their register under a live guest. In practice the new runtime-PM write never fires there (an open vfio fd keeps runtime_status at "active"), so only the pre-existing register read occurs.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 1327): Only delete_tracked_vm evicts cache entries; the other world-removal paths (reboot at line 1153, create rollbacks at 2890/3081/3324, failed-reattach hiding at 3618) do not. The impact is bounded by the 60s TTL — a seeded entry goes stale and the card is re-probed once it is unattached — and the create-rollback entries reflect the real hardware state the gate just read, so this is fine as-is; just noting the invariant is "evict on the normal delete path, TTL absorbs the rest".

foxpatch-aleph (changes requested):
The core engineering is good: the resume-before-probe step with RAII restore, the all-ones fail-closed rejection, the shared two-tier TTL, and the refresh pass lock are all correct, and the refresh-vs-create-gate overlap is genuinely closed by the world RwLock (the gate probes under the write lock). The blocking issue is that the review-fix commit only stops a stopped SNP VM's card from being seeded; it does not evict the entry the create gate (or an earlier seed) wrote while the VM was live, so a VM stopped in-session keeps advertising its card CC-on until it is deleted, contradicting the newly added invariant and the stated rationale of that commit. Two smaller notes: A test that seeds the cache before stopping would pin the in-session case.

**`rust/crates/supervisor-daemon/src/service.rs`** (line 532): The `stopped_at_ns == 0` guard stops a stopped SNP VM's card from being *seeded*, but does nothing about an entry that already exists from while the VM was live — and that is the common path. CreateVm's gate inserts `ProbedCcMode::now(Some(On))` at lifecycle.rs:2137; StopVm stamps `stopped_at_ns` but never touches the cache; the card remains in `attached` so the sweep never probes it; and `cc_mode_of` ignores `probed_at`. So after an in-session stop the card keeps advertising CC-on in `gpu_inventory_json` and `VmInfo` until the VM is deleted — precisely the scenario the review-fix commit was written to prevent, and the comment at service.rs:524-529 ("a stopped VM's card simply advertises nothing until the VM is deleted") does not hold in that case. Suggest evicting the cache entries for a VM's cards when it stops, symmetric to what `delete_tracked_vm` now does (lifecycle.rs:1321-1331), e.g. next to the `stopped_at_ns` stamping at lifecycle.rs:859. A test that seeds the cache (or goes through the real gate) and then stops the VM would pin this.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 719): Minor, related to the guard at service.rs:532: a VM adopted with the systemd bus unreachable (the `None` arm here) keeps `stopped_at_ns == 0`, so the seed treats "unknown" as live and seeds CC-on even though the guest may be dead and the card re-modable. Requiring `started_at_ns != 0` as well would exclude both this case and never-started Defined entries, while still covering adopted-running VMs (world.rs stamps `started_at_ns` for those).

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line ): Head's up on a latency consequence: the gate probe now goes through `hold_runtime_power_on`, which can block up to 200 ms per suspended card — while the world *write* lock is held (`build_written_config` runs inside `blocking_write()` at line 2816). Every concurrent world reader (GetHostInfo, ListVms, Health) stalls for that long, where before this PR the gate probe was a single mmap read under the same lock. Not a correctness bug — creates are rare and the card was likely idle/suspended anyway — but it may deserve a word in the comment that documents the write-lock cost, mirroring the read-guard budget note in `refresh_cc_modes_with`.

**`rust/crates/supervisor-daemon/src/gpu_cc.rs`** (line 292): Nit: only `active` short-circuits here, so a device reporting `unsupported` or `disabled` in `runtime_status` (i.e. no runtime PM, device already on) still gets `on` written to `power/control` and then burns the full 200 ms poll waiting for an `active` that can never arrive. Harmless (the read proceeds and fails closed if wrong), but holding only when the status is `suspended` would avoid both the pointless write and the wait.

foxpatch-aleph (approved):
Correct and well-tested fix. The all-ones BAR0 answer from a runtime-suspended card is now rejected as unreadable instead of decoding as devtools, the resume-before-read hold is bounded and restores power/control on Drop, and the 60s cache (with failures cached) plus the gpu_cc_refresh pass lock eliminate both the repeated suspend-cost latency and the two-writer power/control corruption. Lock ordering between the refresh pass (gpu_cc_refresh -> world read) and the create gate (creation_lock -> world write) has no inversion. The CC-on seeding predicate (live SNP VMs only, started and never stopped) is enforced at create time and evicted on stop/program-stop/delete; the StartVm-after-stop re-seeding residual is acknowledged and acceptable. Two minor non-blocking notes below, one on test-fixture fidelity and one informational on parent-bridge runtime PM.

**`rust/crates/supervisor-daemon/src/service.rs`** (line 2026): Minor test-fidelity nit: `snp_entry.times.started_at_ns = 0` combined with the `stopped_at_ns` stamp means this fixture models a VM that never started, which is already covered by `a_never_started_snp_vms_card_is_not_seeded_cc_on`. A real stopped VM keeps a nonzero `started_at_ns` (only `stopped_at_ns` becomes nonzero). Keeping `started_at_ns` nonzero and stamping only the stop would make this test exercise the actual "started then stopped" shape of the predicate at line 542 (`started_at_ns != 0 && stopped_at_ns == 0`) rather than overlapping with the never-started case. Non-blocking — the predicate itself is correct.

**`rust/crates/supervisor-daemon/src/gpu_cc.rs`** (line 281): Informational: `hold_runtime_power_on` writes the device's own `power/control`, which keeps the GPU itself in D0 but does not pin a parent bridge that may itself runtime-suspend. In practice the probe is still correct because the 0xffffffff all-ones check fails closed as `GpuUnreadable` rather than decoding a bogus mode, so this is at most a missed optimization, not a correctness issue. No change requested.

foxpatch-aleph (approved):
This is a careful, well-documented fix for a real problem (an idle vfio card reads back all-ones, which decodes as devtools, and the unauthenticated host-info path could drive per-request device mmaps). The concurrency design holds up under scrutiny: the pass lock serializes refresh passes, the world read guard spans the probes, the create gate's probe under the write lock cannot overlap a pass, the stop/delete eviction is correctly ordered against the stopped_at write, and the SNP seed's or_insert_with non-overwrite is safe because validate_spec_gpus rejects already-attached cards and the create gate always overwrites the entry with its own reading. The two-tier cache and its shortest() outer gate are consistent (a per-card unreadable answer can never sit behind a sweep gate longer than its own tier). I traced every interleaving I could construct and found no bug. The findings below are non-blocking: a doc sentence that is ahead of the code (no start gate exists in this tree), a realistic CI flake in the Instant-based test helpers, and a note on TTL=0.

**`docs/architecture/confidential.md`** (line 315): This sentence (and the matching module doc in gpu_cc.rs line 24: "the create and start gates read the hardware themselves") is ahead of the code in this tree: there is no start gate here. snp_config_slice is only reached via build_written_config on the create path; start_vm_execution does not probe, and the PR description's own "Known and left as is" note says a StartVm after a stop re-seeds CC-on from the gate's pre-stop reading without re-probing. Until #1240 lands, a reader of this doc could wrongly conclude that a restart re-verifies the hardware. Consider wording it as "the create gate" (or "the create gate (and the start gate, once #1240 lands)") so the doc matches the code at merge time.

**`rust/crates/supervisor-daemon/src/gpu_cc.rs`** (line 700): These Instant::now().checked_sub(age).expect(...) helpers (also service.rs lines 1891 and 1918) will panic on any host whose monotonic clock is younger than the age being subtracted — 3601 s in a_decoded_mode_outlives_an_answer_that_carries_none and 61 s in a_pass_rereads_the_unreadable_card_and_leaves_the_known_one_alone. On Linux, Instant is the CLOCK_MONOTONIC since-boot clock, so a freshly provisioned CI runner (booted, built, tested inside an hour) will fail these tests with a panic rather than an assertion. A guard such as Instant::now().checked_sub(age).unwrap_or_else(|| Instant::now() - Duration::from_secs(1) ... ) or skipping when the clock is too young would make the suite uptime-independent; at minimum worth knowing these are uptime-sensitive.

**`rust/crates/supervisor-daemon/src/config.rs`** (line 198): Minor: gpu_cc_mode_ttl = 0 (or a very small value) disables the cache, which re-opens the per-request register-read amplification on the unauthenticated host-info path that the windows exist to bound — the very DoS lever the PR closes. Since the field doc already recommends shortening on nodes re-moded often, a documented floor (or an explicit note that 0 turns rate-limiting off) would prevent an operator from silently undoing the protection. Non-blocking: it requires an operator-supplied env var.

PR #1219, opened 2026-09-09
fix(agent): skip a pool whose size or free space cannot be read

odesenfans:
## Summary
Two places measured a storage pool with a call that reported 0 on an `OSError`, and 0 is indistinguishable from a real measurement: a full pool, or a pool with no room to retain anything. Both then deleted retained data on that guess.

The per-pool retention budget is a share of the pool's size (`VOLUME_RETENTION_BUDGET`, `10%` by default), and `parse_budget("10%", 0)` is 0. A zero budget means every directory marked reclaimable on the pool is evicted, so a pool the agent could not measure for a moment (an unmounted disk, a stale handle, an EIO from a failing device) had all of its retained volumes purged, with the owner recorded on the marker (the erase authorization) and the parent-image pins it carried.

`make_room` had the same failure, bounded rather than total: with free space reported as 0 it took the path it takes for a full pool, summed the retained directories, and evicted them oldest-first until it had freed `needed_bytes`. That is retained data deleted for a create that may already have fitted, repeated on the next placement attempt. Its docstring already claimed such a pool is never evicted from; now that is true.

Both now treat an unreadable pool as unknown rather than zero: log a warning, touch nothing there, carry on with the other pools. Enforcement and room making are only ever deletions, so waiting for the pool to be readable again costs a pass, while guessing costs the data. Reap needs no size at all and still drains such a pool, so a node switched from `keep` to `reap` empties as before.

Stacked on #1216.

## Changes
- `src/aleph/vm/storage_pools.py`: `pool_usage_bytes()` returns the pool filesystem's `(total, free)` as a `PoolUsage` named tuple, or `None` when it cannot be read, instead of a zero that looks like a real measurement. It stays silent, since what an unreadable pool means differs per caller; `_pool_free_bytes` keeps its own log line and now builds on it.
- `src/aleph/vm/agent/vm/reconciler.py`:
  - `_pool_total` is replaced by `_retention_budget(pool)`, which returns `None` when the size is unknown (and 0 under `reap`, which needs no size). `_enforce_retention_budget` skips such a pool with a warning naming how many directories it left alone.
  - `make_room` reads the pool through the same helper: it returns 0 without evicting when the entry measurement fails, and breaks out of the eviction loop when the re-check between two evictions fails. `_pool_free` had no other caller and is gone.
- `docs/architecture/storage.md`: the admission section and step 4 of a pass both document the skip and why an unmeasurable pool is a reason to keep the data, not to delete it.

## Test plan
- `tests/supervisor/test_reconciler.py::test_a_pool_whose_size_cannot_be_read_keeps_what_it_retains`: `disk_usage` raises for pool 0 and answers for pool 1, both over budget; pool 0's retained volume survives and pool 1's is still evicted in the same pass. Fails on the parent commit (the volume is gone).
- `::test_reap_still_reclaims_on_a_pool_whose_size_cannot_be_read`: the guard on the other side, `reap` drains an unmeasurable pool. Passes before and after, deliberately.
- `::test_make_room_evicts_nothing_on_a_pool_it_cannot_measure`: fails without the fix (`assert 8192 == 0`, the retained VM was evicted).
- `::test_make_room_stops_when_the_pool_stops_answering`: the pool answers for the entry check and the first eviction, then raises; only the first directory goes. Fails without the fix (`assert 16384 == 8192`, both went).
- `_fake_disk_usage` in that file grew a `total` and an `unreadable` argument so a test can pin a small pool and make one path raise.
- Checks: `pytest tests/supervisor` (1211 passed, 1 skipped, and the pre-existing journald-stub failure in `test_log.py`); `pytest` on `test_reconciler.py`, `test_cache_eviction.py`, `test_storage_pools_room_maker.py` (133 passed) and earlier on `test_storage_cli.py`, `test_agent_capacity.py`, `test_retire.py`, `test_allocation_verdict.py`, `test_allocations_v2.py`; `ruff format --diff`, `isort --check-only --profile black`, `mypy` on the Python files, all clean. No Rust touched.

- `storage_pools.py`: the comment promising a dead or unmounted pool a last chance from the evictor is now accurate, since `make_room` refuses a pool whose free space cannot be read.
- `storage_pools.py`: every eligible pool's free space is measured once per placement attempt instead of twice (once for the roomiest pool, once inside the sort key ordering the eviction targets), so a pool that cannot answer logs its error half as often. Behaviour otherwise identical.
- `tests/supervisor/test_storage_pools_room_maker.py`: a test that pins one reading per pool per attempt; it fails with the double read.

### DIFF
--- a/docs/architecture/storage.md
+++ b/docs/architecture/storage.md
@@ -384,7 +384,12 @@ message. One pass does, in order:
 4. Enforce the retention budget per pool: sum the markers' `size_bytes` and
    evict oldest-first (by `reclaimable_since`) until under
    `VOLUME_RETENTION_BUDGET`. Under `reap` everything marked is given back,
-   which is how a node switched from `keep` to `reap` drains.
+   which is how a node switched from `keep` to `reap` drains. A pool whose
+   size cannot be read is skipped with a warning under `keep`: the budget is
+   a share of that size, so an unreadable pool would come out as zero bytes
+   allowed and evict everything it retains. Staying over budget until the
+   next pass costs a pass; guessing costs the data. `reap` needs no size and
+   still drains such a pool.
 5. Bring the four download caches under `CACHE_BUDGET` (see below). Skipped
    entirely, with an ERROR, when the supervisor could not be listed: what a
    cache holds for a live VM is read from that VM's message, so a live set
--- a/docs/architecture/storage.md
+++ b/docs/architecture/storage.md
@@ -120,7 +120,10 @@ Admission counts retained (reclaimable) bytes as free:
 and `select_pool`, before refusing a placement, calls the room maker the
 agent registered at startup (`storage_pools.set_room_maker`, wired to
 `reconciler.make_room`) so the retained directories on the target pool are
-evicted oldest-first until the create fits. See "Reclamation" below: a
+evicted oldest-first until the create fits. A pool whose free space cannot
+be read is left alone there too: free space is the whole measure of "does
+this create fit", so without it there is no telling a pool that already fits
+the create from one that never will. See "Reclamation" below: a
 retained disk is a cache entry, not usage, so counting it as used would sell
 less capacity than the node actually has. The rule holds for all three disk
 figures the agent produces, and they have to agree: the aggregate
--- a/src/aleph/vm/agent/vm/reconciler.py
+++ b/src/aleph/vm/agent/vm/reconciler.py
@@ -76,7 +76,12 @@ from aleph.vm.conf import settings
 # MOUNT_ROOT is /mnt, where a volume's mount point is {namespace}_{volume name}.
 from aleph.vm.storage import DEVICE_MAPPER_DIRECTORY, MOUNT_ROOT, is_vm_namespace
 from aleph.vm.storage_budget import parse_budget
-from aleph.vm.storage_pools import StoragePool, get_pools, iter_namespace_dirs
+from aleph.vm.storage_pools import (
+    StoragePool,
+    get_pools,
+    iter_namespace_dirs,
+    pool_usage_bytes,
+)
 from aleph.vm.supervisor_interface.abc import Supervisor
 from aleph.vm.utils import create_task_log_exceptions
 
@@ -607,21 +612,29 @@ def _sweep_side_dirs(
             report.side_dirs_removed += 1
 
 
-def _pool_total(pool: StoragePool) -> int:
-    try:
-        return shutil.disk_usage(str(pool.path)).total
-    except OSError:
+def _retention_budget(pool: StoragePool) -> int | None:
+    """Bytes of retained data ``pool`` may hold, or ``None`` when its size
+    cannot be read and the budget therefore cannot be worked out.
+
+    Reap needs no size at all: it retains nothing, so a node switched from
+    keep to reap gives back everything marked, whatever the pool reports.
+    """
+    if settings.VOLUME_RETENTION == "reap":
         return 0
+    usage = pool_usage_bytes(pool)
+    if usage is None:
+        return None
+    return parse_budget(settings.VOLUME_RETENTION_BUDGET, usage.total)
 
 
 def _pool_free(pool: StoragePool) -> int:
-    try:
-        return shutil.disk_usage(str(pool.path)).free
-    except OSError:
+    usage = pool_usage_bytes(pool)
+    if usage is None:
         # Unknown free space: report none, so make_room falls back to its
         # "stop once needed_bytes have been freed" bound instead of looping.
         logger.warning("Volume pool %s not accessible; freeing on the evicted bytes alone", pool.path)
         return 0
+    return usage.free
 
 
 def _reclaimable_on(pool: StoragePool) -> list[tuple[Path, ReclaimableMarker]]:
@@ -695,9 +708,19 @@ def _enforce_retention_budget(
         entries = _reclaimable_on(pool)
         if not entries:
             continue
-        # Under reap nothing is retained: a node switched from keep to reap
-        # gives back everything marked, whatever the budget says.
-        budget = 0 if reap else parse_budget(settings.VOLUME_RETENTION_BUDGET, _pool_total(pool))
+        budget = _retention_budget(pool)
+        if budget is None:
+            # A budget is a share of the pool's size, so a size nobody can
+            # read would come out as zero bytes allowed and take every
+            # marked directory here with it. Enforcement is only ever a
+            # deletion: leaving the pool over budget until it can be
+            # measured again costs a pass, guessing costs the data.
+            logger.warning(
+                "Volume pool %s not accessible; leaving the %d directory(ies) it retains alone this pass",
+                pool.path,
+                len(entries),
+            )
+            continue
         total = sum(marker.size_bytes for _, marker in entries)
         for directory, marker in entries:
             if not reap and total <= budget:
--- a/src/aleph/vm/agent/vm/reconciler.py
+++ b/src/aleph/vm/agent/vm/reconciler.py
@@ -627,16 +627,6 @@ def _retention_budget(pool: StoragePool) -> int | None:
     return parse_budget(settings.VOLUME_RETENTION_BUDGET, usage.total)
 
 
-def _pool_free(pool: StoragePool) -> int:
-    usage = pool_usage_bytes(pool)
-    if usage is None:
-        # Unknown free space: report none, so make_room falls back to its
-        # "stop once needed_bytes have been freed" bound instead of looping.
-        logger.warning("Volume pool %s not accessible; freeing on the evicted bytes alone", pool.path)
-        return 0
-    return usage.free
-
-
 def _reclaimable_on(pool: StoragePool) -> list[tuple[Path, ReclaimableMarker]]:
     """The pool's reclaimable directories, oldest marker first.
 
@@ -796,7 +786,8 @@ def make_room(pool: StoragePool, needed_bytes: int, *, live: Collection[str] | N
     tries). The third bound (``freed >= needed_bytes``) only exists so a
     lying filesystem cannot turn this into a loop over every retained VM on
     the pool; a pool whose free space cannot be read at all is never evicted
-    from.
+    from, on the same rule the retention budget applies to a pool whose size
+    it cannot read: unknown is not zero, and the guess deletes data.
 
     A marker on a directory a live VM owns is a bug (``_is_orphan`` clears
     those on every pass), but this runs on its own, off the create path, so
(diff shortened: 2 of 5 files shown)


foxpatch-aleph (approved):
Correct and well-scoped fix for a genuine data-loss bug: an unreadable pool previously yielded a zero total, hence a zero retention budget, hence eviction of everything the pool retained. Routing the failure through an explicit None (pool_usage_bytes -> PoolUsage | None) and skipping the pool with a warning is the right fail-safe direction, and the reap path correctly bypasses the size read. Behavior of _pool_free and _pool_free_bytes is preserved bit-for-bit for make_room, admission, and capacity accounting. Both new tests are well-constructed (the keep-side test provably fails on the parent code, and the reap-side test guards the opposite direction), and the docs accurately match the new behavior. All 74 reconciler tests and the related suites pass, ruff format and mypy are clean. Only nits: the test helper patches reconciler_module.shutil though the code under test now reads via storage_pools (works only because both reference the same shutil module object), and settings.VOLUME_RETENTION is read twice (reap local plus inside _retention_budget).

**`tests/supervisor/test_reconciler.py`** (line 54): Minor: the reconciler no longer calls shutil.disk_usage itself — the code under test reads it via storage_pools.pool_usage_bytes. This patch only works because reconciler_module.shutil and storage_pools.shutil are the same module object, so patching is process-global. Patching storage_pools.shutil (or shutil.disk_usage directly) would make the helper target the module actually read and be robust if the reconciler later drops its shutil import.

**`src/aleph/vm/agent/vm/reconciler.py`** (line 615): Nit: _retention_budget re-reads settings.VOLUME_RETENTION while _enforce_retention_budget keeps its own reap local from the same setting; harmless since settings are static during a pass, but threading reap through (or reading it once) would keep the two views from ever diverging.

foxpatch-aleph (approved):
This PR fixes a genuine data-loss bug: an OSError while measuring a pool collapsed to 0, which was indistinguishable from a real measurement, so a momentarily unreadable pool either had all retained volumes purged (0 total -> 0 retention budget) or had retained data evicted for a create that may have fit (0 free). Making the failure explicit via None and skipping the pool is the correct fail-safe for code whose only action is deletion, and the reap path still drains unreadable pools since it needs no size. I verified the fix end-to-end: all 76 reconciler tests plus the related storage suites pass; temporarily reverting pool_usage_bytes to the old zero-on-error behavior makes the three fix tests fail while the reap guard passes both ways, exactly as the PR claims; ruff format, isort, and mypy are clean. Two minor non-blocking nits below concern a now-stale comment and log noise.

**`src/aleph/vm/storage_pools.py`** (line 353): Non-blocking: this comment ("a pool that reports no free space at all (dead disk, unmounted) still gets a turn last, since it is where this placement would land if the evictor brings it back") is now partly stale. After this PR, make_room refuses a pool whose free space cannot be read, so offering it here is a guaranteed no-op (plus a warning log) rather than a last chance at bringing the pool back; only the readable-but-full case still fits the comment. Consider rewording or dropping the dead-disk parenthetical.

**`src/aleph/vm/storage_pools.py`** (line 305): Minor observation on log noise: with a dead pool in the candidate list, a single placement attempt now logs this ERROR several times (once in the candidate loop, once per known_free sort-key comparison, once in the re-check) plus make_room's WARNING and _enforce_retention_budget's WARNING. Diagnostic value is fine, but the repetition per placement attempt is worth knowing about; not worth blocking on.

foxpatch-aleph (approved):
Well-reasoned, well-tested safety fix. The bug is real: a pool whose size or free space could not be read was conflated with a zero measurement, causing retained volumes to be purged from the retention-budget path (parse_budget('10%', 0) == 0) and from make_room. Replacing the silent 0 with None and skipping the pool (with a warning) is the correct fail-safe: both paths are deletions, so the cost of skipping is one pass while the cost of guessing is the data. The reap short-circuit is handled correctly, the measure-once refactor in _select_from is behavior-preserving (verified: StoragePool is hashable, the post-eviction re-measure is kept, sort stability and ordering semantics unchanged), and I confirmed empirically that the three new tests fail against a restored pre-PR implementation while the reap test passes both before and after, exactly as claimed. ruff format, isort and mypy are clean on the changed files; the only ruff-check hits are pre-existing PLR091x on functions this PR does not touch.

**`src/aleph/vm/storage_pools.py`** (line 287): Nice factoring of the shared measurement, and the 'deliberately silent' docstring is a good call — the callers' log lines really do differ in meaning. Minor observation, non-blocking: `pools_disk_usage()` below (line 485) still keeps its own `shutil.disk_usage` + OSError handling rather than reusing this; defensible since it also needs `st_dev` inside the same try block, but worth a thought if you ever touch it again.

**`src/aleph/vm/agent/vm/reconciler.py`** (line 828): Pre-existing behavior, just noting: the first iteration re-measures a pool that was measured a few lines up at entry, so a readable pool costs one extra stat per make_room call. Only worth caring about if you ever want to thread the entry measurement through; harmless as-is and it does catch a pool that goes bad between the entry check and the loop.

**`src/aleph/vm/agent/storage_cli.py`** (line 201): Unrelated to this PR's correctness, but this status path still shows the pre-PR pattern (`total = free = 0` on OSError), so an unreadable pool prints a budget of 0B alongside the new reconcile behavior of skipping it. Display-only, so feel free to ignore — or it could reuse `pool_usage_bytes` for consistency one day.

foxpatch-aleph (approved):
This PR fixes a genuine data-loss bug: an unreadable pool (unmounted disk, stale handle, EIO) was indistinguishable from a full one, and both the retention-budget pass and make_room deleted retained data on that guess. The None-propagation is handled correctly at every call site, under reap the pool is still drained as before (pinned by a dedicated test), and _select_from's once-per-attempt measurement is a small behavior-preserving optimization verified by a new test. The four new tests, docs updates, and log messages are clear and accurately worded. All targeted tests pass (137/137); mypy is clean; the 29 wider-suite failures in this sandbox are environmental (PermissionError on /var/cache/aleph, requiring root) and in files this PR does not touch.

**`src/aleph/vm/agent/vm/reconciler.py`** (line 828): Minor: the loop re-measures the pool on its first iteration even though the entry check at line 803 just did, so a make_room that proceeds to the loop costs one extra statfs per call. Not worth restructuring the loop over, just noting it in case you want to hoist the entry measurement.

**`src/aleph/vm/agent/vm/reconciler.py`** (line 829): Nit: because the `remaining is None` check comes before the `freed >= needed_bytes` bound, a pool that goes unreadable after enough bytes were already freed logs "Stopping the room making... can no longer be read" when the job was in fact done and the next bound would have exited silently. Only a potentially misleading log line; the deletion behavior is correct.

**`src/aleph/vm/storage_pools.py`** (line 485): pools_disk_usage() still measures with its own shutil.disk_usage + OSError handling rather than building on pool_usage_bytes. Keeping it separate is defensible since the st_dev dedup needs os.stat in the same try block, but if it ever needs to share the PoolUsage shape, this is the remaining spot.

foxpatch-aleph (approved):
This PR correctly fixes a genuine data-loss hazard: an unreadable pool previously measured as 0 bytes, which made parse_budget("10%", 0) yield a zero retention budget and evict everything retained on that pool, and made make_room evict oldest-first against a phantom full pool. The new pool_usage_bytes() returning None, the _retention_budget() None-propagation, and the two fail-closed checks in make_room are all correctly wired; StoragePool is frozen and hashable so the new free_by_pool dict is safe, insertion order preserves the lowest-index tie-break, and the reap short-circuit keeps the keep-to-reap drain working. Tests cover both new skip paths, the reap guard, the mid-loop stop, and the once-per-attempt measurement, and all pass locally along with the surrounding suites. The docs match the new behavior. Only cosmetic nits remain.

**`src/aleph/vm/agent/vm/reconciler.py`** (line 828): Cosmetic: `remaining = pool_usage_bytes(pool)` runs before the `freed >= needed_bytes` check, so when the freed bound already holds and the pool happens to have gone silent in that instant, the "Stopping the room making" warning fires even though the loop was about to break anyway — one extra stat syscall and a slightly misleading log line. Matching the pre-PR structure, so not blocking.

foxpatch-aleph (approved):
This PR correctly fixes a genuine data-loss bug: a storage pool whose size or free space could not be read was treated as measuring zero, which under keep mode made parse_budget('10%', 0) == 0 and evicted every retained directory on that pool, and made make_room evict retained data for a create that may already have fit. The refactor is clean: pool_usage_bytes returns a (total, free) named tuple or None, _retention_budget propagates the unknown as a skip with a warning (while reap still drains, since it needs no size), make_room refuses to evict from an unmeasurable pool at entry and stops mid-loop if the pool stops answering, and _select_from now measures each pool once per placement attempt (behavior-identical, since the sort key always ran pre-eviction). StoragePool is a frozen dataclass so the new dict keying is valid, mypy/ruff are clean, and all tests pass locally (77 reconciler, 7 room-maker, 206 across the storage/capacity/cache suites). Tests are well-targeted: two fail on the parent commit as claimed, one pins the reap guard, and one enforces the single-read-per-pool change. Docs accurately describe the new fail-closed behavior. Only trivial nits, none blocking.

**`tests/supervisor/test_reconciler.py`** (line 628): Nit (non-blocking): under reap, _retention_budget returns 0 without touching the disk, so the unreadable= path in this test never actually fires during the budget pass — it only guards against a future regression where the reap short-circuit is removed. Consider a brief comment noting the fake is belt-and-braces, since the test would pass without it.

**`src/aleph/vm/agent/vm/reconciler.py`** (line 715): Nit (non-blocking): "%d directory(ies)" reads awkwardly in a log line; something like "leaving %d retained directory(ies)" or rewording to avoid the plural construction would scan a little better. Purely cosmetic.

foxpatch-aleph (approved):
A genuine data-loss bug fixed correctly: an OSError from shutil.disk_usage was indistinguishable from a real zero, so a transiently unreadable pool (unmounted disk, EIO) had all its retained volumes purged by retention enforcement and make_room. Converting the measurement to a PoolUsage | None that callers treat as unknown-and-skip is the right fail-safe design, since both paths are only ever deletions. The reap exemption, the mid-loop re-check in make_room, and the once-per-attempt measurement in _select_from are all correct (StoragePool is a frozen dataclass, so the dict-key cache is safe). Verified locally: the new and existing reconciler/room-maker/cache/storage-cli/capacity/retire/allocation tests pass (the only errors in the sandbox are environmental aiohttp_client fixture issues, unrelated), and ruff format, isort, and mypy are clean. The two comments below are minor and non-blocking.

**`src/aleph/vm/agent/vm/reconciler.py`** (line 627): Minor: when VOLUME_RETENTION_BUDGET is an absolute size (e.g. "50G"), parse_budget ignores total_bytes, so the budget could be enforced even on a pool whose size cannot be read. Skipping is the conservative choice and I would keep it, but the docstring could acknowledge it — an operator on an absolute budget may wonder why an unreadable pool is left over budget.

**`tests/supervisor/test_storage_pools_room_maker.py`** (line 163): Small docstring nit: "and this read every pool twice" is missing its subject — presumably meant "the old code read every pool twice". The sentence as written is a bit hard to parse.

PR #1220, opened 2026-09-09
refactor(tee): share the ECDSA and validity helpers, keep openssl out of the API

odesenfans:
## Summary
A quality pass over `aleph-tee` turned up a real hole, so this PR is a fix
plus the cleanups that exposed it. The Intel collateral issuer chain carries
the root itself in second position, not an intermediate. The code called that
certificate "the intermediate" and verified it under the pinned root's key,
which any self-signed root satisfies against itself, so nothing was ever
compared to the pin. Nor did anything check what the signer was for: feeding
the collateral's own PCK CRL issuer chain to `verify_signer_chain` was
accepted, handing back the Intel SGX PCK Platform CA as a trusted TCB Info and
QE Identity signer. That is a certificate-purpose confusion the chain
arithmetic alone cannot catch.

The rest is the cleanup that made it visible: the raw ECDSA builder existed
three times, the certificate validity window twice, openssl certificates
leaked across the crate's API, and the SEV-SNP chain check read the wall clock
from inside itself.

Stacked on #1212.

## Changes
- **Fix**, `src/tdx/certs.rs`: the collateral issuer chain is described and
  treated as what it is (signer, then the root). The root is pinned there byte
  for byte, exactly as in a PCK chain; the signer is verified under the pinned
  root directly; and its Common Name must name Intel's TCB signing
  certificate. The doc claiming the root is not embedded is replaced.
- `src/pki.rs` (new, crate-private): the pieces both verifiers need. Raw
  `r || s` ECDSA assembly with context on every step, the validity window
  check and its certificate wrapper, the SystemTime to ASN.1 conversion, and
  the two root pin comparisons.
- `src/tdx/verify.rs`, `src/tdx/tcb.rs`, `src/sev_snp/verify.rs`: the three
  copies of the ECDSA builder are gone. The copy in the collateral document
  check used bare `?` on BigNum and EcdsaSig, so a malformed Intel signature
  surfaced with no hint of which document it came from; it now names the
  document.
- `src/sev_snp/verify.rs`: `verify_cert_chain_at` takes the verification time.
  The window check used `Asn1Time::days_from_now(0)` internally, so the SNP
  chain path could not be driven at a chosen instant at all. `verify_cert_chain`
  keeps its signature and passes the wall clock, so production behaviour and
  every caller are unchanged.
- Root pin style: the two styles stay, and the reason is now written down in
  the shared module and at each pin. AMD re-issues an ARK with the same key,
  so SEV-SNP pins the SubjectPublicKeyInfo; pinning the bytes would turn a
  routine re-issue into a fleet-wide outage. Intel publishes one fixed SGX Root
  CA that every genuine chain carries verbatim, so TDX pins the whole
  certificate. No production behaviour changes on either side.
- `verify_tdx_quote_chain`, `evaluate_tcb`, `pinned_intel_root`,
  `verify_pck_chain` and `verify_signer_chain` are `pub(crate)`: they pass
  `openssl::x509::X509` around, and nothing outside the crate calls them (the
  attest agent uses report_data, traits, types, owner_auth, x509 and the
  SEV-SNP backend; the supervisor controller only the SEV-SNP QEMU argument
  generator; the aleph-rs SDK reimplements its own verifier). `verify_tdx_quote`
  stays the public entry point and returns owned data.
- `owner_auth.rs`, `report_data.rs`, `types.rs`: three comments that ended by
  citing a design document or a divergence ledger now say the thing itself.

### Review follow-ups
- `src/tdx/certs.rs`: the TCB signing gate used `contains`, so a subject
  reading "Not the Intel SGX TCB Signing CA" passed it, and it took the first
  of however many Common Names the subject carried. It now requires exactly
  one Common Name, equal to the constant in full.
- `src/tdx/pck_extension.rs`: `parse_pck_platform` is `pub(crate)`. It takes an
  openssl `X509`, the very type this PR keeps out of the crate's API, and
  nothing outside the crate calls it.
- `src/sev_snp/verify.rs`: `verify_cert_chain_at` is `pub(crate)`. Injecting
  the verification time is a testing affordance; the doc link to it from the
  public `verify_cert_chain` becomes plain text so rustdoc stays clean.
- `src/tdx/certs.rs`: the collateral chain checks the validity window of the
  pinned copy of the root rather than the presented one. That is the same
  check, because the two were just compared byte for byte, and a comment now
  says so.
- `nix/golden-measurements.json`: re-seeded with
  `nix/check-golden-measurements.sh --update`. The attest agent binary that
  links aleph-tee is hashed into the measured initrd, so every aleph-tee edit
  in this stack moves all the pinned launch measurements. The file now carries
  five entries: the base rebase picked up `gpuMeasurement` from the
  confidential-GPU runtime flavor.

## Test plan
- `tdx/certs.rs::collateral_signer_must_be_the_tcb_signing_certificate`: the
  PCK Platform CA presented as a collateral signer. Before the fix
  `verify_signer_chain` returned it as a trusted signer; now it is refused on
  its subject.
- `tdx/certs.rs::signer_common_name_must_match_exactly_and_stand_alone`: the
  exact name is accepted, a subject that merely embeds it is refused (this one
  passed under the old `contains` gate), and a subject carrying a second
  Common Name is refused.
- `tdx/certs.rs::collateral_chain_must_end_in_the_pinned_root`: a chain whose
  second certificate is a genuine Intel certificate that is not the root.
- `tdx/certs.rs::genuine_collateral_chain_verifies`: the fixture's real TCB
  Info chain still verifies.
- `sev_snp/verify.rs::test_cert_windows_follow_the_injected_clock`: expired
  and not-yet-valid at the injected instant. Fails if the verifier reads the
  wall clock (checked by reverting the injection).
- `sev_snp/verify.rs::test_verify_cert_chain_accepts_a_same_key_reissued_ark`:
  an ARK carrying the pinned key in a different envelope must verify, which
  pins the SEV-SNP pin style as a deliberate contract.
- `tdx/certs.rs::a_root_reissued_with_the_pinned_key_asks_for_a_refresh`: the
  byte pin has two rejections, and only the forgery one was covered. A root
  built around the pin's own public key in a fresh envelope must be refused
  with the "same public key ... the pin needs refreshing" wording and not with
  the forged-root wording, so that maintenance diagnostic cannot rot. Watched
  fail with the same-key branch disabled.
- `pki.rs`: a real P-256 signature round trips through the raw form, the halves
  are not interchangeable, and an odd-length signature is refused.
- The existing impostor-chain test still rejects, one step earlier and with the
  pin's message; its assertion follows.
- Checks after the rebase onto the #1212 branch, exit codes read bare: in
  `rust/`, `cargo fmt --check` 0,
  `cargo clippy --locked --all-targets -- -D warnings` 0,
  `cargo test --locked -p aleph-tee` 0 (142 passed, 2 ignored; 133 before this
  branch). With `--manifest-path crates/aleph-attest-agent/Cargo.toml`: fmt 0,
  clippy 0. Earlier on this branch, `cargo doc --locked -p aleph-tee --no-deps`
  clean (it warned before) and the attest agent's `cargo test` 0 (49 passed).
  `nix/check-golden-measurements.sh` reports "Golden measurements match" after
  the re-seed.


foxpatch-aleph (approved):
This PR fixes a real vulnerability and the fix is correct: the old verify_signer_chain verified the chain's second certificate (the root itself) under the pinned root's key, a self-signature check any self-signed cert passes, so the pin was never actually enforced; and the signer was only required to be issued by the root, so the PCK Platform CA (genuinely Intel-root-issued) was accepted as a TCB Info/QE Identity signer — I confirmed both from the fixture. The new code byte-pins the root (fixture cert matches the vendored PEM), verifies the signer directly under the pinned root, and requires the TCB signing CN, which both collateral issuer chains in the fixture carry. The refactor into pki.rs is behavior-preserving at every call site (64-byte typed arrays, toolchain 1.90 supports is_multiple_of, no external callers of the newly pub(crate) functions), the SEV-SNP pin-style contract (key for AMD, bytes for Intel) is documented and test-pinned, and the injected-clock refactor enables proper window testing without flakiness. Test coverage for the fix is precise: the Platform-CA-as-signer case now refuses on subject, a chain ending in a non-root genuine Intel cert refuses on the pin, and the genuine chain still verifies.

**`rust/crates/aleph-tee/src/tdx/certs.rs`** (line 221): Minor nit: `cn.contains(TCB_SIGNING_CN)` is a substring match where equality would be stricter. It is not exploitable here — only certs genuinely issued by the byte-pinned Intel root reach this check, and only the TCB signing cert carries that CN — but `cn == TCB_SIGNING_CN` (or comparing against the full subject) would state the intent more precisely and would not silently accept some future Intel CN like "Intel SGX TCB Signing (Staging)". Non-blocking.

**`rust/crates/aleph-tee/src/sev_snp/verify.rs`** (line 846): Minor consistency nit: this test asserts through the wall-clock `verify_cert_chain` while the PR's stated philosophy is that verification time is a parameter; `verify_cert_chain_at` would match the neighbouring `test_cert_windows_follow_the_injected_clock` and remove any dependence on the wall clock. Practically safe (the re-issued ARK is valid ±2h), so non-blocking.

**`rust/crates/aleph-tee/src/pki.rs`** (line 35): Observation: `ecdsa_from_raw`'s even-length check is slightly looser than the old typed `&[u8; 64]` signature — a 66-byte slice would be split at 33 rather than rejected. Every current call site passes a typed `[u8; 64]` (and tcb.rs still pre-checks `len() == 64`), so it is unreachable today; just worth knowing the type-level guarantee was traded for a runtime one. Non-blocking.

foxpatch-aleph (approved):
This PR fixes a genuine security hole in the TDX collateral verification path — the collateral issuer chain carries the pinned Intel root itself in second position, so the old 'intermediate under the pinned root's key' check was tautological and the pin was never enforced, and the PCK Platform CA was accepted as a TCB Info/QE Identity signer. I independently confirmed the chain structure against the fixture, that the CN constants match the real certificates, that all seven visibility restrictions to pub(crate) have no external callers, that SEV-SNP production behavior is unchanged by the clock injection, and that the golden-measurement re-seed is expected. Tests are thorough: the purpose-confusion regression test, the exact-CN gate, the pin-must-be-the-root test, the injected-clock window test, and the same-key ARK re-issue contract test all directly pin the new behavior. The shared pki.rs module is a clean factoring with accurate documentation of the asymmetric pin rationale. Only trivial nits remain.

**`rust/crates/aleph-tee/src/pki.rs`** (line 77): Non-blocking: the "same public key but a different certificate, so the pin needs refreshing" diagnostic branch has no test coverage — the TDX impostor test exercises the different-key branch. A test presenting an Intel root re-issued with the same key would cover both this branch and the documented maintenance story.

**`rust/crates/aleph-tee/src/tdx/certs.rs`** (line 207): Minor observation: the window of the pinned root (rather than the presented root) is checked here. This is exactly equivalent since check_pinned_root has already established byte identity, but a one-line comment noting that equivalence would save the reader the double take.

foxpatch-aleph (approved):
A genuine security fix plus a clean refactor. The TDX collateral signer chain is now described and treated as what it actually is — signer, then the root — with the root pinned byte-for-byte exactly as in a PCK chain and the signer required to be Intel's TCB signing certificate by exact single-CN match. I independently verified the premise against the fixture: the TCB Info and QE Identity chains carry [TCB Signing, SGX Root CA] while the PCK CRL chain carries the Platform CA, which the old code would have accepted as a TCB Info signer. The shared pki.rs helpers are behavior-preserving (fixed 64-byte quote arrays and the retained explicit length check mean the lenient &[u8] signature loosens nothing; the SEV-SNP wall-clock entry point keeps its signature), all pub(crate) narrowings have no external callers, and the new tests directly demonstrate both the old accept of the Platform CA and the exact-CN gate. Minor non-blocking notes below; the golden-measurement re-seed rests on CI, as no cargo or nix toolchain was available in this environment to execute the suite locally.

**`rust/crates/aleph-tee/src/tdx/certs.rs`** (line 180): The exact 2-cert requirement fails closed if Intel ever inserts an intermediate between the TCB signing cert and the root — a service outage rather than a security hole, which is the right direction, but it is stricter than a generic chain walk. Worth knowing it is a deliberate contract; the module and PR docs cover this.

**`rust/crates/aleph-tee/src/tdx/certs.rs`** (line 220): check_signer_identity gates on the Common Name only, not O or other subject attributes. That is fine as defense-in-depth since the signer must also verify under the pinned root's key (only Intel can mint such a cert), and the CN is precisely what distinguishes certificate purpose — but if Intel ever issues two root-signed certs with the same CN, this gate would not separate them.

**`rust/crates/aleph-tee/src/pki.rs`** (line 34): ecdsa_from_raw accepts any non-empty even length, and openssl's from_private_components does not range-check r/s, so a malformed short signature (e.g. r = 0) constructs successfully and fails later at verify time. Every current caller enforces 64 bytes up front (fixed [u8; 64] arrays in quote.rs, the explicit length check retained in tcb.rs), so nothing is loosened in practice — just something to keep in mind if a new caller reaches for this helper without its own length check.

**`rust/crates/aleph-tee/src/pki.rs`** (line 77): Nice touch: distinguishing the same-key/different-envelope rejection ("pin needs refreshing") from the forged-root case in check_pinned_root. If key extraction itself fails, the message falls through to "possible forged or cache-poisoned root", which is slightly misleading for a parse error, but it is an unreachable-in-practice diagnostic path.

**`rust/crates/aleph-tee/src/sev_snp/verify.rs`** (line 203): verify_cert_chain reading the wall clock is the one deviation from the "verification time is a parameter everywhere" claim in pki.rs's asn1_now doc, and it is consistently documented as the production entry point delegating to verify_cert_chain_at. No action needed; just noting the two docs are in tension and both already explain themselves.

**`nix/golden-measurements.json`** (line 1): The re-seed is consistent with the head commit, but it cannot be validated without a nix build (no toolchain available in this review environment) — CI must run nix/check-golden-measurements.sh before merge, since a stale seed here would mask tampering with the measured agent while looking green.

foxpatch-aleph (approved):
This PR fixes a real security hole (any Intel-root-issued certificate — e.g. the PCK Platform CA — was accepted as a trusted TCB Info/QE Identity signer, and the presented 'intermediate' was never compared to the pin) and ships the cleanup that made it visible. I verified the fix's premise against the real collateral fixture (all issuer chains carry [signer, root], both TCB Info and QE Identity signed by the same 'Intel SGX TCB Signing' cert), confirmed the pub(crate) narrowing breaks no external callers (aleph-attest-agent and supervisor-controller touch none of the newly private items), confirmed every error-string assertion matches the new messages, and confirmed the shared ecdsa_from_raw split is behaviorally identical for all existing call sites (quote fields are fixed [u8; 64]; tcb.rs pre-checks 64). The new adversarial tests are well-designed, especially the purpose-confusion proof using genuine Intel certificates and the same-key re-issued root diagnostic test. The golden-measurement re-seed is value-only and its mechanism is documented; I could not independently rebuild the measurements without nix, relying on the author's stated checker run. No correctness, security, or coverage issues found.

**`rust/crates/aleph-tee/src/pki.rs`** (line 34): Non-blocking observation: the old TDX helper took `&[u8; 64]`, so the 32/32 split was a compile-time guarantee; `ecdsa_from_raw` now splits at whatever midpoint it is given, making the width a caller-owned contract. All current call sites are safe (quote fields are fixed `[u8; 64]` arrays; `tcb.rs` rejects `len() != 64` first), and the midpoint split is documented — but a future P-384 call site routing through `ecdsa_from_raw` with a 48-byte slice would get a silent 24/24 split instead of an error. The `ecdsa_from_components` path is the right one for variable widths (as SEV-SNP uses it), so just worth keeping in mind if new callers appear.

foxpatch-aleph (approved):
This PR fixes a genuine vulnerability in the TDX collateral signer verification: the old code verified the presented "intermediate" (actually the root itself) under its own key, so any self-signed root passed and nothing was ever compared to the pin, letting an attacker fabricate TCB Info/QE Identity collateral with their own keys; the new code byte-pins the presented root and requires the signer's subject to be exactly the TCB signing Common Name. I verified against the fixture that the issuer chains really do carry the root in second position and that the PCK Platform CA chain is accepted by the old arithmetic, confirming both the bug and the fix. The shared pki.rs helpers preserve the previous semantics exactly (SPKI pin for AMD, byte pin for Intel), the clock injection into verify_cert_chain keeps the public entry point's wall-clock behavior unchanged, and the pub(crate) demotions are safe — no caller outside the crate touches any demoted function. Tests are precise and adversarial (same-key re-issue, exact-CN, injected-clock, reissued-root diagnostics), and the documentation is exemplary. I could not run cargo or nix in this environment, so the reported green test runs and the re-seeded golden measurements should be confirmed by CI.

**`rust/crates/aleph-tee/src/pki.rs`** (line 77): Minor, non-blocking: the same-key diagnostic branch is attacker-triggerable — the pinned public key is public, so anyone can build a fresh self-signed certificate carrying it and make the verifier print "the pin needs refreshing" for a submitted chain. It is cosmetic (the rejection still happens and the message opens with "is not {pin_label}"), but an operator triaging logs from an untrusted input path should not read this branch as evidence the genuine vendor re-issued its root. Consider it awareness rather than a request for change.

**`nix/golden-measurements.json`** (line 1): The re-seeded hashes cannot be verified without running the nix build, which this review environment lacks; CI must run nix/check-golden-measurements.sh and confirm "Golden measurements match" before merge, since a mismatch here means the measured boot chain drifted. The addition of the gpuMeasurement entry alongside the base rebase is consistent with the flake's five outputs.

PR #1221, opened 2026-09-09
fix(agent): hold the GPU lock inside the collection, not in the handler

odesenfans:
## Summary
The GPU evidence route bound its serialization guard in the request handler and then ran the collector through `web::block`, which cannot be cancelled. When actix dropped the handler future because the client disconnected, the guard was released while `nvattest` kept talking to the guest driver for up to its full 60 second timeout, so a peer that opened `/.well-known/attestation/gpu` and hung up repeatedly started one concurrent SPDM exchange per attempt, bounded only by the blocking pool. The route now takes an owned guard (same `lock_wait` timeout, so a queued caller still gets 503 with `Retry-After`) and moves it into the blocking closure, so the GPU is released when the collection actually ends rather than when the client goes away.

Stacked on #1218.

## Changes
- `rust/crates/aleph-attest-agent/src/proxy.rs`: `GpuState::lock` becomes `Arc<tokio::sync::Mutex<()>>` so the handler can acquire an `OwnedMutexGuard` and move it into the `web::block` closure; the guard now lives exactly as long as `collect()`. `GpuAttestationResponse` borrows the client nonce and the boot claims instead of cloning the whole claims document on every response (they never change after startup).
- `rust/crates/aleph-attest-agent/src/main.rs`: construct the lock in its `Arc`.
- `nix/golden-measurements.json`: re-seeded, since the agent binary is baked into the measured initrd.

## Test plan
- `rust/crates/aleph-attest-agent/src/proxy.rs`: new `a_dropped_client_keeps_the_gpu_taken_until_the_collection_ends`. A gated collector parks inside `collect()` and records how many callers are inside at once; the first request is polled until its collection starts and its handler future is then dropped (what actix does on client disconnect); the second caller must be answered 503 "gpu attestation busy" and must never enter the collector; the collector's concurrency high-water mark stays at 1, and the route serves 200 again once the abandoned collection ends. Against the old code the collector was entered twice at once (peak 2).
- `cargo fmt --check --manifest-path crates/aleph-attest-agent/Cargo.toml` (0), `cargo clippy --locked --all-targets --manifest-path crates/aleph-attest-agent/Cargo.toml -- -D warnings` (0), `cargo test --locked --manifest-path crates/aleph-attest-agent/Cargo.toml` (50 passed).
- `nix/check-golden-measurements.sh` verifies clean after the re-seed.
- The agent's `Cargo.lock` is untouched: no dependency change.

## Rebased onto dev-2.1 (#1199 + #1200), plus review fixes

The proxy change replayed clean. The stale golden re-seed commit was dropped during the rebase and a fresh one produced at this branch's tip, against the base that now carries #1199's `gpuMeasurement`: all five values, from `nix/check-golden-measurements.sh --update`, then verified by a plain run of the same script, which prints "Golden measurements match."

Review fixes, as one commit:
- `Retry-After` was hard-coded to 10, which happened to equal the default wait and would go stale the moment a deployment tuned it. It now comes from the state's own `lock_wait` through `retry_after_secs`: the caller has just spent that long queueing without getting in, so the exchange ahead of it is wedged rather than slow, and one more wait is the honest estimate. Rounded up to a whole second (the header has no finer unit) and never zero. New test `the_busy_answer_advertises_the_configured_wait` reads the header against a three second wait and pins the rounding edges.
- The disconnect test polled two conditions in unbounded loops (the first request reaching the collector, the abandoned collection leaving it). A regression in either would have hung the suite instead of failing it. Both now stop at the gate's own cap and say what did not happen.

Checks re-run after the rebase: `cargo fmt --check` and `cargo clippy --locked --all-targets -- -D warnings` exit 0 in both `rust/` and `--manifest-path crates/aleph-attest-agent/Cargo.toml`. Test suites left to CI this round.


foxpatch-aleph (approved):
Correct and well-targeted fix for a real concurrency bug: holding the tokio mutex guard in the handler future meant actix dropping the handler on client disconnect released the lock while the uncancellable web::block collector was still mid-SPDM-exchange, letting trivial connect-and-hang-up requests start concurrent exchanges against the GPU driver. Moving to lock_owned() with the guard transferred into the blocking closure ties lock release to the actual end of collect(), and the details are all right: timeout cancellation of a queued lock_owned() future dequeues the waiter cleanly, the let _serialized idiom keeps the guard alive for the closure body (and releases it on unwind), queued callers never consume blocking threads since the lock is acquired before entering the pool, and the borrowed GpuAttestationResponse fields outlive serialization with no API impact on the binary-only crate. The regression test is thorough and fails deterministically against the old code, with GATE_CAP preventing parked-thread leaks on failure, and the golden measurements are re-seeded per the crate's documented policy. Two tiny non-blocking nits below; could not run cargo in this environment, but static analysis fully supports the PR's claims.

**`rust/crates/aleph-attest-agent/src/proxy.rs`** (line 168): Nit (pre-existing, kept as-is by this diff): "Retry-After" is hardcoded to 10 while the actual wait bound is GPU_LOCK_WAIT (also 10s today). Deriving the header from gpu.lock_wait (e.g. Retry-After: secs) would keep the two from drifting apart if the constant changes.

**`rust/crates/aleph-attest-agent/src/proxy.rs`** (line 553): Minor robustness thought for the new test: the poll loop at lines 553-555 has no overall deadline, so a hypothetical regression that made the first request fail before entering collect() would hang the test until the CI timeout instead of failing with a message. I verified no current path can do that (gpu is Some, "00" is valid hex, the lock is free), so this is purely hypothetical — a bounded wrapper like the GATE_CAP comment already documents for the blocking side would just make a failure mode louder.

foxpatch-aleph (approved):
The fix is correct and well-executed: moving the OwnedMutexGuard into the web::block closure ties the lock's lifetime to the collection rather than the handler future, and since spawn_blocking closures run to completion even when the JoinHandle-like future is dropped on client disconnect, the guard is released exactly when collect() ends. The timeout-wrapped lock_owned() is cancellation-safe on the pinned tokio, the borrowed response struct is sound (query and state outlive the immediate serialization, JSON unchanged), and retry_after_secs correctly derives and rounds the Retry-After from the configured lock_wait. Both new tests are properly bounded (GATE_CAP caps every wait, and a regression fails with a message instead of hanging the suite) and the disconnect test demonstrably distinguishes old from new behavior via the peak-concurrency counter. The golden re-seed follows the repo's documented procedure, and rewriting all five entries is expected since --update rebuilds the whole file and the agent binary is baked into every measured initrd; I could not re-run nix or cargo in this environment, so I take the PR's stated verification of the re-seed and the passing suite.

**`rust/crates/aleph-attest-agent/src/proxy.rs`** (line 193): Nit: the moved guard reads a bit oddly as `_serialized`; a name like `_guard` (or `_serialized_guard`) would signal at a glance that the binding exists purely to pin the lock for the duration of `collect()`. The comment above already explains the intent, so feel free to leave as-is.

**`rust/crates/aleph-attest-agent/src/proxy.rs`** (line 191): Worth being aware of the availability trade-off this fix makes explicit (and which the comment gestures at): a client that disconnects mid-collection now leaves the GPU held for the remainder of that collection, so the route can serve 503 for up to COLLECT_TIMEOUT (60 s) after a disconnect, where the old code released immediately. That is the right call — a wedged exchange must not become a concurrent SPDM session — but it does mean a disconnect storm holds the route busy one collection at a time rather than freeing it early. A comment line in the struct doc could spell out this DoS surface; optional.

foxpatch-aleph (approved):
Correct and well-targeted fix. Moving the owned mutex guard into the web::block closure is the right repair: the blocking task is detached when the handler future is dropped, so the GPU is now held exactly for the life of collect() instead of the life of the request, eliminating the concurrent-SPDM-exchange DoS vector. I verified statically that the timeout-wrapped lock_owned is cancellation-safe, that no await point exists between guard acquisition and web::block, that panics/failures inside collect() still release the guard, and that the borrowed GpuAttestationResponse (client_nonce/boot_claims) cannot outlive its referents. The new tests are thorough and race-free under my tracing (bounded polling, capped gate, correct condvar ordering), and the retry_after_secs derivation is correct at every asserted edge. The golden re-seed is consistent with the agent binary being a measured initrd file, though I could not re-run the nix check locally (no toolchain available) and rely on CI for it. Two non-blocking notes below.

**`rust/crates/aleph-attest-agent/src/proxy.rs`** (line 192): Trade-off worth stating in the PR, not a blocker: a client that opens the route and hangs up still holds the GPU for up to COLLECT_TIMEOUT (60 s in gpu.rs) per abandoned collection, and can do that back-to-back, keeping the route at 503 for everyone else. That is inherent to running an uncancellable child collector and strictly better than the previous behavior (unbounded concurrent exchanges), and the fail-fast queue keeps worker threads from piling up, so this is acceptable as-is.

**`rust/crates/aleph-attest-agent/src/proxy.rs`** (line 504): Nit: this test sleeps the full 3-second lock_wait before the 503 arrives (the handler only answers after the timeout elapses). Using lock_wait: Duration::from_secs(1) and expecting Retry-After of 1 would prove the same end-to-end property 2 seconds faster; the fractional edges are already covered by the retry_after_secs unit asserts just below.

PR #1222, opened 2026-09-09
fix(agent): serialise the starts of one VM so one does not delete the other

odesenfans:
## Summary
Every start path (the allocation reconciler, the v1 `/control/allocations` handler, `notify_allocation`) runs the same read, record, download, create sequence, and the download in the middle of it lasts seconds. Nothing serialised that per hash, so a push landing on two paths at once left both callers reading "this node does not have this VM", both recording it and both downloading into the same directory. The supervisor refused the second create with `VmAlreadyExistsError`, the create path read that refusal as its own failure and retired the VM, and the retire deleted the VM the first caller had just brought up, forgot its record and, on fresh disks, purged its volumes.

`start_persistent_vm` now takes a per-hash lock around the whole sequence, and a create refused because the VM already exists returns instead of retiring.

Stacked on #1217.

## Changes
- `src/aleph/vm/agent/create_lock.py` (new): `vm_create_lock(namespace)`, one `asyncio.Lock` per hash, created on demand and dropped once nobody holds or waits for it. Starts of different VMs still run in parallel. It is deliberately not the storage reconciler's `creating()`, which is a refcount because a reconcile pass must count overlapping spans; this one excludes them.
- `src/aleph/vm/agent/run.py`: `start_persistent_vm` runs its whole body under that lock, placed inside the function rather than at the call sites so a new caller cannot forget it. The three create branches in `create_vm_execution` (program, instance, V-PROGRAM) catch `VmAlreadyExistsError` before the generic teardown branch, log it and return: that covers the create paths the lock does not reach (an operator recreate, a migration import racing a scheduler push) and makes the destructive outcome impossible rather than merely unreachable.

## Test plan
- `tests/supervisor/test_run_create_lock.py` (new, 6 tests): two concurrent `start_persistent_vm` calls on one hash create it once, delete nothing and raise nothing, and leave no lock behind; the second start does not even read the supervisor until the first is done; a create whose `create_vm` raises `VmAlreadyExistsError` neither deletes the VM nor purges its volumes, on all three create paths. All five fail on the base commit, the first one on `deleted == []` (the winner's VM was deleted). A sixth pins the release itself: a start whose `create_vm` raises, then a second start of the same hash under a timeout, since a lock left held by a raising body would make every later start of that hash wait forever rather than fail an assert.
- `ruff format --diff`, `isort --check-only --profile black` on the changed files and `mypy src/aleph/vm/ tests/` from the worktree root: all exit 0. The test suites are left to CI on this round.

### DIFF
--- a/src/aleph/vm/agent/create_lock.py
+++ b/src/aleph/vm/agent/create_lock.py
@@ -1,0 +1,64 @@
+"""One lock per VM hash, held across a whole create.
+
+A VM is started from several places at once: the allocation reconciler, the v1
+``/control/allocations`` handler and the single-VM ``notify_allocation``
+endpoint all run the same read, record, download, create sequence, and the
+download in the middle of it lasts seconds. With nothing serialising them, two
+callers read "this node does not have this VM", both record it, both download
+into the same directory and both create it. The supervisor refuses the second
+create, the create path reads that refusal as its own failure and retires the
+VM, and the retire deletes the VM the first caller had just brought up,
+forgets its record and, when the disks were fresh, purges its volumes.
+
+The lock is per hash, so starts of different VMs still run in parallel. It is
+created on demand and dropped once nobody holds or waits for it, so a node
+that has started thousands of VMs over its life keeps no lock for each.
+
+This is a different thing from ``creating()`` in the storage reconciler, which
+is a refcount on purpose: it tells a reconcile pass that a directory is being
+built and must count every span, including two that overlap. This one is
+mutual exclusion between the creates themselves.
+"""
+
+from __future__ import annotations
+
+import asyncio
+from collections.abc import AsyncIterator
+from contextlib import asynccontextmanager
+
+# Hash to the lock serialising its creates, and to the number of callers
+# holding or waiting for it. The count is what says when the lock can be
+# dropped: the last one out removes it.
+_locks: dict[str, asyncio.Lock] = {}
+_waiting: dict[str, int] = {}
+
+
+@asynccontextmanager
+async def vm_create_lock(namespace: str) -> AsyncIterator[None]:
+    """Serialise the creates of one VM hash.
+
+    Wraps the whole read-record-create sequence, not just the create call: it
+    is the window between reading the supervisor and creating the VM that lets
+    a second caller decide the VM is missing.
+    """
+    lock = _locks.get(namespace)
+    if lock is None:
+        # There is no await between the miss and the insert, so the event loop
+        # cannot interleave here and two callers cannot end up holding two
+        # different locks for one hash.
+        lock = asyncio.Lock()
+        _locks[namespace] = lock
+    # Counted before the acquire: a caller queued behind the lock has to keep
+    # it alive, or the holder's release would drop it and the next caller
+    # would build a second one and walk straight in.
+    _waiting[namespace] = _waiting.get(namespace, 0) + 1
+    try:
+        async with lock:
+            yield
+    finally:
+        remaining = _waiting[namespace] - 1
+        if remaining:
+            _waiting[namespace] = remaining
+        else:
+            del _waiting[namespace]
+            del _locks[namespace]
--- a/src/aleph/vm/agent/run.py
+++ b/src/aleph/vm/agent/run.py
@@ -24,6 +24,7 @@ from multidict import CIMultiDict
 
 from aleph.vm.agent.aggregate import get_user_settings
 from aleph.vm.agent.capacity import CapacityManager, requested_gpu_ids
+from aleph.vm.agent.create_lock import vm_create_lock
 from aleph.vm.agent.expiry import ExpiryManager
 from aleph.vm.agent.snp_instance_launch import (
     build_snp_instance_spec,
@@ -48,6 +49,7 @@ from aleph.vm.supervisor_interface.abc import Supervisor
 from aleph.vm.supervisor_interface.errors import (
     FileTooLargeError,
     ResourceDownloadError,
+    VmAlreadyExistsError,
     VmNotFoundError,
     VmSetupError,
 )
@@ -463,6 +465,19 @@ async def _retire_after_create_failure(
         logger.exception("Teardown of half-started %s %s failed", what, vm_hash)
 
 
+def _log_lost_create_race(vm_hash: ItemHash, what: str) -> None:
+    """Say that this create found the VM already there.
+
+    The supervisor refuses a create for an id it already holds, which means
+    another start path built this VM while this one was downloading. That VM
+    is the live one: this create has nothing left to do, and must not treat
+    the refusal as its own failure. The teardown that would follow deletes
+    the VM the other caller just brought up, drops its record and, when the
+    disks were fresh, purges its volumes.
+    """
+    logger.info("%s %s already exists: another start built it, leaving it alone", what, vm_hash)
+
+
 async def create_vm_execution(
     vm_hash: ItemHash,
     *,
@@ -515,6 +530,9 @@ async def create_vm_execution(
                 spec, _resources = await build_program_create_vm_spec(vm_hash, content)
                 info = await supervisor.create_vm(spec)
                 await _wait_until_running(supervisor, info.vm_id)
+            except VmAlreadyExistsError:
+                _log_lost_create_race(vm_hash, "Program VM")
+                return None
             except Exception:
                 # Admission, build, create or readiness failed: retire the
                 # record and whatever was started, but never let a teardown
@@ -579,6 +597,9 @@ async def create_vm_execution(
                         resolved_gpus = await capacity.resolve_gpus(requested_gpus, owner=content.address)
                         spec = replace(spec, gpus=resolved_gpus)
                 info = await supervisor.create_vm(spec)
+            except VmAlreadyExistsError:
+                _log_lost_create_race(vm_hash, "Instance")
+                return None
             except Exception:
                 # build or create failed: retire the early record so a failed
                 # create never leaves a dangling owner-identity entry behind (a
@@ -652,6 +673,9 @@ async def create_vm_execution(
                     )
                     spec = replace(spec, gpus=resolved)
                 info = await supervisor.create_vm(spec)
+            except VmAlreadyExistsError:
+                _log_lost_create_race(vm_hash, "V-PROGRAM")
+                return None
             except Exception:
                 # build, GPU resolution or create failed: retire the early
                 # record, and drop any bundle build_vprogram_spec may have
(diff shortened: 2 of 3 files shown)


foxpatch-aleph (approved):
Well-executed fix for a genuine race: the per-hash lock correctly serialises the whole read-record-download-create sequence with safe lazy creation and last-one-out cleanup (the _waiting counter incremented before acquire is the key subtlety, and it is handled right), and catching VmAlreadyExistsError before the generic teardown branch in all three create paths makes the destructive outcome impossible rather than merely unreachable, also covering the operator reboot/reinstall paths that don't take the lock. The lock has no re-entrancy or deadlock risk (single acquisition, reconciler semaphore always acquired before it), the docstrings accurately explain the design and its distinction from the storage reconciler's creating() refcount, and the five new tests pin both halves of the fix across program, instance and V-PROGRAM paths — all pass. Remaining observations are non-blocking nits.

**`src/aleph/vm/agent/create_lock.py`** (line 32): Non-blocking: the state is split across two parallel dicts keyed identically. A single dict mapping namespace -> (lock, count) would keep the pair atomic and make the increment/cleanup invariants harder to break. The split does mirror the existing creating() pattern in reconciler.py, so it is defensible as-is.

**`src/aleph/vm/agent/run.py`** (line 600): Worth being aware: this except block also wraps capacity.check_message, build_create_vm_spec, build_snp_instance_spec and GPU resolution — not just create_vm — so any future VmAlreadyExistsError raised by those phases would be silently treated as "another start built it, leave it alone". Today those raise download/setup errors, so the semantics hold, but if one of them ever legitimately raises VmAlreadyExistsError this handler would mask it.

**`tests/supervisor/test_run_create_lock.py`** (line 160): Nice-to-have: consider adding a test that the lock is released when the body raises — e.g. one start whose create_vm throws, then a second start that must proceed. The finally guarantees it, but a regression there would deadlock every subsequent start of that hash, which seems worth pinning. Relatedly, line 171's assert on create_lock._locks couples the test to private module state (acceptable for a leak test, just noting it).

foxpatch-aleph (approved):
Correct fix for a genuinely destructive race: two start paths both reading 'VM absent', both downloading and both creating, with the loser's teardown deleting the winner's live VM. The per-hash asyncio.Lock is sound for this single-process, single-loop agent (the no-await get-or-create and the release/decrement adjacency are both airtight, and the _waiting refcount correctly keeps a lock alive while a caller queues behind it); placing it inside start_persistent_vm rather than at call sites covers all three scheduled start paths, and the new VmAlreadyExistsError branches correctly defang the create paths the lock does not reach (operator recreate, migration import, on-demand program VM), where the generic except would have retired a VM another path just built. The loser skipping persist_record is safe since the winner persists the identical record. Tests are well targeted: I ran them and 5/6 pass; the vprogram parametrization fails only in this review sandbox due to a stale aleph-message (1.2.1.dev instead of the pinned 1.5, confirmed by test_vprogram.py failing identically on the base code) — CI with pinned deps should be consulted for it. ruff format and isort verified clean. Only nits remain: no test pins the per-hash (vs global) nature of the lock, the two parallel dicts in create_lock.py duplicate a keying that must stay in sync, and _log_lost_create_race is a thin wrapper around one logger.info call.

**`tests/supervisor/test_run_create_lock.py`** (line 153): All six tests use a single hash, so nothing pins that the lock is per-hash rather than global: a regression that keyed every lock on one constant (or namespace="") would pass this whole file while serialising all starts node-wide. A two-hash variant asserting both creates interleave (or simply that two different hashes can hold the lock at once) would close that gap.

**`src/aleph/vm/agent/create_lock.py`** (line 32): Two module-level dicts keyed by the same namespace must stay in sync by hand; a single dict mapping namespace -> (lock, waiters) would make that invariant local and drop the double dict lookup in the finally. Non-blocking — the current code is correct.

**`src/aleph/vm/agent/run.py`** (line 468): _log_lost_create_race is a wrapper around a single logger.info line, and its name/docstring suggest more machinery than it holds. Inlining the log at the three call sites (with the existing comment) would be simpler; if kept, a shorter docstring would match its one-line body.

**`tests/supervisor/test_run_create_lock.py`** (line 275): Note for CI: the [vprogram] parametrization depends on VerifiableProgramContent.gpu, which exists in the pinned aleph-message==1.5 but not in older installs; it fails in this review sandbox with a stale 1.2.1.dev (the pre-existing test_vprogram.py fails identically there). Expected to pass on the pinned dependency set.

foxpatch-aleph (approved):
The core fix is correct and well-built. The per-hash lock in create_lock.py is race-free (the miss-to-insert window contains no await, so the event loop cannot interleave it), its refcount cleanup handles exceptions and cancellation via the finally, and dropping empty entries avoids unbounded growth. The start_persistent_vm rewrite is a faithful wrap of an unchanged body. The VmAlreadyExistsError catches sit before the generic except Exception in all three create branches, and only supervisor.create_vm within those try blocks can raise it, so nothing real is silently swallowed. Of the six new tests, five pass locally; the sixth ([vprogram]) fails only because this environment has aleph-message 1.2.1.dev4 installed where the project pins 1.5 — the pre-existing test_vprogram.py suite errors identically here — so it is an environment artifact, not a PR defect, and CI should confirm. Remaining points are minor observations rather than blockers.

**`src/aleph/vm/agent/create_lock.py`** (line 44): The lock is correct, but note the module state is bound to a single event loop: if start_persistent_vm were ever run from multiple threads/loops the dict get-and-insert would race. That holds today (one loop per agent process) but may be worth one sentence in the docstring. Separately, the on-demand program path (_ensure_program_vm in run.py, which uses program_client.creation_lock) is a create path this lock does not reach: an on-demand request can still retire a VM a concurrent persistent start just brought up, by design ("a VM this agent process did not configure is recreated"). Pre-existing and deliberate, but since the module docstring enumerates the start paths it covers, a line acknowledging the on-demand path and its separate lock would help future readers.

**`tests/supervisor/test_run_create_lock.py`** (line 153): The module promises that starts of different VMs still run in parallel, but no test pins that: a regression that made the lock process-global rather than per-hash would pass this suite. A two-hash test asserting interleaved progress (e.g. different hashes both inside their slow builds at once) would close it.

**`src/aleph/vm/agent/migration/runner.py`** (line 362): Not covered by this diff, but relevant to the PR description: a migration import that loses the race to a scheduler push raises VmAlreadyExistsError here and lands in the generic handler, reporting IMPORT_FAILED even though the VM is up. It is safe (vm_created stays False so no retire_vm, and _vm_exists keeps the rmtree away from the winner's disks), but the operator sees a failed import for a VM that exists. Fine as a follow-up; just noting the new VmAlreadyExistsError handling in create_vm_execution does not literally reach this path.

foxpatch-aleph (approved):
This PR correctly fixes a real destructive race: two concurrent start paths for one hash both saw the VM as missing, both built it, and the loser's teardown deleted the winner's live VM and purged its volumes. The per-hash refcounted asyncio.Lock in create_lock.py is sound — I traced the creation window (no await between the dict miss and insert, safe under single-loop asyncio), the increment-before-acquire that keeps a queued waiter's lock alive, the last-out cleanup, and cancellation-while-waiting; there is no reentrancy (start_persistent_vm is the sole taker and never recurses) and no lock-ordering cycle with the storage reconciler's creating() refcount. Placing the lock inside start_persistent_vm rather than at call sites is the right call, and the three VmAlreadyExistsError catches make the destructive retire impossible even on the create paths the lock doesn't reach (operator recreate, migration import — the runner is already safe in its own direction via its vm_created guard and _vm_exists rmtree check). I ran the new tests (5/6 pass; the [vprogram] param fails only in this sandbox due to a stale aleph-message 1.2.1.dev4 installed instead of the pinned 1.5 — a pre-existing vprogram test fails identically on the same pre-existing content.gpu read at run.py:669) and 76 pre-existing tests around the re-indented start_persistent_vm, all green. ruff format and mypy are clean; the ARG/FBT ruff findings on the new test file are a version artifact, matching findings on existing test files under the sandbox's newer ruff. Two minor nits below, neither blocking.

**`tests/supervisor/test_run_create_lock.py`** (line 171): Minor: consider also asserting `create_lock._waiting == {}` here (and at lines 229/235). A `_waiting`-only leak would not be caught by the current `_locks == {}` assertion, and it is the worse failure mode: a leftover count makes every later exit compute `remaining == 1`, pinning the lock entry for that hash permanently. Asserting both dicts pins the full no-leak invariant.

**`src/aleph/vm/agent/create_lock.py`** (line 33): Nit: `_waiting` counts the holder as well as the waiters (the refcount is incremented before acquire), so the name slightly under-describes it. Something like `_refcounts` would read more precisely, but the comment above makes the intent clear enough that this is optional.

foxpatch-aleph (approved):
Solid, well-documented fix. The per-hash lock is correctly implemented: the lock creation is race-free on a single event loop (no await between miss and insert), the waiter count is incremented before acquire so a queued caller keeps the lock alive, and the finally-based release survives exceptions and cancellation — I verified the cancellation accounting with a direct stress test and traced all four callers (reconciler, v1 allocations handler, notify_allocation, reconciler _start_one) into the same event loop with no reentrancy path. The VmAlreadyExistsError handling is correctly ordered before the generic teardown branch in all three create branches, and the migration runner's unlocked raw create_vm is exactly the residual path it covers. start_persistent_vm's rewrite is a pure indentation change with no behavioral drift. Tests are meaningful and pass (5/6 in my environment; the [vprogram] parametrization fails only because the sandbox has aleph-message 1.2.1 instead of the pinned 1.5, and pre-existing tests fail identically), and the neighboring suites (run routing, reconciler, allocations_v2, vprogram_port_map) show no regressions.

**`src/aleph/vm/agent/run.py`** (line 873): Noting a residual race the lock does not cover: the on-demand program path (_ensure_program_vm) still calls supervisor.create_vm without vm_create_lock, so a scheduled start and an on-demand request for the same hash can race. Its failure mode is an HTTP error to one request rather than the destructive retire, so this is fine for this PR — just worth keeping in mind if the on-demand path ever gains a teardown-on-create-failure.

**`src/aleph/vm/agent/create_lock.py`** (line 33): Minor: _locks and _waiting are two parallel dicts that must stay in sync by hand. A single dict mapping namespace -> (lock, count) would make divergence impossible and save the double lookup in the finally. Cosmetic only.

**`tests/supervisor/test_run_create_lock.py`** (line 171): assert create_lock._locks == {} asserts global emptiness; if some future test leaked a lock for a different hash, the failure would point at this test confusingly. assert str(_HASH) not in create_lock._locks would target the assertion. Very minor.

foxpatch-aleph (approved):
The fix is correct and well-targeted. The per-hash lock's refcount logic is sound (increment before acquire, decrement in finally, no await points that could interleave, so waiters always keep the lock object alive and the last caller out cleans up), and placing it inside start_persistent_vm covers all three start paths. Catching VmAlreadyExistsError before the generic teardown branch on all three create paths makes the destructive retire impossible even for create paths the lock does not reach, which is the right defense-in-depth. I ran the new tests in a sandbox: 5/6 pass; the vprogram param fails only due to aleph-message version skew in my environment (1.4.0 installed vs 1.5 pinned, missing VerifiableProgramContent.gpu), not a defect. mypy and ruff format are clean on the changed files; the ruff check complexity warnings on create_vm_execution/start_persistent_vm are a pre-existing, non-gated class of violation. Two non-blocking observations: the lock is process-local and single-event-loop only (fine for this aiohttp app, but the assumption is only implicit in a code comment), and _ensure_program_vm's separate program_client.creation_lock means an on-demand program create can still race a persistent start for the same hash — the loser gets an HTTP error rather than a destructive teardown, so it is residual, not blocking.

**`src/aleph/vm/agent/create_lock.py`** (line 37): The lock is process-local and single-event-loop only (module-global state, one loop per aiohttp app), which the code relies on but only states implicitly in the comment about the miss-and-insert window. A one-line note in the module docstring that this serialises within one agent process would save a future reader from wondering about multi-loop or multi-process scenarios.

**`src/aleph/vm/agent/run.py`** (line 1129): Note for a follow-up: _ensure_program_vm (line 848) serialises on program_client.creation_lock(vm_id), a different lock for the same hash, so an on-demand program create can still race a persistent start_persistent_vm. The loser there fails via _raise_http_for_program_error rather than retiring the winner's VM, so it is not destructive — just residual rather than fixed. Worth tracking separately.

**`tests/supervisor/test_run_create_lock.py`** (line 208): The purge fixture is used here purely as a stub (to keep a retire from touching real storage) but never asserted on, while its docstring says "spy". A brief note in the test, or asserting purge.assert_not_called() alongside delete_vm, would make the intent clearer.

**`tests/supervisor/test_run_create_lock.py`** (line 153): Nice coverage overall — the ordering test and the lock-release-after-raise test pin exactly the two subtle properties of the lock. One small gap: no test pins that starts of different hashes still proceed in parallel (the whole point of keying by hash rather than one global lock).

PR #1223, opened 2026-09-09
fix(daemon): cap a confidential VM at one GPU and name the card it refuses

odesenfans:
## Summary
The confidential-GPU feature that landed in #1200 already refuses a card on the SEV-SNP arm that brings its own measured kernel cmdline (the LUKS, writable-root arm), which is what the first version of this PR was for. Two things it does not do are still worth having, and they are what this PR is now.

The refusal does not say which card it is about, so an operator reading the error has to guess which slot to empty; it now names every `pci_host` in the spec. And nothing caps the number of cards: the reserved 64-bit MMIO window and the guest's PCI port and chassis numbering were sized for a single one, so a second card silently shared a window that may not hold it. A spec carrying more than one is refused with the count. The agent refuses a V-PROGRAM asking for more than one card before it stages anything (`vprogram_launch.py`, `gpu.count > 1`), so the daemon-side cap is the backstop for a request that reaches CreateVm another way, not a second opinion about the same input.

Stacked on #1218.

## Changes
- `rust/crates/supervisor-daemon/src/lifecycle.rs`: in `snp_config_slice_with`, the opaque-arm refusal that #1200 added now interpolates the spec's pci addresses into its message and keeps both reasons (the guest carries no verified GPU attestation stage on that arm, and the bounce-buffer kernel parameter a confidential card needs is neither required nor measured there). New check below it: more than one GPU on an SEV-SNP spec is refused with the count. Both run before the confidential-computing probe, so a spec that cannot be admitted never touches the hardware. The verity arm is otherwise unchanged.
- The cap's comment states the honest reason after review: the window sizing does sum every card's BARs and the argv builder does emit a port and chassis per card, so the mechanics are not what stops a second one. What stops it is that a single-card confidential guest is what has been exercised end to end, guest runtime included, and nothing at the create gate can check that a second card's attestation, bounce buffer and NVLink topology behave. Fail-closed policy until a multi-card guest has actually been run.

No controller change: the SNP conformance fixtures already carry exactly one GPU on the verity arm (`controller_argv_snp/gpu.json`) and none on the opaque arm (`controller_argv_snp/luks.json`), so no argv moves.

## Test plan
- `lifecycle.rs::snp_config_slice_rejects_a_gpu_on_the_opaque_cmdline_arm`: #1200's test, extended to assert the refusal names the card. This PR's own near-duplicate of it was dropped rather than kept alongside.
- `lifecycle.rs::snp_config_slice_takes_a_gpu_on_the_verity_arm`: the same card on the measured arm still builds a slice with the window set, so nothing else narrowed.
- `lifecycle.rs::snp_config_slice_refuses_more_than_one_gpu`: two cards on the verity arm are refused, with a window closure that returns a size that fits, so the count alone is what refuses the spec.
- Checks from `rust/`: `cargo fmt --check` exit 0, `cargo clippy --locked --all-targets -- -D warnings` exit 0 (which type-checks every test). The test suites were not run locally this round (some daemon tests need root to unwind DHCP servers); CI runs them.

### DIFF
--- a/rust/crates/supervisor-daemon/src/lifecycle.rs
+++ b/rust/crates/supervisor-daemon/src/lifecycle.rs
@@ -2103,18 +2103,34 @@ fn snp_config_slice_with(
             "SEV-SNP measured boot requires kernel_path and initrd_path".to_string(),
         ));
     }
-    // Only the measured V-PROGRAM arm derives its own cmdline here, and only
-    // that image runs the guest-side GPU attestation stage. The opaque arm
-    // passes the agent's cmdline through verbatim and the daemon cannot tell
-    // whether the guest verifies the card at all, so a GPU there would be
-    // owner hardware nothing in the boot chain checks. Fail closed before the
-    // per-card rules run.
+    // A confidential GPU is admitted only on the arm whose cmdline this
+    // function derives (the dm-verity one), and only that image runs the
+    // guest-side GPU attestation stage. Two things go wrong on the other
+    // arm. It passes the agent's cmdline through verbatim, so the daemon
+    // cannot tell whether the guest verifies the card at all, and a card in
+    // confidential-computing mode does no plaintext DMA, so the guest has to
+    // bounce every transfer through a shared buffer whose size is a kernel
+    // parameter the measured runtime pins and the launch digest covers. On
+    // the opaque arm that parameter is neither required nor measured. Fail
+    // closed before the per-card rules run, and say which arm takes the card.
     if !spec.gpus.is_empty() && !tee.kernel_cmdline.is_empty() {
-        return Err(RpcError::InvalidBackend(
-            "confidential GPUs are only supported on measured V-PROGRAM specs, not on the \
-             opaque-cmdline SNP instance arm"
-                .into(),
-        ));
+        let hosts: Vec<&str> = spec.gpus.iter().map(|g| g.pci_host.as_str()).collect();
+        return Err(RpcError::InvalidBackend(format!(
+            "a GPU (pci_host {}) cannot be attached to an SEV-SNP VM that brings its own kernel \
+             cmdline; confidential GPUs are only supported on measured V-PROGRAM specs, not on \
+             the opaque-cmdline SNP instance arm",
+            hosts.join(", ")
+        )));
+    }
+    // One card per confidential VM: the reserved 64-bit MMIO window and the
+    // guest's PCI port and chassis numbering were both sized for a single
+    // card, so a second one is refused here rather than handed a window that
+    // may not hold it.
+    if spec.gpus.len() > 1 {
+        return Err(RpcError::InvalidBackend(format!(
+            "an SEV-SNP VM takes at most one GPU, the spec carries {}",
+            spec.gpus.len()
+        )));
     }
     // A GPU may enter a confidential guest only in NVIDIA CC mode: the card
     // then refuses plaintext DMA and answers attestation, and the guest
@@ -5097,9 +5113,11 @@ mod tests {
     #[test]
     fn snp_config_slice_rejects_a_gpu_on_the_opaque_cmdline_arm() {
         // A confidential instance renders its own measured cmdline, so the
-        // guest carries no verified GPU attestation stage. A CC-mode card must
-        // not be admitted there even though the card itself would pass: the
-        // injected probe says "on" and is never consulted.
+        // guest carries no verified GPU attestation stage, and the
+        // bounce-buffer parameter a confidential card needs is neither
+        // required nor measured there. A CC-mode card must not be admitted
+        // even though the card itself would pass: the injected probe says
+        // "on" and is never consulted.
         let harness = harness_with_gpus(vec![nvidia_card("06:00.0")]);
         let state = &harness.state;
         let root = state.host.settings.execution_root.clone();
@@ -5113,7 +5131,11 @@ mod tests {
         let cc_on = |_: &str, _: &str| Ok(Some(crate::gpu_cc::CcMode::On));
         match snp_config_slice_with(state, &spec, cc_on, |_| Ok(1024)) {
             Err(RpcError::InvalidBackend(msg)) => {
-                assert!(msg.contains("measured V-PROGRAM specs"), "{msg}")
+                assert!(msg.contains("measured V-PROGRAM specs"), "{msg}");
+                assert!(
+                    msg.contains("06:00.0"),
+                    "the refusal must name the card the operator has to detach: {msg}"
+                );
             }
             other => panic!("a GPU on the opaque arm must be InvalidBackend, got {other:?}"),
         }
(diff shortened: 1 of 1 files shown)


foxpatch-aleph (approved):
The fix is correct and well-scoped: the opaque SNP arm does pass the agent's cmdline verbatim and splice nothing, so a CC-mode GPU there would boot without the required (and measured) swiotlb bounce-buffer parameter — refusing it fails closed. Both new checks run before the CC probe and MMIO sizing, sit on the single SNP-slice choke point (lifecycle.rs:1734), and are scoped to SEV-SNP only. The three tests isolate exactly the new behavior and their 'red before the fix' claims match the pre-patch code path. One non-blocking nit: the multi-GPU cap's rationale comment (lifecycle.rs:2103-2106) is imprecise — gpu_bar.rs already sums BARs across multiple cards (see window_sums_every_card_under_the_devices_dir) and snp_gpu_args enumerates rp{index}/chassis={index+1} correctly — so the honest justification is that the design and guest runtime were validated for a single card, not that the window cannot hold two. The cap itself is a sound fail-closed policy either way.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 2104): The stated rationale is imprecise about the code: `gpu_bar::gpu_mmio64_mb` does sum BARs across all cards (gpu_bar.rs:105 even has a multi-card test) and the controller's `snp_gpu_args` enumerates `rp{index}`/`chassis={index+1}` per card. The stronger, accurate justification is that the SNP GPU design (docs/superpowers/specs/2026-09-04-nvidia-cc-design.md §7.2) and the guest runtime were validated for exactly one card, so this cap is fail-closed policy. Consider rewording so a future reader doesn't conclude the window code can't handle two cards when it can. Non-blocking.

foxpatch-aleph (approved):
The change is small, correct, and well-scoped: the opaque-arm refusal now names every pci_host it is about, and a new fail-closed cap refuses an SEV-SNP spec carrying more than one GPU. Both checks run before the per-card probe so an inadmissible spec never touches the hardware, the interpolated pci_host strings only ever reach an error message (no injection surface), and the multi-GPU test genuinely pins the cap since its harness would otherwise admit the spec. The PR description's claims all check out: the agent-side gpu.count > 1 guard exists in vprogram_launch.py, and the controller fixtures carry one GPU (gpu.json) and none (luks.json) so no argv output moves. The honest-reason rewrite of the cap's comment (a policy limit until a multi-card guest is exercised, not a code limit) is accurate to what the code does. Only nit: the new verity-arm acceptance test nearly duplicates the existing snp_config_slice_accepts_a_cc_mode_gpu_and_sizes_the_window test.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 5193): Minor: `snp_config_slice_takes_a_gpu_on_the_verity_arm` is nearly identical to the existing `snp_config_slice_accepts_a_cc_mode_gpu_and_sizes_the_window` (line 5156) — same spec, same cc_on probe, same 524288 MB window, and the same `pci_mmio64_mb == Some(524288)` assertion. The existing test already proves the verity arm still takes a card, so this one adds little beyond documentation; if it stays, a brief note that it is the deliberate counterpart of the refusal test would help a future reader see why both exist.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 2112): For an opaque spec carrying two cards, the opaque-arm refusal above fires first, so the operator sees the "brings its own kernel cmdline" message rather than the count message. Both are refusals so there is no gap, but if you ever want the more specific diagnosis first, moving this check above the opaque-arm one would also cover a 2-card opaque spec with the count message.

foxpatch-aleph (approved):
The background task's exit code 0 was misleading — that was just `tail`'s status; the log confirms `cargo: command not found`, so the tests were indeed never run on this machine, exactly as I noted in the review. My assessment stands: I verified the PR by close reading rather than execution, the author states clippy (which type-checks all tests) passes, and CI runs the suites.

Final verdict is unchanged — **APPROVE**, with the three minor, non-blocking nits:

1. `snp_config_slice_takes_a_gpu_on_the_verity_arm` (lifecycle.rs:5253) largely duplicates the existing `snp_config_slice_accepts_a_cc_mode_gpu_and_sizes_the_window` test.
2. The multi-GPU refusal (lifecycle.rs:2134) reports only the count; interpolating the pci_hosts — as the opaque-arm refusal now does — would be more actionable for the operator.
3. The instance path (`run.py:671`) resolves N CC cards with no early agent-side cap, so an over-carded instance churns capacity holds before the daemon backstop refuses it at CreateVm; a `vprogram_launch.py:199`-style early check would fail faster.

```json
{
  "verdict": "APPROVE",
  "summary": "Correct, well-commented fail-closed policy change. The cap sits in the single SNP config gate ahead of the hardware probe; the opaque-arm refusal now names the offending pci_hosts. I verified the PR's claims: the agent-side V-PROGRAM cap exists (vprogram_launch.py:199), the controller SNP fixtures carry 0 or 1 GPU so no argv moves, and the 'honest reason' comment is accurate (gpu_bar.rs sums every card's BARs and snp_gpu_args emits a root port and chassis per card, so the cap is indeed policy, not a code limit). The bounce-buffer rationale is corroborated by manifest.py:41 and docs/architecture/confidential.md:370. No Rust toolchain exists on this review machine, so the tests could not be executed locally (a background build attempt confirmed cargo is absent); they were instead verified by close reading and are structurally sound, including the two-GPU test's fitting window that isolates the count check. Only minor, non-blocking nits: the new verity-arm test duplicates an existing test, the count error could name the cards for operator parity, and the instance path still resolves N cards before the daemon backstop refuses them."
  "comments": [
    {
      "path": "rust/crates/supervisor-daemon/src/lifecycle.rs",
      "line": 5253,
      "body": "Non-blocking: this new test is nearly identical to the pre-existing `snp_config_slice_accepts_a_cc_mode_gpu_and_sizes_the_window` (same verity-arm spec, same injected cc_on probe, same `pci_mmio64_mb == Some(524288)` assertion) — the existing test additionally records the probe's arguments and the cache update. The PR dropped the near-duplicate of the opaque-arm test; this one could arguably be folded into the existing acceptance test too. Harmless as a regression guard for 'the rules above narrow nothing else', so fine either way."
    },
    {
      "path": "rust/crates/supervisor-daemon/src/lifecycle.rs",
      "line": 2134,
      "body": "Non-blocking: for operator ergonomics and consistency with the opaque-arm refusal above, this message could also interpolate the offending pci_hosts (the `hosts` vec pattern is already there) — 'the spec carries 3' tells the operator the count, but the addresses tell them what to detach."
    },
    {
      "path": "src/aleph/vm/agent/run.py",
      "line": 671,
      "body": "Observation outside the diff: the instance path resolves `count` CC cards with no agent-side cap (unlike the V-PROGRAM path, which refuses `gpu.count > 1` in vprogram_launch.py before staging), so an SNP instance asking for two will place holds on both cards and only be refused at CreateVm by the new daemon backstop. Fail-closed and teardown handles it, but an early check mirroring vprogram_launch.py:199 would fail faster and avoid churning capacity holds — maybe a follow-up."
    }
  ]
}
```

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 5253): Non-blocking: this new test is nearly identical to the pre-existing `snp_config_slice_accepts_a_cc_mode_gpu_and_sizes_the_window` (same verity-arm spec, same injected cc_on probe, same `pci_mmio64_mb == Some(524288)` assertion) — the existing test additionally records the probe's arguments and the cache update. The PR dropped the near-duplicate of the opaque-arm test; this one could arguably be folded into the existing acceptance test too. Harmless as a regression guard for 'the rules above narrow nothing else', so fine either way.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 2134): Non-blocking: for operator ergonomics and consistency with the opaque-arm refusal above, this message could also interpolate the offending pci_hosts (the `hosts` vec pattern is already there) — 'the spec carries 3' tells the operator the count, but the addresses tell them what to detach.

**`src/aleph/vm/agent/run.py`** (line 671): Observation outside the diff: the instance path resolves `count` CC cards with no agent-side cap (unlike the V-PROGRAM path, which refuses `gpu.count > 1` in vprogram_launch.py before staging), so an SNP instance asking for two will place holds on both cards and only be refused at CreateVm by the new daemon backstop. Fail-closed and teardown handles it, but an early check mirroring vprogram_launch.py:199 would fail faster and avoid churning capacity holds — maybe a follow-up.

PR #1224, opened 2026-09-09
fix(cli): make aleph-vm storage status and list truly read-only

odesenfans:
## Summary
`aleph-vm storage status` and `list` are documented read-only, and were not. Reading a `.reclaimable` marker that does not parse unlinked it, `settings.setup()` created every configured directory before the missing-database refusal, and `setup_pools()` adopted a pool on first sight (in-pool marker plus adoption registry). Run by an operator without the agent's privileges the unlink raised an uncaught PermissionError and killed the command; run against a mistyped execution root, `status` refused and left a tree of empty directories at the typo; run on a node whose second disk is not mounted, it adopted that mountpoint as a pool, which is the write the adoption guard exists to prevent. Two more ways the listing lied: a marker timestamp without a UTC offset parsed naive, so subtracting it from the aware clock raised TypeError and aborted the whole listing over one row, and a future-dated marker printed a negative age. Finally `reclaim` blamed device-mapper for any directory the purge left behind and sent the operator to `storage reconcile`, but the purge also leaves a directory on any rmtree/unlink OSError (EACCES, EROFS, EBUSY, an immutable file), which no teardown fixes.

Stacked on #1215.

## Changes
- `agent/vm/reclaimable.py`: `read_marker(dir, *, repair=True)`, threaded through `iter_reclaimable` and `reclaimable_bytes`, so a caller that may not write reads a corrupt marker without removing it; the removal itself is now best effort instead of an exception out of a read. `from_json` reads a timestamp with no offset as UTC (every writer here uses UTC).
- `storage_pools.py`: `setup_pools(read_only=False)` and `_adopt_pool(read_only=...)`. Read-only still validates, classifies and refuses a pool whose marker vanished after adoption; it just adopts nothing.
- `agent/vm/purge.py`: `purge_vm_storage` returns a `PurgeResult` (files deleted, and the `KeptDirectory` entries with their reason and whether a dm target is the cause) instead of a bare count. Callers that ignored the count are unaffected. Foxpatch review: the device-mapper hold used to be recomputed inside the per-directory loop, re-walking every pool on each iteration; hoisted above the loop, computed once per purge.
- `agent/storage_cli.py`: `status`/`list` refuse a missing database before any setup runs, skip `settings.setup()` entirely and call `setup_pools(read_only=True)`; they read markers with `repair=False`. `_age` clamps at zero. `reclaim` reports each surviving directory with its real reason and the partial count on stderr, and only points at `storage reconcile` when a dm target is actually the cause. The env file loads with `interpolate=False`, matching systemd's `EnvironmentFile=`, so a value containing `${...}` reaches the settings intact.
- `agent/storage_cli.py` (second commit): `_pool_usage` and `_retention_budget` return `None` when the pool's filesystem cannot be stat'ed, and `_figure` prints `unknown` rather than `0 B`, so an unreadable pool is no longer indistinguishable from a full one; the cache budget takes the same treatment, and a `reap` budget stays a real zero. `_is_marked_reclaimable` reads with `repair=False`, so a reclaim about to be refused (a running agent) no longer unlinks the corrupt markers of unrelated VMs on its way to refusing.
- `docs/architecture/storage.md`: what read-only means for these two verbs, the two reasons a purge leaves a directory and the different answers they need, the UTC rule for a bare marker timestamp, and the no-interpolation rule for the env file.

## Test plan
- `tests/supervisor/test_storage_cli.py`: a corrupt marker survives `list` and `status` (exit 0, other rows printed); a naive marker timestamp shows a real age and a future-dated one shows `0d 0h`; `reclaim` under a monkeypatched `rmtree` raising EROFS names the OSError, never mentions device-mapper as the cause and does not advise `storage reconcile`, with stdout empty; `status` adopts no pool and writes no `volume-pools.json`; `status` creates no directory before refusing a missing database; the env file is not `${}` expanded; `_env_file_path` covers the flag, `$ALEPH_VM_ENV_FILE` and the packaged default.
- `tests/supervisor/test_vm_purge.py`: the result names a dm-held directory (`device_mapper=True`) and an EROFS one (`device_mapper=False`), and a complete purge keeps nothing.
- `tests/supervisor/test_reclaimable.py`: `repair=False` keeps a corrupt marker (through `read_marker` and `reclaimable_bytes`), a marker that cannot be removed still reads as None, and a timestamp with no offset reads as UTC.
- Review round: `test_status_says_unknown_for_a_pool_it_cannot_measure` (disk_usage raising EACCES on one pool; BUDGET and FREE read `unknown`, the other pool still shows figures), `test_status_says_unknown_for_a_cache_budget_it_cannot_compute`, and `test_a_refused_reclaim_removes_no_marker` (agent running, exit 3, a corrupt marker on a directory sorting before the target survives).
- Foxpatch review round: the device-mapper hold hoist is behaviour-preserving, pinned by the existing `test_vm_purge.py` coverage (`held`/`held_here` are unchanged in what they classify, only in when they are computed).
- Checks: `pytest tests/supervisor/test_storage_cli.py tests/supervisor/test_vm_purge.py tests/supervisor/test_reclaimable.py tests/supervisor/test_reconciler.py` = 188 passed; `ruff format --diff`, `isort --check-only --profile black`, and `mypy src/aleph/vm/ tests/` all clean. A whole-`tests/supervisor` collection currently fails to collect on this checkout (`test_snp_instance_run.py`, `test_vprogram_launch.py`: `ImportError: cannot import name 'ConfidentialGpuRequirement'`), from the confidential-GPU feature merged into `dev-2.1` after this chain was opened (#1200) needing a newer `aleph_message` than this venv has installed; unrelated to this PR's files. Also driven by hand against a real execution root: the refusal creates nothing, `reconcile --dry-run` adopts the pool and migrates the DB, and a following `list` leaves a corrupt marker in place.

### DIFF
--- a/docs/architecture/storage.md
+++ b/docs/architecture/storage.md
@@ -334,8 +334,12 @@ because the marker outlives every other record of it (the registry record is
 forgotten and the DB rows are deleted), and it is what lets the node
 authorize the owner's own erase of retained data; it is optional, so markers
 written before the field and orphan markers for a VM whose message the node
-never held simply have none. Under `reap` no marker is ever written: the
-purge is immediate.
+never held simply have none. `reclaimable_since` is written in UTC with an
+offset; a timestamp that arrives without one (a marker restored or edited by
+hand) is read as UTC rather than as a naive datetime, since a naive one
+cannot be compared with the aware clock the rest of the agent uses and every
+subtraction against it raises, costing a whole listing or eviction pass for
+one marker. Under `reap` no marker is ever written: the purge is immediate.
 A create for the same hash adopts the directory (the marker is unlinked and
 the volumes are reused through the existing sticky placement in
 `volume_path_for`), which is what `reconciler.creating()` does on entry.
@@ -585,6 +589,25 @@ in its `--help` epilog) needs no running *agent process*:
   `live` when the registry knows its hash and `unmarked` otherwise (an
   orphan no pass has reached yet is not a live VM). Read-only, registry
   live set only.
+
+  Read-only here is literal, and it took work to make it so, because three
+  writes were hiding under a listing. `settings.setup()` makes every
+  configured directory, so the two verbs skip it entirely and refuse a
+  missing database before it would have run (an operator who mistyped the
+  execution root used to get the refusal and a tree of empty directories at
+  the typo; nothing these verbs read needs `setup()`, since every path is
+  derived when the settings object is built). `setup_pools()` adopts a pool
+  on first sight, writing the in-pool marker and the adoption registry, so
+  they call it with `read_only=True`: it still validates and classifies, and
+  still refuses a pool whose marker vanished after adoption, but it adopts
+  nothing (on a node whose second disk is not mounted, adopting that path is
+  precisely the write the guard exists to prevent). And a `.reclaimable`
+  marker that does not parse is read with `repair=False`: removing it is the
+  reconciler's repair, a listing that unlinks a file is not read-only, and
+  the unlink raises for an operator without the agent's privileges, which
+  cost the whole command rather than that one row. The one write they do
+  make is the schema migration of a database that is already there, without
+  which the registry cannot be read at all.
 - `storage reclaim <hash> [--trust-registry]`: purge one reclaimable
   directory now. Checks the name and the `.reclaimable` marker first, purely
   locally, so a typo or an unrelated hash fails instantly rather than
@@ -601,12 +624,16 @@ in its `--help` epilog) needs no running *agent process*:
   retained directories by clearing their markers, and one that started
   while the supervisor was being asked is in no answer this process holds,
   so the second read is what keeps the purge off the disks a create is at
-  that moment building on. Finally, if the purge itself leaves the
-  directory behind (a device-mapper target still holds one of its volumes,
-  the same guard `purge_vm_storage` always applies), reclaim reports that
-  and exits non-zero rather than claiming success. `reclaim` never tears
-  devices down itself: `storage reconcile` is the CLI path that does, for
-  every VM nothing owns, so the refusal points the operator there.
+  that moment building on. Finally, if the purge itself leaves a directory
+  behind, reclaim names each one with the reason `purge_vm_storage` gives
+  for it and exits non-zero rather than claiming success. There are two
+  such reasons and they need different answers: a live device-mapper target
+  holding the volumes (the guard `purge_vm_storage` always applies), which
+  is freed by tearing that target down, and the removal itself failing (a
+  read-only filesystem, an immutable file, a directory this user may not
+  write), which no teardown fixes. Only the first sends the operator to
+  `storage reconcile`, which is the CLI path that tears devices down, for
+  every VM nothing owns; `reclaim` never does it itself.
 - `storage reconcile [--dry-run] [--trust-registry]`: run one
   `reconcile_storage()` pass. Two processes decide what it may do, and they
   are not the same one. The **agent** is this Python service
@@ -707,7 +734,10 @@ anything:
   `.env` in the working directory. The CLI therefore loads that file itself
   (`--env-file`, else `$ALEPH_VM_ENV_FILE`, else the packaged path) and
   rebuilds the settings singleton from it, logging which file it used or
-  that there was none. Values already in the environment win, so a one-off
+  that there was none. The file is read without `${}` expansion, the way
+  systemd's `EnvironmentFile=` reads it, so the CLI sees exactly the value
+  the daemon sees and a secret containing a dollar sign is not silently
+  truncated. Values already in the environment win, so a one-off
   override on the command line still works, and a `--env-file` that does
   not exist is an error rather than a silent fall back to the defaults: a
   pass run with `VOLUME_RETENTION` at its `reap` default on a node
@@ -723,7 +753,10 @@ anything:
   with a raw sqlite error. `status` and `list` are the exception; being
   read-only, they refuse a database that does not exist and exit `1` rather
   than create an empty one, so an operator pointed at the wrong execution
-  root sees the mistake instead of an empty listing.
+  root sees the mistake instead of an empty listing. That refusal is the
+  first thing they do, before the configuration is turned into directories
+  and before the pools are set up, so the wrong execution root is reported
+  rather than created.
 
 `reconcile` and `reclaim` can purge, so they apply the same fail-closed rule
 `reconciler._startup_refusal` applies to the daemon's own startup pass: a
--- a/docs/architecture/storage.md
+++ b/docs/architecture/storage.md
@@ -583,7 +583,15 @@ subparser on `agent/cli.py`'s own parser, which points at `storage --help`
 in its `--help` epilog) needs no running *agent process*:
 
 - `storage status`: per pool, live / reclaimable / cache / free bytes
-  against the budgets. Read-only, registry live set only.
+  against the budgets. Read-only, registry live set only. A figure this
+  process could not measure prints `unknown`, never `0 B`: the free space and
+  the budget of a pool whose filesystem cannot be stat'ed (a mountpoint that
+  went away, a directory this user may not read) would otherwise be
+  indistinguishable from a pool that is genuinely full, which is the state an
+  operator runs this command to find. The cache budget is a share of the
+  filesystem holding the cache root, so it says `unknown` for the same
+  reason. A budget under `VOLUME_RETENTION=reap` is a real zero and prints
+  as one.
 - `storage list [--reclaimable]`: hash, pool, size, reason, age. The reason
   is the marker's for a reclaimable directory; an unmarked directory reads
   `live` when the registry knows its hash and `unmarked` otherwise (an
--- a/src/aleph/vm/agent/storage_cli.py
+++ b/src/aleph/vm/agent/storage_cli.py
@@ -527,6 +527,43 @@ def _cli_live_set(registry: AgentVmRegistry) -> LiveSet:
     )
 
 
+def _pool_usage(path: Path) -> tuple[int | None, int | None]:
+    """(total, free) bytes of the filesystem holding ``path``, or (None,
+    None) when it cannot be read: a mountpoint that went away, a directory
+    this user may not stat, a filesystem that is gone."""
+    try:
+        usage = shutil.disk_usage(str(path))
+    except OSError:
+        logger.warning("Could not read the usage of %s", path, exc_info=True)
+        return None, None
+    return usage.total, usage.free
+
+
+def _retention_budget(total: int | None) -> int | None:
+    """The retention budget in bytes, or None when it cannot be computed.
+
+    Under ``reap`` it is zero and nothing else, whatever the filesystem
+    says. Otherwise it is measured against the pool's size (typically a
+    percentage of it), so a size this process could not read leaves the
+    budget unknown rather than zero.
+    """
+    if settings.VOLUME_RETENTION == "reap":
+        return 0
+    if total is None:
+        return None
+    return parse_budget(settings.VOLUME_RETENTION_BUDGET, total)
+
+
+def _figure(size: int | None) -> str:
+    """A byte figure, or ``unknown`` for one this process could not measure.
+
+    Zero is a measurement, and a pool printed as 0 bytes free with a 0 byte
+    budget looks exactly like a full pool, which is the state an operator
+    runs this command to find. An unreadable pool has to say so.
+    """
+    return "unknown" if size is None else _human(size)
+
+
 def _status(registry: AgentVmRegistry, out: TextIO) -> int:
     live = live_hashes(registry)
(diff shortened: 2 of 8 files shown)


foxpatch-aleph (approved):
Solid, carefully-reasoned PR. The read-only enforcement is complete: a missing DB is refused before any setup, settings.setup() is skipped (verified it is the step that creates cache/volume directories), setup_pools(read_only=True) bypasses the adoption write and heal while keeping the vanished-marker refusal, and read_marker(repair=False) stops a listing from unlinking a corrupt marker — leaving only the documented DB schema migration as a write. The new liveness probe fails closed in every direction I could attack it: STOPPED requires a positive connection refusal, wildcard binds probe both loopbacks so an IPv6-only listener is not mistaken for a dead agent, family-unavailable errnos stay silent, and PermissionError classifies as UNKNOWN. _probe_agent targets the agent's own bind address (supervisor.py:608), so it cannot be fooled by a foreign service. The purge_vm_storage int→PurgeResult change is safe: all four callers (cache.py:620, retire.py:234, reconciler.py:438, reconciler.py:652) ignore the return value, and the reclaim output now reports a per-directory survival reason, with the device_mapper distinction driving the reconcile advice. interpolate=False matches systemd EnvironmentFile= semantics, and _age clamping at zero is correct. All 123 new tests pass. Two non-blocking notes below.

**`src/aleph/vm/agent/storage_cli.py`** (line 462): Nit (non-blocking): on Python 3.10 (requires-python = ">=3.10"), asyncio.TimeoutError is not a subclass of the built-in TimeoutError — they were only unified in 3.11 — so a timeout raised by asyncio.wait_for lands in the final else branch and calls _socket_reach() instead. The result is the same UNKNOWN verdict the explicit check produces whenever the socket file exists, so the code is correct today, but only because the fallback coincides. Consider catching (TimeoutError, asyncio.TimeoutError) to make the intended classification explicit rather than accidental.

**`src/aleph/vm/agent/vm/reclaimable.py`** (line 398): Nit (non-blocking): _reclaimable_cache is keyed on pool_path alone, not on the repair mode, so a repair=False caller can be served a value computed by a repair=True walk (or vice versa). It is benign today because a corrupt marker contributes 0 to the sum either way, but if the two modes ever diverge numerically the cache would mask it. A repair flag in the key would make that impossible.

foxpatch-aleph (approved):
This PR is well-executed: the read-only contract of `storage status`/`list` is now enforced end-to-end (missing-DB refusal before any setup, `setup_pools(read_only=True)` that still validates but adopts nothing, `read_marker(repair=False)` threaded through `iter_reclaimable`/`reclaimable_bytes`), the `PurgeResult` API correctly separates device-mapper holds from failed removals so `reclaim` no longer gives unusable advice, and the naive-timestamp/UTC fix and `_age` clamp close two genuine crash and display bugs. I verified the breaking `purge_vm_storage` return-type change is safe (every other caller ignores the return), that `Settings.__init__` really is side-effect-free so skipping `setup()` loses nothing these verbs read, that `setup_pools(read_only=True)` still refuses a vanished pool marker, and that the new tests trace correctly through the fixtures. Tests cover each behavior change including the refusal-order property (a refused reclaim removing no unrelated marker) and the no-interpolation property. Docs match the code precisely. Only minor nits remain.

**`src/aleph/vm/agent/vm/purge.py`** (line 219): Nit (pre-existing, carried along by this change): `held` is recomputed inside the per-directory loop, and each recomputation re-walks every pool via `iter_volume_files(namespace)` — O(n²) over the VM's directories. Hoisting `held = [...]` above the loop would also let the loop body drop to a single `held_here` filter. Not a bug, and not introduced here, so non-blocking.

**`src/aleph/vm/agent/vm/reclaimable.py`** (line 398): Nit: the `_reclaimable_cache` is keyed on `pool_path` only, not on `repair`, so a `repair=True` call can be served a value cached by a `repair=False` walk (or vice versa). Today both modes sum identically (a corrupt marker reads as None and contributes 0 either way), so this is safe, but if `repair` ever changes the sum the cache would silently serve the stale mode. Within a single process the mode is never mixed in practice, so nothing to fix now — just worth remembering if the semantics of `repair` ever widen.

**`src/aleph/vm/agent/storage_cli.py`** (line 717): Minor observation: the old post-purge `_still_on_disk` check also caught a namespace directory that appeared after the purge's own snapshot (an extreme race); `result.kept` only reports what the purge itself saw. The replacement is semantically equivalent for every realistic case and strictly better information-wise (real reason per directory), so this is fine — just noting the tiny behavioral delta.

foxpatch-aleph (approved):
Solid, well-documented PR that genuinely makes status/list read-only and gives reclaim honest diagnostics. The marker repair threading, read-only pool setup, missing-DB-before-setup refusal, naive-timestamp UTC handling, and PurgeResult reporting are all correct; I verified the settings derivation in conf.py makes skipping settings.setup() safe, that all other purge_vm_storage callers ignore the return value, and that interpolate=False is supported by the pinned python-dotenv. All 130 tests across the three touched files pass; the remaining supervisor-suite failures in this environment are unrelated artifacts. Only minor nits: the mixed dm/OSError purge report advice, the cache key ignoring the repair flag, and a redundant local import in a test.

**`src/aleph/vm/agent/storage_cli.py`** (line 750): When a purge leaves a mix of device-mapper-held and OSError-kept directories, the closing advice only mentions 'storage reconcile'; the non-dm directories are named per-line above but get no place in the final 'then retry' sentence. Consider a combined message (e.g. also telling the operator to fix the removal errors) so the advice covers every kept directory.

**`src/aleph/vm/agent/vm/reclaimable.py`** (line 398): The reclaimable_bytes cache key does not record the repair flag. Harmless today because a corrupt marker contributes 0 to the sum in both modes, but it is a subtle invariant — worth a one-line comment saying both repair modes produce the same total.

**`tests/supervisor/test_storage_cli.py`** (line 1085): The local 'import errno' and 'import shutil as shutil_module' are redundant: this PR already adds both to the module's top-level imports. Harmless, but they can be dropped.

foxpatch-aleph (approved):
Solid, careful PR. Every behavioral claim checks out against the code: the repair=False threading through read_marker/iter_reclaimable/reclaimable_bytes keeps the reconciler's repair while making status/list/reclaim-check genuinely read-only; setup_pools(read_only=True) retains the vanished-marker refusal while adopting nothing; the DB refusal for read-only verbs now runs before any directory creation; the naive-timestamp-to-UTC coercion also fixes a latent TypeError in the reconciler's eviction sort; and the hoisted device-mapper hold computation in purge_vm_storage is behavior-preserving (computed after purge_vm_volumes, and one directory's rmtree cannot change another's dm state). All 132 tests in the touched files pass, mypy is clean on the changed sources, and the PurgeResult int-to-dataclass change is safe (all other callers discard the return). Only cosmetic nits remain.

**`src/aleph/vm/agent/storage_cli.py`** (line 745): Minor: when a purge keeps both kinds of directories (a dm-held one plus an OSError one), the trailing advice only addresses the dm-held subset. The per-directory reasons are printed so an operator can see the distinction, but the closing line is half-right in that mixed case; consider advising per reason class, or saying explicitly that the reconcile advice only covers the dm-held ones. Non-blocking.

**`src/aleph/vm/agent/vm/reclaimable.py`** (line 146): Nit: invalidate_reclaimable_cache() runs before the now best-effort unlink, so a failed unlink still invalidates the cache without any state change. Harmless (invalidation is never incorrect, only unnecessary), just noting the ordering.

**`tests/supervisor/test_storage_cli.py`** (line 1166): Nit: the test re-imports errno and shutil locally (import shutil as shutil_module) though both are already imported at module scope. Trivial.

foxpatch-aleph (approved):
This PR is correct and carefully built. The read_marker(repair=False) threading is complete for every marker read status/list/and a refused reclaim make, and the reconciler keeps the repair where it belongs; the naive-timestamp-as-UTC rule and the _age clamp are right; PurgeResult/KeptDirectory cleanly separate the two ways a directory survives a purge, and every existing caller of purge_vm_storage either uses the new shape or ignores the return; setup_pools(read_only=True) provably performs no writes (both the marker write and _record_adopted are gated, and the vanished-marker refusal still runs); the held-hoist out of the per-directory loop is behavior-preserving; interpolate=False is supported by the pinned python-dotenv, and skipping settings.setup() for the read-only verbs is safe because every path they read is validator-derived in conf.py, not created by setup(). Tests pin each new behavior at the exact point a regression would reappear, and the docs match the implementation. The observations below are minor and non-blocking.

**`src/aleph/vm/agent/vm/reconciler.py`** (line 619): Out of this PR's stated scope, but worth noting for a follow-up: `reconcile --dry-run` in the CLI process still unlinks corrupt markers, because the reconciler's own `iter_reclaimable()` (here) and `read_marker()` calls keep repair=True. A preview is the closest analogue of a listing, and it is the same 'listing that unlinks a file' pattern this PR removes from status/list. The read_marker docstring says the reconciler keeps the repair, and a real reconcile pass legitimately needs it — but a dry-run does not.

**`src/aleph/vm/agent/vm/reclaimable.py`** (line 405): Minor subtlety: the reclaimable_bytes cache is keyed on pool_path but not on repair, so a cached total computed by a repair=True walk is served to repair=False callers (and vice versa). This is benign today because a corrupt marker contributes 0 to the sum whether it is removed or ignored, but a one-line comment saying the totals are identical by construction would save a future reader from having to re-derive that.

**`src/aleph/vm/agent/storage_cli.py`** (line 745): When one purge keeps both kinds of directories (a dm-held one and an EROFS/EACCES one), the single advice line points at 'storage reconcile', which only fixes the dm one. The per-directory reasons printed above let the operator see the difference, and a retry re-reports, so this is fine — just flagging the mixed case as the one where the advice is only half right.

**`src/aleph/vm/storage_pools.py`** (line 210): Deliberate and documented tradeoff, but worth stating in review: under read_only a pool whose marker vanished after adoption still raises, so status and list refuse to run at all on a node with an unmounted second disk — the operator cannot even report on the pools that are healthy. Fail-closed is the right default here (reporting against the filesystem below an unmounted path would mislead), and the refusal message directs them to the real problem.

**`tests/supervisor/test_storage_cli.py`** (line 1119): Nit: the local `import errno` / `import shutil` here duplicate the module-level imports at lines 3 and 9 (same in the new test in test_vm_purge.py at test_vm_purge.py:357). Harmless, but the module-level imports are enough.

foxpatch-aleph (approved):
This PR delivers exactly what it promises: 'aleph-vm storage status' and 'list' are now truly read-only, refusing a missing DB before any setup, skipping settings.setup() and pool adoption, and reading reclaimable markers with repair=False so corrupt markers are ignored rather than unlinked. The purge path is made honest by PurgeResult, whose KeptDirectory.device_mapper flag lets 'storage reclaim' advise 'storage reconcile' only when a dm-target hold is actually the reason — instead of pointing operators at a fix that cannot help with EROFS/EACCES failures. I verified the int→PurgeResult change is backward compatible across all callers, that the device-mapper hoisting is behavior-preserving, that skipping settings.setup() is safe for everything status/list read, and that naive marker timestamps now read as UTC. The test plan is excellent and all ~188 tests pass locally, including fault-injection for unremovable directories and unmeasurable pools. The single non-blocking note concerns the reclaimable_bytes TTL cache, which is keyed on pool path + mtimes but not on the new 'repair' flag; it is safe today because corrupt markers contribute 0 to the total either way, but a code comment would spare future readers the re-derivation. No security, correctness, testing, or documentation issues found.

**`src/aleph/vm/agent/vm/reclaimable.py`** (line 398): Non-blocking observation: `_reclaimable_cache` is keyed on `pool_path` (plus the pool-mtime fingerprint) but not on the new `repair` flag, so a `repair=True` call can serve a later `repair=False` call within the 5s TTL. This is currently harmless — a corrupt marker contributes 0 to the total whether `read_marker` unlinks it or returns None — so both flag values produce identical sums. Since the safety of this conflation depends on that invariant, it may be worth a one-line comment here (e.g., "the total is identical for either repair value, so the cache needs no repair key"), or a note in the PR description, so a future change to read_marker's handling of corrupt files doesn't silently break it.

foxpatch-aleph (approved):
Solid, carefully scoped PR. The read-only guarantees are enforced at the right layers (marker repair made opt-in, pool adoption skipped, settings.setup() bypassed with the DB refusal hoisted before it), the naive-timestamp UTC fix and age clamping are correct, and PurgeResult is a clean interface whose return-type change breaks no existing caller. Tests genuinely pin the new behavior and I verified they pass alongside ruff and mypy. Only minor nits, neither blocking.

**`src/aleph/vm/agent/vm/reclaimable.py`** (line 77): Non-blocking, pre-existing: `reason` is read from the marker file without validation against the declared Literal ("gone"/"orphan"), so a hand-edited marker containing a tab or newline would corrupt the TSV output of `storage list`. Since this PR already hardens from_json against another hand-edit hazard (naive timestamps), a type/range check here would close the loop.

**`src/aleph/vm/agent/storage_cli.py`** (line 745): Cosmetic: when a purge leaves a mix of dm-held and OSError-kept directories, the report advises only `storage reconcile`, which fixes the dm one but not the others. The per-directory reasons printed above make this diagnosable, so this is fine as-is.

PR #1225, opened 2026-09-09
fix(agent): tell a push the truth about a VM whose teardown is running

odesenfans:
## Summary
A reconciler pass re-reads the plan right before it tears a VM down, so a push that arrives first is honoured, but the delete itself parks for as long as the VM takes to stop and the supervisor keeps listing the VM while it runs. A push that re-added that VM in the window was answered `unchanged`, read straight off that listing, while the retire went on to GONE and reaped the disks: the scheduler believed nothing had happened to a VM this node was destroying and would rebuild from scratch on the next pass.

The reconciler now keeps the hashes whose teardown is in flight, and the verdict consults them: a planned VM being removed is judged as a candidate rather than reported unchanged, so the answer is `accepted` (or `pending`, when the push carried no message to size it by), which is what the loop actually does about it once the delete returns.

Stacked on #1222.

## Changes
- `allocation/reconciler.py`: `_removing`, a set of the hashes whose teardown is running, added before the await in `_teardown_dropped` and cleared in a `finally` so a delete the supervisor refuses (retried on the next pass) leaves nothing behind. `removing_hashes()` exposes it to the answer.
- `allocation/verdict.py`: `compute_verdict` takes `removing_now` and skips the `unchanged` branch for any planned hash in it, so the VM falls through to the candidate loop and is answered accepted or pending. A VM the push does not name is untouched by the set and is still reported as `removing`.
- `views/allocations_v2.py`: the handler reads the set after its two awaits, in the same turn as the verdict, so a teardown that started while it was parked in `list_vms` is accounted for (the list it holds says that VM is running).

## Test plan
- `tests/supervisor/test_allocations_v2.py`: the whole scenario through HTTP. An empty plan starts the teardown, which parks in the supervisor's delete; the push re-adding the VM lands in that window and must answer `accepted` with an empty `unchanged`; once the delete returns the loop starts the VM again. Failed before the fix on `unchanged == []`.
- `tests/supervisor/test_allocation_reconciler.py`: the hash is reported removing only while the delete is parked, a teardown that raises leaves the set empty, and a VM re-added mid-teardown is started by the next pass.
- `tests/supervisor/test_allocation_verdict.py`: a planned VM in `removing_now` is accepted (or pending with no message) and never unchanged; one the push drops is still `removing`.
- Checks: 6 new tests failed before the fix; after it, the allocation, verdict, v2, teardown, verify and executions-list files pass (154), the wider set that imports the touched modules passes (414), and `tests/supervisor tests/migration` is `1 failed, 1237 passed, 1 skipped`, the one failure being the pre-existing journald `test_log.py::test_make_logs_queue`. `ruff format --diff`, `isort --check-only --profile black` and `mypy` on the three sources and the three test files: all exit 0. No Rust touched.

### DIFF
--- a/src/aleph/vm/agent/allocation/reconciler.py
+++ b/src/aleph/vm/agent/allocation/reconciler.py
@@ -95,6 +95,14 @@ class AllocationReconciler:
         self._wakeup = asyncio.Event()
         self._failures: dict[ItemHash, FailureRecord] = {}
         self._states: dict[ItemHash, AllocationState] = {}
+        # The VMs whose teardown is running right now. A delete parks for as
+        # long as the VM takes to stop, and the supervisor keeps listing it
+        # meanwhile, so a push arriving in that window reads a VM that is up
+        # and is in fact on its way to GONE with its disks reaped. The pass
+        # re-reads the plan before it starts a teardown, so a push that gets
+        # there first is honoured; past that point the retire cannot be
+        # called off, and the answer has to say so.
+        self._removing: set[ItemHash] = set()
 
     # ── Public surface ──
 
@@ -148,6 +156,16 @@ class AllocationReconciler:
         if self._desired is None:
             return set()
         return {vm_hash for vm_hash, planned in self._desired.entries.items() if planned.verified is None}
+
+    def removing_hashes(self) -> frozenset[ItemHash]:
+        """The VMs whose teardown is in flight, for the answer to a push.
+
+        A hash in here is one the supervisor still lists and this node is
+        already committed to destroying, so a push that re-adds it is asking
+        for the VM to be built again rather than left alone. Read by the
+        verdict, which must not answer "unchanged" for any of them.
+        """
+        return frozenset(self._removing)
 
     def state_for(self, vm_hash: ItemHash) -> tuple[AllocationState | None, FailureRecord | None]:
         """What the agent is doing about this VM, for the executions list."""
@@ -215,6 +233,11 @@ class AllocationReconciler:
             if record is None or not is_removable_by_allocation(record, info):
                 continue
             logger.info("Plan %s dropped %s; tearing it down", current.plan_id, vm_hash)
+            # Marked before the await and cleared however it ends, including
+            # on a delete the supervisor refuses: that one is retried on the
+            # next pass, and a hash left behind here would have every later
+            # push answered as a rebuild of a VM that is up and staying up.
+            self._removing.add(vm_hash)
             try:
                 await teardown_vm(vm_hash, supervisor=self.supervisor, registry=self.registry)
             except Exception:
@@ -223,6 +246,8 @@ class AllocationReconciler:
                 # before a single start ran, and teardowns carry no backoff, so
                 # it did that on every interval for as long as it kept failing.
                 logger.exception("Tearing down %s failed; leaving it for the next pass", vm_hash)
+            finally:
+                self._removing.discard(vm_hash)
 
     async def _start_missing(self, plan: AllocationPlan, known: dict[ItemHash, VmInfo]) -> None:
         live = {
--- a/src/aleph/vm/agent/allocation/verdict.py
+++ b/src/aleph/vm/agent/allocation/verdict.py
@@ -162,6 +162,7 @@ def compute_verdict(
     capacity: _Capacity,
     node_hash: str | None = None,
     available_gpus: list[GpuDevice] | None = None,
+    removing_now: frozenset[ItemHash] = frozenset(),
 ) -> PlanVerdict:
     """The immediate answer: what we take, what we drop, what we refuse.
 
@@ -170,6 +171,11 @@ def compute_verdict(
     the top of the module, so the handler reads them first and hands them in.
     Without them simulate refuses every candidate that asks for a card, which
     is the right answer to give when the cards were not looked at.
+
+    ``removing_now`` is the set of VMs the convergence loop is deleting as
+    this answer is computed. The supervisor goes on listing such a VM until
+    its delete returns, so its status alone would have this call report it as
+    running and untouched.
     """
     verdict = PlanVerdict()
     known = by_hash(infos)
@@ -185,7 +191,18 @@ def compute_verdict(
             # sizing it as a candidate would let a node that is tight on room
             # refuse a VM it is already holding, and a refusal is what takes
             # the VM out of the plan the loop converges on.
-            if info.status in LIVE_STATUSES or info.status in STOPPED_STATUSES or info.awaiting_confidential_init:
+            #
+            # A VM whose teardown is already running is the exception, however
+            # alive or however stopped the supervisor says it is: the retire is
+            # past the point where a push can call it off, so the VM is going
+            # away with its disks and this node will have to build it again.
+            # Judged as a candidate instead, it is answered accepted, or
+            # pending while the push carried no message to size it by, which is
+            # what the loop will actually do about it on the pass after the
+            # delete returns.
+            if (
+                info.status in LIVE_STATUSES or info.status in STOPPED_STATUSES or info.awaiting_confidential_init
+            ) and vm_hash not in removing_now:
                 unchanged.add(vm_hash)
                 verdict.unchanged.append(vm_hash)
             # A planned VM the supervisor holds dead is about to be created
--- a/src/aleph/vm/agent/views/allocations_v2.py
+++ b/src/aleph/vm/agent/views/allocations_v2.py
@@ -58,6 +58,7 @@ async def update_allocations_v2(request: web.Request) -> web.Response:
     # invalidate the answer about to be returned; see allocation.verdict.
     infos = await app["supervisor"].list_vms()
     available_gpus = await app["capacity"].available_gpus()
+    reconciler = app["allocation_reconciler"]
     verdict = compute_verdict(
         plan,
         infos=infos,
@@ -65,9 +66,13 @@ async def update_allocations_v2(request: web.Request) -> web.Response:
         capacity=app["capacity"],
         node_hash=node_identity.get_node_hash() if node_identity else None,
         available_gpus=available_gpus,
+        # Read after the two awaits, in the same turn as the verdict, so a
+        # teardown that starts while this handler is parked in list_vms is
+        # still accounted for: the list it holds says that VM is running.
+        removing_now=reconciler.removing_hashes(),
     )
     verdict.rejected.update(rejected)
-    app["allocation_reconciler"].submit(narrow_plan(plan, verdict))
+    reconciler.submit(narrow_plan(plan, verdict))
 
     logger.info(
         "Plan %s: %d accepted, %d pending, %d unchanged, %d rejected, %d removing, %d retained",
--- a/tests/supervisor/test_allocation_reconciler.py
+++ b/tests/supervisor/test_allocation_reconciler.py
@@ -794,6 +794,97 @@ async def test_a_vm_waiting_on_its_confidential_session_is_not_started_again(rec
     assert starts.attempts == 0
 
 
+# ── A teardown that is still in flight ─────────────────────────────────────
+
+
+def _park_the_teardown(reconciler, monkeypatch):
+    """Hold the teardown of a VM inside the supervisor's delete.
+
+    A real delete takes as long as the VM takes to stop, and that is the
+    window a re-push lands in. Returns the event that lets it finish and the
+    one that says it has started.
+    """
+    entered = asyncio.Event()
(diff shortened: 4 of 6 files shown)


foxpatch-aleph (approved):
Correct, minimal fix for a real bug: a push re-adding a VM mid-teardown was answered "unchanged" because the supervisor still lists the VM as running while its delete is parked, so the scheduler believed nothing had happened to a VM this node was destroying. Tracking in-flight teardown hashes in the reconciler and consulting them in compute_verdict is the right seam, and the guard is wired in the same event-loop turn as the verdict and submit(), so no interleaving can skew the answer. Verified end-to-end: all 154 allocation tests pass, ruff/isort/mypy are clean, and reverting just the verdict condition makes the new tests fail, confirming they guard the behavior. The three test layers (pure verdict, reconciler, HTTP with the real background loop) give good confidence, and the teardown-parks-in-delete simulation matches production reality. One non-blocking note: a teardown that completes while the handler is parked in list_vms() (between reading infos and reading removing_hashes()) is still answered unchanged for a VM that just vanished — the mirror image of the fixed race, but the window shrinks from the VM's full stop time to the handler's two awaits, and the next pass rebuilds the VM anyway since the plan re-lists it.

**`src/aleph/vm/agent/views/allocations_v2.py`** (line 72): Reading removing_now after the two awaits is the right call, but note the residual window in the other direction: a teardown that *completes* while this handler is parked in list_vms() or available_gpus() has already been discarded from _removing, so the answer is still "unchanged" for a VM that just disappeared. This is much narrower than the bug being fixed (the handler's two awaits vs. the whole delete) and self-corrects — the submitted plan still lists the VM, so the next pass rebuilds it — so it's fine to leave as is, but worth knowing the answer is best-effort in that direction.

foxpatch-aleph (approved):
Correct and well-guarded fix for a real race: a push re-adding a VM whose teardown is in flight was answered unchanged even though the delete was committed and would retire the VM GONE with its disks reaped. Tracking the in-flight hashes in the reconciler (added before the await, discarded in a finally) and having the verdict consult them is sound under asyncio: the add shares a turn with the plan re-read, the handler reads the set after its awaits and in the same turn as submit(), and the refused-delete path leaves nothing stale behind. Judging such a VM as a candidate also interacts correctly with capacity simulation, and a VM the push drops is still answered removing. I verified the touched and adjacent suites pass (146 tests), ruff format and mypy are clean, and that reverting the source changes makes the meaningful new tests fail. Two non-blocking notes: a sliver remains where a teardown completing between the handler's list_vms and its removing_now read still answers unchanged, but the level-triggered loop rebuilds the VM on the next pass so it self-corrects; and one of the new reconciler tests (test_a_vm_re_added_while_it_was_being_torn_down_is_started_again) passes with the fix reverted, so it is a regression guard rather than a pin of the fix.

**`src/aleph/vm/agent/views/allocations_v2.py`** (line 72): Reading removing_now after the two awaits and in the same turn as compute_verdict and submit() is the right call: nothing can mutate _removing in between. Note for completeness that a teardown that completes between list_vms and this read is still answered unchanged — the VM is dead but the snapshot said RUNNING and _removing has already been discarded. The next pass finds the VM missing from the supervisor's list and rebuilds it, so this sliver self-corrects exactly like a VM dying right after the list read; not worth closing.

**`tests/supervisor/test_allocation_reconciler.py`** (line 923): This test passes with the fix reverted (it exercises the loop's half of the promise, which was already correct), so it is a regression guard rather than a pin of the fix — unlike its sibling test_a_teardown_in_flight_is_visible_while_it_runs, which fails without the set. Not a problem, just worth knowing when reading the PR description's failure count.

foxpatch-aleph (approved):
This PR fixes a real and subtle race: a push re-adding a VM whose teardown delete was parked in the supervisor used to be answered `unchanged` off a stale VM listing, telling the scheduler nothing had happened to a VM the node was about to destroy and rebuild from scratch. Tracking in-flight teardowns in the reconciler and consulting them in the verdict is the right fix, and the concurrency reasoning is airtight: `_removing.add()` sits in the same event-loop turn as the plan re-read, the handler reads `removing_hashes()` in the same await-free turn as `compute_verdict` and `submit()`, and the `finally` discard keeps failed deletes from poisoning later answers. I verified that `CapacityManager.simulate` excludes a candidate's own registry record and volumes, so a mid-teardown VM sized as a candidate cannot refuse itself on its own still-committed record. All 158 allocation/teardown/verify/executions tests pass, mypy is clean, and `ruff format` is clean; the wider-suite failures in this sandbox are environmental (permission and library-version errors) in files this PR does not touch. Tests are thorough — the end-to-end HTTP test pins the exact scenario, and the reconciler tests pin set membership to the parked window. Only non-blocking note: `removing_now` reflects only teardowns the reconciler's own pass ordered, so retires from other paths (expiry RECREATE, update watcher, the legacy endpoint) can still be answered `unchanged` mid-delete — materially less harmful since RECREATE keeps disks and the legacy route is a separate plane, but worth keeping in mind if those planes converge.

**`src/aleph/vm/agent/allocation/reconciler.py`** (line 160): Non-blocking observation: `removing_hashes()` reflects only teardowns this reconciler's own pass ordered. Other retire paths can also be mid-delete while a push is answered — the expiry manager's RECREATE (`expiry.py:58`), `update_watcher.py:93`, and the legacy `/control/allocations` endpoint (`views/__init__.py:635`) — and a VM being retired through those is still reported `unchanged`. The window is materially less harmful there (RECREATE keeps disks, the legacy route is a separate plane), so this scoping looks right for this PR, but a word in the docstring saying the set is about the reconciler's own teardowns would prevent it being read as exhaustive.

**`src/aleph/vm/agent/views/allocations_v2.py`** (line 72): Verified the ordering claim that makes this correct: reading `removing_hashes()` after the two awaits but in the same await-free turn as `compute_verdict` and `submit` means any teardown either began (hash present in the set) or its re-read of `self._desired` will see the just-submitted plan and skip. No interleaving is possible on a single event loop. Might be worth adding a half-sentence to the comment noting the symmetric invariant in the reconciler (`reconciler.py:229-240`), since the two halves only close the race together.

foxpatch-aleph (approved):
A well-scoped fix for a real race: while a delete parks in the supervisor, the VM stays listed, so a push re-adding it was answered "unchanged" for a VM about to be destroyed and rebuilt. Tracking teardowns in an in-flight set, guarded so the add is await-adjacent to the plan re-read and the discard lives in a finally, and consulting it from the verdict within the handler's await-free critical section is airtight under asyncio; the scoping is precise (only RUNNING/STOPPED VMs change classification, BOOTING/DEFINED are never in the set), and capacity accounting stays conservative. Tests cover the fix at three levels, including a full HTTP round-trip with the real reconciler loop, and all pass along with the surrounding suites; ruff and mypy are clean. The only nit is a doc comment that slightly understates the possible outcomes.

**`src/aleph/vm/agent/allocation/verdict.py`** (line 199): Minor doc nit: this comment (and the PR description) says such a VM is answered accepted or pending, but the candidate path can also reject it — via capacity.simulate (insufficient_capacity) or a node-hash pin — after which narrow_plan drops it from the submitted plan so it is not rebuilt here. That is still an honest answer (the node will not run it after the delete returns), so purely a wording nit; consider saying "accepted, pending, or honestly refused".

foxpatch-aleph (approved):
This PR correctly fixes a real race: a push re-adding a VM mid-teardown was answered `unchanged` because the supervisor still lists the VM during its (long) delete, so the scheduler believed nothing had happened to a VM the node was about to retire GONE with disks reaped. The fix is minimal and well-placed: an in-flight set marked before the await and cleared in `finally` (so a failed delete retried on the next pass cannot leave a stale entry mis-judging every later push), exposed via `removing_hashes()` and read by the handler in the same event-loop turn as the verdict and `submit()`, which excludes such VMs from the `unchanged` branch so they are honestly answered `accepted`/`pending`. I verified all 118 tests in the three touched files pass, ruff/isort/mypy are clean, and the new HTTP regression test fails at exactly the `unchanged == []` assertion when the handler wiring is reverted. The residual window (a teardown completing while the handler is parked in `available_gpus()`) is irreducible given a snapshot answer, bounded, and self-healing on the next pass. Comments and docstrings match the codebase's unusually thorough house style.

**`src/aleph/vm/agent/views/allocations_v2.py`** (line 72): Verified the ordering matters and is right: reading `removing_hashes()` after the two awaits catches a teardown that starts while the handler is parked in `list_vms()`. Note there remains a small, irreducible residual window — a teardown that *completes* while the handler is parked in `available_gpus()` leaves the handler holding a stale `infos` listing and an already-cleared set, so that VM is answered `unchanged` even though it just vanished. It is bounded, self-healing (the plan keeps the VM, so the next pass rebuilds it), and strictly better than the pre-fix behavior, so no change requested — just worth knowing the snapshot cannot be made fully airtight.

**`src/aleph/vm/agent/allocation/reconciler.py`** (line 249): The `finally` here is the load-bearing detail: without it, a delete the supervisor refuses (retried next pass) would leave the hash in `_removing` forever, and every later push would size that up-and-staying-up VM as a candidate — where a capacity rejection would wrongly evict it from the plan. Confirmed the dedicated test covers this.

foxpatch-aleph (approved):
A correct and well-tested fix for a real race: a push re-adding a VM mid-teardown was answered "unchanged" off a supervisor listing that still shows the VM running. The in-flight set is maintained soundly (added before the await, discarded in a finally, so a failed or cancelled delete leaves nothing behind), the handler reads it after its awaits in the same turn as the verdict and submit() so the answer and recorded plan stay consistent, and capacity accounting remains correct because simulate() discounts a candidate's own registry record. I verified the new end-to-end test fails without the verdict change, that all 158 allocation-related tests pass, that the event-gated tests are not flaky across repeated runs, and that isort/mypy are clean; the wider-suite failures I saw are environmental (sandbox /var/cache permissions and a stale aleph_message package), not caused by this PR. One residual, non-blocking observation: if a teardown completes in the millisecond window between the handler's list_vms() and removing_hashes() (which spans the available_gpus() await), the answer can still say "unchanged" for a VM already gone — but that case self-heals since the VM stays in the narrowed plan and the next pass rebuilds it, unlike the bug fixed here.

**`src/aleph/vm/agent/views/allocations_v2.py`** (line 72): Non-blocking observation: a residual race remains. If a teardown *completes* between list_vms() returning and removing_hashes() being called (the window spans the available_gpus() await), the stale infos still yield "unchanged" for a VM that is already destroyed. Unlike the bug fixed here, that case self-heals — the VM stays in the narrowed plan and the next pass rebuilds it — and the window is milliseconds rather than the teardown's full duration, so I don't think more machinery is warranted. Worth a comment noting the boundary if you want to be thorough.

**`src/aleph/vm/agent/allocation/verdict.py`** (line 205): Nit: `removing_now` vs `verdict.removing` is a subtle naming pair — one is teardowns in flight, the other VMs this push stops. The docstring disambiguates them well, so this is fine as-is; just flagging it in case a future reader conflates the two.

foxpatch-aleph (approved):
The fix correctly closes a genuine race: a push re-adding a VM during its teardown was answered unchanged off a supervisor listing that still reports the VM as running, while the retire completed and reaped the disks. Tracking in-flight teardowns in a set (add before the await, discard in a finally so refused deletes leave nothing behind) and consulting it in the verdict makes the answer honest, and judging the VM as a candidate is capacity-safe because simulate discounts a candidate's own registry record. I verified the concurrency reasoning (no awaits between the removing_hashes read, the verdict and submit, so no interleaving gap on a single loop), ran the full touched-area suites (147 passed), reverse-applied the fix to confirm the new tests fail without it, and confirmed ruff, isort and mypy are clean. Two non-blocking observations below; otherwise this is a correct, minimal, well-documented and thoroughly tested change.

**`src/aleph/vm/agent/views/allocations_v2.py`** (line 72): Non-blocking observation: one residual window remains. If a teardown *completes* during the `available_gpus()` await — i.e. after `list_vms()` was read but before `removing_hashes()` — the hash is no longer in the in-flight set while `infos` still lists the VM as running, so the answer is `unchanged` for a VM that was just retired GONE. The loop still converges (the plan names it, so the next pass rebuilds it), making this the same snapshot staleness that applies to any state change across the two awaits, so I don't think it needs code — but a sentence in the comment noting the read is also only a snapshot could help a future reader who assumes the answer is exact.

**`src/aleph/vm/agent/allocation/reconciler.py`** (line 105): Minor observation: since `_teardown_dropped` awaits each teardown sequentially within a pass, this set holds at most one hash in production. The set is still the right shape (idempotent `discard` and it survives a future move to concurrent teardowns), so no change needed — just noting the invariant in case someone reads it as evidence teardowns can overlap.

PR #1226, opened 2026-09-09
test(agent): record why the fake collector is never the exec target

odesenfans:
## Summary
The intermittent `Text file busy (os error 26)` in the attest-agent's GPU collector tests is already fixed on `dev-2.1`: #1200 changed `script_collector` to run each fake collector through `/bin/sh <script>` instead of exec'ing the written script itself, which removes the failure at its source. The retry helper this PR originally added is therefore no longer reachable, and it has been dropped.

What is left worth having is the reason, stated correctly, so that nobody shortens the command line back to the script's own path and brings the flake straight back. The account this PR shipped with was wrong in a way that made the `/bin/sh` indirection look optional: it claimed an inherited writable descriptor makes ANY exec in the process fail. ETXTBSY is per inode. `exec` refuses a file that someone still holds open for writing, counting writers of the inode being exec'd, not writers of anything anywhere. That is exactly why exec'ing `/bin/sh` cannot fail here (nothing in the suite ever opens `/bin/sh` for writing, so its writer count is always zero) while exec'ing a just-written script races the fork window of every other test thread.

Stacked on #1221.

## Changes
- `rust/crates/aleph-attest-agent/src/gpu.rs`: the doc comment on `script_collector` now records that the written file is deliberately never the exec target and why, with the per-inode semantics spelled out and the two observed failures (CI and locally) named. Removed the `collect_retrying_etxtbsy` helper and its six call sites: a retry cannot fire against an inode whose writer count is always zero, and keeping it would suggest the `/bin/sh` indirection is belt and braces rather than the fix.
- `nix/golden-measurements.json`: re-seeded. A comment inside `gpu.rs` is still a source edit of the crate the measured binary is built from, so the whole chain moves. Produced by `nix/check-golden-measurements.sh --update` at this branch's tip, then verified by a plain run of the same script, which prints "Golden measurements match."

## Test plan
- No behaviour under test changed: this is a comment plus the removal of a wrapper that only ever forwarded a successful call. `cargo clippy --locked --all-targets --manifest-path crates/aleph-attest-agent/Cargo.toml -- -D warnings` exit 0 type-checks every test call site.
- `cargo fmt --check --manifest-path crates/aleph-attest-agent/Cargo.toml`: exit 0.
- `nix/check-golden-measurements.sh` (no flag) after the re-seed: "Golden measurements match."
- The test suites were not run locally this round; CI runs them.

## Note on reproducibility
The wart the earlier round found still stands and is still worth its own issue: the measured attest-agent binary is not invariant to source changes confined to `#[cfg(test)]`, nor to comments. rustc parses the whole file before cfg-stripping and some of that per-file bookkeeping reaches the compiled binary, so every edit to the agent crate forces a re-seed of all five measurements. Worth investigating whether stripping or a different debug-info level makes the release binary invariant to test-only edits.


foxpatch-aleph (approved):
A correctly diagnosed and correctly scoped flake fix. The cross-test fork-inherited writable fd → ETXTBSY race is real and matches the known cargo/rustc spawn flake, and retrying on exactly ETXTBSY (surfacing through the anyhow chain via chain().find_map(downcast_ref::<io::Error>())) is the standard mitigation. All six spawn-based test call sites route through the helper, retries are side-effect-free since a failed spawn did nothing, the expected-error tests can never look like ETXTBSY, and no non-test code or Cargo.lock changed so the measured initrd is untouched. Two non-blocking nits below concern one inaccurate sentence in the new doc comment and a hardcoded errno.

**`rust/crates/aleph-attest-agent/src/gpu.rs`** (line 238): This sentence overstates the mechanism: ETXTBSY is per-inode on Linux: `deny_write_access()` rejects only execs of the specific file that is open for writing. So the exec that fails is precisely the one targeting the script whose `fs::write` was still in flight (the fd the other test's fork inherited) — not any exec regardless of target. The retry is right either way, but as written the comment could lead a future reader to apply this workaround more broadly than the failure mode warrants. Consider rewording to say any test's spawn can be rejected when its own script's write was inherited mid-flight.

**`rust/crates/aleph-attest-agent/src/gpu.rs`** (line 246): Minor: `libc` is already a direct dependency of this crate, so `libc::ETXTBSY` could replace the hardcoded 26. Test-only code, so either is fine.

foxpatch-aleph (approved):
A correct, minimal, test-only deflake: the retry helper walks the anyhow chain to the wrapped io::Error and only retries on ETXTBSY, which in collect() can only originate from the spawn itself — a failed spawn has no child, so a retry never double-spawns a collector and cannot mask the tests' real failure modes (timeout, exit status, output cap). All six subprocess-spawning tests are covered, production code is untouched, and the timeout tests' elapsed-time assertions keep a large margin over the added ~50ms worst-case retry overhead. The golden-measurement re-seed matches the repo's documented same-PR convention. Two non-blocking nits: The doc comment's last sentence overstates the blast radius: ETXTBSY is a per-inode check (deny_write_access() counts writers of the file being exec'd), so an inherited writable descriptor only fails execs of that specific script, not 'ANY exec... regardless of which file it targets' — worth rewording given the PR's whole point is to document this race precisely. And the crate already depends on libc, so libc::ETXTBSY would be clearer than the hard-coded 26.

**`rust/crates/aleph-attest-agent/src/gpu.rs`** (line 238): This last sentence misstates the kernel behavior: the ETXTBSY check (deny_write_access) is per-inode: it counts writers on the file being exec'd, so an inherited writable descriptor on script X only fails execs of script X — which in this suite is always the affected test's own spawn, exactly matching the observed CI failures in collect_handles_large_output_without_deadlock and collect_rejects_output_beyond_the_cap. 'ANY exec in the process (regardless of which file it targets)' is not accurate; execs of unrelated files (/bin/sh, other tests' scripts) are unaffected. The fix is still right, but since this comment exists specifically to document the race precisely, it would be worth rewording to say the inherited descriptor keeps a write count on that one script until the forking child execs, so only that script's exec fails.

**`rust/crates/aleph-attest-agent/src/gpu.rs`** (line 246): Nit: libc is already a direct dependency of this crate, so `libc::ETXTBSY` would be clearer and self-documenting compared to the hard-coded 26 (alternatively, `err.kind() == std::io::ErrorKind::ExecutableFileBusy`). Non-blocking; the value is correct on Linux.

odesenfans:
Closing: #1200 (merged into dev-2.1) already removed the ETXTBSY exposure at its root by never exec'ing the written script, and its comment on script_collector already records why. What remained here after the rebase was a longer version of that comment plus a golden re-seed forced by the source-hash embedding, which is not worth moving every measurement for.

PR #1227, opened 2026-09-09
fix(daemon): bound the GPU MMIO window sizing and its address budget

odesenfans:
## Summary
Sizing a confidential VM's 64-bit PCI MMIO window from the card's sysfs BARs did unchecked arithmetic and had no upper bound. A BAR line whose `end` is below its `start` was skipped in silence (which under-sizes the window and leaves the card's BARs unassigned in the guest), and a line spanning the whole address space (`start` 0, `end` `0xffffffffffffffff`) made `end - start + 1` overflow: a panic in a debug build, a wrapped total in release. Separately, nothing checked the resulting window against the address space the guest can actually reach: an SEV-SNP guest launches with `EPYC-v4`, whose default physical address width is 40 bits, so the firmware has 1 TiB to place everything in. The 512 GiB window a 128 GiB BAR1 card asks for lands exactly at that ceiling, but a larger SKU asks for one the firmware cannot place at all, and the guest then sees a device that enumerates and does nothing.

Stacked on #1223.

## Changes
- `rust/crates/supervisor-daemon/src/gpu_bar.rs`: `parse_resource_file` validates and sums only the BARs that count, with `checked_sub`/`checked_add`, and rejects a reversed range, a region size that overflows 64 bits, and a running total that does. The 64-bit prefetchable flag mask is a module constant instead of being rebuilt on every line. New `check_mmio64_budget(window_mb, guest_ram_mb)` models where the firmware puts the window (above the RAM, aligned to the window's own size, both powers of two) and refuses anything whose top passes the 1 TiB the guest can address; the two constants carry the reasoning for the ceiling and for the low-memory allowance.
- `rust/crates/supervisor-daemon/src/error.rs`: two typed variants, `GpuBarRange { line, reason }` and `GpuMmioBudget { window_mb, guest_ram_mb, top_mb, budget_mb }`, so the refusal names every number rather than formatting a string at the call site.
- `rust/crates/supervisor-daemon/src/lifecycle.rs`: `snp_config_slice_with` runs the budget check on the sized window against `spec.memory_mib` and returns `InvalidBackend` when it does not fit. The arithmetic stays general (the window is still summed over however many cards it is given) even though #1223 caps a confidential VM at one card.

## Test plan
- `gpu_bar.rs`: `a_reversed_bar_range_is_an_error`, `a_bar_spanning_the_whole_address_space_is_an_error_not_a_panic` and `a_bar_total_that_overflows_is_an_error` pin the three arithmetic cases (the middle one was a real panic before, at `end - start + 1`); `the_budget_takes_one_cards_window_next_to_ordinary_ram`, `the_budget_refuses_a_window_the_guest_cannot_address` and `the_budget_refuses_ram_that_pushes_the_window_over_the_ceiling` pin the budget in both directions, including the case where the window fits on its own but the RAM forces it up to the next alignment boundary.
- `lifecycle.rs`: `snp_config_slice_refuses_a_window_the_guest_cannot_address` pins the create-time refusal and that the message carries the numbers. The pre-existing `snp_config_slice_accepts_a_cc_mode_gpu_and_sizes_the_window` and `snp_config_slice_takes_a_gpu_on_the_verity_arm` are the controls: the 512 GiB window a real card asks for still passes.
- All new tests were watched red first (three against the restored old parse loop, one against the fix removed from the call site).
- Checks from `rust/`: `cargo fmt --check` exit 0, `cargo clippy --locked --all-targets -- -D warnings` exit 0, `cargo test --locked -p supervisor-daemon` exit 0 (393 lib + 1 bin + 1 grpc + 2 read_only passed, 0 failed).

## Rebased onto dev-2.1 (#1199 + #1200), plus review fixes

#1200 added its own ceiling to `mmio64_window_mb`: a 4 TiB `MAX_WINDOW_MB` clamp with checked rounding, so an absurd BAR total cannot wrap into a tiny window. Both bounds are kept and the relationship between them is now written down. The clamp keeps the number handed to fw_cfg finite; `check_mmio64_budget` is the tighter and later gate that actually refuses the create, and today's guest address space is a quarter of the clamp, so a window reaching the clamp is refused rather than launched. New test `a_window_that_hits_the_clamp_is_refused_by_the_budget` pins that. The BAR validation and the address budget merged with no textual conflict.

Review fixes, as one commit:
- `GUEST_PHYS_MB`: a spec-supplied `cpu_model` does not move the 1 TiB figure. The confidential argv passes the model name alone, with no `phys-bits` and no `host-phys-bits`, and QEMU's default for a named x86 model is 40 bits whichever model it is (verified against `supervisor-controller/src/qemu.rs`). If a future argv widens it, this check can only over-refuse, which fails a create rather than booting a card into a guest that cannot reach it.
- `check_mmio64_budget`: the placement model is named at its source, `OvmfPkg/Library/PlatformInitLib` (`PlatformDynamicMmioWindow`, `PlatformAddressWidthFromCpuid`), so the arithmetic can be re-verified rather than trusted.

Not taken: the `saturating_add` across cards, which #1236 further up the stack already makes checked.

Checks re-run after the rebase: `cargo fmt --check` exit 0, `cargo clippy --locked --all-targets -- -D warnings` exit 0. Test suites left to CI this round.


foxpatch-aleph (approved):
This PR fixes real arithmetic hazards (a debug panic / release wrap on a full-span BAR, silent under-sizing on reversed ranges) and adds a missing address-space budget check that prevents booting confidential VMs with a dead GPU. The checked arithmetic in parse_resource_file is correct and fails closed; check_mmio64_budget's align-up math is sound with every saturating operation biased toward refusal, and its boundary (a window ending exactly at 1 TiB is allowed) is right for a 40-bit guest. The 1 TiB figure matches QEMU's default for named CPU models, the single enforcement site covers the only path that sets pci_mmio64_mb, and the tests pin the former panic, all three overflow cases, both budget directions, and the create-time wiring while keeping realistic windows passing. Only non-blocking nits remain.

**`rust/crates/supervisor-daemon/src/gpu_bar.rs`** (line 26): The 1 TiB budget is applied even when the spec names a different cpu_model (snp_config_slice_with passes tee.cpu_model through). That is safe today because QEMU gives every named (non-host) model the same 40-bit default, and it can only over-refuse, never under-refuse — but a one-line note that a spec-supplied cpu_model does not change this figure (and why) would help the next reader. Non-blocking.

**`rust/crates/supervisor-daemon/src/gpu_bar.rs`** (line 94): The placement model (window sits above RAM, aligned to its own size) is an empirical fact about the firmware, and the ceiling check silently drifts if OVMF ever changes it. Worth citing the OVMF source that implements this placement (OvmfPkg PlatformInitLib) so the next maintainer can re-verify the model. The doc comment is otherwise thorough. Non-blocking.

**`rust/crates/supervisor-daemon/src/gpu_bar.rs`** (line 127): This saturating_add across cards is now unreachable saturation: parse_resource_file errors on a total that overflows, so the cross-card sum can never saturate. Harmless, but a checked_add returning a GpuBarRange error would be consistent with the rest of the module. Non-blocking.

foxpatch-aleph (approved):
This PR is correct and its central claims verify against the real firmware and hypervisor sources. I confirmed that a named x86-64 model (EPYC-v4, no phys-bits/host-phys-bits) yields a 40-bit guest physical width in QEMU, that reduced-phys-bits is NOT subtracted from CPUID 0x80000008 (it appears only in 0x8000001F, which OVMF's PlatformAddressWidthFromCpuid ignores), and that OVMF's AmdSev build — the one this project ships — places the fw_cfg-overridden window at align_up(max(top of RAM, 4 GiB), 1 GiB) naturally aligned to power-of-two(size), with no bounds check against the address space in the override path. The check's model, align_up(ram + 4 GiB, window) + window <= 1 TiB, is therefore exact and conservative, and the inclusive boundary (512 GiB window + 64 GiB RAM ending exactly at 1 TiB) is right. The BAR validation fixes a real debug-build panic (end - start + 1 overflow) and closes a silent under-sizing path; the typed errors, the documented relationship between the 4 TiB clamp and the 1 TiB budget, and the test suite (which I verified case-by-case by hand, since cargo is unavailable here) are all sound. Two non-blocking comments below.

**`rust/crates/supervisor-daemon/src/gpu_bar.rs`** (line 49): Verified against QEMU and OVMF: this constant is right for the placement model. One thing worth adding to the doc comment: the Linux guest kernel reduces its own usable physical width by the SEV value reported in CPUID Fn8000_001F EBX[11:6] (arch/x86/kernel/cpu/amd.c: `c->x86_phys_bits -= (cpuid_ebx(0x8000001f) >> 6) & 0x3f`), which is 1 on the current EPYC host — so the kernel-side budget is 39 bits (512 GiB), half of this figure. The kernel does not reject PCI BARs based on x86_phys_bits, so nothing breaks today, but the real card's BARs sit at [512 GiB, 640 GiB), right on that line, and a sentence noting that OVMF's budget is the binding constraint (and that the kernel's is tighter) would help whoever debugs a boundary report.

**`rust/crates/supervisor-daemon/src/error.rs`** (line 57): Nit: the Display text "does not describe an addressable region" fits the first two GpuBarRange reasons but not the third ("it pushes the BAR total past 64 bits"), which is a statement about the running total rather than the line. Something like "invalid PCI resource line {line:?}: {reason}" would read correctly for all three. Purely cosmetic.

foxpatch-aleph (approved):
This PR is correct and unusually verifiable. The unchecked sysfs BAR arithmetic is now checked with typed errors (the 0/0xffffffffffffffff line was a real panic at end - start + 1), and the new address-space budget rests on claims I verified against upstream sources rather than taking on faith: QEMU's default physical width for a named model with no phys-bits/host-phys-bits is 40 bits (cpu.c falls back to TCG_PHYS_ADDR_BITS, host-cpu.c only overrides under host-phys-bits), SEV's reduced-phys-bits is only surfaced via CPUID 0x8000001F EBX[11:6] and never shrinks the width OVMF derives from 0x80000008, and OVMF's PlatformGetFirstNonAddress places the fw_cfg-override window at align_up(top-of-RAM, its own size) with no validation against the address space (PlatformDynamicMmioWindow returns early on the override path), so the daemon-side refusal guards a real, otherwise-unguarded failure mode. The 4 GiB low-memory allowance always over-models the real top of RAM (4 GiB + high RAM), so every approximation can only over-refuse, never under-refuse; hostile memory_mib saturates into a refusal rather than a panic. Test arithmetic all checks out by hand, including the exact-1-TiB boundary and the clamp-vs-budget interaction. Two comment-level nits below.

**`rust/crates/supervisor-daemon/src/gpu_bar.rs`** (line 115): Nit: the OVMF function cited here doesn't implement the placement model for this daemon's windows. `PlatformDynamicMmioWindow` is the *dynamic* path and returns early when `PcdPciMmio64Override` is set (which it always is when `opt/ovmf/X-PciMmio64Mb` is present); the fw_cfg-override window this daemon produces is actually placed by `PlatformGetFirstFirstNonAddress` -> `PlatformGetFirstNonAddress` in `OvmfPkg/Library/PlatformInitLib/MemDetect.c` (ALIGN(top_of_ram, SIZE_1GB) then ALIGN(base, GetPowerOfTwo64(size))). Since the stated purpose of this comment is letting a reader re-verify the arithmetic, pointing at the function that no-ops on the override path undermines that. `PlatformAddressWidthFromCpuid` is cited correctly.

**`rust/crates/supervisor-daemon/src/gpu_bar.rs`** (line 45): Nit: the widening scenario is reachable today, not only by a future argv. `cpu_model` is interpolated verbatim into `-cpu {model}` (`snp_tee_fragment`), so a spec supplying `"host"` or `"EPYC-v4,host-phys-bits=on"` already widens the guest's address space past 40 bits, and this check would over-refuse windows the guest could address. That is exactly the fail-closed direction the comment describes, so nothing to fix — but "whichever model it is" reads slightly stronger than the model-string passthrough actually guarantees. Consider mentioning that a comma-suffixed model string is one of the forms that can widen the space.

foxpatch-aleph (approved):
Verified the placement model against OVMF master (OvmfPkg/Library/PlatformInitLib/MemDetect.c): with the fw_cfg override set, the window base is ALIGN_VALUE(top_of_RAM_floored_at_4GiB, GetPowerOfTwo64(size)) and OVMF does not itself bound it to the address width, so the daemon-side budget check is genuinely necessary and the model in check_mmio64_budget matches, with the 4 GiB low allowance over-approximating (fail-closed). The 40-bit/1 TiB guest width is confirmed: CPUID phys-bits defaults to 40 for named models, snp_tee_fragment passes the model name alone, and PlatformAddressWidthFromCpuid applies no SEV width reduction on this path. The parse-loop reordering (skip the flag check before range validation) preserves prior behavior for lines that do not count, and the checked arithmetic turns a real debug-build panic into typed errors that name the offending line. All new tests' arithmetic hand-checks pass (no cargo in this review environment), including the exact-boundary acceptance of the real 512 GiB window + 64 GiB RAM case and the clamp-vs-budget interplay. Two non-blocking documentation nits below.

**`rust/crates/supervisor-daemon/src/gpu_bar.rs`** (line 114): Small doc-precision nit: for the fw_cfg-override path — which is always the daemon's case — `PlatformDynamicMmioWindow` short-circuits (`using fwcfg override`) and does no placement work; the align-above-RAM arithmetic this function models lives in `PlatformGetFirstNonAddress` in the same file (`PcdPciMmio64Base = ALIGN_VALUE(FirstNonAddress, SIZE_1GB)` then `ALIGN_VALUE(base, GetPowerOfTwo64(size))`). Since the comment tells readers to re-read that code before changing anything here, pointing at the function that actually implements the model for the override case would make the pointer exact. Relatedly, OVMF does not refuse the override window when it exceeds the address width — it programs base+size as-is, and the guest simply cannot reach it — so "the firmware assigns no window at all" on line 111 is slightly off (the guest-visible outcome, a dead device, is the same, and the refusal is right either way).

**`docs/architecture/confidential.md`** (line 337): Optional parity nit: this section (and the invariant list around line 436) describes the BAR-sized `X-PciMmio64Mb` window but not the new guest-address-space budget from this PR. An operator debugging a create refusal with a large-SKU card will look here first, so a sentence noting that the sized window is refused (with the numbers) when it cannot fit next to the guest's RAM inside the 1 TiB a 40-bit guest can address would help.

foxpatch-aleph (approved):
This PR fixes genuine bugs (a debug panic/release wrap on `end - start + 1` in the BAR parse loop, and a silent skip of reversed ranges that under-sized the window) with correct checked arithmetic, and adds a well-modeled budget check that faithfully mirrors OVMF's dynamic 64-bit MMIO placement (window above top of RAM, aligned to its own size) with a conservative 4 GiB low allowance. I verified the check runs before both the opaque-cmdline and verity arms, that the budget input `spec.memory_mib` matches the actual QEMU RAM sizing, and re-derived the arithmetic for every test case and realistic production configuration — the 512 GiB window of the real 128 GiB BAR1 card lands exactly at the 1 TiB ceiling and passes, while larger windows or large-RAM combinations are correctly refused. Error variants carry the numbers rather than a formatted string, fail-closed is chosen consistently, and test coverage pins the three arithmetic cases, both refusal directions, the clamp/budget interplay, and the create-time refusal. Two non-blocking nits below.

**`rust/crates/supervisor-daemon/src/gpu_bar.rs`** (line 122): Non-blocking: `window_mb.max(1)` guards the division but makes `check_mmio64_budget(0, ram)` pass trivially (base = top-of-RAM, top = base + 0). Unreachable from the production caller, which floors the window at 1 GiB, but since this is a `pub` function whose input contract (`mmio64_window_mb` output, power of two, ≥ 1 GiB) is only stated in prose, a `debug_assert!(window_mb > 0)` (or one doc line on the function) would pin it against a future caller passing 0 and silently getting a pass.

**`rust/crates/supervisor-daemon/src/gpu_bar.rs`** (line 49): Non-blocking: the doc thoroughly justifies why a `cpu_model` doesn't move the 1 TiB figure, but doesn't address the `reduced-phys-bits=1` that the same SNP argv passes via `sev-snp-guest` (qemu.rs:523). Since the C-bit/reduction lives in CPUID Fn8000_001F rather than Fn8000_0008, and OVMF's `PlatformAddressWidthFromCpuid` reads the latter, the 40-bit figure should hold — but as this constant is the linchpin of the whole budget, a one-line note confirming the reduction was considered (and that `PlatformAddressWidthFromCpuid` doesn't subtract it) would let the next reader re-verify rather than wonder. The shipped 512 GiB-window configuration working in the field corroborates the figure either way.

foxpatch-aleph (approved):
This PR replaces unchecked arithmetic and an unbounded window with checked operations and a create-time address-space budget, and it is correct. I verified the two external claims it rests on against upstream sources: QEMU defaults named x86 models to 40 physical bits when neither phys-bits nor host-phys-bits is passed (the SNP argv in supervisor-controller/src/qemu.rs passes the model name alone), and SEV reduced-phys-bits is not subtracted from CPUID 0x80000008 (it is reported via Fn8000_001F EBX[11:6]), so the 1 TiB guest budget is right. OVMF's PlatformGetFirstNonAddress places the aperture at ALIGN_VALUE(top_of_ram, GetPowerOfTwo64(size)) and PlatformDynamicMmioWindow returns early on the fw_cfg override, so the daemon's align_up(top_of_ram, window) + window model matches exactly, and with the daemon always sending powers of two >= 1 GiB the two-step alignment collapses to the daemon's formula. The parse validation (checked_sub/checked_add, only on wanted-flag lines) is correct, the budget errs conservative in the right direction, the flagship 512 GiB window + 64 GiB RAM case fits exactly at the ceiling, and the typed error variants carry the numbers. The new tests pin all three arithmetic cases, both refusal directions, the clamp/budget interaction, and the create-time refusal through the right error path. Only non-blocking notes below.

**`rust/crates/supervisor-daemon/src/gpu_bar.rs`** (line 119): Non-blocking: the placement model documented here matches OVMF's PlatformGetFirstNonAddress, but that function also raises FirstNonAddress to the memory-hotplug end advertised via the `etc/reserved-memory-end` fw_cfg entry before placing the aperture. Today's SNP argv uses a fixed-size memfd backend with no hotplug region, so the model holds, but a one-line note that the check assumes no memory-hotplug area (and that adding hotplug to the argv would need this revisited) would keep the model's fidelity claims airtight, in keeping with the docstring's "re-read that code before changing anything here".

PR #1228, opened 2026-09-09
docs: fix architecture docs left stale by the Python supervisor removal

odesenfans:
## Summary
The docs/architecture set was last verified before the Python supervisor daemon, its per-VM controllers and the implementation-dispatch launcher scripts were deleted (2026-08). Several pages still cited files that no longer exist (`error_mapping.py`, `grpc_server.py`, the qemu controllers package, `src/aleph/vm/network/`), described the daemon as one of two implementations, or had function names and counts that later refactors moved (`roomiest_pool_free_bytes` vs `eligible_pool_free_bytes`, the `Supervisor` ABC's 23 methods not 29, the V-PROGRAM stop-guard now living in `allocation/teardown.py`). This PR verifies every flagged claim against the current worktree and fixes the docs.

## Changes
- `networking.md`: drop the deleted `src/aleph/vm/network/` pointer entry, describe the SNP dnsmasq server's DHCPv6 range and `--enable-ra` (not just a single IPv4 `--dhcp-range`), and note `firewall.py` is deleted.
- `storage.md`: fix `_check_max_volume` (uses `eligible_pool_free_bytes()`, not the unused `roomiest_pool_free_bytes()`), the daemon's looser `VOLUME_POOLS` validation (accepts `hdd`, skips path-exists checks the agent enforces), the devmapper writable file (fallocate only, no filesystem of its own), the cloud-init parity claim (byte-compared fixtures, not a live parsed-Python comparison, since that Python builder is gone), the download-cache in-stream check (a hard ceiling on written bytes, not a comparison against a possibly-lying `Content-Length`), and the four caches' `CACHE_BUDGET` (not deduplicated by `st_dev`, unlike the pool figure). Bumped the stale verified-against stamp.
- `wire-contract.md`, `divergences.md`: reframe the removed Python daemon/`error_mapping.py`/`grpc_server.py`/`STATUS_CODE_BY_ERROR`/`LocalSupervisor` as history instead of a live oracle; mark divergence entries 59/60 (the launcher-dispatch packaging notes) obsolete now that the launchers, the flag and their test are gone.
- `process-model.md`: drop "Rust or Python"/"either implementation" framing, fix who writes `{vm_hash}-controller.json` (the daemon, not the agent) and the `Supervisor` ABC's method count (23, not 29).
- `overview.md`: fix the repo-layout table (no more launcher scripts, no more fixture-generation scripts, no `tests/network/` suite) and add the `agent/allocation/` and `/v2/control/` allocation surface it omitted.
- `vm-lifecycle.md`: fix expiry/update-watcher reaping (both go through `retire_vm(..., RetireReason.RECREATE, ...)`, not a bare `supervisor.delete_vm`), and add the v2 allocation surface (`/v2/control/capacity/check`, `/v2/control/allocations`) that the admission section omitted.
- `confidential.md`: add the third capability probe (the daemon's BAR0 GPU confidential-computing-mode probe feeding `nvidia_cc`, previously described as a "future" key even though it's advertised today), the EIP-191 owner-signature gate on `/confidential/inject-secret` for confidential-instance images, correct the V-PROGRAM stop-guard's real location (`allocation/teardown.py`, shared with the v2 reconciler, not `views/__init__.py`), and note aleph-tee already carries TDX quote-verification code (software path implemented, no hardware backend yet), not just SEV-SNP.
- `testing.md`: fix the droplet CI OS/leg counts (five OSes built, four droplet-tested; ubuntu-26.04 was missing), add the `golden-measurements.yml` workflow to the inventory, fix the shellcheck job's file patterns (`*.sh`/`*.script`, the `*-launcher` pattern is gone), and drop the stale "firewall tests in tests/supervisor" claim (none remain there).
- `proto/README.md`, `README.md`: fix the proto codegen output path and a stale README link label.

## Test plan
Docs-only change per the dispatch; no tests added or run. Every corrected claim was checked directly against the code in this worktree (grep/read of the relevant Python and Rust modules, workflow YAML files, and test files) before editing; see the report for the specific commands.

Stayed out of: the `vm-lifecycle.md` paragraph another branch (`od/2.1-alloc-owner-stop-respected`) is adding right after the stop/idle-expiry boundary, and the `storage.md` CLI/validation sections three other branches (`od/2.1-storage-cli-safety`, `od/2.1-storage-strict-namespace`, `od/2.1-storage-live-snapshot-on-retire`) are rewriting; confirmed by diffing against those branches before editing.


foxpatch-aleph (commented):
This docs-only PR is exceptionally thorough and nearly every corrected claim checks out against the code: I independently verified the _check_max_volume/eligible_pool_free_bytes fix (capacity.py:790-798), the daemon's looser VOLUME_POOLS validation (parse_pool_entry accepts hdd and skips path checks that setup_pools enforces), the fallocate-only writable volume file, the download cache's independent in-stream ceiling, the per-root CACHE_BUDGET without st_dev dedup, the Supervisor ABC's 23 methods (AST-counted, the 29 was stale), the daemon writing {vm_hash}-controller.json (lifecycle.rs:2972), the retire_vm(RECREATE) reaping in expiry.py:58 and update_watcher.py:93, the teardown.py stop-guard, the gpu_cc.rs probe_cc_mode/nvidia_cc-only-alongside-sev_snp claim, the EIP-191 owner gate with 403 in the attest-agent, the TeeType TDX/NvidiaCc variants with no TdxBackend, the five-OS/four-droplet CI counts, the golden-measurements.yml triggers, the shellcheck *.sh/*.script patterns, and the removal of the launchers, ALEPH_VM_SUPERVISOR_IMPL and tests/test_controller_launcher.py. The one real inaccuracy introduced is the cloud-init fixture path: the docs point to tests/fixtures/cloudinit/, which does not exist at the repo root (there is no tests/fixtures/ directory there); the fixtures live at rust/crates/supervisor-daemon/tests/fixtures/cloudinit/, resolved via CARGO_MANIFEST_DIR in test_fixtures.rs:32-34, so both mentions should be corrected. Also worth a follow-up outside this PR's scope: abc.py's own docstring still says "implements all 29 methods", now disagreeing with the corrected count.

**`docs/architecture/storage.md`** (line 225): The fixtures path is wrong: there is no `tests/fixtures/` at the repo root. The committed fixtures live at `rust/crates/supervisor-daemon/tests/fixtures/cloudinit/` (resolved via `CARGO_MANIFEST_DIR` joined with `tests/fixtures` in `test_fixtures.rs:32-34`). Since every other path in these docs is repo-root-relative, this sends readers hunting for a nonexistent directory — the very class of staleness this PR fixes.

**`docs/architecture/storage.md`** (line 736): Same fixture-path issue as above: `tests/fixtures/cloudinit/` should be `rust/crates/supervisor-daemon/tests/fixtures/cloudinit/`.

**`docs/architecture/process-model.md`** (line 51): Minor nit: `save_controller_configuration` is not fully unused — it is still exercised by `tests/supervisor/test_save_controller_configuration.py` (its atomicity unit test). "Unused outside its own test" or "unused by production code" would be precise. Defensible as-is; just flagging.

**`src/aleph/vm/supervisor_interface/abc.py`** (line 4): Out of scope for this docs PR, but worth a follow-up: the module docstring here still says a concrete supervisor "implements all 29 methods", which the corrected doc (and an AST count) puts at 23. The doc fix is right; now the source docstring disagrees with both.

foxpatch-aleph (approved):
Docs-only PR; I independently verified essentially every corrected claim against the worktree and found no errors. The 23-method ABC count, `_check_max_volume`'s use of `max()` over `eligible_pool_free_bytes()` plus reclaimable bytes, the daemon's looser `VOLUME_POOLS` parsing (accepts `hdd`, no path-exists check vs the agent's `setup_pools()`), fallocate-only `create_volume_file`, the byte-compared cloudinit fixtures, the dual download cap, per-root CACHE_BUDGET without st_dev dedup, `retire_vm(..., RetireReason.RECREATE, ...)` in both expiry.py and update_watcher.py, the `is_removable_by_allocation` teardown guard shared by the legacy endpoint and the v2 reconciler, the BAR0 GPU-CC probe feeding per-card `cc_mode`, the EIP-191 owner-auth 403 gate, aleph-tee's TDX module (with no `TdxBackend` and QGS explicitly deferred), the five-built/four-droplet-tested OS matrix, the shellcheck `*.sh`/`*.script` patterns, golden-measurements.yml's triggers, the dnsmasq dual single-address ranges with `--enable-ra` and `--port=0`, the proto codegen path, and the removal of `src/aleph/vm/network/`, `src/aleph/vm/supervisor/`, `tests/network/` and `tests/test_controller_launcher.py` all match the code exactly. Only three cosmetic nits, noted below. The reframing of removed Python-daemon references as history (divergences 59/60, wire-contract) is accurate, including the controller unit now exec'ing the Rust binary directly.

**`docs/architecture/divergences.md`** (line 14): The rewritten preamble's Dispositions list documents keep / fix-in-python-later / fix-in-both-later / fix-in-rust, but entries 59 and 60 below use a new disposition value, `resolved`, which the list does not define. Since the paragraph above mentions entries are "marked obsolete/resolved", consider adding `resolved` to the glossary line, e.g. "...a known Rust gap with a planned increment), or **resolved** (the Python side it compared against is gone, kept for its stable entry number)".

**`docs/architecture/divergences.md`** (line 3): Pre-existing line, but the reframe leaves it reading "Living registry of deliberate Rust-vs-Python divergences" directly under a title that now says "(historical)". Consider softening to match, e.g. "Registry of deliberate Rust-vs-Python divergences; entry numbers are stable and referenced by tests."

**`docs/architecture/confidential.md`** (line 3): This file's verified-against stamp is `973b8fd7 (2026-09-10)` while every other file in this PR is stamped `06c30936 (2026-09-09)`. If confidential.md was deliberately verified against a different (stacked) commit that includes the owner-auth work, fine; otherwise this looks like it should match the rest.

PR #1229, opened 2026-09-09
fix(agent): publish a failure code, not the exception, on the public list

odesenfans:
## Summary
`/v2/about/executions/list` is unauthenticated and `@cors_allow_all`, and the allocation block it renders carried `str(error)[:200]` of whatever `start_persistent_vm` raised, under a code that fell back to the exception's class name. The exceptions a create raises quote what it was working with: the capacity refusal names the node's free memory and the cache path a download would not fit in, a download failure names the URL, and an unhandled one names whatever its raiser put in it. Reproduced on the base commit: a start failing with `RuntimeError("could not open /var/lib/aleph/vm/deadbeef/private-volume.img")` served `{"code": "RuntimeError", "message": "could not open /var/lib/aleph/vm/deadbeef/private-volume.img"}` to any anonymous reader. v1 has always masked an unclassified failure as "Unhandled Error".

A failure record now carries a code out of a closed set (`AllocationFailureCode`) and no free text at all, so there is no field left for an exception to write into. Each code has one sentence, written for a reader and published with it, and the exception's own text goes to the log. The code is derived from the exception's type, never from matching on its message.

Stacked on #1225.

## Changes
- `src/aleph/vm/agent/allocation/failures.py` (new): `AllocationFailureCode`, the one public sentence per code, and `classify_start_failure`. Supervisor boundary errors map through their `ErrorCode`; the agent's own `InsufficientResourcesError` and `VmStartupError` map by class; the message fetch's `HTTPNotFound` / `HTTPServiceUnavailable` map to `message_unavailable`; the unsupported-content-type `HTTPBadRequest` maps to `unsupported`; anything unrecognised is `internal` with "Unhandled error".
- `src/aleph/vm/agent/allocation/plan.py`: `FailureRecord.code` is the enum, and the `message` field is gone.
- `src/aleph/vm/agent/allocation/reconciler.py`: `_record_failure` classifies instead of reading `type(error).__name__` and `str(error)`, and logs the code alongside the full exception. The rebuild path records `vm_failed` rather than a code built from the status string.
- `src/aleph/vm/agent/views/__init__.py`: `_allocation_block` renders the code's value and its published sentence.
- `docs/architecture/vm-lifecycle.md`: a sentence on what the allocation loop is allowed to report about a failure, in the paragraph that already describes that reporting.

### One correction to the original plan
The plan said the message field could carry a typed error's own message for codes that are safe by construction. It cannot: the typed errors the create path raises are exactly the ones with unsafe text. `capacity.py` raises `InsufficientResourcesError("Insufficient capacity to create VM. ...")` with the host's committed and available figures, `vm/cache.py` raises it naming `/var/cache/...`, and `vm/downloader.py` raises `ResourceDownloadError(str(ClientResponseError))`, which is the URL. So the message published is a function of the code and of nothing else. The typed errors still earn their keep: they decide which code, so a full node, a message that cannot be fetched, a runtime that will not download and a VM this CRN cannot run stay four different answers to the scheduler.

## Test plan
- `tests/supervisor/test_allocation_failures.py` (new): every exception the start path can raise lands on the code that describes it, every `ErrorCode` is covered, every failure code has a distinct message, and `internal` says "Unhandled error".
- `tests/supervisor/test_executions_list_allocation.py`: end to end through the real reconciler and the real app, a start raising a `RuntimeError` that quotes a host path serves `{"code": "internal", "message": "Unhandled error"}` with the path absent from the whole JSON body, and a capacity refusal serves `insufficient_capacity` with its own sentence and none of the refusal's figures. The two tests that asserted the record was echoed verbatim now assert the sanitized shape.
- `tests/supervisor/test_allocation_reconciler.py`: the record itself carries the code and none of the exception's text, for an unclassified failure and for a typed capacity one. The assertion that a failed start recorded `code == "RuntimeError"` now expects `internal`.
- Checks: the seven allocation and views test files plus `tests/supervisor/views` (264 passed, exit 0); the whole `tests/supervisor` suite (1238 passed, 1 skipped, exit 1 on `test_log.py::test_make_logs_queue`, the pre-existing journald environment failure, identical on the base); `ruff format --diff .`, `isort --check-only --profile black .` and `mypy src/aleph/vm/ tests/ examples/example_fastapi runtimes/aleph-debian-12-python` all exit 0.


foxpatch-aleph (approved):
This PR correctly fixes an information-disclosure issue on the unauthenticated, CORS-open /v2/about/executions/list endpoint by replacing verbatim exception text with a code from a closed enum plus a fixed per-code sentence, leaving exception detail to the operator log. The design is fail-safe: classification is by exception type (never message matching), unrecognized exceptions fall through to a fixed "Unhandled error" ("internal"), and I traced every exception the start path can raise (capacity admission, supervisor boundary errors, VmStartupError, message-fetch HTTP errors, unsupported-content-type HTTPBadRequest, download errors) to confirm each lands on a sensible code. All FailureRecord constructors and readers were updated consistently, there is no circular import risk, and test coverage is strong — per-exception classification, full ErrorCode enumeration, and end-to-end tests asserting the secret path is absent from the whole JSON body. I could not execute the tests locally (environment lacks project dependencies), but static analysis found no issues. Two non-blocking nits: an unused `logger` in the new failures.py, and a non-404 ClientResponseError from the message fetch is classified as DOWNLOAD_FAILED rather than MESSAGE_UNAVAILABLE. The changed code vocabulary is a breaking change for consumers of the old raw codes, which is intentional.

**`src/aleph/vm/agent/allocation/failures.py`** (line 29): `logger = logging.getLogger(__name__)` is unused in this module — either drop it or note why it's here.

**`src/aleph/vm/agent/allocation/failures.py`** (line 97): Minor imprecision: a non-404 `ClientResponseError` escaping the message fetch (messages.py re-raises it rather than converting to `HTTPNotFound`/`HTTPServiceUnavailable`) lands on DOWNLOAD_FAILED instead of MESSAGE_UNAVAILABLE. Harmless since no text is published either way, but if the distinction ever matters to the scheduler, the fetch path could wrap its ClientResponseError like the downloader does (downloader.py:165-174).

**`src/aleph/vm/agent/views/__init__.py`** (line 308): Note for rollout: the published `code` vocabulary changes (e.g. `"RuntimeError"` → `"internal"`, `"resource_download_error"` → `"download_failed"`), which is a breaking change for any scheduler consumer keying on the old raw values. Intentional per the PR, just flagging it for the consumer-side rollout.

foxpatch-aleph (approved):
A clean, well-tested security fix. The leak is closed structurally rather than by sanitization: FailureRecord no longer has a free-text field, so no exception text can reach the unauthenticated, cross-origin executions list, and the published error is a code out of a closed enum paired with a fixed sentence keyed by code. Classification is by exception type rather than message matching, which is the robust choice, and I verified against the real start path that every exception create_vm_execution can raise lands on a code, with unknowns falling to internal/'Unhandled error'. Wire compatibility is preserved (the old vm_{status.value} scheme produced the same 'vm_failed' string the new enum publishes), no stale consumers of the removed message field remain, the docs accurately describe the new contract, and the tests are genuinely good — including end-to-end tests asserting the secret host path and the capacity figures are absent from the entire JSON body, and a guard that every supervisor ErrorCode classifies. All 86 tests in the three touched files pass, and ruff/isort/mypy are clean.

**`src/aleph/vm/agent/allocation/failures.py`** (line 95): Nit, non-blocking: any ClientResponseError maps to download_failed, but a non-404 one can also escape the message fetch (messages.py:33 and :45 re-raise raw), so a connector 500 on the message itself would publish download_failed rather than message_unavailable. Still a closed code with fixed text, so nothing leaks — purely a granularity nit.

**`src/aleph/vm/agent/allocation/failures.py`** (line 101): public_failure_message does a bare _PUBLIC_MESSAGES[code] lookup, so an enum member added without a message would raise KeyError inside the request handler and 500 the endpoint. test_every_code_has_a_message_of_its_own guards this today, which is probably enough, just worth knowing the failure mode is a request-time 500 rather than a missing sentence.

**`tests/supervisor/test_allocation_failures.py`** (line 71): This assertion is trivially true for any return value (and mypy already guarantees the return type), so the test's real value is that no ErrorCode makes classify_start_failure raise — the name slightly oversells it. The specific SUPERVISOR_ERROR cases are properly covered by the parametrized test via PortUnavailableError and InternalSupervisorError, so this is fine as is.

foxpatch-aleph (approved):
A genuine information-leak fix, done at the right layer: FailureRecord lost its free-text field entirely, so the unauthenticated executions list can only publish a code from a closed enum plus its fixed sentence, and the type system now enforces that policy for future call sites. I traced every exception class in classify_start_failure against the actual reconciler start path (start_persistent_vm -> create_vm_execution) and the mapping is correct for each one, including the subtle cases (the HTTP-wrapping functions that raise HTTPServiceUnavailable with a capacity meaning are not on this path; in the reconciler path it comes only from the message fetch). All 182 allocation tests pass, and the new tests are strong: per-exception mapping, full ErrorCode enumeration, and end-to-end assertions that secret text is absent from the whole JSON body. Only non-blocking nits: the every-code-covered test asserts little (membership is already guaranteed by the return type, so a code dropped from the supervisor table would not fail it), and the public code values changed shape, which any scheduler consumer must track.

foxpatch-aleph (approved):
This PR correctly closes an information-disclosure hole on the unauthenticated executions list: the failure record now carries only a code from a closed enum plus a fixed per-code sentence, with no field left for exception text to leak through. The classification is by exception type (a contract), never message matching, and I verified the ordering hazards: the boundary InsufficientResourcesError is caught via SupervisorError before the agent-side same-named class, and in aiohttp 3.10 HTTPNotFound/HTTPServiceUnavailable are not subclasses of HTTPBadRequest, so the HTTP-exception checks cannot shadow each other. The reconciler keeps the full exception in the node log for the operator, and the end-to-end tests assert leaked content is absent from the whole JSON body. All 360 relevant allocation/verdict/executions tests pass and mypy is clean on the touched files. Two non-blocking nits on test strength and the breadth of the HTTPBadRequest mapping; neither affects correctness or the security guarantee.

**`tests/supervisor/test_allocation_failures.py`** (line 71): This test cannot fail as written: classify_start_failure always returns an AllocationFailureCode (and dict.get with a default cannot raise), so `assert ... in AllocationFailureCode` holds for any input. To actually enforce the docstring's intent — a code added to ErrorCode must not fall off the map — assert the destination for the unmapped codes explicitly, e.g. that PORT_UNAVAILABLE, VM_NOT_FOUND, VM_ALREADY_EXISTS, HOST_NOT_FOUND and INTERNAL each land on SUPERVISOR_ERROR. As is, the test only exercises lines.

**`src/aleph/vm/agent/allocation/failures.py`** (line 90): Any HTTPBadRequest on this path is published as UNSUPPORTED ("This node cannot run this VM"), which reads as permanent. Today that is accurate — the only HTTPBadRequest the reconciler path can reach is the unsupported-content-type fallback in create_vm_execution (run.py:717) — but the "no later attempt will do better" property belongs to the raiser, not the exception type. Worth keeping in mind if a transient condition ever starts raising HTTPBadRequest from inside the start path; create_vm_execution_or_raise_http_error already translates ResourceDownloadError into HTTPBadRequest, and would misclassify if ever routed through the reconciler. Non-blocking given the closed-set design limits the damage to a slightly wrong sentence.

foxpatch-aleph (approved):
This PR correctly closes a genuine information-disclosure hole: I verified that /v2/about/executions/list is unauthenticated and CORS-allow-all, and that the old code published str(error) verbatim. The replacement design — a closed enum of codes with one fixed sentence each, classification purely by exception type — is sound and correctly grounded: I checked the aiohttp class hierarchy to confirm no branch ordering can shadow, traced each classified exception to a real raiser on the start path (try_get_message for HTTPNotFound/HTTPServiceUnavailable, run.py:717 for HTTPBadRequest, _wait_until_running for VmStartupError, capacity admission for InsufficientResourcesError), and confirmed via grep that nothing still reads the removed FailureRecord.message field. The full exception detail is retained in the log for operators. The new and updated tests pass (20 + 63); the only failures I encountered are pre-existing sandbox permission issues in test_operator_retire.py, untouched by this PR. Minor nits only, none blocking.

**`src/aleph/vm/agent/allocation/failures.py`** (line 80): Nit: `ErrorCode.INTERNAL` (raised as `InternalSupervisorError`, e.g. a NotimplementedSupervisorError) falls through to `supervisor_error`, published as "The hypervisor refused to run this VM". An internal error is a bug on the node rather than a refusal — the unclassified-Python-error path gets the more honest `internal`/"Unhandled error" here. Consider mapping `ErrorCode.INTERNAL` to `AllocationFailureCode.INTERNAL` explicitly.

**`src/aleph/vm/agent/allocation/failures.py`** (line 94): Nit: a `ClientResponseError` with a non-404 status raised while fetching the *message* (messages.py re-raises it from `try_get_message`) lands on `download_failed` rather than `message_unavailable`. Still a closed-set sentence, so no leak — just a slightly-off code for a scheduler trying to tell a bad hash from a bad download.

**`tests/supervisor/test_allocation_failures.py`** (line 71): This assertion is tautological: `classify_start_failure` has return type `AllocationFailureCode`, so `in AllocationFailureCode` holds for any input. It would still pass if a future `ErrorCode` that deserves a specific mapping (say, a new `VM_NOT_FOUND`-style code) were added. The per-subclass parametrized test above covers the intended mappings, so this is minor, but asserting against an expected mapping per code would make this test catch drift.

**`tests/supervisor/test_executions_list_allocation.py`** (line 141): Nit (test-only): with `real_reconciler=True`, `setup_webapp`'s startup hook also spawns the reconciler's background `run()` loop, which can converge on the submitted plan concurrently with the explicit `await reconciler._converge_once()`. The tests don't assert on `attempts`, so it passes, but the count of recorded attempts is nondeterministic in principle — worth knowing if a later assertion is added.

foxpatch-aleph (approved):
This PR correctly closes an information leak on the unauthenticated /v2/about/executions/list endpoint: the failure record now carries only a code from a closed enum plus a fixed sentence, with no free-text field for an exception to write into. I verified the classifier covers every exception type the start path can actually raise (the raw SupervisorError vocabulary from create_vm_execution, HTTPNotFound/HTTPServiceUnavailable from load_updated_message, VmStartupError from _wait_until_running, the agent's own InsufficientResourcesError, and the HTTPBadRequest for unsupported content types), with correct isinstance ordering between the two distinct InsufficientResourcesError classes. The reconciler keeps the full exception text in the operator-only log, the docstring and architecture docs explain the invariant, and no circular import or stale FailureRecord consumer remains. All new and modified tests pass, as do ruff format and mypy; the 8 pre-existing failures in tests/supervisor/views/test_operator_retire.py are environment issues (PermissionError on /var/cache/aleph) unrelated to this change. Only nit: test_every_supervisor_error_code_is_covered cannot fail as written and so does not test the invariant its name claims.

**`tests/supervisor/test_allocation_failures.py`** (line 71): This assertion is vacuous: classify_start_failure is annotated to (and always does) return an AllocationFailureCode member, so `result in AllocationFailureCode` passes for every input and cannot detect a code falling off the map. A stronger check that matches the docstring's intent would be `assert classify_start_failure(error) is not AllocationFailureCode.INTERNAL` — a boundary error is a real answer and should never be classified as unhandled. Non-blocking.

foxpatch-aleph (approved):
This PR closes a real information-disclosure hole (exception text quoting host paths, URLs and node capacity figures was rendered on an unauthenticated, cross-origin endpoint) and does so with a design that makes recurrence impossible rather than merely patched: FailureRecord has no free-text field left, the public message is a pure function of a closed enum, and classification is by exception type rather than message matching, which is the stable contract. All call sites are updated consistently, the docstring and architecture-doc updates accurately describe the new behavior, and the test coverage is strong (per-exception classification, every ErrorCode exercised, end-to-end assertions that the secret text is absent from the whole JSON body, and the pre-existing verbatim-echo tests converted to assert the sanitized shape). I verified the classification ordering handles the two InsufficientResourcesError classes correctly, ran the relevant suites (83 passing in the three touched files; only pre-existing sandbox-environment failures elsewhere), and confirmed ruff/isort/mypy are clean on the changed files. Two nits, neither blocking.

**`src/aleph/vm/agent/allocation/failures.py`** (line 94): Non-blocking observation: a ClientResponseError escaping the message fetch with a non-404 status (try_get_message and get_latest_ref both re-raise these, e.g. a 500 or 429 from the connector) lands on DOWNLOAD_FAILED rather than MESSAGE_UNAVAILABLE. The published sentence for download_failed ("A resource this VM needs could not be downloaded") is still a fair description, so this is a coarse-bucket judgment call rather than a bug — and nothing leaks either way.

**`tests/supervisor/test_allocation_failures.py`** (line 77): Non-blocking: `assert classify_start_failure(error) in AllocationFailureCode` is guaranteed by the return type annotation, so this test's real value is exercising every ErrorCode without a KeyError from _BY_SUPERVISOR_CODE's .get default. It does that job; a marginal improvement would be asserting the folded codes (INSUFFICIENT_RESOURCES -> insufficient_capacity, etc.) per ErrorCode, which would also guard against a future edit pointing one at the wrong bucket.

foxpatch-aleph (approved):
This PR fixes a real information-disclosure vulnerability: the unauthenticated, CORS-open executions list published up to 200 characters of raw exception text (host paths, download URLs, node capacity figures) from any failed VM start. The fix is structurally sound rather than cosmetic — FailureRecord loses its free-text field entirely, so exception text cannot reach the public response by construction, while the full detail is preserved in the operator log. I verified the classifier cannot raise (SupervisorError.code always exists, dict lookup has a fallback, terminal return), that all _BY_SUPERVISOR_CODE keys match real ErrorCode members, that the HTTP exception mappings match what the reconciler path actually raises, and that no stale .message consumers remain. All 83 tests in the three touched files pass, including the end-to-end assertions that the secret text is absent from the whole JSON body; the 8 failures in tests/supervisor/views/test_operator_retire.py are a sandbox permission artifact unrelated to this diff. Only minor nits: the ErrorCode-coverage test is tautological (the fallback makes it unable to fail), and FileTooLargeError mapping to DOWNLOAD_FAILED slightly stretches the published sentence.

**`tests/supervisor/test_allocation_failures.py`** (line 66): This exhaustiveness test is tautological: because classify_start_failure has a SUPERVISOR_ERROR fallback, the assertion that the result is an AllocationFailureCode member can never fail. It documents intent, but the real coverage is the parametrized test above; consider asserting instead that codes absent from _BY_SUPERVISOR_CODE land on SUPERVISOR_ERROR.

**`src/aleph/vm/agent/allocation/failures.py`** (line 64): Minor semantic nit: FILE_TOO_LARGE maps to DOWNLOAD_FAILED ("A resource this VM needs could not be downloaded") when the resource actually downloaded but exceeded a limit. The don't-retry signal is right; only the sentence is a slight stretch.

**`src/aleph/vm/agent/allocation/failures.py`** (line 90): HTTPBadRequest -> UNSUPPORTED is accurate today (the only raise in the reconciler path is the unsupported-content-type at run.py:717), but it is a wide exception class to hang that meaning on; a future unrelated HTTPBadRequest in the start path would be misclassified as "This node cannot run this VM". Not blocking.

PR #1230, opened 2026-09-09
docs: operator documentation for what 2.1 adds

odesenfans:
## Summary

> **Merge order.** This PR documents the `vms[<i>]` rejection keys and the
> closed set of `error.code` values that #1229 introduces. It merges after
> #1229; landing it first would document answers the node does not yet give.
Nothing an operator would read documented what 2.1 adds. The storage
reclamation CLI existed only in the architecture notes, the `VOLUME_*`,
`CACHE_BUDGET` and `ALLOCATION_*` settings only in `conf.py`, and the v2
allocation endpoints nowhere at all. An operator who upgraded a node and
saw free space jump, or a scheduler integrator reading the v2 answer, had
no document to go to. This PR adds two operator documents, a 2.1 section
to the confidential-computing document, and fixes the agent README's
install instructions, which still told operators to install
`python3-nftables python3-jsonschema` and to `pip3 install pydantic-dotenv
'aleph-message~=1.0.1'`.

Stacked on #1228.

## Changes
- `doc/storage-reclamation.md` (new): what happens to the disks of a VM
  that is gone (including the payment-shortfall change from 1.x), the
  retention and cache budgets with their defaults, the `aleph-vm storage`
  verbs, the read-only guarantee on `status`/`list`, the environment-file
  loading a hand-run command does, the three-state rule that decides
  whether a purge may run, the `--trust-registry` caveat, and the exit
  codes (0 / 1 / 2 / 3) a wrapper script needs.
- `doc/scheduler-allocations.md` (new): `POST /v2/control/allocations` and
  its verdict buckets, `POST /v2/control/capacity/check`, the `allocation`
  block in `GET /v2/about/executions/list`, the rule that a node is
  governed either by v1 pushes or by v2 plans and not both, the four rules
  a reader of the answer needs (a refused hash is never torn down by its
  own push, a crash loop backs off, a stopped VM restarts only on a push,
  one create at a time per VM), and the `ALLOCATION_*` settings. The eight
  published `error.code` values from #1229's `AllocationFailureCode` are
  tabled with their published sentence and what each tells a scheduler to
  do, so an external consumer need not read source.
- `doc/confidential.md`: a "What 2.1 changes for a confidential CRN"
  section: TDX instance messages rejected up front, what a node must
  satisfy before it advertises `tee.nvidia_cc` (CC mode `on`, card free
  and bound to vfio-pci, GPU support and `pciutils`, SNP launch
  capability), and the note that the measured guest image changed so
  V-PROGRAM runtime bundles must be rebuilt and republished.
- `src/aleph/vm/agent/README.md`: apt list matched to the deb (plain
  `nftables`, no `python3-jsonschema`), and section 2.e replaced with
  `pip3 install .` against the current `pyproject.toml` pins.
- `README.md`: links the operator documents.

## Test plan
- Docs only, no tests, as the dispatch states.
- Content read off the final rules on the in-flight branches:
  `origin/od/2.1-storage-cli-readonly-and-diagnostics` (storage CLI and the
  storage architecture document), `origin/od/2.1-alloc-inflight-removing`
  (the seven allocation fixes and the vm-lifecycle document),
  `origin/od/2.1-packaging-prune-deps` (deb and pyproject), and
  `src/aleph/vm/conf.py` on this branch for every default.
- The apt-line edit is byte-identical to the one on
  `od/2.1-packaging-prune-deps`, so the two merge without a conflict.


foxpatch-aleph (commented):
Docs-only PR whose claims I verified against the codebase with unusual success: every settings default, endpoint shape, exit code, rejection/retention code, and confidential-computing claim that can be checked against this branch matches the implementation exactly (conf.py defaults, allocations_v2.py/allocation_auth.py/verdict.py, storage_cli.py verbs and exit codes, translate.py's TDX refusal, resources.py nvidia_cc gating, pyproject deps for the README pip fix). The one systematic caveat is that the PR documents behavior that lives on the in-flight stacked branches rather than this branch: the storage CLI's three-state agent-running rule, --env-file loading, 'unknown, never 0 B', and read-only guarantees are not in storage_cli.py here; the scheduler doc's crash-loop backoff and stopped-VM single-start rules contradict the current AllocationReconciler (whose _forget clears the failure record on a successful create and whose notify_vm_down wakes the loop for any planned STOPPED/FAILED event); and the pciutils deb-dependency claim does not hold for this branch's packaging control. That is the PR's stated stacked-PR design, so it is not blocking, but the merge order is a hard coupling worth locking in.

**`doc/storage-reclamation.md`** (line 133): This three-state rule (and the `--env-file` loading at line 115, the "unknown, never `0 B`" guarantee at line 85, and the read-only guarantees in the section above) describes the storage CLI as it exists on `od/2.1-storage-cli-readonly-and-diagnostics`, not on this branch: `src/aleph/vm/agent/storage_cli.py` here has no agent-running probe, no `--env-file` flag, `_status` prints `0 B` via `_human(0)` on an OSError, and `main()` unconditionally creates the registry/pools. Not a problem for a stacked docs PR, but if that branch slips past the 2.1 release this section makes false claims to operators — worth an explicit merge-order dependency in the release notes.

**`doc/scheduler-allocations.md`** (line 57): Verified the backoff math (base × 2^(attempts−1), capped) and the per-hash `creation_lock`, but on this branch's reconciler these two rules are only half-true: `_forget` (src/aleph/vm/agent/allocation/reconciler.py:237) drops the failure record as soon as the create succeeds, so a guest that dies after reaching RUNNING restarts on the next wake-up with no backoff and no "stayed up longer than the longest wait" rule; and `notify_vm_down` wakes the loop for any planned STOPPED/FAILED supervisor event, so a guest-shutdown VM is restarted on the event, not only on a push. Presumably `od/2.1-alloc-inflight-removing` changes this — a scheduler reading the doc against a node running only the merged base would observe different behavior, so same merge-order note.

**`doc/confidential.md`** (line 91): "It is a dependency of the Debian package from 2.1 on" — `packaging/aleph-vm/DEBIAN/control` on this branch has no `pciutils` (and still depends on `python3-jsonschema`, which the agent README's new apt line drops). Both presumably arrive with `od/2.1-packaging-prune-deps`; same merge-order coupling as above.

**`doc/storage-reclamation.md`** (line 175): Minor nit: the exit-1 row says "the environment file is missing", but the env-file section (line 117) says an absent default `/etc/aleph-vm/supervisor.env` is fine and logs "none". Only an explicitly named `--env-file` that does not exist is the error — consider "an explicitly named `--env-file` is missing" to keep the two sections from reading as contradictory.

foxpatch-aleph (changes requested):
The documentation itself is excellent: every claim I could check against this tree is accurate (settings names/defaults, endpoint shapes, verdict buckets, retained reasons, exit code 3, 413/8MiB rule, TDX message, nvidia_cc gating, the apt/pip fixes, all README links). The blocking problem is that a large part of both new documents describes behavior that does not exist in the tree this PR is based on — it documents #1229's AllocationFailureCode table and rules, plus the storage-CLI env-file/read-only/agent-liveness work and the packaging branch's pciutils change. Most seriously, scheduler-allocations.md:117 promises 'nothing an exception wrote is published here', but the code on this branch publishes str(error)[:200] on an unauthenticated, CORS-open endpoint and its error codes are exception class names (e.g. resource_download_error, per the existing test), not the eight documented codes. Similarly, none of the four reading rules hold in this tree (narrow_plan strips refused hashes and _teardown_dropped removes them; a plan-listed stopped VM restarts on every pass; no per-hash create lock exists; failure records are dropped on successful start), and storage_cli.py has no --env-file handling, no agent-running refusal, and status prints 0 B rather than unknown. Please re-base this PR on the last of the code PRs it documents (or annotate the hard ordering dependency so it cannot merge ahead of them); everything else is ready.

**`doc/scheduler-allocations.md`** (line 118): This is a security promise the code on this branch does not keep: `src/aleph/vm/agent/allocation/reconciler.py:259` publishes `str(error)[:200]` as `error.message`, and the published `code` is the exception class name (`resource_download_error` — see tests/supervisor/test_executions_list_allocation.py:124), not the closed eight-code set in the table below. Neither `AllocationFailureCode` nor codes like `message_unavailable`, `startup_failed` or `vm_failed` exist anywhere in this tree. This document must not merge before #1229; if the stack ordering is the intent, please re-target the base branch so that ordering is enforced rather than only stated in the PR body.

**`doc/scheduler-allocations.md`** (line 52): None of the four rules in this section hold on this branch: (1) `narrow_plan` (verdict.py:231) strips refused hashes from the plan and `_teardown_dropped` tears a running, removable one down; (2) a STOPPED VM is not in LIVE_STATUSES, so `_start_missing` restarts it on every pass, not only on a push; (3) there is no per-hash lock anywhere in the reconciler; (4) `_forget` drops the failure record as soon as a start succeeds, not after a max-backoff uptime window. Same fix as the error table: land this after the branch that introduces these rules, or reword to what the merged code will actually do.

**`doc/storage-reclamation.md`** (line 116): The `--env-file`/`ALEPH_VM_ENV_FILE` precedence, the exit-1 on a missing explicit file, and the exit-1 for `status`/`list` on a missing agent database (line 179) are all absent from `src/aleph/vm/agent/storage_cli.py` on this branch — `main()` calls `settings.setup()` and never reads an env file. This section documents `od/2.1-storage-cli-readonly-and-diagnostics`, which this PR is not based on; it cannot merge before that branch.

**`doc/storage-reclamation.md`** (line 87): Neither the read-only guarantees nor this `unknown` behavior exist in this tree's CLI: `_status` prints `0` for total/free/budget on an OSError, and `main()` unconditionally creates the agent database via `_load_registry()`. As with the env-file section, this is the in-flight branch's behavior — please make the stacking explicit or rebase onto that branch.

**`doc/confidential.md`** (line 92): "It is a dependency of the Debian package from 2.1 on" — `packaging/aleph-vm/DEBIAN/control` on this branch has no `pciutils`. True only once `od/2.1-packaging-prune-deps` lands; worth the same stacking note (the deb here still lists `python3-jsonschema` too, so the agent README's apt line is now matched to that branch's deb, not this one).

foxpatch-aleph (approved):
Docs-only PR that is unusually rigorous: every claim I could cross-check against this branch is accurate, including all settings defaults in both tables (verified against conf.py), the v2 allocation answer buckets and allocation-block shape (allocations_v2.py, views/__init__.py), the eight-MiB plan cap with its 413-not-401 behavior (allocation_auth.py), the five retained reasons and the V-PROGRAM exception (teardown.py, verdict.py), the backoff doubling, the storage CLI verbs, tables and exit code 3, and the TDX rejection message. The one structural caveat is that a substantial part of the documented surface — vms[<i>] rejected keys, the eight AllocationFailureCode values, --env-file, read-only status/list, the agent-running refusal, the stdout/stderr split and pciutils in the deb — exists only on the in-flight stacked branches, not on this branch, so merge order matters; the PR description discloses this and the content is internally consistent with it. Minor doc gap: the capacity-check response envelope is not shown.

**`doc/scheduler-allocations.md`** (line 46): This paragraph (and the eight error.code rows, the --env-file chain, the read-only status/list guarantee and the agent-running refusal in storage-reclamation.md) describes behavior that is not on this branch: build_plan here keys a rejected entry by str(raw_hash) (verdict.py:102) and storage_cli.py has no --env-file, no agent-running probe, and writes refusals to stdout. That is fine for a stacked docs PR — the description says the content was read off the in-flight branches — but it makes the merge order a hard dependency: this document must not merge before od/2.1-alloc-inflight-removing and od/2.1-storage-cli-readonly-and-diagnostics, or it will document behavior that does not exist. If any of those branches moves before this merges, re-check the load-bearing details (the vms[<i>] key format, the eight published sentences, the env-file precedence) since a wrapper script written against them breaks silently otherwise.

**`doc/scheduler-allocations.md`** (line ): The response envelope is missing: the endpoint answers {"results": {...}, "capacity": {...}} (allocations_v2.py:131), with the per-hash verdicts nested under results and the node's headroom beside them. An integrator reading only this section would expect the per-hash map at the top level. One sentence naming the two fields would close the gap.

**`doc/storage-reclamation.md`** (line 125): Same stack caveat as the scheduler doc: on this branch _reclaim writes its refusal to out (stdout, storage_cli.py:273) and main() unconditionally runs settings.setup() + storage_pools.setup_pools() for every verb, so neither the stdout/stderr split nor the read-only status/list guarantee holds until the od/2.1-storage-cli-readonly-and-diagnostics branch lands. Both are load-bearing for the wrapper scripts this section targets, so worth a re-check at merge time.

foxpatch-aleph (approved):
Well-written operator documentation; every claim I could check against this branch's code is accurate: all settings defaults match conf.py, the storage CLI verbs and exit-code semantics match storage_cli.py, the reconciler scope/pass-triggers/startup-preview match reconciler.py, the v2 endpoint shapes, auth precedence, 413-on-oversize, headroom fields, allocation-block shape and VmStatus values match allocations_v2.py/capacity.py/views, and the confidential-computing section matches the TDX rejection and nvidia_cc advertisement code. The caveat is one the PR description already discloses: a material part of the two new documents describes behavior that lives on the sibling in-flight branches (od/2.1-storage-cli-readonly-and-diagnostics, od/2.1-alloc-inflight-removing, od/2.1-packaging-prune-deps) rather than on this branch, and in two places the checked-out code does the opposite of what the doc states, so this PR must land after (or together with) those. With the stack order respected, this is good work; the comments below are merge-order and completeness notes, not defects in the writing.

**`doc/scheduler-allocations.md`** (line 59): Merge-order dependency: on this branch the opposite holds. `narrow_plan` (verdict.py:231) strips refused hashes from the plan, and `_teardown_dropped` (reconciler.py:165) tears down anything not in `plan.entries`, so a running VM whose pushed message fails verification IS torn down by its own push here. The rule as written is only true once `od/2.1-alloc-inflight-removing` lands — worth a note in the PR body that this file must not merge before that branch.

**`doc/scheduler-allocations.md`** (line 130): Same merge-order caveat, and it is security-relevant: on this branch `error.message` is `str(error)[:200]` (reconciler.py:262, `_record_failure`), i.e. exactly the raw exception text, and there is no `vms[<i>]` keying (an unreadable hash is echoed back verbatim, verdict.py:102). The claim that "nothing an exception wrote is published here" and the `vms[<i>]` keying only hold after #1229's `AllocationFailureCode` lands, so the 8-row table should be re-checked against that branch at merge time.

**`doc/storage-reclamation.md`** (line 101): Not what the checked-out code does: `storage_cli.main()` calls `storage_pools.setup_pools()` and `_load_registry()` unconditionally (storage_cli.py:337-342), so on this branch `status`/`list` do adopt pools and do create the database, and there is no `--env-file` handling at all. Accurate for `od/2.1-storage-cli-readonly-and-diagnostics` (and the refusal/`--env-file` exit-code-1 rows at line 179), but not for the branch as it stands — another entry for the merge-order note.

**`doc/confidential.md`** (line 93): Minor: `packaging/aleph-vm/DEBIAN/control` on this branch lists `lshw` and has no `pciutils` (and still depends on `python3-jsonschema`), so "dependency of the Debian package from 2.1 on" is true only after `od/2.1-packaging-prune-deps` merges. The README apt-line edit itself matches this branch's control file (`nftables`, no `python3-nftables`) and the two merge cleanly, as stated.

**`doc/scheduler-allocations.md`** (line 51): Nit: `_retention_reason` (verdict.py:136) also has an `operator_policy` fallback, currently unreachable because the branches mirror `is_removable_by_allocation`. The five-reason list reads as closed, so if that seam ever fires the doc silently goes wrong — a half-sentence "or a future operator-policy reason" would make it future-proof.

foxpatch-aleph (changes requested):
The documentation itself is high quality and the parts written against this tree verify cleanly: every settings default matches conf.py, the v2 endpoint shapes, verdict buckets, allocation block, retained reasons, TDX and nvidia_cc claims, reconciler triggers, storage CLI verbs, and the agent README fix (neither nftables bindings nor jsonschema is imported by the agent; pip3 install . matches pyproject) are all accurate. However, a substantial part of the documented behavior exists only on the in-flight branches listed in the description, not in this tree, and several rules are the opposite of what the code here does: the closed eight-code error table (this branch publishes str(error)[:200] and an exception-derived code on the unauthenticated executions list, so the security claim that nothing an exception wrote is published is currently false), the vms[<i>] rejected keying (this branch echoes str(raw_hash) back), the refused-hash teardown protection (narrow_plan plus _teardown_dropped means a refused-but-running VM is torn down by its own push), the stopped-VM-stays-stopped rule (start_persistent_vm resumes a STOPPED VM), the per-hash create lock (creating() is explicitly not a lock), and the storage CLI's --env-file loading, agent-liveness refusal, lost-database refusal, and read-only status/list guarantees (main() unconditionally runs setup_pools() and loads the registry for every verb; reclaim refusals go to stdout, not stderr). Merged before the code branches, these docs would actively mislead operators and scheduler integrators. Please land this strictly after the code PRs it documents, or note the dependency in the documents/PR so main never carries docs contradicting the code.

**`doc/scheduler-allocations.md`** (line 143): This security claim ("error.code comes from a closed set, and error.message is the one sentence that belongs to that code ... so nothing an exception wrote is published here") is false for the code in this tree: allocation/reconciler.py:257-261 derives code via getattr(error, "code", None).value or the exception class name, and publishes message=str(error)[:200] — raw exception text — on this unauthenticated, CORS-open endpoint. The closed set and fixed sentences exist only on the in-flight #1229 branch. Either gate this table on that branch landing first or note the dependency; as written it understates what the endpoint publishes today.

**`doc/scheduler-allocations.md`** (line 48): The vms[<i>] keying is not what this branch does: allocation/verdict.py:102 keys an unreadable-hash rejection by str(raw_hash), echoing back whatever string the push sent (the very concern the second paragraph raises). The index keying is one of the in-flight fixes; the doc should not land before it.

**`doc/scheduler-allocations.md`** (line 59): "A refused hash is never torn down by its own push" does not hold on this branch: narrow_plan (verdict.py:242) removes refused entries from the plan, and _teardown_dropped (allocation/reconciler.py:165) tears down any running VM not in plan.entries — so a VM the node already runs whose pushed entry was refused (invalid_message, node_mismatch, or a capacity refusal) is treated as dropped and torn down by that same push. A scheduler relying on this rule as documented would lose VMs.

**`doc/scheduler-allocations.md`** (line 72): The opposite is true on this branch: start_persistent_vm (agent/run.py:1116-1122) explicitly resumes a STOPPED VM ("is stopped, starting it" -> supervisor.start_vm) when a plan names it, and compute_verdict re-sizes a stopped, held VM as a candidate and can return accepted for it. This rule comes from the in-flight branch; the doc must not merge before it does.

**`doc/scheduler-allocations.md`** (line 91): On this branch only the program path takes a per-hash lock (program_client.creation_lock, run.py:824); the storage reconciler's own comment (vm/reconciler.py:86-89) states that a scheduler push and an operator reinstall take no common per-hash lock and that two creating() spans for one hash can overlap. The blanket "every start path" claim documents the in-flight fix, not this tree.

**`doc/storage-reclamation.md`** (line 101): "They do not create the agent database, do not create any configured directory, and do not adopt a storage pool" is not true of this branch's CLI: main() (agent/storage_cli.py:339-341) unconditionally runs storage_pools.setup_pools() — which writes pool markers and the adoption registry — and loads the registry for every verb including status and list. This guarantee is the whole point of the read-only claim, so the doc should land with the branch that implements it.

**`doc/storage-reclamation.md`** (line 116): No --env-file option or ALEPH_VM_ENV_FILE handling exists in agent/storage_cli.py: parse_args defines only --reclaimable, --trust-registry and --dry-run, and main() calls settings.setup() with no environment file. Same for the agent-liveness refusal (line 137, exit 3), the lost-database refusal (line 169), and the exit-1 "no agent database" case (line 179) — none are implemented on this branch. Also note line 126's "every warning, refusal and diagnostic goes to stderr": _reclaim writes its refusals to stdout here (run(args, registry, sys.stdout), storage_cli.py:318-327).

**`doc/confidential.md`** (line 93): "It is a dependency of the Debian package from 2.1 on" — packaging/aleph-vm/DEBIAN/control on this branch has no pciutils (and still depends on python3-jsonschema, which the agent README edit here removes from the manual apt line). True only once the packaging branch lands; worth gating or rewording so the two don't disagree on main in between.

foxpatch-aleph (changes requested):
This is a documentation-only PR whose claims I verified extensively against the code on this branch and on the in-flight branches it cites, and nearly everything checks out: every settings default, the storage CLI verbs and exit codes, the v2 allocation answer shape and buckets, the closed set of eight failure codes and their published sentences, the stopped-VM and crash-backoff rules, and the confidential-GPU advertisement gating all match the implementation (including byte-for-byte matches against od/2.1-alloc-executions-list-sanitized and od/2.1-alloc-quality). However, two statements are factually wrong: storage-reclamation.md says a missing file at $ALEPH_VM_ENV_FILE is logged and the command continues, when the code refuses with exit 1 the same as a missing --env-file; and scheduler-allocations.md's retained-reason list omits the reachable value 'refused'. For a PR whose whole value is accuracy, these should be fixed before merge; both are one-line edits.

**`doc/storage-reclamation.md`** (line 122): This contradicts the code. A file named via $ALEPH_VM_ENV_FILE is treated exactly like an explicit --env-file: _env_file_path() in src/aleph/vm/agent/storage_cli.py:836-848 returns named=True for both, so a missing file at $ALEPH_VM_ENV_FILE logs 'No environment file at ...' and exits 1 — it does not continue on the process environment. Only a missing file at the default path (/etc/aleph-vm/supervisor.env) is logged and tolerated. The exit-code table row 1 ('a --env-file was named and does not exist') should also mention $ALEPH_VM_ENV_FILE for the same reason. As written, an operator will expect the command to continue where it actually refuses.

**`doc/scheduler-allocations.md`** (line 52): The retained-reason list is missing 'refused'. compute_verdict (src/aleph/vm/agent/allocation/verdict.py:206) publishes retained[vm_hash] = "refused" for a VM that is running here and whose hash the push named but the node refused — present on this branch and on every in-flight branch in the stack. Since the stated goal is that an external consumer need not read source, a scheduler parsing the retained map will hit an undocumented value. Suggest adding 'refused (the push named it and this node turned it down)' to the list.

**`src/aleph/vm/agent/README.md`** (line 79): Nit, pre-existing: the shell block above ends with four backticks, so some renderers treat everything after it as still inside the code fence. Not introduced by this PR, but since this section was reworked it is a cheap fix while you are here.

foxpatch-aleph (commented):
This is a docs-only PR and I verified its factual claims against the code on this branch in unusual depth: every settings default, both CLI table layouts, the verdict buckets, retained reasons, allocation states, exit-code semantics, auth details (8 MiB cap answering 413 before the 401 check), the TDX rejection message, and the nvidia_cc advertisement rules all match the source exactly, and all README links resolve. The agent README fix is correct: the apt line matches the deb's dependency list and the pyproject pins it names all exist. Two things need a look before merge. First, the docs describe behavior from PR #1229 that does not exist on this branch: verdict.py:109 keys an unreadable-hash rejection by echoing the raw string the push sent, while the doc says it is answered as `vms[<i>]` — precisely the behavior the doc says is avoided — and the eight `error.code` values in the table also appear nowhere in this tree. That is fine for a stacked PR, but this must not merge before #1229, and the discrepancy should be conscious. Second, the storage exit-code table undersells exit 2: it is also returned for invalid node configuration, not just argparse usage errors.

**`doc/scheduler-allocations.md`** (line 39): This row does not match the code on this branch: `build_plan` keys an unreadable-hash rejection by echoing the raw string back (`rejected[str(raw_hash)]`, src/aleph/vm/agent/allocation/verdict.py:109) — exactly the unbounded echo the third paragraph says is avoided. The `vms[<i>]` scheme appears only in the doc, so it must come from #1229. Same for the eight `error.code` values at lines 150-159, which exist nowhere in this tree. Both are fine for a stacked PR, but this branch then has a hard merge-order dependency on #1229 — if this lands first, the doc documents rejection keying and failure codes the node does not emit. Worth an explicit statement in the PR description that this merges after #1229.

**`doc/storage-reclamation.md`** (line 180): Exit 2 is not only argparse: `run_parsed` also returns 2 when loading the env file raises a pydantic `ValidationError` (src/aleph/vm/agent/storage_cli.py:893-897), i.e. an invalid node configuration value such as a misspelled `ALEPH_VM_CACHE_BUDGET`. A wrapper distinguishing a typo in its own command line from a bad node config would want that second cause on this row.

foxpatch-aleph (approved):
Documentation-only PR that is accurate to an unusual degree: every settings default, endpoint response shape, exit code, error string, and packaging claim I could check against the source matches exactly, including subtle behaviors like the 413-not-401 body-cap design, the env-file named-vs-default distinction, and the `Invalid node configuration for <field>` message. The retained-reason set, allocation state enum, and headroom keys all match the code. The only caveat is that a meaningful subset of the documented behavior (the `vms[<i>]` rejection keys, the closed 8-code error set, stopped-VMs-stay-stopped, the per-hash create lock, `unknown`-instead-of-`0 B`, and the read-only `status`/`list` guarantee) does not exist in this branch's tree and comes from #1229 and the in-flight storage/allocation branches; the PR declares this merge order, so it is fine, but the doc will describe behavior the node does not have if this lands ahead of its dependencies or if those branches change before merge. Recommend re-diffing those sections against the final state of #1229 and `od/2.1-storage-cli-readonly-and-diagnostics` at merge time.

**`doc/scheduler-allocations.md`** (line 66): Merge-order note: in the base tree, `compute_verdict` answers `unchanged` only for `LIVE_STATUSES` (running/booting/defined), so a held-but-stopped VM falls through to the candidate path here, and `LIVE_STATUSES` in `reconciler.py` means the convergence loop restarts it. The rule as written (unchanged in any state, stopped stays stopped) is the behavior of the in-flight allocation branch, not this tree — correct only once `od/2.1-alloc-inflight-removing` lands, and worth re-checking against its final state.

**`doc/scheduler-allocations.md`** (line 145): The closed eight-code set and the fixed published sentences are #1229's `AllocationFailureCode`; the tree currently publishes `str(error)` (raw exception text, truncated to 200 chars) and codes derived from exception names. Given the PR merges after #1229 this is fine, but the exact `error.message` sentences should be re-diffed against #1229's final wording before merge so the published table does not drift from what the node actually emits.

**`doc/storage-reclamation.md`** (line 87): In this branch's `_status` (src/aleph/vm/agent/storage_cli.py:517-544), an OSError from `shutil.disk_usage` sets `total = free = 0`, so an unreadable pool prints `0 B`, not `unknown`. The `unknown` behavior is on `od/2.1-storage-cli-readonly-and-diagnostics`; accurate only after that branch lands.

**`doc/storage-reclamation.md`** (line 102): Same dependency: in this branch, `run_parsed` unconditionally runs `settings.setup()` (which `makedirs`s the caches and the execution root) and `storage_pools.setup_pools()` before every verb, including `status` and `list`. The literal read-only guarantee described here is implemented on the in-flight storage CLI branch, not in this tree — correct only once it lands.

PR #1231, opened 2026-09-09
refactor(agent): storage code quality: enums, shared helpers, honest comments

odesenfans:
## Summary

The code-quality pass over the agent's storage modules for 2.1: the duplicated
helpers, the untyped and unvalidated boundaries, the stale comments, and the
small correctness items the S3, S4 and S5 reviews left behind. Twenty commits,
one per item, each with its own tests.

Six of them are behaviour fixes rather than refactors, and they are the reason
this is not a pure cleanup:

- The cache pass purges retained VM directories to free the parent images they
  pin and reported none of it, so a pass that deleted a VM's disks logged
  `evicted=0, freed=<a few kilobytes of cache entries>`, and so did the startup
  preview an operator reads before an upgrade's first pass.
- Writing the reclaimable marker moves the directory's mtime, so every VM
  retired under `keep` looked like a create in flight for the next
  `VOLUME_CREATE_GUARD` and kept its device-mapper devices, which are precisely
  what makes the purge refuse the directory: the marker delayed the reclamation
  it announced.
- Download admission evicted against one measure (bytes really on disk) and
  refused against another (those plus the room an unmeasured download is
  holding), so a measured download could evict every unreferenced entry on a
  root and still be refused. Cache lost, create failed.
- A retire cleared the live snapshot the admission hook reads but not the last
  supervisor listing the room maker reads, so a placement could refuse to take
  back the disks of a VM that is gone until the next pass, up to an hour later.
- A create that failed while another create for the same hash was still running
  restored a reclaimable marker onto the directory that create was building,
  and a marker on a live directory is counted as reclaimable capacity the node
  offers to someone else.
- `_reconcile_namespace` re-asked the liveness question on the line after
  `_is_orphan` had answered it, and then walked every pool measuring the VM
  before purging it: the repeat could catch nothing, and the window that does
  exist (the pass runs in a worker thread, creates enter `creating()` on the
  event loop) spanned the measurement and the purge unguarded.

Stacked on #1219.

## Changes

- `agent/vm/reconciler.py`: `SideDirNaming` enum replaces the `"exact"` /
  `"prefix"` mode strings threaded through three functions and owns the
  name-to-hash conversion that was open-coded twice; the last liveness check
  moves to immediately before the mark or the purge; `_within_create_guard`
  skips a directory that carries a marker; `forget_supervisor_hash`;
  `creating()` leaves the marker to a create still running; the cache pass's
  reclaims are folded into the report; `_dir_bytes`, `_still_on_disk`,
  `_plausible` and `_reclaimable_on` give way to shared implementations;
  `teardown_orphan_devices` and `release_cache_devices` lose their underscore.
- `agent/vm/reclaimable.py`: `from_json` validates the reason and the schema
  version (an unknown version is kept and reported, not removed as corrupt);
  `file_size_bytes` uses one lstat and `S_ISREG` instead of three stats;
  `reclaimable_entries` is the single listing (and eviction order) the cache
  pass and the retention budget share; `namespace_size_bytes` moves here.
- `agent/vm/cache.py`: `admit_download` decides the refusal before evicting and
  reads the cache disk once; `_unknown_length_charge` loses a dead OSError
  branch that held the whole budget; `_iter_loop_backings` is the one sysfs
  walk; `_record_refs` is typed; the reclaim ledger reaches the caller.
- `storage.py`: `remove_devmapper` calls `remove_base_device` for its last step
  rather than repeating the sibling scan that function is.
- `storage_pools.py`: `has_namespace_dirs`, the one answer to whether a VM
  still has directories.
- `agent/vm/retire.py`: drops the retired hash from the last supervisor
  listing; the teardown guard's comment no longer claims to be unreachable
  (the reinstall path calls that function directly).
- `docs/architecture/storage.md`: the transient overshoot when a chunked body
  outgrows its reserve, and the renamed helpers.

`agent/storage_cli.py` is owned by another change in flight and is untouched.
The four reconciler names it imports (`_plausible`, `_still_on_disk`,
`_teardown_orphan_devices`, `_release_cache_devices`) are kept as aliases at
the bottom of the module; they go when that module is pointed at the real
names.

## Test plan

- `tests/supervisor/test_reclaimable.py`: an unknown marker reason does not
  parse and is removed as corrupt, a marker from a newer schema does not parse
  and is kept; a symlink, a dangling symlink, a directory and a missing file
  all measure 0.
- `tests/supervisor/test_reconciler.py`: a create landing during the
  measurement keeps its directory; a VM the cache pass reclaims reaches
  `report.evicted` and `bytes_freed`; a marker write does not re-age a
  directory for the device teardown; a failed create leaves the marker to the
  create still running.
- `tests/supervisor/test_cache_eviction.py`: a refused download evicts nothing;
  the cache disk is read once per admission and the reserve comes from that
  figure; a body bigger than its reserve is recovered by the next pass.
- `tests/supervisor/test_retire.py`: a record-dropping retire drops the hash
  from the last supervisor listing and a RECREATE does not; the record-less
  teardown test now has its asyncio marker and actually runs (it was silently
  skipped).
- `pytest tests/supervisor`: 1226 passed, 1 failed. The failure is
  `test_log.py::test_make_logs_queue` (journald stub), identical to the base
  commit's run (1211 passed, 1 skipped, same single failure). Fifteen tests
  added, and the base's one skip is gone.
- `ruff format --diff`, `isort --check-only --profile black`, `mypy src/aleph/vm/`
  and `mypy src/aleph/vm/ tests/`: all clean. No Rust touched.

## Not done, and why

`get_existing_file` bounding immutable volumes at `MAX_RUNTIME_ARCHIVE_SIZE`
"instead of the message's `size_mib`" is not implementable as described:
`aleph_message`'s `ImmutableVolume` has no size field (only `ref` and
`use_latest`), and neither does a V-PROGRAM's verified volume. Every other
caller of that function (runtime manifests, bundle tarballs, workload images
and hash trees, TEE firmware) fetches something no message states a size for
either. The cap stays, and its comment already explains why it is the runtime
one.

- `reconciler.py`: two overlapping creates of one hash could lose the markers between them. Only the first create in finds a marker to adopt, so when it failed with a second still running it left the restore to that one, which had adopted nothing. What was adopted is now held per namespace beside the count of creates in flight: nothing goes back while any create runs, a create that returns normally drops the markers for good, and the last create out with none committed restores whatever any of them adopted. The docstring and `docs/architecture/storage.md` say so, and a test drives the two-task overlap in the order that used to lose the marker.
- `cache.py`: `_evictable_bytes` is only computed when the root is over its budget, so an admission on a healthy cache no longer walks every pool's markers and stats sysfs per runtime entry on the event loop.
- `reclaimable.py`: `UnsupportedMarkerVersion` becomes `UnsupportedMarkerVersionError` (ruff N818), with its two references.
- Rebase note: the merged base moved under `agent/vm/cache.py`; this branch's own commit that threads the cache disk total down from `admit_download` removes the OSError fallback that #1211 had just reworded, so #1211's two fallback tests were adapted to the admission level, where the unreadable-disk path now lives.
- Checks on the rebased branch: `ruff format --diff`, `isort --check-only --profile black`, `mypy src/aleph/vm/ tests/` clean. Test suites are left to CI.

### DIFF
--- a/docs/architecture/storage.md
+++ b/docs/architecture/storage.md
@@ -545,7 +545,12 @@ truth as the body lands, since the bytes on disk are counted the moment they
 are written. A ceiling never sets the eviction target either: a root's usage
 counts a guessed reservation only at the bytes it has actually written when
 the question is how much to evict, and at the full figure when the question
-is whether the next download fits. What is admitted is then charged to
+is whether the next download fits. A body bigger than its reserve therefore
+takes its root over the budget while the extra bytes land: a `.part` is not an
+entry and nothing may unlink it under the download writing it. The overshoot
+is transient rather than permitted. Those bytes are a measurement, so the next
+pass counts them and evicts least recently used for them, and once the
+download finishes its entry is evictable like any other. What is admitted is then charged to
 the download's `.part` path (`reserve_download`) until `download_file`
 releases it, so a second create arriving while the first is still writing
 sees the room the first was promised rather than only the bytes it has
--- a/docs/architecture/storage.md
+++ b/docs/architecture/storage.md
@@ -685,7 +685,7 @@ in its `--help` epilog) needs no running *agent process*:
   - **The agent is down and the supervisor answers.** The full pass runs.
     Nothing is being created, so the age of the directory is a sound guard
     again, and the live set is a known one (registry union `list_vms`), so
-    `_teardown_orphan_devices` removes the dm devices of every namespace no
+    `teardown_orphan_devices` removes the dm devices of every namespace no
     live VM owns before the walk, or the purge that follows would be refused
     on a volume file a target still holds.
   - **The supervisor does not answer.** The live set is the registry alone,
--- a/docs/architecture/storage.md
+++ b/docs/architecture/storage.md
@@ -353,6 +353,12 @@ turn retained disks into an owner-less orphan directory with no parent-image
 pins and a timestamp that resets on every attempt. A directory the failed
 create's own teardown purged is not re-marked, and a marker written while
 the create ran (a retire of the same hash) is the newer record and stays.
+Two creates of one hash can overlap (a scheduler push beside an operator
+reinstall), and only the first of them finds a marker to adopt, so what was
+adopted is held per namespace rather than per create: while any create is
+still running nothing goes back, a create that returns drops the markers for
+good (the directory is a live VM's now), and the last create to leave with
+none of them committed restores whatever any of them adopted.
 
 ### The reconciler
 
--- a/docs/architecture/storage.md
+++ b/docs/architecture/storage.md
@@ -414,7 +414,7 @@ Everything a pass touches is under a directory the agent created and is
 keyed by a hash that `storage.vm_namespace` accepted: the name has to parse
 as an `ItemHash` (64 lowercase hex characters, or an IPFS CID), so a
 directory an operator dropped on a volume pool is never a VM. The delete
-paths go through `purge._checked_namespace`, which refuses on the same rule
+paths go through `purge.checked_namespace`, which refuses on the same rule
 before any filesystem access; the passes that walk the pools ask
 `is_vm_namespace` first and skip what the guard would refuse.
 
--- a/src/aleph/vm/agent/capacity.py
+++ b/src/aleph/vm/agent/capacity.py
@@ -26,7 +26,7 @@ from aleph_message.models import ExecutableContent, ItemHash, VerifiableProgramC
 from aleph_message.models.execution.instance import InstanceContent
 
 from aleph.vm import storage_pools
-from aleph.vm.agent.vm.purge import ROOTFS_STEM, _checked_namespace
+from aleph.vm.agent.vm.purge import ROOTFS_STEM, checked_namespace
 from aleph.vm.agent.vm.reclaimable import (
     MARKER_NAME,
     file_size_bytes,
@@ -228,7 +228,7 @@ def existing_volume_files(vm_hash: ItemHash | str) -> dict[str, Path]:
     Symlinks and the marker are skipped, and pools are walked in order with
     the first match winning, so the result does not depend on iteration luck.
     """
-    namespace = _checked_namespace(vm_hash)
+    namespace = checked_namespace(vm_hash)
     files: dict[str, Path] = {}
     for directory in storage_pools.iter_namespace_dirs(namespace):
         try:
--- a/src/aleph/vm/agent/vm/backup.py
+++ b/src/aleph/vm/agent/vm/backup.py
@@ -30,7 +30,7 @@ from typing import BinaryIO
 
 from aleph_message.models import ItemHash
 
-from aleph.vm.agent.vm.purge import ROOTFS_STEM, _checked_namespace, iter_volume_files
+from aleph.vm.agent.vm.purge import ROOTFS_STEM, checked_namespace, iter_volume_files
 from aleph.vm.backup.archive import (
     backup_metadata,
     check_disk_space_for_multiple,
@@ -85,7 +85,7 @@ def purge_vm_backups(vm_hash: ItemHash | str) -> int:
     primitive: it is interpolated into a glob pattern here, and a delete path
     never builds a pattern or a path out of an unchecked name.
     """
-    namespace = _checked_namespace(vm_hash)
+    namespace = checked_namespace(vm_hash)
     backup_dir = get_backup_directory()
     removed = 0
     for archive in backup_dir.glob(f"{namespace}-*.tar"):
--- a/src/aleph/vm/agent/vm/cache.py
+++ b/src/aleph/vm/agent/vm/cache.py
@@ -38,7 +38,7 @@ import json
 import logging
 import os
 import shutil
-from collections.abc import Callable, Collection, Iterable
+from collections.abc import Callable, Collection, Iterable, Iterator
 from dataclasses import dataclass
 from pathlib import Path
 
@@ -357,27 +357,30 @@ def parent_refs_of(evicted: list[Path]) -> list[str]:
     return [path.name for path in evicted if path.parent == runtime]
 
 
-def _loop_devices_backing(path: Path) -> list[str]:
-    """The loop devices backed by ``path``, found through sysfs.
+def _iter_loop_backings() -> Iterator[tuple[str, str]]:
+    """``(loop device, backing file)`` for every loop device the kernel lists.
 
-    ``storage.detach_loop_devices`` asks ``losetup -j``, which has to stat
-    the backing file; an evicted cache entry is already unlinked, and the
-    loop that still pins its blocks is exactly the one that has to go. sysfs
-    keeps the path, marked " (deleted)".
+    Read from sysfs rather than with ``losetup -j``, which has to stat the
+    backing file: an evicted cache entry is already unlinked, and the loop
+    that still pins its blocks is exactly the one that has to go. sysfs keeps
+    the path of an unlinked file, marked " (deleted)", and the marker is left
+    on the string here because it is what tells the two callers apart.
     """
-    devices: list[str] = []
     try:
         backing_files = sorted(SYS_BLOCK.glob("loop*/loop/backing_file"))
     except OSError:
-        return devices
+        return
     for backing_file in backing_files:
         try:
             backing = backing_file.read_text().strip()
         except OSError:
             continue
-        if backing.removesuffix(DELETED_SUFFIX) == str(path):
-            devices.append(f"/dev/{backing_file.parent.parent.name}")
-    return devices
+        yield f"/dev/{backing_file.parent.parent.name}", backing
+
+
+def _loop_devices_backing(path: Path) -> list[str]:
+    """The loop devices backed by ``path``, unlinked or not."""
+    return [device for device, backing in _iter_loop_backings() if backing.removesuffix(DELETED_SUFFIX) == str(path)]
 
 
 async def remove_parent_device(ref: str) -> None:
@@ -832,20 +835,12 @@ def _deleted_cache_backings() -> list[tuple[str, Path]]:
     cache entry that is already unlinked."""
     roots = cache_roots()
     leaked: list[tuple[str, Path]] = []
-    try:
-        backing_files = sorted(SYS_BLOCK.glob("loop*/loop/backing_file"))
-    except OSError:
-        return leaked
-    for backing_file in backing_files:
-        try:
-            backing = backing_file.read_text().strip()
-        except OSError:
-            continue
+    for device, backing in _iter_loop_backings():
         if not backing.endswith(DELETED_SUFFIX):
             continue
         path = Path(backing.removesuffix(DELETED_SUFFIX))
         if path.parent in roots:
-            leaked.append((f"/dev/{backing_file.parent.parent.name}", path))
+            leaked.append((device, path))
     return leaked
 
 
--- a/src/aleph/vm/agent/vm/cache.py
+++ b/src/aleph/vm/agent/vm/cache.py
@@ -50,6 +50,7 @@ from aleph.vm.agent.vm.reclaimable import (
     ReclaimableMarker,
     file_size_bytes,
     iter_content_refs,
+    namespace_size_bytes,
     reclaimable_entries,
     refs_from_content,
 )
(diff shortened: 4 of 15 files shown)


foxpatch-aleph (approved):
Approve. All six behavior fixes are traced end-to-end and correct: the refusal in admit_download is decided before any unlink (a refused download evicts nothing, and the optimistic estimate is only used to decline, never to commit); the create-guard marker skip is safe because creating() clears markers before a create writes, so a mid-create directory never carries one; forget_supervisor_hash sits after the RECREATE early-return so recreates keep their hash; the nested-creating() fix restores markers only when the last owner fails; _reconcile_namespace measures before the final liveness re-check, closing the purge-a-live-VM window; and reclaimed_bytes from evict_caches is folded into the cache report without double counting. The renames are non-breaking (underscore aliases kept for storage_cli.py, no stale references in src/), the new imports introduce no cycles, and no security issues were found (namespaces are hash-validated before glob/dmsetup use). The four changed test files pass 204/204; the remaining tests/supervisor failures reproduced here are all PermissionError on /var/cache/aleph from settings.setup(), i.e. sandbox limitations, not regressions. The two comments below are minor performance/style nits.

**`src/aleph/vm/agent/vm/cache.py`** (line 821): Minor: `_evictable_bytes` is computed even when `usage <= budget`, in which case the refusal condition `usage - evictable > budget` is impossible whatever the estimate says. On a healthy cache every admitted download therefore pays the walk over `reclaimable_entries()` (a stat+read of every `.reclaimable` marker across the pools) for nothing. Gating it — `if may_evict and usage > budget else 0` — keeps the behavior identical and skips the walk on the common path. The `evict_caches` call below could take the same shortcut, though its `needed`-driven stop makes that one cheaper already.

**`src/aleph/vm/agent/vm/reconciler.py`** (line 618): Nit: `naming.namespace_of(child)` is computed here and again inside `_is_stale_side_dir` (line 568) for every stale candidate. Returning the namespace from `_is_stale_side_dir` (e.g. `str | None` instead of a bare bool) would compute it once and let the caller log the namespace without re-deriving it.

foxpatch-aleph (approved):
A high-quality quality pass: the renames are mechanical and applied consistently (storage_cli.py keeps importing via documented transitional aliases), and the six behavior fixes are real, well-reasoned, and each pinned by a test. I verified the new admission pre-check is arithmetically sound (usage - evictable > budget is exactly "a full eviction would still leave it over budget" under the same ceilings-inclusive measure the old post-eviction check used), that _within_create_guard's marker skip is safe because creating() registers before adopting, that read_marker's new exception clause is ordered correctly before ValueError, and that the remove_devmapper/remove_base_device dedup preserves behavior. All storage-related tests pass in this environment (the wider suite's failures are root-privilege paths untouched by this diff), and ruff format/isort are clean with the pinned versions. The remaining findings are non-blocking: an unconditional full marker/sysfs walk on every download admission even when the root already fits, a narrow concurrent-create race in which the adopted markers can be lost entirely despite the creating() docstring claiming otherwise, and an exception name that violates the project's selected ruff N818 rule (not CI-enforced, since CI only runs ruff format).

**`src/aleph/vm/agent/vm/cache.py`** (line 821): Minor perf: `_evictable_bytes` runs on every download admission whenever `may_evict` is true, even when `usage <= budget` (the common case). It walks every pool's namespace directories reading markers (`reclaimable_entries()`), recomputes `live_refs(registry)`, and does a sysfs stat per runtime-cache entry via `_may_evict` — all on the event loop, inside the download. Since the refusal branch can only trigger when `usage > budget`, computing it lazily (e.g. `evictable = _evictable_bytes(registry, entries) if may_evict and usage > budget else 0`) keeps the new semantics at zero extra cost for a root that already fits. Before this PR an under-budget root short-circuited in `evict_caches` before the marker walk.

**`src/aleph/vm/agent/vm/reconciler.py`** (line 143): Edge case in the new skip-restore rule: when create A adopts the markers and then fails while create B for the same hash is still running, A correctly skips the restore — but B adopted nothing, so if B also fails, `restore_markers` puts nothing back and the marker is lost entirely. The next pass re-marks the directory as reason "orphan" with no owner and no depends_on, which is precisely the degradation `restore_markers` exists to prevent (lost erase authorization and parent-image pins). The docstring's claim that the still-running create "puts back what it adopted if it fails in its turn" is inaccurate in this scenario, since the second create adopted nothing. The nested-in-one-task case is fine (the outer's except runs after the inner exits, so the count is back to 1 and it restores), but the two-task overlap this module's own comment at line 96 says can happen is not. Worth either softening the docstring or sharing the adopted markers per-namespace so the last create out restores them.

**`src/aleph/vm/agent/vm/reclaimable.py`** (line 41): Nit: ruff's N818 (selected via the "N" ruleset in pyproject) wants exception classes named with an Error suffix — `UnsupportedMarkerVersionError`. CI only runs `ruff format`, so this won't fail anything, but the rest of the file's exception naming conventions are otherwise consistent.

foxpatch-aleph (approved):
A careful, well-tested quality pass in which all six behavior fixes are genuine and correctly implemented. The creating() shared-state rework survives exhaustive scenario analysis (overlapping creates, nested creates, retire re-marking mid-create), the admission refusal-before-eviction fix uses an estimate that errs in the safe direction, and the marker/guard/report fixes are each backed by a test that drives the original failure. All 241 tests in the touched files pass in my environment, imports are cycle-free, and the doc updates match the code. The remaining findings are narrow races or corners that do not block.

**`src/aleph/vm/agent/vm/reconciler.py`** (line 298): This is the first in-place mutator of `_last_supervisor_hashes` — every other writer (in `_live_set`) rebinds it. `known_live_hashes` can iterate this set from a worker thread (the room maker is invoked via `asyncio.to_thread` on the placement path), and CPython raises `RuntimeError: Set changed size during iteration` if a `discard` lands mid-iteration. The window is tiny and the consequence is one failed placement that retries, but rebinding a fresh set (e.g. building `known_live_hashes`'s union over a copy, or storing a new set here) would close it for free.

**`src/aleph/vm/agent/vm/reconciler.py`** (line 191): Minor corner worth a docstring note: `committed` here means "this context manager returned normally", which for a *nested* `creating()` span (an inner volume rebuild inside an outer create) does not mean the VM went live. If the inner span returns normally and the outer create then fails, the shared `adopted` has been cleared and nothing is restored — the directory falls to the orphan flow and is re-marked without its owner and pins, where the pre-PR code restored the marker. I couldn't find a production call path that nests spans, and the docstring's stated contract ("a create that returns normally commits") matches the code, so this is observation, not a request.

**`src/aleph/vm/agent/vm/reconciler.py`** (line 476): The measurement now runs even when the liveness re-check on the next line will skip the namespace, so a directory claimed mid-walk costs a full multi-pool walk for nothing. Negligible in practice, but if it ever matters, moving the measurement after the liveness check for the `keep` branch only (where `freed` is never used) would save it.

**`src/aleph/vm/agent/vm/reclaimable.py`** (line 369): `adopt()` clears a marker even when `read_marker` returned None because the schema version is unknown, so a create for that hash still unlinks a newer agent's marker that `read_marker` deliberately keeps. The create is going to write the directory anyway, so this is defensible, but it does mean the v1/v2 interop story for kept markers ends the moment the hash is re-created.

foxpatch-aleph (approved):
Thorough storage-quality refactor. All six behavior changes are correct: the _CreateState sharing between overlapping creates is sound in every commit/fail ordering I traced, admit_download correctly decides refusal on an optimistic evictable estimate before evicting, forget_supervisor_hash is dropped only for record-dropping retires, and the marker-aging, create-guard and cache-report changes behave as described. Docs match the code. The 29 full-suite failures I could reproduce are all environment artifacts (outdated installed aleph_message, a sqlalchemy version mismatch, and unprivileged /var/cache/aleph) — every test file the PR touches passes completely. Two minor nits, neither blocking.

**`src/aleph/vm/storage_pools.py`** (line 494): Nit: `any(True for _ in iter_namespace_dirs(namespace))` is a roundabout way to ask whether the iterator yields anything. `next(iter_namespace_dirs(namespace), None) is not None` (or restructuring `iter_namespace_dirs` to yield the dirs and using `bool(...)`) would say it directly.

**`src/aleph/vm/agent/vm/reclaimable.py`** (line 90): Minor observation, no change needed: an unhashable `reason` (e.g. a JSON list instead of a string) makes `reason not in RECLAIM_REASONS` raise TypeError rather than ValueError. That TypeError is caught by read_marker's corrupt branch, so the marker is removed either way — the classification works, it just leans on the broader exception set. Confirmed intentional-looking and correct.

foxpatch-aleph (commented):

foxpatch-aleph (approved):
A high-quality code-quality pass that is more than cleanup: the six behavior fixes are genuine (unreported reclaim bytes, marker-write re-aging directories into the create guard, admission evicting-then-refusing on two different measures, stale supervisor listings blocking room-making, lost markers between overlapping creates, and a mis-ordered liveness check before a purge). I traced the new creating() state machine through every overlapping-create interleaving and it is correct in each; the admit_download reordering decides refusal against the same _is_evictable predicate the eviction uses, so nothing is evicted for a download that is then refused; _within_create_guard's marker skip is sound because creates register in _creating and adopt before writing. All renames are complete with no stale references (the four underscore aliases match storage_cli.py exactly), the version/reason marker validation has correct exception ordering, and the previously silently-skipped async retire test now runs. All PR-relevant tests pass in this sandbox (241 + 198 + 23 across reconciler, reclaimable, cache eviction, retire, purge, capacity, storage, CLI, devmapper); broader failures and lint/type findings are environment artifacts in untouched code, and CI only enforces ruff format, which the changed files satisfy. Remaining comments are cosmetic.

**`src/aleph/vm/agent/vm/reclaimable.py`** (line 159): Nit: the docstring promises "the operator is told which agent has to look at it", but the log line below only reports the marker path and the version error — it does not identify any agent or version to compare against. Either drop the claim or add the running agent's version to the log message.

**`src/aleph/vm/agent/vm/cache.py`** (line 890): Nit: this docstring closes with `."""` on the text line, unlike every other docstring in the file which puts the closing quotes on their own line.

**`src/aleph/vm/agent/vm/reconciler.py`** (line 1150): Fine as documented scaffolding for the in-flight storage_cli change — just make sure the follow-up that repoints the CLI drops these aliases rather than leaving them behind; `_plausible`/`_still_on_disk` in particular are easy to mistake for the real API from a grep.

foxpatch-aleph (approved):
A well-executed quality pass with six genuine behavior fixes, each correctly reasoned and tested. The creating() marker-sharing refactor holds up under every interleaving I traced (nested and non-nested overlap); the admission reordering decides refusal before evicting against a single disk reading; the marker-aware create guard fixes a real reclamation deadlock; and the report now counts the retained disks the cache pass purges. All 241 tests in the changed files pass, mypy and ruff format are clean on the changed modules, the storage CLI aliases resolve, and the documentation matches the code. No correctness or security issues found.

**`src/aleph/vm/agent/vm/reconciler.py`** (line 960): Minor, non-blocking: the marker skip in _within_create_guard is sound in-process (is_creating is checked first), but a marker written by a same-hash retire landing while a create is in flight re-enables this window for the out-of-process storage CLI, which cannot see the creating set and previously would have been protected by the fresh mtime. That sequence is already a broken/racy one (the retire purges or marks the very storage the create is building), so I would just note it in the comment or docstring for the next reader rather than change the code — the fix this makes (retired VMs no longer keeping their dm devices for a whole guard, blocking reclamation) is worth far more than the residual window.

**`src/aleph/vm/agent/vm/reconciler.py`** (line 485): Cosmetic, in an already exotic scenario: when mark_reclaimable's exclusive write is refused (e.g. a kept, newer-schema marker from a downgraded agent), report.marked_orphans still records the namespace even though nothing was written. If you ever touch this path again, distinguishing "wrote" from "was refused" in the report would make the operator log exactly honest; not worth blocking anything.

foxpatch-aleph (approved):
Correct, well-structured, and thoroughly tested storage refactor. All six behavior fixes are sound: the cache-pass accounting fix, the marker re-aging fix, refusal-before-eviction in download admission (verified: a refused download evicts nothing), retire dropping the hash from the last supervisor listing, shared adopted-markers across overlapping creates (all commit/fail permutations check out, including exclusive restore so a mid-create GONE marker survives), and the post-measurement liveness re-check. Security is solid: namespaces validated, marker JSON schema-validated, dm-guarded purges. All 247 tests in the changed files pass. Remaining comments are cosmetic and non-blocking.

**`src/aleph/vm/agent/vm/reconciler.py`** (line 176): Minor: `setdefault(namespace, _CreateState())` constructs a throwaway `_CreateState()` on every call even when the key exists. Harmless, but `_creating.get(namespace)` with a conditional insert would avoid the churn if this path is hot. Not worth blocking on.

**`src/aleph/vm/agent/vm/reclaimable.py`** (line 88): Question (non-blocking): a marker with an unknown `version` is kept intact (UnsupportedMarkerVersionError handled before the corrupt-marker path), but a marker with an unknown `reason` is deleted as corrupt. If a reason is ever added in the future the asymmetry is fine (reasons are a closed set here), but worth confirming the divergence in treatment is intentional — an unknown-version marker may describe data a future agent can still interpret, while an unknown-reason one cannot.

**`src/aleph/vm/agent/vm/cache.py`** (line 1): Minor docs nit: the `_unknown_length_charge` docstring ends flush against the closing quotes. Cosmetic only.

PR #1232, opened 2026-09-09
fix(agent): refuse a v1 allocation push once a plan governs the node

odesenfans:
## Summary
An allocation plan is total: the reconciler tears down every supervisor VM the plan does not list. The legacy `POST /control/allocations` push names only the VMs one scheduler knows about, so a scheduler mixing the two modes on one CRN had every VM it started through the legacy route swept within `ALLOCATION_RECONCILE_INTERVAL` (disks reaped, records dropped) and then started again by the next push, with the push and the loop undoing each other for as long as it kept pushing. The legacy stop loop is just as destructive the other way round: it retires every running VM the body leaves out, which under a plan is most of them. This PR makes it one mode per node: once the reconciler holds a plan, the legacy route answers 409 Conflict and does nothing.

Stacked on #1229.

## Changes
- `src/aleph/vm/agent/allocation/reconciler.py`: `has_plan()`, "is this node governed by a plan". Not the same question as "does the plan list anything": the empty plan is the instruction to run nothing here, and a node under it is still under it.
- `src/aleph/vm/agent/views/__init__.py`: `update_allocations` refuses with 409 and `{"success": false, "code": "governed_by_v2_plan", "message": ..., "endpoint": "/v2/control/allocations"}` when a plan governs the node. The check sits right after authentication, before the stop loop and before any start. The docstring states the one-mode-per-node rule. `notify_allocation`, the owner-signed single-VM path, is not a scheduler push and is untouched.
- `docs/architecture/vm-lifecycle.md`: one paragraph on the rule, next to the allocation loop's stop semantics.

## Test plan
- `tests/supervisor/test_allocations_v2.py`, three tests through the real app and the real reconciler:
  - `test_a_v1_push_is_refused_once_a_v2_plan_governs_the_node`: after a plan submit, a legacy push answers 409 with the code and the endpoint, and starts nothing (failed before the fix with 200).
  - `test_a_v1_push_refused_by_the_plan_tears_nothing_down`: the refusal comes before the stop loop, so a legacy push with an empty body does not retire the VM the plan wants kept (failed before the fix).
  - `test_a_v1_push_still_starts_vms_before_any_v2_plan`: the compatibility half, a node no plan has reached starts what the legacy route gives it (green before and after).
- pytest: the allocation files, the v1 view tests and the create-lock test, 254 passed; `tests/supervisor/views` 86 passed.
- `ruff format --diff`, `isort --check-only --profile black` on the three changed Python files: exit 0. `mypy` the CI way (`src/aleph/vm/ tests/ examples/example_fastapi runtimes/aleph-debian-12-python`): Success, 211 files.


foxpatch-aleph (approved):
This PR is correct and well-executed. The distinction between 'governed by a plan' and 'the plan lists something' is the right one, and it is documented at both the method and the doc level. The 409 guard sits exactly where it must: after authentication (so unauthenticated callers learn nothing) and before the stop loop (which is the destructive half the second new test pins down). The three new tests are meaningful, go through the real app and real reconciler, and all related suites pass (23 + 30 tests verified locally, ruff clean). Two non-blocking observations: (1) there is a small TOCTOU window between the has_plan() check and the stop loop where a concurrently-submitted plan would not be seen, but the plan's next reconcile pass sweeps whatever the push did, so it self-heals; (2) because _desired is in-memory, an agent restart briefly re-opens the legacy route on a plan-governed node until the next plan push — consistent with the reconciler's documented post-restart behavior, but worth a sentence in the docstring. Minor nit: an oversized legacy body on a governed node answers 413 (cap enforced by the auth verifier) rather than the 409, which is slightly surprising but harmless.

**`src/aleph/vm/agent/views/__init__.py`** (line 615): Non-blocking: there is a TOCTOU window here — a v2 plan submitted while this handler is parked in an await (request.json(), list_vms(), a start) will not be seen, and the legacy push proceeds against a newly governed node. It self-heals on the next reconcile pass, and it only arises when a scheduler is actively mixing modes, so it is acceptable; worth keeping in mind if the guard is ever relied on for isolation rather than error reporting.

**`src/aleph/vm/agent/allocation/reconciler.py`** (line 168): Since _desired is in-memory, an agent restart reverts a plan-governed node to v1 mode until the next plan arrives (consistent with _converge_once's 'no plan means delete nothing'). Consider saying so in the docstring so operators know a restart briefly re-opens the legacy route on plan-governed nodes.

foxpatch-aleph (approved):
Small, correct, and well-fenced change. The 409 gate is placed at exactly the right spot (after auth, before the stop loop and any starts), and has_plan() draws the right line between 'a plan governs' and 'the plan lists something' — the empty plan still governs, and narrow_plan() always yields a real plan so even a fully-rejected v2 push locks the legacy route out. The in-memory-only semantics deliberately mirror the reconciler's own post-restart rule, and the docstring and docs paragraph are accurate. The three new tests drive the real app plus the real reconciler, so has_plan() is genuinely exercised, and I verified no existing legacy-route test hits a false 409 (they all use the real reconciler with no plan submitted). All relevant tests pass locally (the only failures are a pre-existing dependency skew in the v-program create path, untouched by this diff). Only one non-blocking observation about a race window that resolves itself in the plan's favor.

**`src/aleph/vm/agent/views/__init__.py`** (line 615): Non-blocking observation: this check is a point-in-time read, and a v2 push can still land between the has_plan() check and this handler's own awaits (request.json(), list_vms(), the starts). That push would then run to completion and the reconciler would sweep what the plan doesn't list on the next pass — last writer wins with the plan authoritative, which is the right tiebreak, so no change requested. Might be worth one line in the comment noting the residual window, alongside the existing before-the-stop-loop rationale, since the comment currently reads as if the check were the only ordering that matters.

**`tests/supervisor/test_allocations_v2.py`** (line 49): Good catch adding has_plan=MagicMock(return_value=False) here rather than letting the MagicMock auto-stub return a truthy value — that would have silently turned every legacy-route test into a 409. Related nit: the SimpleNamespace double in tests/supervisor/test_executions_list_allocation.py (_reconciler()) has no has_plan, so it can't be reused for a test that hits the legacy route; not worth changing now, just something to remember if that double gets copied.

foxpatch-aleph (approved):
Correct, well-tested, and well-documented fix for a real and nasty failure mode: with both allocation modes live on one node, the total plan and the partial v1 push would fight forever, each tearing down what the other wanted. has_plan() is exactly the right predicate (a plan governs whether or not it names anything, and it is only ever set by the v2 route), the 409 is correctly placed after auth and before the stop loop, and the tests pin down both the refusal and the ordering that makes it safe, plus the compatibility half. The mock fix in _app() (has_plan=MagicMock(return_value=False)) is necessary and easy to miss. The residual TOCTOU window between the has_plan() check and the per-VM teardowns is non-blocking: it only bites a request already in flight at the instant a plan lands, and the consequence is a one-time rebuild rather than indefinite thrash, though the codebase's own standard (the reconciler re-reads _desired before each teardown because GONE reaps volumes) suggests re-checking has_plan() inside the stop loop would be in keeping.

**`src/aleph/vm/agent/views/__init__.py`** (line 615): Non-blocking: this check is not re-done inside the stop loop below, so a plan submitted while a v1 push is parked in request.json(), allocation_lock or list_vms() still lets that in-flight push tear down VMs the plan wants kept (volumes reaped, then rebuilt by the loop). The reconciler guards this exact hazard in _teardown_dropped by re-reading _desired before each teardown, quoting that a teardown cannot act on a snapshot one push out of date since GONE reaps the volumes; a `if reconciler.has_plan(): break` at the top of the stop loop would match that standard and close most of the window. Consequence is bounded to one request at the moment of transition, so this is an observation rather than a request.

foxpatch-aleph (approved):
The change is correct, minimal, and well-placed: the 409 gate sits after authentication (preserving 401/413 precedence) and before body parsing, the allocation lock, the stop loop, and any start, which is exactly the ordering the teardown test pins. has_plan() is a sound synchronous read of in-memory state, and its docstring accurately describes the post-restart window in terms consistent with _converge_once's early return. The fixture change (has_plan returning False on the mock) was necessary and correctly avoids flipping existing legacy-route tests to 409. Tests are focused, each pinning a distinct behavior through the real app and real reconciler; all 23 pass locally (remaining suite failures are sandbox permission issues unrelated to this diff). Non-blocking notes: a plan submitted between the has_plan() check and the legacy stop loop still has its VMs exposed to that loop's teardowns (the reconciler converges afterwards, so harm is transient); the v2 endpoint path is duplicated as a literal in three places; and notify_allocation on a plan-governed node still exposes hold-based persistent VMs to the sweep unless the plan names them, which the lifecycle doc could state more explicitly.

**`src/aleph/vm/agent/views/__init__.py`** (line 615): Non-blocking: a plan submitted while this handler is parked in `list_vms()` or `request.json()` still has its listed VMs exposed to the stop loop below, since the check isn't repeated before each `teardown_vm`. The reconciler's own teardown re-reads `self._desired` for exactly this reason. Harm is transient — the loop rebuilds what the plan lists — so this is fine as-is, just worth being aware of.

**`src/aleph/vm/agent/views/__init__.py`** (line 624): The `"/v2/control/allocations"` literal now appears here, in the route registration (supervisor.py:420), and as the test's `PLAN` constant. A shared constant would prevent the three drifting apart.

**`docs/architecture/vm-lifecycle.md`** (line 293): It might be worth one more sentence here about the consequence for the unaffected route: a hold-based persistent VM started via `POST /control/allocation/notify` is still swept by the loop within the reconcile interval unless the plan names it (stream/credit PAYG VMs are exempt via `is_removable_by_allocation`). The PR is explicit this is deliberate, but the doc's "unaffected" could be read as "safe from the sweep", which it isn't.

foxpatch-aleph (approved):
This PR is a focused, well-reasoned fix: once a plan governs a node, the legacy allocation push is refused with 409 rather than being half-honoured against a total plan. The implementation is correct — has_plan() distinguishes 'governed' from 'plan lists anything' (the empty plan still governs), the check sits after authentication (401 takes precedence) and before the body parse and, critically, before the destructive stop loop, which the third test pins. notify_allocation is genuinely untouched as claimed. The reconciler's _desired is set-only, so the governance flag is stable until restart, which the docstring documents as deliberate. All new tests pass in this environment (23/23 in test_allocations_v2.py, 86 across the allocation files); the other failures I reproduced are pre-existing sandbox issues unrelated to this diff. The change is sound and I would merge it.

**`src/aleph/vm/agent/views/__init__.py`** (line 615): Non-blocking: the refusal is invisible on the node side. A scheduler mixing modes will keep receiving 409s, but nothing lands in the agent log, so the operator only learns of the misconfiguration from the caller's side. A logger.info("Refusing a v1 allocation push: this node is governed by a plan") here (or warning, since it means a v1 scheduler is still pointed at a v2 node) would surface it.

**`src/aleph/vm/agent/views/__init__.py`** (line 615): Worth noting the TOCTOU window: has_plan() is read before request.json() and before allocation_lock, so a plan submitted while a legacy push is in flight still lets that one push run its stop loop and starts once before the reconciler sweeps the result within ALLOCATION_RECONCILE_INTERVAL. It self-heals (the plan stays authoritative), so it is not a bug — but a sentence in the comment saying why the window is acceptable would match the rest of this file's reasoning style.

**`src/aleph/vm/agent/views/__init__.py`** (line 624): Nit: the "/v2/control/allocations" literal appears in both the response body and the docstring, while the route is registered in supervisor.py:420. A module-level constant shared with the registration point would keep them from drifting.

foxpatch-aleph (approved):
This PR correctly closes the destructive interference between the legacy v1 allocation push and the v2 plan reconciler: once a plan reaches the node, the legacy route answers 409 and does nothing. The check placement is the crux — after authentication but before the stop loop and before any start — and that ordering is explicitly regression-tested, including the case where a half-honoured push would otherwise reap VMs the plan wants kept. The has_plan() semantics are right (an empty plan still governs), the restart window is honestly documented and consistent with the reconciler's existing post-restart silence, notify_allocation is correctly left available, and the docs match the code. All 23 tests in test_allocations_v2.py pass locally; the only failures elsewhere are pre-existing environment issues in suites this diff does not touch. The minor observations below are non-blocking.

**`src/aleph/vm/agent/views/__init__.py`** (line 615): Non-blocking: the has_plan() read and the stop loop are separated by an await (request.json()), so a plan landing in that window is not caught by this refusal and the push proceeds as before. That is inherent to the check-then-act shape (submit() is deliberately await-free and lock-free), and the level-triggered loop heals the result within one reconcile interval, so it is not worth a lock. Might deserve one line in the docstring for the next reader, alongside the stop-loop comment.

**`src/aleph/vm/agent/views/__init__.py`** (line 619): Deliberate-looking and, I think, correct: the refusal precedes body validation, so a malformed body on a governed node answers 409 rather than 400. Worth a half-sentence in the comment saying so, since it is the one observable consequence of putting the check before the try/parse block rather than after it.

foxpatch-aleph (approved):
A minimal, well-targeted fix for a genuinely destructive interaction: the v2 plan is total (the loop tears down every VM it omits), so a legacy v1 push naming only one scheduler's VMs had the push and the reconcile loop undoing each other indefinitely, and the legacy stop loop could reap plan-governed VMs. The 409 is correctly placed after authentication (so node mode is not disclosed to unauthenticated callers) and before both the stop loop and the starts, which is the crux; has_plan() is properly distinct from planned_hashes() truthiness since the empty plan still governs. The documented restart window (in-memory plan means the legacy route briefly reopens) is consistent with the reconciler's existing post-restart silence, and leaving notify_allocation untouched is coherent because is_removable_by_allocation retains PAYG VMs, so the loop never fights that path. Tests cover both halves of the rule (compatibility before any plan, refusal after) plus the critical assertion that the refusal precedes the destructive stop loop, and all 23 tests in the touched file pass locally alongside the reconciler, executions-list, and v1 view suites. The one non-blocking observation: has_plan() is read outside allocation_lock and submit() takes no lock, so a plan landing between the check and the stop loop still gets fought once before the loop converges back — inherent to the design and not worth changing.

**`src/aleph/vm/agent/views/__init__.py`** (line 615): Non-blocking observation: this check runs outside allocation_lock and submit() takes no lock, so a plan submitted between the check and the stop loop below still gets fought once (the push retires VMs the new plan wants kept, until the next pass converges back). Inherent to the await-free submit() design and not worth surgery; the level-triggered loop bounds the damage to one interval. If you ever wanted to narrow it, re-checking has_plan() right after acquiring the lock would shrink the window, though submit() being lock-free means it can't be eliminated.

foxpatch-aleph (approved):
Clean, correct fix for a real footgun: mixing the partial legacy push with the total v2 plan made the push and the reconcile loop undo each other indefinitely, and the legacy stop loop could reap everything the plan kept. has_plan() is the right question (empty plan still governs), the 409 is placed after auth and before the stop loop with sensible precedence (401 -> 409 -> 400), the response shape bridges the v1 success key and the v2 code/message convention, and the post-restart window is a documented, deliberate trade-off consistent with the loop's own post-restart silence. Tests pass locally (23/23 in test_allocations_v2.py; the 9 failures in the wider suites are environment issues — /var/cache/aleph permissions and a pydantic model mismatch — in files this PR does not touch). Only non-blocking nits: the 409 path logs nothing, and there is a small inherent TOCTOU window where a plan landing mid-legacy-push still lets that push's stop loop run before the loop reverts to the plan.

**`src/aleph/vm/agent/views/__init__.py`** (line 615): Non-blocking: this 409 path emits no log line, while other operator-relevant transitions in this route log at INFO. A scheduler silently failing with 409s would be easier to diagnose from the node's log — consider `logger.info("Refused a legacy allocation push: this node is governed by a plan")`.

**`src/aleph/vm/agent/views/__init__.py`** (line 615): Worth being aware of (no change suggested): has_plan() is a snapshot, and the handler yields several times after it (request.json(), list_vms()), so a plan submitted while a legacy push is mid-flight still lets that push's stop loop run before the reconciler reverts to the plan. It self-heals on the next pass but a torn-down VM's disks are gone, which is the same failure the PR is closing. Acceptable given how narrow the window is and that the alternative is invasive locking.

foxpatch-aleph (approved):
A clean, well-scoped fix for a real operational hazard: a scheduler mixing v1 pushes with v2 plans on one CRN would have had the push and the reconciler undoing each other indefinitely, with the legacy stop loop additionally retiring plan-kept VMs. The 409 is correctly ordered (after auth, before the stop loop and the starts), has_plan() properly distinguishes 'governed' from 'lists something' so the empty plan still locks the route, and the test double was updated with has_plan=False where a bare MagicMock would have been truthy. All three new tests and the surrounding suites pass in this environment (the only failures are pre-existing, unrelated pydantic/environment issues reproducing at the head commit). Docs are accurate and the post-restart window is honestly documented in three places.

**`src/aleph/vm/agent/views/__init__.py`** (line 615): Non-blocking: this guard is check-then-act. reconciler.submit() does not take allocation_lock, and request.json() plus the lock acquisition sit between this check and the stop loop, so a v2 plan landing inside one in-flight legacy push is still half-honored (and a VM its stop loop stopped would stay stopped, since the reconciler never restarts a stopped VM). The window is one request wide — strictly narrower than the post-restart window has_plan() documents for itself — but the view docstring could own it the same way has_plan() does.

**`tests/supervisor/test_allocations_v2.py`** (line 490): Nice precision patching aleph.vm.agent.views.teardown_vm rather than the reconciler's import — it asserts the route's own stop loop never runs, without accidentally masking a reconciler teardown. Worth keeping in mind for future readers: a reconciler-side teardown would go through reconciler_module.teardown_vm, which this deliberately does not cover.

PR #1233, opened 2026-09-09
fix(daemon): report a guest that died on its own as FAILED

odesenfans:
## Summary
A persistent VM whose QEMU exited on its own had no status that said so. `vm_status` took the lifecycle timestamps plus a plain "is the unit active" flag, so with no stop stamped and the unit no longer active, a VM created on this daemon reported BOOTING for ever and one adopted at boot reported DEFINED for ever. The agent's allocation reconciler counts RUNNING, BOOTING and DEFINED as live (`allocation/plan.py` `LIVE_STATUSES`), so a crashed guest was never torn down and never rebuilt, and the crash-loop backoff of #1214 was unreachable for exactly the failure it was written for. The daemon announced nothing either: every event came from an RPC path, and nothing calls the daemon when a guest dies. This PR gives the status mapping the unit's real state and adds the FAILED arm, and turns the observation into one event.

Stacked on #1227.

## Changes
- `units.rs`: `UnitStateSource` now yields a `UnitLiveness` per unit (`Active`, `Transitional`, `Dead`, `Unknown`) instead of a bare bool. `Transitional` is `activating` and `deactivating`, the two states systemd has a job in flight for; `Dead` is the settled `failed`, `inactive` and "not loaded at all"; an unrecognized ActiveState is `Unknown`, never a guess that a live guest died. `active_states` stays as the coarse view for the callers that only ask "is it up", so adoption and the start gates are untouched.
- `service.rs`: `vm_status(times, running, unit)` gains one arm, between RUNNING and BOOTING: a VM the daemon has seen alive (`started_at_ns` stamped, which happens once the controller is confirmed ready and at adoption for a VM already running), with no stop stamped, whose unit is `Dead`, is FAILED. `Transitional` stays BOOTING and `Unknown` changes nothing. `guest_liveness` suppresses the arm for the three VMs whose unit is down for a reason of their own: an ephemeral program runs no unit, a SEV or SEV-ES VM's controller is held down until its owner uploads the session certificates, and a VM with a reboot in flight is between two lives. SEV-SNP has no session and is judged like any other VM. `is_awaiting_confidential_init` moved out of `vm_info_message` into a named function so both users share it.
- `service.rs`: `GetVm` and `ListVms` fetch the unit state rather than the flag, and hand the resulting status to the event hub. A bus failure degrades to `Unknown`, which still reads as "not running" but never as death. The mutation RPCs pass no observation: they just settled the unit themselves.
- `events.rs`: `EventHub::observe` announces a VM that has gone FAILED since the hub last reported on it, once per transition rather than once per read. The hub remembers the last status it reported for each VM (whether a lifecycle path emitted it or a read observed it), so a deliberate stop is not announced twice and a poll of a VM that is still dead stays quiet. The first status seen only seeds the map: with nothing reported before it there is no transition and no old status to name. `forget` drops the record on delete, so a hash created again is not diffed against its previous life.
- `lifecycle.rs`: `status_snapshot` and the start/reboot emits go through the same liveness lookup, so a pre-mutation snapshot of a crashed VM says FAILED too.

### Second round: the two windows where a read has no lock

Review found a race the new arm opens. `ListVms` and `GetVm` take no per-VM lock, so a poll can land inside a stop or a reboot, and in both windows the unit is down with nothing stamped to say why.

- `lifecycle.rs` `stop_vm_execution` stamped `stopping_at_ns` only after the guest was gone, and the graceful ACPI shutdown it waits on runs for up to 75 seconds. The stamp now goes in before `stop_and_disable` is issued, and is put back if the stop never started, so a unit systemd refused to touch does not leave the VM in STOPPING for ever.
- `lifecycle.rs` `reboot_vm` stamped nothing across `RestartUnit` plus the readiness wait, and a restart takes the unit out of `active` and back in. The entry now carries a `restarting` marker across that window (new `VmEntry` field), set together with a fresh `starting_at_ns` so the VM reports BOOTING, and cleared on the failure paths too, where the VM really is down and should say so. `start_vm_execution` clears it as well, so a reboot that died between setting and clearing cannot leave a VM permanently exempt from the arm.

### Review follow-ups, round 2
- `src/service.rs`: a read that computes FAILED confirms it against a fresh
  entry before returning or announcing. The read path clones the entry and
  then asks systemd with no lock held, so a clone taken just before a reboot
  or stop began carries neither the restarting marker nor the stop
  pre-stamp, and a unit answer that arrives after the job took the unit
  down computed a death for a VM in a deliberate transition. Every mutation
  sets its marker or stamps under the world write lock before it touches
  systemd, so a re-read after the unit was observed dead sees the window.
  One shared helper serves `GetVm` and `ListVms`; a VM deleted meanwhile is
  NOT_FOUND for the former, left out of the listing for the latter, and
  never announced. The lock is taken only on the FAILED arm and only after
  the D-Bus query, never around it. Tests race a read against the reboot
  marker (BOOTING), the stop pre-stamp through ListVms (STOPPING), the
  converse with no window (FAILED, announced once), and a delete.
- `src/lifecycle.rs`: when `stop_and_disable` fails, the stop stamps are
  restored only if the unit did not settle. A disable that fails after the
  stop job already took the unit down leaves the stop recorded, so a
  deliberately stopped VM does not read FAILED and announce a death. One
  extra `get_active_state` read, on the error path only. Tests cover the
  refused-disable and refused-stop halves.

### Review follow-ups, round 3
- `src/lifecycle.rs`, `src/units.rs`: the stop path no longer infers
  whether the stop happened from a point-in-time ActiveState, which cannot
  tell a refused stop from a queued one (StopUnit returns as soon as the
  job is accepted, and the unit then sits active or deactivating for the
  guest's shutdown). `stop_and_disable` returns a two-variant error that
  names the step that failed: a refused stop restores the pre-stop stamps,
  a failed disable after an accepted stop keeps the stop recorded. The
  fake systemd now models a queued stop by leaving the unit deactivating,
  so the test that only passed because the fake stop was synchronous is
  gone; both halves are covered.
- `src/world.rs`: the `restarting` marker's doc says what sets and clears
  it now (RebootVm and StartVm, both stamping a fresh `starting_at_ns`).
- `docs/architecture/divergences.md`: entry for `VmInfo.starting_at_ns`
  reporting the most recent start or reboot rather than the creation-time
  start, with the reason (the BOOTING mapping needs the fresh value).
- Test: the no-op `drop(probe.probing)` and the field only it read are
  removed.

## Test plan
- `service.rs`: `a_running_adopted_vm_whose_unit_died_reports_failed` (was `..._reports_defined`), `a_started_vm_whose_unit_died_reports_failed_instead_of_booting_for_ever`, and `a_list_that_finds_a_dead_unit_reports_failed_and_announces_it_once` all fail on the base commit (DEFINED and BOOTING where FAILED is expected) and pass with the fix.
- `lifecycle.rs`: `a_poll_during_a_stop_sees_stopping_not_a_death` and `a_poll_during_a_reboot_sees_booting_not_a_death` drive a fake systemd that pauses where the real one leaves a job in flight and runs the read path (status plus `observe`) from inside the pause. Both report `[Failed]` without the second round's fix and `[Stopping]` / `[Booting]` with it, and neither announces a death. `a_unit_that_fails_after_a_reboot_still_reports_failed` pins that the exemption is the window, not the VM.
- `service.rs` regression guards: `an_explicitly_stopped_vm_stays_stopped_under_a_dead_unit`, `a_confidential_vm_awaiting_its_session_is_never_reported_failed`, `an_ephemeral_program_ignores_the_unit_state`, plus the transitional and unanswered-bus cases inside the two mapping tests.
- `lifecycle.rs`: the SEV-ES create test now asserts that its deliberately dead unit still reports BOOTING.
- `events.rs`: `an_observed_death_is_announced_exactly_once` and `a_read_is_not_an_event_source`.
- Checks: `cargo fmt --check` (0), `cargo clippy --locked --all-targets -- -D warnings` (0), `cargo test --locked -p supervisor-daemon` (403 lib + 4 integration, 0 failed), `cargo test --locked --workspace` (0 failures). Python untouched; `tests/supervisor/test_proto_convert.py` and `tests/supervisor/test_allocation_reconciler.py` run green (46 passed) as evidence the FAILED mapping is already wired.

## Follow-ups, not in this PR
- The SEV / SEV-ES exemption never re-engages: `awaiting_confidential_init` cannot tell "the owner has not uploaded a session yet" from "the guest died after it did", so a dead post-session SEV guest still reports BOOTING. Closing it needs a daemon-side record that `InitializeConfidential` ran and a matching agent-side change, since `start_persistent_vm` leaves an awaiting VM untouched by design. SEV-SNP, the 2.1 confidential path, is unaffected.
- `EventHub::forget` is called only from `delete_tracked_vm`. Other paths that retire a VM without going through it leave one small record behind, which costs a stale `old_status` on the next life of the same hash rather than a wrong status.

### Third round: the start window was open too

Review (CHANGES_REQUESTED) found the same race on the start path, and a wider one. `start_vm_execution` cleared `stopping_at_ns` and `stopped_at_ns` at its very first step and then spent the tap creation, the nftables rules, the SNP DHCP server, `RestartUnit` and the readiness wait with the unit down, while `started_at_ns` still held its pre-stop value. That is the dead-unit arm's exact shape, so a concurrent `GetVm` or `ListVms` reported FAILED where it used to report BOOTING, and `EventHub::observe` announced a spurious STOPPED to FAILED death that wakes the agent's reconciler against a VM that is coming up as asked.

- `lifecycle.rs`: `start_vm_execution` now sets the same `restarting` marker `reboot_vm` uses, with a fresh `starting_at_ns` so the VM reports BOOTING. Its body moved into `start_vm_execution_marked` so the marker is cleared exactly once on the way out, success or failure at any step. A start that failed leaves a VM that really is down, and it reports FAILED.
- `lifecycle.rs`: the already-active short circuit in `start_vm` returns before any of that, so a marker left by a reboot that died mid-window survives it. The comment there now says why that is harmless: the marker only suppresses the dead-unit arm, and a VM whose unit that branch just found active reports RUNNING regardless.

New tests: `a_poll_during_a_start_sees_booting_not_a_death` (the same mid-job fake systemd, driven through stop then start) and `a_start_that_fails_before_the_unit_still_reports_failed` (an injected tap-create failure, the error path furthest from the exit).

Not taken: `EventHub::forget` on the other retirement paths, already recorded as a follow-up below.

## Rebased onto dev-2.1 (#1199 + #1200)

Both commits replayed with no conflict; #1200 touched `service.rs` and `lifecycle.rs` in regions this PR does not.

Checks re-run after the rebase: `cargo fmt --check` exit 0, `cargo clippy --locked --all-targets -- -D warnings` exit 0. Test suites left to CI this round (some daemon tests need root to unwind DHCP servers).


foxpatch-aleph (approved):
A well-executed fix for a real gap: a guest whose QEMU exited on its own was reported BOOTING or DEFINED forever, which the agent's allocation reconciler counts as live, so crashed guests were never rebuilt. The new UnitLiveness enum correctly distinguishes activating/dead/unanswered-bus, the FAILED arm in vm_status is correctly gated on started_at_ns (which I verified is only stamped after the controller is confirmed ready, at adoption for running VMs, or for a confidential VM awaiting its session), and deliberate stops are protected by the stopping/stopped short-circuits. The EventHub observe/record/emit design gives exactly-once death announcements with no double-announce races under the mutex, and forget on the tracked-delete path is the only removal site that needs it. Agent-side wiring (proto mapping, LIVE_STATUSES, TEARDOWN_STATUSES, event and gap-reconcile handling) checks out. Remaining findings are minor, documented, and non-blocking: a narrow stop-window race that can announce a spurious FAILED for a deliberate stop (benign since the agent treats FAILED and STOPPED identically and the reconciler is list-driven), and the SEV/SEV-ES blindness that persists for guests that die after session upload. Test coverage of both the new behavior and its regression guards is thorough.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 816): Non-blocking: the unit is stopped before stopping_at_ns is stamped, so a GetVm/ListVms racing a deliberate stop (or the RestartUnit inside a reboot, where the unit passes through deactivating) can see started_at set, no stop stamps, and a Dead unit — and announce a spurious RUNNING→FAILED just before the path's own RUNNING→STOPPED. Impact is benign (the agent's handling of FAILED and STOPPED is identical, and the reconciler converges from the list, not from events), but stamping stopping_at_ns before issuing the stop would close the window entirely.

**`rust/crates/supervisor-daemon/src/service.rs`** (line 613): The awaiting_confidential_init gate uses !running, which conflates 'never started' with 'was running and died': a SEV/SEV-ES guest that dies after its session was uploaded and the controller ran is still never reported FAILED, i.e. the exact bug this PR fixes persists for that VM family. You document this honestly, so it is not blocking — but a stamp set when the session is uploaded (or in initialize_confidential before enable_and_start) would let the FAILED arm cover a running confidential guest that later dies. Worth a follow-up issue.

**`rust/crates/supervisor-daemon/src/service.rs`** (line 614): Nit: guest_liveness passes unit.is_active() to awaiting_confidential_init while vm_info_message uses the caller's running for the wire field. They are identical for every current caller (GetVm/ListVms derive running from the unit; mutations pass Unknown, which short-circuits), but threading the caller's running through would remove the subtlety.

foxpatch-aleph (changes requested):
The core mechanism is correct and well-built: the FAILED arm in vm_status, the UnitLiveness tri-state (bus failure degrades to Unknown, never Dead), EventHub::observe's exactly-once semantics (the last_status Mutex serializes record-and-check, so concurrent polls cannot double-announce), the stopping_at_ns pre-stamp in stop_vm_execution, and the restarting marker in reboot_vm together close the stop and reboot no-lock windows, and mutated_vm_info passing UnitLiveness::Unknown keeps mutation responses backwards compatible. Agent-side compatibility checks out (FAILED is not in LIVE_STATUSES and the FAILED event path already drops state and wakes the reconciler). However, the start window was missed: start_vm_execution clears the stop stamps without setting any guard while started_at_ns retains its pre-stop value (stop never clears it) and the unit is Dead for the whole window -- tap/nft/DHCP setup, units.restart, and wait_for_controller_ready, the last of which waits for the guest to boot -- so a concurrent poll (ListVms takes no per-VM lock) now computes FAILED where it previously computed BOOTING, and observe announces a spurious STOPPED-to-FAILED death event that wakes the reconciler to converge on a VM that is mid-start. This is the exact race class the PR's second round fixed for stop and reboot, newly introduced on the start path by the FAILED arm; set the same restarting marker around the window (cleared on success and error paths) as reboot_vm does. No Rust toolchain was available in this environment, so the test suite could not be executed; the review is by reading only. Minor, non-blocking: the panic-safety-net clear of restarting at line 1083 is skipped when start_vm early-returns on an already-active unit, so a marker leaked by a panicked reboot would persist (panic-only), and EventHub::forget runs only on delete_tracked_vm, leaving stale last_status records on other retirement paths -- already documented as a follow-up and harmless while observe only fires on death.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 1077): The start window is unguarded. start_vm_execution clears stopping_at_ns and stopped_at_ns here without setting any marker, while started_at_ns retains its pre-stop value (stop_vm_execution never clears it) and the unit is Dead for the entire rest of the window -- tap/nft/DHCP setup, units.restart, and wait_for_controller_ready, the last of which waits for the guest to boot. A poll landing in that window (ListVms/GetVm take no per-VM lock) computes: stopped_at_ns == 0, stopping_at_ns == 0, running == false, started_at_ns != 0, unit == Dead -> FAILED, where this same window previously reported BOOTING; EventHub::observe then announces a spurious STOPPED -> FAILED death for a VM that is mid-start, waking the agent's reconciler to converge on it (and the read itself reports it not-live). guest_liveness only exempts this for SEV/SEV-ES VMs, via the accidental cover of awaiting_confidential_init. This is exactly the race the second round fixed for stop and reboot; please set the same restarting marker reboot_vm uses (or equivalent) around the window, cleared on both success and error paths.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 1166): Minor: this early return skips the safety-net clear of restarting at line 1083, so a restarting marker leaked by a reboot that panicked between set and clear would persist on an already-active VM. Panic-only, so fine to leave, but worth a comment noting the safety net does not cover the already-active short circuit.

**`rust/crates/supervisor-daemon/src/events.rs`** (line 90): Non-blocking: forget is only called from delete_tracked_vm, so a hash that retires through any other path keeps its last_status record. Harmless today since observe only fires on death and the first-seen seed absorbs the staleness, and the author has documented broader coverage as a follow-up -- just confirming that follow-up is worth doing before observe grows more transitions.

foxpatch-aleph (approved):
This is a well-engineered fix for a real gap (a spontaneously-dead guest reporting BOOTING/DEFINED forever, invisible to the agent's reconciler). The FAILED arm is guarded conservatively: it requires a positively observed dead unit under a VM the daemon has seen alive, bus failures degrade to Unknown rather than Dead, and unrecognized systemd states map to Unknown. All four windows where the unit is transiently down (stop, start, reboot, create) are correctly marked, and the mid-job fake-systemd tests exercise the exact interleavings a concurrent ListVms would hit. EventHub::observe announces a death exactly once per transition and the emit-before-early-return ordering correctly prevents a post-stop read from double-announcing. I traced every call path (create, adopt, readopt, InitializeConfidential, delete, all RPC responses) and found no unmarked window or status/running mismatch. Remaining nits are minor and mostly already acknowledged as follow-ups in the PR description; none block merging.

**`rust/crates/supervisor-daemon/src/events.rs`** (line 90): Minor, already acknowledged as a follow-up: `forget` is only called from `delete_tracked_vm`, so a VM retired outside it (a failed create after a ListVms observed it, a hidden-VM discard) leaves a stale `last_status` record. The cost is bounded — a wrong `old_status` (e.g. Stopped->Failed instead of Defined->Failed) if the same hash is observed Failed in its next life — but it may be cheap to also call `forget` on the create-failure teardown path while you're here.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 1090): The `restarting` marker can only leak across the set/clear window via a caught panic inside `restart`/`wait_for_controller_ready` (a process exit resets in-memory state anyway), and the `start_vm` short-circuit comment at line 1191 correctly explains why a leaked marker is benign (it only suppresses the dead-unit arm, and that branch requires the unit active). Agreed with the documented reasoning — just noting the marker is also not cleared by `stop_vm_execution`, so a leaked marker would persist through a stop/start cycle until a successful start or reboot clears it.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 1091): Behavioral note: clearing `stopping_at_ns`/`stopped_at_ns` here means a deliberately-stopped VM whose start then fails permanently loses its STOPPED status and reports FAILED instead. That is the PR's stated intent ("a start that fails really is a VM that is down"), and neither STOPPED nor FAILED is in the agent's LIVE_STATUSES, so reconciler behavior is unaffected — just flagging the semantic loss of "was deliberately stopped" fidelity.

**`rust/crates/supervisor-daemon/src/events.rs`** (line 78): Cosmetic: during a systemd bus outage a read records BOOTING/DEFINED for a running VM (Unknown unit, not running), so a genuine death announced right after reports old_status=BOOTING or DEFINED rather than RUNNING. Also, in the narrow window where a lifecycle `emit` races a read that already computed FAILED, a death can be announced twice. Both are benign — the agent is idempotent on teardown — and neither justifies the per-VM lock the PR deliberately avoids.

foxpatch-aleph (approved):
This PR is a careful, well-reasoned fix: the UnitLiveness taxonomy is exhaustive and fails closed (unrecognized ActiveState and bus errors both map to Unknown, never Dead), the dead-unit FAILED arm is guarded by stop stamps on every deliberate-teardown path, and the three unlocked windows (stop, reboot, start) are each closed with the right marker and cleared on all exits. EventHub::observe's once-per-transition logic is race-free under its mutex, and the agent-side contract (LIVE_STATUSES excluding FAILED, event handler reacting to STOPPED/FAILED) matches the daemon's new behavior. The mid-job fake systemd tests genuinely reproduce the races they pin. Remaining issues are documentation-level: the restarting marker's doc comments still describe only the RebootVm window even though StartVm now sets it too, and EventHub::forget is only wired to the delete path (acknowledged follow-up).

**`rust/crates/supervisor-daemon/src/world.rs`** (line 257): This doc says the marker is set while "RebootVm has a systemd restart job in flight", but after the third round `start_vm_execution` sets it too (as does its `guest_liveness` counterpart in service.rs, which still speaks only of "a VM in the middle of a reboot"). Worth updating both to name the start window, since that is the longer of the two.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 1226): Nit: "The next stop, start or reboot clears it" — `stop_vm_execution` never touches `restarting`; only a start or reboot clears the marker. The claim is harmless (a stopped VM reports STOPPED on the stop stamps regardless), but the comment overstates the recovery paths.

**`rust/crates/supervisor-daemon/src/events.rs`** (line 90): `forget` is only called from `delete_tracked_vm`, but several other `entries.remove` paths (the create-failure cleanups in `create_vm_inner`, quiesce.rs) retire a VM without it. Beyond the acknowledged map growth, note one subtle behavioral edge: a stale `Failed` record would suppress the once-per-life death announcement for a recreated same-hash VM (observe sees previous == Failed). Worth filing the acknowledged follow-up so it is not lost.

foxpatch-aleph (approved):
This PR is a careful, well-documented fix for a real gap (a guest that dies on its own is never noticed), and the three review rounds have closed the stop, reboot and start windows convincingly. The UnitLiveness four-state mapping is sound (unrecognized ActiveStates map to Unknown, bus failures degrade to Unknown rather than Dead, ListUnits misses map to Dead), the FAILED arm's preconditions (started_at stamped, no stop recorded, unit settled Dead) are enforced by ordering in vm_status and by the guest_liveness exemptions, and EventHub::observe's record-then-diff makes the death announce exactly once per transition without double-announcing deliberate stops. The VmEntry.restarting marker is always cleared on the success and error paths of both start_vm_execution and reboot_vm, is in-memory only (VmEntry carries no serde), and start_vm's already-active short circuit is correctly analyzed as harmless. The MidJobProbe tests genuinely exercise the mid-job windows from inside the fake systemd, and the agent-side handling of FAILED (LIVE_STATUSES excludes it; supervisor.py and the reconciler's TEARDOWN_STATUSES include it) confirms the event is consumable. I could not run cargo in this environment, but all changed call sites were traced by hand and are consistent.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 839): Minor: stop_and_disable is probe + stop + disable, so if units.stop() succeeds but units.disable() fails, this restore puts stopping_at_ns back to 0 even though a stop job is genuinely in flight. In that window a poll can hit the dead-unit arm and announce a FAILED for a VM whose stop is actually underway. Extremely narrow (disable failing after is_enabled returned true) and arguably an acceptable degradation since the StopVm RPC errored, but the restore could be scoped to the case where the stop was never issued (e.g. move the restore into stop_and_disable's caller only for the pre-stop failure, or have stop_and_disable report how far it got).

**`rust/crates/supervisor-daemon/src/service.rs`** (line 1185): Theoretical TOCTOU, non-blocking: the statuses are computed at time T and observed at T+delta with no per-VM lock in between. If a VM that read FAILED is started again by a concurrent StartVm whose emit records Running inside that gap, this observe(Failed) will see previous=Running and announce a spurious Running→FAILED death against a healthy, just-started VM (and leave the map saying Failed until the next poll corrects it). The window is sub-millisecond and a start must complete inside it, so the probability is tiny, but since the consequence is the agent tearing down a healthy VM, it may be worth a note in the observe() docstring acknowledging the read is not atomic with the lifecycle paths.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 6707): Nit: drop(probe.probing) is a no-op — probe.state.units still holds an Arc to the same MidJobProbe, so nothing is actually dropped. It reads like it has an effect; either drop it from the test or replace with a comment saying the probe simply is not invoked again after this point.

**`rust/crates/supervisor-daemon/src/events.rs`** (line 32): Confirmed the documented follow-up: last_status entries are only removed by delete_tracked_vm (via forget), so entries retired through other paths (e.g. the program-recreate branch of reboot_vm, which removes the entry and re-creates under the same vm_id, is fine, but any path that removes an entry without delete_tracked_vm) leave one small record per distinct hash. Cost is only a stale previous status on the next life of the same hash, as the PR notes — agreed it is acceptable to defer.

foxpatch-aleph (approved):
A well-designed and carefully tested change. The FAILED arm is placed correctly in vm_status (stop stamps and running short-circuit first), the UnitLiveness taxonomy cleanly separates "dead" from "unanswered", and all three mutation windows (stop, reboot, start) are marked and unmarked on every exit path, including error paths, which the _marked body split guarantees. The event hub's record/seed/forget semantics are coherent with the lifecycle emits, and the mid-job test harness is a convincing reproduction of the races the second and third rounds fixed. I found no correctness or security issues; the findings below are non-blocking observations.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 1226): "The next stop, start or reboot clears it" overstates: stop_vm_execution only touches stopping_at/stopped_at, never entry.restarting, so a stale marker survives a stop (it is harmless because the stop stamps short-circuit vm_status before the dead-unit arm, and the next start/reboot does clear it, but the comment should say only a start or reboot clears it).

**`rust/crates/supervisor-daemon/src/service.rs`** (line 1126): Residual TOCTOU, worth being aware of: the entry is cloned (with restarting=false / stopping_at=0) and the unit is queried afterwards, so if a reboot or stop starts inside that gap, the poll can still pair a pre-marker snapshot with a unit that systemd has already taken down, report a one-off FAILED, and announce a death against a healthy VM — the exact false positive this PR works to prevent. A cheap mitigation would be to re-check the live entry's marker/stamps before believing a Dead unit (or before calling events.observe). The window is small and the design accepts lock-free reads, so this is an observation, not a blocker.

**`rust/crates/supervisor-daemon/src/events.rs`** (line 83): observe() is not atomic: two concurrent polls that both read previous != Failed can both emit, so a death can be announced twice. The duplicate is harmless to the agent (its teardown/rebuild is idempotent on FAILED), but holding the record/compare/emit under one lock (or a check-and-set on last_status) would preserve the once-per-transition property under concurrency. Nit: emit() then records the same status a second time.

**`rust/crates/supervisor-daemon/src/units.rs`** (line 73): from_active_state's Unknown arm — the safety net that keeps a future systemd state from declaring a live guest dead — has no direct test; only active/deactivating/inactive/failed are exercised, and only indirectly through the other suites. One small table test would pin it.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 839): Note on the restore: if stop_and_disable fails after the stop job was accepted (e.g. the disable step errors), stopping_at is restored to 0 while the unit is now genuinely going down, so the next read reports FAILED rather than STOPPING. That is defensible (the VM really is down with nothing stamped), just flagging that the restore assumes the whole call is all-or-nothing.

foxpatch-aleph (approved):
Solid, carefully-reasoned change. The UnitLiveness abstraction keeps the new FAILED status honest (only a positively-observed dead unit under a VM the daemon has seen alive can read as death; bus failures and unrecognized systemd states degrade to Unknown), the three race windows (stop, reboot, start) are closed coherently with stamps and the restarting marker plus rollback, and EventHub::observe's record-then-compare design makes the once-per-transition guarantee atomic under concurrent polls. The test plan targets exactly the races fixed, with a clever mid-job fake systemd harness. Findings below are documentation corners and narrow residual race notes, none blocking.

**`rust/crates/supervisor-daemon/src/world.rs`** (line 257): The doc comment says 'Set while RebootVm has a systemd restart job in flight', but the third round of this PR also has `start_vm_execution` set the marker across its window (lifecycle.rs:1121), which is arguably the longer window. Worth updating the doc so a future reader doesn't conclude only the reboot path touches it. The same reboot-only phrasing appears in the `guest_liveness` doc in service.rs ('a VM in the middle of a reboot has a restart job in flight').

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 1227): 'The next stop, start or reboot clears it' overstates: `stop_vm_execution` never touches `entry.restarting`, so a stop does not clear a stale marker. It's harmless today because a stopped VM reports STOPPING/STOPPED via the stop stamps regardless, and a subsequent start (unit not active) or reboot does clear it — but if the marker is ever set for a new reason, this comment could mislead. Consider rewording to 'the next start or reboot clears it'.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 839): Narrow corner in the rollback: `stop_and_disable` can fail on the disable step *after* the stop job was successfully queued (stop_and_disable, units.rs:157). In that case the stamp is rolled back even though a real stop is now in flight; once the unit settles the VM reads started_at + Dead and gets announced as a death for what was a deliberate (if erroring) stop. The RPC also errors, so the caller knows the stop was attempted. Probably fine to leave, but worth a sentence in the comment if you agree that's the intent.

**`rust/crates/supervisor-daemon/src/world.rs`** (line 730): Residual hole worth a follow-up note alongside the SEV one: a VM adopted during a bus outage (the `None` arm) never gets `started_at_ns` stamped, so if it later dies its dead unit reports DEFINED, not FAILED — the agent holds it live, which is the exact bug this PR fixes, resurfaced through a different door. Narrow (requires a bus outage at boot, then a later spontaneous death with no intervening RPC), but the same class as the one being fixed.

**`rust/crates/supervisor-daemon/src/events.rs`** (line 32): Note for the record: `last_status` only ever shrinks via `forget` on `delete_tracked_vm`; the ephemeral-reboot path (lifecycle.rs:1264) and the failed-readopt rollback (lifecycle.rs:2729) also remove entries without forgetting, leaving one small stale record per retired hash. The worst case is a stale `previous` on the next life of the same hash (and since `observe` only fires on FAILED, that mostly means a missed announcement, never a spurious one). Already documented as a follow-up in the PR description; just confirming the assessment is right.

foxpatch-aleph (changes requested):
The design is sound and unusually well-documented: the UnitLiveness taxonomy with Unknown never reading as death, the once-per-transition announce built on insert-returns-previous, the stop pre-stamp, and the restarting marker across the reboot/start windows are all correct, and the exemption set (programs, SEV/SEV-ES awaiting session, restart windows) is right. One residual race remains in the new mechanism: get_vm/list_vms clone the entry and only then query the unit, so a snapshot taken before a mutation window is not covered by the window's marker or pre-stamp. A read that clones before a reboot begins (restarting=false, started_at set, no stop stamps) and whose unit query is answered while the restart job has the unit down computes FAILED from the stale clone and announces a spurious death, which the agent acts on by rebuilding a healthy rebooting VM. Since FAILED is the only status that announces and triggers rebuilds, re-validating the entry from the world when the dead-unit arm fires would close this cheaply. A minor related gap: the stop path restores stopping_at_ns even when stop_and_disable failed after the stop job was issued.

**`rust/crates/supervisor-daemon/src/service.rs`** (line 1129): The dead-unit arm can fire on a stale snapshot here (and in list_vms at line 1181). The read path clones the entry, then queries the unit with no lock held, so the `restarting` marker and the stop pre-stamp only protect reads whose snapshot was taken *inside* the window. Consider a read that clones the entry just before a reboot starts (restarting=false, started_at_ns set, no stop stamps), and whose unit query is answered only after the restart job has taken the unit down (the clone-to-query gap is normally ~1 D-Bus round trip, but spawn_blocking queueing behind lifecycle threads or a busy bus can stretch it well past the reboot's marker-set point). The status is computed from the pre-reboot clone: started_at != 0, unit Dead -> FAILED, and observe announces a spurious death for a VM that is rebooting as asked. On the agent side, watch_supervisor_events turns that into notify_vm_down(vm_id, FAILED), which makes the reconciler converge immediately and rebuild -- precisely the harm this PR exists to prevent; the same stale status in the ListVms response flips the reconciler's FAILED-vs-STOPPED decision for a deliberately stopped VM. Suggested fix: when the computed status is FAILED, re-read the entry under the world lock and recompute before observing/returning -- the mutations always set their marker/stamps under the write lock *before* touching systemd, so a fresh re-read after the unit observation sees the window; a cheap guard on the only arm that announces.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 839): Minor: this restore also fires when stop_and_disable failed *after* the stop job was issued -- the disable step can error with the unit already stopped. In that case the stamp is put back even though the unit is genuinely down with no stop recorded, so the VM reads FAILED and gets a death announcement even though it was deliberately stopped (a retry of StopVm would settle it, but the interim announcement wakes the reconciler). Consider keeping the stamp when get_active_state reported the unit already settled, or documenting that this failure mode is accepted.

foxpatch-aleph (changes requested):
Well-designed and exceptionally well-tested PR: the UnitLiveness taxonomy, the FAILED arm with its program/SEV-awaiting/restart-window exemptions, the once-per-transition EventHub::observe, and the confirm-on-fresh-entry read path are all correct, and the race analysis across the stop/start/reboot windows is sound. One blocking issue in the newest commit: the stop_and_disable error path decides whether the stop happened by re-probing the ActiveState, but on production systemd StopUnit returns as soon as the job is queued, so when disable is the failing step the probe reads active/deactivating/unknown, the pre-stop stopping_at_ns gets restored, and the queued job then settles the unit dead with no stop recorded — the VM reads FAILED and a death is announced for a deliberate stop, the exact outcome that commit exists to prevent. The unit test passes only because FakeSystemd::stop is synchronous. Propagating which step failed (or calling stop and disable separately in stop_vm_execution) would make the decision exact. Remaining comments are minor: a stale doc comment on the restarting field, a wire-visible starting_at_ns semantic change, and a no-op drop in one test.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 848): The ActiveState re-probe cannot distinguish "stop refused" from "stop queued and in flight". ZbusUnitStates::stop queues StopUnit and returns as soon as the job is accepted (see unit_job), and the unit then sits in active/deactivating for the guest's ACPI shutdown (up to 75s per wait_for_controller_stopped). So when disable is the failing step, the probe here reads active, deactivating, or unknown (bus hiccup) in nearly every case, the pre-stop stopping_at_ns is restored, and the queued stop job proceeds to settle the unit dead with no stop recorded — the VM then reads FAILED and EventHub announces a death for a VM the operator deliberately stopped. That is the exact outcome this commit exists to prevent, and a_stop_that_fails_with_the_unit_down_keeps_the_stop_recorded only passes because FakeSystemd::stop is synchronous (the unit is inactive the moment stop returns). The stop-refused half of the guard is fine (no job was queued, the unit is genuinely still up); only the disable-failed half is wrong. Since stop_and_disable knows which step failed, consider returning a structured error from it (or calling stop and disable directly here) instead of inferring it from a point-in-time ActiveState.

**`rust/crates/supervisor-daemon/src/world.rs`** (line 257): This doc comment is stale after the third-round change: start_vm_execution sets the same marker for the start window (not just RebootVm), and both paths stamp a fresh starting_at_ns, so "a reboot stamps neither a stop nor a fresh start until the controller is back" no longer holds. Worth updating so the field's contract matches what now sets and clears it.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 1134): Note this re-stamp (and the one in reboot_vm at line 1328) is wire-visible: VmInfo.starting_at_ns now reports the most recent start/reboot rather than the original creation-time start, which the base code preserved. The fresh value is needed for the BOOTING mapping, but if this is a deliberate divergence from the Python timestamp semantics it may deserve an entry in the divergences ledger mentioned in units.rs.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 6719): drop(probe.probing) is a functional no-op: state still holds another Arc to the MidJobProbe, and the OnceLock closure stays registered, so nothing is disarmed. If the intent was to guarantee the probe cannot fire for the rest of the test, that is not achieved; otherwise the line can simply be removed.

foxpatch-aleph (approved):
This PR is a well-reasoned, carefully-raced fix for a real bug: a guest whose QEMU exited on its own was reported BOOTING/DEFINED for ever and never rebuilt by the agent's reconciler. The design is sound — UnitLiveness distinguishes transitional/dead/unknown so a bus failure or in-flight job can never read as death; the FAILED arm sits correctly after the stop/stopped/running short-circuits; the deliberate windows (stop pre-stamp, reboot/start restarting marker with fresh starting_at_ns) are all set under the world write lock before systemd is touched, which is precisely what the read path's re-confirmation in observed_vm_info relies on; the StopAndDisableError two-variant restore correctly handles the queued-stop case that a point-in-time ActiveState cannot; and EventHub::observe's seed/dedup/forget logic produces exactly one event per transition. Test coverage is exceptional, including mid-job fake systemd probes and direct race reproductions for the reboot, stop, start, and delete windows. Three prior review rounds already addressed the races I would have raised. The remaining nits below are non-blocking.

**`rust/crates/supervisor-daemon/src/units.rs`** (line 188): Non-blocking observability nit: both `StopAndDisableError` variants use `#[error(transparent)]`, so when the error is converted via `From<StopAndDisableError> for RpcError` (`error.to_string()`), the RPC error the caller sees does not say *which* step failed. Since the whole point of the two-variant enum is that only the failing step can distinguish a refused stop from a failed disable, consider `#[error("stop failed: {0}")]` and `#[error("disable failed: {0}")]` so operators reading the gRPC error aren't left with the same ambiguity the type exists to resolve.

**`rust/crates/supervisor-daemon/src/service.rs`** (line 435): Residual (accepted) TOCTOU, worth stating in the doc comment: the re-read closes the window only while a mutation is *in flight*. If a mutation begins and *fully completes* between the read's unit query and its re-read, the fresh entry shows no marker and no stop stamps while the carried-over `Dead` observation produces a false FAILED for one poll (and one announcement). The window requires an entire mutation — including the controller readiness wait — to fit inside a D-Bus round trip plus one lock acquisition, so it is negligible in practice, and re-querying systemd under the lock would be worse. Fine to leave as is; just noting the design boundary.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 1473): Confirmed `events.forget` is wired only into `delete_tracked_vm`; other retirement paths leave one stale `last_status` record, costing a stale `old_status` on the next life of the same hash. This is already recorded as a follow-up in the PR description — just confirming that assessment is accurate.

PR #1234, opened 2026-09-09
fix(agent): verify a plan's entries in a worker thread, not on the loop

odesenfans:
## Summary
An allocation push is capped at 8 MiB of body, which is thousands of entries, and `build_plan` judged every one of them on the event loop: a pydantic parse and an ecrecover each. Measured locally, 300 signed entries cost 1.3 seconds of straight-line work, and for that time the agent answers nothing else: not a status request, not a supervisor callback, not its own convergence pass. The per-entry work now runs in a worker thread, 32 entries to a hop, while the verdict (which reads the shared capacity view) stays on the loop exactly where it was.

Stacked on #1232.

## Changes
- `allocation/verdict.py`: the per-entry half of `build_plan` is now `judge_entry` / `judge_entries`, returning a `JudgedEntry` derived from the entry alone (no shared state read or written), and `assemble_plan` folds those into the plan in arrival order. `build_plan` is a coroutine that dispatches the batches through `asyncio.to_thread`. What it hands back is unchanged, refused hashes included: the fold is the same code, in the same order, so the rule that a hash the push named is never torn down by that push holds as before.
- The awaits `build_plan` adds land before the handler reads anything the answer is computed from, so the window in which nothing may yield (supervisor list, GPU read, `compute_verdict`, `submit`) is untouched. Both module docstrings now say which half awaits and why the other must not.
- `allocation/verify.py`: `VerifiedMessage` no longer carries a deep copy of its own message. It existed for a caller that resolves `use_latest` refs in place, and there is no such caller: `create_vm_execution` loads the message from the CCN again through `load_updated_message`. The docstrings that described verification as saving that fetch are corrected: what it buys is the immediate answer, an unverifiable entry is answered `pending` and sized a pass later.
- `views/allocations_v2.py`: `_read_plan` awaits `build_plan`.

## Test plan
- `tests/supervisor/test_allocation_verdict.py::test_a_large_plan_is_verified_without_holding_the_event_loop`: 300 genuinely signed entries, with a concurrent task that must get its turns before the plan is built. Watched fail on the old code (the task had not started when the answer was ready), passes now.
- `tests/supervisor/test_allocation_verdict.py::test_the_entries_reach_the_thread_in_bounded_batches`: 70 entries reach the thread as 32, 32, 6, so one push cannot hold a worker for the length of the plan.
- `tests/supervisor/test_allocation_verify.py`: the deep-copy test is gone with the field; the remaining verification cases are untouched.
- Checks: `pytest` over the allocation, capacity, registry and v2 files, 278 passed. `ruff format --diff`, `isort --check-only --profile black` on the five changed files, both exit 0. `mypy src/aleph/vm/ tests/ examples/example_fastapi runtimes/aleph-debian-12-python` (the CI invocation), `Success: no issues found in 211 source files`.


foxpatch-aleph (approved):
This PR is correct and well-executed. The per-entry verification work from build_plan is moved into a worker thread in bounded 32-entry batches, while the plan assembly (which is ordering-sensitive) stays on the loop; I verified the fold in assemble_plan reproduces the old inline loop exactly, including both duplicate-hash orderings, the refusal/refused-set semantics, and the unusable-hash path. The await-free window between the supervisor read and submit() is untouched, and build_plan's new awaits land strictly before any shared state is read. The removed VerifiedMessage.original field has no remaining consumers — the reconciler starts VMs by hash only and re-fetches messages itself, so the deep copy was dead weight, as the docstring claims. judge_entry is genuinely pure (parse + ecrecover, no shared state), so the thread hop is safe. I ran the three relevant test files (90 tests) and all pass. Two non-blocking nits below.

**`src/aleph/vm/agent/allocation/verdict.py`** (line 89): Minor: `VERIFICATION_BATCH_SIZE` is defined mid-module between `compute_plan_id` and the `JudgedEntry` dataclass; it would sit more naturally near the top of the module with the other module-level definitions, or directly above `build_plan` where it is consumed.

**`src/aleph/vm/agent/allocation/verdict.py`** (line 184): Behavioral delta, benign and non-blocking: in the old code a duplicate hash whose predecessor was already rejected was skipped before `verify_entry` ran; now every entry is judged in the thread and only discarded at the fold. Duplicate entries therefore pay a wasted parse/ecrecover before being dropped. No behavior change since `judge_entry` cannot raise, just a small cost note in case duplicates ever become common.

foxpatch-aleph (approved):
This PR is correct and careful. The move of per-entry verification into a worker thread preserves arrival order (sequential batches, extend-in-order), touches no shared state (pydantic parse, ecrecover, constants, thread-safe logging only), and confines its new awaits to _read_plan, well before the no-yield window between the supervisor read and submit() that the handler's invariants depend on. The removed VerifiedMessage.original deep copy is genuinely dead: I verified by grep that planned.verified is only read for .message.content sizing in compute_verdict and capacity_check plus a None-check in the reconciler, and the launch path re-fetches via load_updated_message, making the corrected docstrings accurate. I confirmed at runtime that ItemHash is a validating class, so judge_entry's try/except still rejects garbage hashes. All 68 tests in the changed files pass, the wider 268-test allocation/capacity/registry/v2 suite passes, and the event-loop pin test was stable over five repeated runs. Two non-blocking nits noted in the comments.

**`src/aleph/vm/agent/allocation/verdict.py`** (line 218): Minor: entries are now all verified before dedup, so a push repeating the same large message N times pays a parse plus an ecrecover per copy, where the old inline loop skipped re-verifying a hash it had already refused (`if vm_hash in rejected: continue` ran before verify_entry). Semantics are unchanged and the work stays bounded by the 8 MiB body cap, so this is only a note — but if pathological pushes ever matter, deduplicating by raw hash before batching would recover the old behavior.

**`src/aleph/vm/agent/allocation/verdict.py`** (line 110): Nit: "The body's list of entries, or the ValueError that says it is not one" reads as if the exception is a return value; the function raises on an unreadable body. A phrase like 'raises ValueError otherwise' would be slightly clearer.

foxpatch-aleph (approved):
Correct and well-executed offload of per-entry verification to a worker thread. The split of build_plan into judge_entry/judge_entries (pure, thread-safe, no shared state) and assemble_plan (ordering, same fold in the same order as before) is faithful to the old semantics, including the subtle duplicate-hash rules; the handler's no-yield window between the supervisor read and submit() is preserved since _read_plan is awaited before it. The VerifiedMessage.original deep-copy removal is safe: the launch path re-loads messages from the CCN via load_updated_message and the verified message is now consumed only for sizing. New tests meaningfully pin both the yielding behavior and the batching. All related suites pass (249 tests), ruff/isort/mypy clean.

**`src/aleph/vm/agent/allocation/verdict.py`** (line 156): Minor behavioral delta: verify_entry now also runs for entries the fold discards as duplicates of an already-rejected hash (the old code short-circuited on "vm_hash in rejected" before verifying). The result is discarded either way, so this only adds log lines and a bit of parsing for duplicated entries — fine to leave, but a one-line note in the docstring would help future readers. Non-blocking.

**`tests/supervisor/test_allocation_verdict.py`** (line 802): Hardcoding [32, 32, 6] duplicates VERIFICATION_BATCH_SIZE; if the constant is retuned, this test breaks for a reason unrelated to what it protects. Deriving the expectation from the constant (e.g. building the expected list from VERIFICATION_BATCH_SIZE and 70) would make the test robust to retuning. Non-blocking.

foxpatch-aleph (approved):
This is a well-executed and carefully documented refactor. The per-entry verification half of build_plan is genuinely pure (pydantic parse + ecrecover over a per-request body, no shared state), so moving it to a worker thread via asyncio.to_thread is safe, and assemble_plan reproduces the old fold exactly — same arrival order, same duplicate/refusal semantics, same rejection keys and reasons, including the non-dict-entry path landing under the "None" key. The critical invariant (nothing between the supervisor read and submit() may yield) is preserved: both routes call _read_plan before reading anything the answer is computed from. All build_plan callers were updated, the removed VerifiedMessage.original field has no remaining callers, and the empty-body path still answers as an empty plan with no thread hops. The new tests are meaningful: one pins event-loop responsiveness against 300 genuinely signed entries, the other pins bounded batching. Two minor, non-blocking observations are noted below.

**`src/aleph/vm/agent/allocation/verdict.py`** (line 127): Minor behavior change: in the old code a duplicate hash short-circuited on the `vm_hash in rejected` check before verify_entry ran, whereas now every entry is verified in the thread even if assemble_plan will skip it. The result is identical (the fold skips duplicates before checking outcome), so this is only extra work and extra warning logs for a duplicate-heavy push — fine as-is, just noting the difference.

**`src/aleph/vm/agent/views/allocations_v2.py`** (line 49): Theoretical nit: `except ValueError` around `await build_plan(...)` now also catches ValueErrors raised inside the worker thread and would answer them with the misleading 400 \"Body is not a plan: 'vms' must be a list\". In practice no reachable path in judge_entry can raise (all verification failure modes return strings), and the exposure is the same as before this PR, so this is non-blocking.

foxpatch-aleph (approved):
A careful, behavior-preserving refactor that moves per-entry verification (pydantic parse + ecrecover) off the event loop into batched worker threads. The split into judge_entry/judge_entries (pure, thread-safe) and assemble_plan (ordering, on the loop) is clean, and the fold logic is byte-for-byte equivalent to the old inline loop, including the subtle duplicate-hash and refusal semantics. The awaits are correctly placed outside the no-yield window between the supervisor read and submit(), and both module docstrings now document that contract. Removal of VerifiedMessage.original is safe — I confirmed no caller resolves refs in place on the verified message. The two new tests are well-designed and not flaky (the concurrent task is the only ready task while the loop parks on the to_thread future), and the batch-size test correctly patches the module global. I ran the changed test files (72 passed) plus the wider allocation/capacity/registry/v2 suite (218 passed); all green. Only a minor pre-existing quirk: a non-dict entry and a null item_hash both map to the literal rejected key "None".

**`src/aleph/vm/agent/allocation/verdict.py`** (line 149): Minor, pre-existing and deliberately preserved: a non-dict entry and an entry with `item_hash: null` both produce the rejected key `"None"` (via `str(None)`), so two such entries collide in the `rejected` dict and the second overwrites the first. That matches the old behavior exactly, but now that the mapping lives in `unusable(None)` it might deserve a sentence in the docstring noting the collision is intentional.

foxpatch-aleph (approved):
Faithful refactor: the per-entry fold in assemble_plan reproduces the old build_plan logic exactly (order, duplicate handling, refused-set carrying, plan_id), the new awaits land before the handler reads the supervisor's view so the no-yield window is preserved, and the work moved to the thread is genuinely pure (pydantic parse + ecrecover on entry-local data only). The deep-copy removal is dead-code elimination backed by the fact that the launch path reloads messages via load_updated_message and nothing mutates verified.message in place. All 290 allocation-related tests pass in my environment, and the two new tests genuinely pin the concurrency property (the served event can only set while the loop yields) and the batching. ruff and mypy are clean on the changed files.

**`src/aleph/vm/agent/allocation/verdict.py`** (line 155): Minor behavior delta vs. the old loop: verify_entry now runs for entries whose hash duplicates an already-refused one, where the old `vm_hash in rejected` short-circuit skipped it. Results are identical; a push full of duplicates of one bad hash just pays an extra parse and ecrecover per duplicate (plus verify_entry's own warning logs, in addition to the repeat-entry warning). Not worth blocking, just noting it's the one place the refactor is not strictly work-preserving.

**`tests/supervisor/test_allocation_verdict.py`** (line 776): Nit: if build_plan raises, ticking is left pending and the failure surfaces with a side of "Task was destroyed" noise. A try/finally around the await (or awaiting ticking in a finally) keeps the failure report clean. Trivial, not blocking.

foxpatch-aleph (approved):
This PR is correct and well-executed. The split of build_plan into a thread-safe per-entry half (judge_entry/judge_entries, pure, run via asyncio.to_thread in bounded batches) and an ordering-only fold (assemble_plan) preserves the exact semantics of the old inline loop, including duplicate handling and the refused-set protections, and the new awaits land strictly before the no-yield window between the supervisor read and submit() that the module docstring guards. Removing VerifiedMessage.original is safe: no consumers remain, verified is used only for sizing and pending checks, and the launch path re-loads the message via load_updated_message (run.py:501), so the corrected docstrings are accurate. verify_entry cannot raise for dict entries, so the only behavior delta (duplicates now re-verified in the thread) is harmless redundant work. All 145 tests across the touched and adjacent files pass locally. Two non-blocking nits below.

**`src/aleph/vm/agent/allocation/verdict.py`** (line 184): Minor behavior delta: the old inline loop skipped verify_entry for entries whose hash was already in rejected (the `if vm_hash in rejected: continue` came before verify_entry), whereas judge_entry now verifies every entry in the thread. Harmless since the duplicate's judgement is discarded identically by assemble_plan and verify_entry never raises, but duplicated hashes now pay a redundant parse+ecrecover and can double the warning lines. Not worth changing for the small sizes involved, just noting the delta.

**`tests/supervisor/test_allocation_verdict.py`** (line 799): Nit: hardcoding `batches == [32, 32, 6]` couples the test to VERIFICATION_BATCH_SIZE == 32, so resizing the constant later breaks the assertion even though the property still holds. Deriving the expected list from verdict_module.VERIFICATION_BATCH_SIZE (and the entry count) would keep them in sync; the trailing `max(batches) <= VERIFICATION_BATCH_SIZE` assertion only partially covers this.

foxpatch-aleph (approved):
This is a careful, well-documented refactor that moves the CPU-heavy per-entry verification (pydantic parse + ecrecover) off the event loop and into a worker thread, batched 32 entries at a time, while keeping the verdict computation and the critical no-yield window (supervisor list -> GPU read -> compute_verdict -> submit) untouched. I verified the assemble_plan fold reproduces the old inline loop exactly, including duplicate-hash ordering in both directions, non-dict entries, and unusable-hash keys; that JudgedEntry is genuinely derived from the entry alone so the thread is safe; that VerifiedMessage.original had no remaining consumers and the launch path re-fetches via load_updated_message; and that the batch slicing has no off-by-one. The two new tests are meaningful — the event-loop pin reads the event before awaiting the task, so the await cannot mask a failure — and all allocation-related tests (178), mypy, ruff format, and isort pass locally. Only non-blocking observations: duplicate entries of an already-rejected hash are now re-verified in the thread where the old code skipped them (a minor performance nuance bounded by the 8 MiB body cap), and the sequential batch dispatch is the right choice over gather since it bounds worker occupancy.

**`src/aleph/vm/agent/allocation/verdict.py`** (line 127): Minor, non-blocking: a duplicate entry whose hash was already refused by an earlier entry now gets fully parsed and verified here, where the old code skipped it (the `vm_hash in rejected` check ran before `verify_entry`). Behavior is identical and the extra work is bounded by the 8 MiB body cap, so nothing to change — just noting the trade.

**`src/aleph/vm/agent/allocation/verdict.py`** (line 219): Good call keeping the batches sequential rather than gathering them: a gather would hold several pool workers at once and still give the loop only one checkpoint, while this bounds both. No action needed.

foxpatch-aleph (approved):
This PR correctly moves per-entry verification (pydantic parse + ecrecover) off the event loop and into worker threads in bounded batches, restoring agent responsiveness for large pushes without disturbing the careful no-yield window between the supervisor read and submit(). The split into judge_entry/judge_entries/assemble_plan preserves the old fold's semantics exactly — including duplicate handling, the refused-set rules, plan-id computation, and the 'unusable hash' edge cases (a non-dict entry is keyed under the literal string "None" in both versions). The judge path is genuinely thread-safe: it reads only per-entry data and read-only module constants, and verify_entry is total for dict entries. The VerifiedMessage.original removal is safe — no consumer remains, and the launch path re-fetches via load_updated_message without ever mutating the plan's message. I verified 262 tests across the allocation/capacity/registry/v2 suites pass, ruff and isort are clean, and mypy reports no issues in any changed file. The two new tests are well-designed: the responsiveness test is deterministic (the first to_thread await necessarily yields to the loop) and the batching test's monkeypatch works because build_plan resolves judge_entries from module globals at call time. Only two trivial nits, neither blocking.

**`tests/supervisor/test_allocation_verdict.py`** (line 803): Nit: this `max(batches) <= VERIFICATION_BATCH_SIZE` assertion is redundant with the exact `batches == [32, 32, 6]` match on the previous line. Harmless if it's meant to guard the invariant if the literal list is ever updated, but one of the two could go.

**`src/aleph/vm/agent/allocation/verdict.py`** (line 219): Behavioral nuance worth being aware of (no action needed): because every entry is now judged up front, a repeat of a hash already refused still pays its pydantic parse and ecrecover, where the old inline fold skipped the duplicate via the `vm_hash in rejected` check before calling verify_entry. The outcome is identical since the discarded judgement is dropped in assemble_plan, and verify_entry is total so a duplicate cannot throw — just slightly more work for a duplicate-heavy push.

foxpatch-aleph (approved):
Clean, well-reasoned change. The per-entry verification split into judge_entry/judge_entries is genuinely free of shared state, batch ordering is preserved through the sequential to_thread awaits so assemble_plan's duplicate/refusal fold is bit-for-bit the old behavior, and both call sites await build_plan strictly before the supervisor-read-to-submit no-yield window, so the concurrency invariant is untouched. The VerifiedMessage.original removal is justified: no reader exists in src or tests, the verified message is only used for sizing, and the launch path reloads from the CCN. The two new tests are correctly constructed (the event-loop pin reads the flag before awaiting the pinning task, so the read can't be what lets the task run) and the removed deep-copy test went with the field it covered. Minor note: an entry whose hash an earlier duplicate already refused is now fully verified in the thread before assemble_plan skips it, where the old code short-circuited verification — wasted CPU only, no behavior change.

**`src/aleph/vm/agent/allocation/verdict.py`** (line 184): Minor: an entry whose hash was already refused by an earlier duplicate is now fully verified (parse + ecrecover) in the thread before being skipped here, where the old inline code skipped before calling verify_entry. Wasted work on a malformed/redundant push only, no behavior change — not worth blocking, just noting the per-entry cost is now unconditional.

PR #1235, opened 2026-09-09
refactor(agent): install the storage hooks as one object, not three globals

odesenfans:
## Summary
The agent hands three callbacks down to the storage layer: what admits a download against the cache budget, what evicts retained volumes when a placement does not fit, and what runs a reconcile pass after a VM is retired for good. Each lived as a module global in the module that calls it, with its own setter, and startup called the three setters one after another, so a raise between two of the calls left a node that would evict retained volumes for a placement but let downloads blow the cache budget (or the reverse), with neither module able to say which half had landed. Verified against the current tree: `set_room_maker(f)` followed by a raise leaves `room_maker` installed and the other two empty. The three slots become fields of one frozen `AgentHooks`, published in a single assignment, so a wiring that fails installs nothing. A second commit adds the autouse fixture that empties the five agent state globals around every test, which is where the real test-order coupling in these files lives.

Stacked on #1231.

## Scope note
The brief asked for the hooks to be passed explicitly to the functions that need them. Counted first, as the brief's stop condition requires: that design touches 48 call sites in `src/` and 66 in `tests/` at first order (`retire_vm` alone has 19 source call sites, `download_file_in_chunks` reaches its callers only through `download_file` and the five cache getters, 23 more), and the second order adds at least 10 more source sites because `check_payment`, `teardown_vm`, `start_persistent_vm` and `_ensure_program_vm` hold no `app`. That is well past the ~15 ceiling for what the brief calls the riskiest refactor of the series, so this is the smaller alternative the coordinator approved: one object with one owner, no threading. The lookup stays ambient, so "wire before the first create" survives; what goes is the three-globals shape and the half-wired startup window.

## Changes
- `src/aleph/vm/hooks.py` (new): `AgentHooks`, a frozen dataclass with the three slots, plus `current_hooks` / `install_hooks` / an `installed_hooks` context manager that restores the previous instance rather than clearing. The module imports nothing from `aleph.vm` at runtime, which is what lets both the storage modules and the agent reach it; the three callbacks all point back into the reconciler, and the reconciler imports the storage modules itself. The three callable type aliases (`AfterGoneHook`, `CacheAdmission`, `RoomMaker`) move here with the rationale that was attached to them.
- `src/aleph/vm/agent/supervisor.py`: startup builds one `AgentHooks` and installs it once, in place of the three sequential setter calls. `current_hooks()` is the only way to read it back: an extra copy on the app would go stale the moment a single-slot setter installed a new instance.
- `src/aleph/vm/storage.py`, `src/aleph/vm/storage_pools.py`, `src/aleph/vm/agent/vm/retire.py`: the module global goes; each consumer reads its slot from `current_hooks()` at the point of use, as it read its global before. `set_cache_admission`, `set_room_maker` and `set_after_gone_hook` stay as thin shims that write one field into the installed object (`replace`), so the existing fixtures and anything out of tree keep working, and so a single-slot setter cannot silently unwire the other two.
- `tests/supervisor/conftest.py`: an autouse fixture empties `cache._live_snapshot`, `reconciler._creating`, `reconciler._last_supervisor_hashes`, `reclaimable._reclaimable_cache` and `storage._reserved_downloads` before and after every test. These are caches and in-process bookkeeping, not wiring, so nothing reset them; the tests that reach into them by hand only ever cleaned up the one they touched.
- `docs/architecture/storage.md`: the four places that named a setter now name the slot, plus one paragraph saying the three are installed as one object. The four code comments that said the agent registers via `set_cache_admission` / `set_room_maker` (`storage.py`, `agent/vm/downloader.py`, `agent/migration/runner.py`, `agent/vm/cache.py`) name the slot too.

## Test plan
- `tests/supervisor/test_agent_hooks.py` (new, 5 cases): a wiring that raises while being built installs no slot at all (the property three setters cannot have); the three consumers read one installed object; a single-slot setter leaves the other two alone; `installed_hooks` puts back what was there; an unwired node has every slot empty.
- `tests/supervisor/test_agent_state_isolation.py` (new, 2 cases): one test leaks all five state globals, the next checks that none of it arrived. Run against the base tree without the fixture, the second case fails, which is the leak it pins.
- `pytest` on `test_agent_hooks.py`, `test_agent_state_isolation.py`, `test_reconciler.py`, `test_reclaimable.py`, `test_vm_purge.py`, `test_cache*.py`, `test_retire.py`, `test_storage*.py`, `test_build_supervisor.py`: 311 passed on the base, 318 passed here.
- `pytest tests/supervisor`: base `06b7fb39` (temporary worktree) 1 failed, 1226 passed; this branch 1 failed, 1233 passed. Identical FAILED set, only the pre-existing `test_log.py::test_make_logs_queue` journald-stub failure.
- `ruff format --diff`, `isort --check-only --profile black` on every changed file: exit 0. `mypy src/aleph/vm/ tests/ examples/example_fastapi runtimes/aleph-debian-12-python` (the CI invocation, from the worktree root): exit 0, 210 source files. `ruff check` reports the same three complexity findings on `storage.py` and `storage_pools.py` as the base commit does, none introduced here.

- `tests/supervisor/conftest.py`: the reset moved out of the fixture into `reset_agent_state()`, and it now also reinstalls an empty `AgentHooks`, so a hook leaked by a test that dies before its cleanup cannot contaminate later tests.
- `tests/supervisor/test_agent_state_isolation.py`: one test that leaks the state, calls that reset, and asserts the emptiness it promises, instead of two tests that only worked in definition order.
- Not changed: the `install_hooks` / `installed_hooks` naming.
- Checks on the rebased branch: `ruff format --diff`, `isort --check-only --profile black`, `mypy src/aleph/vm/ tests/` clean. Test suites are left to CI.


foxpatch-aleph (approved):
This is a well-executed refactor: the three module-global hooks become one frozen AgentHooks object installed in a single assignment, giving all-or-nothing startup wiring, and the read-at-point-of-use semantics are faithfully preserved in all three consumers. I verified the seven new tests pass, 200+ related tests pass, there are no stale references to the removed globals anywhere in the tree, the conftest fixture's five reset globals all exist and are safe to reset by reassignment, and hooks.py introduces no import cycle. The full-suite failures I observed are all sandbox permission issues on /var/cache/aleph and /var/lib/aleph, unrelated to the PR. Only non-blocking nits remain.

**`src/aleph/vm/storage.py`** (line 131): Non-blocking: this read-modify-write (and the twins in storage_pools.py:318 and retire.py:73) is not atomic — two concurrent shim calls could lose one update, which the old independent globals could not. In practice production never calls the shims (startup uses install_hooks once) and tests are sequential, so it's fine, but a half-sentence in the docstring noting the shims are not safe for concurrent use would make the contract explicit.

**`tests/supervisor/test_agent_hooks.py`** (line 38): The raise here happens during argument evaluation, so this case pins Python's evaluation order rather than the production construction in supervisor.run(). It demonstrates the property, but exercising run()'s actual AgentHooks construction (or a helper that builds it) would make the all-or-nothing guarantee cover the real wiring path too. Non-blocking.

**`src/aleph/vm/storage.py`** (line 464): This comment (and the similar ones in downloader.py:71 and agent/migration/runner.py:302) still points at storage_pools.set_room_maker / storage.set_cache_admission, which now describe the test-facing shims rather than the production wiring (install_hooks at startup). Still accurate since the shims exist, but updating the wording would avoid confusion. Very minor.

foxpatch-aleph (approved):
This is a well-executed, carefully documented refactor. The three module-global hooks are consolidated into one frozen AgentHooks published by single assignment, which genuinely fixes the half-wired-startup window described; the consumers read current_hooks() at the point of use (storage.py:164, storage_pools.py:344, retire.py:223) with no stale caching, and the single-slot shims built on dataclasses.replace preserve the other two slots as documented. Import-graph placement of hooks.py is correct (nothing from aleph.vm at runtime; the string forward-ref is never resolved at runtime), the removed supervisor.py imports are genuinely unused, and the conftest resets match how each module owns its global. New and existing tests pass locally with no regressions; the mypy/ruff findings are pre-existing on untouched lines. Remaining points are non-blocking nits.

**`tests/supervisor/test_agent_state_isolation.py`** (line 32): Non-blocking: this isolation pin relies on pytest executing the two tests in definition order (documented in the module docstring). If a shuffling plugin (e.g. pytest-randomly) is ever adopted, both tests silently become trivially green. An alternative that doesn't depend on ordering would be a single test that leaks the state, calls the fixture's reset logic (or drives the fixture via a nested fixture request), and then asserts emptiness.

**`tests/supervisor/conftest.py`** (line 163): Non-blocking: the fixture clears the five state globals but not the installed hooks, yet a hook leaked by a test that dies between set_room_maker(fn) and its cleanup contaminates later tests in exactly the same way. Today every test installs hooks via try/finally or installed_hooks so the leak can't happen, but adding install_hooks(AgentHooks()) to reset() would close the same class of cross-test leak the fixture exists for, and make the 'wiring is clean' property just as guaranteed as the state property.

**`src/aleph/vm/hooks.py`** (line 84): Minor readability nit: install_hooks vs installed_hooks are near-identical names for publish-forever vs publish-and-restore. The tests use both correctly and the docstrings disambiguate, so this is purely observational — but a reader skimming supervisor.py's install_hooks(hooks) next to the tests' installed_hooks(...) needs the leading 'd' to tell them apart.

foxpatch-aleph (approved):
This PR cleanly replaces three module-global hooks with one frozen AgentHooks dataclass installed in a single assignment. The design is correct: hooks.py sits at the top of the import graph with no runtime aleph.vm imports (TYPE_CHECKING only), consumers read current_hooks() at the point of use exactly as they read their globals before, and the single-slot shims preserve sibling slots via dataclasses.replace, so existing fixtures keep working. I verified the 7 new tests pass, 271 tests in the directly affected suites pass, and the full tests/supervisor suite passes except 29 failures and 2 collection errors that I traced to environment causes (sandbox permission denied creating /var/cache/aleph, and the installed aleph-message version missing ConfidentialGpuRequirement/TdxRegisters) — none in PR-touched modules. ruff format is clean on all changed files and mypy is clean on the new files (remaining findings are pre-existing and stem from a newer local mypy than CI's pinned 1.8.0). The remaining observations are non-blocking nits: the read-modify-write in the three single-slot shims is not atomic (safe today since production only calls install_hooks once at startup, but worth a comment), the install_hooks/installed_hooks names differ by two characters, and the new ARG005 findings in test_agent_state_isolation.py are invisible to CI since ruff check is not enforced.

**`src/aleph/vm/hooks.py`** (line 86): Minor: the three single-slot shims (set_room_maker, set_cache_admission, set_after_gone_hook) do a read-modify-write via replace(current_hooks(), ...), which is not atomic — two concurrent shim calls could lose one slot write. Production never hits this (the agent publishes once via install_hooks), but since this module's raison d'être is install atomicity, a one-line note in the shims' docstrings saying they are not safe to call concurrently would document the boundary.

**`src/aleph/vm/hooks.py`** (line 91): Nit: install_hooks vs installed_hooks differ by two characters and read almost identically — in test_agent_hooks.py I initially expected the context manager where the plain function was used. A name like hooks_installed_during or temporarily_install would be unambiguous. The PR notes the naming was deliberately kept; fine as a follow-up.

**`tests/supervisor/test_agent_state_isolation.py`** (line 31): Nit: the lambda's unused pool/needed args trip ARG005 under ruff check (not CI-enforced, since CI only runs ruff format + isort). The sibling test file uses lambda *_args: 0; matching that would keep the new files finding-free.

**`tests/supervisor/test_agent_hooks.py`** (line 33): Observation: this case pins Python's argument-evaluation order (AgentHooks(...) raises before install_hooks is entered) more than production behavior — the real call site in supervisor.py builds lambdas that cannot realistically fail. It documents the property well, but a companion case exercising the supervisor's actual startup shape would make it airtight. Non-blocking.

**`tests/supervisor/conftest.py`** (line 186): The function-body imports mean a rename of any of the five private globals now breaks every supervisor test at call time rather than at collection time. The lazy import is a reasonable trade (avoids the cost when the fixture is unused elsewhere), just flagging the coupling; the suite would still fail loudly.

foxpatch-aleph (approved):
This is a well-designed, well-tested refactor. Replacing three module globals (each with its own setter, installed sequentially at startup) with one frozen AgentHooks published in a single assignment is correct: argument evaluation precedes install_hooks, so a failed startup wires nothing; consumers read their slot once at the point of use, preserving the old read-at-call-time semantics; the single-slot setters use dataclasses.replace, so a test calling one cannot silently unwire the other two; and the type aliases and setters remain importable from their old locations for in-tree and out-of-tree callers. I verified the diff matches the tree, that no references to the removed globals remain in src/tests/docs, that hooks.py introduces no import cycle and does not violate the import-linter contract, and ran the tests: all focused and broader suites pass (63 + 154), and the only full-suite failures unique to this branch in my sandbox trace to environment issues (/var/cache/aleph permissions and a stale local aleph_message), not the PR. ruff format is clean on the changed hunks and mypy is clean on the new files. Two nits below are cosmetic and non-blocking.

**`tests/supervisor/test_agent_state_isolation.py`** (line 31): Minor nit: `lambda pool, needed: 0` trips ARG005 (unused lambda arguments) under the repo's selected `ARG` rules with a current ruff, which slightly contradicts the PR description's claim that `ruff check` introduces no new findings. CI pins ruff==0.4.6 and its style job only runs `ruff format --diff`, so nothing fails, but `lambda *_args: 0` would keep `ruff check` quiet for anyone running it locally.

**`tests/supervisor/conftest.py`** (line 200): Non-blocking observation for future readers: because this autouse fixture reinstalls an empty AgentHooks around every test, a future module- or session-scoped fixture that installs hooks once would have its installation silently wiped between tests. All current installations are function-scoped (verified), so this is fine today; a one-line note in the docstring that hook installation must stay per-test would pin that assumption.

foxpatch-aleph (approved):
This PR cleanly replaces three module-global hook variables with one frozen AgentHooks dataclass published in a single assignment, and the execution is faithful to that design: every consumer reads its slot via current_hooks() at the exact point where the global was read before, the module sits at the top of the import graph with no runtime aleph.vm imports, and the three single-slot setters are preserved as shims via dataclasses.replace so existing tests and out-of-tree callers keep working. I verified the new hooks module compiles and behaves as documented (frozen instances, nested installed_hooks restore), ran the new tests plus the affected suites (217 tests, all passing), confirmed no stale references to the removed globals anywhere, and found that the full supervisor suite's only failures are pre-existing sandbox-environment issues (root-required /var/cache/aleph paths, DNS setup) unrelated to this change. The autouse clean_agent_state fixture correctly resets the five process-wide globals and reinstalls empty hooks around every test, and the isolation test is properly self-contained rather than order-coupled. Docs and comments are updated consistently. The only nit is that the single-slot setters now do a non-atomic read-modify-write on the installed hooks, which is theoretically less race-safe than the old independent globals, though nothing in-tree calls setters concurrently.

**`src/aleph/vm/hooks.py`** (line 84): Minor observation: the single-slot shims (set_room_maker, set_cache_admission, set_after_gone_hook) do a non-atomic read-modify-write via install_hooks(replace(current_hooks(), ...)), so two threads calling different setters concurrently could lose an update — the old independent globals were race-free in that case. Nothing in-tree calls the setters concurrently (tests are single-threaded; the agent installs via install_hooks in run() before serving), so this is theoretical, but a one-line note on install_hooks that concurrent slot mutation is not supported would document the contract.

foxpatch-aleph (approved):
This is a clean, well-executed refactor. The three hook globals become one frozen dataclass installed in a single assignment, with read-at-point-of-use preserved in all three consumers, backward-compatible single-slot shims that use dataclasses.replace so they cannot silently unwire the other two slots, and no import cycles (hooks.py pulls nothing from aleph.vm at runtime; the type aliases remain importable from their original modules). The new autouse fixture addresses real test-order coupling, and its resets are safe because all internal consumers resolve those globals through module lookups at call time. I verified the 6 new tests and all affected suites pass, that no stale references to the old globals or setters remain anywhere in the repo, that the docs changes are accurate, and that the only local failures are environmental (sandbox filesystem permissions and a stale aleph_message package). The only nits are minor test-quality suggestions; nothing blocking.

**`tests/supervisor/test_agent_hooks.py`** (line 57): The assertion compares against an empty AgentHooks(), but the previous wiring in this test is also empty, so the test cannot distinguish \"the previous wiring was untouched\" from \"a buggy install_hooks caught the exception and installed empty hooks\". Installing a non-trivial wiring before the failing construction and asserting it survives would pin the \"previous wiring untouched\" property the docstring promises more strongly.

**`tests/supervisor/test_agent_hooks.py`** (line 31): Minor: this autouse clean_hooks fixture is now redundant with the conftest's clean_agent_state, which already reinstalls empty hooks around every test. Harmless and arguably good for locality, but worth knowing it overlaps.

**`src/aleph/vm/agent/supervisor.py`** (line 560): Worth noting for the commit message or docs, the atomicity win here is mostly a tested invariant rather than a live production risk: the three old setter calls were adjacent in run() and none of them raise in practice. The real wins are the single read point (current_hooks()), snapshot semantics for readers when replace() installs a new instance, and the test isolation the second commit adds.

foxpatch-aleph (approved):
Clean, behavior-preserving refactor. The frozen AgentHooks object with a single install gives the all-or-nothing startup property that three sequential module setters could not, and the single-slot setter shims (via replace()) keep every existing test working without risking partial un wiring. Consumers read slots at the point of use, matching the old global-read semantics; hooks.py introduces no import cycles; the autouse state reset targets real test-order coupling and cannot break wider-scoped fixtures since none exist in tests/supervisor. I verified the new tests pass, the full supervisor suite passes except pre-existing environment-caused failures, and ruff/mypy findings on changed files are all pre-existing. Docs are updated accurately and the scope-note deviation is honestly justified.

**`tests/supervisor/test_agent_hooks.py`** (line 31): Non-blocking: the module-level autouse clean_hooks fixture is now partially redundant with the conftest's clean_agent_state, which reinstalls empty AgentHooks around every supervisor test. Keeping it is reasonable belt-and-braces, but worth knowing the conftest already covers this.

**`tests/supervisor/conftest.py`** (line 163): Non-blocking: reset_agent_state reaches into five private module globals by name, which will silently drift if those modules are refactored. It is documented and test-only, so acceptable, but a targeted test asserting the reset's coverage (as test_agent_state_isolation.py does) is the right mitigation.

foxpatch-aleph (approved):
Sound refactor that replaces three module-global hooks with one frozen AgentHooks object published in a single assignment, closing the half-wired startup window. The import graph is clean (hooks.py imports nothing from aleph.vm at runtime, the StoragePool forward ref never resolves), no references to the old globals or moved type aliases remain, the three consumers preserve read-at-point-of-use semantics, and the backwards-compatible single-slot shims keep existing fixtures working. Verified 6 new tests plus 414 existing tests across the touched suites pass; the only full-suite failures are environmental (root-only /var/cache/aleph and a stale aleph_message version), none attributable to this PR. The autouse reset in tests/supervisor/conftest.py correctly clears five genuine cross-test leaks and cannot clobber session-scoped setup (none exists). Docs are updated consistently. Two non-blocking nits below.

**`tests/supervisor/test_agent_hooks.py`** (line 49): This test never reaches install_hooks: the RuntimeError fires while evaluating the AgentHooks(...) arguments, so it pins Python's argument-evaluation order rather than any library behavior — any implementation, including three separate setters with the same call shape, would pass it. Harmless as a readable statement of the property the refactor relies on, but it cannot catch a regression; if you want it to bite, have the explode-on-argument path coexist with a second, real install attempt that must observe the untouched previous wiring.

**`src/aleph/vm/storage_pools.py`** (line 330): The single-slot shims here (and in storage.py:131, retire.py:73) are a read-modify-write, so two concurrent single-slot sets can lose one. Fine today since production only calls install_hooks once at startup, but worth a one-line note in the docstring if these shims ever gain a production caller outside startup.

foxpatch-aleph (approved):
This is a well-executed, well-documented refactor. The three module-global hooks (cache admission, room maker, after-GONE) are consolidated into one frozen AgentHooks dataclass published in a single atomic assignment, closing the real half-wired-startup window the PR describes. The new module correctly sits at the top of the import graph with no runtime imports from aleph.vm; all consumers read their slot at the point of use; the old setters are preserved as shims that use dataclasses.replace so a single-slot write cannot silently unwire the other two. The behavior-preserving nature of the change is backed by focused new tests (all pass), a full tests/supervisor run whose 31 failures I traced individually to sandbox environment issues (stale aleph_message package, root-only paths, internet-dependent tests) with none touching the PR's code paths, plus clean ruff-format and no mypy errors in changed files. Docs and comments were updated consistently. Two non-blocking nits: the single-slot shims' read-modify-write is not atomic where the old per-global setters were (harmless in practice since they are only called single-threaded at startup or from tests), and the lambda in test_agent_state_isolation.py trips ARG005 under a local `ruff check`.

**`src/aleph/vm/agent/vm/retire.py`** (line 73): Non-blocking: this read-modify-write (also in set_cache_admission and set_room_maker) is not atomic. With the old design each setter wrote its own distinct global, so two threads setting different slots could not lose an update; now two concurrent shim calls on different slots can drop one. In practice these are only called single-threaded at startup and from tests, so this is theoretical, but a one-line note in the docstrings ("not safe to call from two threads") would pin the assumption.

**`tests/supervisor/test_agent_state_isolation.py`** (line 31): Minor: `lambda pool, needed: 0` trips ARG005 (unused lambda arguments) under `ruff check`. CI's style script only runs `ruff format`, so nothing fails, but a local `ruff check` will flag this new file. `lambda _pool, _needed: 0` or a def would silence it.

PR #1236, opened 2026-09-09
refactor(daemon): typed GPU probe errors, deduped walks, and comments that state their reason

odesenfans:
## Summary

Refactor-only cleanup of the Rust supervisor daemon and controller: the two
stringly `DaemonError` variants become structured ones, four duplicated walks
become two helpers, the last saturating BAR sum becomes a checked one, and
every comment that cited a document instead of stating its reason now states
it. Nothing changes at runtime except the three failure paths named below,
which report the same facts through typed fields instead of pre-formatted
strings.

The comment half is not cosmetic. A reader who hit "design doc section 3",
"ledger entry 13" or `docs/plans/2026-07-04-rust-supervisor-daemon-design.md`
got a pointer they could not follow: that design doc was never committed, the
divergence ledger moved, and two dozen module headers introduced themselves as
"a port of src/aleph/vm/<something>.py" for files the agent/supervisor split
deleted. Two of the citations were also simply wrong (the NUMA single-node note
cited the ledger entry about duplicate CLI secret keys; a DHCP teardown note
cited the hugepages entry), which is its own argument for inlining the reason
rather than referencing it.

Stacked on #1233.

## Changes

- `error.rs`, `lspci.rs`, `gpu_bar.rs`: `Lspci(String)` and `GpuProbe(String)`
  are replaced by `LspciSpawn { arguments, source }`, `LspciStatus
  { arguments, status }`, `LspciLine { line }`, `GpuResourceRead
  { path, source }`, `GpuResourceField { field, source }` and
  `GpuResourceLine { line }`. Display text is equivalent to what each site
  formatted before, and the `io::Error` / `ParseIntError` are attached as
  sources instead of interpolated.
- `gpu_bar.rs`: `gpu_mmio64_mb_under` summed each card's BAR total with
  `saturating_add`, the one arithmetic site #1227 left alone. A saturated sum
  silently under-sizes the window, which is the exact failure the module
  exists to prevent (the guest enumerates the card and finds its BARs
  unassigned). Now `checked_add` into a new `GpuBarTotal { pci_host }`.
- `gpu_bar.rs`, `lifecycle.rs`: the window sizer takes `&[&str]`, so
  `snp_config_slice_with` no longer clones every pci address into a
  throwaway `Vec<String>` for a callee that only joins them into a path.
- `lifecycle.rs`: `inventory_gpu(state, pci_host)` replaces the inventory
  `find` spelled out in both `validate_spec_gpus` and `snp_config_slice_with`.
- `service.rs`: `attached_gpus(world)` replaces the fold over the world view's
  attached cards in both `host_info` and `refresh_cc_modes_with`, yielding each
  card with whether the VM holding it is confidential (the only difference
  between the two copies).
- `lspci.rs`: `GpuDevice.cc_mode` reads like a field the parser fills and never
  can (lspci reports nothing about the mode). The doc comment now says it is a
  reporting field written only by the host info path, on its own clone.
- Daemon and controller comments: every `docs/plans` path, "design doc section
  N", "divergence N", "ledger entry N" and "(ledgered)" now states its reason
  in words, and the module headers name the Python donor by symbol
  (`LocalSupervisor.get_host_info`, `QemuVM.start`, `TapInterface`, the
  SQLAlchemy `PortMapping`) rather than by a path that no longer resolves. The
  three Python modules that do still exist keep their paths.
- Controller comments that introduced a function as "the port of" a Python
  symbol (`__main__.main`, `QemuVM.stop()`, `QemuConfidentialVM.start()`) now
  describe what the code does, since that controller does not exist on this
  branch. `place_vm_numa` says what honouring a requested NUMA node means (that
  node or a refusal, never a quiet placement elsewhere) instead of naming a
  decision number.
- `proto/supervisor.proto`, `proto/README.md`: the header and README described
  the agent/supervisor split by pointing at a design document this repository
  does not contain, and the README's error-enum section at "Annex A.6" of it.
  Both now state the split. `supervisor_pb2.pyi` carries the header as a module
  docstring and is regenerated; `supervisor_pb2.py` does not encode comments
  and is unchanged, so `scripts/check_proto_clean.sh` stays green.

## Test plan

- `gpu_bar.rs`: `malformed_lines_are_errors` now pins the variants and their
  fields rather than only `is_err()`; `a_card_without_a_resource_file_is_an_error`
  pins `GpuResourceRead` and its path; new
  `a_bar_total_across_cards_that_overflows_is_an_error` drives two cards whose
  BARs each fill half the address space, so the per-file parse succeeds and only
  the cross-card sum overflows, and names the card that pushed it over.
- `lspci.rs`: the multibyte parse test pins `LspciLine`; the cc_mode
  serialization test says why the parser leaves the mode absent.
- `lifecycle.rs`: the unreadable-probe test now builds a typed
  `GpuRegisterRead` instead of a `GpuProbe(String)`.
- Checks, from `rust/`, run bare with the exit code read directly:
  `cargo fmt --check` (0), `cargo clippy --locked --all-targets -- -D warnings`
  (0), `cargo test --locked --workspace` (0, 657 passed / 0 failed),
  `cargo doc --no-deps -p supervisor-daemon` (0; the 5 intra-doc warnings it
  prints are pre-existing and on lines this PR does not touch).
  `scripts/check_proto_clean.sh` (0).
- `aleph-tee` and `aleph-attest-agent` are deliberately untouched, so the
  golden measurements do not move.
- The sweep's own gate:
  `git grep -nE 'docs/plans|divergence [0-9]|ledger entry|decision [0-9]|the port of' rust/crates/supervisor-daemon rust/crates/supervisor-controller proto`
  returns exactly one line, the in-code "NUMA ledger entry" in
  `lifecycle.rs`, which is the placement ledger and not a document.

## Rebased onto dev-2.1 (#1199 + #1200), plus a review fix

One conflict, in the walks-dedupe commit. `attached_gpus` yields each attached card with a flag, and the flag's meaning tightened further down the rebased stack: it is now "a live confidential guest vouches for this card's CC mode", which needs the VM to be confidential AND not stopped. The helper carries that definition (the other caller, the inventory subtraction, ignores the flag), and the field is named `vouched_cc_on` rather than `confidential` so it cannot be read as the plain config question. `inventory_gpu` merged with #1200's new ordering comment on the membership check kept above it.

Review fix, as its own commit: the citation sweep re-run with a wider pattern found two more.
- `tests/grpc.rs`: "ledgered" now says the reason. Persistence means a per-VM systemd controller unit, and this Firecracker path runs the microVM in-process for one request, so there is no unit for a persistent VM to be adopted from or restarted through.
- `rust/rust-toolchain.toml`: dropped the design-doc and section citation for the actual reason (CI gates fmt and clippy with this version, so a local run and a CI run agree on what a warning is). The claim that this file has no measurement impact was verified: `nix/flake.nix` narrows the measured attest-agent derivation's source to the root `Cargo.toml`, `aleph-tee` and `aleph-attest-agent`, and `rust-toolchain.toml` is not in that fileset; the nix build takes its toolchain from the flake.

Left alone deliberately: three `rust-port-divergences` references in `aleph-tee/src/{types,report_data}.rs` and `aleph-attest-agent/src/secrets.rs`. Those crates are in the measured chain, so purging them moves every golden measurement and belongs on a branch that re-seeds, not on a daemon-side quality PR. The many `increment A1/B1/C1/C2` markers are likewise out of scope here.

Checks re-run after the rebase: `cargo fmt --check` exit 0, `cargo clippy --locked --all-targets -- -D warnings` exit 0, and the same pair with `--manifest-path crates/aleph-attest-agent/Cargo.toml` exit 0. Test suites left to CI this round.


foxpatch-aleph (approved):
A clean, faithful refactor. The typed DaemonError variants (LspciSpawn/LspciStatus/LspciLine, GpuResourceRead/GpuResourceField/GpuResourceLine, GpuBarTotal) reproduce the old Display text while attaching io::Error/ParseIntError as #[source]; every call site was updated consistently and no stale GpuProbe(String)/Lspci(String) usages or stale error-message pins in tests remain. The saturating-to-checked BAR sum is a genuine correctness improvement with a well-constructed test (per-card parses succeed at half the address space so only the cross-card sum overflows, and the test pins which card is named). The dedup helpers inventory_gpu and attached_gpus preserve the exact semantics of the walks they replace, including the snp().is_some() confidential flag that is the sole difference between the two attached_gpus callers. The &[String] -> &[&str] window-sizer change eliminates the throwaway clone correctly. I verified the PR's factual claims: all removed docs/plans paths indeed do not exist, the 'entry 82' reference in docs/architecture/divergences.md was unresolvable (the file has no such numbering), the reworded SNP rootfs comment accurately describes aleph-tee's generator, and the regenerated .pyi matches the proto header. Only nit: a leftover '(ledgered)' reference in tests/grpc.rs was missed.

**`rust/crates/supervisor-daemon/tests/grpc.rs`** (line 190): Minor: this comment still says "; ledgered)", a dangling reference to the divergence ledger this PR removes everywhere else (the sibling comment at lifecycle.rs:2756 was rewritten, this one in the integration test seems to have been missed). Same class of unfollowable citation the PR set out to eliminate.

foxpatch-aleph (approved):
A clean, well-scoped quality pass. The six typed DaemonError variants replace the two stringly ones with equivalent Display text and properly attached sources; the saturating BAR sum is correctly upgraded to checked_add with a test that isolates the cross-card overflow; the inventory_gpu and attached_gpus dedups preserve per-site error semantics and set contents exactly; and the &[&str] window-sizer signature matches its only production caller. I verified the PR's own sweep grep returns the single claimed line, that no remaining src/aleph/vm path references dangle except one brace-expansion nit in world.rs, and that the .pyi regeneration is consistent with supervisor.proto while supervisor_pb2.py is correctly untouched. No behavioral risk found; two minor non-blocking comments below.

**`rust/crates/supervisor-daemon/src/world.rs`** (line 887): This comment still cites `src/aleph/vm/network/{hostnetwork,interfaces,ipaddresses}.py`, a path that no longer resolves — the same class of dangling citation this PR sweeps, and it is inside the swept crate. The gate grep only matches `docs/plans|divergence|ledger|decision|the port of`, so `src/aleph/vm` paths are not covered. Consider naming the donors by symbol only, like the other rewritten headers.

**`rust/crates/supervisor-daemon/src/gpu_bar.rs`** (line 139): Nit: `GpuBarTotal` is currently unreachable from the production path — `snp_config_slice_with` rejects specs with more than one GPU (lifecycle.rs:2194), and a single card's total is already rejected inside `parse_resource_file` (GpuBarRange). The guard is still correct for a public function taking a slice, and the test exercises it directly; just be aware the new test is its only caller. (Also noted: `aleph-tee/src/owner_auth.rs:8` still cites a nonexistent `docs/plans/2026-08-18-...` file, though that crate is declared out of scope.)

foxpatch-aleph (approved):
Solid refactor: the typed error variants are complete and source-chained, the saturating-to-checked BAR sum fix is correct and well-tested (the new test properly isolates the cross-card overflow that the per-file guard cannot catch), and both deduplicated walks preserve their call sites' semantics exactly. The comment sweep achieves its stated goal and the inlined reasons I spot-checked against the code are accurate. One leftover: tests/grpc.rs:190 still carries a "(ledgered)" reference that the sweep gate's regex will never catch. Test results were not independently reproduced (no Rust toolchain available in this environment) but the changes compile-consistent by inspection.

**`rust/crates/supervisor-daemon/tests/grpc.rs`** (line 190): This "(ledgered)" is the one leftover of the sweep: the PR prose says every "(ledgered)" citation now states its reason, and the gate's regex (`ledger entry`) does not match the bare word "ledgered", so this one will stay invisible to the check forever. Reword to state the reason inline ("they boot under systemd controller units and follow with the controller port"), like the sibling comment in lifecycle.rs, and consider widening the gate to `ledger(ed| [0-9])?\b` style patterns so the class stays closed. Note rust/rust-toolchain.toml:1 also still cites the design doc, though that file is outside the swept crates.

foxpatch-aleph (approved):
This is a clean, well-executed quality pass. The substantive changes are all correct: the saturating-to-checked BAR sum fix in gpu_bar.rs closes a real failure mode (a saturated total silently under-sizing the MMIO64 window, leaving guest BARs unassigned) and its new cross-card overflow test drives exactly the right path; the attached_gpus and inventory_gpu dedupes faithfully reproduce the walks they replace (the vouched_cc_on flag matches the original confidential computation exactly); and the &[&str] signature change removes a needless clone with all callers updated. The typed error variants preserve the diagnostic facts of the old strings while attaching io::Error/ParseIntError as #[source], and I confirmed no consumer depends on the old message text. The citation sweep's gate holds: only the in-code NUMA ledger reference remains, and I verified the referenced design docs and Python donor paths genuinely do not exist in this repo. Two non-blocking nits noted in comments.

**`rust/crates/supervisor-daemon/src/error.rs`** (line 54): Non-blocking nit on the PR description rather than the code: it claims "Display text is equivalent to what each site formatted before", which is true for the Lspci* variants but the GpuResource* messages are reworded ("GPU probe failed: bad resource field ..." became "PCI resource field ... is not a hexadecimal number: ..."). I grepped for consumers of the old "GPU probe failed: " prefix and found none (the nvidia-cc operator doc references the separate tracing::warn! at service.rs:609, which is unchanged), so this is harmless — just an overstated claim in the description.

**`rust/crates/supervisor-daemon/src/service.rs`** (line 464): The helper is clean and the doc comment accurately explains the flag semantics (confidential AND not stopped, i.e. a live guest vouches for the card's CC mode) and why host_info ignores it. One tiny observation: host_info computes vouched_cc_on per GPU only to discard it, which is acknowledged in the helper's documentation — fine as is.

**`rust/crates/supervisor-daemon/src/gpu_bar.rs`** (line 174): path: path.clone() could be a move since the binding is dead after read_to_string, but the clone is negligible and arguably clearer than restructuring the closure; not worth changing. The checked_add fix itself and the new overflow test are exactly right — I verified the test fixture flags (0x14220c) contain WANTED_FLAGS so the per-card parse succeeds and only the cross-card sum overflows, pinning the second card as the one that trips it.

foxpatch-aleph (approved):
This is a careful, well-executed refactor. The typed GPU/lspci error variants are a strict improvement with all call sites migrated and no stale references remaining; the saturating-to-checked BAR sum change is correct (the per-file parse already guards its own total, so only the cross-card sum could overflow, and the new test pins exactly that case) and cannot regress behavior since the old saturating path was refused downstream by check_mmio64_budget anyway. Both dedup helpers preserve the semantics of the code they replace, and the &<&str> signature change is applied consistently. The comment sweep's factual claims were verified: the cited docs/plans files genuinely don't exist, the nix flake fileset indeed excludes rust-toolchain.toml, the surviving 'NUMA ledger entry' is the placement ledger and not a document, the .pyi header matches the .proto verbatim, and the check_proto_clean.sh enforcement paths don't encode comments. All PR-description claims held up under verification; I found no bugs, no security issues, and no test-coverage gaps for the new behavior.

**`rust/crates/supervisor-daemon/src/gpu_bar.rs`** (line 178): Nit on the comment's rationale: a saturating sum would not under-size the window — it would saturate to u64::MAX, clamp the window to MAX_WINDOW_MB, and be refused by check_mmio64_budget — so the practical outcome was already a refusal. The checked_add is still the right change (the refusal now names the card and happens earlier with a clearer message), but 'Saturating here would quietly under-size the window' overstates what the old path did. Consider softening to say the saturating path silently produced a garbage total whose only rejection came later, at the budget check.

**`rust/crates/supervisor-daemon/src/error.rs`** (line 38): LspciSpawn/LspciStatus carry a pre-formatted arguments String ("-s 06:00.0 -nnk") while the sibling GpuResourceRead/GpuResourceField variants use structured fields. The string form reproduces the prior Display text exactly, which is fine for parity, but it's a slight inconsistency within this same PR's goal of replacing pre-formatted strings with typed data. Non-blocking.

**`src/aleph/vm/agent/vprogram_launch.py`** (line 16): Outside this PR's swept paths, so consistent with its stated scope, but note that a dangling reference to docs/plans/rust-port-divergences.md survives here on the Python side. Worth a follow-up sweep of the same citation class so the next reader of the agent doesn't hit the same dead pointer the daemon PR just removed.

**`rust/crates/supervisor-daemon/src/lspci.rs`** (line 281): Minor test-coverage observation: every new GPU error variant is pinned by an updated test except LspciSpawn and LspciStatus, which sit behind the real-subprocess boundary and remain unconstructed in tests, same as before the change. Acceptable given the seam design (is_kernel_enabled_gpu is injectable, spawn is not), just noting the asymmetry with the rest of the variant work.

foxpatch-aleph (approved):
A disciplined, refactor-only PR that I verified line by line: the six typed error variants fully replace the two stringly ones with no stale references anywhere in the workspace, display text is equivalent to the pre-formatted messages with sources attached, the saturating-to-checked BAR sum is a genuine hardening (the old path silently under-sized the window, which is the exact failure the module exists to prevent, and the new cross-card overflow test is well-designed to pin the guard the per-file tests cannot reach), and the inventory_gpu / attached_gpus helpers faithfully preserve the two duplicated walks' behavior — host_info discards the flag exactly where the old code had no flag. Every rewritten comment I spot-checked states its reason accurately: place_vm_numa's 'honoured or refused' matches the allocator, the aleph-tee 'never a disk line' claim matches sev_snp_qemu_args, the removed Python donor paths genuinely no longer resolve while the kept ones do, the flake.nix fileset really excludes rust-toolchain.toml, and the citation sweep returns exactly the one in-code NUMA placement-ledger match the PR claims. The proto change is comment-only and supervisor_pb2.py does not encode comments, so check_proto_clean.sh stays green, which I confirmed by reading the regenerated .pyi header against supervisor.proto. I could not execute cargo or python in this environment, so fmt/clippy/tests were verified statically; nothing I read suggests a failure. One trivial nit below is non-blocking.

**`rust/crates/supervisor-daemon/src/gpu_bar.rs`** (line 175): Nit, non-blocking: `path: path.clone()` inside the `map_err` closure is only needed because the closure is constructed while `&path` is borrowed for the read in the same expression. Hoisting the read (`let contents = std::fs::read_to_string(&path);`) before the `.map_err` would let the closure move `path` instead of cloning it. Purely cosmetic — it is an error path on a small PathBuf — feel free to ignore.

foxpatch-aleph (approved):
Refactor-only cleanup verified line-by-line: the two stringly DaemonError variants are correctly replaced by six structured ones (sources attached via #[source], Display output preserved for the lspci variants), the checked_add in gpu_mmio64_mb_under closes the last silent-saturation path and sits in the right place (the cross-card loop, with a deterministic test that isolates it from the already-guarded per-file sums), inventory_gpu and attached_gpus are faithful dedups of their prior inline walks (host_info ignores the flag; refresh_cc_modes_with computes the same snp().is_some() && live predicate), and the &[&str] signature change is applied at every call site. The regenerated .pyi matches the new proto header, the rust-toolchain.toml claim checks out against nix/flake.nix's fileset, and the PR's own sweep gate returns exactly the claimed single line. I could not compile locally (no cargo/python toolchain available), but nothing in the tree depends on the old error-message strings and no correctness issue was found in the changed paths. Minor findings: an operator doc still names the deleted GpuProbe variant, and the same class of dead citation survives outside the sweep's scope.

**`docs/operators/nvidia-cc.md`** (line 108): This PR deletes the `DaemonError::GpuProbe` variant, but this line still tells operators to look for it. (gpu_cc.rs never returned it either — it produces GpuRegisterRead/GpuUnreadable — but now the identifier does not exist at all.) A one-line follow-up here would keep the operator doc from naming a symbol this sweep removed.

**`packaging/debian-12.dockerfile`** (line 17): Same class of unfollowable pointer the sweep removes, but outside its declared scope: all five packaging dockerfiles still cite the never-committed design doc (2026-07-04-rust-supervisor-daemon-design.md, section 9), and rust-port-divergences is still cited in nix/flake.nix, nix/init.sh, nix/udhcpc.script and src/aleph/vm/agent/vprogram_launch.py even though no such file exists in the repo. Non-blocking — just noting the cleanup is incomplete repo-wide.

**`rust/crates/supervisor-daemon/src/gpu_bar.rs`** (line 175): Nit: `path` is not used after the read, so the closure could capture it by move (`path: path`) instead of cloning; the `read_to_string(&path)` borrow ends before `map_err` consumes the closure. Purely cosmetic given this runs once per card on the create path.

foxpatch-aleph (approved):
Clean, well-verified refactor. The typed error variants are complete and consistent (no stale references to the removed Lspci(String)/GpuProbe(String) anywhere in the workspace), the attached_gpus and inventory_gpu helpers faithfully deduplicate the previous inline walks with identical semantics, and the saturating-to-checked BAR sum is correctly placed at the only site that can overflow (the cross-card loop, since per-file totals are already guarded by GpuBarRange) with a test that drives exactly that case. I independently verified the .pyi/proto header regeneration, the citation-sweep gate output, the nix fileset claim behind the rust-toolchain.toml comment, and that kept Python donor paths exist. Minor nits only: LspciLine's Display hardcodes "-mmnnn" where its sibling variants carry arguments as a field, and the new Display wording differs slightly from the old strings despite the PR's "equivalent" claim (nothing pins the old text).

**`rust/crates/supervisor-daemon/src/error.rs`** (line 51): Minor nit: LspciLine's Display hardcodes "-mmnnn", while LspciSpawn and LspciStatus carry the invocation arguments as a field. Today this variant is only constructed in the -mmnnn parser so the message is accurate, but the asymmetry means a future reuse for the -s <addr> -nnk path would silently mislabel the line's source command. Consider deriving the text (or carrying the arguments) if the variant ever gets a second call site.

**`rust/crates/supervisor-daemon/src/gpu_bar.rs`** (line 183): Verified: a single file's total cannot reach this checked_add (parse_resource_file already errors via GpuBarRange "it pushes the BAR total past 64 bits"), so GpuBarTotal is unreachable except on the cross-card sum — and the new test constructs exactly that, with domain-less addresses that sysfs_device_dir_under normalizes so the second card ("07:00.0") is correctly named. Good placement of the guard.

foxpatch-aleph (approved):
Refactor-only PR that holds up under static review. The one behavioral fix (saturating_add -> checked_add for the cross-card BAR sum, gpu_bar.rs:181) is correct and necessary, with a test that genuinely isolates the cross-card overflow from the already-guarded within-file one; I verified the fixture flag word carries WANTED_FLAGS so both lines are counted. The two deduped walks (inventory_gpu, attached_gpus) reproduce the prior inline code exactly at every call site, and the &amp;[String] -> &amp;[&amp;str] signature change is propagated to the sole production caller and all test closures. No references to the removed stringly variants (GpuProbe, Lspci) remain, and #[source] attachments are correct. The PR's factual claims check out: the citation sweep returns exactly the one claimed in-code 'NUMA ledger entry' (a data structure, not a document), the nix flake's fileset excludes rust-toolchain.toml so the golden measurements cannot move, and the proto change is comment-only with a matching regenerated .pyi. The rewritten comments I spot-checked state reasons that are accurate to the code. Minor non-blocking note: a few error Display strings changed prefixes (e.g. 'GPU probe failed:' -> 'cannot read the PCI resource file ...'), which affects InvalidBackend message text but no error codes and nothing pinned the old text. I could not run cargo (no toolchain available in this review environment), so fmt/clippy/test were not independently re-executed; all other verification was done directly against the source.

**`rust/crates/supervisor-daemon/src/gpu_bar.rs`** (line 174): Minor nit: `path: path.clone()` inside the map_err closure is only executed on the error path, so it's fine, but since `path` is unused after this point you could move it into the closure instead of cloning (`map_err` takes `FnOnce`). Harmless either way; not worth a change on its own.

**`rust/crates/supervisor-daemon/src/error.rs`** (line 54): Heads-up on wire-visible text: GpuResourceRead's Display changes the message prefix from the old 'GPU probe failed: cannot read {path}' to 'cannot read the PCI resource file {path}: {source}'. This text flows into InvalidBackend RPC payloads via snp_config_slice_with. Nothing pinned the old text and the error code is unchanged, so this is fine — just noting that any consumer matching on message text would see new strings.

foxpatch-aleph (approved):
A careful, well-executed refactor-only PR. The typed-error conversion is complete and consistent (all Lspci/GpuProbe call sites migrated, sources attached rather than interpolated, no tests pinning the old strings), the saturating-to-checked BAR sum is a strict fail-closed improvement whose new test correctly isolates the cross-card overflow from the per-file checks, both walk dedupes are semantically faithful, and the signature change to &[&str] removed the throwaway clones as described. I verified the PR's own gate: the citation grep returns exactly the claimed one in-code 'NUMA ledger entry' line, the .pyi matches the regenerated proto header, the rust-toolchain.toml claim holds against nix/flake.nix's narrowed fileset, and the rewritten comments' factual claims (NUMA hint-or-refuse, firecracker direct-children, checks.rs gating, world.rs negative vm_index) all check out against the code. Two non-blocking nits: world.rs:894 still cites the deleted src/aleph/vm/network/{hostnetwork,interfaces,ipaddresses}.py path (the same class of dangling citation this PR sweeps, missed here), and the new freeze-timeout comment in service.rs:1546 slightly overclaims NaN-safety (f64::clamp propagates NaN, so GUEST_FREEZE_TIMEOUT=NaN would still panic from_secs_f64; pre-existing behavior, comment text only).

**`rust/crates/supervisor-daemon/src/world.rs`** (line 894): This sweep's own class of miss: the citation here still names src/aleph/vm/network/{hostnetwork,interfaces,ipaddresses}.py, a path the agent/supervisor split deleted. The equivalent donor in tap.rs was reworded to name the symbols only; this one (and the TapInterface properties it describes) should get the same treatment for consistency with the PR's stated gate.

**`rust/crates/supervisor-daemon/src/service.rs`** (line 1546): Minor accuracy nit on the new comment: "a crafted value cannot panic Duration::from_secs_f64" is not quite true for NaN, since f64::clamp propagates NaN (Rust std documents that an initial NaN returns NaN regardless of min/max) and GUEST_FREEZE_TIMEOUT=NaN parses successfully through get_f64, so from_secs_f64 would still panic. The clamp predates this PR and the behavior is unchanged; only the comment wording is new. Consider "a negative or oversized value" instead of "a crafted value".

foxpatch-aleph (approved):
High-quality refactor. The typed Lspci/GpuResource variants are strictly better (sources attached via #[source] instead of interpolated), the checked_add fix in gpu_mmio64_mb_under is real but behavior-compatible (the old saturated sum always ended in a refusal via check_mmio64_budget anyway, so this fails the same creates earlier with a better-typed error), and the attached_gpus and inventory_gpu dedupes are field-for-field equivalent to the walks they replace, with the vouched_cc_on flag semantics documented where the one caller that ignores it can't misread it. The new overflow test is well constructed. I verified the PR's claim that the old citations were wrong against the surviving ledger (entry 73 is duplicate-secret-key handling, entry 77 is hugepages, not DHCP/NUMA), and that docs/plans and the deleted Python donors are indeed unresolvable. Remaining nits are external to the swept paths: docs/operators/nvidia-cc.md still names the deleted DaemonError::GpuProbe variant, and the divergence ledger's preamble still says entry numbers are 'referenced by tests and comments', which after this PR holds only for the aleph-tee crates. Note cargo/python were unavailable in my environment, so the green fmt/clippy/test/proto-clean claims are taken on the PR's word; the diff is internally consistent with them.

**`docs/operators/nvidia-cc.md`** (line 108): This PR deletes `DaemonError::GpuProbe`, so this operator doc now points at a symbol that no longer exists — exactly the class of dangling reference the PR removes elsewhere, and it falls outside the sweep gate because the gate only covers rust/crates and proto. The BAR0 read failures are now `DaemonError::GpuRegisterRead` / `DaemonError::GpuUnreadable` in gpu_cc.rs. (The `GPU CC mode probe failed` log-line hint on the preceding lines is still accurate — service.rs still logs that literal.)

**`docs/architecture/divergences.md`** (line 12): Now that the daemon and controller comments no longer cite entry numbers, this sentence ("entry numbers are stable and referenced by tests and comments") is only true for the three aleph-tee/aleph-attest-agent references deliberately left in this PR. A one-line touch-up here would keep the ledger's self-description truthful. Also worth noting the header at line 3 already drops 'and comments', so the two lines now disagree.

**`rust/crates/supervisor-daemon/src/error.rs`** (line 59): Minor description nit, not a code issue: the PR description says the Display text is equivalent to what each site formatted before, but GpuResourceField changed wording ("bad resource field ..." → "PCI resource field ... is not a hexadecimal number") as did GpuResourceRead ("cannot read {path}" → "cannot read the PCI resource file {path}"). Per ledger entry 33 these texts are explicitly unpinned, so nothing breaks; just the PR text slightly oversells equivalence.

PR #1237, opened 2026-09-09
refactor(agent): one refusal type, one retention rule, and the allocation dead surface

odesenfans:
## Summary
Code quality on the allocation modules, after the eight fixes that came before it. The allocation plane said "no" in four hand-written dict shapes with two independent vocabularies of failure codes, the verdict wrote the retention rules out a second time by hand (already one branch behind the copy next door), the verdict declared its own Protocols restating two signatures that have drifted before, and a handful of comments described code the chain has since changed.

Two real defects fell out of the review items and are fixed with tests: a candidate whose volume directory still carries its `.reclaimable` marker was credited those bytes twice (free space and discount), which had the plan simulation promise room the node did not have; and four boundary error codes that mean "this node is confused" were published to the scheduler as "the hypervisor refused to run this VM".

Stacked on #1234.

### The one behaviour change: five refusal sentences

Every refusal code now publishes one sentence, written in `refusal.py`, so the answer to a push and the executions list cannot describe the same code two ways. The codes are unchanged, and the `invalid_message` refusal still carries the verifier's own reason. What changed is the prose, in all five places the verdict used to author it by hand:

| code | before | after |
| --- | --- | --- |
| `insufficient_capacity` | not enough capacity on this CRN | This node has no room for this VM |
| `node_mismatch` | this instance is allocated to a different node | This VM is allocated to a different node |
| `node_hash_unknown` | this node has not discovered its own hash yet | This node has not discovered its own hash yet |
| `message_required` | embed the signed message for this VM to be sized | Embed the signed message for this VM to be sized |
| `gpu_unavailable` | no available GPU matches this request | No available GPU matches this request |

The last three differ by their first letter only; the first two are reworded, `insufficient_capacity` onto the sentence the executions list already published for that code and `node_mismatch` because a v-program is not an instance.

## Changes
- `allocation/refusal.py` (new): `AllocationFailureCode` (now covering the admission codes the verdict used to spell by hand) plus `Refusal`, one code and one sentence (the five in the table above), serialized in one place by `as_dict`. Its own module because the capacity manager imports it and the exception classifier next door imports the create path, which imports the capacity manager.
- `allocation/failures.py`: keeps `classify_start_failure`. `VM_NOT_FOUND`, `VM_ALREADY_EXISTS`, `HOST_NOT_FOUND` and `INTERNAL` now publish `internal` rather than `supervisor_error`; `PORT_UNAVAILABLE` keeps the default, since that one really is the hypervisor declining.
- `capacity.py`: `AdmissionVerdict` carries a `Refusal` and reads `accepted` off it, so a verdict cannot refuse without saying why. `existing_volume_files` skips a directory that still carries its reclaimable marker. `PlanAdmission` Protocol published here and inherited by `CapacityManager`.
- `vm_registry.py`: `RecordLookup` Protocol published here and inherited by `AgentVmRegistry`.
- `allocation/verdict.py`: uses the two published Protocols instead of restating their signatures; `_retention_reason` gone; an entry with no usable hash is answered under its position in the body (`vms[3]`) rather than under `str(None)`, which collapsed every such entry into one; `compute_verdict` documents what a refusal means for a VM whose teardown is already running.
- `allocation/teardown.py`: `retention_reason` returns the reason or None, and `is_removable_by_allocation` is that answer read as a boolean.
- `allocation/plan.py`: `FailureRecord.first_failed_at` dropped (written, never read).
- `allocation/reconciler.py`: `TEARDOWN_STATUSES` comment corrected; the failed-start warning gains `exc_info=True`, the log being the only place the reason can be read.
- `run.py`, `views/operator.py`: `create_vm_execution`'s ignored `persistent` parameter and its call sites are gone; persistence comes from the message.
- `views/__init__.py`: `_times_dict` answers for a VM the supervisor does not know too, so the executions list builds that block one way; the v1 handler's docstring says it stops RUNNING VMs only where a plan's pass stops RUNNING, STOPPED and FAILED.
- `views/allocations_v2.py`: serializes through `Refusal.as_dict`; the comment claiming nothing yields between the VM list and `submit()` now describes the two awaits that are actually there.

Brief items that were already done or moot: `VerifiedMessage.original` was dropped by #1234; `AllocationState`, `state_for`, `planned_hashes` and every other `FailureRecord` field are read by `_allocation_block` or the executions list; `dict.fromkeys(_TIMES_KEYS)` had one call site, not two, so the helper unification is between it and `_times_dict`.

## Test plan
- `test_allocation_failures.py`: the four boundary codes that now say `internal`; one wire shape and one code behind the refusal a full host answers with, wherever it answers.
- `test_allocation_teardown.py`: `retention_reason` per reason, and the v-program case where the two used to disagree.
- `test_allocation_verdict.py`: every entry with no usable hash is answered on its own; a VM whose teardown is in flight that the host refuses is dropped and stays gone, with its hash still in the plan's refused set.
- `test_agent_capacity.py`: a directory marked reclaimable discounts nothing (fails on the base: `assert 0 == 20000`); `existing_volume_files` skips it whole.
- `test_allocation_reconciler.py`: a cancellation inside the delete leaves no VM marked removing (fails if the discard leaves its `finally`).
- Checks: `tests/supervisor` `1 failed, 1258 passed, 1 skipped` against `1 failed, 1242 passed, 1 skipped` on the base, the same single failure both times (`test_log.py::test_make_logs_queue`, journald, environment-only); `tests/migration` 24 passed; `ruff format --diff` exit 0; `isort --check-only --profile black` exit 0; `mypy src/aleph/vm/ tests/ examples/example_fastapi runtimes/aleph-debian-12-python` from the worktree root, `Success: no issues found in 212 source files`.


foxpatch-aleph (approved):
A careful, well-documented quality refactor that unifies the allocation plane's refusal vocabulary into one Refusal type and one retention rule, and fixes two real defects: reclaimable-marked directories being credited twice in disk admission (free space and discount), and node-confusion boundary errors being published as hypervisor refusals. I verified both fixes are sound against the actual code: _available_disk_bytes adds reclaimable_bytes to free space so the discount skip is the correct arithmetic, and the create path is unaffected because creating()/adopt() clears the marker before admission runs. The positional vms[i] keying cannot collide with hash keys (ItemHash validation rejects that form) and eliminates both the collapse of multiple no-hash entries into a single "None" key and the echo of unbounded request text. mypy passes on all changed modules (remaining errors are in the untouched pubsub.py), all PR-touched test suites pass (336 tests), and the full-suite failures I reproduced are sandbox-environmental in files this PR does not touch.

**`src/aleph/vm/agent/allocation/verdict.py`** (line 164): This is a breaking wire-format change for entries with unreadable hashes: the response key changes from echoing the raw string the push sent (previously `str(raw_hash)`), to the positional `vms[i]`. The old shape was broken two ways (multiple no-hash entries collapsed into one `"None"` key, and it echoed unbounded request text back), so the change is an improvement — but if the official scheduler correlates rejected keys against what it sent, it needs the new positional scheme. Worth a sentence in the release notes.

**`src/aleph/vm/agent/views/allocations_v2.py`** (line 127): Nit: `results` mixes keys here — the raw `key` (an `ItemHash` for hash-keyed rejections) while the candidate results a few lines below use `str(vm_hash)`. Since `ItemHash` is a `str` subclass the serialized output is identical, so this is purely cosmetic, but `str(key)` would keep the two halves reading the same.

foxpatch-aleph (commented):

foxpatch-aleph (approved):
A clean, well-factored quality pass over the allocation plane. The new refusal.py gives the whole agent one closed vocabulary of failure codes with a single published sentence per code, and the module split keeps it a leaf so capacity.py can depend on it without pulling in the create path (no import cycles, confirmed at runtime). Both defect fixes are correct: existing_volume_files now skips directories still carrying the .reclaimable marker, ending the double credit (free space + discount) that let plan simulation promise room the node lacks; and the four boundary error codes are no longer misreported as hypervisor refusals. The positional vms[i] keys fix both the str(None) key collision and the unbounded echo of request text in responses, and the unauthenticated executions list now publishes only codes plus fixed sentences, with exception details confined to the log. Dead surface (first_failed_at, the ignored persistent parameter) is genuinely dead, and the inherited PlanAdmission/RecordLookup Protocols prevent signature drift. Tests are thorough and pass: 181 across capacity/verdict/failures/teardown, plus the reconciler, allocations_v2, and executions_list suites; the only failures encountered were environmental (stale aleph_message wheel, /var/cache/aleph permissions), unrelated to the PR.

**`src/aleph/vm/agent/views/allocations_v2.py`** (line 130): Minor nit, not blocking: this MESSAGE_REQUIRED branch is currently unreachable. build_plan (verdict.py:177) only places verified entries into plan.entries and rejects the rest, so planned.verified is never None here. Fine as a defensive guard with a clear comment, but if the plan shape ever changes, note that this branch would silently shadow the more specific INVALID_MESSAGE refusal _read_plan already put in `rejected` for the same entry.

foxpatch-aleph (changes requested):
The refactor itself is correct and well-executed: the unified Refusal vocabulary with one wire shape (as_dict), the reclaimable-marker double-count fix in existing_volume_files (verified arithmetically sound for both the create path, where the creating() guard adopts the directory before admission, and the plan simulation), the four boundary error codes now publishing internal rather than supervisor_error, the retention_reason unification that fixes a real latent disagreement for v-programs, and the vms[i] positional keys that stop both the collapse of multiple unhashable entries into one 'None' key and the echo of unbounded request text. The one genuine defect is a missed test migration: tests/supervisor/test_allocation_verdict.py:610 still subscripts verdict.rejected[HASH_A]["code"]", but this PR changed PlanVerdict.rejected from dict values to Refusal objects, so the test fails on the PR head with TypeError: 'Refusal' object is not subscriptable — contradicting the PR's own test plan of a single known journald failure. Fixing that one line (e.g. verdict.rejected[HASH_A].code is AllocationFailureCode.INSUFFICIENT_CAPACITY) would make this an approval.

**`tests/supervisor/test_allocation_verdict.py`** (line 610): This assertion was missed in the migration from dict-shaped refusals to the Refusal dataclass: `verdict.rejected[HASH_A]["code"]` now raises `TypeError: 'Refusal' object is not subscriptable`, so `test_a_newcomer_is_refused_the_room_a_stopped_vm_holds` fails on the PR head (confirmed by running it in isolation). It should read `verdict.rejected[HASH_A].code is AllocationFailureCode.INSUFFICIENT_CAPACITY`, matching the other migrations in this file. As shipped this contradicts the PR's stated test plan of only the journald failure.

foxpatch-aleph (approved):
A careful, well-tested quality refactor that fixes two real defects and removes a genuinely dead parameter. I verified the double-credit analysis in existing_volume_files against _available_disk_bytes (marked directories' bytes do count as free, and the create path clears the marker via creating()/adopt() before admission runs), confirmed the four boundary-code remappings to internal are correct, confirmed the positional refusal keys cannot collide with hash keys (ItemHash rejects 'vms[3]'), and confirmed the removed persistent parameter was already ignored in create_vm_execution's body, making its removal behavior-neutral. All 236 tests in the five touched allocation test files pass, migration tests are 24/24, and mypy is clean on the changed modules; the remaining suite failures in this sandbox are environment-caused (stale aleph_message version and root-only /var/cache/aleph) in code this PR does not touch. The retention_reason unification fixes a real v-program inconsistency, and the new Refusal/Refusals types make the wire shape impossible to drift. One stale docstring nit below.

**`tests/supervisor/test_allocation_verdict.py`** (line 197): Stale docstring: it still says \"_retention_reason has to stay a mirror of is_removable_by_allocation\" and mentions the operator_policy catch-all, but _retention_reason was deleted in this PR and merged into retention_reason (teardown.py), so the hazard the docstring describes no longer exists. Worth rewording to describe what the test now pins (each retained VM is reported under the reason teardown.retention_reason returns) rather than a mirror-drift that can no longer happen.

foxpatch-aleph (commented):

foxpatch-aleph (approved):
This PR is a well-executed consolidation: one refusal vocabulary (allocation/refusal.py) replaces four hand-written dict shapes, and two real defects are fixed with tests — the reclaimable-marker double-credit in existing_volume_files and the mislabeling of node-confused supervisor codes as hypervisor refusals. I verified the reclaimable semantics against _available_disk_bytes (the double-credit was real), confirmed the retention_reason unification fixes a genuine disagreement (the old verdict-side copy lacked the is_vprogram branch, so a dropped v-program was reported retained while the loop tore it down), traced all imports for cycles (none), confirmed every enum member has a published sentence and every rejection message comes from a closed set (verify.py reasons), and confirmed clean removals (first_failed_at, raw_hash, the persistent parameter, AdmissionVerdict's old shape). The 236 pure-logic tests across verdict, teardown, failures, reconciler and capacity all pass locally; the aiohttp-client test failures I observed are environmental (sandbox has pytest-asyncio 0.20.3 vs 0.23.7 pinned), not related to this change.

**`src/aleph/vm/agent/allocation/verdict.py`** (line 164): Behaviour change worth flagging for the release notes: the key for an entry with no usable hash changes from the echoed raw value (or the literal "None") to its position, vms[3]. A scheduler that correlates its rejections by echoing back the key it sent will see a wire-format change for invalid entries — strictly an improvement (duplicates were collapsed into one refusal before, and the raw string was unbounded request text), but it is visible to clients.

**`src/aleph/vm/agent/capacity.py`** (line 257): The per-directory skip is the right shape (the marker is written per namespace directory, and only that pool's bytes count as reclaimable), and checking the filename directly rather than read_marker() is fine here — unlike the reconciler, this path must not have the side effect of repairing/deleting a malformed marker. TOCTOU between the iterdir() listing and the per-entry checks is also harmless: a file deleted in between fails is_file() and is skipped, which errs conservative.

foxpatch-aleph (approved):
High-quality refactoring PR that fixes two real defects with regression tests and consolidates a fragmented refusal vocabulary into a single well-documented module. I verified the reclaimable-marker double-counting fix (including that the create path keeps its discount because the creating() guard adopts before admission), the positional vms[N] keying (ItemHash validation makes key collisions impossible; also stops echoing unbounded request text), the boundary-codes-to-internal remapping, the retention_reason consolidation, the dead first_failed_at field, and the persistent-parameter removal across all call sites. All 270+ tests in the affected modules pass; the only failures in this environment are pre-existing (stale aleph_message package, /var/cache permissions, journald), and the mypy/ruff findings are version drift on lines untouched by the diff.

**`src/aleph/vm/agent/allocation/refusal.py`** (line 62): Minor observation: the INVALID_MESSAGE entry here is never published via Refusal.for_code, since an invalid_message refusal always carries the verifier's own fixed reason instead (verdict.py:164, :172). It exists so public_failure_message works for every enum member — harmless, but a one-line note in the docstring would save a future reader the search.

**`src/aleph/vm/agent/views/allocations_v2.py`** (line 127): Worth noting for release notes if this API has any external consumers: the advisory capacity_check response keys change from the old str(None)/echoed-junk keys to positional vms[N] keys. Deliberate and an improvement, just a wire-shape change for advisory clients.

foxpatch-aleph (approved):
This PR is a quality refactor that also fixes two real defects, and every claim in its description checks out. The double-credit fix in existing_volume_files is correct: _available_disk_bytes() already adds reclaimable_bytes() to free space, so a directory still carrying its .reclaimable marker was credited as both free space and a discount, letting the plan simulation promise room the node does not have; the create path is unaffected because the creating guard adopts and clears markers before admission. The four boundary ErrorCode -> internal remappings are sound (all 11 ErrorCode values accounted for, with PORT_UNAVAILABLE deliberately kept as supervisor_error), and the vms[i] positional keys are collision-free since ItemHash rejects that format. The retention_reason unification removes a genuine past disagreement for v-programs, the dead first_failed_at and persistent surfaces are fully removed with no stragglers, and the behavior changes (five refusal sentences, positional rejection keys, in-flight-teardown refusal) are documented, reasoned, and pinned by tests. All PR-relevant tests pass locally (93 + 179 + 24 migration); the 30 failures in the full local run are environment artifacts (stale aleph_message install missing TdxRegisters/ConfidentialGpuRequirement/gpu, and /var/cache/aleph sandbox permissions) unrelated to the diff.

**`src/aleph/vm/agent/capacity.py`** (line 257): Verified the double-credit reasoning: _available_disk_bytes() adds reclaimable_bytes() to the pools' own free figure, so skipping the whole marker-carrying directory here is the right arithmetic, and the create path keeps its discount because the creating guard (vm/reconciler.py) adopts and clears markers before admission runs. The new test failing on the base (disk_mib 0 == 20000) confirms this was a real hole.

**`src/aleph/vm/agent/allocation/verdict.py`** (line 164): The vms[{index}] keying is collision-free with hash keys: ItemHash("vms[3]") raises UnknownHashError, so the later "vm_hash in rejected" duplicate check keeps its exact meaning, and the change both stops unbounded request text being echoed back and un-collapses the entries that used to all answer under the single key "None". Worth being aware that this is a wire-format change for any scheduler that read the old str(raw) keys, but it is a strict improvement and documented in the PR description.

**`src/aleph/vm/agent/allocation/failures.py`** (line 38): Agreed that all four codes mean "this node is confused" rather than "the hypervisor declined"; VM_ALREADY_EXISTS is the only one with a normal producer (a lost create race, which create_vm_execution handles as a no-op), and internal is the honest published code for it. If a scheduler ever needs to distinguish races from bugs this would be the place to split, but nothing today does.

**`src/aleph/vm/agent/allocation/refusal.py`** (line 74): Keeping public_failure_message alongside Refusal.for_code gives two ways to reach the same sentence; not a problem since both read _PUBLIC_MESSAGES, but if anything ever needs the sentence it could go through Refusal.for_code(code).message to leave one path. Purely informational.

foxpatch-aleph (approved):
This PR is well-executed: the refusal consolidation into refusal.py is internally consistent (no stale callers of the old AdmissionVerdict shape, public_failure_message, or raw_hash anywhere in src), and both defect fixes are correct. I verified that the reclaimable-marker fix is sound: _available_disk_bytes adds reclaimable_bytes() to pool free space, so a marked directory's bytes are already counted as free and discounting them would double-credit them; the create path keeps its discount because creating() adopts the directory and clears its marker before check_message runs. The positional rejection keys ("vms[N]") cannot collide with real hashes (ItemHash("vms[3]") raises UnknownHashError) and the invalid_message message text comes from verify_entry's closed set of fixed reasons, so nothing unbounded is echoed back. The retention_reason refactor preserves is_removable_by_allocation's semantics exactly, and the boundary-code reclassification is deliberate, documented, and tested. All allocation/capacity/reconciler/teardown tests pass (236), tests/migration passes (24), mypy is clean on the allocation modules, and ruff format is clean; the other failures in this environment stem from an outdated aleph_message 1.2.1-dev install versus the pinned 1.5, not from the PR. Only one trivial doc nit found.

**`src/aleph/vm/agent/capacity.py`** (line 246): Tiny nit: the closing sentence still says "Symlinks and the marker are skipped", but with the new early `continue` the marker now skips the whole directory rather than being skipped as an entry. The paragraph above describes the new behavior correctly, so this is only a leftover phrase in the last line.

foxpatch-aleph (approved):
High-quality refactor that unifies the allocation refusal vocabulary into one Refusal type with one serialization path, and fixes two real defects: the reclaimable-marker double-credit in existing_volume_files (verified against _available_disk_bytes, which counts marked bytes as free) and the boundary error codes that misreported node bugs as hypervisor refusals. The vms[index] re-keying fixes a genuine collapse bug (three hash-less entries answered as one "None") and stops echoing unbounded request text into responses. All affected test suites pass; no stale references to the removed shapes remain; the import split's rationale (avoiding capacity->failures->run->capacity cycle) checks out. The failures I could reproduce locally are environment issues (aleph-message version lacking ConfidentialGpuRequirement/gpu, /var/cache/aleph permissions) in code paths untouched by the diff. Two minor nits below, neither blocking.

**`tests/supervisor/test_agent_capacity.py`** (line 1054): This assertion is now vacuous: the message is the fixed published sentence ("No available GPU matches this request"), which can never contain "evil". The property it pinned — that the requested device_id must not be echoed back — is now structurally guaranteed by Refusal.for_code, so the test could assert refusal.message equals the published sentence, or simply be dropped to the code assertion above.

**`src/aleph/vm/agent/allocation/teardown.py`** (line 23): retention_reason has 7 return statements, which trips PLR0911 under the repo's configured ruff rules (the rest of the file set has ~126 such warnings, so clearly not gating — just noting it adds one more than the two functions it replaces).

PR #1238, opened 2026-09-10
fix(vprogram): measure runtimes for EPYC-Genoa first, EPYC-v4 as fallback

odesenfans:
## Problem

A V-PROGRAM user reported llama.cpp running at 4.2 tok/s on a V-PROGRAM against 19.4 tok/s on a legacy confidential instance, same model (Hermes-3-8B Q5_K_M), same flags, same 16 vCPU. The guest reports `AMD EPYC-v4 Processor` with no AVX-512 flags.

QEMU's `EPYC-v4` is the **Naples** model (family 23, model 1), despite the name. Genoa is `EPYC-Genoa` (family 25, model 17). Verified with `query-cpu-model-expansion` on QEMU 10.2.1:

| model | family | model | avx512f |
|---|---|---|---|
| EPYC-v4 | 23 | 1 | no |
| EPYC-Genoa | 25 | 17 | yes |

The legacy confidential-instance path launches with `-cpu host`, which is why that side sees the full feature set. The SNP path pins the model from the runtime manifest, and every published manifest (exec-1.1, compose-1.1, snp-1.0) lists `["EPYC-v4"]` from `DEFAULT_CPU_MODELS`. The flake comment describing EPYC-v4 as "Genoa" is what let this through.

## Change

`DEFAULT_CPU_MODELS = ["EPYC-Genoa", "EPYC-v4"]`, Genoa first. #1145 already made the launch model message-driven: the agent launches the first listed model its QEMU supports, and the scheduler places a V-PROGRAM on any node matching at least one measurement. So Genoa-capable hosts get AVX-512 and older hosts fall back to EPYC-v4; nobody is stranded. All seven SEV-SNP CRNs live on the network (2.0.1 / 2.0.2, both contain #1145) advertise `EPYC-Genoa`.

`EPYC-Genoa` and `EPYC-Genoa-v2` share family/model/stepping, so the alias resolving to a different version per QEMU release does not move the launch measurement. The aleph-rs CLI (`sev` crate) and the nix `sev-snp-measure` both already accept `EPYC-Genoa`.

Also corrects the flake comment and pins the new list in the bundle test.

## Rollout

The republish is manifest-only: bundle contents do not depend on the CPU model, and clients compute one measurement per listed model at launch. New exec-1.2 / compose-1.2 / snp-1.1 manifests pointing at the existing bundle refs, then repoint the vm-images aggregate (keeping the old entries for running VMs). Cost: one extra client-side measurement per launch.

## Tests

`tests/vprogram/`, `test_vcpu_select.py`, `test_vprogram_launch.py`, `test_snp_instance_launch.py`: 134 passed locally. `ruff format` clean.

### DIFF
--- a/nix/flake.nix
+++ b/nix/flake.nix
@@ -267,7 +267,8 @@
 
       # Parameterized SEV-SNP launch measurement builder.
       # vcpus:            number of vCPUs (affects the launch measurement)
-      # vcpuType:         QEMU CPU model ("EPYC-v4" for Genoa, "EPYC-v3" for Milan)
+      # vcpuType:         QEMU CPU model. Note "EPYC-v4" is QEMU's Naples model
+      #   (family 23), not Genoa; Genoa is "EPYC-Genoa" (family 25, model 17).
       # workloadRoothash: when null (default), the cmdline is the
       #   platform-only form `...roothash=<platform>` (workload-less parity,
       #   what test_vm_snp and the baked `measurement` below expect). When
--- a/src/aleph/vm/vprogram/bundle.py
+++ b/src/aleph/vm/vprogram/bundle.py
@@ -204,7 +204,15 @@ CMDLINE_TEMPLATE_EXEC_V1 = (
     " workload_roothash={workload_roothash}"
     " verified_volumes={verified_volumes}"
 )
-DEFAULT_CPU_MODELS = ["EPYC-v4"]
+# QEMU CPU models the published runtimes are measured for, in preference
+# order: the CRN launches the first one its QEMU can run. Despite the name,
+# "EPYC-v4" is QEMU's Naples model (family 23, model 1): no AVX-512, so
+# vector-heavy workloads such as llama.cpp fall back to AVX2 and run several
+# times slower than on the host silicon. "EPYC-Genoa" (family 25, model 17)
+# exposes AVX-512 and is what every SEV-SNP CRN on the network advertises;
+# "EPYC-v4" stays as the fallback so a Milan or Rome host is never stranded.
+# Each entry costs one client-side measurement per launch; keep the list short.
+DEFAULT_CPU_MODELS = ["EPYC-Genoa", "EPYC-v4"]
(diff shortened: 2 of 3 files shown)


foxpatch-aleph (approved):
Correct, minimal fix. The change is a data-only edit to DEFAULT_CPU_MODELS, and the runtime already supports ordered multi-model lists: requested_vcpu_types preserves manifest order and select_snp_vcpu_type picks the first model the host can launch, failing closed on empty intersection (src/aleph/vm/agent/vcpu_select.py:45). The legacy single-model fixtures still exercise the EPYC-v4 fallback and default paths, the baked measurement.hex staying EPYC-v4 is consistent since EPYC-v4 remains listed and the sidecar is informational, and the new bundle test pins the expected list. Two documentation nits: two remaining "EPYC-v4 (Genoa)" comments in nix/flake.nix repeat the misconception this PR corrects, and docs/architecture/confidential.md still describes the SNP launch digest as pinning a fixed -cpu EPYC-v4, which is stale since #1145. Verified locally that tests/vprogram/ and tests/supervisor/test_vcpu_select.py pass (82 tests); the launch-suite failures in this sandbox are /var/lib/aleph permission errors unrelated to the diff.

**`nix/flake.nix`** (line 316): The corrected comment at the vcpuType parameter (lines 270-271) fixes the misleading "(Genoa)" annotation, but two more instances of it survive in this same file: here ("Default measurement: 2 vCPUs, EPYC-v4 (Genoa)") and at line 332 ("Default compose measurement: 2 vCPUs, EPYC-v4 (Genoa)"). Since this PR is specifically about correcting that misconception, these two should be updated too so the wrong annotation doesn't get copied into new call sites.

**`docs/architecture/confidential.md`** (line 76): This sentence still says the SNP launch digest "pins the fixed -cpu EPYC-v4 instead" — stale since #1145 made the launch CPU model message-driven, and now doubly misleading given this PR switches the published runtimes to EPYC-Genoa first. Worth a follow-up (or folding in here) since the doc is about exactly this launch path.

PR #1239, opened 2026-09-10
fix(agent): rebuild a crashed VM without re-admitting it against today's headroom

odesenfans:
## Summary
A VM this node holds and the supervisor reports FAILED is rebuilt by the
convergence loop, and the rebuild was admitted against the node's current
headroom twice. `compute_verdict` sized the listed hash through `simulate` and
could refuse it `insufficient_capacity`, after which #1209's refusal
protection keeps the VM but the loop never builds it back; and the create path
ran the full `check_message`. On a node that sits over its caps for historical
reasons, which this project has, either one strands the VM: one crash and it
never comes back.

The VM's memory was never released. Its registry record is still there and
still counted in `_committed_resources`, so a rebuild is not new load: it is a
reservation the node is already holding being used again.

The answer side now acknowledges a listed hash this node holds a record for,
whatever its status, and never sizes it. Together with #1217 that makes the
whole rule "only a hash this node holds nothing for is a candidate". The
exception stays the VM whose teardown is already running (#1225): that one is
going away with its disks and its record, so this node really would have to
find room for it from nothing.

The create side gains `check_recreate`, which judges the disk and nothing
else, with the same held-volume discount `check_message` applies, so a message
that grew a volume or one whose files a reinstall purged still has to find
room for what is new. It is reached only when the caller says it may be
rebuilding and the registry turns out to hold the hash before the create
records anything of its own: the convergence loop and the operator's reboot
and reinstall say so, the legacy allocation route and the on-demand program
paths do not, so `check_message`'s enforced path is unchanged for a newcomer.

What bounds a rebuild the node cannot really run is #1214's crash-loop
backoff, not admission: the VM dies again and the next attempt waits longer.

Stacked on #1237.

## Changes
- `allocation/verdict.py`: a plan entry the supervisor lists and this node
  holds a registry record for is answered `unchanged` in any state, and is not
  handed to `simulate`. A hash with no record is sized as before.
- `capacity.py`: new `check_recreate(content, vm_hash=...)`, disk only, with
  the held-volume discount. The disk half of `_check_against` moves into
  `_disk_errors`, shared by both, so the two paths cannot come to different
  answers about the same volumes.
- `run.py`: `create_vm_execution` takes `recreate`, reads whether the registry
  already held the hash before any branch records one, and routes admission
  through `_admit_create`. `create_vm_execution_or_raise_http_error` and
  `start_persistent_vm` pass it through.
- `allocation/reconciler.py`: `_start_one` marks its start as a rebuild, and
  its docstring says the backoff ladder is the safety net.
- `views/operator.py`: the reboot and reinstall rebuilds do the same; both
  deliberately keep the record for the rebuild.

## Test plan
- `tests/supervisor/test_allocation_verdict.py`: a listed dead hash with a
  record here is `unchanged`, is not refused and reaches `simulate` with no
  candidates; one with no record is still sized and accepted. The recreate
  fixtures that expected a dead VM to be judged as a candidate now expect it
  to be acknowledged, and the two refusal-carry tests use a hash this node
  holds nothing for, since that is the only kind the answer can still refuse.
- `tests/supervisor/test_agent_capacity.py`: `check_recreate` admits on a node
  whose records already exceed its memory bucket while `check_message` on the
  same message is refused; it still refuses disk it cannot find room for; it
  discounts the volumes the VM already holds.
- `tests/supervisor/test_create_path_disk_admission.py`: a rebuild of a
  recorded hash goes through `check_recreate`, a first create and a caller
  that does not claim a rebuild go through `check_message`, and against the
  real capacity manager on an over-committed node the rebuild gets past
  admission while a first create is refused before the download.
- `tests/supervisor/test_allocation_reconciler.py`: the loop's own starts are
  marked as rebuilds.
- Checks: `ruff format --diff` (exit 0), `isort --check-only --profile black`
  (exit 0), `mypy src/aleph/vm/ tests/` from the worktree root (Success, exit
  0). The test suites are left to CI on this round.


foxpatch-aleph (approved):
This PR is well-engineered and its core premise holds everywhere it is applied: I verified that RECREATE retires keep the registry record (retire.py returns before registry.forget), that the rebuild flag is read before any branch records (run.py:528), that the disk rule is genuinely shared between check_recreate and check_message via _disk_errors and discounted_disk, and that removing_now still takes precedence over held so a VM in teardown is not silently acknowledged. All 221 tests across the four touched suites pass locally and mypy is clean on the changed capacity code. The observations below are non-blocking design notes, not defects: the memory-growth gap is a documented trade-off bounded by the crash-loop backoff, and the legacy-route omission is an explicit scope decision, though both are worth a follow-up.

**`src/aleph/vm/agent/capacity.py`** (line 462): check_recreate never compares the message against what the record actually committed. The record's commitment was made from the message at first admission; if the owner amends the message to request more memory (update-watcher retires RECREATE and the loop rebuilds with the grown message), the rebuild IS partly new load, yet nothing judges the delta. The crash-loop backoff only bounds a VM that fails to boot, not one that boots and over-commits the node. Since the record conveniently stores its message, a cheap tightening would be: when the new content's memory/vCPUs exceed what record.message committed, fall back to the full check (or judge just the delta). This preserves the fix for the unchanged-message case, which is the crash-recovery case the PR is about. Non-blocking — the trade-off is documented in the docstring and the PR description.

**`src/aleph/vm/agent/run.py`** (line 528): The rebuild flag is read before load_updated_message's await, so a concurrent teardown that forgets the record during the download leaves admission running check_recreate for a VM that is no longer recorded. The window is narrow (the reconciler serializes per hash, but the operator paths call create_vm_execution_or_raise_http_error outside that lock) and the direction of error is the permissive one, so this is acceptable — but worth being aware that the flag is advisory, not a guarantee the record survives to admission.

**`src/aleph/vm/agent/views/__init__.py`** (line 711): Follow-up candidate, not for this PR: the legacy allocation endpoints call start_persistent_vm without recreate=True, so a crash-recovery rebuild through the legacy route still goes through check_message and can be refused on a node over its caps — the exact stranding this PR fixes, surviving only on that route. Since create_vm_execution gates the skip on registry.get(vm_hash) is not None anyway, passing recreate=True here would be equally safe and would close the gap. The on-demand program path (_ensure_program_vm) has the same shape.

**`tests/supervisor/test_agent_capacity.py`** (line 1640): Nice test. One small hardening suggestion: assert the refusal reason from check_message is about memory (e.g. check \"instance bucket\" or \"Memory\" in the error text) so the test fails informatively if _patch_host or the reserved-memory patches ever stop producing a memory-bucket refusal and the failure comes from disk instead — currently any InsufficientResourcesError from check_message makes the test pass.

foxpatch-aleph (approved):
This PR fixes a real stranding bug: a crashed VM's memory was charged twice at rebuild (once via its still-held registry record in _committed_resources, once via the request), so an over-capacity node refused every rebuild and refusal protection then kept the VM forever. The fix is sound on both sides. The verdict side now reports any hash the registry holds a record for as unchanged, which cleanly extends coverage to the FAILED status that LIVE_STATUSES/STOPPED_STATUSES missed, while still letting removing_now take precedence so mid-teardown VMs are sized. The create side gains check_recreate, which skips only the memory/vCPU checks (whose reservation retire_vm(RECREATE) provably keeps) while still judging disk by the same discounted_disk rule, so grown or added volumes must fit. I verified the rebuild flag is read before any branch records, that the _disk_errors extraction preserves error ordering and the unconditional availability call, and that the crash-loop backoff ladder -- not admission -- bounds a hopeless rebuild. Tests are thorough and all pass here. Remaining observations are documented design tradeoffs, not defects: the scheduler is never told a rebuild keeps failing on a tight node, and the legacy allocation route still has the original double-charge behavior.

**`src/aleph/vm/agent/run.py`** (line 528): This read races with any path that drops the record between here and _admit_create (terminal message, unpaid eviction, reconciler reap): rebuild would be stale, so admission would skip the memory/vCPU checks for a VM whose committed memory has just been given back and will be re-committed -- a one-shot over-admission on a node that already sits over its caps. The per-hash creating() guard and the reconciler's serialization keep the window tiny, and nodes under their caps cannot over-commit at all, so this is an observation rather than a blocker. A comment noting the read is unsynchronized by design might help future readers.

**`src/aleph/vm/agent/allocation/verdict.py`** (line 271): Agreed with the rule. One consequence worth stating explicitly in the docstring: a crash-looping VM on a tight node is answered unchanged forever, so the scheduler never learns that the rebuild is not working and cannot react (e.g. by migrating the allocation elsewhere). The backoff ladder bounds the local cost, and reporting a VM we hold any other way would reintroduce the double-charge, so this is the right call -- but it is a real loss of signal, not just an accounting detail.

**`src/aleph/vm/agent/views/__init__.py`** (line 711): These legacy start_persistent_vm calls do not pass recreate, so a persistent VM re-pushed through the old allocation route keeps the original double-charge behavior and can still be refused on an over-cap node. That matches the PR's stated scope boundary, but a short TODO or follow-up ticket would help make the gap deliberate rather than overlooked.

**`src/aleph/vm/agent/capacity.py`** (line 488): Good reuse: routing through declared_volumes/discounted_disk means a volume the message grew, one it added, or one whose files a reinstall purged still has to fit, so check_recreate is the check_message disk rule rather than a parallel copy. The _disk_errors extraction is behavior-preserving (order kept, _available_disk_bytes still called unconditionally). Minor nit: the InsufficientResourcesError payload only carries disk_mib, which matches what is actually judged here, but callers keying on the full required shape from check_message may want to know.

foxpatch-aleph (approved):
Well-designed and consistently implemented fix for a real stranding bug: a crashed VM's rebuild was double-charged against headroom its own registry record already commits. The verdict-side rule (a listed hash this node holds a record for is acknowledged, whatever its status), the new disk-only check_recreate sharing the same _disk_errors arithmetic as the enforced path, and the flag-plus-registry routing on the create side are all correct, and the #1225 removing_now exception is preserved. All 221 tests in the touched files pass locally; remaining failures are environmental (wrong aleph-message version, /var/cache permissions) and pre-date this PR. GPU admission is unaffected, RECREATE retires provably keep the record so the operator's recreate=True paths are sound, and no check is lost for first creates or unrecorded hashes.

**`src/aleph/vm/agent/allocation/verdict.py`** (line 271): Boundary note: the `held` rule only fires for a hash the supervisor lists. A hash with a phantom record that the supervisor does *not* list is still sized as a candidate and can be refused INSUFFICIENT_CAPACITY on an over-committed node, even though the enforced rebuild path (check_recreate) would admit it — the same stranding failure in the 'supervisor forgot the VM' shape, and then the advisory answer is stricter than the enforced one. Arguably the honest answer here, but a sentence in the comment saying why the un-listed shape stays a candidate would help future readers.

**`src/aleph/vm/agent/run.py`** (line 528): The flag is snapshotted here, but `load_updated_message` below can download for seconds: a concurrent GONE retire that forgets the record during that window leaves the create skipping memory admission where `check_message` would have judged it. Very narrow race, and the existing lost-create-race handling covers the concurrent-teardown case, but worth knowing the flag can be stale rather than merely conservative.

**`src/aleph/vm/agent/run.py`** (line 481): Minor: `content` is unannotated here (`ExecutableContent`) where the rest of the module annotates everything.

**`tests/supervisor/test_allocation_reconciler.py`** (line 670): Typo in the test name: 'test_the_loops_starts...' should read 'the loop's'.

foxpatch-aleph (commented):
Correct and well-scoped fix for a real stranding bug: a crashed VM held by this node was rebuilt against headroom it didn't have, so on over-committed nodes #1209's refusal protection meant one crash stranded the VM for good. Acknowledging a listed hash this node holds a record for as `unchanged` is sound because `simulate` excludes each candidate's own record from committed sums, so the reservation is never freed by the rebuild. The new `check_recreate` path (disk-only, with held-volume discount) is correctly gated: the `recreate` flag is only set by the convergence loop and the operator's reboot/reinstall, and only after the registry confirms a record exists, so first-time creates keep the full enforced check. Migration and the legacy allocation route deliberately keep `check_message`, which is right. Tests are thorough — including an end-to-end reproduction against a real CapacityManager on an over-committed 64 GiB host — and all relevant suites pass locally. Two minor nits below, neither blocking.

**`src/aleph/vm/agent/run.py`** (line 481): Minor: `content` lacks a type annotation, while both `CapacityManager.check_message` and `check_recreate` annotate their first parameter as `ExecutableContent`. Adding it here would keep the three consistent and help IDEs confirm the routing.

**`src/aleph/vm/agent/run.py`** (line 1212): Observation: the `recreate` pass-through through `start_persistent_vm` itself is only exercised with `start_persistent_vm` monkeypatched (tests/supervisor/test_allocation_reconciler.py, tests/supervisor/test_executions_list_allocation.py), so a future edit dropping this argument would not be caught by a test. Not worth much — the reconciler and operator-path assertions would catch most regressions — but a one-line assertion on the forwarded kwargs would close the gap.

foxpatch-aleph (approved):
Correct fix for a real stranding bug: a crashed VM's rebuild was sized against headroom its own registry record had already subtracted, so on an over-committed node one crash left the VM kept-but-never-rebuilt. The three pieces (verdict acknowledges held hashes, check_recreate judges disk only with the same held-volume discount, create path detects a pre-existing record before recording) are consistent with each other, and the enforcement chain checks out: retire_vm(RECREATE) keeps the record so the rebuild detection works, disk refusals still flow into the crash-loop backoff, and every other caller keeps the full check_message path. Tests are thorough and pass locally, along with ruff format; the failures I saw in the wider suite are environmental (root-only cache dirs, a stale aleph-message wheel), not from this change. Two non-blocking observations below, plus a minor annotation nit.

**`src/aleph/vm/agent/allocation/verdict.py`** (line 311): The stated rule — "only a hash this node holds nothing for is a candidate" — is only implemented for supervisor-*listed* hashes, because the `held` check lives in the `known.items()` loop. A planned hash the registry holds a record for but the supervisor no longer lists (state loss on the supervisor, or the record kept by a RECREATE retire after a failed create) still falls through to `candidates` here, while the reconciler passes `recreate=True` even when `known.get(vm_hash)` is None and the create path skips its memory check. On an over-committed node the verdict can therefore refuse it and `narrow_plan` drops it, reproducing the stranding this PR fixes, in the unlisted case. The direction is conservative (advisory stricter than enforced), so no over-admission risk — but either the comment overstates the rule or this case is a missed spot. Consider acknowledging a held record here too, or narrowing the comment to listed hashes.

**`src/aleph/vm/agent/capacity.py`** (line 462): Worth a sentence in the docstring: because `create_vm_execution` calls `load_updated_message` and then re-records the message, a rebuild whose *updated* message grew memory/vCPUs is admitted with no memory check at all — disk is still judged for growth, memory is not. The record then commits the larger figure to every subsequent admission, and the crash-loop backoff is the only bound, which is the trade-off the PR description states; it would be good to have that asymmetry written down next to the "not new load" argument so a future reader doesn't take it for an oversight.

**`src/aleph/vm/agent/run.py`** (line 481): Minor nit: `content` is the only unannotated parameter in the new code (everywhere else in this module signatures are fully typed, e.g. `ExecutableContent`).

foxpatch-aleph (approved):
Well-reasoned and internally consistent fix for a genuine stranding bug: a held record's memory/vCPUs are already subtracted from headroom via _committed_resources, so re-charging a rebuild against that same headroom double-counts and refuses forever on an over-committed node. The rebuild flag is read before every registry.record() call site (essential, and correctly commented), the _disk_errors extraction is behavior-preserving (same checks, same error order), removing_now still correctly wins over the new held branch, and check_recreate reuses the exact same discounted_disk rule as check_message so the two paths cannot diverge. Test coverage is strong: I ran all touched suites locally and they pass (56+95+52+18); the operator-retire failures in this sandbox are purely environmental (aiohttp loop-plugin breakage, /var/cache/aleph permissions, a stale installed aleph_message version that also breaks untouched v-program tests). Non-blocking notes below concern a deliberate but real behavior change (the advisory verdict now never judges disk for a held rebuild, so a rebuild that cannot find room loops with backoff instead of being refused to the scheduler), unjudged memory growth on message amendment during rebuild, a missing type annotation, and the error payload reporting the discounted rather than declared disk figure.

**`src/aleph/vm/agent/allocation/verdict.py`** (line 271): Worth being aware of a deliberate asymmetry this introduces: with a held dead VM now answered `unchanged` without ever being sized, the advisory verdict no longer judges disk for a rebuild, while the enforced path (check_recreate) still does. A rebuild that genuinely cannot find room for a grown volume now loops on the node with backoff rather than being refused back to the scheduler for placement elsewhere. The PR description owns this (backoff as the bound), and the memory-stranding bug this fixes is far more common, so it's a reasonable trade-off — just flagging that disk, unlike memory, has no record committing it.

**`src/aleph/vm/agent/run.py`** (line 528): One consequence worth a comment here: `load_updated_message` may fetch an amended message requesting more memory/vCPUs than the record commits, and check_recreate will not judge the growth — the record is re-recorded with the new figures only after admission. Bounded by the supervisor's physical backstop and the crash-loop ladder, and consistent with the PR's stated philosophy, but it does mean an amended message can relaunch beyond the caps on rebuild.

**`src/aleph/vm/agent/run.py`** (line 481): Minor: `content` is the only untyped parameter on an otherwise fully annotated signature (the rest of the codebase annotates message params as ExecutableContent). Consider typing it for consistency.

**`src/aleph/vm/agent/capacity.py`** (line 500): Nit: `required={"disk_mib": disk.disk_mib}` reports the post-discount figure rather than the declared total, so an operator reading the refusal sees a smaller number than the message asks for. The detail string carries both required and available, so this is cosmetic, but `declared_volumes(content)` summed would be the more self-explanatory figure.

foxpatch-aleph (approved):
This PR fixes a real stranding bug (a crashed VM on an over-committed node is never rebuilt) with a coherent, well-scoped change across all three layers: the advisory verdict acknowledges a held hash in any state instead of sizing it, a new check_recreate judges only disk through the same _disk_errors helper the enforced path uses, and create_vm_execution routes on a caller-supplied flag gated by a registry lookup read before any branch records anything — so a first create or a caller that does not claim a rebuild keeps the full check_message path. The #1225 removing_now exception is preserved and tested, the crash-loop backoff is the documented bound on a rebuild the node cannot really run, and the new tests cover every boundary including a real-capacity-manager reproduction of the failure. All 221 tests in the target suites pass, and ruff/isort/mypy are clean on the changed lines. Only trivial nits remain.

**`src/aleph/vm/agent/run.py`** (line 481): Minor: `content` lacks a type annotation while the rest of the module annotates signatures — `def _admit_create(content: ExecutableContent, vm_hash: ItemHash, *, ...)`. mypy infers it fine, so this is cosmetic only.

**`src/aleph/vm/agent/run.py`** (line 528): Small race worth being aware of: `rebuild` is read here, then `load_updated_message` awaits, and each branch records before admission. If a concurrent teardown (e.g. an operator erase racing a reboot) drops the record inside that window, admission skips the memory checks for a VM whose reservation just disappeared. The window is narrow, the reconciler path serializes per hash, and the concurrent case is operator-initiated, so this is not a blocker — but a comment noting the read is best-effort against concurrent retires could help future readers.

**`tests/supervisor/test_allocation_reconciler.py`** (line 670): Typo in the test name: `test_the_loops_starts_are_marked_as_rebuilds_for_admission` → `test_the_loop_starts_are_marked_as_rebuilds_for_admission`.

foxpatch-aleph (approved):
This PR is correct and well-executed. The central invariant — that a RECREATE retire keeps the registry record, so a rebuild reuses a reservation the node never gave back — holds on every recreate-marked path, and the rebuild detector (`recreate and registry.get(...) is not None`, read before any branch records) is ordered correctly against all three record-then-admit branches. `check_recreate` shares the disk arithmetic with `check_message` via `_disk_errors`/`discounted_disk`, so the two admission paths cannot diverge, and the verdict change keeps the `removing_now` exception intact so a mid-teardown VM is still sized as a candidate. Test coverage is thorough (unit admission, verdict boundary, create-path routing against the real capacity manager, and the loop's flag), and all new tests pass; the only failures are pre-existing sandbox permission errors in tests this PR didn't touch. Remaining observations are documented design choices rather than defects: the verdict still sizes a held hash the supervisor no longer lists, the legacy allocation route deliberately keeps the full check, and an amended message that grew memory is rebuilt without re-judging memory.

**`src/aleph/vm/agent/allocation/verdict.py`** (line 271): The `held` check only reaches a hash the supervisor lists (`known`), so a planned VM with a registry record that the supervisor no longer lists — e.g. the agent DB survives a node reboot but supervisor state doesn't — is still sized as a candidate below. On a node whose out-of-plan records push it over caps, the verdict refuses it and the loop never rebuilds it, which is the same stranding this PR fixes for the listed case. `simulate` excludes all candidates' own records, so this only bites when the over-commitment comes from records outside the plan; may be worth a follow-up or at least a comment noting the boundary is "listed + held", not just "held".

**`src/aleph/vm/agent/views/__init__.py`** (line 1052): Non-blocking: the legacy allocation route deliberately doesn't set `recreate`, so a crashed VM rebuilt through `notify_allocation` on an over-committed node still goes through `check_message` and can be refused — the same one-crash stranding via a different door. Since `create_vm_execution` reads the registry itself, the flag is only an opt-in and this route could pass it with the same safety. The PR scopes this out explicitly, but it might be worth a follow-up issue so the fix isn't plan-governed nodes only.

**`src/aleph/vm/agent/capacity.py`** (line 462): One consequence worth being aware of: `check_recreate` doesn't judge memory at all, so a message amended to a larger memory/vCPU size than the recorded one is rebuilt without re-admission — the record only commits the old size. The crash-loop backoff bounds a rebuild that dies, but a silent amendment that fits-but-degrades on a full node would be admitted where a first create would be refused. Acceptable per the PR's stated trade-off; just flagging that the reservation being reused is the old message's size.

**`src/aleph/vm/agent/run.py`** (line 481): Minor nit: `content` is untyped here while the rest of the module is annotated (`ExecutableContent`). mypy infers it fine, but a type would match house style.

foxpatch-aleph (approved):
The change is correct, carefully scoped, and well tested. The admission relaxation requires both an opt-in caller flag and a pre-existing registry record, so a newcomer or an unvouching caller keeps the full check; the disk half is shared verbatim with check_message via _disk_errors, so the two paths cannot diverge; and the #1225 removing_now exception is preserved. The two-sided test coverage (verdict boundary, enforced-path routing, and end-to-end against a real CapacityManager on an over-committed node) pins the important invariants. Findings below are documentation nits and one non-blocking design observation, none of which warrant blocking the PR.

**`src/aleph/vm/agent/capacity.py`** (line 430): This docstring still opens with "The single admission path of every create path" — with `check_recreate` added, it no longer is. Worth a sentence distinguishing the two (message admission vs. disk-only rebuild admission) now that the create paths route through `_admit_create`.

**`src/aleph/vm/agent/run.py`** (line 619): Pre-existing comment, but now slightly stale for the rebuild route: "Disk, memory and vCPUs were judged before the download, by check_message above." On a rebuild, memory and vCPUs are deliberately not judged and the judge is `check_recreate`. Same for the `_retire_after_create_failure` docstring at line 453 ("admits through ``check_message`` with ``exclude_vm_hash`` set"), which no longer describes the rebuild retry.

**`src/aleph/vm/agent/allocation/verdict.py`** (line 271): One corner the new rule does not cover, worth knowing about: a hash this node holds a record for but the supervisor does *not* list at all still becomes a candidate (the `held` check only fires inside the `known` loop) and can still be refused for capacity on an over-cap node — the same stranding the PR fixes for the listed case, in the much narrower inconsistent-state window where a record exists but no supervisor listing does. Not a regression (this behaved the same before), and the enforced path would admit it via `check_recreate`; the loop just never gets there because the answer refused it.

**`src/aleph/vm/agent/run.py`** (line 481): `content` is unannotated while the rest of the file annotates parameters (`check_message(self, content: ExecutableContent, ...)`). Adding the `ExecutableContent` import would keep this consistent.

**`src/aleph/vm/agent/capacity.py`** (line 497): Design observation, not a blocker: on a node that genuinely can no longer run a held VM (host shrank below the VM's memory), the VM now crash-loops under the backoff ladder and the scheduler is never told to move it — before, the answer could refuse it and let the scheduler relocate it. The PR description makes this tradeoff explicitly (backoff is the bound, memory was committed anyway), so this is noted for the record only.

**`tests/supervisor/test_allocation_reconciler.py`** (line 670): Small nit: `test_the_loops_starts_...` reads awkwardly; `test_the_loop_marks_starts_as_rebuilds_for_admission` would match the file's phrasing conventions. Also, no test pins the held-dead-VM + `removing_now` combination (that a record held while teardown runs stays a candidate); the semantics are clear from the condition ordering but a pinning test would match the thoroughness of the rest of this suite.

PR #1240, opened 2026-09-10
fix(daemon): read a confidential VM's cards again before starting it

odesenfans:
### Summary

A confidential VM's GPU is read once, by the create gate, and that reading
does not survive the guest. When the VM is stopped, `forget_cc_modes` drops
the cache entry, and rightly so: once QEMU is gone the card is idle hardware
an operator can re-mode with NVIDIA's tool. But `StartVm` brought the guest
back without reading the card again, and the next refresh sweep re-seeded it
CC-on purely from the fact that a live SNP VM was holding it. A card
switched out of CC mode while its VM was down would therefore carry a
confidential guest the host had never checked, and `GetHostInfo` would go on
advertising "on" about a card nobody had read since the guest left. The
guest's own boot-time GPU attestation catches this for the owner, but the
daemon should refuse the start the way the gate refuses the create.

### Changes

- The create gate's per-card rule (inventory membership first, then the
  probe, then the cache write, then the `On` requirement) moves into
  `require_gpu_cc_mode`, so the two admissions cannot drift. Error wording
  and check order are unchanged.
- `start_vm_execution` runs that rule on every card of a VM whose config is
  SNP, before the stop stamps are cleared and before the tap, the nftables
  rules, the DHCP server and the unit restart. A refusal returns the gate's
  error and leaves the VM reporting `STOPPED`, with nothing stood back up. A
  plain passthrough VM is not gated, matching the create.
- No world write lock is taken on the start path: the card is already
  claimed by this VM's config, the caller holds the VM's lock, no create can
  take an attached card and the refresh sweep skips it. The probe's resume
  wait of up to 200 ms per suspended card therefore sits under the VM lock
  alone.
- The sweep's comment in `refresh_cc_modes_with` now says the start is what
  re-reads the card and seeds the cache, and that the sweep's seed only
  restates an answer the start already took.

### Test plan

- `starting_a_confidential_vm_refuses_a_card_re_moded_while_it_was_down`:
  the start is refused with the gate's wording naming the card, the
  inventory carries the fresh `off` rather than the create's `on`, the VM
  still reports `STOPPED`, and no DHCP server comes back up.
- `starting_a_confidential_vm_seeds_its_card_from_the_fresh_reading`: the
  card still answers `on`, the start proceeds, and the inventory says `on`
  at once rather than after a refresh pass.
- `starting_a_plain_vm_holding_a_card_reads_nothing`: a non-SNP VM holding
  the same card starts with the card answering `off` and caches nothing.
- The daemon test suite, `cargo fmt --all -- --check` and
  `cargo clippy --locked --all-targets -p supervisor-daemon -- -D warnings`
  on CI.


foxpatch-aleph (approved):
Correct and well-documented fix. The extraction of the per-card admission rule into require_gpu_cc_mode is behavior-preserving (identical check order — inventory membership first, probe, unconditional cache write, On requirement — and identical error wording), and placing the gate at the top of start_vm_execution means a refusal leaves the stop stamps and all teardown work untouched. I verified the no-world-lock concurrency argument: start_vm holds the per-VM lock, validate_spec_gpus rejects cards claimed by any entry's config (stopped or not) so no create can race the probe, and the refresh sweep skips attached cards and seeds with or_insert so it cannot overwrite the start's fresh reading; during the gate the entry still carries its stop stamps so attached_gpus does not vouch for the card. The three new tests are meaningful and mutually exclusive, and the thread-local switchable probe is sound (CcMode is Copy+Send, synchronous test thread). Two non-blocking observations: (1) the reboot path also releases and re-hands the card yet remains ungated and never forgets the cache, so the stated invariant 'a card enters an SEV-SNP VM only after the create gate or the start has read CC-on' still has a narrow hole through reboot_vm's RestartUnit window — pre-existing, not a regression, possibly worth a follow-up; (2) trivial nit: the block braces around the single cache-insert statement in require_gpu_cc_mode are a leftover from the inlined site.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 1128): Verified the concurrency claim in the comment: start_vm holds the per-VM lock, validate_spec_gpus (lifecycle.rs:1799-1807) rejects any card claimed by an entry's config including a stopped one, so no create can take this card mid-probe, and refresh_cc_modes_with skips attached cards while its known_cc_on seed uses or_insert_with and so cannot overwrite the start's fresh reading. During the gate the entry still carries its stop stamps, so attached_gpus does not vouch "on" for the card either. The argument holds.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 1347): Non-blocking observation: reboot_vm is the third admission that hands a confidential guest its card, and it stays ungated — RestartUnit releases QEMU and re-acquires it without re-reading the card, and reboot never calls forget_cc_modes, so the sweep keeps vouching "on" through the window. The window is narrow (a systemd restart) versus the unbounded stop-to-start window this PR fixes and the guest's own boot attestation covers the owner, so this is pre-existing rather than a regression — but the service.rs invariant "a card enters an SEV-SNP VM only after the create gate or the start has read CC-on from it" still has a hole here. Worth a follow-up or a comment acknowledging it.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 2255): Trivial nit: this { } block is a leftover from the inlined site in snp_config_slice_with and now encloses a single statement; it can go.

foxpatch-aleph (approved):
A correct, well-scoped fail-closed fix: the per-card CC admission rule is extracted verbatim into require_gpu_cc_mode and applied on the start path for SNP configs, before any stop-stamp clearing or irreversible step, so a refusal leaves the VM STOPPED with nothing stood back up. The lock-free reasoning (VM lock held, creates can't claim attached cards, sweep skips claimed cards, probe on the blocking pool) checks out against the code. Tests cover the refusal, the successful re-seed, and the plain-VM no-op, and no existing test regresses. Only residual doc-staleness nits remain.

**`rust/crates/supervisor-daemon/src/service.rs`** (line 502): Minor doc staleness: this doc on refresh_cc_modes still says a confidential guest's card "gets its mode from the create gate that admitted it, with no read at all" — with this PR the start reads the card too (and it is the start, not the create, that last vouched for a stopped-then-restarted VM's card). The comment inside refresh_cc_modes_with was updated; this outer doc could say "the create gate or the start that admitted/re-admitted it".

**`rust/crates/supervisor-daemon/src/service.rs`** (line 181): Same nit: the gpu_cc_modes field doc says it is populated "by the SEV-SNP GPU gate when a create probes its cards" — worth adding the start path now that it also seeds the cache.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 808): forget_cc_modes' doc says the cache was learned "either before the cards were attached or from the create gate that read them on the way in" — after this PR the start's fresh reading is a third source that the stop must also drop. One clause would cover it.

foxpatch-aleph (changes requested):
The extraction of require_gpu_cc_mode is faithful (check order, error wording and cache behavior all preserved), its placement in start_vm_execution correctly precedes every irreversible step so a refusal leaves the VM STOPPED with no partial teardown, and the no-world-lock reasoning holds (validate_spec_gpus rejects cards claimed by stopped entries, and the refresh sweep skips attached cards, so the probe under the VM lock is race-free). However, the fix has a reachable hole: RebootVm on a stopped VM also resurrects the guest — reboot_vm unconditionally restarts the unit (which starts an inactive unit) and even recreates the SNP DHCP server — without ever going through start_vm_execution or the new gate, so the exact scenario this PR closes (confidential VM stopped, card re-moded while down, guest booted on an unverified card) is still possible via a sibling RPC. The existing test reboot_of_a_stopped_vm_keeps_the_stop_stamps_and_empty_forwards pins that the unit really comes back up, and its comment already admits the daemon should either refuse the reboot or run the full start path. Either run require_gpu_cc_mode in reboot_vm for a stopped VM, or refuse the reboot, before merging; the test suite otherwise covers the change well.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 1128): This gate is reachable from StartVm but not from RebootVm, and RebootVm on a *stopped* VM boots the guest too: reboot_vm (line 1276) never checks whether the unit is active, calls RestartUnit (which starts an inactive unit), and even recreates the SNP DHCP server at line 1321 — all without going through start_vm_execution. The existing test `reboot_of_a_stopped_vm_keeps_the_stop_stamps_and_empty_forwards` (line 6931) pins that the unit really restarts (`assert!(running, "the unit itself was restarted")`), and its own comment concedes "both daemons should either refuse the reboot or run the full start path". As written, an operator can re-mode a card while its SNP VM is stopped and bring the confidential guest back via RebootVm with no CC read — the precise scenario this PR closes. Please route the reboot path through `require_gpu_cc_mode` (or refuse a reboot of a stopped VM), or document why it is out of scope.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 2219): "applied wherever a guest is about to be handed it" is not yet true while reboot_vm can hand a stopped VM its card without the gate (see the comment on the start gate). If the reboot path is intentionally left out of this PR, this doc comment should say so; otherwise it overstates the invariant.

foxpatch-aleph (approved):
Faithful extraction of the create gate's per-card CC rule into require_gpu_cc_mode, with byte-identical messages and ordering, and a correct re-application on the start path: the gate runs before the stop stamps are cleared and before any side effect, so a refusal leaves the VM STOPPED with nothing stood up. The no-world-write-lock reasoning checks out (creates are rejected for attached cards at validate_spec_gpus, the sweep skips attached cards, gpu_cc_modes is a leaf lock, and the only caller holds the per-VM lock). SNP-only gating matches the create predicate. The three new tests cover the refusal, the cache-seeding success path, and the plain-VM no-op, and the service.rs comment updates are accurate. Two non-blocking observations: the reboot path (reboot_vm restarts the unit directly) still re-boots the guest without re-reading the card and keeps the sweep vouching CC-on through the window, which is the same shape this PR closes with a much narrower window; and the inherent TOCTOU between probe and QEMU attach (identical to create) could use a sentence in the doc comment. Verified statically only — cargo is unavailable in this review environment, so CI must confirm build, fmt, clippy and tests.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 1347): The reboot path still re-boots a persistent SNP VM's guest on its card without re-reading it: it restarts the unit directly rather than going through `start_vm_execution`, and because a reboot never stamps a stop, `attached_gpus` keeps returning `vouched_cc_on = true` through the window, so the sweep keeps seeding CC-on as well. That is the same exposure this PR closes, just with a seconds-wide window instead of an arbitrarily long stop. Either `reboot_vm` should run the same `require_gpu_cc_mode` loop before its restart, or the gate's comment (and the service.rs sweep comment) should say why the reboot is accepted — presumably the guest's own boot-time attestation backstop, which the PR description already leans on for the residual TOCTOU.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 2218): The doc covers why the read is live rather than cached, but not the residual race: the card is free between this probe and QEMU actually taking it (tap, nftables, DHCP and RestartUnit all run in between), so an operator can still flip the mode after the probe — the same race the create gate has. The guest's boot-time attestation is the backstop in both cases; a sentence saying the probe is admission, not a continuous guarantee, would set the right expectations.

foxpatch-aleph (approved):
This PR cleanly closes a real security gap: a card re-moded while its SNP VM sat stopped would have been handed a confidential guest on the stop's next start, with the inventory re-seeded 'on' purely from a live VM holding it. The create gate's per-card rule is extracted verbatim into require_gpu_cc_mode (same check order, same error wording) and run on the start path before any stop-stamp clearing or irreversible step, so a refusal leaves the VM STOPPED with nothing stood back up. I verified the concurrency claims: validate_spec_gpus rejects cards attached to any entry config (including stopped ones) so no create races the probe; attached_gpus walks configs regardless of liveness so the refresh sweep skips the card throughout; and the sweep's or_insert_with seed cannot overwrite the start's own reading. The gate condition (entry.config.snp().is_some()) matches the create gate's SNP condition, and plain passthrough VMs are ungated on both paths. Tests pin the refusal, the seeding, and the plain-VM non-gating, with a reasonable hermetic harness workaround for the real-sysfs BAR reads. One non-blocking observation: reboot_vm's persistent branch restarts the unit without going through start_vm_execution, so a reboot never re-reads the card either — a much narrower window, but the same reasoning applies and may deserve a comment.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 1347): Non-blocking: the persistent branch of `reboot_vm` restarts the unit directly and so never re-reads the card, and unlike a stop it does not drop the cache entry either. During the RestartUnit down-window the card is briefly idle hardware an operator could re-mode, and a mid-reboot re-mode would carry a confidential guest on a card whose cached 'on' nobody re-read — the same shape this PR fixes for the stop/start path, just in a window of seconds instead of an arbitrary gap. Worth a comment here (or a follow-up) explaining why the reboot is considered safe, since the start gate now exists one function over.

foxpatch-aleph (changes requested):
The extraction into require_gpu_cc_mode is behavior-preserving (identical check order, cache write, and error wording), the gate is placed before every irreversible step so a refusal leaves the VM STOPPED with nothing stood back up, and the no-world-lock reasoning is sound (VM lock held by the sole caller, validate_spec_gpus rejects creates for config-claimed cards, the sweep's attached set skips them). The tests are well-targeted and the sweep comment updates are accurate. However, the fix is incomplete for the security property it claims: reboot_vm (lifecycle.rs:1288) restarts the controller unit for persistent VMs with no liveness check, so RebootVm on a stopped SNP VM holding a re-moded GPU boots the confidential guest with no CC re-read — the exact hole this PR closes for StartVm, reachable through the same RPC surface (pinned by the existing test reboot_of_a_stopped_vm_keeps_the_stop_stamps_and_empty_forwards). Either gate the reboot path the same way or refuse the reboot of a stopped VM, as that test's own comment already suggests.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 1361): This is the remaining hole in the fix: for a persistent VM, reboot_vm restarts the controller unit with no liveness check, so a RebootVm on a *stopped* SNP VM holding a GPU boots the confidential guest directly — the same 'card re-moded while the VM was down' scenario, bypassing the new gate entirely (the test at reboot_of_a_stopped_vm_keeps_the_stop_stamps_and_empty_forwards pins that a stopped VM's unit is restarted, and its own comment says the daemon should either refuse the reboot or run the full start path). Since the card is ungated hardware exactly as long as the VM is down, this needs require_gpu_cc_mode on the same terms as start_vm_execution, or a refusal when stopped_at_ns != 0. As it stands, the daemon enforces the card rule on CreateVm and StartVm but not on the third path that hands a guest the card.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 2247): The doc comment says the rule is applied "wherever a guest is about to be handed it" — with the ungated reboot path above, it overpromises. Worth either tightening the wording or covering reboot_vm before this lands, since the two admissions are supposed to be unable to drift.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 1140): Minor, non-blocking: the start gate conditions on entry.config.snp().is_some() (the persisted config) while the create gate conditions on spec.tee.backend == SevSnp. They agree for anything this daemon creates, and snp() fails closed on a partially-written config, but a line noting that the start path derives its gate from the persisted config (and why that's equivalent) would help future readers confirm the two admissions really can't drift.

foxpatch-aleph (approved):
A well-reasoned, correctly placed fix. The extraction of the create gate's per-card rule into require_gpu_cc_mode guarantees the two admissions cannot drift, and applying it on the start path closes a real hole: a card re-moded while its confidential VM was stopped would otherwise boot a guest the host had not vouched for. The gate runs before every irreversible step (verified: stop stamps, restarting marker, tap/nft/DHCP/unit), so a refusal leaves the VM cleanly STOPPED. The no-world-lock justification holds up under scrutiny: the caller holds the VM lock, a create cannot claim an attached card, and the sweep skips attached cards. Lock ordering with gpu_cc_modes is consistent with existing paths. The injection-safety invariant (inventory-membership check before the probe argv/sysfs path) moved intact with its comment. Tests cover refusal, fresh-seeding, and plain-VM non-gating, and existing create-gate tests exercise the shared helper. Only minor nits.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 5165): Nit: `msg.contains("confidential-computing mode")` matches both the off-mode arm and the probe-error arm ("cannot read the confidential-computing mode of GPU..."). The cache assertion below disambiguates in practice, but asserting a phrase unique to the off-mode arm (e.g. "is not in NVIDIA confidential-computing mode") would pin which refusal fired.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 2255): Nit: the full `impl Fn(&str, &str) -> Result<Option<CcMode>, DaemonError>` signature is now spelled out at three sites in this file; the existing `crate::gpu_cc::CcProbe` type alias would express the same thing. Purely stylistic, and the expanded form matches snp_config_slice_with's existing style, so feel free to ignore.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 5148): Low-risk gap: there is no start-path test for the other two refusal arms (card no longer in the host inventory, probe error), but both are shared with the create gate, which has coverage through the pre-existing snp_config_slice_with tests, so the residual risk is small.

foxpatch-aleph (approved):
A clean, behavior-preserving extraction of the create gate's per-card CC rule into require_gpu_cc_mode and a correctly-placed application of it on the start path. I verified the refactor is byte-identical in error wording and check order; that the gate runs before the stop-stamp clear and every irreversible step; and that all three documented lock-free claims hold (vm_lock serializes lifecycle, validate_spec_gpus refuses attached cards to any create, and the refresh sweep probes only unattached cards, so no probe ever races the gate and the sweep's insert-if-absent seed cannot overwrite the gate's fresh reading). The start condition entry.config.snp().is_some() provably matches the create condition since only configs built by snp_config_slice set the SNP marker. Tests are well-designed: the refusal test pins wording, cache contents, STOPPED status, and DHCP teardown; the plain-VM test catches both an unwanted probe and an unwanted refusal. I could not run cargo (no toolchain in this review environment), but the tests read coherent against the harness. Two non-blocking observations below; the second is a scope question about the reboot path.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 1145): Scope observation, non-blocking: reboot_vm's non-program arm (lifecycle.rs:1364, `state.units.restart(&unit)`) still restarts the controller without this gate and without dropping the cache entry, so a card re-moded during the reboot's transient down-window would restart its guest without the host re-reading it. The window is seconds rather than the unbounded stop→start gap, and the guest's own boot-time attestation still catches it, so this is a policy question rather than a bug — but the service.rs comment "Neither the create gate nor the start comes through here" is true only because reboot comes through neither, which may be worth a sentence here or there if the reboot path is meant to stay deliberately ungated.

**`rust/crates/supervisor-daemon/src/gpu_cc.rs`** (line 9): Minor doc nit: the module doc's "The read only ever runs on a card no VM owns, so it never races a guest" was exact before this PR (sweep and create gate both read only unattached cards), but the start gate now also reads a card a stopped VM's config claims. The invariant that matters — never read under a live guest — still holds, so consider updating the wording to say the read never runs on a card a guest holds.

foxpatch-aleph (approved):
This PR correctly closes a real security gap: StartVm could bring back a confidential guest onto a GPU whose CC mode had not been read since the create gate, even though forget_cc_modes deliberately drops that reading on stop because an operator can re-mode an idle card. The extraction of require_gpu_cc_mode is behavior-preserving on the create path (same check order, same fail-closed wording, same inventory-membership-first safety property), and the start gate is placed before every state mutation and irreversible step, so a refusal leaves the VM STOPPED with nothing stood up. The no-world-lock reasoning is sound: the caller holds the VM lock, the card is claimed by this VM's config so no create can take it, and the refresh sweep's attached set makes the seed only ever restate the start's own answer. The three new tests pin the refusal, the fresh seeding, and the plain-VM pass-through. Remaining observations are minor and non-blocking: a start that passes the gate but fails later (unit restart, DHCP) leaves the CC-on reading cached with no forget path for the now-stopped VM, the start path's probe-error and not-in-inventory arms are untested, and reboot_vm's RestartUnit path still relaunches without a re-read.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 1133): Minor, non-blocking: when the gate passes but a later step of start_vm_execution_marked fails (unit restart, DHCP start), the gate's CC-on reading stays in the cache for a VM that reports stopped, and nothing forgets it — stop_vm_execution's already-stopped early return (line 832) skips forget_cc_modes. The card then keeps advertising "on" in GetHostInfo/VmInfo, diverging from the "a stopped VM's card advertises nothing" invariant documented in refresh_cc_modes_with (service.rs). No admission path trusts the cache, so this is reporting-only; consider forgetting the cards on the failure exits of the marked start if you want the invariant to hold there too.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 5077): Test-coverage suggestion: the start gate's probe-error arm ("cannot read the confidential-computing mode of GPU at pci_host ...", e.g. a card read during its reset) and the not-in-inventory arm ("is not in the host inventory") are covered on the create path but not on the start path. Since snp_vm_holding_a_card already builds the fixture, each would be a small test — worth having, as those are the arms a future refactor of require_gpu_cc_mode could silently break.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 1349): Observation, not a request: reboot_vm still relaunches the guest via RestartUnit without re-reading the card, so the reboot window (seconds, vs the potentially hours-long stop window this PR fixes) remains a way for a re-moded card to receive a confidential guest, with only the guest's boot-time attestation as backstop. If the invariant you want is "the host never launches a confidential guest onto an unread card", the reboot path is the remaining entry point; if the seconds-wide window is judged acceptable, a word in the start-gate comment saying so would help future readers.

foxpatch-aleph (approved):
The fix is correct and well-reasoned. The per-card admission rule is extracted verbatim (same check order, same fail-closed semantics, same error strings) so the create and start gates cannot drift. The gate is placed before the stop-stamp clear and every irreversible step, so a refusal leaves the VM STOPPED with nothing stood up, exactly as the PR claims. I traced the no-world-lock argument and it holds: the card stays attached in the world view while the VM is stopped, so creates cannot claim it and the refresh sweep skips it, and the sweep's or_insert_with seed cannot clobber the gate's unconditional insert. The three new tests cover the refusal, the seeding success path, and the plain-VM non-gating, asserting the right invariants. Two non-blocking observations: (1) reboot_vm still restarts an SNP VM's unit without re-reading the card, leaving a narrower window of the same exposure class - worth a follow-up or an explicit out-of-scope note; (2) the new sweep-comment sentence 'the seed below only ever restates the start's own answer' slightly overstates things for a VM adopted at daemon boot, where the seed is the only source since no start ran in that daemon's lifetime, though the surrounding comment retains the adoption rationale.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 1133): Scope question rather than a defect: `reboot_vm` (RebootVm RPC) restarts the unit directly and never comes through this gate, so a reboot of an SNP VM holding a card releases the GPU to idle hardware and re-attaches it without a re-read. That is the same class of exposure with a much smaller window, and the sweep keeps treating the VM as live throughout, so it is arguably out of the hole this PR closes — but it may deserve a follow-up or a one-line note in the comment saying the reboot is deliberately not gated.

**`rust/crates/supervisor-daemon/src/service.rs`** (line 617): "the seed below only ever restates the start's own answer" reads slightly absolute: for a VM adopted at daemon boot, no start runs in that daemon's lifetime, so the seed is the only source (the neighboring comment explains that adoption case, which is why this is cosmetic). Something like "the seed below restates the start's own answer (or vouches for an adopted VM)" would cover both.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 4130): Minor: "a plain fn pointer behind an Arc" is a little confusing — CcProbe is a plain fn pointer stored directly in DaemonState (which is itself what sits behind the Arc). The sentence parses either way, but "stored in the Arc'd DaemonState" would be clearer.

PR #1241, opened 2026-09-10
feat(agent): attach as many confidential GPUs as the V-PROGRAM names

odesenfans:
## Summary

`build_vprogram_spec` refused any V-PROGRAM asking for more than one confidential GPU, with a comment that single-card passthrough was the configuration this CRN validates. That refusal was the only single-card assumption left in the stack, and it sat in the one place the scheduler cannot see: the capacity resolver already consumes any `count` of cards, all or nothing; the daemon's SNP gate probes every card in the spec and sizes the MMIO window from all of their BARs; the guest init creates one device node per card and nvattest collects evidence for every GPU, which the attestation route returns as an array; the message schema caps at eight, NVIDIA's own ceiling. So a two-card message was placed by a scheduler that counts cards and then refused at launch, an allocation that churns instead of failing early.

This PR lifts the cap. The count is the resolver's business.

What the cap encoded was NVIDIA's validation matrix, not a code limit: multi-GPU confidential computing is validated on HGX B200/B300 with driver R590 or newer over encrypted NVLink, Hopper is one card per VM, and the RTX PRO 6000 Blackwell Server Edition lists it as not yet validated. That is now stated in the operator runbook as a failure signature to expect, next to the note that the guest's `swiotlb` reservation is sized per VM rather than per card. Multi-card joins the hardware-unverified list for the Tier 2 pass. If a SKU turns out to refuse multi-card, the right fix is a per-VM maximum in the daemon's per-card architecture table, advertised under `tee.nvidia_cc` so the scheduler can honour it, not a blanket refusal here.

## Changes

- `src/aleph/vm/agent/vprogram_launch.py`: the `count > 1` refusal and its comment are gone; the remaining GPU checks (runtime has a gpu block, vendor and arch match, memory floor) are unchanged.
- `tests/supervisor/test_vprogram_launch.py`: `test_gpu_vprogram_rejects_a_second_gpu` becomes `test_gpu_vprogram_accepts_a_multi_card_count`, which builds a spec for a two-card message and checks that the cards are left for run.py to resolve.
- `docs/operators/nvidia-cc.md`: the `VmSetupError ... one confidential GPU per VM` failure signature is replaced by what a multi-card launch can actually run into.

The plan sketch under `docs/superpowers/plans/` still shows the old refusal; #1205 drops those documents, so it is left alone here.

## Verification

`ruff format --check` clean; `ruff check` reports the same pre-existing count on both files as dev-2.1; `tests/supervisor/test_vprogram_launch.py` and `test_snp_instance_run.py` pass locally (44 tests). No measured file changes, so the golden measurements do not move.

### DIFF
--- a/docs/operators/nvidia-cc.md
+++ b/docs/operators/nvidia-cc.md
@@ -140,10 +140,15 @@ If `nvidia_cc` is absent from `/about/capability` while
   plain `gpu_device_id` shortage (`resolve_confidential_gpus`,
   `src/aleph/vm/agent/capacity.py`), so the log and the scheduler can tell
   a confidential-GPU shortage from an ordinary one.
-- **`VmSetupError ... asks for N GPUs; this CRN attaches one confidential
-  GPU per VM`.** The schema allows up to eight cards, but only single-GPU
-  passthrough is validated on the RTX PRO 6000 Blackwell Server Edition.
-  The scheduler should not have placed a multi-GPU V-PROGRAM here.
+- **A multi-card V-PROGRAM boots but the driver refuses one or more
+  cards.** The CRN attaches as many CC-mode cards as the message's `count`
+  names (up to the schema's eight), and every stage is per card. What is
+  not per card is NVIDIA's validation: multi-GPU confidential computing is
+  validated on HGX B200/B300 with driver R590 or newer over encrypted
+  NVLink, Hopper is one card per VM, and the RTX PRO 6000 Blackwell Server
+  Edition lists it as not yet validated. The guest's `swiotlb` reservation
+  is also sized per VM, not per card. Until a multi-card host has run the
+  hardware pass, expect a two-card V-PROGRAM to be an experiment.
 
 ## 6. What the CRN never does
 
--- a/src/aleph/vm/agent/vprogram_launch.py
+++ b/src/aleph/vm/agent/vprogram_launch.py
@@ -192,13 +192,10 @@ async def build_vprogram_spec(vm_hash: ItemHash, content: VerifiableProgramConte
 
     gpu = content.gpu
     if gpu is not None:
-        # The schema allows up to eight cards, the ceiling of NVIDIA's
-        # multi-GPU CC mode. One card per VM is the configuration this CRN
-        # validates: single-GPU passthrough on the RTX PRO 6000 Blackwell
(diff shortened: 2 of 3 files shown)


foxpatch-aleph (approved):
This PR removes the last single-card assumption in the confidential-GPU launch path, and every supporting claim checks out: the capacity resolver already handles any count all-or-nothing, the daemon's SNP gate probes each card and sums all BARs for the MMIO window, the guest init creates one device node per card, and the schema caps count at eight. The old refusal sat after placement on a scheduler that counts advertised CC cards, so it was indeed a churn source. Tests (44 total) and ruff format pass locally, the rewritten test appropriately guards against reintroducing the refusal, and the operator doc correctly replaces a now-impossible failure signature with a realistic one. Two minor observations: the replacement comment's "every stage below" names stages that are downstream in the pipeline rather than below in this function, and the removed cap was also the only launch-path re-check of the schema's le=8 ceiling for model_copy/model_construct routes (the count now relies solely on the resolver's host-inventory bound) — neither is blocking.

**`src/aleph/vm/agent/vprogram_launch.py`** (line 195): Minor wording: "every stage below" names stages that are downstream in the launch pipeline (the resolver in run.py, the daemon, the guest init), not below in this function — the remaining checks in this block don't depend on count at all. "every stage downstream" would be more precise.

**`src/aleph/vm/agent/vprogram_launch.py`** (line 194): Worth being aware of: the removed `count > 1` check was also the only launch-path enforcement of the schema's le=8 ceiling for routes that bypass field validation (model_copy/model_construct), which this file otherwise re-checks deliberately (roothash patterns, MAX_VERIFIED_VOLUMES) because the daemon trusts the downstream values. With the cap gone, a hand-built content with count=9 would pass build_vprogram_spec and attach nine cards on a host that has them. It's policy-only (every card is still individually probed and attested, and the resolver fails closed below the requested count), and per the PR text per-VM hardware maximums belong in the daemon's per-card table — so this seems intentionally accepted, just flagging the asymmetry with the file's own re-check convention.

**`docs/operators/nvidia-cc.md`** (line 143): Good replacement: the old signature described a failure state the scheduler can no longer produce, and the new entry correctly tells operators what a multi-card launch can actually run into (validation matrix, per-VM swiotlb sizing). The "experiment" framing is honest about the untested-hardware status.

PR #1242, opened 2026-09-10
feat(agent): advertise the allocation protocol version in /status/config

odesenfans:
Adds `"api": {"allocations": 2}` to `GET /status/config`.

The scheduler switches to `POST /v2/control/allocations` per node from this field: it already fetches `/status/config` every poll, so discovery costs no extra request and no per-node memory. A node without the block is a 2.0 node and keeps receiving the legacy call. Bump the number when a newer allocation endpoint ships.

One test through the real app.

### DIFF
--- a/src/aleph/vm/agent/views/__init__.py
+++ b/src/aleph/vm/agent/views/__init__.py
@@ -521,6 +521,13 @@ async def status_check_version(request: web.Request):
         return web.HTTPForbidden(text=f"Outdated: version {current} < {reference}")
 
 
+# The highest /control/allocations protocol this node serves. The scheduler
+# reads it from /status/config, which it already polls, and calls
+# /v2/control/allocations from 2 up; a node without the field gets the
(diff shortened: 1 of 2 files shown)


foxpatch-aleph (approved):
Minimal, correct change that adds a protocol-version advertisement (api.allocations: 2) to the public /status/config response. The advertised version is accurate — POST /v2/control/allocations is registered in setup_webapp (src/aleph/vm/agent/supervisor.py:420) — and the feature-detection contract is sound: the key is additive, doesn't collide with existing keys, and its absence on older nodes cleanly means legacy. No sensitive data is exposed beyond a capability number on an already-public endpoint. The test goes through the real app and asserts the literal value, which pins the wire contract and will force a conscious update when the constant is bumped; it passes locally, alongside its sibling suite (12 passed). The constant's comment documents the consumer contract (scheduler calls v2 from 2 up, absence means legacy) precisely. No issues found.

PR #1243, opened 2026-09-14
refactor(agent): point the storage CLI at the reconciler's API names

odesenfans:
## Summary
#1231 dropped the underscore from the reconciler helpers the storage CLI imports (`is_vm_namespace`, `teardown_orphan_devices`, `release_cache_devices`, `has_namespace_dirs`) and left four aliases under the old names so #1224's CLI kept importing while the two branches were separate. Both are merged, so the CLI now imports the names themselves and the aliases are gone. No behaviour change.

## Changes
- `src/aleph/vm/agent/storage_cli.py`: imports and calls use `is_vm_namespace`, `teardown_orphan_devices` and `release_cache_devices`.
- `src/aleph/vm/agent/vm/reconciler.py`: the four compatibility aliases at the end of the module are removed.

## Test plan
- The existing storage CLI tests cover every changed call site (`reclaim` refusal on a non-hash name, the device teardown before a full pass, the cache-device release after it). No test change: nothing they assert moved.
- `ruff check` and `ruff format --check` clean on both files.


foxpatch-aleph (approved):
This is a minimal, mechanical rename that swaps the storage CLI's imports and call sites from the reconciler's underscore-prefixed aliases to the canonical names and deletes the four compatibility aliases from reconciler.py. I verified the change is complete: no references to _plausible, _still_on_disk, _teardown_orphan_devices, or _release_cache_devices remain anywhere in the repo, the canonical functions exist at reconciler.py:976 and reconciler.py:1001 (with is_vm_namespace available via the import at reconciler.py:80), both modified modules import cleanly, reconciler.py ends cleanly with no leftover alias block, and docs/architecture/storage.md already uses the underscore-free name. The existing suites exercise every changed call site and pass in full (72 storage CLI tests, 82 reconciler tests). The only ruff findings in this environment are on lines untouched by the diff and pre-date the PR.

**`src/aleph/vm/agent/storage_cli.py`** (line 117): Nit: `is_vm_namespace` is not defined in reconciler.py — it reaches that module via `from aleph.vm.storage import ... is_vm_namespace` (reconciler.py:80) — so this import resolves through an implicit re-export. Importing it from `aleph.vm.storage` directly would be marginally more direct. Non-blocking; importing via the reconciler is consistent with where the old `_plausible` alias lived and with this PR's framing of pointing the CLI at the reconciler's names.

PR #1245, opened 2026-09-14
feat(agent): give the owner a route to start a VM they stopped

odesenfans:
### Summary

The agent's allocation loop treats a STOPPED VM as a decision somebody made and
refuses to overrule it: a plan naming a stopped VM does not restart it, on
purpose, because the owner is who stopped it. The loop's own docstring says
"only its owner undoes it", and until now nothing implemented the undo. The
only route that reached the start path was the legacy
`/control/allocation/notify` push, which a plan-governed node refuses with 409
and which we want decommissioned, so a VM its owner stopped stayed stopped for
good while the node went on committing its memory, vCPUs and disk.

This adds `POST /control/machine/{ref}/start` next to the existing stop and
reboot routes, and makes `operate_reboot` stop claiming it starts a VM it never
touches.

### Changes

- **`POST /control/machine/{ref}/start`**, authenticated and answered exactly
  like `stop`: owner or delegate, 403 otherwise, 404 for a hash this node holds
  no registry record for. **No payment check**, deliberately, because `stop`
  does not check either and the two are a pair; a node that wants to stop
  serving an unpaid VM has expiry and the plan for that.
- It goes through the existing `start_persistent_vm(..., recreate=True)`, so it
  inherits the behaviour the convergence loop already relies on: a STOPPED VM is
  resumed in place (definition, disks and port forwards preserved), a FAILED one
  is rebuilt, a STOPPING one is waited out and rebuilt, and the whole thing is
  serialised on the per-hash create lock so a double click cannot build the VM
  twice. `recreate=True` is the same claim the loop makes: the record already
  committed the memory and the vCPUs, so admission only has to find room for the
  disk.
- A VM that is already RUNNING or BOOTING is answered `200 ... is already
  running` without being restarted, and one awaiting its owner's confidential
  session is answered `409` naming `/confidential/initialize`, since only those
  certificates can start it.
- A VM the supervisor has forgotten (it forgets across a restart while the
  record and the disks stay) is rebuilt rather than 404'd, following
  `operate_erase`'s precedent.
- **`operate_reboot` is honest now.** It only ever acted on RUNNING or BOOTING;
  every other status fell through to a `200` reading `Starting VM (was not
  running) with ref ...` having called nothing at all. It answers `409` and
  names the start route instead.
- **A start that fails answers like the create route, not with a bare 500.** A
  rebuild can fail for the ordinary reasons a create fails (no room on the node,
  a base image that will not download, a guest that never reaches RUNNING), and
  `create_vm_execution_or_raise_http_error` has mapped those to 503 / 400 / 500
  with a reason for a long time. Rather than repeat that ladder in the view,
  `start_persistent_vm` now takes the create callable it should use: it defaults
  to `create_vm_execution`, so the convergence loop and the legacy push are
  unchanged and keep classifying raw exception types onto their backoff ladder,
  and the view passes the wrapper. One parameter and one call site in `run.py`,
  kept deliberately small because the two error ladders there are being merged
  on another branch.
- The lifecycle guide's stop section already promised the owner "starts it
  again, through the same operator API". It now names the route that does it.

### Test plan

- CI runs the suite. New cases in `tests/supervisor/views/test_operator_retire.py`,
  on the same fake-supervisor harness as the existing stop and reboot tests:
  start of a STOPPED VM goes through the start path once and as a rebuild and
  answers 200; start of a RUNNING VM answers 200 without starting anything;
  start of a hash the node holds no record for is 404; a stranger and an
  unsigned caller are refused with exactly the status `stop` gives them; reboot
  of a STOPPED VM answers 409 and names the start route; and a start whose
  rebuild hits `InsufficientResourcesError` answers 503, through the real
  `start_persistent_vm` and the real wrapper rather than a mocked one.
- Locally: `ruff check` and `ruff format --check` on the touched files (no new
  findings; every remaining one is pre-existing and identical on the base
  commit), `python3 -m py_compile` on each.
- Allocation state was traced rather than instrumented: a restarted VM is simply
  live again, `_start_missing` skips it, and `_forget_settled` clears any failure
  record it was carrying. No new machinery, and the executions list reports it
  RUNNING because the supervisor does.


foxpatch-aleph (approved):
Solid, well-tested PR that fills a real gap (owner-stopped VMs were unrecoverable) with a route that correctly reuses the serialised start path and inherits its semantics. Auth mirrors stop exactly, the pre-lock checks are advisory with the authoritative state machine inside start_persistent_vm's lock, and error mapping through create_vm_execution_or_raise_http_error is a clean reuse of the create route's status ladder without disturbing the convergence loop or legacy push. Tests pass and genuinely exercise the new behavior, including the real 503 mapping path. Only minor observations remain.

**`src/aleph/vm/agent/views/operator.py`** (line 507): Unlike operate_stop/operate_reboot, operate_start doesn't branch on record.persistent. For an owner-stopped ephemeral VM (stop is a retire RECREATE, record kept) this route rebuilds it via create_vm_execution — consistent with the ephemeral reboot cycle — but start_persistent_vm then calls expiry.cancel(vm_id) and arms the update watcher with no idle-teardown timer re-armed, so an owner-started on-demand program would run indefinitely. Probably the intended edge-case behavior, but a comment here (or in the docstring) saying the route is really about persistent VMs and that an ephemeral start is a recreate would help future readers.

**`src/aleph/vm/agent/views/operator.py`** (line 529): This awaiting_confidential_init -> 409 branch is the only branch of the new handler not covered by the new tests, even though the harness already supports it (_vm_info takes awaiting_confidential_init=True). A one-case addition would close the gap.

**`src/aleph/vm/agent/run.py`** (line 750): Callable[..., Awaitable[None]] loses all parameter checking for the injected create callable — a small Protocol would let type checkers verify callers pass something start_persistent_vm can actually call. Noted as deliberate pending the error-ladder merge; feel free to ignore.

**`tests/supervisor/views/test_operator_retire.py`** (line 602): Minor: _app_with_a_recorded_vm is defined mid-file at line 602 while the other shared helpers (_vm_info, _fake_supervisor) sit at the top; the three earlier stop/reboot tests above it duplicate its setup and could use it. Pure style — moving it up (and folding the older tests onto it) would shave a bit of duplication.

foxpatch-aleph (approved):
A well-scoped, correct implementation. The new start route mirrors stop's authorization and 404 semantics exactly, defers all state decisions to start_persistent_vm's per-hash-locked re-check (so the view's pre-checks are advisory and races are handled), and reuses the create route's HTTP error ladder for rebuild failures without changing the reconciler's raw-exception backoff contract. Skipping the payment check is sound because a fully-retired unpaid VM has no registry record and start 404s on it. The reboot fix removes a response that claimed a start it never performed. Tests are thorough and pass (all 20 in the touched module, 159 across the views/allocation suites), the docs match the code, and lint/format are clean; the only failures in the environment are pre-existing dependency-version issues in code the PR does not touch.

**`src/aleph/vm/agent/views/operator.py`** (line 544): Minor, non-blocking: the HTTP-mapping wrapper only covers the rebuild path. When a STOPPED VM is resumed in place (supervisor.start_vm + _wait_until_running inside start_persistent_vm), a VmStartupError or a failing start_vm escapes unmapped and the owner gets aiohttp's bare 500 rather than the "VM failed to start" reason the create ladder gives for the equivalent never-reached-RUNNING outcome. The PR's framing ("a start that fails answers like the create route") fully holds only for the rebuild path. Could be worth a brief comment noting the resume path intentionally answers 500, or wrapping it too — but the behavior is acceptable and not a bug.

PR #1246, opened 2026-09-14
fix(daemon): adopt a VM whose unit failed as a death, not as a stop

odesenfans:
**Summary**

At adoption the daemon read one bool per controller unit, so a unit systemd
had marked `failed` and a unit an owner had stopped looked identical: both
were stamped stopped, and `vm_status` short-circuits on that stamp. A planned
VM whose guest died just before a host reboot (or whose unit failed to start
after the daemon restarted) reported STOPPED for the rest of its life. The
agent leaves STOPPED to the owner and answers the scheduler `unchanged`, so
the VM never ran again and the scheduler believed it was placed and healthy.

Adoption now reads the unit liveness it already fetches. A `failed` unit is a
death: the entry takes a start stamp and no stop, reports FAILED through the
existing dead-unit arm, and the agent rebuilds it. An `inactive` or unloaded
unit keeps the stopped stamp, which is what a stop through StopVm leaves
behind. A unit mid-job, an ActiveState systemd grows later, and an unreachable
bus all stamp nothing and defer to the live unit queries.

The rebuild of an adopted failure is bounded by the same backoff ladder as any
other death (`allocation/reconciler.py::_start_one` with `recreate=True`), so
a VM the node cannot really run does not spin.

**Changes**

- `UnitLiveness` gains `Failed`, plus `is_dead()` for the "settled, however it
  got there" question `vm_status` asks.
- `build_world_view` matches on the liveness instead of a bool: failed stamps
  a start and marks the entry, inactive stamps a stop, everything undecided
  stamps nothing.
- `VmEntry::adopted_failed` records the verdict, since the two predicates that
  read "started with no stop" as a live guest run with no unit observation:
  the GPU CC-mode seed (a dead guest vouches for no card) and
  `awaiting_confidential_init` (a dead SEV controller is not waiting for its
  session). StartVm and RebootVm clear it.
- `UnitStateSource::active_states`, the bool projection, is deleted; its two
  remaining callers ask the liveness directly.

**Test plan**

- `cargo fmt --all -- --check` and `cargo clippy --locked --all-targets -p
  supervisor-daemon -- -D warnings` are clean; CI runs the suites.
- New hermetic tests: adoption of a failed unit (FAILED, start stamped, no
  stop, no GPUs), adoption after a real `stop_and_disable` (STOPPED),
  adoption of a mid-job unit (nothing stamped), the bus-failure case extended;
  the CC seed and awaiting-session cases for an adopted failure; an async
  ListVms case that reports FAILED without announcing it twice; and the
  liveness mapping itself in `units.rs`.


foxpatch-aleph (approved):
Well-reasoned, well-documented fix. The adoption path now distinguishes a failed unit (a death: start stamp, no stop, FAILED reported so the agent rebuilds) from an inactive one (the owner's stop), from a mid-job unit (defers to live queries). I verified the load-bearing assumption that stop_and_disable leaves a unit inactive rather than failed, that the agent-side reconciler rebuilds FAILED and leaves STOPPED alone, that CreateVm treats a failed unit as a fresh create, and that both predicates reading 'started with no stop' as a live guest (GPU CC seed, awaiting_confidential_init) are gated by the new flag. All flag-clear points (StartVm, RebootVm, CreateVm entry replacement, readopt) are handled, and StopVm on an adopted failure can't misreport because the stop stamps short-circuit the status. Test coverage of the new behavior is thorough and hermetic.

**`rust/crates/supervisor-daemon/src/units.rs`** (line 227): This doc comment still references `active_states`, which this very PR deletes — it should now say `unit_states`/`controller_units`.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 1135): The clearing of `adopted_failed` in `start_vm_execution` (and the equivalent in `reboot_vm` at lifecycle.rs:1329) has no direct test — the existing tests only cover the create-path entry replacement. A small lifecycle test (start an adopted-failed entry, then assert the card vouches again / awaiting_confidential_init resumes its usual shape) would pin the reset.

**`rust/crates/supervisor-daemon/src/world.rs`** (line 1419): Nit: the comment says "The cards stay attached, but no guest of this VM holds them" while the assertion is `entry.gpus.is_empty()` — the entry no longer claims the cards at all (the claim lives in `config.gpus`, which is what `attached_gpus` reads). The comment and the assertion describe two different lists; worth a word clarifying that `config.gpus` is what stays claimed.

**`rust/crates/supervisor-daemon/src/units.rs`** (line 869): Trivial: `a_failed_unit_is_dead_without_being_a_deliberate_stop` asserts Transitional and Unknown are not dead but skips `assert!(!UnitLiveness::Active.is_dead())` — the one remaining variant.

foxpatch-aleph (commented):
Clean, well-tested fix. Replacing the bool-per-unit projection with UnitLiveness::Failed makes the failed/stopped distinction explicit at the type level, the adoption stamping in build_world_view correctly produces a death shape for failed units, and guarding attached_gpus' CC vouching and awaiting_confidential_init on !adopted_failed closes the exact hole that stranded reboots in STOPPED forever. vm_status semantics are preserved via is_dead(), and the agent-side contract checks out: FAILED now reaches the delete+recreate path with the backoff ladder instead of the resume-in-place path. Two non-blocking comments below on the death-mark clearing asymmetry and the mid-job adoption arm.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 1135): Minor asymmetry with the reboot path: here `adopted_failed` is cleared before the restart attempt (line 1139 runs `start_vm_execution_marked` after this), while `reboot_vm` (line 1329) clears it only after `restarted?` succeeds. If this StartVm fails, the entry is left with started≠0/stopped==0 and `adopted_failed == false` — precisely the shape `attached_gpus` vouches CC-ON for and `awaiting_confidential_init` treats as live, for a guest that is still dead. Low severity since the same shape already exists for pre-PR mid-run guest deaths, and the agent's FAILED recovery is delete+create rather than StartVm — but since this PR's whole point is to never vouch for a dead guest's cards, consider moving the clear to the success path (alongside the `started_at_ns` stamp, as reboot does) or noting why pre-clearing is intended.

**`rust/crates/supervisor-daemon/src/world.rs`** (line 745): Nit on the comment: "defers to live unit queries" is slightly optimistic for the never-started case. If the unit is Transitional/Unknown at adoption and later settles dead, `vm_status`'s dead-unit arm requires `started_at_ns != 0`, so an entry stamped with nothing here reports DEFINED indefinitely — there's no re-stamp once the unit settles. Self-healing in practice (the agent starts a DEFINED VM), but worth a word in the comment so a future reader doesn't assume the status converges on its own.

PR #1247, opened 2026-09-14
refactor: the 2.1 series' duplicated rules and vestigial wrappers, in one place each

odesenfans:
**Summary**

Ten behaviour-preserving clean-ups from the 2.1 code-bloat review, plus the
cheap extras it flagged for the same files. One commit per item: six copies of
a certificate signature check in the attestation code, two copies of one
HTTP error table, two mutexes and a bookkeeping struct for one GPU CC pass,
four copies of the program-versus-unit liveness rule, two identical TCB
ladders, three obsolete hook setters, three spellings of one systemd query, an
emptied-out message wrapper, dead arithmetic in front of `BigNum::from_slice`,
and the storage CLI's private copies of two shared helpers. 531 lines out, 272
in.

**Changes**

- `aleph-tee`: `pki::check_signed_by` and `check_crl_signed_by` carry every
  chain signature check; the TDX module and QE identities share one
  `IsvsvnTcbLevel` and one `walk_isvsvn_ladder`; the SEV-SNP report signature
  goes to openssl untrimmed, with a single-pass byte reversal; the pinned
  Intel root's own validity window is asserted by its unit test instead of on
  every collateral document.
- `supervisor-daemon`: the CC-mode refresh keeps one lock (the per-card
  freshness window was already the gate that mattered); `service::liveness_of`
  and `units::query_unit` replace four and three copies of the liveness and
  single-unit-query rules; the trait drops its lossy `active_states`
  projection; `GpuMmioBudget` stops carrying a constant field.
- agent (Python): the instance/v-program create path maps failures through the
  same table as the program path (and gains its `MicroVMInitError` and
  supervisor-side `HostNotFoundError` arms); the three per-module hook setters
  are gone now that `hooks.py` installs one object; the plan carries the
  verified `ExecutableMessage` rather than a one-field wrapper; the storage
  CLI reads pools and retention budgets through `storage_pools` and the
  reconciler; `_check_against` reads `HostCaps` directly.

Two deliberate behaviour deltas, both fail-closed and both in the commit
bodies: a message-less `FileTooLargeError` no longer raises out of
`args[0]`, and a TDX module ISVSVN above 255 now meets no ladder rung instead
of failing to parse.

**Test plan**

- `cargo fmt --all -- --check` and `cargo clippy --locked --all-targets -D
  warnings` for `aleph-tee`, `supervisor-daemon` and the attest agent: clean.
- `ruff check` / `ruff format --check` on the touched Python files: no new
  findings against the base.
- CI runs the Rust and Python suites. Tests that pinned a deleted spelling
  were updated, none deleted.
- The golden measurements need re-seeding (aleph-tee is measured into the
  initrd).


foxpatch-aleph (approved):
A disciplined, well-executed cleanup. Every consolidation was verified against the code it replaces: the PKI helpers preserve all six signature checks and their error contexts; the strip_leading_zeros removal is safe because BigNum::from_slice already handles leading zeros; the TDX ladder unification with its u8->u16 widening is fail-closed as documented; the liveness/query_unit consolidation preserves the bus-failure degradation semantics exactly; the unified Python error table is a superset of both old tables and fixes a real args[0] IndexError hazard; and no stale references to any deleted symbol remain. Both deliberate behaviour deltas are correctly characterized as fail-closed. Tests were updated rather than weakened, and the golden measurements were re-seeded in the head commit as required.

**`rust/crates/aleph-tee/src/tdx/certs.rs`** (line 184): Moving the pinned root's validity-window check from runtime to the unit test is defensible (the pin is byte-identical and its window is a compile-time fact), but note that CI is now the only tripwire: a deployed build carrying an expired pin (valid to 2049) will silently continue accepting the root where the old code failed closed. The comment makes the trade-off explicit, so this is just something to keep in mind for the next pin rotation.

**`src/aleph/vm/agent/storage_cli.py`** (line 520): The status command now stats each pool twice (once directly via pool_usage_bytes, once inside _retention_budget) where the deleted _pool_usage helper shared one statvfs. Harmless for a CLI, just a tiny redundancy if anyone ever cares.

foxpatch-aleph (approved):
A disciplined, well-executed deduplication. Every consolidated rule traces back faithfully to its duplicates: the extracted pki::check_signed_by/check_crl_signed_by carry the same error specificity, the strip_leading_zeros removal is genuinely dead arithmetic (BigNum::from_slice handles leading zeros; the all-zeros edge case yields 0 both ways), the TCB ladder unification preserves highest-rung-first ordering with the documented fail-closed u8→u16 delta, and the removed CC sweep gate leaves the per-card TTL as the DoS-relevant bound, which still holds. On the Python side the unified error table is a verified strict superset of the deleted inline one (arm-by-arm), the FileTooLargeError message-less case now degrades gracefully instead of raising IndexError, and the reconciler's set-union iteration is safe because _forget pops both dicts with defaults. All deleted identifiers (VerifiedMessage, hook setters, active_states, _caps, CcSweep, _RootBudget.root, shortest) have zero remaining references, and tests were updated rather than deleted. The pinned-root window moving to a unit test is sound since the pin is an include_bytes! compile-time constant that callers must byte-match. Only negligible nits remain.

**`src/aleph/vm/agent/storage_cli.py`** (line 525): Nit: _status now stats each pool's filesystem twice — pool_usage_bytes(pool) here for the free figure, and again inside _retention_budget(pool) for the total — where the deleted local helper read it once. Negligible for a CLI status dump, but threading the one PoolUsage through both would restore the single read if you care.

**`rust/crates/aleph-tee/src/tdx/certs.rs`** (line 184): Agreed this is a compile-time fact and the unit test will catch expiry in CI. Worth noting the trade explicitly somewhere: with the runtime check gone, a deployment running past the pin's notAfter keeps verifying (the previous fail-closed behavior was a self-inflicted DoS), and nothing but CI notices. That is the right call for a pinned trust anchor — pinning is exactly not relying on expiry — but the reasoning now lives split between this comment and the test assertion.

**`src/aleph/vm/agent/storage_cli.py`** (line 126): Importing the private-named _retention_budget from aleph.vm.agent.vm.reconciler follows the existing convention in this file (_startup_refusal, live_hashes), so this is consistent — but if this helper is now the shared source of truth for the retention budget, promoting it out of the underscore namespace would make the cross-module contract honest.

foxpatch-aleph (approved):
Verified all ten consolidations against the tree at b1d2563. The extracted pki.rs helpers (check_signed_by / check_crl_signed_by) preserve signature contexts and bailing error strings, including the self-signed ARK and the pinned Intel root's runtime window check (still enforced plus a new unit-test assertion). The unified _raise_http_for_program_error covers every arm of the old instance-path table and adds MicroVMInitError and the supervisor-side HostNotFoundError arm, so no handler regresses; its two documented deltas are genuinely fail-closed (a message-less FileTooLargeError now yields a 400 with a fallback string instead of raising IndexError, and a TDX module ISVSVN above 255 parses but can never meet a u8-provided SVN, correctly reporting below-every-level). The GPU CC pass is functionally replaced by per-card probe windows: CcModeWindow equals the old sweep gate for fresh cards, and CcUnreadableWindow cleanly separates transient bus failures from the hard one-hour Unknown gate, removing the failure mode where a newly-stale card could not be re-probed because the sweep gate was closed. walk_isvsvn_ladder is a faithful unification of the two TCB ladders (explicit u16 widening preserves ordering for both TDX-module and QE identities). Replacing the four copies of the program-vs-unit liveness rule with UnitLiveness/query_unit preserves the old semantics including the 'dead guest stops the init poller' case (Unknown no longer short-circuits early exit). The hook-setter consolidation into AgentHooks is strictly better: the install is now atomic, installed_hooks restores rather than clears, and no remaining references to the old setters exist. The reconciler's submit() union loop is safe because _forget is idempotent (double pop with defaults). Reusing the CLI's helpers in reconciler.py and pool_usage_bytes is a clean drop; storage_pools.pool_usage_bytes correctly preserves the 'free' count and the None-on-OSError semantics. The remaining refactors (VerifiedMessage removal, _RootBudget root field, liveness_of, is_removable_by_allocation) are all behaviour-preserving with no remaining references to the removed definitions. No bugs or security issues found; tests were updated coherently across the affected suites.

**`src/aleph/vm/agent/storage_cli.py`** (line 310): Minor, non-blocking: _status now calls pool_usage_bytes(pool) and retention_budget(pool) where the old code called shutil.disk_usage once, so a status pass issues an extra statvfs per pool. Cosmetic given the CLI's interval; ignore if deliberate to keep each helper self-contained.

foxpatch-aleph (approved):
This is a careful, genuinely behaviour-preserving cleanup. I verified each consolidation against the code it replaces: the six certificate-signature checks in pki.rs preserve semantics and error quality; dropping strip_leading_zeros is safe because BigNum::from_slice accepts leading zeros (the all-zero case included); the ISVSVN u8->u16 widening is fail-closed exactly as the description claims; liveness_of/query_unit preserve the bus-failure->Unknown->inactive and times-based-program rules at all four call sites; and removing the whole-sweep CC gate leaves the per-card is_fresh TTL (both windows.mode and windows.unreadable) as the rate limiter for the unauthenticated path. On the Python side, _raise_http_for_program_error is order-compatible with the removed instance-path chain, gains the MicroVMInitError and supervisor-side HostNotFoundError arms, and fixes the args[0] IndexError on a message-less FileTooLargeError; the reconciler submit() union loop is equivalent given _forget is an idempotent double-pop; the removed hook setters leave no dead imports; and the storage CLI's new shared helpers match the deleted private copies exactly. I ran the affected Python suites (348 tests across hooks, allocation verify/verdict/reconciler, retire, storage, capacity and run-path tests) — all pass; the only failures are a pre-existing sandbox aleph_message version mismatch in code this PR does not touch. One note: the PR description says the pinned Intel root's validity window is 'asserted by its unit test instead of on every collateral document', but the runtime check at tdx/certs.rs:186 remains — the new test is an added CI canary, not a replacement, which is the safer direction and fine as-is.

**`rust/crates/aleph-tee/src/tdx/certs.rs`** (line 186): The PR description says the pinned root's window is 'asserted by its unit test instead of on every collateral document', but this runtime check is still here — the new test at line 362 is an addition (a CI canary for pin expiry), not a replacement. Keeping the runtime check is the safer direction, so this is only a description inaccuracy; maybe reword the PR body so a future reader doesn't think the runtime check was dropped.

**`rust/crates/aleph-tee/src/tdx/tcb.rs`** (line 561): walk_isvsvn_ladder and its two call sites have no direct unit test, only indirect coverage through the module-identity and QE-identity tests. The u16 widening and the 'never trust document order' sort are both security-relevant enough that one direct test (e.g., levels listed low-to-high, and a rung above 255) would pin the contract cheaply. Non-blocking — the existing tests do exercise both ladders end to end.

PR #1248, opened 2026-09-14
test: prove each 2.1 rule once, and share the test doubles the series retyped

odesenfans:
### Summary

Acts on the fifteen items the test-bloat assessment ranked for 2.1, tests
only. Three shared-helper hoists first, then one deletion or merge per file.
753 lines out, 297 in, and 3 seconds of wall clock off every run of the
attest-agent suite. No production code is touched: every Rust change is
inside a `#[cfg(test)]` module or helper.

### Changes

Hoists:

- `units.rs` gets a delegation macro for `UnitStateSource`, so the three test
  wrappers the 2.1 series added spell out only the one or two methods they
  override.
- `pki.rs` gets the self-signed impostor certificate builder the three TDX
  test modules each hand-rolled.
- `service.rs` gets `cards_state`/`cards_state_in`, replacing nine copies of
  the `HostState` literal and two hand-rolled card constructors.

Deletions and merges:

- `vm_namespace` gets the direct table test it never had, and the five call
  sites that re-proved its alphabet keep only what is specific to them.
- Two SEV-SNP chain refusals their neighbours already make, the second of
  which makes it against AMD's real pin.
- The two TDX `walk_*` tests become one, driven by the level helper this
  series added.
- Three tautological cases in `test_agent_hooks.py`, and the whole of
  `test_agent_state_isolation.py`, whose subject is a conftest helper.
- The two parser-level marker tests, keeping the two observable
  `read_marker` twins.
- The three "unchanged" verdict tests become one parametrized case, and the
  headroom test whose assertion is fixed by its own setup goes.
- `TestRemovability`, now covered case for case by `TestRetentionReason`.
- The CLI copy of the naive-timestamp marker, the cache budget assertion that
  holds by arithmetic, and the busy-wait test that spent 3 seconds repeating
  its neighbour.

Every deletion names its surviving test in the commit body; every merge keeps
all the inputs and assertions of the tests it absorbed.

### Test plan

- `cargo fmt --all -- --check`, clean.
- `cargo clippy --locked --all-targets` for `supervisor-daemon`, `aleph-tee`
  and `aleph-attest-agent`, clean with `-D warnings`.
- `ruff check` / `ruff format --check` on the touched Python files: no new
  findings against `dev-2.1` (these files do not lint clean on either side).
- CI runs the suites. Worth watching: the merged TDX walk test, the busy-route
  `Retry-After` assertion, and the TDX impostor-certificate tests that now
  build their certificate through `pki::test_cert`.


foxpatch-aleph (approved):
Test-only cleanup, verified behavior-preserving throughout. The three hoists are faithful: the delegation macro's signatures match every UnitStateSource method and all three converted wrappers delegate exactly the sets they hand-wrote before; pki::test_cert reproduces the three hand-rolled impostor builders (subject/issuer/keys/validity/signing identical); cards_state/cards_state_in collapse nine HostState literals without changing any input. Every deletion is genuinely redundant with a surviving twin: the removability predicate is a one-line delegation to retention_reason (whose disagreement test keeps the direct assertion), the SEV-SNP pinned-root refusal survives in the stronger real-AMD-pin test, the expired-ARK branch survives in the injected-clock test (ARK is checked first), the marker parser's observable behavior is covered at the read_marker level, and vm_namespace's alphabet is now pinned once at the guard all eight delete paths call. The allocation-verdict merge strengthens the STOPPING case, and the new busy-route Retry-After assertion goes through the real endpoint while the pure-function test recovers the 3s case. Touched Python files compile clean with no orphaned imports.

**`rust/crates/supervisor-daemon/src/units.rs`** (line 840): The `reload` arm currently has no callers — none of the three wrappers delegates reload (they keep the trait's no-op default, as they did before). Keeping it completes the trait's method set, which is defensible, but if it stays unused it is dead macro surface a future reader has to reason about.

**`tests/supervisor/conftest.py`** (line 163): Deleting test_agent_state_isolation.py leaves `reset_agent_state` with no direct test — the deleted file was its only consumer. The fixture runs around every supervisor test, so a regression there will now surface as order-dependent cross-test failures rather than one focused failure. Defensible for a test-cleanup PR (the deleted test's subject was a conftest helper), but worth a conscious call: a broken reset is exactly the kind of thing that silently degrades a suite over time.

PR #1249, opened 2026-09-14
fix(packaging): ship the aleph-message the code is written against

odesenfans:
## Summary
Every V-PROGRAM create on a node installed from the .deb fails, and has since #1200 landed. The launch path reads `content.gpu` (`vprogram_launch.py:193`, also `run.py:697`), a field aleph-message added in 1.5, but the .deb installs **aleph-message 1.4**: `packaging/Makefile` carries its own dependency list and had drifted from `pyproject.toml`.

On the node the create raises, the VM is retired as `failed_create` and the scheduler is told the node cannot run the workload:

```
ERROR aleph.vm.agent.views:775 | Unhandled Error while starting VM '94369341...':
'VerifiableProgramContent' object has no attribute 'gpu'
  File "/opt/aleph-vm/aleph/vm/agent/vprogram_launch.py", line 193, in build_vprogram_spec
    gpu = content.gpu
AttributeError: 'VerifiableProgramContent' object has no attribute 'gpu'
```

Found by the aleph-testnets full-suite run of dev-2.1 on the TEE server (run 34347251256): the three V-PROGRAM tests failed with the V-PROGRAM never becoming reachable, while every other confidential and instance test passed. The unit suites cannot see it: they run against pyproject's pin, where the field exists.

## Changes
- `packaging/Makefile`: `aleph-message==1.4` becomes `==1.5`, which is what `pyproject.toml` pins and what the code is written against.
- `tests/test_packaging_pins.py` (new): parses both dependency lists and fails on any package the two spell differently. The Makefile already asked for this in a comment ("Versions below MUST mirror pyproject.toml"); a comment did not hold. 13 packages are shared today.

## Test plan
- The new test fails on the drifted Makefile with `aleph-message is '==1.4' in the deb and '==1.5' in pyproject`, and passes on the fixed one (checked both ways).
- The real proof is a deb install: the droplet legs build and install the package, and the next aleph-testnets run on this branch exercises a V-PROGRAM create end to end.
- `ruff check`, `ruff format --check`, `isort` clean on the new file.


foxpatch-aleph (approved):
The fix is correct and minimal, and I verified both directions: pyproject.toml pins aleph-message==1.5 while the deb now ships the same, and the code genuinely reads content.gpu (vprogram_launch.py:193, run.py:694), a field absent from 1.4, so the described AttributeError on every deb-installed node checks out. The new regression test passes on the fixed tree and, on a simulated drifted Makefile, fails with a clear actionable message. All test-coverage claims hold (13 shared packages confirmed). Two non-blocking observations below about the test's blind spots; neither warrants blocking a release-blocking fix.

**`tests/test_packaging_pins.py`** (line 54): The intersection-only comparison means the other half of the drift class goes undetected: a dependency added to pyproject.toml but never added to the Makefile list (11 packages are pyproject-only today) would pass this test silently, and that is arguably how the next content.<field> breakage would arrive — via a new or renamed package. The deb list is intentionally a subset, so a full containment check would need a maintained allow-list (e.g. deps provided by apt or pulled transitively: aiodns, systemd-python, aiohttp via aiohttp-cors, pydantic via pydantic-settings...). Worth a follow-up, or at least a comment here naming the intentional omissions so the asymmetry is documented.

**`tests/test_packaging_pins.py`** (line 21): This regex matches the first ^dependencies = [ block in the file, which works only because [project] precedes [tool.hatch.envs.testing], whose own dependencies = [ (pyproject.toml:88) also starts at column 0. If sections are ever reordered (pyproject-fmt or a refactor), the test would silently parse the hatch env list instead. Anchoring to the [project] table (e.g. match text after the first ^\[project\] header) or asserting the parse yields the known top-level entries would make this self-evident rather than order-dependent.

PR #1250, opened 2026-09-14
feat(agent): advertise the owner routes a client cannot assume

odesenfans:
## Summary
#1245 added `POST /control/machine/{ref}/start`, and nothing tells a client whether a node has it. Probing cannot answer the question either: the route answers 404 for a VM this node holds no record for, which is exactly what an older node answers for a route it does not serve.

`/status/config` already carries `api.allocations`, the version the scheduler negotiates on, so this uses the same object with a sibling field:

```json
"api": {"allocations": 2, "control": ["machine_start"]}
```

Names rather than a version number, because the owner-facing control surface is not uniformly ordered: a confidential route needs a TEE, a GPU route needs a card, so a node can have a later route and not an earlier one, and each route ships when it ships. A name means exactly one route and never changes meaning. `api.allocations` keeps its own shape: one client, one ordered protocol.

## Changes
- `views/__init__.py`: `CONTROL_API_FEATURES` next to `ALLOCATIONS_API_VERSION`, published as `api.control`.
- `docs/architecture/vm-lifecycle.md`: one sentence next to the start route saying where a client looks and why probing does not work.
- `tests/supervisor/test_status_config.py`: the new field is asserted; the allocations assertion is narrowed to its own key so the two grow independently.

## Test plan
- `test_public_config_advertises_the_control_routes_a_client_cannot_assume` asserts `machine_start` is listed.
- The existing allocation-protocol test still pins `api.allocations == 2`.
- `isort`, `ruff format --check` and `py_compile` clean on both files.

Future routes worth discovering get a name here. If the list grows past a few, or grows names that depend on hardware rather than on the build, `/about/capability` is the better home and this field should move there.


foxpatch-aleph (approved):
Small, correct, well-scoped change. The advertised name matches the route registered at supervisor.py:389, api.allocations keeps its exact shape so the scheduler's existing negotiation is untouched, and the empty-list-vs-missing-key distinction (new node with no control routes vs old node) is preserved, which matters for exactly the discovery problem this PR solves. Both tests pass locally, the doc sentence lands in the right place, and the "names, not a version" rationale is sound given the surface is hardware-conditional. Two non-blocking notes: nothing statically couples the advertised names to the registered routes, and the new test's "in" assertion won't catch accidental extra names.

**`src/aleph/vm/agent/views/__init__.py`** (line 551): Non-blocking: the name here is linked to the route registration in supervisor.py ("/control/machine/{ref}/start") only by convention. If a future route is conditionally registered (TEE, GPU), this constant will need to become a function of settings — the comment already anticipates that, and the PR description names /about/capability as the escalation point if the list grows, so this is fine as-is.

**`tests/supervisor/test_status_config.py`** (line 44): Minor: "machine_start" in body["api"]["control"] is future-tolerant but would not catch an accidentally advertised extra name. Pinning == ["machine_start"] would make the contract exact, though "in" is a reasonable choice too. Just something to reconsider if the list stays tiny.

odesenfans:
Closing: the capability field is an optimization, not a contract. A client has to implement the 404 fallback anyway for nodes older than this release, and once it has, trying POST /control/machine/{ref}/start and falling back to /control/allocation/notify on 404 is correct in every branch: a hash the node holds no record for reaches notify, which creates it, and an old node without the route reaches notify, which is the only way. Not worth a field that only saves a round trip.

PR #1251, opened 2026-09-14
fix(agent): a log viewer hanging up is the end of the stream, not a 500

odesenfans:
## What

A client closing the `stream_logs` websocket while the VM was still producing lines made the next send raise `ClientConnectionResetError` out of the handler, so the error middleware logged a full traceback and answered 500 on a connection nobody was reading. The testnets' `test_instance_logs` does exactly that on every run (seen in run 34891355482's TEE agent log).

## How

- `ConnectionResetError` and aiohttp's `ClientConnectionResetError` are the peer leaving: logged at info, the socket closed, no 500.
- The streaming loop runs under `contextlib.aclosing`, so the supervisor's gRPC stream is cancelled the moment the client goes instead of waiting for the garbage collector to finalize the generator.
- The websocket body moves into `_serve_log_stream` so one `try/finally` owns the close, in place of the five explicit `close(); return ws` pairs it had.
- `Supervisor.stream_logs` is typed as the async generator every implementation already is (the conformance test asserts as much), since `aclosing` needs `aclose()`.

## Test

`tests/supervisor/views/test_operator_stream_logs.py` calls the view directly with a mocked request and a stub websocket whose second log-line send raises `ClientConnectionResetError`, the exception aiohttp's writer raises in production once the transport is closing. It asserts the handler returns normally, the socket is closed, "went away" is logged at info with nothing at ERROR, and the generator's `finally` has run by the time the handler returns (only `aclose` achieves that; a collected generator is finalized on a later loop iteration).

A transport-level test cannot observe the disconnect: aiohttp's `TestServer` hardcodes `handler_cancellation=True`, so a dropped peer cancels the handler instead of letting a send raise, and a graceful client `close()` keeps draining the server's frames until the stream ends.

## Also

`runtimes/instance-rootfs/create-debian-1*-disk.sh` now fall back to the umu.se cluster hosts (laotzu, gemmei, saimei) when the download through `cloud.debian.org` fails. The redirector has been sending every client to `chuangtzu.ftp.acc.umu.se`, whose certificate expired on 2026-09-14 21:20 UTC, which killed every pytest job at the image download. No third-party mirror carries the cloud images, so the siblings are the only fallback. `curl` also runs with `-f` now, so an HTTP error page is a failure rather than something saved as the image.


foxpatch-aleph (changes requested):
The production change is correct and verified: with a production-style runner (handler_cancellation=False, the aiohttp default), a client hanging up mid-stream surfaces in the handler as aiohttp's ClientConnectionResetError, which subclasses ConnectionResetError, so the new except-clause catches it; aclosing deterministically finalizes the supervisor's generator even when the handler receives CancelledError on a hard abort; and ws.close() in the finally is safe because aiohttp swallows exceptions there. The AsyncIterator -> AsyncGenerator annotation matches the conformance suite's isasyncgenfunction assertions, and the 95 existing ABC/view tests pass. However, the new test is broken and fails deterministically (~107s in this environment): a graceful ws.close() keeps draining the server's frames during the close handshake (each frame resets the 0.2s per-read timeout), so the server never raises and the generator runs to natural exhaustion -- meaning stream_finalized is set by exhaustion, not by aclose, and the 'went away' assertion can never pass. Even a hard TCP abort cannot produce 'went away' under the test, because aiohttp's TestServer hardcodes handler_cancellation=True (test_utils.py:118), converting disconnects into CancelledError rather than ConnectionResetError. Recommend replacing the transport-level test with a direct unit test of _serve_log_stream using a stub ws whose send_json raises ClientConnectionResetError, or serving the app via AppRunner/TCPSite and aborting the client transport.

**`tests/supervisor/views/test_operator_stream_logs.py`** (line 80): This never simulates a disconnect the server can observe. In aiohttp 3.10.11, ClientWebSocketResponse.close() sends the close frame and then loops on self._reader.read() with a per-read timeout -- and every frame the server sends resets that 0.2s timer. Since the generator keeps sending every 0.01s, the client consumes the whole stream during its "close": the server never sees an error, the generator runs to natural exhaustion (~100s for 10_000 lines, so this test also burns CI time), and `await ws.close()` blocks that entire duration. Additionally, even a hard TCP abort (transport.abort()) cannot reach the `except ConnectionResetError` branch here, because aiohttp's TestServer hardcodes handler_cancellation=True (aiohttp/test_utils.py:118): the handler task is cancelled with CancelledError before a send ever raises. I verified this end-to-end: the test fails at line 85 after ~107s. Suggested fix: unit-test `_serve_log_stream` directly with a stub `ws` whose `send_json` raises `ClientConnectionResetError` on the second call, asserting (a) the exception does not escape, (b) the 'went away' record is logged at info, and (c) the generator's finally runs (aclose). Alternatively, serve the app with AppRunner/TCPSite (production semantics, handler_cancellation=False) and abort the client transport.

**`tests/supervisor/views/test_operator_stream_logs.py`** (line 82): As written, this assertion passes for the wrong reason: because the blocked ws.close() above consumes the entire stream, the generator is exhausted (not aclosed) by the time this runs, so the event fires via generator completion. It does not demonstrate that `aclosing` cancels the supervisor stream on early exit, which is one of the two behaviors under test. With a raising send in a direct test of _serve_log_stream, this would genuinely verify the aclose path.

**`src/aleph/vm/agent/views/operator.py`** (line 310): Worth a brief comment that this catches aiohttp's ClientConnectionResetError (raised by WebSocketWriter._send_frame / _drain_helper when the peer is gone), which works only because it subclasses ConnectionResetError in aiohttp 3.10.11. It's a subtle, version-dependent dependency that the whole fix hinges on; naming the concrete exception would help whoever debugs this after an aiohttp upgrade.

odesenfans:
Review addressed in 9d93553c, all three points confirmed against aiohttp 3.10.11.

- The transport-level test is gone. It now calls `stream_logs` directly with a mocked request and a stub websocket whose `send_json` raises `ClientConnectionResetError` on the second log line. It asserts the handler returns normally, "went away" is logged at INFO with nothing at ERROR, the socket is closed, and the generator's `finally` has already run when the handler returns (so by `aclose`, not by a later collection).
- The except clause now names `aiohttp.ClientConnectionResetError` alongside `ConnectionResetError` instead of relying on the subclass relationship.

foxpatch-aleph (approved):
The fix is correct and well-targeted: I confirmed the error middleware (src/aleph/vm/agent/supervisor.py:167) turns any escaping exception into a logged 500, that aiohttp 3.10's WebSocketResponse.close() swallows non-cancel exceptions so the new finally cannot mask the handled reset, and that aclosing deterministically runs the gRPC client's 'finally: call.cancel()' instead of waiting for GC. The AsyncIterator -> AsyncGenerator annotation change is annotation-only, the pinned aiohttp version guarantees ClientConnectionResetError exists, and the existing callers (tests/integration/test_vm_management.py, conformance tests) remain compatible. The new test passes locally (after satisfying the root-owned directories and kernel path that TESTING.md documents; it fails only in an unprivileged sandbox for environmental reasons). The extraction of _serve_log_stream genuinely simplifies the five close-and-return paths into one try/finally with no behavior change on the auth-failure paths.

**`src/aleph/vm/agent/views/operator.py`** (line 310): Nit: with aiohttp pinned at 3.10.11, ClientConnectionResetError subclasses ConnectionResetError, so the second tuple entry is redundant — except ConnectionResetError alone covers both. Keeping both is harmless and self-documenting, so non-blocking; alternatively catching the broader ConnectionError would also cover BrokenPipeError if you ever see EPIPE from the write path.

**`src/aleph/vm/agent/views/operator.py`** (line 315): Heads-up: the unified close log drops the old "closed (past logs for stopped VM)" variant that used to distinguish the stopped-VM branch, so anything grepping logs for that suffix loses a signal. Consider keeping a branch-specific message if operators rely on it.

**`tests/supervisor/views/test_operator_stream_logs.py`** (line 69): Minor: because operator.py does "import aiohttp.web as web", this patch rebinds the shared aiohttp.web.WebSocketResponse attribute process-wide for the test's duration rather than just operator's reference. mocker restores it so it's fine in isolation, just worth knowing if this file ever grows parallel websocket tests.

foxpatch-aleph (approved):
The fix is correct and well-tested: catching ConnectionResetError (of which aiohttp's ClientConnectionResetError is a subclass) turns a viewer hangup into a clean stream end, and aclosing deterministically cancels the supervisor's gRPC stream instead of relying on GC — the ABC retyping to AsyncGenerator is a necessary consequence and matches every implementation plus the conformance test. The refactor into _serve_log_stream with a single try/finally owning ws.close() is a strict improvement over the five close-and-return pairs. The shell mirror fallback is sound under set -e, and I verified all three umu.se hosts actually serve the hardcoded paths (with gemmei redirecting trixie to the broken chuangtzu host — a case the loop handles because curl verifies TLS). The new test is thoughtfully designed, particularly its assertion that the generator's finally ran before the handler returned, which distinguishes aclose from GC finalization. Remaining nits are non-blocking: the broad ConnectionResetError catch also covers supervisor calls (gRPC errors are translated, so mostly theoretical), the mocker.patch target string hides that it patches aiohttp.web globally, and a pre-existing "Download Ubuntu image" typo.

**`src/aleph/vm/agent/views/operator.py`** (line 310): Note the catch also wraps supervisor.get_vm/get_logs and get_message in _serve_log_stream. Today only websocket writes realistically raise ConnectionResetError here (gRPC errors are translated to SupervisorError by translate_rpc_error), but a raw untranslated socket reset from the supervisor side would be mislogged as "client went away" at info. Consider narrowing the try around the send loop if that distinction ever matters, or a comment noting the assumption.

**`tests/supervisor/views/test_operator_stream_logs.py`** (line 69): Because operator.py does "from aiohttp import web", this target resolves to patching the attribute on the shared aiohttp.web module itself, i.e. a global patch for the test's duration rather than a name in the operator module. It works and mocker reverts it, but writing it as "aiohttp.web.WebSocketResponse" (or patching web.WebSocketResponse via patch.object) would make the actual scope obvious.

**`runtimes/instance-rootfs/create-debian-12-qemu-disk.sh`** (line 19): Pre-existing nit, not from this PR: the echo says "Download Ubuntu image" on a script that downloads Debian 12.

PR #1252, opened 2026-09-15
docs: point the install instructions at the 2.1.0 release

odesenfans:
The README's Debian package download URLs still reference release 1.2.0. Bump them to the upcoming 2.1.0 so a fresh developer setup fetches a current package.

Discovered during the pre-2.1.0 release audit (stale install instructions were flagged as release-blocking).


foxpatch-aleph (approved):
Clean, mechanical version bump of the four Debian package download URLs (plus the instructional line) in the README from 1.2.0 to 2.1.0. I verified that all five stale references were updated with no stragglers elsewhere in the repo, that the asset filenames match those produced by .github/workflows/release-on-tag.yml, and that the bare-semver tag format in the URL matches the project's actual release tagging (as used by scripts/crn_upgrade.sh). No logic, security, or testing concerns for a docs-only change.

**`README.md`** (line 89): Since 2.1.0 is described as upcoming, these wget URLs will 404 until the release is published — merging this should be coordinated with (or immediately followed by) publishing the 2.1.0 release, otherwise a fresh developer setup breaks.

**`README.md`** (line 93): Out of scope for this bump, but the release workflow also builds aleph-vm.debian-13.deb and the README (like crn_upgrade.sh's OS detection) skips Debian 13. Worth a follow-up PR if Debian 13 is a supported target.

PR #1253, opened 2026-09-15
fix(agent): retire a VM for payment only after consecutive sweeps agree

odesenfans:
## What

The payment sweep retired a VM as GONE on the first sweep that found it short, while the same function waits for three consecutive confirmations before acting on a terminal message status. A `404` from `/api/v0/addresses/<addr>/balance` reads as a balance of zero, and under the 2.1 default `VOLUME_RETENTION=reap` a GONE retirement purges the disks at once. One misrouted API answer therefore deleted every hold-paid VM's disks on the node, irreversibly. Found by the 2.1.0 pre-release audit.

## How

- The three payment loops (hold, credit, stream) collect the VMs they would stop instead of retiring inline; a VM is retired only once `STOP_AFTER_CONFIRMATIONS` (3) consecutive sweeps named it, through the same helper the terminal-status branch now uses. A sweep that finds the VM covered clears its strikes.
- The stop order within an address is fixed to "most recently started first". It used to follow the supervisor's listing order, and a listing that changed between sweeps would have named a different VM each time, so no VM ever reached three confirmations.

At the default 60 s sweep interval a genuine shortfall is enforced after three minutes instead of one. The 30 s balance cache cannot span two sweeps, so a cached bad answer counts once.

## Tests

The existing shortfall tests sweep three times and assert nothing is retired before the third. Two new ones: a shortfall that clears mid-way resets the count, and the youngest VM is the one found short whatever order the supervisor lists them in.


foxpatch-aleph (approved):
The change is correct and well-targeted: the three payment loops now collect the VMs they would stop and retire them only via the shared _confirmed helper after STOP_AFTER_CONFIRMATIONS consecutive sweeps, and the fixed youngest-first stop order is the necessary companion fix (a supervisor listing that permutes between sweeps would otherwise never let confirmations accumulate). check_payment has a single caller (the monitor_payments loop), so the module-level strike dicts have no concurrency concern, and the clearing of strikes for VMs not found short gives the consecutive semantics their correct reset-on-top-up behavior. All 11 tests in test_checkpayment.py pass, and the two new tests cover the two failure modes the PR exists for. One non-blocking gap: _shortfall_strike_count entries are never pruned for VMs that leave the supervisor snapshot while short.

**`src/aleph/vm/agent/tasks.py`** (line 532): Non-blocking: strikes are only cleared for VMs still present in `infos`, so a VM that goes short for 1-2 sweeps and then leaves the supervisor listing (user stop, recreate, crash) keeps its entry in `_shortfall_strike_count` forever — a slow leak over agent uptime. More relevant to this PR's goal: since re-running the same program message produces the same item hash, a re-run VM would inherit its old strike count and could be retired after a single short sweep, where a fresh VM gets three. Consider pruning to the current snapshot here, e.g. `_shortfall_strike_count = {h: c for h, c in ... if h in current_ids}` or popping keys not in `infos`. (`_terminal_strike_count` shares the pattern, so this is a pre-existing weakness rather than a regression.)

**`tests/supervisor/test_checkpayment.py`** (line 204): Non-blocking: only the superfluid tier gets the confirmation treatment in tests; the hold and credit loops route through the same `short` dict and `_confirmed` helper so the risk is low, but if you want belt-and-braces, one hold- or credit-tier variant going through `_sweep_until_confirmed` would pin the shared path for all three loops.

PR #1254, opened 2026-09-15
fix(agent): the migration reaper only deletes export files

odesenfans:
## What

The startup migration reaper removed any VM directory holding a `.part` file unless the supervisor listed that VM. It runs as the first `on_startup` hook, before the registry is rehydrated and before the storage reconciler, and it consulted neither the registry, nor the `.reclaimable` markers, nor the create guard. A retained directory with a stale download beside it, or a VM the registry knows and the supervisor does not, was removed with its marker and its owner on the supervisor's word alone. Stale `.part` files are what older releases leaked, so a node upgrading to 2.1 is where this fired. Found by the 2.1.0 pre-release audit.

## How

The storage reconciler already handles that case with every guard this hook lacked: it runs after rehydration, sweeps `.part` and `.tmp` files older than `VOLUME_CREATE_GUARD` outside any namespace a create holds, judges an orphan directory against the registry and the supervisor together, and treats a marked directory as reclaimable rather than orphaned. So the reaper keeps the one job nothing else does, deleting orphan `*.qcow2.export.qcow2` files, and drops its directory removal and its `list_vms` call.

## Tests

The "removes an orphan dir with .part files" test is replaced by one asserting such a directory, marker included, is left alone; the hook test asserts it needs nothing from the supervisor.


foxpatch-aleph (commented):
The fix is correct and well-verified: the reaper ran before registry rehydration and would rmtree a namespace directory containing .part files on the supervisor's word alone, destroying retained volumes with stale downloads beside them. The storage reconciler, which now owns that job, does consult the registry, the supervisor, the create guard, and the reclaimable markers before removing anything, and its existing tests already cover the handoff cases (tests/supervisor/test_reconciler.py:946-960 for a migration import in flight, :479-493 for the part sweep). The retained export sweep is still needed and is the only job nothing else does. All four reaper tests pass. The one gap is documentation: docs/architecture/storage.md:271-277 still describes the removed rmtree-of-.part-directories behavior in detail and should be updated to describe the export-only reaper and the handoff to the reconciler.

**`docs/architecture/storage.md`** (line 272): This paragraph is now stale: it says the reaper `rmtree`s any namespace directory that is not a known live VM and still has `.part` files, but that pass is exactly what this PR removes. Please update it to describe the new behavior — the reaper deletes only stray `*.export.qcow2` files, and half-imported directories (`.part` files included) are left to the storage reconciler, which judges them against the registry, the supervisor, the create guard, and the reclaimable markers.

PR #1255, opened 2026-09-15
fix(agent): refuse at admission what the node's settings rule out

odesenfans:
## What

Nothing in the allocation plane read `ENABLE_CONFIDENTIAL_COMPUTING` or `ENABLE_GPU_SUPPORT`. A node with confidential computing off sized a confidential VM or a V-PROGRAM on memory and disk, answered the scheduler `accepted` (or `pending`, then fetched and admitted it), and had the create refused by the daemon at start. The plan still listed the VM, so the loop rebuilt it at the capped backoff for as long as the scheduler kept it there. A GPU request on a node with GPU support off went the same way, or was answered `gpu_unavailable`, which tells a scheduler to wait for a card that will never be handed out. Found by the 2.1.0 pre-release audit.

## How

- `ResourceRequirements` gains `confidential` and `wants_gpu`, read from the message: a V-PROGRAM and an instance with `trusted_execution` need confidential computing; a GPU requirement or a V-PROGRAM `gpu` block needs GPU support.
- `CapacityManager.simulate` refuses such a candidate as `unsupported` before any sizing, which covers both the plan verdict and `/v2/control/capacity/check`.
- `check_message` raises `UnsupportedWorkloadError` on every create path. It subclasses `InsufficientResourcesError` so every route that already answers an admission refusal answers this one; the allocation classifier publishes it as `unsupported`, and the HTTP mapper answers 400 with the reason instead of 503 "at this time".
- `doc/scheduler-allocations.md` lists the new refusal.

The retry policy of the loop is unchanged on purpose: a plan is the authority on what runs, so a listed VM stays owed an attempt. The point of this change is that the scheduler is told `unsupported` at push time and places the VM elsewhere, so the loop never sees it.

## Tests

`requirements_from_message` on a plain, a confidential and a V-PROGRAM message; `simulate` refusing a confidential candidate with TEE off and admitting it with TEE on; a GPU candidate with GPU support off refused as `unsupported` even with the card in inventory; `check_message` raising and then admitting; the classifier mapping. The V-PROGRAM disk-admission test now switches confidential computing on, since a V-PROGRAM is only sized on a node that can launch one.


foxpatch-aleph (changes requested):
The design is right and the wiring is correct — the exception hierarchy, the before-the-parent ordering in both classifiers, and the simulate() refactor all check out, and the scheduler-facing refusal is a genuine improvement over 503/gpu_unavailable. But the branch does not pass its own test suite: test_check_message_buckets_a_vprogram_as_an_instance now fails (a V-PROGRAM with ENABLE_CONFIDENTIAL_COMPUTING off, the default, raises UnsupportedWorkloadError where main passes), and the new test_check_message_refuses_what_the_node_has_switched_off leaks into the real storage pools for its second assertion, failing in isolation and in a full-file run here. Both are small, well-scoped fixes in the same file the PR already touches.

**`tests/supervisor/test_agent_capacity.py`** (line 1430): This existing test fails on this branch (passes on main): the content is a V-PROGRAM, so `check_message` now raises `UnsupportedWorkloadError` when `ENABLE_CONFIDENTIAL_COMPUTING` is off, which is the default. The PR fixed the same interaction in test_create_path_disk_admission.py:199 — this test needs the same treatment, e.g. `mocker.patch.object(settings, "ENABLE_CONFIDENTIAL_COMPUTING", True)`. Verified by running it: on main 1 passed, on this branch `UnsupportedWorkloadError: confidential computing is disabled on this node`.

**`tests/supervisor/test_agent_capacity.py`** (line 291): The second half of this test (`assert _manager().check_message(message) is None` with TEE on) falls through into disk admission, and the per-pool max-volume check reads real storage pools — `_patch_host` only stubs `_available_disk_bytes`, not `storage_pools.eligible_pool_free_bytes`. It fails in isolation and in a full-file run here with `required 10000 MiB, roomiest pool has 0 MiB` (passing in some orders only by accident of what earlier tests left behind). Follow the file's own convention from line ~696 ("stub it too or it hits the real filesystem") and add `_patch_pools(mocker, ...)`.

**`src/aleph/vm/agent/capacity.py`** (line 503): Non-blocking: `check_recreate` skips `unsupported_reason`, so a rebuild of a recorded confidential/GPU VM on a node whose operator turned the setting off still passes admission and dies at the daemon, keeping the crash-loop-backoff behavior the PR description sets out to avoid — just on the local rebuild path instead of the scheduler path. If deliberate ("a rebuild is not new load"), a line saying so in the docstring would help.

**`src/aleph/vm/agent/views/__init__.py`** (line 1119): Non-blocking: the reserve endpoint catches the parent `InsufficientResourcesError`, so a GPU reservation on a node with `ENABLE_GPU_SUPPORT` off answers 503 "cannot reserve the requested resources at this time" — the exact "at this time" semantics this PR moves the scheduler away from. Fine to leave (operator route, not scheduler), but a distinct 400 here would be consistent.

**`src/aleph/vm/agent/capacity.py`** (line 487): Nit: the unsupported check could run before `discounted_disk` (a storage-pool walk) — the settings check is the cheaper one and the refusal doesn't need the disk figures. Harmless as written.

**`src/aleph/vm/agent/run.py`** (line 805): The new 400-mapping branch has no test, while its sibling branches in `_raise_http_for_program_error` do (test_supervisor_run_routing.py:866). One case asserting `HTTPBadRequest` for `UnsupportedWorkloadError` would lock in the before-the-parent ordering that makes this branch reachable.

foxpatch-aleph (approved):
This PR is a well-scoped, correct fix for a real gap: nothing in the allocation plane consulted ENABLE_CONFIDENTIAL_COMPUTING or ENABLE_GPU_SUPPORT before sizing a candidate, so a node could answer 'accepted' for a VM its daemon would refuse, and the reconciler would then retry it forever at capped backoff. The design is sound: the requirements DTO gains two derived booleans, unsupported_reason() is judged before sizing in both check_message and simulate, UnsupportedWorkloadError subclasses InsufficientResourcesError so every existing refusal route inherits it, and the classifier and HTTP mapper place the new type before its parent so it is published as 'unsupported' rather than 'no room'. I traced every admission route (plan verdict, advisory capacity check, all three create paths, reserve, migration) and none turns the new error into a 500, and the simulate refactor is safe: the unsupported branch never touches the disk variable and correctly re-enters the record-commitment restore. Tests are thoughtful (asserting check_capacity is not called proves the before-sizing claim; the GPU-off test passes a card in inventory to prove the refusal is 'unsupported' rather than 'gpu_unavailable'), all new tests pass, and a base-vs-head run of tests/supervisor shows zero regressions. Remaining observations are minor: the recreate path and the reserve endpoint's 503 still answer the old way for a settings-ruled-out VM, and the new HTTP-mapper branch lacks a test.

**`src/aleph/vm/agent/capacity.py`** (line 490): Minor asymmetry with simulate: here the unsupported check runs after discounted_disk, which walks the storage pools when a VM hash is given, so a settings-ruled-out rebuild candidate still pays the directory walk before being refused. Moving the requirements/unsupported_reason pair above discounted_disk would skip it and keep the two admission paths symmetric.

**`src/aleph/vm/agent/capacity.py`** (line 503): check_recreate deliberately skips the unsupported check, so a rebuild of a TEE/GPU VM the registry already holds is still attempted after the operator flips the corresponding setting off, and fails at the daemon each time. Since compute_verdict answers 'unchanged' for held VMs, the push-time refusal never sees these either. This matches pre-PR behavior (not a regression), but it is a residual gap relative to the PR's goal — worth confirming it is deliberate, or a note for the settings-change runbook.

**`src/aleph/vm/agent/views/__init__.py`** (line 1119): Since UnsupportedWorkloadError subclasses InsufficientResourcesError, reserving resources for a GPU message on a GPU-off node is still answered with 503 "at this time" on this route — the exact "wait for room that will never come" semantics this PR removes elsewhere. It is handled (no 500) and pre-existing, but a scheduler acting only on this endpoint could be misled.

**`src/aleph/vm/agent/run.py`** (line 805): This new 400 branch has no test, while the neighbouring branches in _raise_http_for_program_error each have a test_program_error_mapper_maps_* case in tests/supervisor/test_supervisor_run_routing.py (see test_program_error_mapper_maps_file_too_large_to_bad_request). A one-liner asserting HTTPBadRequest with reason "Unsupported workload" would keep that pattern complete.

odesenfans:
Review addressed.

- The two blocking findings were the first CI run's failures, fixed in aab9ed2a (the V-PROGRAM bucketing test switches confidential computing on; the new check_message test stubs the sizing and asserts the refusal comes first, instead of reaching the pools).
- The four non-blocking notes are all taken, in the follow-up commit: `check_recreate` judges the settings the same way a first create does, the reserve endpoint answers 400 with the reason before the parent's 503, the settings check in `check_message` runs ahead of the pool walk, and `_raise_http_for_program_error` has a test for the 400 mapping and its before-the-parent ordering.

foxpatch-aleph (approved):
Solid, well-scoped fix. The unsupported-workload gate is placed before sizing on all three admission paths (check_message, check_recreate, simulate), the subclassing of InsufficientResourcesError is sound, and every new branch is correctly ordered before its parent check. I verified all existing except InsufficientResourcesError handlers, the simulate loop's GPU-pool/record-commitment accounting for refused candidates, the create-path record cleanup, and that requirements_from_message is the only production constructor of ResourceRequirements (so every path gets the new flags). Tests cover extraction, all three admission paths, the classifier and the HTTP mapper, and the required fixture flips for existing tests were made. No correctness, security or coverage issues found.

**`tests/supervisor/test_agent_capacity.py`** (line 273): Minor: this could also pin the V-PROGRAM's `wants_gpu`, since `content.gpu`-block detection is one of the two things the new field exists for — a fixture without a GPU block asserting `wants_gpu is False` (or True, if it has one) would keep that half of `requirements_from_message` honest.

**`src/aleph/vm/agent/capacity.py`** (line 359): Nit: `UnsupportedWorkloadError` living in capacity.py is fine (it's the only raiser), but note the deliberate asymmetry that the HTTP paths publish `str(error)` ("confidential computing is disabled on this node") while the scheduler-facing refusal uses the closed-vocabulary "This node cannot run this VM" — consistent with the stated no-details policy for public refusals, just worth keeping intentional.

foxpatch-aleph (approved):
Well-executed change: the unsupported-feature check is judged before sizing on every create path (check_message, check_recreate, simulate), the new exception subclasses InsufficientResourcesError so all existing refusal routes answer it, and every new dispatch site correctly precedes its parent-class check. simulate's restructured loop is safe (the disk variable is only read on the accept path, and a refused candidate no longer takes cards or walks pools), already-running VMs are still answered 'unchanged' rather than re-judged or torn down, and the fixed enum-derived response sentences avoid leaking host figures. Tests cover requirements extraction, both refusal paths, simulate with each flag off/on, the classifier mapping, and the 400 HTTP mapping, and the author correctly identified every pre-existing test that needed the flags patched. Only minor nits: requested_gpu_ids(content) is evaluated twice in requirements_from_message, and no test exercises the V-PROGRAM top-level gpu block setting wants_gpu=True.

**`src/aleph/vm/agent/capacity.py`** (line 223): Nit: `requested_gpu_ids(content)` is evaluated twice here (line 221 and again in the `wants_gpu` expression). Reusing the `gpu_device_ids` value computed one line above (`wants_gpu=bool(gpu_device_ids) or ...` via a local) would read a little cleaner and skip a redundant walk of `content.requirements`.

**`tests/supervisor/test_agent_capacity.py`** (line 260): The new `wants_gpu` branch for a V-PROGRAM `gpu` block (`bool(getattr(content, "gpu", None))` in requirements_from_message) has no direct test: this test only asserts `confidential` for a V-PROGRAM and `wants_gpu` for an instance `requirements.gpu`, and the mock-based test at line 1434 explicitly sets `content.gpu = None`. A small assertion that a V-PROGRAM message with a gpu block yields `wants_gpu=True` (and is refused `unsupported` with ENABLE_GPU_SUPPORT off) would lock in the branch the PR description calls out.

PR #1256, opened 2026-09-15
refactor(agent): the storage reconciler sweeps stale migration exports; drop the reaper

odesenfans:
Follow-up to #1254, which reduced the startup migration reaper to deleting orphan export files. Kept for 2.2 as agreed: it merges the reaper into the storage reconciler.

## What

The one case the reduced reaper still covered is an agent restart with an export outstanding: the runner's 30-minute TTL task dies with the process, the job table is in memory, and the export sits beside a live VM's disks, where no reconciler pass looks (the parts sweep only knows `.part` and `.tmp`). Every other way an export is removed is the migration module's own: the TTL task, the failed-export reset, and the cleanup route.

## How

- The parts sweep also removes `*.export.qcow2` files, on the export TTL rather than the create guard: a valid export legitimately waits up to 30 minutes for the destination, so ten minutes is not old enough. Counted as `exports_removed` in the pass report.
- `EXPORT_TTL_SECONDS` moves to `migration.jobs`, which the reconciler can import without a cycle (`runner` imports the reconciler for `creating()`).
- The reaper module, its startup hook and its test file are removed; the runner's comments and the two storage docs no longer describe it. A stale export is now also caught by the hourly pass, not only at startup.

## Tests

A live VM's directory holds an export older than the TTL, an export older than the create guard but younger than the TTL, and its real disk: only the first goes, and it is reported as an export, not a part.


foxpatch-aleph (approved):
Correct and well-tested simplification. The export sweep lives in the parts pass, which already walks live VM namespaces (only creating() is skipped), and applies the export TTL (1800s) rather than the create guard (600s) — exactly matching the runner's own TTL, so nothing the runner would keep is deleted. The constant move to migration/jobs.py is cycle-free (jobs imports only conf). I verified no leftover reaper references, that the race between the runner's TTL task and the reconciler's unlink is covered by the OSError handler, and ran the full supervisor + migration suites on both this branch and base: failure lists are identical, so the PR introduces no new failures (the pre-existing ones are local environment issues with aiohttp/aleph_message versions). Docs are accurate. Only cosmetic nits remain.

**`src/aleph/vm/agent/vm/reconciler.py`** (line 533): Nit: this now logs "Removed stale partial download" for migration exports too, which aren't downloads. The file path makes it identifiable, but a distinct log message ("Removed stale migration export %s") would save an operator a double-take when debugging a migration that lost its export.

**`src/aleph/vm/agent/vm/reconciler.py`** (line 534): Nit: re-deriving the export-vs-part classification from the filename works (the glob guarantees the suffix), but the code already knows which glob produced the entry — carrying the classification alongside the (path, age_guard) pair instead of re-testing the name would be marginally more robust if a future pattern overlaps.

**`src/aleph/vm/agent/migration/runner.py`** (line 59): Nit: the reflowed sentence reads a bit awkwardly ("a complete orphan a failed import left, which the storage reconciler has not purged yet"), and this line is ~108 chars where the surrounding wrap is ~75. Cosmetic only.

PR #1257, opened 2026-09-15
refactor(tee): pure-Rust verifier, openssl out, sev 8 crypto_nossl

odesenfans:
## Why

`aleph-tee` is heading to crates.io so the published aleph-sdk can consume its TDX (and later SNP) verifier. The SDK is deliberately openssl-free (rustls on ring, `sev` 8 with `crypto_nossl`) because it cross-compiles, and the measured guest agent has no business carrying a vendored OpenSSL either. This PR moves every signature and certificate check in the crate onto x509-parser (structure) and ring (math), and bumps `sev` to 8 with `crypto_nossl`.

## What changes

- `pki.rs`: parse, PEM split, chain signature (RSASSA-PSS/SHA-384 for AMD, ECDSA P-256/P-384 for Intel), raw `r || s` ECDSA, pinned-root comparisons, validity windows, single-Common-Name extraction. The outer signature algorithm is now bound to the one inside the signed bytes (RFC 5280 4.1.1.2), a check openssl performed implicitly. Test certificate fabrication moves to rcgen.
- `tdx/*`: same checks, same error strings; certificates travel as DER between steps.
- `sev_snp/verify.rs`: same checks; the report's little-endian P-384 scalars are rebuilt into ring's fixed-width form with an explicit range check, and the VCEK key must now be on P-384.
- New genuine Milan fixtures (report + VCEK, from the `sev` crate's test data via aleph-rs) so AMD's real RSASSA-PSS chain and P-384 report signature are exercised offline. Before this PR only an ignored network test touched a real AMD chain.
- The guest agent's lock drops openssl; golden measurements re-seeded.

## Verification

- `cargo test -p aleph-tee`: 147 passed, 2 ignored (network).
- Both workspaces: fmt, clippy `-D warnings`, test.
- `grep openssl` on both locks: nothing.

Stack: this is PR 1 of 4 (next: Intel PCS collateral client, `guest`/`verify` features, publication).


foxpatch-aleph (approved):
Solid, security-conscious openssl removal. I independently reproduced the cryptographic checks against the shipped fixtures: the Milan report's P-384 signature over bytes 0..0x2A0 with LE zero-padded 72-byte scalars validates under the fixture VCEK (confirming p384_signature_from_le and verify_raw_ecdsa), AMD's RSASSA-PSS profile (SHA-384, MGF1-SHA-384, salt 48) matches ring's RSA_PSS_2048_8192_SHA384, the Intel PCK chain and CRLs are all ecdsa-with-SHA256 on P-256 (inside the accepted set), and the TDX quote/QE-report signatures validate as well. The new RFC 5280 4.1.1.2 outer/inner AlgorithmIdentifier binding is a real improvement over openssl (which verified per the unsigned outer field), the narrow algorithm allowlist fails closed, and the CRL/validity-window semantics preserve the old boundaries exactly. Genuine-Milan fixtures close a real coverage gap since report verification was previously only tested synthetically.

**`rust/crates/aleph-tee/Cargo.toml`** (line 32): Nit: the `tempfile` dev-dependency added in this PR is not used anywhere in the crate (no source or test file references it), but it does land in the lockfile. Drop it or wire up the intended use.

**`rust/Cargo.toml`** (line 68): Worth being aware that the `x509-parser` feature on rcgen is only needed by `cert_for_public_key` in the `#[cfg(test)]` testing module, yet rcgen is a regular dependency (used by src/x509.rs), so the feature rides along for every consumer of the crate once published. The existing comment explains why it exists, which is appreciated; just flagging the trade-off since this crate is headed to crates.io.

**`rust/crates/aleph-tee/src/pki.rs`** (line 225): Non-blocking observation: in the ECDSA branch the signer's key type/curve is explicitly validated via ec_public_key, but the RSASSA-PSS branch (line 215) hands whatever SPKI bytes the signer carries straight to ring, relying on ring to reject non-RSA keys with a generic 'is not signed by' error. Fail-closed either way, and the outer/inner binding check above ensures the declared parameters can't be swapped, so this is fine as-is; a key-type check would only improve the error message.

foxpatch-aleph (approved):
Careful, security-conscious port of every signature and certificate check from openssl onto x509-parser + ring, verified openssl-free across manifests, both lockfiles, and crate sources. The algorithm allowlist fails closed, the new RFC 5280 4.1.1.2 outer/inner AlgorithmIdentifier binding is a genuine hardening over openssl's X509_verify, the P-384 little-endian scalar rebuild is correct and well tested, and validity/CRL window semantics exactly match the previous behavior. The genuine Milan fixtures (properly licensed with Apache-2.0 and a provenance README) are a meaningful test-coverage improvement over the previously ignored-only AMD chain path. Two non-blocking nits below; neither is a correctness or security bug for the vendors in scope.

**`rust/crates/aleph-tee/Cargo.toml`** (line 32): The `tempfile = { workspace = true }` dev-dependency is unused: nothing in the crate's src/ or tests/ references tempfile or TempDir, and the comment above it describes only the `time` dependency. Suggest dropping it so it does not imply a role it does not play.

**`rust/crates/aleph-tee/src/tdx/certs.rs`** (line 75): Nit: the CRL revocation lookup now compares `revoked.raw_serial() == cert.raw_serial()` (raw DER INTEGER bytes), where openssl's `get_by_cert` compared serials as integers. If a CA ever emitted a non-minimal INTEGER encoding (e.g. redundant leading zero) for the same serial in the CRL versus the certificate, a genuinely revoked cert would be treated as not revoked (fail-open for revocation). Intel's DER is consistent so this is theoretical, but normalizing both serials (or comparing as integers) would make the check encoding-independent. Also worth confirming the re-seeded golden measurements via the CI measurement pipeline, since the binary changed with the openssl removal.

PR #1258, opened 2026-09-15
feat(tee): Intel PCS collateral client

odesenfans:
## Why

`verify_tdx_quote` takes a pre-assembled `TdxCollateral`; nothing fetched one. A client that consumes the crate from crates.io needs the crate to be a complete verifier, so this adds the Intel Provisioning Certification Service client.

## What

- `tdx::pcs`: `collateral_request` reads FMSPC and PCK CA off a quote's chain; `PcsClient::fetch` gets TCB Info, QE Identity, the PCK CRL and the root CA CRL (API v4), slices the signed JSON bodies out of the responses verbatim (the detached signatures cover the exact text), percent-decodes the issuer chains from the headers, and caches each item under its own `nextUpdate`. An expired cache is refetched; a failed refetch is an error.
- `fetch.rs`: the body cap and cache helpers, shared with the AMD KDS client.

## Verification

- Offline tests build PCS-shaped envelopes from the vendored Phala collateral and run the assembled result through `verify_tdx_quote` on the v4 fixture; the network paths are exercised against a closed loopback port.
- One ignored live test against Intel PCS.
- `cargo test -p aleph-tee`: 155 passed, 3 ignored; clippy and fmt clean; golden measurements re-seeded.

Stack: PR 2 of 4, on top of #1257.


foxpatch-aleph (approved):
A well-designed PCS client with a sound trust model: nothing fetched is trusted on transport, every piece is re-verified against the pinned Intel root, and the TCB walk independently cross-checks the platform FMSPC against the TCB Info, so the unverified-chain shortcut in collateral_request is safe as documented. The hand-rolled JSON slicer is correct through escape and depth handling, and the cache fails closed on torn, corrupt, or expired entries (window checks are re-enforced at verify time, so cache poisoning cannot extend validity). The shared fetch.rs refactor also fixes a latent bug where an empty KDS cache file was served as a valid cached certificate. Minor non-blocking notes: no request timeout on reqwest::get (consistent with the pre-existing AMD KDS code), non-atomic cache writes, u8::from_str_radix accepting a leading '+' in percent_decode, a duplicated 3-cert chain-length check, and no test for the corrupt-cache-file-as-miss path.

**`rust/crates/aleph-tee/src/tdx/pcs.rs`** (line 383): reqwest::get uses the default client, which has no request timeout — a hung PCS connection stalls fetch indefinitely. The pre-existing AMD KDS code has the same behavior, so this is consistent, but a Client with a timeout would be more robust for a verifier in the request path.

**`rust/crates/aleph-tee/src/fetch.rs`** (line 67): write_cache is non-atomic: a crash or two concurrent fetches can leave a torn file. It fails closed (torn JSON/hex/DER becomes a parse failure → refetch), so it is safe, but a write-to-temp-then-rename would avoid the churn.

**`rust/crates/aleph-tee/src/tdx/pcs.rs`** (line 185): u8::from_str_radix accepts a leading '+' (and other sign quirks), so an escape like "+a" decodes to 0x0a instead of erroring. Harmless since the decoded PEM chain is cryptographically verified afterwards, but checking both bytes are ASCII hex digits would be stricter.

**`rust/crates/aleph-tee/src/tdx/pcs.rs`** (line 69): This 3-cert length check duplicates PCK_CHAIN_LEN in certs.rs verify_pck_chain. Consider reusing the constant (and/or the error message) so the two can't drift.

**`rust/crates/aleph-tee/src/tdx/pcs.rs`** (line 517): The corrupt-cache-file path (a file that is non-empty but unparseable JSON, or invalid hex for a CRL) is not directly tested — only the empty-file case is covered in fetch.rs. A small test writing garbage to one cache file and asserting it is treated as a miss would pin that fail-closed behavior.

foxpatch-aleph (approved):
Solid, security-conscious addition. The trust model is right: the PCS client only obtains bytes, and every piece is verified downstream against the pinned Intel root (signature, window, CRLs), so neither the HTTP path nor the disk cache can become a trust shortcut. The verbatim body slicing logic is correct for well-formed JSON and any mis-slice would fail signature verification (fails closed). Cache currency checks mirror the verifier's own nextUpdate semantics with no off-by-one, and errors on cached entries degrade to a refetch rather than serving stale collateral. The AMD KDS refactor into fetch.rs is behavior-preserving apart from the beneficial shared 30s timeout. Remaining findings are nits: percent_decode's from_str_radix accepts a leading '+' so '%+2' decodes to 0x02 instead of erroring; the hand-rolled JSON scanner could be replaced by serde_json's raw_value feature; crl_is_current ignores thisUpdate; cache writes are non-atomic (benign truncated-read race); and the live network envelope path is only covered by the ignored test. Note: no Rust toolchain was available in the review environment, so the claimed test results (155 passed, 3 ignored) were not independently re-run.

**`rust/crates/aleph-tee/src/tdx/pcs.rs`** (line 185): u8::from_str_radix also accepts an optional leading '+' sign (u8::from_str_radix("+2", 16) == Ok(2)), so an escape like '%+2' silently decodes to byte 0x02 rather than erroring. Harmless here — the header arrives over TLS and feeds PEM parsing — but a strict hex-digit check would keep malformed escapes out.

**`rust/crates/aleph-tee/src/tdx/pcs.rs`** (line 90): Consider serde_json's raw_value feature (Box<RawValue>), which hands back a value's verbatim slice including whitespace without re-serializing — it would replace this hand-rolled string/depth walk entirely. The current scanner is correct and well tested (including the nested-decoy and brace-in-string cases), so this is only a simplification to weigh against the extra feature flag.

**`rust/crates/aleph-tee/src/tdx/pcs.rs`** (line 216): crl_is_current only checks nextUpdate; a cached CRL whose thisUpdate is in the future would be served from cache and only rejected later by the verifier's check_validity_window. Fails closed either way — just noting the asymmetry, a thisUpdate check here would surface the problem at fetch time with a clearer error.

**`rust/crates/aleph-tee/src/fetch.rs`** (line 90): write_cache is a plain write, so a concurrent process can observe a truncated file mid-write; the serde parse then fails and the entry is treated as a miss (benign, but a tmp-file + rename would avoid the refetch thrash). Consistent with the pre-existing KDS behavior, so not blocking.

**`rust/crates/aleph-tee/src/tdx/pcs.rs`** (line 382): The live envelope path — get_bytes -> header_chain -> slice -> cache write — is only exercised by the ignored network test; the default suite covers the pieces (extract_top_level_object, percent_decode, cache currency) but not their wiring. A loopback HTTP-server fixture would let cargo test exercise a realistic PCS envelope end-to-end without leaving the machine.

foxpatch-aleph (approved):
Well-designed Intel PCS collateral client with a sound trust model: nothing fetched is trusted until signature-verified against the pinned Intel root, all error paths fail closed, and URL/cache-key construction is injection-safe (hex-encoded FMSPC, enum CA values). I independently validated the hand-rolled JSON scanner's logic (strings, escapes, nested depth, decoy keys) and confirmed the cache-currency semantics (now <= nextUpdate) exactly match the downstream verification windows. The shared-client refactor of the AMD KDS code is behavior-preserving apart from the intended 30s timeout. Minor observations: freshly fetched documents are not checked for currency before caching (an expired doc is cached and only rejected downstream by check_document_window); the offline tests assume port 9 refuses immediately rather than black-holes; only the Platform PCK CA path has test coverage; and the PCK CRL issuer-chain header is mandatory even though verify_pck_chain does not consume that chain, which could needlessly break PCCS mirrors.

**`rust/crates/aleph-tee/src/tdx/pcs.rs`** (line 335): Unlike the cache-hit path, a freshly fetched document is never checked with document_is_current before being cached and returned. An expired-but-signed document from PCS is therefore cached, then rejected downstream by check_document_window, and the cache write is wasted since the next fetch will reject it. Not a bug — the verifier catches it and the failure mode is correct — but the asymmetry with the cache path deserves a comment, or the same currency check could be applied here for a clearer error.

**`rust/crates/aleph-tee/src/tdx/pcs.rs`** (line 289): Only the Platform CA path is exercised (the fixture chain is a Platform-CA one); the ca=processor query value and pckcrl-processor.json cache key have no test coverage. A small unit test on PckCa::from_common_name for the Processor CN, or a synthetic Processor-chain collateral_request test, would close the gap.

**`rust/crates/aleph-tee/src/tdx/pcs.rs`** (line 568): Port 9 on loopback usually refuses immediately, but on a host where connections are black-holed (or something listens on port 9) each of the four requests would burn the full 30s shared-client timeout before the test fails. Binding a socket and dropping it, or an explicit connect-failure assertion, would make the offline tests airtight; acceptable as-is.

**`rust/crates/aleph-tee/src/tdx/pcs.rs`** (line 412): header_chain makes the PCK CRL issuer-chain header mandatory, but verify_pck_chain never consumes pck_crl_issuer_chain (CRL signatures are checked against certificates taken from the quote's own chain). A PCCS mirror that omits SGX-PCK-CRL-Issuer-Chain would fail here unnecessarily — worth noting given the docstring advertises PCCS mirror support. Fine for Intel's own PCS, which always sends it.

foxpatch-aleph (approved):
A well-designed Intel PCS client with a sound trust model: everything fetched is treated as untrusted input that is re-verified against the pinned Intel root and the injected clock at verify time, so the cache cannot weaken the trust path. The verbatim JSON slicer is byte-correct (escape handling, depth tracking, no panic paths on multi-byte UTF-8), cache semantics are per-item nextUpdate windows with fail-closed refetch, and cache filenames are built from hex/fixed strings with no traversal risk. The sev_snp refactor into the shared fetch.rs module is faithful. Only minor nits remain; none are correctness or security bugs. Note: cargo was unavailable in the review environment, so the test suite was not re-run; this assessment is from full static reading of the diff and surrounding crate.

**`rust/crates/aleph-tee/src/fetch.rs`** (line 90): Non-atomic cache write: `std::fs::write` truncates then writes, so a concurrent verifier can observe a partially-written JSON cache file. It self-heals (serde parse failure is treated as a miss and refetched), but writing to a temp file and renaming would close the window for free.

**`rust/crates/aleph-tee/src/tdx/pcs.rs`** (line 185): `u8::from_str_radix` accepts a leading `+`, so an escape like `%+5` decodes to 0x05 instead of being rejected. Harmless since the decoded bytes feed strict PEM parsing, but a stricter hex check (or rejecting `+`/whitespace) would match the error message's intent. Consider also rejecting a stray trailing `%` explicitly, which currently falls out of the `get(i+1..i+3)` range check.

**`rust/crates/aleph-tee/src/tdx/pcs.rs`** (line 202): `document_is_current` gates only on `nextUpdate`, not `issueDate`: a cached document with a future issueDate is served from cache and rejected only later by `check_document_window`. That is fail-closed, so this is fine — just an asymmetry to be aware of if the cache ever grows a stricter role.

**`rust/crates/aleph-tee/src/tdx/pcs.rs`** (line 67): The rejection paths of `collateral_request` (chain length != 3, intermediate CN that is neither Intel PCK CA) have no direct tests, only the happy path on the v4 chain. Two small negative tests would pin the error behavior the way the rest of the module does.

**`rust/crates/aleph-tee/src/tdx/pcs.rs`** (line 566): The offline tests assume 127.0.0.1:9 is closed; on a host running the discard service they would fail with a confusing connection error. A comment already documents the intent, so this is just a heads-up for exotic CI environments.

**`rust/crates/aleph-tee/src/fetch.rs`** (line 25): `expect` panics inside library code if reqwest's client builder fails (e.g. TLS backend init). Extremely unlikely with static settings, and the prior `reqwest::get` path panicked internally too, so this is an improvement — just noting the library-side panic.

foxpatch-aleph (approved):
A well-scoped Intel PCS collateral client with a sound trust architecture: the module only fetches bytes, and every trust decision (pinned Intel root, detached signatures, validity windows) stays in the existing verifier, so the cache is not a trust boundary and every malformed-cache or extraction path fails closed. The hand-rolled verbatim JSON slicer is correct for well-formed envelopes and its failure modes are signature-check failures downstream; cache currency semantics match the verifier's inclusive window boundaries; the sev_snp extraction into fetch.rs is faithful (same cache layout and caps, plus a new 30s timeout that is an improvement). Tests cover extraction edge cases, percent-decoding, currency, cache hits with end-to-end verification, and expired-cache/failed-refetch behavior offline, with the live network path appropriately gated behind an ignored test. Findings are comment-level only: percent_decode accepts a '+' sign in %XX escapes via u8::from_str_radix, cache keys are not namespaced by base_url so two deployments sharing a cache dir can cross-serve (bounded by root pinning and nextUpdate), write_cache is non-atomic (torn reads fail closed but cause refetch churn), and a leftover 'the closure the brief sketched' comment in the tests reads as drafting noise. Could not run cargo (no toolchain available); review is static.

**`rust/crates/aleph-tee/src/tdx/pcs.rs`** (line 185): `u8::from_str_radix` accepts a leading `+`, so an escape like `%+a` decodes to `0x0a` instead of erroring. Harmless today (the corrupted PEM fails closed at parse), but validating that both characters are hex digits would make the reject path strict. Consider `hex_check` + `from_str_radix`, or `u8::from_str_radix` on bytes pre-checked with `c.is_ascii_hexdigit()`.

**`rust/crates/aleph-tee/src/tdx/pcs.rs`** (line 270): Cache files are keyed only by FMSPC/CA, not by `base_url`, so two `PcsClient`s pointed at different deployments (e.g. Intel and a PCCS mirror) sharing the default cache dir would cross-serve collateral. The impact is bounded — everything still verifies to the pinned Intel root and stays within its own `nextUpdate` — but a one-line note in the `PcsClient` docs (or folding the host into the cache file name) would prevent a future surprise.

**`rust/crates/aleph-tee/src/fetch.rs`** (line 90): `std::fs::write` is truncate-then-write, so a concurrent verifier reading the same cache file can observe a torn copy. It fails closed (parse/decode error → refetch), but a write-to-temp-then-rename would avoid refetch churn when several verifiers share a cache.

**`rust/crates/aleph-tee/src/tdx/pcs.rs`** (line 521): This comment references 'the closure the brief sketched', which is drafting-process context that will read oddly to future maintainers. Consider trimming it to just why a nested generic fn is used, or dropping it.

foxpatch-aleph (approved):
Well-designed PCS client with a sound trust model: collateral is never trusted on arrival (verified against the pinned root downstream), signed-document bodies are sliced verbatim via RawValue so the detached signature's bytes are preserved, and cache-currency semantics (inclusive nextUpdate) exactly match the verifier's own window checks. URL and cache-path construction is injection-free (hex FMSPC, fixed allowlists), the body cap and shared 30s timeout bound memory and time, and error paths fail closed (corrupt cache = miss, failed refetch = error, both tested). Two non-blocking nits: the tmp-file rename in write_cache can race between concurrent writers in one process (benign - degrades to a cache miss, never to unverified trust), and the ignored live test doesn't exercise the CRL wire format.

**`rust/crates/aleph-tee/src/fetch.rs`** (line 93): Nit: the tmp path is deterministic per process (path.with_extension("tmp.{pid}")), so two concurrent fetches of the same item within one process share the same tmp file; interleaved fs::write + rename can land torn content in the cache. It degrades safely (torn JSON -> parse failure -> treated as a miss and refetched, which a test already covers), but the doc comment's atomicity claim only holds for concurrent readers, not concurrent writers. A unique suffix (atomic counter or thread id) would close it, and would also avoid the *.tmp.<pid> litter a crashed process leaves behind.

**`rust/crates/aleph-tee/src/tdx/pcs.rs`** (line 667): Suggestion: the live test verifies the two signer chains and TCB Info currency but never parses the CRLs, so a wire-format drift in the v4 pckcrl endpoint (e.g. Intel returning base64 instead of raw DER) would pass this test while breaking production - and the offline tests use a seeded cache, so they can't catch it either. Running verify_tdx_quote over the fetched collateral would pin all four items end to end.

PR #1259, opened 2026-09-15
feat(tee): guest and verify feature gates

odesenfans:
## Why

The crate has two consumers with disjoint needs: the measured guest agent (report generation, key-bound certificate) and relying parties (chain verification, collateral, TDX). Publishing it means client SDKs would otherwise pull the guest device code, and the agent already carried the whole verifier closure into the measured binary.

## What

- `guest`: the SEV-SNP firmware backend. `verify`: pki, AMD and Intel chains, KDS and PCS clients, TDX, the certificate-parsing half of the X.509 extension helper; it also turns on sev's `crypto_nossl` for the builtin AMD roots. Both on by default. `tracing` stays always-on because the QEMU argv module logs through it.
- The agent depends on `guest` only: its lock loses aleph-tee's x509-parser 0.16, rsa, p384 and x509-cert (rcgen's own x509-parser 0.18 edge remains, it is the agent's TLS dependency, not the verifier's). The controller's dev dependency takes no default features.
- rcgen and time become dev-dependencies (they were test-only already).
- CI builds and lints each half on its own.
- Golden measurements re-seeded for the smaller agent build.

Stack: PR 3 of 4, on top of #1258.


foxpatch-aleph (approved):
The guest/verify feature split is correctly implemented: every ungated module was verified to be free of verify-side dependencies, the sev crate's snp-only base dep correctly covers the firmware/parser paths used by the ungated report module, both Cargo.locks are internally consistent with the new optional-dependency graph, and both consumers (agent with guest-only, controller with no features) only use ungated APIs. The change is security-positive, shrinking the measured guest TCB by dropping the verifier's crypto and HTTP closure. CI gates all three feature combinations with clippy, and the default-feature workspace build still runs the full test suite. Only minor nits remain: a now-dead workspace-level sev pin, one intra-doc link in traits.rs that dangles in guest-only rustdoc builds, and the rewritten agent TLS test asserting extension-value bytes without pinning the OID binding.

**`rust/Cargo.toml`** (line 66): This workspace-level `sev` pin is now dead config: aleph-tee inlines its own sev dep (rust/crates/aleph-tee/Cargo.toml:20) and no other crate inherits `sev.workspace = true`. The comment here still reads as the operative declaration ("the crate's `verify` feature turns on sev's `crypto_nossl`"), which could mislead a future reader into thinking this features list matters at the workspace level. Consider dropping the entry or trimming the comment to say the pin is kept only as documentation.

**`rust/crates/aleph-tee/src/traits.rs`** (line 16): This intra-doc link targets `crate::sev_snp::verify::verify_sev_snp_report`, which is compiled out under `--no-default-features --features guest`. Rustdoc for the guest-only build will report a broken link (rustdoc isn't part of CI so nothing gates it, but wrapping the reference in non-link prose or noting the verify feature would keep guest-only docs clean).

**`rust/crates/aleph-attest-agent/src/tls.rs`** (line 142): The byte-window search proves the extension value made it into the cert, but unlike the old `extract_attestation_from_cert` round-trip it no longer pins the OID binding — if `from_oid_content` were called with a different OID constant this test would still pass. Asserting the DER-encoded `ATTESTATION_OID` bytes (`06 08 2B 06 01 04 01 87 2B 01 01`) also appear in `cert_der` would close that gap without pulling the verify closure into the agent.

**`.github/workflows/test-rust.yml`** (line 83): Minor: the guest-only combination is clippy-compiled but its tests are never executed (and the no-features build is lib-target `check` only). A `cargo test --locked -p aleph-tee --no-default-features --features guest` line would be cheap insurance against a test that behaves differently without the verify closure.

foxpatch-aleph (approved):
Well-executed feature split along a real deployment boundary. I verified against the actual sev 8.0.0 source that the always-on code (report parser, qemu args, x509 encode/decode) needs only sev's snp feature, while crypto_nossl is exactly what drags in p384/rsa/sha2/x509-cert, so the verify -> sev/crypto_nossl mapping is correct. Gating is airtight: every reference to pki/fetch/certs/verify/tdx is confined to verify-gated files, gated tests are cfg'd correctly, both halves get clippy -D warnings and tests in CI, the agent's rewritten TLS test is valid (rcgen embeds the extension content verbatim, so the byte-window search works), and the golden measurements were re-seeded. The change is security-positive, shrinking the measured guest binary's dependency closure. Only minor nits remain.

**`rust/crates/aleph-tee/Cargo.toml`** (line 20): Nit: the `sev` version pin is now duplicated here and in rust/Cargo.toml. Since `verify = [..., "sev/crypto_nossl", ...]` works with workspace deps too, `sev = { workspace = true }` would keep the version pin in one place and avoid drift if the root pin is ever bumped. Non-blocking.

**`.github/workflows/test-rust.yml`** (line 82): Nit: the empty feature set is checked with `cargo check` (lib/bins only), while the two single-feature halves get `clippy --all-targets`, so test/bench targets of the no-features build are never linted. Using `cargo clippy ... --no-default-features --all-targets -- -D warnings` here would keep the three combinations consistent. Practically harmless since any always-on test code is still compiled by the guest and verify clippy runs.

**`rust/crates/aleph-attest-agent/src/tls.rs`** (line 142): The byte-window search is a valid replacement (rcgen's `from_oid_content` embeds the content verbatim inside the extension OCTET STRING), and the comment correctly notes the OID-based extraction path stays covered by the verify-side x509 tests. One observation: the substring assertion can't detect the extension being attached under a wrong OID, but given the guest side never parses certs, this is a reasonable trade-off.

foxpatch-aleph (approved):
Clean and well-executed feature split. Every use of ring, x509-parser and reqwest is confined to verify-gated modules, sev/crypto_nossl correctly rides verify (the always-on report parser only needs sev's snp feature; sev compiles with no crypto backend), the agent's lock genuinely drops the verifier closure without letting openssl in, and both consumers (guest-only agent, featureless controller dev-dep) only touch ungated APIs. The reworked tls.rs test (byte-containment of the DER extension plus a decode) is a sound replacement for the removed structural check, and the CI matrix plus the rustdoc-link downgrades show the feature edges were handled thoroughly. Only two doc-level nits, neither blocking.

**`rust/crates/aleph-tee/src/lib.rs`** (line 10): Minor doc inaccuracy: this says the "quote parsers" are always available, but the TDX quote parser (`tdx::quote`) is gated behind `verify` — only the SEV-SNP report parser (`sev_snp::report`) is unconditional. Worth rewording so the always-available list doesn't over-promise.

**`rust/Cargo.toml`** (line 66): This workspace-level `sev` pin is now dead config: after this PR no crate inherits it (aleph-tee declares `sev` inline at crates/aleph-tee/Cargo.toml:20), so editing it silently has no effect. Either drop the entry, or single-source it by having aleph-tee use `sev = { workspace = true }` and let the `verify` feature add `sev/crypto_nossl` — member features merge on top of inherited ones, so the split still works, and the version pin stays in one place as the agent Cargo.toml comment implies.

foxpatch-aleph (approved):
The guest/verify feature split is correct and internally consistent: I traced every cross-module reference and all uses of the pki/fetch/certs/verify/tdx closure live inside verify-gated modules, while always-on modules (types, traits, none, owner_auth, report_data, sev_snp::{qemu,report}, x509 encode/decode) reference nothing gated. Both consumers match the split (the agent's guest-only build uses only guest and always-on APIs; the controller's featureless dev-dep uses only types and sev_snp::qemu), and gating sev's crypto_nossl under verify is right since only the verify path touches the builtin AMD roots. The security outcome is a genuine win — the measured guest binary drops reqwest, ring, and X.509 parsing. CI now builds, lints, and tests each half independently, the rewritten agent test (byte-substring plus decode round-trip) preserves adequate coverage, and the golden-measurement re-seed follows the documented process. Two non-blocking nits: the now-orphaned workspace-level sev pin duplicates the version pin in two places (aleph-tee could have kept sev.workspace = true since sev/crypto_nossl is additive in the verify feature list), and the rcgen/time comment's 'shared with other workspace members' rationale is inaccurate — the real reason is that dev-dependency workspace inheritance requires the entries.

**`rust/Cargo.toml`** (line 66): The workspace-level `sev` pin is now inherited by no one — aleph-tee/Cargo.toml declares its own local `sev`, so the version/feature pin lives in two places and can drift. Keeping `sev = { workspace = true }` in aleph-tee would work identically, since the verify feature's `sev/crypto_nossl` is additive either way.

**`rust/Cargo.toml`** (line 76): Minor doc nit: no other workspace member uses rcgen or time, so "shared with other workspace members that use them" doesn't hold. The actual reason the pins must stay in [workspace.dependencies] is that aleph-tee's dev-dependencies inherit them via `{ workspace = true }`.

foxpatch-aleph (commented):
Well-executed feature split: I statically verified that no ungated module (none, owner_auth, report_data, sev_snp::{qemu,report}, traits, types, x509 encode/decode) references any verify-gated dependency or module, that the agent uses only ungated APIs plus the guest-gated SevSnpBackend, and that the controller's no-features dev-dependency is satisfiable. The agent lockfile slimming (rsa, p384, x509-cert, x509-parser 0.16 dropped; rcgen's 0.18 edge correctly retained as a runtime TLS dep) is consistent, the main workspace Cargo.lock correctly needs no regeneration (dev-dep reclassification doesn't change the lock graph), tracing is correctly unconditional (qemu.rs:97 logs through it), and the CI matrix (no-features check, clippy -D warnings + tests for each half) matches how the two consumers build. The golden-measurement re-seed is the expected consequence of the measured binary shrinking and follows the documented process. Only minor doc/comment issues: the lib.rs doc claims quote parsers are always available when the entire tdx module (including the quote parser) is verify-gated; the sev version pin is now duplicated between the workspace manifest and aleph-tee's Cargo.toml where workspace inheritance plus a sev/crypto_nossl feature dep would have avoided it; and the agent's Cargo.toml comment about inheriting all pins from the root manifest is now imprecise for sev.

**`rust/crates/aleph-tee/src/lib.rs`** (line 10): Doc inaccuracy: this claims the "report and quote parsers … are always available," but the TDX quote parser lives in tdx::quote and the whole `tdx` module is `#[cfg(feature = "verify")]`. Only the SEV-SNP report parser (sev_snp::report) is always available; quote parsing is verify-only. Since this PR is precisely about documenting the feature boundary, the crate-level doc should state it correctly.

**`rust/crates/aleph-tee/Cargo.toml`** (line 20): Minor: inlining `sev = { version = "8", ... }` duplicates the version pin that already exists at rust/Cargo.toml:66, and the two can drift. `sev = { workspace = true }` plus the existing `verify = ["sev/crypto_nossl", ...]` feature entry works fine with workspace inheritance, so the duplication isn't required. If the inline declaration is deliberate (e.g. to keep the guest half's sev feature set visible next to the feature gates), a one-line comment saying the root pin is the source of truth would help.

**`rust/crates/aleph-attest-agent/Cargo.toml`** (line 15): This comment ("aleph-tee (path dep) keeps inheriting from the root manifest") is now imprecise after this PR: aleph-tee's `sev` dependency is declared inline in its own Cargo.toml rather than inherited from the root [workspace.dependencies]. Worth a small tweak so future readers don't assume a single pin location.

foxpatch-aleph (approved):
Clean, well-executed feature split. I statically verified the gating is airtight: no ungated module references fetch, pki, tdx, sev_snp::certs, sev_snp::verify, or the ring/x509-parser/reqwest crates; the agent's imports (SevSnpBackend, report_data, x509 encode/decode, none, owner_auth, traits, types) and the controller's (sev_snp::qemu only) each match their feature sets; tracing is correctly unconditional because qemu.rs logs through it; rcgen/time are genuinely test-only; the doc-link de-referencing in traits.rs/backend.rs is the only breakage of that kind; and the lockfile deltas (agent drops rsa/p384/x509-parser 0.16/reqwest/ring, sev resolves with no crypto backend) confirm the claimed TCB reduction is real. The golden measurement re-seed is format-valid and consistent with the agent binary changing in every initrd flavor. Only minor nits remain.

**`rust/crates/aleph-tee/Cargo.toml`** (line 20): Nit: this inline sev declaration duplicates the identical workspace spec at rust/Cargo.toml:66. `sev.workspace = true` would work exactly the same here — the member's `verify` feature can still add `sev/crypto_nossl` by name — and would leave only one copy of the pin to keep in sync. Not blocking.

**`rust/Cargo.toml`** (line 76): The new comment says the rcgen/time pins stay at the workspace level "so the version is shared with other workspace members that use them", but after this PR aleph-tee is the only main-workspace member using either (the agent is a separate workspace and pins rcgen = "0.14" independently). Consider softening the comment to just say they are aleph-tee dev-dependency pins.

**`.github/workflows/test-rust.yml`** (line 82): Minor: the no-default-features configuration (what supervisor-controller's dev-dependency uses) only gets `cargo check`, not `--all-targets`/tests. The guest run compiles a superset of the ungated modules so the risk is mostly theoretical, but `cargo clippy -p aleph-tee --no-default-features --all-targets -- -D warnings` would give exact parity with how the controller consumes it.

foxpatch-aleph (approved):
A clean, well-executed feature-gate split. I statically verified the gate closure since no Rust toolchain is available here: every ungated module uses only always-on deps, and the gated deps (ring, x509-parser, reqwest, raw_value, sev::certs) appear exclusively in verify-gated code; SevSnpBackend is confined to the guest-gated backend module; the attest-agent consumes only guest/always-on items and the supervisor-controller only ungated ones; the rewritten agent test's hard-coded OID DER bytes are the correct encoding of 1.3.6.1.4.1.60000.1.1; and the lockfile surgery matches the expected dependency closure. The CI matrix lints all three feature combos with --all-targets and tests each half. Two minor nits below are non-blocking.

**`.github/workflows/test-rust.yml`** (line 86): Nit: the matrix runs `cargo test` for the guest-only and verify-only halves but not for the bare `--no-default-features` combo (it is only clippy-compiled via `--all-targets`). If the bare set ever gains a test-only regression (e.g. a test in an ungated module calling a gated item), clippy would catch the compile but the test would never execute. Cheap to add `cargo test --locked -p aleph-tee --no-default-features` for symmetry.

**`rust/crates/aleph-attest-agent/src/tls.rs`** (line 151): The hard-coded OID DER bytes are correct (verified: 60000 encodes as 83 D4 60), but they silently drift if `ATTESTATION_OID` ever changes — `ATTESTATION_OID` is already imported in this file, so a small helper or a comment tying the byte array to the constant would make the coupling explicit. Mitigated today by `test_attestation_oid_value` pinning the arcs, so non-blocking.

**`nix/golden-measurements.json`** (line 2): Re-seeding is an unavoidable consequence of the agent's dependency closure shrinking, and it was done last (after the review fixes). Worth confirming the nix-built measurements match these values in CI before merge, since they cannot be re-derived in this review environment.

foxpatch-aleph (approved):
Clean, internally consistent feature-gating PR. I statically verified that ring, x509-parser, reqwest and serde_json's raw_value are referenced only from verify-gated modules, that sev/crypto_nossl is only needed by verify-side code (sev::certs::snp::builtin), that the guest-side modules and their tests use only unconditionally-compiled dependencies, and that both consumers' feature edges (agent: guest-only; controller: no features) resolve against always-on modules only. The CI matrix plus the workspace-level build covers all four feature combinations. Golden measurements are re-seeded in the same PR as the measured binary's dependency closure changes, as the repo's own process requires. Only a minor test-level nit on a hardcoded OID DER encoding.

**`rust/crates/aleph-attest-agent/src/tls.rs`** (line 151): Minor nit: the OID's DER encoding is hardcoded here even though `ATTESTATION_OID` is already imported in this file (line 5). If the OID ever changes, this byte array silently becomes stale knowledge rather than following the constant. The bytes themselves are correct (I verified 60000 encodes to 83 D4 60), and pinning the wire format in a test is defensible, but deriving the DER from `ATTESTATION_OID` — or at least noting in the comment that this must be kept in sync with it — would remove the duplication. Non-blocking.

PR #1260, opened 2026-09-15
chore(tee): publish aleph-tee to crates.io

odesenfans:
## Why

aleph-sdk and aleph-types are published crates and cannot carry git dependencies, so for the SDK to consume the TDX (and later SNP) verifier, `aleph-tee` has to be on crates.io. It ships from this repository until the supervisor, runtimes and verification libraries move to their own project.

## What

- README, `readme`/`keywords`/`categories`/`include` metadata, a copy of the MIT licence in the crate directory.
- CI runs `cargo publish --dry-run` so a manifest that stops packaging fails on the PR.
- `publish-aleph-tee.yml`: on an `aleph-tee-vX.Y.Z` tag, checks the version against the manifest and publishes through crates.io trusted publishing (OIDC, no stored token).

## Release procedure for 0.1.0

1. Merge the stack.
2. First publication is manual (crates.io needs the crate to exist before trusted publishing can be configured): from `rust/`, `cargo publish -p aleph-tee` with a personal token.
3. On crates.io, crate settings, add the trusted publisher: repository `aleph-im/aleph-vm`, workflow `publish-aleph-tee.yml`.
4. Tag `aleph-tee-v0.1.0` on the merge commit; later releases are tag-only.

Stack: PR 4 of 4, on top of #1259.


foxpatch-aleph (changes requested):
The packaging workflow, trusted-publishing setup, version guard, and include list are all correct and well-engineered, but the crate cannot actually be published as written: the keyword confidential-computing is 22 characters and crates.io rejects keywords over 20 characters at upload time (server-side check in rust-lang/crates.io src/controllers/krate/publish.rs), while cargo performs no local keyword-length validation, so the new publish dry-run gate in CI passes while the real upload fails with an HTTP 400. Both the manual first release described in the PR and every tag-triggered release will hit this. Shortening or dropping that keyword is a one-line fix, after which this looks good to merge.

**`rust/crates/aleph-tee/Cargo.toml`** (line 9): This keyword is 22 characters and crates.io rejects keywords longer than 20 at upload time: the publish endpoint errors with an invalid keyword message (keywords must have less than 20 characters). Cargo performs no local keyword-length validation, so the `cargo publish --dry-run` gate in test-rust.yml passes while the actual publish fails with an HTTP 400. Confirmed against crates.io: no crate on the registry carries any keyword over 20 characters (confidential-computing, artificial-intelligence, asynchronous-programming all return zero results in the keyword API). Use a keyword of at most 20 characters (e.g. trusted-execution, sev, hardware-security) or drop it and ship four keywords.

**`.github/workflows/test-rust.yml`** (line 89): Minor nit on the comment: the dry-run verification build compiles with test cfg off, so a file missing from `include` is only caught if it is referenced from non-test code (intel_sgx_root_ca.pem yes, tests/fixtures/** no — those are only read by `include_bytes!` inside `#[cfg(test)]` modules). The include list is correct today, but the fixture case would slip past this gate; worth softening the comment or noting the dry run does not validate packaged tests.

**`rust/crates/aleph-tee/README.md`** (line 18): `sev_snp::qemu` (QEMU arg generation) is public, always-compiled API per the Cargo.toml comment, but is missing from this list and from the description. Worth a mention since the old description advertised it.

foxpatch-aleph (approved):
Clean, well-executed publishing PR. The Cargo.toml metadata is valid for crates.io (exactly 5 keywords, all within the 20-char limit; both categories are real slugs; description fits), the `include` list is complete — all files live under `src/**`/`tests/**`, including the `include_bytes!`-embedded Intel SGX root PEM and the test fixtures — and the LICENSE matches the repo root verbatim. The README example compiles against the real API (`fetch_collateral`, `verify_tdx_quote`, `TdxVerification.registers`/`report_data` are all accurate, and the documented caveat about caller-side pinning is correct). The publish workflow has minimal permissions, no stored secrets (OIDC trusted publishing), and a sensible tag/manifest version check; the CI dry-run placement is right. Only minor, non-blocking observations: (1) `release-on-tag.yml` matches `tags: "*"`, so an `aleph-tee-v*` tag will also fire the deb-build matrix and a draft GitHub release — worth excluding that prefix or noting in the release procedure; (2) the publish job gates only on the version match, not on tests having passed; (3) the README's `#`-prefixed hidden lines render literally on crates.io.

**`.github/workflows/publish-aleph-tee.yml`** (line 13): Heads-up: `release-on-tag.yml` already triggers on `tags: "*"`, so pushing `aleph-tee-v0.1.0` will also kick off the 5-OS deb build matrix and create a draft GitHub release named after the tag. Consider excluding `aleph-tee-v*` there (or being ready to delete the spurious draft release) so the release procedure in the PR description isn't surprising.

**`.github/workflows/publish-aleph-tee.yml`** (line 43): The publish job gates only on the tag/manifest version match — it never runs the test suite, so a tag pushed to a commit whose CI hasn't passed (or failed) will still publish. That's a common tradeoff for tag-driven releases (the tests run on the PR/merge), but if you want a stronger gate, a `cargo test --locked -p aleph-tee` step (or a CI-status check) before upload is cheap since the build is already warm.

**`rust/crates/aleph-tee/README.md`** (line 30): Cosmetic nit: unlike rustdoc, crates.io/docs.rs render READMEs literally, so the `#`-prefixed hidden lines (`# async fn run(...)`, `# Ok(()) }`) will appear verbatim in the published docs. Either that's fine, or restructure the example to a full `fn main`.

foxpatch-aleph (approved):
Solid, well-documented publishing setup. All aleph-tee dependencies resolve to registry versions (no git/path deps would block cargo publish), the include list covers every file in the crate including the third-party fixture licenses, LICENSE is an exact copy of the root MIT license, keywords/categories are within crates.io rules, and the workflow uses the official crates.io trusted-publishing action with least-privilege permissions and no stored secrets. The README example was verified against the actual API surface (parse_tdx_quote, collateral_request, fetch_collateral, verify_tdx_quote, TdxTcbPolicy::default, TdxVerification.registers/report_data) and matches. The CI publish dry-run gate is a good practice. Only minor nits: the grep-based tag/manifest version check is positionally fragile, and the README code example is never compiled by CI so it can drift as the pre-1.0 API changes.

**`.github/workflows/publish-aleph-tee.yml`** (line 32): Minor nit: this check works only while `[package]`'s `version = ` stays the first such line in Cargo.toml — the grep is positional, so a reordered manifest or a re-formatted dependency could silently compare against the wrong version (the `sev = { version = "8" }` line would match if reformatted to a multi-line table entry). `cargo read-manifest` piped through `jq -r .version` would be robust, but this is fine as-is today.

**`rust/crates/aleph-tee/README.md`** (line 23): Since the README documents the API as pre-1.0 and changing between minor versions, consider wiring the README into the crate docs (`#![doc = include_str!("../README.md")]` in lib.rs) so this example is compiled by `cargo test --doc` — right now nothing in CI type-checks it, so it will be the first thing to drift. Verified it is accurate against the current API.

foxpatch-aleph (approved):
Well-executed publishing PR: the tag/manifest version guard fails closed, trusted publishing via OIDC avoids long-lived tokens, permissions are minimal, and the CI dry-run guards the packaging manifest. The include list is not just cosmetic — unit tests load fixtures via include_bytes! from tests/fixtures, so tests/** shipping in the package is required for the packaged crate to compile, and losing it would fail the new dry-run step rather than release. Verified that all workspace dependencies are version-only (no path deps leak into the published manifest), that the grep for the version line cannot false-match the sev = { version = ... } line, and that every API symbol used in the README example exists in the source. Two non-blocking nits below.

**`.github/workflows/publish-aleph-tee.yml`** (line 44): Consider pinning rust-lang/crates-io-auth-action to a full commit SHA rather than the mutable @v1 tag: this workflow holds id-token: write, and a repointed tag could exfiltrate the OIDC token (blast radius is small since the token is scoped to publishing aleph-tee, but an auth action is exactly where a SHA pin is worth it).

**`rust/crates/aleph-tee/README.md`** (line 21): This example is never compiled: readme = "README.md" in Cargo.toml does not make cargo test run doctests from the README, and lib.rs does not include it via #![doc = include_str!("../README.md")], so the rust,no_run annotation is currently aspirational and the flagship example can silently rot. Adding the doc-include to lib.rs would keep it honest.

foxpatch-aleph (approved):
Well-executed publishing PR. The include list is complete (all 17 compile-time include_bytes! references live under src/** or tests/**, which are both shipped), the LICENSE is byte-identical to the repo root MIT license, the keyword/category metadata is valid (5 keywords = the crates.io max, both categories are real slugs), and I verified the README example's names and signatures against the actual source. I also confirmed rust-lang/crates-io-auth-action exists with a v1 tag exposing a token output, so the OIDC plumbing is correct, and the version-vs-tag check is fail-safe and runs before authentication. Two non-blocking observations below.

**`.github/workflows/publish-aleph-tee.yml`** (line 48): The publish step trusts the tagged commit without running the test suite or even a dry run first — it relies on the convention of tagging a green merge commit. Tag pushes require write access so this is acceptable, but running `cargo publish --locked --dry-run -p aleph-tee` (or the feature-matrix tests) before the real publish would close the loop for free.

**`rust/crates/aleph-tee/README.md`** (line 23): The example is accurate today (verified against `pcs.rs`, `quote.rs`, `tcb.rs`, `verify.rs`), but nothing compiles it: `lib.rs` doesn't embed the README as a doctest, so the snippet can silently drift from the API. Wiring it via `#![doc = include_str!("../README.md")]` would make CI catch drift — though it would then fail the guest-only CI leg since the example needs `verify`, so a feature-gated doctest elsewhere is the lighter option.

foxpatch-aleph (approved):
Solid PR. The packaging metadata is complete and correct: the include list covers every include_bytes! target (all fixtures live under src/** and tests/**), all dependencies resolve to registry versions so cargo publish won't fail on path deps, the tag-vs-manifest version check logic is sound, and the crates.io metadata (description, keywords, categories, licence) is within the registry's limits and is a faithful copy of the repo's MIT text. The trusted-publishing setup (OIDC, no stored token, minimal permissions, official Rust-org auth action) is a good security posture. The README example compiles against the current API — I verified every function, field and module path it references. Three non-blocking suggestions: the publish workflow could run the test suite (or at least the publish dry-run) before the real publish so a tag on a red/untested commit doesn't ship; the README snippet is written in doctest style (hidden '# async fn run' wrapper) but lib.rs never include_str!s the README, so it's never compiled and can silently rot; and for a workflow holding id-token: write and publish authority, SHA-pinning rust-lang/crates-io-auth-action would be stronger than tag-pinning (though this matches the repo's existing convention of actions/checkout@v4 etc.).

**`.github/workflows/publish-aleph-tee.yml`** (line 46): Non-blocking: this workflow publishes on any matching tag without running the test suite — the only gates are the version-match check and cargo's build verification. A tag pushed on a commit that never saw green CI (or that predates a fix) would ship untested. Consider a `cargo test --locked -p aleph-tee` step before the publish, since a tag event doesn't trigger test-rust.yml.

**`.github/workflows/publish-aleph-tee.yml`** (line 44): Nit: for a workflow with `id-token: write` and publish authority, SHA-pinning the third-party action would be stronger than `@v1` (a moved tag resolves silently). Tag-pinning matches the repo's existing convention for `actions/checkout@v4`/`Swatinem/rust-cache@v2`, so feel free to ignore.

**`rust/crates/aleph-tee/README.md`** (line 23): This snippet is written in doctest style (hidden `# async fn run(...)` wrapper, `no_run`), but `lib.rs` never does `#![doc = include_str!("../README.md")]`, so it is never compiled in CI and can silently drift from the API. Today it is accurate — I verified `collateral_request`, `fetch_collateral`, `parse_tdx_quote`, `verify_tdx_quote`, `TdxTcbPolicy::default`, and `quote.signature.pck_chain_pem` all exist with matching signatures. Either wiring it into the crate docs or dropping the doctest-style scaffolding would keep it honest.

foxpatch-aleph (approved):
Solid, carefully thought-out publishing setup. The manifest metadata is correct, the include list provably covers every include_bytes! fixture, the OIDC workflow uses the official crates.io action with a minimal permission set and a fail-safe tag/version check, and the README example matches the actual API signatures. The CI dry-run gate is the right defense against manifests that stop packaging. Only minor nits below.

**`.github/workflows/publish-aleph-tee.yml`** (line 44): Nit: consider pinning `rust-lang/crates-io-auth-action` to a commit SHA rather than the moving `v1` tag. This step runs with `id-token: write` in the release path, so for a project this focused on supply-chain integrity, an immutable pin is cheap insurance. (Same applies to `actions/checkout@v4` if the repo's other workflows don't already establish a convention.)

**`rust/crates/aleph-tee/README.md`** (line 23): The README example is not exercised anywhere: `lib.rs` has no `#![doc = include_str!("../README.md")]`, so `cargo test` doctests never compile it and it can silently drift as the API evolves. The signatures currently check out (verified against `verify_tdx_quote`, `fetch_collateral`, `collateral_request` and `TdxVerification`), but adding the `include_str!` doc attribute would keep it that way for free.

foxpatch-aleph (approved):
Well-executed publishing PR. The include list is correct and necessary (unit tests embed tests/fixtures via include_bytes!, and cargo's package verification compiles test targets, so the fixtures must ship); the CI dry-run closes the loop on packaging regressions before release; the publish workflow uses the official crates.io trusted-publishing action with a fail-safe tag/manifest version check; and the README example matches the real API exactly. I found no blocking issues.

**`.github/workflows/publish-aleph-tee.yml`** (line 19): Consider moving `id-token: write` from the workflow-level `permissions` down to the job (and to the `publish` job specifically). It works either way, but job-level scoping is the tighter default for a credential this sensitive.

**`.github/workflows/publish-aleph-tee.yml`** (line 44): The workflow pins `rust-lang/crates-io-auth-action` by tag, so a future v1 retag would silently run new code with `id-token: write` and crates.io publish rights. Consistent with `checkout@v4` elsewhere in the repo, so not blocking, but for a publish path a commit-SHA pin is worth considering.

**`.github/workflows/publish-aleph-tee.yml`** (line 32): The `grep -m1 '^version = '` check assumes the `[package]` table stays first in the manifest; if it ever stops matching, `manifest_version` is empty and the check still fails closed, so this is safe today. Just noting the assumption.

**`.github/workflows/publish-aleph-tee.yml`** (line 46): Optional: the publish runs no tests, only the tag-vs-manifest check. Since the procedure says to tag a merge commit that already passed CI this is fine, but if you want belt-and-braces, `cargo test --locked -p aleph-tee` before the publish step is cheap.

**`rust/crates/aleph-tee/README.md`** (line 14): The feature table doesn't mention `sev_snp::qemu` (QEMU arg generation), which is always-available like the parsers. Minor incompleteness in an otherwise accurate table.

foxpatch-aleph (changes requested):
The publishing machinery itself is solid: the include list correctly ships the include_bytes'd PEM and test fixtures, the README doctest matches the real API and is correctly feature-gated, the tag/version check and tag exclusion are right, and secrets handling (job-scoped id-token: write, SHA-pinned action) is exemplary. The one thing I cannot square is the golden-measurement re-seed: nothing in this diff should move a byte of the measured attest-agent binary (the doctest hook is cfg-stripped, README.md/LICENSE are not in agentSrc, and the Cargo.toml changes are package metadata only), yet all five launch measurements changed. Per the check script's own guidance, that pattern is reproducibility drift that should be investigated, not re-seeded. Please verify the cause of the measurement change before merging; everything else is minor.

**`nix/golden-measurements.json`** (line 2): The commit message attributes this re-seed to the README doctest hook, but I can't see how it moves the measurement. The hook is `#[cfg(doctest)]`-stripped before macro expansion — it must be, since `../README.md` is not in the measured source set (commonCargoSources keeps only .rs/.toml/lockfiles, per the flake.nix comment), so if the item were not stripped the agent build would fail on the `include_str!` target. The Cargo.toml changes (description/readme/keywords/include) are package metadata that does not reach codegen, and README.md/LICENSE are not in agentSrc at all, so the attest-agent binary should be byte-identical — and the measurement is a function of file content only (initrd.nix). All five values changing means the binary changed anyway. check-golden-measurements.sh's own failure message says: if nothing in the measured chain was meant to change, do not update the golden file, find the drift. Could you verify the actual cause — e.g. diff the built agent binary at base vs head, or check whether the previous golden was already stale against the pinned inputs (toolchain/nixpkgs)? If the hook genuinely moved the binary, that mechanism should be recorded in the commit message, since program bundles bake measurement.hex and unexplained drift here is exactly what this file exists to catch.

**`rust/crates/aleph-tee/README.md`** (line 9): Minor: `No OpenSSL, so the crate cross-compiles wherever rustls does` holds for the default/verify configuration, but the guest-only configuration — exactly how the measured agent consumes the crate — pulls openssl-sys through `sev` without `crypto_nossl`; nix/flake.nix builds a static openssl for the agent for precisely that reason. Worth qualifying so a `--no-default-features --features guest` consumer is not surprised by an OpenSSL requirement.

**`.github/workflows/publish-aleph-tee.yml`** (line 53): Minor: the tag-push gate re-tests only the default feature set, and tag pushes skip test-rust.yml entirely, so a regression that only manifests in a non-default feature combination (guest-only or verify-only) would publish to crates.io uncaught. Consider mirroring the two `--no-default-features` test invocations from the test-rust.yml feature matrix into this release gate.

foxpatch-aleph (approved):
Well-designed and carefully gated release infrastructure. The publish workflow uses least-privilege permissions (contents: read top-level, id-token: write scoped to the single publishing job), trusted publishing with no stored secrets, and pins the auth action to a full commit SHA which I verified is exactly the commit behind rust-lang/crates-io-auth-action v1.0.5. The tag/manifest version check is correct and fails safe on malformed tags, the !aleph-tee-v* exclusion in release-on-tag.yml uses valid negation syntax in the right order (and release-on-tag is the only other tag-triggered workflow), and the include globs in aleph-tee/Cargo.toml are correct under both legacy glob and current gitignore-style semantics, so the include_bytes!-embedded Intel root CA and the test fixtures both ship. The README doctest is the standard cfg(doctest) pattern and every API item in the example matches the actual signatures. The strip profile in aleph-attest-agent is honored (its own workspace root, built by crane from that manifest) and the golden measurement re-seed is CI-gated by the golden-measurements workflow on od/** PRs. Only minor nits remain.

**`rust/crates/aleph-tee/Cargo.toml`** (line 7): Nit: the new description drops generation, but the guest feature does generate reports over /dev/sev-guest (the old description said generation/parsing). Consider something like report generation and quote parsing so the crates.io listing still reflects the guest-side half.

**`.github/workflows/publish-aleph-tee.yml`** (line 22): Nit: consider adding a concurrency group keyed on the workflow (or on the tag). Two aleph-tee-v* tags pushed close together would race here; crates.io serializes versions so the worst case is a failed run, but a concurrency block makes that explicit.

PR #1261, opened 2026-09-15
fix(agent): parse V-PROGRAM contents when rebuilding the registry

odesenfans:
## Problem

`get_message_executable_content` tries `ProgramContent` then `InstanceContent` and nothing else, so every V-PROGRAM execution record is logged as "unparseable" when the agent rehydrates its registry after a restart. The V-PROGRAM keeps running under the supervisor, but the agent has no record for it.

Observed on a 2.0.1 node at its 2026-09-11 restart and again right after upgrading that node to 2.1.0, where the missing record now costs more:

- the storage reconciler skips its cache pass every hour: `ERROR Skipping the cache pass: 2 live VM(s) have no registry record (…)`
- `aleph-vm storage status` prints a pydantic traceback before its tables
- the VMs are outside owner-auth, the payment sweep and the allocation logic until they are recreated

## Fix

Try `VerifiableProgramContent` between the two. Every content model forbids extra fields, so exactly one of the three accepts a given dict; the order only decides which error is raised for a dict none of them accepts, which stays the instance one.

## Test

`tests/supervisor/test_executable_content_parser.py` parses each example message (program, instance, confidential instance, V-PROGRAM fixture) through the function and checks that what `persist_record` stores (`model_dump_json`) parses back to an equal model, which is the exact path `rehydrate_registry` takes.


foxpatch-aleph (approved):
Correct, minimal fix for a real production bug. The change inserts VerifiableProgramContent parsing between the existing ProgramContent and InstanceContent attempts in get_message_executable_content, and I verified the key invariant it relies on: all three models forbid extra fields and each has required fields the others lack, so at most one model accepts any given dict and insertion order cannot affect existing program/instance records. Error behavior for unparseable dicts is unchanged (the final unguarded InstanceContent.model_validate still raises), the ExecutableContent return type includes all three models, and the fix covers both consumers of the persisted message column (rehydrate_registry and the operator owner-auth fallback in views/operator.py). The new tests pass (7/7) and exercise the exact persist_record/model_dump_json round-trip path used after a restart; surrounding test files that use this function also pass. No security concerns — pure parsing change with no new input surface.

**`src/aleph/vm/utils/__init__.py`** (line 50): Nit (non-blocking): with the second fallthrough the pattern is duplicated; a loop such as `for model in (ProgramContent, VerifiableProgramContent, InstanceContent):` with a final raise would scale to future content types without adding a try/except per model. The explicit form is also fine — just flagging in case more executable content types are expected.

**`tests/supervisor/test_executable_content_parser.py`** (line 33): Nit: this parametrization omits the confidential instance example that test_parses_every_executable_content_type includes. Since it is still an InstanceContent the round trip is arguably redundant, but including it (or a comment noting why it's excluded) would make the symmetry between the two tests obvious.

foxpatch-aleph (approved):
Correct and well-scoped fix. I verified that all three content models (ProgramContent, InstanceContent, VerifiableProgramContent) have extra: "forbid", so at most one accepts a given dict — the docstring's invariant holds, and the try-order cannot misclassify. VerifiableProgramContent is part of the ExecutableContent union, so the return annotation stays accurate, and ValidationError subclasses ValueError, so the excepts work. The round-trip test exercises exactly the persist_record → rehydrate_registry path, and downstream consumers (AgentVmRecord.is_vprogram, persist_record) already handle the V-PROGRAM type. Error behavior for invalid dicts is preserved. The operator view also benefits, fixing the storage status traceback mentioned in the PR description. Only a minor test-coverage nit remains.

**`tests/supervisor/test_executable_content_parser.py`** (line 42): Minor: the round-trip parametrization omits `confidential_instance_message_from_aleph.json` even though the type test covers it — adding it here would keep the two tests consistent. Non-blocking.

PR #1263, opened 2026-09-16
fix(supervisor): give SEV-SNP confidential instances the instance IPv6 hextet

odesenfans:
## Summary

SEV-SNP confidential instances could be assigned the V-PROGRAM static IPv6 (`vm-type` hextet `4`) instead of the instance one (`3`) by the Rust supervisor. Seen in production: instance `abe65110c001...` on a static-policy node was reported as `2001:bc8:702:32d:4:abe6:5110:c001` instead of `2001:bc8:702:32d:3:abe6:5110:c001`. The Python agent (`compute_requested_ipv6` / `snp_instance_launch.py`) already uses `VmType.instance`.

## Root cause

`VmType::of_qemu` (world.rs) returned `VProgram` for any config where `snp().is_some()`, assuming SEV-SNP measured boot is the exclusive V-PROGRAM launch path. That stopped being true with the opaque-cmdline SNP arm for confidential instances (#1130). The doc comment claimed this could never misclassify an instance because the agent supplies the address, but the daemon still recomputes the static scheme when:

- adopting a config without a persisted `guest_ipv6_cidr` (legacy config, or one written under the dynamic policy),
- creating with an empty `requested_ipv6`,
- restart/start paths that use `VmEntry::vm_type()`.

The create path (`create_vm_type` in lifecycle.rs) had the same `snp => VProgram` rule.

## Fix

No new persisted field is needed. The two SNP arms are already distinguishable:

- **Persisted config:** `image_format` is written by the opaque-cmdline arm only. It was added in the same commit as that arm (17b2e2fb, #1130), so every persisted confidential-instance config has it, and the verity (V-PROGRAM) arm never writes it. `VmType::of_qemu` now returns `VProgram` only for `snp().is_some() && image_format.is_none()`. `VmEntry::vm_type()` and the adoption path both use it.
- **Create request:** a non-empty `tee.kernel_cmdline` marks the opaque arm (the same selector `snp_config_slice` uses). `create_vm_type` is now `VProgram` only for `snp && kernel_cmdline.is_empty()`. A new `debug_assert_eq!` checks that this matches `entry.vm_type()` on the written config, so the create and adoption classifications cannot drift apart.

I also corrected the doc comments that said this could not happen. The IPv6 persistence logic (`reproducible_on_adoption`) and the dynamic ordinal logic are unchanged.

## Testing

- `adopting_a_persisted_snp_config_gives_the_v_program_ipv6_hextet` is kept for the verity (V-PROGRAM) config. Its setup moved into a shared `adopt_persisted_snp_config` helper.
- New `adopting_a_persisted_snp_instance_config_gives_the_instance_ipv6_hextet`: adopts a persisted SNP config that has `image_format`/`image_readonly` and no `guest_ipv6_cidr`, and expects hextet `3`.
- New `create_snp_instance_allocates_the_instance_ipv6_hextet`: creates from an opaque-cmdline SNP spec with no `requested_ipv6`, expects hextet `3`, and checks that the written config classifies as `Instance`.

CI runs the tests. Locally I only ran `cargo check -p supervisor-daemon --tests` and `cargo clippy -p supervisor-daemon --tests` (both clean) plus `cargo fmt`. I did not run `cargo test`.


foxpatch-aleph (approved):
Correct and well-scoped fix. The root cause (VmType::of_qemu treating any SNP config as VProgram) is real, and the chosen discriminator is airtight: I verified that `image_format` is written at exactly one site (lifecycle.rs:1993) from `snp_slice.rootfs_format`, which the verity arm always returns as None (lifecycle.rs:2448) and the opaque arm as Some (lifecycle.rs:2309), with `build_written_config` as the only config producer. The codebase was already relying on this same invariant for GetVmSpec's cmdline echo (service.rs:758), so the fix reuses an established marker rather than inventing one. The create-path predicate (`snp && kernel_cmdline.is_empty()`) mirrors `snp_config_slice_with`'s arm selector, and the new `debug_assert_eq!` prevents the two classifications from drifting. All other callers (lazy recompute, guest_ipv4, adoption) route through `of_qemu` and inherit the fix. Python parity (`VmType.instance` for InstanceContent) and backward compatibility for pre-opaque-arm configs check out. Tests exercise both the create fallback and the adoption recompute, with assertions (`guest_ipv6_cidr.is_none()`, written-config round-trip) that genuinely pin the behavior. Only trivial nits below.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 2811): Nit: `snp_opaque_cmdline` only checks `!tee.kernel_cmdline.is_empty()` without also checking the backend is SevSnp, so the name slightly overpromises. It's behaviorally fine (the value is only consumed under `snp &&`, and `kernel_cmdline` on non-SNP backends is already rejected at lifecycle.rs:1815-1824), but folding the backend check in — or renaming to something like `opaque_cmdline` — would make the predicate self-contained. Non-blocking.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 2998): Good addition. One note for awareness: being a `debug_assert_eq!`, this drift guard is compiled out in release builds, so a future divergence between the create predicate and `of_qemu` would only be caught by the test suite. That matches the existing pattern (the adjacent `snp` assert), so this is fine — just something to keep in mind if this invariant ever becomes load-bearing in another direction.

odesenfans:
Closing: the supervisor should not derive guest IPv6 addresses at all, so there's nothing to classify. #1264 is being reworked so the agent owns all IPv6 allocation (static and dynamic) and the supervisor saves whatever address it is given. That removes `VmType`/`of_qemu` from the Rust side.

PR #1264, opened 2026-09-16
refactor: the agent owns guest IPv6 allocation, the supervisor stores it

odesenfans:
## Summary

Guest IPv6 allocation moves entirely to the agent, for both the static and the dynamic policy. The Rust supervisor now only validates, stores and serves the address it is given. The agent and the supervisor ship in the same release, so older agents are not supported.

Production bug behind this: with `IPV6_ALLOCATION_POLICY=dynamic`, a supervisor restart made the API report the wrong guest IPv6 for a running VM.
- Pool `2001:bc8:702:32d::/64`, prefix 124. An SEV-SNP instance was assigned `::b0/124` (guest `::b1`).
- After the restart, `/v2/about/executions/list` reported `::11`, while the VM, `vmtap8` and its per-tap dnsmasq still used `::b1`.
- The rebuilt dynamic ordinal could also hand a new VM a /124 that a running VM already used.

## Root cause

- Under the dynamic policy the daemon picked the address itself and did not persist it (`guest_ipv6_cidr` was only written when the agent supplied the address or the policy was static).
- After a restart, adoption recomputed the address from adoption order. That is not allocation order (VMs deleted in between, stopped VMs skipped, sorted by file name).
- The rebuilt ordinal restarted from the number of adopted running VMs, so a new create could collide with a live VM.

### Supervisor (Rust)
- **Required address:** a networked `CreateVm` must carry `network.requested_ipv6`; otherwise it fails with INVALID_ARGUMENT / `INVALID_BACKEND`. A malformed address, or one that disagrees with `ipv6_prefix_len`, is refused the same way.
- **Always persisted:** the address is stored as `guest_ipv6_cidr` under any policy.
- **Collision backstop:** a create whose requested network overlaps a subnet another entry holds is refused before any side effect, and a stale stopped entry is kept. "Holds" means the live address, or the persisted one for a stopped VM (`WorldView::ipv6_holder`).
- **Removed:** the static and dynamic derivations (`ipv6_static_assignment`, `ipv6_dynamic_assignment`), the dynamic ordinal and its seeding, `VmType` / `VmType::of_qemu` / `VmEntry::vm_type`, and the now-unused `WorldError` variants.
  - `IPV6_*` settings are still parsed, because they are copied into the controller config's settings slice and the startup check still reads them.
- **Adoption:**
  - A persisted `guest_ipv6_cidr` is used verbatim.
  - A legacy config without it uses the address configured on the VM's live tap `vmtap{vm_index}`, under any policy. This goes through the new `TapBackend::global_ipv6_address`, which reads `ip -6 -o addr show dev X scope global`. The address is kept in memory, so the rebuilt spec and a later start agree with it. The config file is not rewritten.
  - With neither, a running VM is hidden like a failed reattach.
- **Start / rederive without a known address:** a start, a lazy rederive or `RecreateNetwork` for an entry with no known IPv6 fails with a clear message. For a start, the check runs before the stop timestamps are cleared, so the VM stays STOPPED.
- **Idempotent create:** the requested IPv6 is compared as a network, not as text. The Python notation the agent sends and the Rust notation of a rebuilt spec (e.g. `0aa0` vs `aa0`) therefore match.

### Agent (Python)
- **Single create path:** every agent create now goes through `guest_ipv6.create_vm_with_ipv6`: `run.py` (program, instance, V-PROGRAM, on-demand program) and `migration/runner.py`.
- **Dynamic allocation** picks the first free `/IPV6_SUBNET_PREFIX` subnet of `IPV6_ADDRESS_POOL`, skipping:
  - subnet 0 (the host's);
  - every network a VM the supervisor lists holds. A stopped VM has no live address, so its address is read from `get_vm_spec`;
  - every allocation still in flight.
- **Allocation rules:**
  - Allocations are serialised by a lock, and each subnet stays reserved until its create returns or fails.
  - A VM the supervisor already knows keeps its address, so idempotent retries and rebuilds do not move it.
  - An exhausted pool raises `InsufficientResourcesError`.
  - No address is allocated for a Firecracker program without internet, for a spec that already has one, or when networking is off.
- **Static path:** unchanged (`compute_requested_ipv6` at spec build time; hextet 3 for instances including SNP instances, 1 for microVMs, 4 for V-PROGRAMs).
- **No other Python code on `main` derives guest addresses.** `supervisor_interface/configuration.py` only carries the settings schema of the controller config.

### A stopped legacy VM (config written before this change)

`start_persistent_vm` (the owner's `/control/machine/{ref}/start` and the scheduler path) now checks the VM's spec before resuming a STOPPED VM:
- If the VM needs a tap and its spec carries no address, the agent rebuilds the VM instead of resuming it: a delete with `RetireReason.RECREATE` (keeps disks and port forwards), then a fresh create that carries an agent-allocated address. A FAILED VM already went through this path.
- The supervisor's refusal stays in place as a backstop.
- Other in-place start callers (backup restore, reinstall, a failed migration export) only meet this case for a legacy VM that was already stopped across the upgrade. They then fail with the supervisor's explicit message.

## Docs and fixtures
- `docs/architecture/networking.md` is updated.
- The three Python-generated controller-config test fixtures now carry the `guest_ipv6_cidr` their static scheme gave them, noted in `test_fixtures.rs`. Legacy-config tests remove the field.

## Testing

- **Rust:** tests tied to the removed derivations were dropped or rewritten. New tests cover:
  - a networked create without `requested_ipv6` is rejected (QEMU and program), as are malformed or mismatched addresses; a create without networking needs none;
  - an overlapping requested subnet is rejected (live, stopped-persisted, and a wider network), and the stale entry is kept;
  - a restart adopts the persisted address under both policies;
  - a create after a restart cannot take an adopted VM's subnet;
  - a legacy running config adopts its tap address under the static policy, and a persisted address wins over the tap;
  - a legacy config with neither is hidden, and one adopted stopped fails to start with a clear error and stays STOPPED;
  - a retry in Python notation matches, and a legacy VM adopted from its tap matches a retry that carries that address.
- **Rust, run locally:** only `cargo check -p supervisor-daemon --tests`, `cargo clippy -p supervisor-daemon --tests` (both clean) and `cargo fmt --check`. None of the Rust tests were run locally; CI will run them.
- **Python:** new tests in `tests/supervisor/test_guest_ipv6.py` cover the first free subnet after the host one, skipping held networks (live, stopped, wider ones), ignoring networks outside the pool, reuse of an existing address, concurrent creates not colliding, a failed create releasing its reservation, pool exhaustion, and specs that need no allocation. The static policy never calls the supervisor. `test_supervisor_run_routing.py` gains a test for rebuilding a stopped legacy VM.
- **Python, run locally:**
  - `test_guest_ipv6.py` and `test_supervisor_run_routing.py`: all 48 tests pass.
  - The translate, run and migration test files also pass, except failures that happen identically on the base commit because this machine lacks `/var/cache/aleph` and `/var/lib/aleph` permissions.


foxpatch-aleph (approved):
A well-engineered fix for a real production bug. The root-cause analysis is correct (dynamic addresses were not persisted, and the adoption-order recompute both misreported live VMs and reset the allocator to hand out in-use subnets), and the layered fix — always persist the assigned /124, adopt it verbatim, fall back to the live tap address for legacy configs, and make the allocator skip held subnets while seeding its ordinal past every held one — is sound. I verified the ordinal/shift arithmetic (including the pool_len==0 mask special case), termination of next_free_dynamic_ipv6 under an all-held pool, lock ordering around the tap probe and world lock, that the same_spec_or_conflict clearing only affects adopted entries, and that all call sites of build_world_view/derive_tap_assignment were updated. The tests directly encode the production incident shape and the regression scenarios, including the defensive ordinal-rewind skip. Non-blocking observations: (1) subnets of active-but-hidden VMs (failed_reattach, duplicate vm_index) are seeded into build_world_view's local held_ipv6 but not carried into the WorldView, so a later create can still reuse one of those subnets; (2) same_spec_or_conflict clears current's ipv6_prefix_len to 0, so a retry with empty requested_ipv6 but non-zero prefix_len conflicts where a live-created VM would match; (3) GetVmSpec for adopted dynamic VMs now returns a non-empty requested_ipv6 where it previously returned empty — intentional, but wire-visible to agent consumers.

**`rust/crates/supervisor-daemon/src/world.rs`** (line 959): The held-set and the highest_held seeding only cover entries. build_world_view seeds its local held_ipv6 from ALL configs (line 692), including hidden ones (duplicate active vm_index, failed IP derivation -> failed_reattach, whose live tap carries a real /124), but that set is discarded. A create whose ordinal reaches a hidden VM's subnet can therefore still collide with it, since WorldView::held_ipv6_networks() only iterates entries. This is a narrow edge case and strictly better than the pre-PR behavior, but carrying those extra networks into the WorldView (e.g. an initial held set) would close it.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 1713): Minor asymmetry: when the request has an empty requested_ipv6, current's ipv6_prefix_len is cleared to 0. A retry carrying requested_ipv6: "" with ipv6_prefix_len: 124 would conflict against an adopted VM, whereas a live (non-adopted) VM created with that same spec would compare equal via the stored spec short-circuit. Presumably agents always send both empty, so this is theoretical — but comparing only requested_ipv6 in the daemon-picks case (or ignoring prefix_len when the address is empty) would be more forgiving.

**`rust/crates/supervisor-daemon/src/tap.rs`** (line 108): The scope-global check is redundant: the command already passes "scope global" as a filter, so every line ip prints has it. Harmless defensiveness against the filter being ignored; the parser is correct either way (verified the \\ line-continuation shape of ip -o output and the link-local/garbage cases).

**`rust/crates/supervisor-daemon/src/service.rs`** (line 771): Worth being aware this is a wire-visible change: under the dynamic policy, GetVmSpec for an adopted VM now returns a non-empty requested_ipv6 and ipv6_prefix_len: 124 where it previously returned empty. The create-path comparison handles it, but any agent consumer that inspects these fields under the dynamic policy will observe the change.

**`rust/crates/supervisor-daemon/src/world.rs`** (line 578): Verified the ordinal math here: ::b0 in fc00:1:2:3::/64 with prefix 124 gives 0xb0 >> 4 = 11, matching the production incident, and ::1:0 gives 0x1000 as asserted in dynamic_ordinals_are_computed_only_for_pool_subnets. The pool_mask special case for pool_len == 0 correctly avoids the 128-bit shift overflow.

foxpatch-aleph (approved):
Well-executed refactor that fixes a real production bug (dynamic IPv6 recomputed from adoption order after a restart). The agent-side allocator is sound: allocations are serialized by a lock, in-flight reservations are released in a finally block, existing VMs keep their address for idempotent retries/rebuilds, and the first-free-subnet search terminates in O(len(held)). The supervisor-side backstop correctly runs under the world lock before any side effect and preserves stale entries on refusal; the start-refusal gate runs before stop timestamps are cleared so the VM stays STOPPED; and the Python-vs-Rust IPv6 notation difference (0aa0 vs aa0) is handled by network-level canonicalization with tests on both sides. Tests are extensive and all 48 new/changed Python tests pass locally. Only minor issues found: a mangled sentence in docs/architecture/networking.md, a misleading comment in vm_spec_message, a backstop blind spot for hidden/failed-reattach VMs (pre-existing in severity), and the N+1 get_vm_spec calls per dynamic create.

**`docs/architecture/networking.md`** (line 38): This sentence got mangled in the rewrite: "turns `(vm_index, requested IPv6 network)` intocloud-init paths consume." — the middle of the sentence was lost. It should read something like "turns `(vm_index, requested IPv6 network)` into an IPv4 pair and the guest IPv6 pair the tap-creation and cloud-init paths consume."

**`rust/crates/supervisor-daemon/src/world.rs`** (line 396): The backstop only iterates `entries`: a hidden VM (failed reattach — bad vm_index, IPv4 derivation failure, or the new no-known-IPv6 case) is in `reserved_vm_indices`/`failed_reattach` instead, yet its live tap still holds a subnet. Neither this check nor the agent's `list_vms`-driven allocation sees such a VM, so a dynamic allocation could take its subnet. The old ordinal seeding had the same blind spot, so this is not a regression, but the doc comment "keeps two VMs from ever sharing a subnet" is stronger than the actual coverage — worth either consulting the failed-reattach taps here or softening the comment.

**`rust/crates/supervisor-daemon/src/service.rs`** (line 769): This comment is inaccurate: the create path's comparison is exact equality on the canonicalized spec, so a request with an empty `requested_ipv6` does NOT match an assigned one (it would hit `requested_ipv6_pair`'s refusal, or `AlreadyExists` on the idempotency check). In practice the dynamic agent always reuses the existing address via `allocate_dynamic_ipv6`, so the behavior is fine — only the comment misleads. Consider rewording to say a dynamic-policy agent reuses the assigned address on retries.

**`src/aleph/vm/agent/guest_ipv6.py`** (line 145): Minor performance note: `_network_of` issues one `get_vm_spec` RPC per VM without a live network (i.e. per stopped VM), per create, and this all runs while holding the allocation lock, so every dynamic create pays O(N) serialized RPCs on a node with many stopped VMs. Correct as written; just worth keeping in mind if create latency ever becomes an issue.

foxpatch-aleph (commented):
A well-executed refactor: the agent now owns guest IPv6 allocation for both policies, the supervisor validates/persists/serves the address, and the production restart bug (dynamic addresses recomputed from adoption order) is genuinely fixed. I verified the tricky parts by hand and by running the Python suites (48 tests pass): the overlap backstop runs under the world lock before stale-entry removal (so self-recreates and refused creates behave correctly), the subnet-skip math in _first_free_subnet is right including wider blockers, ipv6_mask/ipv6_networks_overlap have no shift UB, canonical_ipv6_spec makes Python vs Rust IPv6 notation compare equal, all agent create call sites route through create_vm_with_ipv6, and the three fixture addresses match the static scheme. Two non-blocking fixes requested: a garbled sentence in docs/architecture/networking.md (the edit dropped the middle of the sentence, leaving "intocloud-init paths consume"), and a stale comment in service.rs that describes removed behavior (a dynamic-policy agent now does send an address, and same_spec_or_conflict compares canonical specs rather than treating an empty address as matching any assigned one).

**`docs/architecture/networking.md`** (line 38): This sentence got mangled by the edit: "`derive_tap_assignment` (`rust/crates/supervisor-daemon/src/world.rs`) turns `(vm_index, requested IPv6 network)` intocloud-init paths consume." The original explained that `TapAssignment` wraps them with the device name and the CIDR-formatted host/guest address strings the tap creation and cloud-init paths consume — the middle of the sentence was dropped, leaving "into" glued to "cloud-init". Please restore something like: "turns `(vm_index, requested IPv6 network)` into an IPv4 pair and an IPv6 pair; `TapAssignment` wraps them with the device name and the CIDR-formatted host/guest address strings the tap creation and cloud-init paths consume."

**`rust/crates/supervisor-daemon/src/service.rs`** (line 768): This comment is now inaccurate on both counts: "A dynamic-policy agent sends no address; the create path's comparison treats that as matching any assigned one." After this PR, the dynamic-policy agent *does* send an address — `create_vm_with_ipv6` fills it in right before the create — and `same_spec_or_conflict` compares canonical specs directly, with no rule treating an empty `requested_ipv6` as matching any assigned one. As written it would mislead a maintainer into relying on a comparison behavior that does not exist. Suggest replacing with something like: "A dynamic-policy agent allocates the address right before the create (create_vm_with_ipv6), so a retry of an adopted VM carries the same network and compares equal once canonicalized; empty only for a legacy config that predates the persisted field."

**`src/aleph/vm/agent/guest_ipv6.py`** (line 55): Minor, non-blocking: the mutual exclusion is per event loop (`_allocation_locks` is keyed by loop) while `_in_flight` is process-global, so creates running on two different loops in one process would not be serialized against each other. The agent runs a single loop today so this is fine, but the per-loop `WeakKeyDictionary` machinery only buys test isolation — a plain module-level `asyncio.Lock()` would be simpler if multi-loop operation is out of scope. The line also exceeds the project's usual formatting width.

**`rust/crates/supervisor-daemon/src/lifecycle.rs`** (line 1701): Non-blocking observation: `requested_ipv6_pair` accepts any prefix < 127, so a buggy caller could request a network as wide as the entire pool (e.g. the /64 pool itself) and effectively monopolize it, with future creates failing on the overlap backstop. Since the agent is a trusted, same-release component and the overlap check contains the fallout to a clear error, I don't think a stricter prefix bound is required — just noting the boundary.

odesenfans:
Thanks for the reviews. Addressed in 22a3c612:

- **Garbled sentence in `docs/architecture/networking.md`**: restored.
- **Stale `GetVmSpec` comment in `service.rs`**: reworded. The agent always sends an address now, and the create path compares canonicalized specs.
- **Hidden VMs not covered by the overlap backstop**: fixed on the supervisor side. `FailedReattach` now records the network a hidden VM's live controller still holds (its persisted /124, else the address on its tap). `WorldView::ipv6_holder` checks it, so a create on that subnet is refused. This covers a VM hidden at adoption (Firecracker config, failed IPv4 derivation) and one hidden by a failed boot reconcile. The record goes away with the queue entry (re-adopt, delete, dead controller). Tests: `a_hidden_vm_keeps_holding_its_subnet`, `a_dequeued_hidden_vm_releases_its_subnet`, `a_create_overlapping_a_hidden_vms_subnet_is_rejected`. The doc comment and networking.md now state the remaining limit: `ListVms` doesn't report hidden VMs, so the agent can't skip their subnets, and a dynamic create landing on one is refused rather than moved. A duplicate-`vm_index` VM isn't recorded, since it shares its tap device with the adopted VM.

Left as is:

- **Per-loop lock instead of a module-level `asyncio.Lock()`** (`guest_ipv6.py`): deliberate. An `asyncio.Lock` binds to the first loop it waits on, and the test suite runs many loops in one process. `ruff format` accepts the line length.
- **`requested_ipv6_pair` accepts any prefix**: the agent ships in the same release, and the overlap check turns a bad request into a clear error.
- **One `get_vm_spec` per stopped VM for each dynamic create**: correct, and cheap at realistic stopped-VM counts. Worth revisiting only if create latency shows it.
- **Redundant `scope global` check in `tap.rs`**: kept as a harmless guard.
- **`GetVmSpec` now returns the address for adopted dynamic VMs**: intended. The agent is the only consumer and relies on it for idempotent retries.
- From the first review, the ordinal seeding and the empty-address `prefix_len` asymmetry are gone: the rework removed that code.

foxpatch-aleph (approved):
Well-executed refactor that fixes a real production bug (dynamic-policy address drift after a restart). The allocation responsibility cleanly moves to the agent; the supervisor validates, persists, and serves. The overlap backstop is checked under the world lock before any side effect, idempotent retries are preserved via except-vm carve-out and CIDR canonicalization, and all agent create paths were migrated to create_vm_with_ipv6. Test coverage on both sides is thorough and the fixtures were updated correctly (I verified the static-scheme hextets match the Python allocator's output). Two minor nits below, neither blocking.

**`rust/crates/supervisor-daemon/src/world.rs`** (line 749): This hide path inserts `FailedReattach::new(config.vm_index)` without `.holding(...)`, while the sibling hide paths (the IPv4 failure above and the Firecracker branch) record `hidden_vm_ipv6(...)`. In the case where the persisted `guest_ipv6_cidr` is non-empty but unparseable, `ipv6_from_tap` is never tried here even though `hidden_vm_ipv6` would consult it — so a hidden VM's live tap address would go unheld and a create could take its subnet. Consider `FailedReattach::new(config.vm_index).holding(hidden_vm_ipv6(settings, qemu.guest_ipv6_cidr.as_deref(), taps, config.vm_index))` for consistency (edge case: requires a corrupted persisted config).

**`rust/crates/supervisor-daemon/src/world.rs`** (line 517): The first five lines of this doc comment describe `ipv6_from_tap` but were merged into `hidden_vm_ipv6`'s doc; `ipv6_from_tap` (a `pub` function) now has no doc comment while `hidden_vm_ipv6`'s doc describes two functions. Split them.

**`src/aleph/vm/agent/guest_ipv6.py`** (line 207): Design observation, non-blocking: when the first free subnet happens to be held by a hidden VM, the supervisor refuses the create (agent cannot see hidden VMs in list_vms), so retries deterministically fail until the hidden VM is readopted or dies. Safety over liveness is a defensible and documented choice (networking.md), but it could turn into a permanent create-failure loop on a node with a stuck hidden VM; catching the supervisor's overlap error here and skipping to the next free subnet would make it self-healing.

PR #1265, opened 2026-09-17
fix(attest-agent): stream proxied responses instead of buffering them

RezaRahemtola:
## Problem

The attest agent's reverse proxy read the whole upstream response (`upstream_resp.bytes().await`) before sending anything to the client. Every V-PROGRAM request goes through this proxy, so streaming APIs break: with an OpenAI-compatible server (vLLM, llama.cpp) and `"stream": true`, the client receives all tokens at once when generation ends.

Measured on a V-PROGRAM running vLLM: first byte at 19.16 s, last byte at 19.16 s.

## Fix

Relay the upstream body chunk by chunk: `resp.streaming(upstream_resp.bytes_stream())` (enables reqwest's `stream` feature).

- Response status, headers and body bytes are unchanged. HTTP/1.1 responses are now chunked instead of carrying a `Content-Length`; upstream `Content-Length` was already dropped.
- Large responses are no longer held in agent memory.
- A client disconnect now closes the upstream connection, so the app can stop work nobody will read.
- Behaviour change: if the upstream fails mid-body, the status is already sent, so the client sees an aborted connection instead of a `502` JSON body. The error is still logged (`upstream response body failed mid-stream`).

## Dependencies

`Cargo.lock` gains `tokio-util`, `futures-io` and `futures-macro` from reqwest's `stream` feature. It also gains `wasm-streams`, which is wasm-only and not compiled for this target. `futures-io` is pinned to 0.3.32 to match the existing futures crates. `futures-util` (already in the tree) becomes a direct dependency, for `TryStreamExt::map_err`.

The measured initrd changes, so `nix/golden-measurements.json` is re-seeded. All five entries move, since every flavor embeds the agent. The values come from this PR's golden-measurements CI jobs, the same reference that `main`'s golden file matches.

## Tests

- New `proxy_relays_body_chunks_before_the_upstream_finishes`: a raw upstream sends one chunk and holds the connection open, and the first chunk must reach the client within 5 s.
- New `upstream_failing_mid_body_aborts_the_response`: the upstream closes mid-body, and the client gets a 200 followed by a body error, not a 502.
- Both fail on the old buffering code and pass with the fix.
- `cargo test --locked`: 53 passed. `cargo clippy --locked --all-targets -- -D warnings`: clean. `cargo fmt --check`: clean.
- Not yet validated end to end in a V-PROGRAM, since that needs a runtime built from this change.


foxpatch-aleph (changes requested):
The core change is correct, minimal, and well-tested: relaying the upstream body chunk-by-chunk via resp.streaming(upstream_resp.bytes_stream()) is the right fix, header handling remains sound, no client timeout exists that would truncate streams, and the new test genuinely fails against the buffering code. The blocking issue is process, not code: enabling reqwest's stream feature adds compiled dependencies (tokio-util, fuller futures-util) to the measured attest-agent binary, which moves the initrd launch measurement. The crate manifest explicitly requires re-seeding nix/golden-measurements.json in the same PR, and the golden-measurements workflow triggers on this path, so the PR as-is will report measurement divergence. Minor nits: mid-body upstream errors are now silent where they used to be logged, the test duplicates the AppState construction from the existing state() helper, and the documented mid-body failure path (aborted connection instead of a 502) has no test.

**`rust/crates/aleph-attest-agent/Cargo.toml`** (line 31): Enabling the stream feature adds compiled dependencies (tokio-util, and the fuller futures-util) to the measured agent binary, so the initrd launch measurement moves. The comment at the top of this file and the golden-measurements workflow (which triggers on rust/crates/aleph-attest-agent/**) both require nix/golden-measurements.json to be re-seeded in the same PR, but this diff does not touch it. Please run ./nix/check-golden-measurements.sh --update and commit the new measurements (this is the only blocking item).

**`rust/crates/aleph-attest-agent/src/proxy.rs`** (line 275): Minor observability regression: with the buffered body, a mid-body upstream failure was logged via tracing::error!; now actix's BodyStream silently discards the error when it aborts the connection. Consider wrapping the stream, e.g. upstream_resp.bytes_stream().map_err(|e| { tracing::error!("upstream body failed mid-stream: {e:#}"); e }), so operators keep visibility into upstream failures after the status is already committed.

**`rust/crates/aleph-attest-agent/src/proxy.rs`** (line 328): Nice test design: the single deadline covering both the handler and the first poll is exactly what makes it fail on the buffering code. Would be worth a companion test for the documented mid-body failure path: an upstream that errors (or closes) mid-body, asserting the client sees an aborted connection rather than a 502 body, since that is the one behavior change this PR introduces.

**`rust/crates/aleph-attest-agent/src/proxy.rs`** (line 358): Nit: this duplicates the AppState construction from the state() helper above. A small state_with_upstream(up: &str) helper (with state() delegating to it) would avoid the copy-paste and keep future AppState fields from being maintained in two places.

RezaRahemtola:
Thanks, addressed all four points in the same commit:

- Re-seeded `nix/golden-measurements.json` with `nix/check-golden-measurements.sh --update`. All five entries move, since every flavor embeds the agent.
- Mid-body upstream errors are logged again: the body stream goes through `map_err` with a `tracing::error!`. `futures-util` is already in the lockfile and is now a direct dependency.
- Added `upstream_failing_mid_body_aborts_the_response`: the client gets a 200 followed by a body error, not a 502. Like the streaming test, it fails on the buffering code.
- `state()` now delegates to `state_with_upstream()`. The raw chunked upstream and the single-deadline first-chunk read are shared helpers used by both tests.

`cargo test --locked`: 53 passed. Clippy (`-D warnings`) and fmt are clean.

foxpatch-aleph (approved):
Correct, minimal fix to a real problem: buffering the upstream body broke SSE/streaming workloads and held large responses in agent memory. The streaming relay is implemented properly (hop-by-hop filtering already in place, no client-side timeout or Compress middleware to defeat it, map_err logging compensates for actix silently dropping body-stream errors), the dependency changes are the standard way to enable it, and the golden-measurement re-seed follows the crate's documented same-PR procedure. The two new tests genuinely discriminate against the old buffering code via a raw chunked upstream gated behind a oneshot plus a 5 s deadline. Non-blocking notes: mid-body upstream failures now surface as an aborted connection instead of a 502 (disclosed and inherent to streaming), and the wasm-streams lockfile entries are wasm-only noise with no effect on the measured native binary.

**`rust/crates/aleph-attest-agent/src/proxy.rs`** (line 276): Verified the streaming relay is safe in context: the production reqwest client (main.rs) sets no total timeout, so long-lived SSE streams are not cut off, and the App wraps no Compress middleware that would buffer the stream. The map_err logging is the right pattern given actix drops body-stream errors silently. Non-blocking: mid-body upstream failures now abort the connection instead of returning a 502; that is inherent to committing the status early and is well documented in the PR description.

**`rust/crates/aleph-attest-agent/src/proxy.rs`** (line 397): Good test design: driving proxy_handler directly with a raw chunked upstream gated by a oneshot, under a deadline, correctly fails on the old buffering code (the handler would block in bytes().await while the upstream stays open). Non-blocking observation: since the handler is called directly, the wire-level relay through actix's server (chunked re-framing, per-chunk flush) is not unit-covered, so the pending end-to-end validation in a real V-PROGRAM mentioned in the PR is worth following up on.

**`rust/crates/aleph-attest-agent/Cargo.lock`** (line 1996): Informational: reqwest's stream feature pulls in wasm-streams, which is a wasm-target-only dependency; it is not compiled into the native/musl binary, so it adds no TCB or measurement impact, only lockfile churn. The golden-measurements re-seed moving all five entries is expected since every flavor embeds the agent, and matches the same-PR re-seed procedure the crate's Cargo.toml documents.

RezaRahemtola:
The first re-seed was generated on a machine whose builds don't reproduce CI's: that machine also gets a different `measurement` for unmodified `main` (`d6a8f559…` vs the committed `026e6314…`). Likely cause (unverified): its Nix 2.35 builds in randomized `/nix/var/nix/builds/nix-*` directories, and build paths end up in the agent binary. The golden file now holds the values the CI jobs computed for this exact tree.

foxpatch-aleph (approved):
Correct and well-executed fix: replacing the buffered upstream read with a chunk-by-chunk relay is the right approach, and the surrounding constraints hold — the reqwest client has no total-request timeout (which would have truncated streams mid-flight), no actix middleware re-buffers the body, hop-by-hop/Content-Length stripping is preserved so the framing-desync protections are intact, and reqwest's auto-decompression features remain off so body bytes relay verbatim. The two new tests are well designed: the deadline around handler-plus-first-poll fails on the old buffering code, and the raw socket/oneshot gate upstream tests real streaming semantics, including the mid-body abort behavior. Dependency changes are minimal and justified (futures-util with default features off is enough for TryStreamExt::map_err), and the golden measurements are re-seeded per the crate's documented rule that its lockfile is part of the measured TCB. The behavior change (mid-body upstream failure now aborts the connection instead of returning 502) is inherent to streaming and clearly documented in both the PR and the code comment.

**`rust/crates/aleph-attest-agent/src/proxy.rs`** (line 276): Minor note for future log archaeology: because `map_err` is only invoked when actix actually polls the stream, an upstream error is logged only if the client is still connected when it surfaces — if the client disconnects first, the body is dropped and the upstream failure goes unlogged. That's reasonable behavior (nobody was there to observe it), just worth remembering when grepping logs. No change requested.

**`rust/crates/aleph-attest-agent/src/proxy.rs`** (line 371): Since `proxy_first_chunk` returns only status, body and first chunk, the new tests don't exercise header forwarding on the streaming path. Consider threading the response headers out of the helper and asserting the upstream's `content-type: text/event-stream` survived — one line, and it pins the header-relay behavior (hop-by-hop stripping, Content-Length dropping) that the streaming rewrite now depends on. Non-blocking suggestion.

PR #1266, opened 2026-09-17
fix(attest-agent): keep upstream framing and stream request bodies through the proxy

odesenfans:
Stacked on #1265 (base branch `fix/attest-agent-stream-proxy`); merge that first, then retarget this one to `main`.

## Problem

Reviewing #1265 turned up four things the streaming relay left behind, all in `proxy.rs`:

1. **Re-chunked responses, and a stray terminator on 204/304.** The relay dropped the upstream `Content-Length` and handed every body to actix as an unsized stream, so every response went out chunked. For a 204 or 304 that corrupts the connection: actix's h1 encoder selects chunked encoding from the stream body *before* the status makes it skip the `Transfer-Encoding` header, then still writes `0\r\n\r\n` at end of body. The next response on that keep-alive connection starts with those five bytes (a CORS preflight or a DELETE followed by a POST, for instance).
2. **Invented `Content-Type`.** `HttpResponseBuilder::streaming()` inserts `application/octet-stream` when the upstream sent no `Content-Type`; the old `body()` path never did. Browsers prompt a download for a bare HTML/text response.
3. **Request side still buffered and capped.** The handler took `web::Bytes`, so every request body was held in agent memory and anything over actix's 256 KiB default got a 413 before the handler ran (a chat completion carrying a base64 image, for example). The `main.rs` comment saying the default service never reads `web::Bytes` was wrong. The body was also copied once per request (`to_vec()`).
4. **Duplicate error log.** actix-http already logs a body-stream error at error level in its dispatcher (`Response payload stream error`), so the `map_err` wrapper only doubled the line, and its comment ("actix drops the error silently") was wrong.

## Fix

- The upstream's declared framing decides the response body: a `Content-Length` (with no `Transfer-Encoding` next to it) gives a `SizedStream` with that length, anything else a `BodyStream`, and 204/304 finish with no body. Bytes still relay as they arrive.
- The body goes in through `body()`, not `streaming()`, so no header is invented.
- The request body is a `web::Payload` pumped through a bounded channel into `reqwest::Body::wrap_stream` (actix's payload is `!Send`, reqwest's body must be `Send`). It is framed with the client's own `Content-Length` when there is one, chunked when the client chunked, and no body is attached to a bodiless request. Nothing is buffered on either side any more.
- The `map_err` wrapper is gone. `futures-util` stays a direct dependency for the payload pump (`StreamExt`, `stream::poll_fn`), so `Cargo.lock` does not change.

Unchanged from #1265, worth stating since its description claims otherwise: a client disconnect does not promptly release the upstream request. With actix's default half-closed handling the dispatcher keeps polling the body after a read EOF, so the upstream is dropped only once a write to the client fails, and a silent upstream is held indefinitely. Not addressed here; it needs an idle guard on the reqwest client and deserves its own change.

Residual limitation: an upstream body without `Content-Length` (SSE, close-delimited) still reaches an HTTP/1.0 client as chunked. Sized bodies, which is what such clients ask for, now keep their length.

## Tests

- `upstream_failing_mid_body_aborts_the_response` now runs on the wire through a real actix server: the connection closes without a chunked terminator.
- `bodiless_statuses_leave_no_chunked_terminator_on_the_connection`: a 204 then a 200 on one keep-alive connection; the second response starts with its status line and carries `content-length: 5`, no `transfer-encoding`.
- `sized_upstream_body_keeps_its_content_length`, `no_content_type_is_invented_for_the_upstream`: header relay on the streaming path.
- `request_bodies_stream_to_the_upstream_uncapped`: a 1 MiB POST through the App reaches the upstream in full with the client's length; fails with a 413 on the previous code.
- The four `AppState` test literals collapse into one `app_state` constructor.
- `cargo clippy --locked --all-targets -- -D warnings` and `cargo fmt --check` are clean locally; the test suite runs in CI.

## Measurements

The agent binary changes, so `nix/golden-measurements.json` is re-seeded from a sandboxed build (all five entries move, every flavor embeds the agent).


foxpatch-aleph (approved):
Verified the PR's four behavioral claims against the actual dependency sources: actix-http 3.13.1's encoder really does write a stray chunked terminator after a bodiless head when handed a stream body (encoder.rs:72-77 forces BodySize::None for 204 yet the dispatcher's Chunked encoder still emits 0\r\n\r\n), actix-http does log 'Response payload stream error' at error level in its dispatcher (dispatcher.rs:653, 710) so dropping the duplicate map_err is correct, hyper honors a user-set Content-Length to frame streaming request bodies (role.rs BodyLength::Unknown path), and actix-http preserves Transfer-Encoding in the header map exposed to handlers so the chunked-request detection works. TestRequest::set_payload inserts Content-Length, so the 1 MiB uncapped test genuinely exercises the CL-framed path. The wire-level tests pin exactly the right behaviors and the code is correct. Non-blocking nits below.

**`rust/crates/aleph-attest-agent/src/proxy.rs`** (line 286): The `chunked` detection is the one new behavior without a test: it works because actix-http keeps `Transfer-Encoding` visible in `req.headers()` (verified in actix-http 3.13.1's decoder, which never strips it), but a regression there would silently drop chunked request bodies. A raw-socket test mirroring `counting_upstream` (client sends `transfer-encoding: chunked`, assert the upstream sees chunked framing and all bytes) would pin this.

**`rust/crates/aleph-attest-agent/src/proxy.rs`** (line 294): Nit: a request with `Content-Length: 0` now goes upstream with no body *and* no `Content-Length`, where the old `web::Bytes` path sent `content-length: 0`. Most servers treat the two identically, but a strict upstream can answer 411 on POST/PUT. Attaching the stream whenever a length is declared (including 0) would preserve the old wire shape; either way, worth a line in the comment.

**`rust/crates/aleph-attest-agent/src/proxy.rs`** (line 314): Arguably a fifth leftover in the spirit of the ones this PR fixes: `insert_header` replaces, and `HeaderMap::iter()` yields each value separately, so a multi-valued response header (e.g. two `Set-Cookie`, or `WWW-Authenticate`) collapses to the last value. `append_header` would relay them all; the hop-by-hop filtering here is per-name so the switch is safe.

foxpatch-aleph (commented):
The four fixes are real and correctly implemented: I verified the 204/304 chunked-terminator bug, the invented Content-Type in streaming(), actix-http's existing error-level body-stream log, and hyper's respect for a user-set Content-Length on a streamed body, all against the exact pinned versions (actix-http 3.13.1, actix-web 4.14.0, hyper 1.10.1). The framing decisions on both the request and response paths never let a client's Transfer-Encoding travel next to a length, so the smuggling posture is sound, and the new wire-level tests are a welcome upgrade. Remaining findings are non-blocking: one small behavior regression for chunked GET/HEAD/CONNECT request bodies, one regression test whose headline check is ineffective, and a pre-existing multi-value header collapse.

**`rust/crates/aleph-attest-agent/src/proxy.rs`** (line 294): Behavior regression for chunked GET/HEAD/CONNECT request bodies: with `chunked` true you attach the streaming body but set no framing header, and reqwest's body size hint is unknown. Hyper's `Client::set_length` (hyper 1.10.1, proto/h1/role.rs:1382-1394) then assumes `Encoder::length(0)` for GET/HEAD/CONNECT when neither a content-length nor transfer-encoding header is set ("if you *must* send a body, set the headers explicitly"), so the body is silently swallowed upstream and the payload pump stalls in `tx.send` until the request completes and drops it. The old `web::Bytes` path forwarded such a body with a derived content-length. Rare in practice, but easily fixed: when `chunked` is true, set `transfer-encoding: chunked` explicitly on `proxy_req` — hyper's client respects a user-set TE header for every method (role.rs:1349-1376).

**`rust/crates/aleph-attest-agent/src/proxy.rs`** (line 723): This test's headline check (no stray chunked terminator) is ineffective as written. In the buggy version actix encodes the 204 head and the stray `0\r\n\r\n` into the same write buffer and flushes once, so the first `read_until` receives both in a single `read()` and swallows the terminator into `first`; the `starts_with` assertion then passes anyway. The test does fail against the pre-PR code, but only because of the `content-length: 5` pin on the second response (old code re-chunked it), not the terminator check the test is named for. Suggest also asserting the first response's tail, e.g. `assert!(!first.ends_with(b"0\r\n\r\n"))` — deterministic, since head and terminator co-flush.

**`rust/crates/aleph-attest-agent/src/proxy.rs`** (line 314): Pre-existing (from #1265, unchanged here): `upstream_resp.headers()` yields each value of a multi-valued header separately, and `insert_header` replaces, so an upstream response with two `Set-Cookie` (or any repeated) headers reaches the client with only the last value. `resp.append_header((name.as_str(), v))` would preserve them all. Worth fixing while this loop is being touched.

**`rust/crates/aleph-attest-agent/src/proxy.rs`** (line 225): Nit on the comment: `actix_web::rt::spawn` is `tokio::spawn`, so the pump task is not pinned to "this worker thread" — it may be scheduled on any worker. The load-bearing property is already stated: the channel only transfers `Bytes`, giving the stream its `Send` bound.

odesenfans:
Review follow-up, in three commits on top:

- **Chunked GET/HEAD/CONNECT bodies** (5ff1a64a): confirmed in hyper 1.10.1 `Client::set_length`, the unframed stream is sent as length 0 for those methods. The proxy now sets `transfer-encoding: chunked` itself when the client chunked. `chunked_request_bodies_stay_chunked_upstream` covers it on the wire with a GET, which is also the chunked-request test asked for in the first review.
- **Repeated response headers** (5ff1a64a): `append_header`, with a two-`Set-Cookie` test.
- **204 test** (512e41a6): right, head and terminator co-flush, so the check was on the wrong read. It now asserts the tail of the first read.
- Golden measurements re-seeded on the final tree.

Two points not taken:

- **`Content-Length: 0` going upstream without a length**: it does not. The header is set for any declared length, zero included; only the stream is skipped, and with no body hyper returns early from `set_length` without touching a user-set Content-Length. Pinned on the wire by `an_empty_request_body_keeps_its_content_length`.
- **"this worker thread" comment**: `actix_web::rt::spawn` is `tokio::task::spawn_local` (actix-rt 2.11.0, lib.rs:214), not `tokio::spawn`. It has to be: the pump owns the `!Send` payload, so a `Send`-bound spawn would not compile. The comment stands.

foxpatch-aleph (changes requested):
The four fixes are real and correctly implemented — I verified against the actix-http 3.13.1 / actix-web 4.14.0 sources that the h1 encoder picks its transfer encoding from the body size before encode_headers consults the status (so a stream body on a 204/304 wrote a stray 0\r\n\r\n after a head with no framing), that streaming() injects application/octet-stream while body() does not, and that the dispatcher already logs body-stream errors, making the map_err removal correct. The send_bridge pump is a sound !Send-to-Send bridge with real backpressure, and the wire-level tests (raw sockets against a real actix server) are well built. However, the request-body path has a genuine regression: whether a body is attached is now decided purely from the client's Content-Length/Transfer-Encoding headers, which is complete for HTTP/1.1 but not for HTTP/2, which this server serves (bind_rustls_0_23 advertises h2 via ALPN). An h2 request body is framed by DATA frames and may legally carry no content-length header (RFC 9113 8.6) — exactly what a streaming upload emits — so such a request gets no body attached and its payload is silently dropped, where the pre-PR web::Bytes path forwarded it. All wire tests speak HTTP/1.1, so none would catch this.

**`rust/crates/aleph-attest-agent/src/proxy.rs`** (line 297): This decision is complete for HTTP/1.1 (a request body must carry Content-Length or Transfer-Encoding there) but not for HTTP/2: the agent's rustls listener advertises h2 via ALPN, and an h2 request body is delimited by DATA frames with no content-length header at all (RFC 9113 8.6). For an h2 POST without that header, chunked is false and content_length is None, so no body is attached and the web::Payload is dropped unread: the upstream receives a bodiless request and the body is silently lost — a regression against the pre-PR web::Bytes path, which forwarded it. Consider also attaching the bridge stream when the headers don't frame the request but the transport can still carry a body (e.g. req.version() == Version::HTTP_2, where hyper will frame it as chunked toward the h1 upstream), and add an h2-path test; none of the current wire tests exercise an h2 client.

**`rust/crates/aleph-attest-agent/src/proxy.rs`** (line 336): Minor, non-blocking: RFC 7232 4.1 says a 304 should retain a Content-Length header, and actix's encoder explicitly supports that (its encode_headers keeps an app-supplied CL for NOT_MODIFIED, unlike NO_CONTENT). Since the header-forwarding loop strips content-length for every status and finish() cannot re-add one, a 304 with content-length arrives bare. Browsers tolerate this, so it is only worth handling if you want the 304 path spec-exact.

**`rust/crates/aleph-attest-agent/src/proxy.rs`** (line 22): Pre-existing, not introduced by this PR: RFC 7230 6.1 also requires dropping headers named in the Connection header (a Connection: X-Foo means X-Foo is hop-by-hop), which the fixed list misses. Worth a follow-up or a comment noting the limitation; not blocking.

odesenfans:
Second review follow-up:

- **HTTP/2 bodies without Content-Length** (442d438a): confirmed, actix-http prepends `h2` to the ALPN list on every rustls bind (service.rs:646), so the agent does serve h2 and such a body was dropped. Fixed differently from the suggestion: attaching the stream on `Version::HTTP_2` alone would send every h2 GET upstream with `transfer-encoding: chunked` and an empty body, since no h2 GET carries a content-length. Instead, when the headers frame nothing, the handler polls the payload once: it ends at once for a bodiless request (always on h1, on END_STREAM for h2), otherwise the first chunk is relayed ahead of the rest and the body goes upstream chunked. `http2_bodies_without_a_content_length_reach_the_upstream` covers both halves with a prior-knowledge h2 client against an h2c listener. Golden file re-seeded.
- **304 Content-Length**: not taken here. RFC 7232 4.1 is a SHOULD, clients tolerate its absence, and it is unrelated to body framing.
- **Headers named in Connection**: taken, as its own PR stacked on this one: #1278.

foxpatch-aleph (approved):
This is a careful, correct fix and every load-bearing claim checks out against the pinned dependency sources. I verified in actix-http 3.13.1 that the h1 encoder really does pick chunked TE from the body stream before the 204 status suppresses the TE header, then writes a stray 0\r\n\r\n on the keep-alive connection; the finish() path produces TransferEncoding::empty() and no terminator. I verified in actix-web 4.14.0 that streaming() inserts application/octet-stream, that no_chunking(len) writes the Content-Length header, and that SizedStream/BodyStream accept reqwest::Error. I verified in hyper 1.10.1 that a user-set content-length with a streaming body takes the length encoder (closing the connection on early truncation), that a user-set transfer-encoding takes the chunked encoder, and that hyper strips (rather than errors on) the TE header for h2 upstreams. The request-framing logic is smuggling-safe because headers are re-derived, not relayed, and actix rejects any TE other than exactly "chunked" plus TE+CL combos, so the handler's assumptions hold. The removed map_err log is fine: actix-http's dispatcher already logs body-stream errors at error level. The tests are strong — on-the-wire assertions for the desync, keep-alive reuse, h2 request bodies, and the 1 MiB uncapped upload, all deterministic. Remaining findings are minor: one stale doc line, a deliberate-but-worth-noting behavior change for bodiless requests, and two small consistency notes.

**`docs/architecture/confidential.md`** (line 183): Stale after this PR: the doc says the proxy "strips hop-by-hop headers and `Content-Length` before forwarding", but the proxy now re-derives and forwards Content-Length on requests (and preserves it on responses) instead of stripping it. Worth updating in the same PR since the framing behavior is the point of the change.

**`rust/crates/aleph-attest-agent/src/proxy.rs`** (line 298): Behavior change worth being aware of (not a bug): the old `body(vec![])` made hyper set `content-length: 0` on every request; now a request with no framing headers at all goes out with no Content-Length. RFC 9110-legal (absent CL + absent TE = empty body), and clearly better for GET/HEAD, but a strict upstream that answers 411 to a POST without a Content-Length would regress. The empty-body and h2 GET tests pin the new behavior deliberately; if 411-compatible servers are a concern, synthesizing `content-length: 0` for h1 POST/PUT/PATCH requests without framing would be the follow-up.

**`rust/crates/aleph-attest-agent/src/proxy.rs`** (line 306): Note: if `--upstream` is an https URL where reqwest negotiates h2, hyper strips this `transfer-encoding` header from the request with a `warn!` per request ("Connection header illegal in HTTP/2", src/proto/h2/mod.rs). The body still arrives correctly via DATA frames, so this is only log noise, but it's worth knowing if an https upstream ever appears.

**`rust/crates/aleph-attest-agent/src/proxy.rs`** (line 342): Minor consistency nit: the request side uses the `CONTENT_LENGTH` constant while the response side uses the string literal here. `HeaderMap::get` is case-insensitive so it works; using the constant in both places would be tidier.

**`rust/crates/aleph-attest-agent/src/proxy.rs`** (line 339): This Content-Length passthrough silently depends on reqwest's decompression staying disabled (the manifest pins `default-features = false` with no gzip/brotli features). If someone later enables compression, the upstream CL would no longer match the decoded bytes and this path would desync clients. One line here — or next to the reqwest dependency — stating that reqwest's compression features must stay off for CL relay would prevent that trap.

odesenfans:
Third round:

- **Stale doc line**: fixed in the commit above (`docs/architecture/confidential.md`).
- **"the old `body(vec![])` made hyper set `content-length: 0` on every request"**: it did not. hyper 1.10.1 treats an empty body as already ended (`proto/h1/dispatch.rs:360`, `body.is_end_stream()` gives `body_type = None`), and `Client::set_length` returns `Encoder::length(0)` for `None` without adding a header; reqwest adds none either. The old path stripped the client's Content-Length and sent nothing for an empty body, so there is no regression for 411-style upstreams. The new code is the stricter one: a client's explicit `Content-Length: 0` now reaches the upstream, pinned by `an_empty_request_body_keeps_its_content_length`.
- **h2 upstream warning**: noted; the upstream is loopback plain HTTP, so h2 is never negotiated.
- **Header-name constant and the decompression caveat**: both taken, in #1278, which has to re-seed the golden file after this merges anyway (they touch the measured crate; doing them here would cost a re-seed on both PRs).

foxpatch-aleph (approved):
Solid, well-researched follow-up that I verified against the vendored actix-http 3.13.1, actix-web 4.14.0, hyper 1.10.1 and reqwest 0.12.28 sources: every library-behavior claim checks out. The 204/304 stray-terminator mechanism (the h1 encoder selects chunked from the stream body before the status makes it skip the Transfer-Encoding header, then still writes 0\r\n\r\n) is exact, and routing those statuses to finish() is the right fix; {204, 304} also precisely matches actix-http's own bodiless framing set, so no other status can desync. streaming() really does invent application/octet-stream, actix-http really does log response-body errors at error level itself (h1 and h2 dispatchers), web::Payload really has no extractor cap, and hyper honors user-set content-length and transfer-encoding: chunked on requests with wrap_stream bodies — each pinned by an on-the-wire test. Security posture is maintained: lengths are re-derived, never relayed verbatim, so CL+TE can never cross together, hop-by-hop stripping is retained, and memory stays bounded by the capacity-1 pump. The tests exercise the right behaviors over real sockets (keep-alive desync, sized framing, 1 MiB uncapped upload, chunked/h2/empty request bodies, mid-body abort). The documented leftover (client disconnect not promptly releasing the upstream, needs an idle timeout on the reqwest client) is correctly deferred. I could not run the suite locally (no cargo available), but static verification found no bugs.

**`rust/crates/aleph-attest-agent/src/proxy.rs`** (line 347): Minor, non-blocking: actix-http's encoder notes that a 304 should retain a manually set content-length (actix-http src/h1/encoder.rs, NOT_MODIFIED arm: "304 responses ... should retain a manually set content-length header"), and RFC 7232 says a 304 SHOULD carry the same header fields as the 200 would have. Here the upstream's content-length is stripped in the forwarding loop and finish() adds none, so a 304 loses its length. In practice clients ignore content-length on 304 and the "never copy a Content-Length, only re-derive" invariant is easier to audit, so this is defensible — just flagging the one place where it diverges from what actix would do natively.

**`rust/crates/aleph-attest-agent/src/proxy.rs`** (line 233): Worth a note for future maintainers: actix_web::rt::spawn is LocalSet-based, so this pump only works inside an actix runtime — #[actix_web::main] in production and #[actix_web::test] in the tests both provide one (verified), but the handler would panic if it were ever driven under a plain #[tokio::main]. The doc comment on send_bridge could mention that runtime requirement, but it's fine as is.

PR #1267, opened 2026-09-17
fix(agent): strip the measured binary

odesenfans:
Every source edit under the agent's nix source filter moved the launch measurement, even doc-only or verify-gated ones (seen on #1259 and #1260). Root cause: nix strips debug info only, so the agent shipped with its symbol table, and ThinLTO names promoted locals after a module hash that changes with any edit to the crate's sources. The code itself was unchanged: the binaries built before and after such an edit are byte-identical once stripped, and with `strip = true` in the release profile they are byte-identical as built.

- `[profile.release] strip = true` on the agent workspace.
- Golden measurements re-seeded (also 3 MB off the initrd).

Stacked on #1260 so the two golden re-seeds do not collide. The fib-service workload binary has the same class of noise; left alone since its source almost never changes.


foxpatch-aleph (approved):
Small, correct, and well-documented fix. The [profile.release] strip = true is placed in the agent's own workspace root (the crate declares [workspace] at line 19), so cargo actually honors it — a profile in a member crate would have been silently ignored. The root-cause analysis is technically sound: nixpkgs strips debug info only (-S), the static-musl agent kept its symbol table, and thin-local ThinLTO promotes internal symbols with names derived from the crate metadata hash, which moves with any source edit; stripping the symtab at link time removes that noise while the later nix strip pass stays a deterministic no-op. All five golden measurements moving is expected since the agent is baked into every initrd flavor, and the old values are pinned nowhere else. The comment in Cargo.toml captures the why succinctly. Only non-blocking note: stripped symbols mean anonymous backtraces for in-guest crashes — an acceptable trade for a measured TCB binary.

**`rust/crates/aleph-attest-agent/Cargo.toml`** (line 50): Nit: this profile is only honored because this crate is its own workspace root ([workspace] on line 19) — cargo silently ignores profiles in member manifests. A future reader relocating or splitting the workspace could drop the strip without noticing until the golden check fails. The own-workspace setup is already documented above, so no change needed; just noting the coupling. Also worth being aware: strip = true means anonymous backtraces for in-guest crash debugging — the right trade for a measured TCB binary, but worth remembering.

PR #1268, opened 2026-09-17
feat(agent): advertise per-pool free space on /about/usage/system

odesenfans:
## Problem

A node advertises one aggregate free-disk figure. Since multi-disk volume pools landed (#1049), that figure no longer describes what a single volume can use: a node with two 1 TB disks advertises about 2 TB free, but no volume ever spans pools.

The scheduler places VMs against that aggregate, so it sends a VM whose single 1.5 TB volume fits neither disk. `CapacityManager._check_max_volume` then correctly refuses it, and because the scheduler only logs refusals, the VM stays assigned to that node and is never re-placed. It is a silent, permanent stall.

The node has the right information. It just never publishes it.

## Change

`/about/usage/system` gains an optional `disk.pools` array:

```json
"disk": {
  "total_kB": 2000000000,
  "available_kB": 1900000000,
  "pools": [{"available_kB": 950000000}, {"available_kB": 950000000}]
}
```

Each entry is that pool's free bytes plus its reclaimable bytes, which is exactly the figure `_check_max_volume` judges a volume against. What a node advertises therefore cannot promise what a create would refuse.

Two details worth knowing:

- **The list is deduplicated by `st_dev`.** `eligible_pool_free_bytes` does not dedup, because its only other caller takes a `max()` and duplicates are harmless there. They are not harmless to a consumer packing volumes across the list: two pool directories on one filesystem would advertise twice the room that exists.
- **`sum(pools)` does not equal `available_kB`, by design.** The aggregate is the supervisor's usage-aware figure; these are what each pool would accept. Neither may be derived from the other.

Purely additive. Absent means "this node predates the field"; `[]` means "no room". Consumers that ignore it are unaffected.

## Consumer

Paired with the aleph-vm-scheduler side, linked in a follow-up comment.

Safe to merge in either order. The scheduler models a node without `pools` as one pool holding everything, which reproduces today's behaviour exactly.

## Testing

120 passed, up from a 114 baseline. New tests cover the `st_dev` dedup, reclaimable inclusion, an unreachable pool contributing nothing, the empty-list case, and the serialised wire spelling that the scheduler parses.


odesenfans:
Consumer side: https://github.com/aleph-im/aleph-vm-scheduler/pull/214

Safe to merge in either order. The scheduler treats a node without `pools` as one pool holding everything, so its behaviour is unchanged until this ships.

foxpatch-aleph (approved):
Correct, well-documented and well-tested. The per-pool figure (free + reclaimable_bytes) matches exactly what _check_max_volume admits against, the st_dev dedup mirrors pools_disk_usage and is the right semantics for a list a scheduler packs volumes across, unreachable pools drop out conservatively, and the wire spelling (None omitted via exclude_none, [] meaning no room) is coherent and locked in by a test. All 12 tests in the file pass. Two non-blocking nits below: a shared-filesystem edge case in the dedup and a docs paragraph that now lags the code.

**`src/aleph/vm/agent/resources.py`** (line 113): Minor, non-blocking: when two pool directories share one filesystem, the dedup credits only the first pool's reclaimable_bytes, while _check_max_volume (which does not dedup) takes max() over both pools' free + reclaimable. Advertisement can therefore under-state admission by the difference in reclaimable bytes between the two pools on that filesystem. This is the safe direction (scheduler routes the VM elsewhere; no stall), and it only bites the odd config where two pool dirs share a filesystem with asymmetric reclaimable. Worth a sentence in the docstring at most.

**`docs/architecture/storage.md`** (line 129): The 'three disk figures the agent produces... have to agree' paragraph predates this PR's fourth, per-pool advertisement (disk.pools on /about/usage/system). A one-sentence addition noting that the per-pool list follows the same free-plus-reclaimable rule (and is st_dev-deduplicated) would keep this doc authoritative for the next reader.

odesenfans:
Follow-up issues opened while reviewing this change:

- #1269 - cache filesystem capacity is neither measured, advertised nor admitted against. Same contiguity problem as this PR fixes, one layer over: immutable volumes and runtime bundles land in `CACHE_ROOT`, which is deliberately separate storage from the pools.
- #1270 - node admission under-counts those same files in `disk_mib`.
- aleph-im/aleph-vm-scheduler#215 - refused allocations are only logged.

PR #1271, opened 2026-09-19
fix(nix): confidential-GPU guest fixes from the first H200 run

odesenfans:
First run of the confidential-GPU image on real hardware (H200 NVL, dual EPYC 9224, SEV-SNP, QEMU 10.2.4, kernel 7.0 host). Three things broke in order; this fixes them.

## Fixes

1. **Guest kernel crypto.** The NVIDIA open driver builds its libspdm glue against the kernel crypto API only when `CRYPTO_{AEAD,AKCIPHER,SKCIPHER,HASH,HMAC,ECDH,ECDSA,RSA}` and `X509_CERTIFICATE_PARSER` are all set (`internal_crypt_lib.h`), and compiles stubs otherwise. With our whitelist kernel, GSP init failed in CC mode: `libspdm_check_crypto_backend: Error - libspdm expects LKCA but found stubs!`. The GPU fragment now carries the missing options (plus `GCM` and `SHA512`, used at runtime).
2. **Persistence mode.** In CC mode the adapter initialises once per GPU reset. `nvattest` was the first client, so when it exited the adapter tore down and every later client failed (`osInitNvMapping: *** Cannot attach gpu`), which would have broken the ready-state step and the workload. `nvidia-smi -pm 1` now runs first. Verified in the guest: `CC status: ON`, `CPU CC capabilities: AMD SEV-SNP`, ready state `not-ready` then `ready` after `-srs 1` (the wording the existing checks expect).
3. **Operator doc.** QEMU refuses the card on an SNP guest with "possibly running out of DMA mappings ... Maximum possible DMA mappings: 65535" unless `vfio_iommu_type1 dma_entry_limit` is raised.

The GPU golden measurement moves (kernel and initrd both change); re-seeded from a sandboxed nix build of this branch.

## Not fixed here

Attestation on the test card still fails closed with `RIM Not Found`: its VBIOS (96.00.B7.00.02) has no RIM published by NVIDIA (only 96.00.CE.00.0F, 96.00.D9.00.0E and 96.00.DB.00.11 exist for this board). That needs a VBIOS update, not a code change. A Hopper entry in `gpu.json` (`accepted_models`) follows once the hwmodel claim can be read.

## Update 2026-09-20: end to end on hardware

The branch grew while driving a full run. Beyond the three guest fixes above:

- **Runtime manifest**: the `gpu` block is keyed by architecture (`archs: {hopper: {accepted_models: ["GH100 A01 GSP BROM"]}, blackwell: {...}}`); one measured image serves both.
- **`cuda-probe` workload** (`nix/cuda-probe`, `nix/cuda-workload.nix`): CUDA driver API through `dlopen`, an embedded PTX SAXPY kernel checked exactly, and a bandwidth round trip. Two contract facts it surfaced, now documented: a workload's `/sbin/init` must be a regular file (the init checks it before chrooting), and CC-mode libcuda needs OpenSSL 3 (`libnvidia-pkcs11-openssl3.so`), which the workload brings.
- **Agent**: `/about/usage/system` always carries a GPU `model`. For a card outside the network's whitelist the key was dropped, the scheduler could not decode the response (`missing field model`) and treated the node as unhealthy for every workload.
- **Daemon**: the OVMF MMIO window counted the SR-IOV VF BARs (sysfs rows 7 to 12) and doubled a 512 GiB rounding to 1 TiB, which cannot fit a 40-bit guest. It now counts BAR rows 0 to 5 and falls back to the un-doubled window when only that fits.
- **Docs**: Hopper cards, the VBIOS RIM check, the host driver blacklist, failure signatures, what a GPU workload must ship.

Result: aleph-testnets run 35485093472 (aleph-im/aleph-testnets#43), 33 tests, 0 failures. A CUDA V-PROGRAM created with `--gpu hopper` (aleph-im/aleph-rs#402) was scheduled onto the H200 NVL CRN, launched by the daemon, passed NVIDIA's verifier in the guest (`GPU verified and ready`), and answered attested `/gpu` and `/bandwidth` calls; a `--gpu blackwell` request stayed unscheduled. Measured on the card: about 4 GiB/s host to device, 0.6 GiB/s device to host, single copies up to 4 GiB with `swiotlb=262144`.

Not covered yet: client-side verification of the GPU evidence (the CLI verifies the SEV-SNP side only until `attest::nvidia` lands), multi-GPU guests (two H200s need more than 40 guest physical bits).

## Update 2026-09-20 (2): the GPU requirement is part of the launch measurement

Until now the measured guest verified every GPU it saw but had no idea what the user asked for: with no card attached the GPU runtime booted as a plain guest under the same measurement, nothing in the guest compared count, arch or model to the request, and the message's `models` are PCI ids, which the host controls. This update closes that.

- **Tokens.** The kernel command line (covered by the launch measurement) carries `gpu_arch=<hopper|blackwell> gpu_count=<1..8>` and, when the message pins models, `gpu_models=<sorted, unique, lowercase ids>` (the whole token is dropped otherwise, like `verified_volumes`). The aleph-rs client renders the same bytes when it computes the measurement (aleph-rs PR 403).
- **Guest.** `init-gpu.sh` refuses to boot without a GPU, collects the evidence once, has nvattest verify that same file with the boot nonce, then runs the agent's new `gpu-policy` subcommand before the ready state is set. It enforces: tokens present and canonical, PCI scan count = evidence count = claims count = `gpu_count`, every evidence blob answers this boot's nonce, no GPU answers twice, arch (evidence arch, nvattest's arch check, certificate model in the per-arch list), and for a model pin the board identity the GPU signs in its SPDM opaque data (`PROJECT/PROJECT_SKU/CHIP_SKU`, the same triple that names the VBIOS RIM) against a `boards` table in the measured `/etc/aleph/gpu.json`. Every failure powers the VM off.
- **Host.** The launcher stages a `{rootfs}.gpu_requirement` sidecar; the daemon validates it against a closed grammar and appends it after the verified volumes. The manifest model gains `boards` and validates slot order.
- **Board table.** Six rows ship: H200 NVL (read from real evidence), H100 PCIe, H100 NVL, H100 SXM5 80GB, H200 SXM5, RTX PRO 6000 Blackwell SE (from NVIDIA part numbers, which encode the same project and SKU). A missing row only fails closed for a user who pins that model.

**Validated on the H200 node**, by hand: correct request boots (with and without `gpu_models=10de:233b`); `gpu_count=2`, `gpu_arch=blackwell`, `gpu_models=10de:2331` (an H100 id), no tokens, no card attached, and a mismatched attest nonce each power off with a specific reason. Golden measurements re-seeded (CI and the node agree on all five). Blackwell and multi-GPU requirements are accepted by the format but have not been exercised on hardware.


foxpatch-aleph (approved):
Approving #1271 as a well-constructed conclusion to the confidential-GPU stack. The GPU requirement is rendered in one place (render_gpu_requirement) whose canonical output — arch, single-digit 1..8 count, sorted unique lowercase PCI ids — is exactly what the Rust daemon's closed grammar (is_canonical_gpu_requirement) accepts and what the guest's gpu-policy parser enforces, so the measurement, the hypervisor splice and the guest verdict stay in three-way byte parity. The daemon reads the sidecar, validates the trimmed value, and splices that same value, closing both the injection and TOCTOU concerns; nothing but the exact canonical form can reach -append. The manifest schema now pins each placeholder to its one legal key and enforces the canonical slot order, so tokens like evil{gpu_arch} or rdinit={gpu_arch} are rejected and the substring-based runtime detection stays sound. Guest init is properly fail-closed (empty PCI bus fatal, gpu-policy failure powers off before the workload starts, verdict bound to the boot nonce), and board identity is grounded in signed SPDM evidence rather than PCI config space. The board table is consistent across flake.nix gpuFacts, the operator docs, and the test fixtures, and the gpuFacts file being the single source for both the verity rootfs and the manifest prevents the two from drifting. Test coverage is excellent: canonical-form acceptance, 17 rejection vectors, no-sidecar byte parity, stale-sidecar removal, and manifest validation all pass locally (55 + 61 + 21). The two comments below are non-blocking nits.

**`nix/init-gpu.sh`** (line 360): Non-blocking nit: gpu-policy.log is the only log under /run/aleph not created with the umask 077 discipline used for the neighbouring evidence and claims files (e.g. line 317). It contains no secrets — the verdict is echoed to the console right below anyway — but a one-line umask wrapper would keep the pattern uniform for whoever copies it next time. Same applies to the /run/aleph/tmp listing at line 296.

**`src/aleph/vm/agent/vprogram_launch.py`** (line 251): Nit: `validated = render_gpu_requirement(...)` can raise TypeError rather than ValueError if a non-string sneaks into models — unreachable through the Pydantic message parsing that already enforces CONFIDENTIAL_GPU_DEVICE_ID_PATTERN upstream, so the call is never reached with garbage in practice. The bare-msg format here strips the exception type either way, so the failure message will just be a little terse in that case.

